PR #191, opened 2016-03-09
Gitter notifications

sedubois:
A row needs to be manually entered in the`Notifications` table for the notifications to be fired. `channel` should be `gitter`, `type` can be an activity type or `all`, and `webhookUrl` needs to be copied manually from what the user obtains on the Gitter integrations page when selecting Open Collective. That will require the merge of a little Gitter PR that will follow.

I note that email and (internal) slack notifications are handled in some other places in the code, it might be helpful to integrate with this code (if deemed sufficiently important).

A little extra PR should allow slack webhooks as well, I can look at that when this is merged.

### DIFF
--- a/app/lib/notifications.js
+++ b/app/lib/notifications.js
@@ -4,7 +4,6 @@ const activitiesLib = require('../lib/activities');
 const ACTIVITY_ALL = 'all';
 
 module.exports = (Sequelize, activity) => {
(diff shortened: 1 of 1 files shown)


sedubois:
Tested sending a message to this channel, goes through properly: https://gitter.im/sedubois/opencollective_test

Documented at https://github.com/OpenCollective/OpenCollective/wiki/Slack-and-Gitter

PR #234, opened 2016-03-25
Update payments flow to create a Donation entry

asood123:
Still need to add a couple more hooks (like creating a subscription) and need to update tests.

### DIFF
--- a/test/donations.routes.test.js
+++ b/test/donations.routes.test.js
@@ -929,14 +929,12 @@ describe('donations.routes.test.js', () => {
             })
             .end((err, res) => {
               expect(err).to.not.exist;
-              console.log("res.body: ", res.body);
               links = res.body.links;
               done();
             });
         });
 
         it('creates a transaction and returns the links', (done) => {
-          console.log('links: ', links);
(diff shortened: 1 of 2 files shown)


test/donations.routes.test.js
```diff
@@ -29,7 +30,7 @@ const stripeMock = require('./mocks/stripe');
 /**
  * Tests.
  */
-describe('payments.routes.test.js', () => {
+describe('donations.routes.test.js', () => {
```
sedubois: Not sure why this test still calls `.post('/groups/' + group.id + '/payments')` (line 206) which is supposed to be a deprecated route cfr. comment https://github.com/OpenCollective/api/pull/234/files#diff-6db976b763ed71154781e45807c0d4e5R127

asood123: I didn't update the old routes in tests intentionally since website was
still using them. Don't want to break them unintentionally.

sedubois: So it's in your hands to update the website side?

asood123: Yes feel free to leave it as is. I had a trello task for it. Will move it
to GitHub.

PR #262, opened 2016-04-05
Run e2e tests during circleci build of api

sedubois:
This branch enables e2e tests to be executed fully locally from the api repo on circleci. Additionally, it enables developers to do the same locally. Also, if `WEBSITE_DIR` and `APP_DIR` are added to their `.env`, the script will reuse these checkouts instead of performing a fresh one. 

I also attempted to cache the website and app node_module subdirectories but for some reason it's refused by circleci cfr. "Save cache" phase:

````
  ✗ /home/ubuntu/api/.app_node_modules
  ✗ /home/ubuntu/api/.website_node_modules
````

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -8,13 +8,23 @@ machine:
   post:
     - npm install jscs@1.13.1 -g
     - npm install mocha@2.4.5 -g
+    - cd api && npm run test:e2e website install
+#    - cd api && npm run test:e2e app install
   node:
     version: 5.1.0
   services:
     - postgresql
+dependencies:
+  cache_directories:
+    - "~/cache"
+  pre:
+    - npm run install:debs
 test:
   override:
     - npm run circleci
+  post:
+    - npm run test:e2e website run
+#    - npm run test:e2e app run
 deployment:
   staging:
     branch: staging
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "circleci": "NODE_ENV=circleci npm run hint && mocha --reporter mocha-circleci-reporter --reporter-options mochaFile=$CIRCLE_TEST_REPORTS/junit/junit.xml test",
     "start": "npm run hint && node .",
     "dev": "nodemon -e js,hbs -w app",
-    "hint": "eslint '**/*.js' --quiet",
+    "hint": "eslint 'app/**/*.js' --quiet",
     "db:migrate": "sequelize db:migrate --config config/sequelize_cli.json --models-path app/models/ --env  $SEQUELIZE_ENV",
     "db:migrate:undo": "sequelize db:migrate:undo --config config/sequelize_cli.json --models-path app/models/ --env  $SEQUELIZE_ENV",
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: circle.yml [hunk 1]
```diff
@@ -6,15 +6,23 @@ general:
       - production
 machine:
   post:
-    - npm install jscs@1.13.1 -g
-    - npm install mocha@2.4.5 -g
+    - npm install jscs@1.13.1 mocha@2.4.5 -g
+    - cd api && npm run test:e2e website install
+    - cd api && npm run test:e2e app install
```
xdamman: you may want to add

    machine:
      environment:
        NODE_ENV: test
      node:
        version: 5.1.0

circle.yml
```diff
@@ -6,15 +6,23 @@ general:
       - production
 machine:
   post:
-    - npm install jscs@1.13.1 -g
-    - npm install mocha@2.4.5 -g
+    - npm install jscs@1.13.1 mocha@2.4.5 -g
+    - cd api && npm run test:e2e website install
```
xdamman: I'm not sure I like `e2e` it's cryptic and not very self explanatory. We use `frontend` in the `website` repo. How do you feel about that wording instead? (or any other proposition?)

package.json
```diff
@@ -70,7 +70,7 @@
     "circleci": "NODE_ENV=circleci npm run hint && mocha --reporter mocha-circleci-reporter --reporter-options mochaFile=$CIRCLE_TEST_REPORTS/junit/junit.xml test",
     "start": "npm run hint && node .",
     "dev": "nodemon -e js,hbs -w app",
-    "hint": "eslint '**/*.js' --quiet",
+    "hint": "eslint 'app/**/*.js' --quiet",
```
xdamman: wouldn't that break locally? We cannot make any assumption on where `app` and `api` are locally in development environment (`app` may not even be there if a developer just wants to change and test `website`)

sedubois: That `app` refers to the `app` directory within the `api` repo, nothing to do with the `app` repo ;)

circle.yml
sedubois: `frontend` in my view is synonymous for "client", e.g website and app are the front-ends, API is the backend. So `test:frontend` would mean running all client tests, including unit tests, no? In my view nightwatch tests are not "front-end" tests because they cover both the front-end and back-end, i.e. they are end-to-end. But I'm the new one in web development so please correct me...

Nightwatch's doc seems to say the same: "Nightwatch.js is an easy to use Node.js based End-to-End (E2E) testing solution for browser based apps and websites." http://nightwatchjs.org/

I can change to `nightwatch` otherwise? I just prefer to avoid technology names where possible.

package.json
xdamman: ah yes, we should rename this to `server`, see the guideline on https://github.com/OpenCollective/OpenCollective/tree/master/dev/default_folder_structure

(now is a good time to harmonize all of this across repos)

circle.yml
xdamman: Alright then, you got me. That means we should rename `test/frontend` to `test/e2e` on the website repo.

PR #263, opened 2016-04-05
added burnate and mission column to Groups, plus some housekeeping

xdamman:


sedubois:
What's this burnrate? Something entered purely manually in DB?

xdamman:
Yeah, it's for the new design. I don't think it makes sense to compute it in real time. It's quite expensive for limited benefit. It would be better to have a cron job that updates that values every so often (outside of the scope of this PR)

sedubois:
I'm not a fan of storing derived data in database, it's logically separate. Storing derived data like the original data can lead to issues:
- ends up blurring the fundamental data and their interrelationships
- allows original and derived data to be incoherent (like in this PR)
- derived data is not real-time
- database backups and migrations are bigger/more complex

There are other ways to compute and store derived data, e.g views, materialised views, triggers, query caching, stored procedures, temporary tables...
http://stackoverflow.com/questions/8250389/computed-calculated-columns-in-postgresql
http://www.postgresql.org/about/news/1296/

That being said I agree that performance should not be optimised before it is actually necessary to do so, as there are more pressing matters...

asood123:
I agree. There are better ways to store computed info.

xdamman:
I don't disagree.  
I was more looking for a quick fix at this stage.

How should we handle those situations? I think it's valuable and welcome to have people do quick fixes to move fast. But we should find a place to keep track of the proper way it should have been done.

We could annotate the code with comments (and use a special hashtag or something to easily parse the code in the future?), or we should create a github issue?

asood123:
Worth thinking about property of quickfixes (like check for transaction in
last 10 mins):
- they save significant development time right now
- is likely an inefficient implementation
- creates minimal ongoing maintenance work. This is important. It's ok to
create manual setup work like in case of Tiers but painful if we are
writing scripts to maintain our quick fix or doing a ton of manual work
like in case of fees, it quickly becomes a net negative.
- Note: quick fix doesn't mean skipping tests. If anything, tests are even
more useful for quick fixes.

Let's tag the code with and we can file a github issue for it once it's
working (and we know that it's needed):
```
// #quickfix: [explanation with right way to fix it]
```

For other functionality, we should only build for 10x-100x scale of where
we are now. These are things where we are certain that a
feature/functionality is working and we need to build it to scale to the
next level. New public page falls in this category. Of course, taking 8
weeks is far too long for it.

PR #272, opened 2016-04-12
For ImagePicker component

carlosascari:
Created endpoint that uses clearbit to lookup and return possible avatars for a new user.

Modified users controller: updateUserWithoutLoggedIn function so as to update avatar field from donation page.

Removed auth middleware from /images endpoint, so as to allow img uploads from donation page

NOTE: Currently `image-uploader-picker-component-for-avatar` branch from website repo does not use the endpoint created.


asood123:
FYI, I'm debugging the tests manually on circleci. Don't merge this even if tests are passing.

asood123:
Added test fix here: https://github.com/OpenCollective/api/pull/276

PR #275, opened 2016-04-14
Slack catch fix

asood123:


asood123:
@xdamman looked at this. Merging.

PR #279, opened 2016-04-16
added referrerId to the Users model

xdamman:


xdamman:
See https://github.com/OpenCollective/OpenCollective/issues/23

PR #280, opened 2016-04-16
making sure we don't charge an application fee on development or staging

xdamman:
Otherwise we have this error:

```Can only apply an 'application_fee' when the request is made on behalf of another account (using an OAuth key or the Stripe-Account header).``

### DIFF
--- a/app/controllers/donations.js
+++ b/app/controllers/donations.js
@@ -140,6 +140,10 @@ module.exports = (app) => {
                 }
               };
 
+              if (process.env.NODE_ENV === 'production') {
+                charge.application_fee_percent = constants.OC_FEE_PERCENT;
+              }
+
               return gateways.stripe.createSubscription(
                 results.getGroupStripeAccount,
                 paymentMethod.customerId,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: app/controllers/donations.js [hunk 1]
```diff
@@ -140,6 +140,10 @@ module.exports = (app) => {
                 }
               };
 
+              if (process.env.NODE_ENV === 'production') {
+                charge.application_fee_percent = constants.OC_FEE_PERCENT;
```
asood123: This isn't going to work. Charge is declared below.

Also, this means that our platformFee amount will never get tested in non-production. It's going to break other tests as well. 

I think a better way might be to create a new entry in local .env file:

```
STRIPE_SANITIZE_KEY: "xxxxx"
```
You can get this key by authorizing stripe on your local machine. Then, update the sanitize script to use this key.

xdamman: well, but we need a different stripe account otherwise we have the issue `Can only apply an 'application_fee' when the request is made on behalf of another account (using an OAuth key or the Stripe-Account header)`

xdamman:
@asood123 what do you recommend here? Because creating a new Stripe key won't help since it will have to be linked to a difference Stripe account altogether.

PR #281, opened 2016-04-16
Show accurate balance

asood123:
- Creates a new field called `netAmountInGroupCurrency` in `Transactions`
- Runs migration for it
- Populates that field on every new `Transaction`
- Includes a script that needs to be run in staging and production

Note: if you try this right now, it'll be inaccurate in staging and production. We'll need to manually delete the rows for fees that we added for Feb.

### DIFF
--- a/app/controllers/groups.js
+++ b/app/controllers/groups.js
@@ -110,18 +110,20 @@ module.exports = function(app) {
   };
 
   const getBalance = (id, cb) => {
-    sequelize.query(`
-      select
-        CAST(SUM("netAmountInGroupCurrency") AS FLOAT)/100 AS total
-        FROM "Transactions"
-        WHERE "GroupId"=${id} AND approved = true`,
-    {
-      type: sequelize.QueryTypes.SELECT
-    })
-    .then((result) => {
-      return cb(null, result[0].total)
-    })
-    .catch(cb);
+    Transaction
+      .find({
+        attributes: [
+          [sequelize.fn('SUM', sequelize.col('netAmountInGroupCurrency')), 'total']
+        ],
+        where: {
+          GroupId: id,
+            approved: true
+          }
+        })
+        .then((result) => {
+          return cb(null, result.toJSON().total/100);
+        })
+        .catch(cb);
   };
 
   const getPublicPageInfo = (id, cb) => {
--- a/app/controllers/groups.js
+++ b/app/controllers/groups.js
@@ -110,18 +110,18 @@ module.exports = function(app) {
   };
 
   const getBalance = (id, cb) => {
-    Transaction
-      .find({
-        attributes: [
-          [sequelize.fn('SUM', sequelize.col('amount')), 'total']
-        ],
-        where: {
-          GroupId: id,
-          approved: true
-        }
-      })
-      .then((result) => cb(null, result.toJSON().total))
-      .catch(cb);
+    sequelize.query(`
+      select
+        CAST(SUM("netAmountInGroupCurrency") AS FLOAT)/100 AS total
+        FROM "Transactions"
+        WHERE "GroupId"=${id} AND approved = true`,
+    {
+      type: sequelize.QueryTypes.SELECT
+    })
+    .then((result) => {
+      return cb(null, result[0].total)
+    })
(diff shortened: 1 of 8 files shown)


app/controllers/groups.js
```diff
@@ -110,18 +110,18 @@ module.exports = function(app) {
...
+      select
+        CAST(SUM("netAmountInGroupCurrency") AS FLOAT)/100 AS total
+        FROM "Transactions"
+        WHERE "GroupId"=${id} AND approved = true`,
+    {
+      type: sequelize.QueryTypes.SELECT
+    })
+    .then((result) => {
+      return cb(null, result[0].total)
+    })
```
Philmod: Why don't you use Sequelize query?

asood123: I couldn't get it to work with Sequelize. It's the division by 100 that's the problem. You can divide before the sum. I want to sum and then divide (so it's more accurate). Do you know how to do it?

Philmod: I guess you could divide in the callback?

asood123:
@Philmod switched out the query. That's better.

app/controllers/groups.js
```diff
@@ -113,15 +113,17 @@ module.exports = function(app) {
     Transaction
       .find({
         attributes: [
-          [sequelize.fn('SUM', sequelize.col('amount')), 'total']
+          [sequelize.fn('SUM', sequelize.col('netAmountInGroupCurrency')), 'total']
```
xdamman: Shouldn't the default be `inGroupCurrency`? In which case we shouldn't specify it?

app/controllers/transactions.js
```diff
@@ -35,6 +35,19 @@ module.exports = function(app) {
...
+      // populate netAmountInGroupCurrency for donations
+        transaction.netAmountInGroupCurrency =
+          Math.round((transaction.amountInTxnCurrency
+            - transaction.platformFeeInTxnCurrency
+            - transaction.hostFeeInTxnCurrency
+            - transaction.paymentProcessorFeeInTxnCurrency)
+          *transaction.txnCurrencyFxRate);
+    } else {
+      // populate netAmountInGroupCurrency for "Add Funds" and Expenses
+      transaction.netAmountInGroupCurrency = transaction.amount*100;
```
xdamman: so if there is no FxRate, you assume it's "add funds"? Is that always true? What about a donation without a FxRate?

xdamman: Since the host is likely in the same currency than the collective, it feels like we should first convert the transaction.amount to the right currency, and then apply the host fees. I also wonder if the paymentProcessor takes their fees after or before the FXRate.

app/models/Transaction.js
```diff
@@ -153,7 +154,8 @@ module.exports = function(Sequelize, DataTypes) {
           interval: this.interval,
           platformFee: this.platformFee,
           hostFee: this.hostFee,
-          paymentProcessingFee: this.paymentProcessingFee
+          paymentProcessingFee: this.paymentProcessingFee,
```
xdamman: above you have `paymentProcessorFee`, would be good to be consistent.

app/controllers/groups.js
asood123: It's a little tricky. There can be up to three currencies:
- DonationCurrency
- TxnCurrency
- GroupCurrency

Ideally, DonationCurrency and GroupCurrency is the same. We have had exceptions like wwcode_df page supports donations in USD and EUR (when Github wanted to donate $1000 USD to them). That meant that DonationCurrency != GroupCurrency.

Since there are two currencies in that table and we are going through migrations on renaming some stuff, I'd prefer to be clear on every amount which currency we are talking about until migration is done.

app/controllers/transactions.js
asood123: All donations have an fxRate. If the currencies are same, we populate a '1' for the fxRate.

Stripe only deals with fees (it's own and our platform fee) after fx-conversion. That's the reason, we have to store it this way.

app/models/Transaction.js
asood123: That's a bug. fixed.

PR #282, opened 2016-04-16
add hostFeePercent field to group

asood123:


app/models/Group.js
```diff
@@ -66,6 +67,11 @@ module.exports = function(Sequelize, DataTypes) {
       }
     },
 
+    hostFeePercent: {
+      type: DataTypes.FLOAT,
+      defaultValue: constants.HOST_FEE_PERCENT
```
xdamman: if it's a float, why do you divide it by 100 above?

asood123: I wanted to make it an INT value 0  to 100. Issue is that we won't be able to charge any non-integer percentage like 2.5% then. So, made it a float. With float, we could represent 5% as `5` or `0.05`. Currently, I'm representing 5% as `5`, 2.5% as `2.5`, which is easier for humans to understand. Otherwise, we'll be manually adding `0.05` or `0.10` or `0.025` and it's easy to be off by a decimal.

sedubois: In my view, % isn't a unit (unlike € or $). Percent literally means "* 1/100". So the numeric value for 5% is AFAIK always 0.05, not 5. Storing a value of 5 would be mixing up what the data is (0.05), with its representation ("5 %").

xdamman: That's a very fair point.

asood123: I don't disagree - it's cleaner if computers are the only ones dealing with it. We'll be manually editing it for the time being. One comparison point is that stripe also calls its field: application_fee_percent and it's an Int between 0 and 100.

PR #286, opened 2016-04-18
Check if Slack error present before logging + let it return a promise

sedubois:


app/lib/slack.js
```diff
@@ -41,9 +41,16 @@ module.exports = {
       slackOptions.channel = options.channel;
     }
 
-    return new Slack(options.webhookUrl || config.slack.hookUrl, {})
-    .send(slackOptions, (err) => {
-      console.error(err);
+    return new Promise((resolve, reject) => {
```
asood123: Does this mean we need to update wherever PostMessage is called? I don't know if we always call it as a promise. Like in postActivity()

asood123: Could we add a test for it?

sedubois: I think `report_transactions.js` is the only place where it looks at the result of the call. All other places call the library blindly. 

But I'll add a test anyway.

PR #287, opened 2016-04-18
use different credentials for sanitizing stripe account

asood123:


asood123:
You'll need to add two more lines to your .env file. I'll send you that over slack

PR #288, opened 2016-04-19
better description for the paypal flow

xdamman:
Need to deploy this before 2pm PST (demo to Docker)


app/gateways/paypal.js
```diff
@@ -15,19 +15,18 @@ const getConfig = (connectedAccount) => ({
 
 const getCallbackUrl = (group, transaction) => `${config.host.api}/groups/${group.id}/transactions/${transaction.id}/callback`;
 
-const createBillingPlan = (group, transaction, paypalConfig, cb) => {
+const createBillingPlan = (planDescription, group, transaction, paypalConfig, cb) => {
```
asood123: Where is this called from? We need to update that call to include additional parameter.

PR #290, opened 2016-04-19
Provide selenium download & caching + centralize post-installation

sedubois:


sedubois:
Shouldn't be merged, there's still an issue with script returning OK even in case of failure.

PR #293, opened 2016-04-21
Fix encoding of cloudinary URLs

sedubois:


sedubois:
Fixes https://github.com/OpenCollective/OpenCollective/issues/41

sedubois:
Merging this because needs to be on master to avoid the npm install cyclic dep

asood123:
We'll also need to update the other thankyou templates.

sedubois:
Good point... https://github.com/OpenCollective/api/pull/294

PR #296, opened 2016-04-25
Publish notifications to slack/gitter only when groupId is defined

sedubois:


app/lib/notifications.js
```diff
@@ -5,6 +5,9 @@ const slackLib = require('./slack');
 const ACTIVITY_ALL = 'all';
 
 module.exports = (Sequelize, activity) => {
+  if(typeof activity.GroupId === 'undefined' || typeof activity.type === 'undefined') {
```
xdamman: Are you sure that `GroupId` is `undefined` and not `null`?
Have you tested this?
We pushed this to prod and it doesn't seem to work.

sedubois: Indeed I didn't properly test this, was using an `activity` data structure from one of the unit-tests. I tested with live data and PR is here: https://github.com/OpenCollective/api/pull/307

PR #300, opened 2016-04-25
Don't filter on userId for counting webhook stripe received activities...

sedubois:
... because their `UserId` is `null` anyway


asood123:
We shouldn't have any transactions with userid null. Where are those coming
from?

sedubois:
Right, I titled wrong, I was actually doing the fix for the count of activities, not transactions. But I simplified the PR to make sure to only impact the counting on that table, very simple now...

PR #305, opened 2016-04-27
Stores distribution information passed in payment object

carlosascari:
Currently this information is stored as a JSON object in the Transactions table in the  comments column

### DIFF
--- a/server/controllers/donations.js
+++ b/server/controllers/donations.js
@@ -315,6 +315,7 @@ module.exports = (app) => {
     const amount = payment.amount;
     const interval = payment.interval;
     const isSubscription = _.contains(['month', 'year'], interval);
+    const distribution = payment.distribution ? JSON.stringify({distribution: payment.distribution}) : '';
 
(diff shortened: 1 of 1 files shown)


xdamman:
Can you add a test for that?

carlosascari:
Sure can, is it urgent? I've started working on this issue https://github.com/OpenCollective/OpenCollective/issues/39 and I'd like to finish first if possible

Also, for a test do you mean for API and for WEB with selenium? I can and will do if thats what you mean, I think the changes here alone are too small for a test: 

https://github.com/carlosascari/opencollective-api/blob/store-donation-ditribution-info/server/controllers/donations.js#L318

The stringify method can only fail if you pass it undefined or an object that references itself. This will never happen since it does a falsy check beforehand. I'd hesitate to add just an API test alone, for these changes.

xdamman:
Fair enough :-)

But yes, would be good to add a selenium test on the frontend when you have a chance.

PR #307, opened 2016-04-28
Discard notification of activities with null GroupId or type

sedubois:


xdamman:
This could be simplified by simply checking if it's false:

    if(!activity.GroupId || !activity.type)

See http://stackoverflow.com/questions/5101948/javascript-checking-for-null-vs-undefined-and-difference-between-and

PR #309, opened 2016-04-28
Check if Stripe webhook body.livemode is present

sedubois:


server/controllers/webhooks.js
```diff
@@ -37,6 +37,9 @@ module.exports = (app) => {
 
     // Stripe send test events to production as well
     // don't do anything if the event is not livemode
+    if (!body.hasOwnProperty("livemode")) {
+      console.error("Assertion error: expected a body.livemode property!");
+    }
```
xdamman: Why don't add this `console.error` within the `if (isProduction && !body.livemode)`?

sedubois: Didn't want it to be in an `if (isProduction)` because we could want to test outside of production, e.g when connecting dev localhost to Stripe via ngrok.

OTOH, I just set up ngrok to send a test Stripe donation to my machine, and the `body.livemode` property was properly defined (and was correctly set to `false`). So don't know what was the explanation for production code processing test Stripe webhooks...? Will decline this PR.

PR #312, opened 2016-04-29
Upgrade Node to 5.11.0 and NPM to 3.8.8 + fixes for CircleCi upgrade to Ubuntu 14.04

sedubois:
Hopefully fixes https://github.com/OpenCollective/api/issues/311

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,7 +6,7 @@ general:
       - production
 machine:
   node:
-    version: 5.1.0
+    version: 5.11.0
   environment:
     NODE_ENV: circleci
   services:
--- a/scripts/install_apt_get_debs.sh
+++ b/scripts/install_apt_get_debs.sh
@@ -3,7 +3,7 @@
 
 set -e
 
-APT_PACKAGES=(google-chrome-stable)
+APT_PACKAGES=(google-chrome-stable subversion)
 APT_CACHE=~/cache/apt
 
 # Work from the directory CI will cache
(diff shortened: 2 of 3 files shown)


circle.yml
```diff
@@ -6,7 +6,9 @@ general:
       - production
 machine:
   node:
-    version: 5.1.0
+    version: 5.11.0
+  npm:
+    version: 3.8.8
```
xdamman: CircleCI doesn't like that

xdamman: NPM is included in NODE so you don't need it.

circle.yml
```diff
@@ -6,7 +6,9 @@ general:
       - production
 machine:
   node:
-    version: 5.1.0
+    version: 5.11.0
+  svn:
```
xdamman: what is svn?

sedubois:
Merging this because blocking for other PRs.

sedubois:
It appears this branch accelerated CircleCi API build from 17:19 to 12:30. In particular NPM install (including postinstall) went down from 9:47 to 6:14. 

before: https://circleci.com/gh/OpenCollective/api/1433
after: https://circleci.com/gh/OpenCollective/api/1468

circle.yml
sedubois: Subversion. That's a hack I used in `test_e2e.sh` to check out the client repositories, to avoid fetching the repo history which is big (only need HEAD).

http://stackoverflow.com/questions/9609835/git-export-from-github-remote-repository

Might otherwise be possible to achieve same effect with `git clone --depth=1`.

scripts/install_apt_get_debs.sh
```diff
@@ -3,7 +3,7 @@
 
 set -e
 
-APT_PACKAGES=(google-chrome-stable)
+APT_PACKAGES=(google-chrome-stable subversion)
```
xdamman: Where are we using subversion? Such an antiquated thing.

PR #313, opened 2016-04-29
better error reporting for stripe errors, and sending curl request by email

xdamman:
Whenever there is an error, we will receive an email like this:

**Subject:** Error 402: POST /groups/2/payments/?api_key=0xxxxx

**Body:**

    To reproduce this error, run this CURL command:

    curl 'http://localhost:3060/groups/2/payments/?api_key=xxxxxxxxxxxx' -H 'connection: keep-alive' -H 'accept: application/json' -H 'origin: http://localhost:3000' -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.75 Safari/537.36' -H 'content-type: application/json' -H 'referer: http://localhost:3000/wwcodeaustin' -H 'accept-encoding: gzip, deflate' -H 'accept-language: en-US,en;q=0.8,fr;q=0.6,es;q=0.4' --data '{"payment":{"stripeToken":"tok_xxxxxxxxxxxxxx","email":"xxxxxxxx@gmail.com","amount":1,"currency":"USD","interval":"month"}}' --compressed

    Error: 
    {"code":402,"type":"StripeCardError","message":"Your card was declined."}

PR #314, opened 2016-04-29
Don't test webapp and website when we push api

xdamman:
Fixes https://github.com/OpenCollective/OpenCollective/issues/48 for API


xdamman:
Tests for #313 don't pass because of this. I need that for debugging a payment that doesn't go through for WWCode. Can someone please merge this?

sedubois:
@xdamman `app` and `website` make use of the `scripts/install_apt_get_debs.sh` and `scripts/test_e2e.sh`, if you delete them you will break setup and testing of those repos.

Also why do you want to disable these tests? The whole point of the developments in `test_e2e.sh` and related, was to be able to test everything end-to-end when doing both client and API changes.

xdamman:
As I suggested here: https://github.com/OpenCollective/OpenCollective/issues/51#issuecomment-215245734

I recommend not testing website and app when testing API. It's way too heavy. The original reason why we wanted to do that was to make sure that a change in API wouldn't break website or app. But the proper way to do this, is to make sure that we have tests that cover what website and app are expecting from the API.

xdamman:
I'm really not sure what's the purpose of `scripts/test_e2e.sh`. What does it do exactly? Those bash scripts are very hard to read.

Only website and app should run e2e tests. As such, it seems weird to have the api repo having any mention of e2e all together.

sedubois:
I agree there would be lighter ways to test API, but currently e2e tests are all we have and if we merge this branch we lose that ability. If in an API PR we e.g mistakenly change the signature of an API, it will break the clients but this won't be visible until the API gets merged, i.e too late. 

From my point of view, e2e tests depend as much on client repos as they depend on API repo (which is why I suggested before that e2e test code be extracted to repos other than website/app).

Also imagine you modify the test data in `controllers/test.js` (should move elsewhere BTW), without e2e the tests within API it will also be invisible until merge.

But sure, things would be faster if we merge this as you'll have less coverage (but you still need to leave `scripts/test_e2e.sh` and `scripts/install_apt_get_debs.sh` in this repo as they are remotely fetched by website and app).

Note however that now build time for `api` is down to ~12.5 min whereas it's ~9 min for website/app.

sedubois:
Also @xdamman you asked before the opposite, to run website and app e2e from API: https://github.com/OpenCollective/api/pull/242#issuecomment-202910993

What made you change your mind?

xdamman:
1. I didn't realize it would slow down tests so dramatically
2. I didn't realize the tests would not always pass and sometime need to be re-run
3. The benefits of this change are not worth the cost and impact on our productivity
4. I realized that there was another way to solve the initial problem that we were trying to solve with this, which is namely to write tests to make sure the website and app receive the data they expect.

sedubois:
Sure we can try to write API-level tests. We should still find a way to fix the flaky tests for the clients which suffer the same fate (your point 2, cfr. https://github.com/OpenCollective/OpenCollective/issues/60 + some timeouts).

If you'd like to merge this, still need to restore scripts cfr earlier comment.

xdamman:
I still don't get why app and website expect those scripts to exist?
It seems like a circle reference.
API should have to reference whatsoever to any e2e test.

sedubois:
- The `test_e2e` script was needed in all 3 repos, so I put it in API and made website/app remotely fetch it (`rexec:api` = "remotely execute API script")
- The `install_apt_get_debs` script was needed in website/app but as it's shared, I used same mechanism putting it in API and having it being fetched remotely

That was kind of a hack to be able to share scripts between different repos. More proper solution would be git submodules but takes a bit more effort. Would be nice though, as it would allow to share other things than scripts, e.g configuration files and common Javascript code (which there is probably plenty of).

xdamman:
Sorry but I still don't get why `api` needs any reference to `e2e` (I understand why we would need it in the context of also testing `app` and `website` when testing `api`, but in this new context of not running those tests, I don't get why `website` or `app` should expect the `api` repo to have any reference to e2e)

xdamman:
@sedubois see my latest comment above.

sedubois:
@xdamman `api` doesn't need any reference to e2e if you decide running e2e in API is too heavy (which in the current state I agree). But still `website` and `app` both need the exact same scripts `test_e2e.sh` and `install_apt_get_debs.sh` (and maybe others). Do you want me to simply duplicate these completely, having a copy of each of them in `website` and `app`? I can take care of that if you want.

(NB: I personally didn't get yet why we have separate `website` and `app` repos. I mean, currently they're both simply accessed through the web. Don't know what is the vision behind this split. There's quite some duplication between them. E.g a good part of the config, `actions/transactions/fetch_by_group.js`, etc. But that's a separate discussion.)

sedubois:
I disabled API e2e tests here: https://github.com/OpenCollective/opencollective-api/pull/340/commits/3801d3b8bbe74f0175d44edc8b4a19c0e8c2d457

This could be closed if you agree.

asood123:
@sedubois There's some history behind split of `app` and `website` - we thought `app` would be a real iOS and Android app (at some point). It's becoming clear that most of the functionality of `app` will end up in `website` anyway and we aren't going to be adding a real iOS or Android app anytime soon. 

Going forward, we should think of `app` as a tool for hosts only - approving expenses, adding funds, etc. Most everything else, we'll support from `website`.

PR #315, opened 2016-04-30
Only allow modifying users with donation in last 10 min

sedubois:


test/authorization.middleware.test.js
```diff
@@ -0,0 +1,70 @@
...
+      .catch(done);
+  });
+
+  beforeEach(() => {
+    donation = {
+      UserId: user.id,
+      currency: 'USD',
+      amount: 100
+    };
+  });
```
xdamman: if you just want to instantiate the `donation` variable, it doesn't have to be within a `beforeEach`.
Unless you actually meant to create the donation in the database? In which case you forgot to do `models.Donation.create(donation)`

test/users.routes.test.js
```diff
@@ -566,35 +566,51 @@ describe('users.routes.test.js', () => {
     });
 
     it('fails if the user already has a password', done => {
```
xdamman: So if the user already has a password we don't show the screen to edit its avatar/info after a donation?

xdamman:
I made a few comments but I wanted to say that it's great to see such thorough tests! 👍

test/authorization.middleware.test.js
sedubois: I'm just resetting the `donation` object because the tests fiddle with its `updatedAt` property.

sedubois: That was a leftover from previous code, will fix

test/users.routes.test.js
sedubois: I don't know the exact intention, but that test isn't new. I just added code to have a donation for the user to make the test pass.

PR #319, opened 2016-05-02
Ongoing work for expenses refactoring

sedubois:
Can be merged as-is and resumed later.


xdamman:
Since it's not urgent I'd love to wait for @asood123 to be back to review this.

server/models/Expense.js
```diff
@@ -0,0 +1,142 @@
...
+
+      approve() {
+        this.status = status.APPROVED;
+        return this.save();
+      }
+    }
+  });
+
+  return Expense;
+  // return Temporal(Expense, Sequelize);
```
asood123: Can you make this a TODO: enable Temporal, so all expense changes are recorded?

server/models/index.js
```diff
@@ -78,6 +79,8 @@ function setupModels(client) {
   m.User.hasMany(m.Activity);
 
   m.Activity.belongsTo(m.Transaction);
+  m.Activity.belongsTo(m.Expense);
```
asood123: I don't think we need to add `Expense` field in `Activity` model. Would rather keep everything in the `data` field.

server/models/index.js
```diff
@@ -78,6 +79,8 @@ function setupModels(client) {
...
+  // TODO shouldn't there also be a donation-activity relation?
```
asood123: Same here. Don't need Donation in Activity table either.

asood123:
Overall looks good. Few minor fixes and will be ready to be merged.

asood123:
@sedubois can we wrap this up this week? It shouldn't take more than a few mins

sedubois:
@asood123 review rdy

PR #320, opened 2016-05-02
Fix donation count in weekly transactions report

sedubois:


xdamman:
We will eventually get rid of those tags. 
Why do you need to do that change? It looks like the bug is elsewhere. Counting transactions where `amount>0` should be enough.

sedubois:
@xdamman `amount > 0` can also count funds which aren't donations.

But OK we can simply filter `Transactions` table on `"DonationId" IS NOT NULL`.

PR #321, opened 2016-05-02
Add entry to Donations table when donating with Paypal

sedubois:
To do:
- [x] add test to `donations.routes.test.js`

### DIFF
--- a/server/controllers/transactions.js
+++ b/server/controllers/transactions.js
@@ -20,6 +20,7 @@ module.exports = function(app) {
   var Activity = models.Activity;
   var Notification = models.Notification;
   var User = models.User;
+  const Group = models.Group;
   var PaymentMethod = models.PaymentMethod;
   var errors = app.errors;
   var emailLib = require('../lib/email')(app);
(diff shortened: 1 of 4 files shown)


sedubois:
Fixes https://github.com/OpenCollective/api/issues/318

asood123:
Code looks good. Thanks for cleaning up the stripeDonation bits as well. Let's add the tests and we can merge. I think there will be more paypal bugs, so let's get this merged to test the complete flow.

asood123:
Actually, there's one other change that's worth making in this PR:

Currently, we create a dummy transaction while we wait for user to go to the Paypal website. We need to find a better way of doing it. In the meantime, maybe we can disable creation of Activity in Transactions controller if `deletedAt` is set?

server/controllers/transactions.js
```diff
@@ -65,7 +66,7 @@ module.exports = function(app) {
       addTransactionToUser: ['createTransaction', (cb, results) => {
         var transaction = results.createTransaction;
 
-        if (user && user.addTransaction) {
+        if (user) {
           user
```
asood123: If user is null, this will fail (that's why the `&& user.addTransaction` was there)

server/controllers/transactions.js
```diff
@@ -85,7 +86,7 @@ module.exports = function(app) {
       addTransactionToPaymentMethod: ['createTransaction', (cb, results) => {
         var transaction = results.createTransaction;
 
-        if (paymentMethod.addTransaction) {
+        if (paymentMethod) {
           paymentMethod
             .addTransaction(transaction)
```
asood123: Same comment as above here.

server/controllers/transactions.js
sedubois: If user id `null`, `user` is falsy so the if statement won't be executed, no?

asood123: Ah right. You removed the `{}`.

PR #326, opened 2016-05-04
Les barbares

xdamman:
Fixes https://github.com/OpenCollective/OpenCollective/issues/68


piamancini:
i'm working the content so it will look ugly for a bit

P.
piamancini.com
t: @piamancini
s: piamancini
OpenCollective <https://opencollective.com>

PR #334, opened 2016-05-12
added group.settings and group.whyJoin

xdamman:
also doesn't return an error for @username twitterHandle

the whyJoin is there to allow collectives to customize that paragraph on their page.

### DIFF
--- a/migrations/20160511222234-group-settings-and-whyjoin.js
+++ b/migrations/20160511222234-group-settings-and-whyjoin.js
@@ -3,7 +3,7 @@
 module.exports = {
   up: function (queryInterface, DataTypes) {
     return queryInterface.addColumn('Groups', 'settings', DataTypes.JSON)
-      .then(() => query.queryInterface.addColumn('Groups','whyJoin', DataTypes.TEXT('long')));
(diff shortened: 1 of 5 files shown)


migrations/20160511222234-group-settings-and-whyjoin.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface, DataTypes) {
+    return queryInterface.addColumn('Groups', 'settings', DataTypes.JSON)
+      .then(() => query.queryInterface.addColumn('Groups','whyJoin', DataTypes.TEXT('long')));
```
asood123: don't need query.

server/models/Group.js
```diff
@@ -98,10 +104,9 @@ module.exports = function(Sequelize, DataTypes) {
 
     twitterHandle: {
       type: DataTypes.STRING, // without the @ symbol. Ex: 'asood123'
-      validate: {
-        notContains: {
-          args: '@',
-          msg: 'twitterHandle must be without @ symbol'
+      set(username) {
```
asood123: nice!

PR #337, opened 2016-05-16
GitHub integration

asood123:
Updated backend to store and verify github credentials during the oauth process. @sedubois - pls review. 

I modified one major thing: a github oauth needs to happen before we have a group in place. So, removed all references to `slug` and also don't need `GroupId` in `ConnectedAccounts` (will remove later).

### DIFF
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -12,7 +12,7 @@ var constants = require('../constants/activities');
 /**
  * Controller.
  */
-module.exports = function(app) {
+module.exports = (app) => {
 
   /**
    * Internal Dependencies.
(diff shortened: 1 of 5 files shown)


server/controllers/users.js
```diff
@@ -299,10 +299,28 @@ module.exports = function(app) {
    */
   return {
 
+    getOrCreate: (email, cb) => {
```
sedubois: This looks like a security hole, you allow to get and modify an existing user just by providing an email address? Shouldn't the email address instead be taken from the authenticated user?

sedubois:
OK I hadn't understood the flow, that Github auth came before group creation. And I actually agree it's better to keep the oauth solely linked to a host user; he's logging in with his personal credentials after all. It just means that only that particular user in his group will be able to execute the flow.

PR #339, opened 2016-05-17
Fix name of repos to check out after rename

sedubois:


sedubois:
Merging because client tests need this change on master to pass

PR #343, opened 2016-05-19
Implement Passport auth flow on API side

sedubois:


server/controllers/connectedAccounts.js
```diff
@@ -1,41 +1,38 @@
...
-            res.send({
-              success: true,
-              token: user.generateConnectedAccountVerifiedToken(req.application, caId, username)}))
-          .catch(next);
-      });
+      // TODO should simplify using findOrCreate but need to upgrade Sequelize to have this fix:
+      // https://github.com/sequelize/sequelize/issues/4631
+      User.findOne({ where: { email: { $in: emails }}})
+        .then(u => u || User.create({
+          username: profile.username,
```
asood123: this could be a problem. we expect the username to be unique in our database. If github username clashes with something in our db, it's going to fail. Let's not set their username to be same as on our side.

server/middleware/security/authentication.js
```diff
@@ -224,6 +212,66 @@ module.exports = function (app) {
...
+      });
+    },
+
+    authenticateService: (req, res, next) => {
+      const apiKey = req.required.api_key;
+      const apiKeyEnc = jwt.sign({apiKey}, secret, { expiresIn: '1min' });
+      const service = req.params.service;
+      passport.authenticate(service, {
+        callbackURL: `${config.host.api}/connected-accounts/${service}/callback?api_key_enc=${apiKeyEnc}`,
+        // TODO set proper scope
```
asood123: What else is needed for scope?

sedubois: Might depend what other data you needed to access through the API during the onboarding? As far as the auth flow goes, this is enough. Can be tweaked in separate PR.

PR #344, opened 2016-05-19
GitHub auth flow

asood123:
Still need to add tests for the new call. 

Also, it's very ugly. Open to suggestions on how to make it cleaner and less error prone.

@sedubois can you pls review?

### DIFF
--- a/migrations/20160519162112-make-group.slug-unique-migration.js
+++ b/migrations/20160519162112-make-group.slug-unique-migration.js
@@ -1,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface) {
+    return queryInterface.addIndex('Groups', ['slug'], {
+      indexName: 'UniqueSlugIndex',
+      indicesType: 'UNIQUE'
+    });
+  },
+
+  down: function (queryInterface) {
+    return queryInterface.removeIndex('Groups', 'UniqueSlugIndex');
+  }
+};
--- a/server/controllers/connectedAccounts.js
+++ b/server/controllers/connectedAccounts.js
@@ -37,7 +37,7 @@ module.exports = (app) => {
     get: (req, res, next) => {
       const payload = req.jwtPayload;
       const provider = req.params.service;
-      if (payload.scope === provider && payload.username) {
+      if (payload.scope === 'connected-account' && payload.username) {
         res.send({provider, username: payload.username, connectedAccountId: payload.connectedAccountId})
       } else {
         return next(new errors.BadRequest('Github authorization failed'));
--- a/server/controllers/connectedAccounts.js
+++ b/server/controllers/connectedAccounts.js
@@ -1,16 +1,11 @@
-var async = require('async');
 const config = require('config');
-const roles = require('../constants/roles');
 
 module.exports = (app) => {
   const errors = app.errors;
   const models = app.get('models');
   const ConnectedAccount = models.ConnectedAccount;
-  const Activity = models.Activity;
   const User = models.User;
-  const Group = models.Group;
-  const groups = require('../controllers/groups')(app);
-  const users = require('../controllers/users')(app);
+
 
   return {
     createOrUpdate: (req, res, next, accessToken, profile, emails) => {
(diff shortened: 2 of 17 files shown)


server/controllers/connectedAccounts.js
```diff
@@ -37,11 +43,81 @@ module.exports = (app) => {
...
+                  callback();
+                }
+              }, (err) => {
+                if (err) return next(err);
+                cb();
+              });
+            }]
+          }, (e) => {
+            if (e) return next(e);
+              res.send({success:true});
```
sedubois: REST-wise, the ID of the created resource (the group) should be returned.

server/routes.js
```diff
@@ -121,6 +121,7 @@ module.exports = (app) => {
    * Groups.
    */
   app.post('/groups', aN.authenticateUserByJwt(), required('group'), groups.create); // Create a group. Option `role` to assign the caller directly (default to null).
+  app.post('/groups/create_from_github_repo', aN.authenticateAppByApiKey, aN.parseJwtNoExpiryCheck, required('payload'), connectedAccounts.createGroupFromGithubRepo)
```
sedubois: - resources should be nouns, not verbs; and the resource that is being created is really a group, not a group called "create_from_github_repo"
- use dash instead of underscore in URLs
- missing semicolon
- why isn't the JWT expiry checked? AFAIK only place where that should happen is `POST /subscriptions/refresh_token`
- why is this `POST /groups...` handled by `connectedAccounts` controller whereas line above is handled by `groups` controller?

Maybe have a single endpoint `POST /groups?flow=manual|github`? If too difficult to combine in a single route (because security middleware differs), maybe do something like:

```
app.post('/groups', ifParam('flow', 'manual'), aN.authenticateUserByJwt(), required('group'), groups.create);
app.post('/groups', ifParam('flow', 'github'), aN.authenticateAppByApiKey, aN.parseJwtNoExpiryCheck, required('payload'), groups.createFromGithub);

function ifParam(param, value) {
  return (req, res, next) => {
    if (!req.params[param]) return next(new Error(`missing parameter ${param}`));
    if (req.params[param] !== value) return next('route');
    next();
  }
}
```

server/controllers/connectedAccounts.js
```diff
@@ -37,11 +43,81 @@ module.exports = (app) => {
...
+      const contributors = payload.users;
+      const creatorGithubUsername = payload.github_username;
+
+      ConnectedAccount
+        .findOne({
+          where: { id: connectedAccountId },
+          include: { model: User }
+        })
+        .then(ca => creator = ca.User)
+        .then(() => Group.create(group))
```
sedubois: Doesn't sequelize allow something like `ca.User.addGroup(group)`?

server/controllers/connectedAccounts.js
```diff
@@ -37,11 +43,81 @@ module.exports = (app) => {
...
+        .then(group => {
+
+          const options = {
+            role: roles.MEMBER,
+            remoteUser: creator
+          };
+          async.auto({
+            createActivity: (cb) => {
+              Activity.create({
+                type: 'group.created',
```
sedubois: should be a constant in `constants/activities.js`

migrations/20160519162112-make-group.slug-unique-migration.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface) {
+    return queryInterface.addIndex('Groups', ['slug'], {
```
sedubois: Are you sure you need an entirely separate index to enfore slug unicity? Can't it be an index on the combination of ID and slug columns or similar? Not super familiar with SQL details, but in general the least indices, the better the performance.

server/controllers/connectedAccounts.js
```diff
@@ -37,11 +43,81 @@ module.exports = (app) => {
...
+            }],
+
+            addContributors: ['addCreator', (cb) => {
+              // TODO: find a cleaner way of doing this
+              async.each(contributors, (user, callback) => {
+                // since we added the creator above with an email, avoid double adding
+                if (user !== creatorGithubUsername) {
+                  var newUser = {
+                    name: `${user}.github`,
+                    email: `${user}.github@opencollective.com`, // TODO: see if we can get their real email
```
sedubois: Why store a fake email?

server/controllers/connectedAccounts.js
asood123: fixed.

migrations/20160519162112-make-group.slug-unique-migration.js
asood123: yes, unfortunately. Wasn't able to find any other way of doing this. If you do combination of iD and slug, that means that slug can repeat (because ID is always unique). And we tie slug directly to serve the group content.

server/routes.js
asood123: Moved it to groups (which required a little bit of juggling to avoid recursion in our controllers - we need to refactor them). Added an ifParam as described above.

server/controllers/connectedAccounts.js
asood123: `email` field in the `Users` table can't be null. That's the main reason for stuffing an email address.

We could remove that constraint - i had a feeling it'll cause other issues throughout the codebase, since we expect `email` field to be present for any user.

Creating a ConnectedAccount and storing the github info there makes a lot of sense. I'll add that.

sedubois: Yes it makes sense to me to remove the unique email constraint, if we start allowing users to authenticate directly with oauth or other means

asood123: Ok I'll remove it.

asood123:
@sedubois if you get a chance, can you take one more look at the last commit. Cleaned up a bunch of logic and routes, etc. Am eager to merge this weekend to get it in staging.

server/middleware/if_param.js
```diff
@@ -0,0 +1,24 @@
...
+/**
+ *  Parameters optionally for a route.
+ */
+module.exports = function(param, value) {
+
+  return function (req, res, next) {
+    // if param doesn't exist, go to next route
+    if (!(req.query[param]) &&
+      !(req.headers[param]) &&
+      !(req.body[param])) {
```
sedubois: I'm not a fan of defining multiple ways to pass parameters to fulfill the same function. Saw it in other places in the code too. Only seems to increase complexity of the code.

server/routes.js
```diff
@@ -120,8 +121,8 @@ module.exports = (app) => {
   /**
    * Groups.
    */
+  app.post('/groups', ifParam('flow', 'github'), aN.authenticateAppByApiKey, aN.parseJwtNoExpiryCheck, aN.checkJwtExpiry, required('payload'), groups.createFromGithub); // Create a group from a github repo
   app.post('/groups', aN.authenticateUserByJwt(), required('group'), groups.create); // Create a group. Option `role` to assign the caller directly (default to null).
```
sedubois: For a bit more coherence (and to avoid having the need for line 124 to be before 125), `ifParam` could also be added on this route + add e.g `?flow=manual` in client call.

test/groups.routes.test.js
```diff
@@ -231,6 +231,137 @@ describe('groups.routes.test.js', () => {
...
+  describe('#createFromGithub', () => {
+
+    it('fails creating a group if param value is not github', (done) => {
+      request(app)
+        .post('/groups?flow=blah')
+        .send({
+          payload: privateGroupData
+        })
+        .expect(400)
+        .end((e, res) => {
```
sedubois: can omit the braces for one-liners

server/routes.js
asood123: I did that first and then realized that it makes a ton of tests fail and we'll have to add the field everywhere. Maybe still worth doing as a separate PR.

server/middleware/if_param.js
asood123: I found it confusing too. We can stick to query params on the next refactor.

PR #347, opened 2016-05-23
email confirmation for github and handling duplicate slug issue better

asood123:


server/controllers/groups.js
```diff
@@ -429,7 +430,13 @@ module.exports = function(app) {
           include: { model: User }
         })
         .then(ca => creator = ca.User)
-        .then(() => Group.create(group))
+        .then(() => Group.findOne({where: {slug: group.slug.toLowerCase()}}))
+        .then(existingGroup => {
+          if (existingGroup) {
+            group.slug = `${group.slug}+${Math.floor((Math.random() * 1000) + 1)}`;
```
sedubois: Technically it could still return an existing slug ;)

asood123: True. I was only building for 10x scale. We have about 100 groups in database. Should be fine to get us to 1000 groups - maybe even 10k!

PR #351, opened 2016-05-24
Load .env for dev mocha tests

sedubois:


asood123:
Why is this needed now?

sedubois:
I don't know, do they run for you? It's only when running locally, as CircleCi has the env vars it needs.

asood123:
Hmm they were running for me earlier. Will try after this merge.

PR #365, opened 2016-05-24
add avatar for oauth user as well

asood123:


asood123:
merging as @carlosascari already reviewed it.

PR #371, opened 2016-05-25
Remove need for DEBUG var to avoid developers debugging blindly

sedubois:


server/middleware/error_handler.js
```diff
@@ -53,16 +53,15 @@ module.exports = (err, req, res, next) => {
...
-    if(req.app.set('env') === 'production') {
+  switch (process.env.NODE_ENV) {
+    case 'production':
       sendErrorByEmail(req, err);
-    }
-
-    console.error('Error Express : ', err);
-    if (err.stack) console.log(err.stack);
+      break;
+    case 'development':
```
asood123: What happens in Staging and circleci?

sedubois: The stack then isn't printed, as was already the case whenever `DEBUG` wasn't defined.

sedubois: Actually it did print it for nightwatch tests.

asood123: would be nice to have for circleci - those things are hard to debug as it is.

PR #372, opened 2016-05-25
Replace Sequelize find with findById, needed in more recent version

sedubois:


sedubois:
That API change isn't backwards-compatible so will provide PR along with Sequelize upgrade

PR #379, opened 2016-05-26
Public activity feed

asood123:
Cleaned up activity feed to support 

- Private activity feed (like we have today)
- Public activity feed (with scrubbed info - no emails, etc)
- gitter and slack support for our users to plug their own hookurls
- also brought back attachments for private feed
- updated language a little bit, that needs more cleaning. Will do that in a separate PR after this goes live.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -43,7 +43,8 @@
     "redirectUri": "http://localhost:3060/stripe/oauth/callback"
   },
   "slack": {
-    "activityChannel": "#activity_test"
+    "privateActivityChannel": "#activity-private-test",
+    "publicActivityChannel": "#activity-public-test"
   },
   "clearbit": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
 }
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "cors": "2.7.1",
     "express": "4.13.4",
     "express-server-status": "1.0.3",
+    "flat": "^2.0.0",
     "form-data": "0.2.0",
     "handlebars": "4.0.5",
     "joi": "7.3.0",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "cors": "2.7.1",
     "express": "4.13.4",
     "express-server-status": "1.0.3",
-    "flat": "^2.0.0",
+    "flat": "2.0.0",
     "form-data": "0.2.0",
     "handlebars": "4.0.5",
     "joi": "7.3.0",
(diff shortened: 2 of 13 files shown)


config/default.json
```diff
@@ -43,7 +43,8 @@
     "redirectUri": "http://localhost:3060/stripe/oauth/callback"
   },
   "slack": {
-    "activityChannel": "#activity_test"
+    "privateActivityChannel": "#activity-private-test",
```
sedubois: I don't see any such channel

server/lib/activities.js
```diff
@@ -104,7 +89,7 @@ module.exports = {
 
       case activities.USER_CREATED:
         // Ex: New user joined: Alice Walker (alice@walker.com) | @alicewalker | websiteUrl
```
sedubois: could update example or simply remove comment

server/lib/activities.js
```diff
@@ -127,13 +112,100 @@ module.exports = {
       default:
         return `Oops... I got an unknown activity type: ${activity.type}`;
```
sedubois: Maybe just `return 'New event: ${activity.type}'` or similar?

server/lib/notifications.js
```diff
@@ -1,48 +1,71 @@
...
-      active: true
-    }
-  }).then(notifConfigs =>
-    Promise.all(notifConfigs, notifConfig => {
-      if (notifConfig.channel === 'gitter') {
-        return publishToGitter(activity, notifConfig);
-      } else if (notifConfig.channel === 'slack') {
-        return publishToSlack(activity, notifConfig);
+  // publish everything to our private channel
+  return publishToSlackPrivateChannel(activity,
```
sedubois: If I understood correctly, just noting this will add quite some redundancy, basically we'll get most events twice in public and private channel. Would be nice to just publish to private channel what's not published to the public one.

config/default.json
asood123: I created them. Haven't shared with the group yet. I'll do it before pushing.

server/lib/notifications.js
asood123: We'll get some duplication. I debated that but felt that piecing things together will be a bit harder if they are split. Let's try it this way. If it gets annoying, I'll change it.

package.json
```diff
@@ -21,6 +21,7 @@
     "cors": "2.7.1",
     "express": "4.13.4",
     "express-server-status": "1.0.3",
+    "flat": "^2.0.0",
```
sedubois: please change to `2.0.0`

PR #380, opened 2016-05-26
More promise chain fixes + fix testing of mail sending

sedubois:
Will be good to double-check email sending still works as intended after merge because of this change https://github.com/OpenCollective/opencollective-api/pull/380/files#diff-fff4a1217d1f3a712ce7f239dc28daaaR38

Mocha tests don't cover that line but the one below (L48).

### DIFF
--- a/server/controllers/groups.js
+++ b/server/controllers/groups.js
@@ -339,7 +339,7 @@ module.exports = function(app) {
         return usergroup;
       })
       .then((usergroup) => usergroup.destroy())
-      .then(() => {
+      .tap(() => {
         // Create activities.
         const remoteUser = (req.remoteUser && req.remoteUser.info) || (req.application && req.application.info);
         const activity = {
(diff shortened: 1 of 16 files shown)


server/controllers/groups.js
```diff
@@ -404,101 +407,80 @@ module.exports = function(app) {
...
 
       ConnectedAccount
         .findOne({
           where: { id: connectedAccountId },
           include: { model: User }
         })
-        .then(ca => creator = ca.User)
+        .tap(ca => {
+          creator = ca.User;
+          options = {
```
asood123: Will these options get passed all the way through?

server/controllers/groups.js
```diff
@@ -404,101 +407,80 @@ module.exports = function(app) {
...
+          type: activities.GROUP_CREATED,
+          UserId: creator.id,
+          GroupId: dbGroup.id,
+          data: {
+            group: dbGroup.info,
+            user: creator.info
+          }
+        }))
+        .then(() => _addUserToGroup(dbGroup, creator, options))
+        .then(() => {
```
asood123: don't need the braces here

server/controllers/transactions.js
```diff
@@ -49,75 +49,22 @@ module.exports = function(app) {
...
-
-      addTransactionToPaymentMethod: ['createTransaction', (cb, results) => {
-        var transaction = results.createTransaction;
-
-        if (paymentMethod) {
-          paymentMethod
-            .addTransaction(transaction)
-            .then(() => cb())
-            .catch(cb);
+    Transaction
```
asood123: This piece of code makes me very happy. So concise and easy to follow!

asood123:
tests ran fine locally.

PR #381, opened 2016-05-27
add errorhandler dependency

asood123:


asood123:
merging, since it's blocking push to production

PR #382, opened 2016-05-27
fixing mailgun promisify

asood123:


asood123:
merged this to get us live in staging again

PR #386, opened 2016-05-31
update weekly script slack

asood123:


asood123:
This broke last week with my slack notification changes.

PR #389, opened 2016-06-01
Don't display svgs in emails

sedubois:


sedubois:
NB: forgot to remove in other `thankyou` templates

PR #395, opened 2016-06-02
fetch user profile data

xdamman:
New route: `/users/:username?profile=true` to fetch the user data and its groups

### DIFF
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -325,14 +325,13 @@ module.exports = (app) => {
      * Show.
      */
     show: (req, res, next) => {
-  
+
       var userData = req.user.show;
 
-      if (req.remoteUser && remoteUser.id === req.user.id) {
+      if (req.remoteUser && req.remoteUser.id === req.user.id) {
         req.user.getStripeAccount()
           .then((account) => {
             var response = _.extend(userData, req.user.info, { stripeAccount: account });
-
             res.send(response);
           })
           .catch(next);
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -8,6 +8,7 @@ var userlib = require('../lib/userlib');
 var generateURLSafeToken = require('../lib/utils').generateURLSafeToken;
 var imageUrlToAmazonUrl = require('../lib/imageUrlToAmazonUrl');
 var constants = require('../constants/activities');
+var sequelize = require('sequelize');
 
 /**
  * Controller.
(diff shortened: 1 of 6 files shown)


server/controllers/users.js
```diff
@@ -324,16 +325,39 @@ module.exports = (app) => {
...
-      if (req.remoteUser.id === req.user.id) {
+
+      var userData = req.user.show;
+
+      if (req.remoteUser && req.remoteUser.id === req.user.id) {
         req.user.getStripeAccount()
           .then((account) => {
-            var response = _.extend({}, req.user.info, { stripeAccount: account });
-
+            var response = _.extend(userData, req.user.info, { stripeAccount: account });
```
sedubois: Could use native JS `Object.assign`

server/controllers/users.js
```diff
@@ -324,16 +325,39 @@ module.exports = (app) => {
...
             res.send(response);
           })
           .catch(next);
-      } else {
-        res.send(req.user.show);
+      } else if (req.query.profile) {
+        var groups;
+        req.user
+        .getGroups()
+        .then(results => {
```
sedubois: can simplify to `.then(results => results.map(g => Object.assign(g.minimal, { role: g.UserGroup.role })))`

server/controllers/users.js
```diff
@@ -324,16 +325,39 @@ module.exports = (app) => {
...
+        .then(groups => UserGroup.findAll({
+          where: { GroupId: { $in: groups.map(g => g.id) } }, 
+          attributes: ['GroupId', [ sequelize.fn('count', sequelize.col('GroupId')), 'members' ]],
+          group: ['GroupId']
+        }))
+        .then(counts => {
+          const membersByGroupId = {};
+          counts.map(g => { membersByGroupId[parseInt(g.GroupId,10)] = parseInt(g.dataValues.members, 10); } );
+          userData.groups = groups.map(g => _.extend(g, { members: membersByGroupId[parseInt(g.id, 10)] }));
+          res.send(userData);
```
sedubois: you're returning `undefined` so should change `then` to `tap` on line 352

server/controllers/users.js
```diff
@@ -324,16 +325,39 @@ module.exports = (app) => {
...
+        })
+        .then(groups => UserGroup.findAll({
+          where: { GroupId: { $in: groups.map(g => g.id) } }, 
+          attributes: ['GroupId', [ sequelize.fn('count', sequelize.col('GroupId')), 'members' ]],
+          group: ['GroupId']
+        }))
+        .then(counts => {
+          const membersByGroupId = {};
+          counts.map(g => { membersByGroupId[parseInt(g.GroupId,10)] = parseInt(g.dataValues.members, 10); } );
+          userData.groups = groups.map(g => _.extend(g, { members: membersByGroupId[parseInt(g.id, 10)] }));
```
sedubois: can use `Object.assign`

server/controllers/users.js
```diff
@@ -324,16 +325,39 @@ module.exports = (app) => {
...
+          res.send(userData);
+        })
+        .catch(next);
+      }
```
sedubois: (unnecessary line break)

server/params/index.js
```diff
@@ -34,17 +34,31 @@ module.exports = (app) => {
...
-        .catch(next)
+
+      const callback = (user) => {
+        if (!user) {
+          return next(new errors.NotFound(`User '${userid}' not found`));
+        } else {
+          req.user = user;
+          next();
+        }
+      }
```
sedubois: missing semicolon

server/params/index.js
```diff
@@ -34,17 +34,31 @@ module.exports = (app) => {
...
-          if (!user) {
-            return next(new errors.NotFound(`User '${userid}' not found`));
-          } else {
-            req.user = user;
-            next();
-          }
-        })
-        .catch(next)
+
+      const callback = (user) => {
```
sedubois: Or to avoid callback and duplication with `groupid` code:

```
  function getByIdOrName(model, idOrName, nameKey) {
    if (isNaN(idOrName)) { // name
      return model.find({
        where: { [nameKey]: idOrName.toLowerCase() }
      });
    } else { // id
      return model.findById(idOrName);
    }
  }

  /**
   * Public methods.
   */
  return {

    /**
     * UserId or username.
     */
    userid: (req, res, next, idOrName) => {
      getByIdOrName(User, idOrName, "username").tap(user => {
        if (!user) {
          throw new errors.NotFound(`User '${idOrName}' not found`);
        }
      })
      .then(user => req.user = user)
      .asCallback(next);
    },

    /**
     * GroupId or slug.
     */
    groupid: (req, res, next, groupIdOrName) => {
      getByIdOrName(Group, groupIdOrName, "slug").tap(group => {
        if (!group) {
          return next(new errors.NotFound(`Group '${groupid}' not found`));
        }
      })
      .then(group => req.group = group)
      .asCallback(next);
    },
    ...
  }
```

server/controllers/users.js
xdamman: What's the difference between `_.extend` and `Object.assign`?
The reason I did it this way is because we need to assign `g.minimal` to the `groups` variable that is outside of that scope.

server/controllers/users.js
xdamman: I like aligning the `.`

server/controllers/users.js
sedubois: `Object.assign` doesn't use an external library so no need for `require`

PR #398, opened 2016-06-02
Add pay flow for new expenses route

sedubois:
- [x] add tests

### DIFF
--- a/scripts/populate_expenses_table.js
+++ b/scripts/populate_expenses_table.js
@@ -1,0 +1,59 @@
+const app = require('../index');
+const config = require('config');
+const constants = require('../server/constants/transactions');
+const models = app.set('models');
+
+// Updates a transaction row, given an expense
+const updateTransaction = (expense, transaction) => {
+  transaction.type = constants.type.EXPENSE;
+  transaction.ExpenseId = expense.id;
+  return transaction.save();
+};
+
+// Creates an Expense row, given a transaction
+const createExpense = transaction => {
+  console.log("Processing transaction id: ", transaction.id);
+
+  const expense = {
+    UserId: transaction.UserId,
+    GroupId: transaction.GroupId,
+    currency: transaction.currency,
+    amount: -transaction.amount*100,
+    category: transaction.tags && transaction.tags[0],
+    title: transaction.description,
+    status: getStatus(transaction),
+    createdAt: transaction.createdAt,
+    updatedAt: transaction.updatedAt,
+    deletedAt: transaction.deletedAt,
+    incurredAt: transaction.createdAt
+  };
+
+  return models.Expense.create(expense)
+    .then(expense => updateTransaction(expense, transaction));
+};
+
+function getStatus(tx) {
+  if (tx.status === 'REIMBURSED') {
+    return 'PAID';
+  }
+  if (tx.approvedAt && !tx.approved) {
+    return 'REJECTED';
+  }
+  if (tx.approvedAt && tx.approved) {
+    return 'APPROVED';
+  }
+  return 'PENDING';
+}
+
+// Get all transactions
+models.Transaction.findAll({
+  where: {
+    amount: { $lt: 0 },
+    ExpenseId: null // this ensures that we don't reprocess a transaction
+  },
+  order: 'id'
+})
+.map(transaction => createExpense(transaction))
+.catch(err => console.log('err', err))
+.then(() => console.log('Done!'))
+.then(() => process.exit());
--- a/server/controllers/expenses.js
+++ b/server/controllers/expenses.js
@@ -4,6 +4,7 @@
 
 const Promise = require('bluebird');
 const constants = require('../constants');
+const includes = require('lodash/collection/includes');
 
 /**
  * Controller.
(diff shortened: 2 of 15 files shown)


server/models/Expense.js
```diff
@@ -16,32 +16,44 @@ module.exports = function (Sequelize, DataTypes) {
...
       set(val) {
         if (val && val.toUpperCase) {
           this.setDataValue('currency', val.toUpperCase());
         }
       }
     },
 
-    amount: DataTypes.INTEGER,
-    title: DataTypes.STRING,
-    description: DataTypes.TEXT,
```
asood123: Is there a reason to remove description? I think we'll need two fields: one short like `title` and one long for comments, optional notes, etc. `notes` maybe?

server/routes.js
```diff
@@ -173,7 +171,15 @@ module.exports = (app) => {
   app.post('/groups/:groupid/expenses', commonLegacySecurityMw, mw.authorizeIfGroupPublic, mw.authorizeAuthUserOrApp, mw.authorizeGroup, required('expense'), mw.getOrCreateUser, expenses.create); // Create an expense for a group.
   app.post('/groups/:groupid/expenses', commonLegacySecurityMw, required('expense'), mw.getOrCreateUser, expenses.create); // Create an expense for a group.
   app.post('/groups/:groupid/expenses/:expenseid/approve', aZ.authorizeAccessToGroup(), aZ.authorizeGroupAccessTo('expense'), required('approved'), expenses.setApprovalStatus); // Approve an expense.
+  app.post('/groups/:groupid/expenses/:expenseid/pay', aZ.authorizeAccessToGroup({userRoles: [HOST, MEMBER]}), aZ.authorizeGroupAccessTo('expense'), required('payoutMethod'), expenses.pay); // Pay an expense.
```
asood123: Currently, only `HOST` should be able to approve an expense.

server/models/Expense.js
```diff
@@ -125,19 +142,22 @@ module.exports = function (Sequelize, DataTypes) {
...
+        this.status = status.PAID;
         return this.save();
       }
     }
   });
 
-  return Expense;
-  // TODO: enable Temporal
-  // return Temporal(Expense, Sequelize);
+  return Temporal(Expense, Sequelize);
```
asood123: If I remember right, we'll also need to add a migration file for the new Temporal table.

server/controllers/expenses.js
```diff
@@ -13,46 +14,50 @@ module.exports = (app) => {
...
     const group = req.group;
-    const user = req.remoteUser || req.user || {};
-    Expense.create(attributes)
-      .tap(expense => expense.setUser(user))
-      .tap(expense => expense.setGroup(group))
+    const attributes = Object.assign({}, req.required.expense, {
+      UserId: user.id,
+      GroupId: group.id
+    });
+    // TODO make client specify 'title' instead of 'description' then remove this
```
asood123: Change to `notes` and keep both fields?

scripts/populate_expenses_table.js
```diff
@@ -0,0 +1,59 @@
...
+
+  const expense = {
+    UserId: transaction.UserId,
+    GroupId: transaction.GroupId,
+    currency: transaction.currency,
+    amount: -transaction.amount*100,
+    category: transaction.tags && transaction.tags[0],
+    title: transaction.description,
+    status: getStatus(transaction),
+    createdAt: transaction.createdAt,
```
asood123: We need to set `incurredAt` also

server/models/Expense.js
sedubois: fixed

scripts/populate_expenses_table.js
sedubois: fixed

server/routes.js
sedubois: fixed

migrations/20160606164500-add-expenses-table.js
```diff
@@ -0,0 +1,113 @@
+'use strict';
+
+const status = require('../server/constants/expense_status');
+
+module.exports = {
```
asood123: I didn't realize that we could create a model without creating the table

migrations/20160606164500-add-expenses-table.js
```diff
@@ -0,0 +1,113 @@
...
+          type: Sequelize.INTEGER,
+          validate: { min: 0 },
+          allowNull: false
+        },
+
+        title: {
+          type: Sequelize.STRING,
+          allowNull: false
+        },
+        notes: Sequelize.STRING,
```
asood123: Needs to support longer text. Sequelize.TEXT?

migrations/20160606164500-add-expenses-table.js
sedubois: what do you mean?

asood123: Well, the model `Expense` is already checked in and pushed. I thought it would have given an error when it doesn't find the corresponding table in the database?

sedubois: Right. Well no code is actually using that model but it's true that it's still created.

server/models/Expense.js
sedubois: I removed the migration file from this PR so can be fixed separately

server/lib/payExpense.js
```diff
@@ -0,0 +1,46 @@
...
+      var amount = expense.amount;
+      var payload = {
+        requestEnvelope: {
+          errorLanguage: 'en_US',
+          detailLevel: 'ReturnAll'
+        },
+        actionType: 'PAY',
+        currencyCode: expense.currency.toUpperCase() || 'USD',
+        feesPayer: 'SENDER',
+        memo: `Reimbursement transaction ${expense.id}: ${expense.description}`,
```
asood123: If the user will see this memo, no point in printing the txnId. It won't mean anything to the user. If group name can be had, I'd rather print that. `Reimbursement from ${group.name}: ${expense.description}`

server/lib/transactions.js
```diff
@@ -0,0 +1,71 @@
...
+      }
+    }
+
+    return models.Transaction.create({
+      netAmountInGroupCurrency: -expense.amount * 100,
+      ExpenseId: expense.id,
+      // TODO remove #postmigration, info redundant with joined tables?
+      type: expenseType,
+      amount: expense.amount,
+      currency: expense.currency,
```
asood123: FYI, we probably want to keep `amount` and `currency` in the `Transactions` even after migration. It is duplicated but useful to have in one place as a record of the original transaction (especially while dealing with multiple currencies).

server/lib/payExpense.js
sedubois: Fix here: https://github.com/OpenCollective/opencollective-api/pull/410

server/lib/transactions.js
sedubois: BTW about multiple currencies https://github.com/OpenCollective/opencollective-api/pull/411

PR #401, opened 2016-06-02
add script to pull contribution data for each group

boneskull:
- add migration for `data` col
- update `Group` model

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "node-slack": "0.0.7",
     "node-uuid": "1.4.7",
     "nodemailer": "1.11.0",
+    "opencollective-jobs": "github:opencollective/opencollective-jobs#be6d0a2014e36cdd073a5dda2b1f81d71ce07a7d",
     "passport": "0.3.2",
     "passport-github": "1.1.0",
     "paypal-adaptive": "0.4.2",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "node-slack": "0.0.7",
     "node-uuid": "1.4.7",
     "nodemailer": "1.11.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -37,6 +37,7 @@
     "node-slack": "0.0.7",
     "node-uuid": "1.4.7",
     "nodemailer": "1.11.0",
+    "opencollective-jobs": "github:opencollective/opencollective-jobs#be6d0a2014e36cdd073a5dda2b1f81d71ce07a7d",
```
asood123: Should we remove the code at the end?

boneskull: that references a particular commit.  if we remove it, we'll get whatever the `master` branch happens to be when `npm install` is run.  so, no, we  want to keep it until we start doing releases for that project

asood123: What about using version numbers instead of commits? Easier to identify and keep track of.

boneskull: I haven't tagged anything yet, but I can do that.

boneskull: done

scripts/update-contributions.js
```diff
@@ -0,0 +1,47 @@
...
+      })
+      .then(data => {
+        group.settings = group.settings || {};
+        group.settings.githubOrg = org;
+        group.data = data;
+        return group.save();
+      })
+      .then(() => {
+        console.log(`Successfully updated contribution data for group "${group.name}"`);
+      })
```
sedubois: this could be a one-liner and return value not used so can use tap `.tap(() => console.log(...))`

asood123: Btw, we should probably call the field inside `data` as `githubContributors`, so we can add other stuff to `data` later.

PR #402, opened 2016-06-02
add name field to be submitted via expenses

asood123:


server/controllers/middlewares.js
```diff
@@ -54,12 +56,16 @@ module.exports = function(app) {
...
       models.User.findOne({
         where: {
-          $or: userData
+          $or: {
+            email: userData.email,
+            paypalEmail: userData.paypalEmail
+          }
         }
       })
       .then(user => user || users._create(userData))
```
sedubois: could also leave `userData` intact and use `Object.assign({}, userData, {name})`

PR #403, opened 2016-06-03
add support for UYU and INR in email

asood123:


asood123:
Even though we don't need UYU support anymore, this is still good to merge.

PR #404, opened 2016-06-03
tweaks to contribution pulling script

boneskull:
- exit process when complete
- use `githubContributors` prop of `data` field to store data
- remove unused var

### DIFF
--- a/scripts/update-contributions.js
+++ b/scripts/update-contributions.js
@@ -2,7 +2,6 @@
 
 'use strict';
 
-const Promise = require('bluebird');
 const app = require('../index');
 const GitHubClient = require('opencollective-jobs').GitHubClient;
 const Group = app.set('models').Group;
@@ -32,10 +31,13 @@ Group.findAll({
             return acc;
           }, {});
       })
-      .then(data => {
-        group.settings = group.settings || {};
-        group.settings.githubOrg = org;
-        group.data = data;
+      .then(contributorData => {
+        group.settings = _.assign(group.settings || {}, {
+          githubOrg: org
+        });
+        group.data = _.assign(group.data || {}, {
+          githubContributors: contributorData
+        });
         return group.save();
       })
       .then(() => {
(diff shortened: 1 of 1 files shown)


scripts/update-contributions.js
```diff
@@ -44,4 +46,7 @@ Group.findAll({
       .catch(err => {
         console.log(`WARNING: ${err.message}`);
       })
+  })
+  .finally(() => {
+    process.exit();
```
sedubois: no need for braces, can be a one-liner

scripts/update-contributions.js
```diff
@@ -44,4 +46,7 @@ Group.findAll({
       .catch(err => {
         console.log(`WARNING: ${err.message}`);
```
sedubois: no need for braces, can be a one-liner

boneskull: Well, it *can*, but should it be?

Given the following:

```js
const baz = () => {
  bar();
};
const foo = () => bar();
```

What does `baz()` return?  `baz()` returns `undefined`.  

What does `foo()` return?  *Neither one of us knows* what `foo()` returns, because we don't know what `bar()` is, or what it returns.

Since I am not intending to return a value [here](https://github.com/OpenCollective/opencollective-api/pull/404#discussion-diff-65847947R47), I explicitly use braces.  This will make the code easier to reason about for future developers.  Sure, `console.log()` doesn't return anything, but keeping the braces means a developer doesn't have to worry about it.

So, in fact, I'd recommend others follow this pattern:  if you don't intend to return a value from a lambda, *use braces*.

sedubois: First looking at the more general case of `then`, just use `tap` to clarify return value (it's ignored), and results in more readable syntax so it does make it more maintainable for future devs IMHO:

```
  .tap(bla => console.log(`bla is ${bla}`)
```

Knowing that, I'd simply follow the same pattern for `catch` for consistency and whatever the return value is can be ignored, like [here](https://github.com/OpenCollective/opencollective-api/blob/c0a45245c32e25b0a703fb86a46ecc3a6d7e8706/server/lib/userlib.js#L47-L47).

PR #411, opened 2016-06-07
Fix expense payment currency (TODO)

sedubois:


server/models/Expense.js
```diff
@@ -4,6 +4,8 @@ const status = require('../constants/expense_status');
 
 module.exports = function (Sequelize, DataTypes) {
 
+  const allowedCurrencies = ['USD', 'EUR', 'GBP', 'SEK', 'UYU', 'INR'];
```
asood123: I think easiest thing right now is to extract this out to a single file and keep it in this repo. When we implement the currency support for expenses on the frontend, we can make it a separate package.

There's only one other place that this is referenced and we can link to the list there as well: https://github.com/OpenCollective/opencollective-api/blob/a87f0824ccc7a640ef9d790cfefe287778ba0313/server/lib/email.js#L62

PR #416, opened 2016-06-07
Added route to pull github repositories using users access token

carlosascari:
Added github lib, has fetchUser method used to retrieve a contributor's info like website, name and email

Covers the last tasks described here:
https://github.com/OpenCollective/OpenCollective/issues/81#issuecomment-221390767

And this issue:
https://github.com/OpenCollective/OpenCollective/issues/117#event-683413872

This PR is related the branch with the same name `fixes-for-github-flow` in website repo, and should be merged first, as the other changes rely on this repo, and not vice versa.

### DIFF
--- a/server/controllers/connectedAccounts.js
+++ b/server/controllers/connectedAccounts.js
@@ -1,4 +1,5 @@
 const config = require('config');
+const request = require('request')
 
 module.exports = (app) => {
   const errors = app.errors;
@@ -47,7 +48,25 @@ module.exports = (app) => {
       } else {
         return next(new errors.BadRequest('Github authorization failed'));
       }
-    }
+    },
 
+    fetchAllRepositories: (req, res, next) => {
+      const payload = req.jwtPayload;
+      ConnectedAccount
+      .findOne({where: {provider: 'github', id: payload.connectedAccountId}})
+      .then(ca => {
+        const options = {
+          url: `https://api.github.com/user/repos?per_page=1000&sort=stars&access_token=${ca.secret}&type=all`,
+          headers: {
+            'User-Agent': 'OpenCollective'
+          },
+          json: true
+        };
+        request(options, (error, response, body) => {
+          if (error) return next(error)
+          res.json(body)
+        });
+      })
+    }
   };
 };
--- a/server/controllers/connectedAccounts.js
+++ b/server/controllers/connectedAccounts.js
@@ -1,5 +1,6 @@
 const config = require('config');
-const request = require('request')
+const request = require('request');
+const Promise = require('bluebird');
 
 module.exports = (app) => {
   const errors = app.errors;
@@ -62,11 +63,10 @@ module.exports = (app) => {
           },
           json: true
         };
-        request(options, (error, response, body) => {
-          if (error) return next(error)
-          res.json(body)
-        });
+        return Promise.promisify(request, {multiArgs: true})(options).then(args => args[1]);
       })
+      .then(body => res.json(body))
+      .catch(next)
     }
   };
 };
--- a/server/controllers/connectedAccounts.js
+++ b/server/controllers/connectedAccounts.js
@@ -53,7 +53,7 @@ module.exports = (app) => {
     fetchAllRepositories: (req, res, next) => {
       const payload = req.jwtPayload;
       ConnectedAccount
-      .findOne({where: {provider: 'github', id: payload.connectedAccountId}})
+      .findOne({where: {id: payload.connectedAccountId}})
       .then(ca => {
         const options = {
           url: `https://api.github.com/user/repos?per_page=1000&sort=stars&access_token=${ca.secret}&type=all`,
--- a/server/controllers/groups.js
+++ b/server/controllers/groups.js
@@ -489,14 +489,12 @@ module.exports = function(app) {
                 }
               })
               .then(user => user || User.create(userAttr))
-              .then(user => {
-                return githubLib.fetchUser(contributor)
-                .then(json => {
-                  user.name = json.name;
-                  user.website = json.blog;
-                  user.email = json.email;
-                  return user.save();
-                });
+              .then(() => githubLib.fetchUser(contributor))
+              .then(json => {
+                user.name = json.name;
+                user.website = json.blog;
+                user.email = json.email;
+                return user.save();
               })
               .tap(user => user.addConnectedAccount(connectedAccount))
               .then(user => _addUserToGroup(dbGroup, user, options));
--- a/server/controllers/groups.js
+++ b/server/controllers/groups.js
@@ -27,6 +27,7 @@ module.exports = function(app) {
   const roles = require('../constants/roles');
   const activities = require('../constants/activities');
   const emailLib = require('../lib/email')(app);
+  const githubLib = require('../lib/github');
 
   /**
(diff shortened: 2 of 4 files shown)


server/controllers/connectedAccounts.js
```diff
@@ -47,7 +48,25 @@ module.exports = (app) => {
       } else {
         return next(new errors.BadRequest('Github authorization failed'));
       }
-    }
+    },
 
+    fetchAllRepositories: (req, res, next) => {
+      const payload = req.jwtPayload;
+      ConnectedAccount
+      .findOne({where: {provider: 'github', id: payload.connectedAccountId}})
```
asood123: Do we need a `provider` attribute in this clause? Shouldn't `id` identify it precisely?

carlosascari: No, you are right the id will identify it precisely alone

server/controllers/groups.js
```diff
@@ -419,37 +420,43 @@ module.exports = function(app) {
...
+        include: { model: User }
+      })
+      .tap(ca => {
+        creator = ca.User;
+        options = {
+          role: roles.MEMBER,
+          remoteUser: creator
+        };
+      })
+      .then(() => {
```
asood123: this can be tap

server/controllers/groups.js
```diff
@@ -419,37 +420,43 @@ module.exports = function(app) {
...
+        creator.website = githubUser.blog;
+        creator.name = githubUser.name;
+        return creator.save();
+      })
+      .then(() => Group.findOne({where: {slug: group.slug.toLowerCase()}}))
```
asood123: same here

server/controllers/groups.js
```diff
@@ -419,37 +420,43 @@ module.exports = function(app) {
...
+      .then(existingGroup => {
+        if (existingGroup) {
+          group.slug = `${group.slug}+${Math.floor((Math.random() * 1000) + 1)}`;
+        }
+        return Group.create(group)
+      })
+      .tap(g => dbGroup = g)
+      .then(() => Activity.create({
```
asood123: same here

server/controllers/groups.js
sedubois: you sure? Seems the found group needs to be returned to be used on line 451

server/controllers/groups.js
sedubois: AFAIK the `then` on line 467 should be aligned with this line

server/controllers/groups.js
```diff
@@ -480,6 +487,15 @@ module.exports = function(app) {
                 }
               })
               .then(user => user || User.create(userAttr))
+              .then(user => {
+                return githubLib.fetchUser(contributor)
```
sedubois: can simplify to `.then(user => githubLib.fetchUser(contributor)).then(json => ...`

server/lib/github.js
```diff
@@ -0,0 +1,26 @@
+var config = require('config');
```
sedubois: can be `const`

server/controllers/groups.js
asood123: Oh yea, you are right. ignore the original comment.

server/lib/github.js
```diff
@@ -0,0 +1,26 @@
...
+
+  fetchUser(username) {
+    const options = {
+      url: `https://api.github.com/users/${username}?client_id=${config.github.clientID}&client_secret=${config.github.clientSecret}`,
+      headers: {
+        'User-Agent': 'OpenCollective'
+      },
+      json: true
+    };
+    return new Promise((resolve, reject) => {
```
sedubois: this can probably be simplified to something like `return Promise.promisify(request, {multiArgs: true})(options).then(args => args[1])`.

Guessing from bluebird doc: http://bluebirdjs.com/docs/api/promise.promisify.html

server/routes.js
```diff
@@ -227,6 +227,11 @@ module.exports = (app) => {
   app.get('/connected-accounts/:service/verify', aN.authenticateAppByApiKey, aN.parseJwtNoExpiryCheck, connectedAccounts.get);
 
   /**
+   * Github API - fetch all repositories using the user's access_token
+   */
+  app.get('/fetch-github-repositories', aN.authenticateAppByApiKey, aN.parseJwtNoExpiryCheck, aN.checkJwtExpiry, connectedAccounts.fetchAllRepositories);
```
sedubois: API resources shouldn't contain verbs (like `fetch` here). Can just use `app.get('/github-repositories', ...);`.

sedubois: Should probably change `aN.authenticateAppByApiKey, aN.parseJwtNoExpiryCheck, aN.checkJwtExpiry` to `aN.authenticateUserByJwt()`.

server/controllers/connectedAccounts.js
```diff
@@ -47,7 +48,25 @@ module.exports = (app) => {
...
+      .findOne({where: {id: payload.connectedAccountId}})
+      .then(ca => {
+        const options = {
+          url: `https://api.github.com/user/repos?per_page=1000&sort=stars&access_token=${ca.secret}&type=all`,
+          headers: {
+            'User-Agent': 'OpenCollective'
+          },
+          json: true
+        };
+        request(options, (error, response, body) => {
```
sedubois: promise chain is broken here, need to return, e.g cfr other comment: `return Promise.promisify(request, {multiArgs: true})(options).then(args => args[1])`

Then adding at the end of the promise chain:
```
  .then(body => res.json(body))
  .catch(next)
```

server/controllers/connectedAccounts.js
```diff
@@ -47,7 +49,24 @@ module.exports = (app) => {
...
+      .findOne({where: {id: payload.connectedAccountId}})
+      .then(ca => {
+        const options = {
+          url: `https://api.github.com/user/repos?per_page=1000&sort=stars&access_token=${ca.secret}&type=all`,
+          headers: {
+            'User-Agent': 'OpenCollective'
+          },
+          json: true
+        };
+        return Promise.promisify(request, {multiArgs: true})(options).then(args => args[1]);
```
asood123: @sedubois why does args[1] work instead of args[2]?

sedubois: Well I was just guessing that as `request` returns two values, `response` and `body`, by using `multiArgs: true` you're asking bluebird to return these into an `args` array in the same order, so in that array, `response` would be at index 0 and `body` at index 1.

sedubois: @carlosascari I would actually get rid of `request` and use `request-promise` instead, we already use the package. Then no need to promisify at all, already returns a promise.

server/controllers/groups.js
```diff
@@ -419,37 +420,45 @@ module.exports = function(app) {
...
+        };
+      })
+      .tap(() => {
+        if (githubUser) {
+          creator.website = githubUser.blog;
+          creator.name = githubUser.name;
+          return creator.save();
+        }
+      })
+      .then(() => Group.findOne({where: {slug: group.slug.toLowerCase()}}))
```
sedubois: `toLowerCase()` is normally redundant, `Group.js` already has a validator to make slug lower-case.

server/lib/github.js
sedubois: @carlosascari same comment here, use `request-promise`

PR #425, opened 2016-06-09
Add transaction type for convenience in clients + cleanup

sedubois:


server/models/Donation.js
```diff
@@ -1,3 +1,5 @@
+const type = require('../constants/transactions').type.EXPENSE;
```
asood123: should this be type.DONATION? since it's the donation model.

asood123: I'll take care of it today

PR #437, opened 2016-06-13
making sure we don't return the members that are not anymore part of 

xdamman:
…the collective


package.json
```diff
@@ -25,6 +25,7 @@
     "form-data": "0.2.0",
     "handlebars": "4.0.5",
     "joi": "7.3.0",
+    "json2csv": "^3.4.2",
```
sedubois: better to freeze version

PR #439, opened 2016-06-14
updated activities and tests based on new expenses

asood123:


server/lib/activities.js
```diff
@@ -16,9 +16,12 @@ module.exports = {
     var interval = '';
     var recurringAmount = null;
     var currency = '';
-    var tags = [];
     var description = '';
     var eventType = '';
+    var provider = '';
+    var caUsername = '';
+    var caLink = '';
```
xdamman: what does `ca` stand for? It's ok to use such abbreviations deep inside the code (e.g. in a limited closure), otherwise it's better to be more explicit (also with autocomplete it's not a big issue), e.g. `ConnectedAccountUsername`

test/lib.activities.test.js
```diff
@@ -63,6 +58,22 @@ describe('lib.activities.test.js', () => {
       var actual = activitiesLib.formatMessageForPrivateChannel(activitiesData[19], true);
       expect(actual).to.equal('New user: someone (UserId: 2) added to group: <blah.com|Blah>');
     });
+
+    it (`${constants.GROUP_EXPENSE_CREATED}`, () => {
```
sedubois: (using string template isn't necessary, just use var)

PR #442, opened 2016-06-15
Profile route

xdamman:
Needed for https://github.com/OpenCollective/opencollective-website/pull/295

### DIFF
--- a/server/controllers/profile.js
+++ b/server/controllers/profile.js
@@ -31,7 +31,7 @@ module.exports = function(app) {
             return controllers.users.show(req, res, next);
             }
             else {
-              return next(new errors.NotFound());
+              return next(new errors.NotFound("There is no collective or user with this slug"));
             }
           })
         }
--- a/server/controllers/profile.js
+++ b/server/controllers/profile.js
@@ -1,0 +1,40 @@
+/**
+ * Dependencies.
+ */
+const _ = require('lodash');
+
+/**
+ * Controller.
+ */
+module.exports = function(app) {
+
+  const models = app.set('models');
+  const errors = app.errors;
+
+  return (req, res, next) => {
+    const slug = req.params.slug;
+    const controllers = req.app.set('controllers');
+
+    models.Group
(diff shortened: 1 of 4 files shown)


server/controllers/profile.js
```diff
@@ -0,0 +1,35 @@
...
+          req.group = group;
+          return controllers.groups.getOne(req, res, next);
+        }
+        else {
+          models.User.findOne({where: {username: slug}})
+          .then((user) => {
+            if(user) {
+            req.user = user;
+            req.query.profile = true;
+            return controllers.users.show(req, res, next);
```
asood123: @sedubois any ideas on how to make this type of flow better using promises?

test/profile.routes.test.js
```diff
@@ -0,0 +1,65 @@
...
+ * Variables.
+ */
+var userData = utils.data('user1');
+var groupData = utils.data('group1');
+var models = app.set('models');
+
+/**
+ * Tests.
+ */
+describe('groups.routes.test.js', () => {
```
asood123: should be `profile.routes.test.js`, so we don't misidentify the test when it fails.

server/controllers/profile.js
sedubois: @asood123 The complexity comes from mixing users and groups within the same endpoint, although they are resources of different nature. I find this query counterintuitive: `User.findOne({where: {username: slug}})`. A slug isn't a username. What if a user has the same username as a group's slug? Then it will be shadowed.

Haven't been in the loop for profiles but AFAIK using separate routes for group and user profiles would provide better separation of concerns. Then code complexity above would collapse.

asood123: Agreed. It's hacky. We need to find a better solution.

sedubois: I agree and that's e.g how GitHub does, e.g github.com/sedubois vs github.com/opencollective. But we then need to ensure unicity across group slugs + usernames.

PR #443, opened 2016-06-15
Supercollective - part 1(DON'T MERGE)

asood123:
- fixed a bug where balance wasn't showing up in app
- changed balance to return in cents (slowly changing every amount to be stored and return in cents)
- added columns 'tags' and 'supercollective' to groups
- refactored groups controller to remove async


server/controllers/groups.js
```diff
@@ -552,70 +530,22 @@ module.exports = function(app) {
...
-      }
-
-      res.send(group);
-    });
-
+    const group = req.group.info;
+
+    req.group.getStripeAccount()
+      .tap(account => group.stripeAccount = account && _.pick(account, 'stripePublishableKey'))
+      .then(() => req.group.getConnectedAccount())
```
xdamman: shouldn't this be: `.then(req.group.getConnectedAccount)` ?

sedubois: both should work, because `getConnectedAccount ` takes no param. But yes I prefer `.then(req.group.getConnectedAccount)`.

server/controllers/groups.js
```diff
@@ -172,12 +151,11 @@ module.exports = function(app) {
       .then((result) => {
         const json = result.toJSON();
 
-        cb(null, {
+        return Promise.resolve({
```
sedubois: you can normally just `return { donationTotal: ... }`, will be made thenable. I actually even prefer

```
.then(result => result.toJSON())
.then(json => {
  donationTotal: ...
  ...
});
```

server/controllers/groups.js
```diff
@@ -552,70 +530,22 @@ module.exports = function(app) {
...
-      if (_.get(results, 'getConnectedAccount.provider') === 'paypal') {
-        group.hasPaypal = true; // hack for the prototype, to refactor
-      }
-
-      res.send(group);
-    });
-
+    const group = req.group.info;
+
+    req.group.getStripeAccount()
```
sedubois: To store temporary properties `account`, `hasPaypal`, `balance` and `yearlyIncome` you can also use `Promise.props`.

server/controllers/groups.js
```diff
@@ -552,70 +530,22 @@ module.exports = function(app) {
...
+      .tap(account => group.stripeAccount = account && _.pick(account, 'stripePublishableKey'))
+      .then(() => req.group.getConnectedAccount())
+      .tap(account => group.hasPaypal = account && account.provider === 'paypal')
+      .then(() => getGroupBalance(group.id))
+      .tap(balance => group.balance = balance)
+      .then(() => getYearlyIncome(group.id))
+      .tap(yearlyIncome => group.yearlyIncome = yearlyIncome)
+      .then(() => getPublicPageInfo(group.id))
+      .tap(publicPageInfo => Object.assign(group,
```
sedubois: can normally just `.tap(publicPageInfo => Object.assign(group, publicPageInfo))`

server/controllers/users.js
```diff
@@ -1,14 +1,14 @@
...
-var Bluebird = require('bluebird');
-var async = require('async');
-var userlib = require('../lib/userlib');
-var generateURLSafeToken = require('../lib/utils').generateURLSafeToken;
-var imageUrlToAmazonUrl = require('../lib/imageUrlToAmazonUrl');
-var constants = require('../constants/activities');
-var sequelize = require('sequelize');
+const _ = require('lodash');
+const Bluebird = require('bluebird');
+const async = require('async');
```
sedubois: would be nice to get rid of `async` ;)

server/controllers/users.js
```diff
@@ -326,12 +327,12 @@ module.exports = (app) => {
...
     show: (req, res, next) => {
 
-      var userData = req.user.show;
+      const userData = req.user.show;
 
       if (req.remoteUser && req.remoteUser.id === req.user.id) {
         req.user.getStripeAccount()
           .then((account) => {
-            var response = Object.assign(userData, req.user.info, { stripeAccount: account });
+            const response = Object.assign(userData, req.user.info, { stripeAccount: account });
```
sedubois: could rename `account` => `stripeAccount` and simplify to `{ stripeAccount }`

server/models/Group.js
```diff
@@ -124,6 +124,15 @@ module.exports = function(Sequelize, DataTypes) {
       get() {
         return `${config.host.website}/${this.get('slug')}`;
       }
+    },
+
+    tags: {
+      type: DataTypes.ARRAY(DataTypes.STRING)
+    },
+
+    supercollective: {
```
sedubois: maybe `isSuperCollective` if it's a boolean?

sedubois: Also I'm not very familiar with the supercollective idea, but depending on how much investment we want to consider and what you'd like to be able to achieve later (e.g only targeting small structures or going toward bigger NGOs, companies, etc):

From a data structure point of view, `isSuperCollective` isn't something to be persisted but to be based on whether it has child collectives; those parent-child relationships would be the thing to store in DB.

Some potential advantages:
- nest collectives within each other on several levels (to have super-super-collectives etc)
- easily move a collective from a supercollective to another or allow having multiple parent collectives
- more easily perform search operations (e.g "find all members under this supercollective")
- implement more generic security rules
- centralized management, e.g send a tweet to everyone who donated over X € in last month
- in long run potentially easier to integrate within entreprise directories (LDAP or whatnot)

E.g create an `OrgTree` table with columns `parent` and `child`, or even just a single column containing a materialized path (a supercollective with ID 50: `50/`; a collective with ID 60 belonging to that supercollective: `50/60/`). (+ potentially register all of them under a root collective: `1/50/60/`).

Optionally users could also be registered within the OrgTree by replacing `UserGroups`, so the materialized path's leaf node could be a groupId or userid, distinguished through an additional column `nodeType` with e.g values `COLLECTIVE`, `HOST`, `MEMBER` or `BACKER`.

(Unclear to me right now whether same table should mix collectives, users and security rules (`HOST`). It could be a slippery slope where we're tempted to put everything in there?)

Materlialized paths normally also allow for fast search operations, even for structures with many levels, using a proper index and JOIN and LIKE operators.

Some similarities with centralized security management tools (turning them on their head, using transparency): https://www.f-secure.com/en/web/business_global/management-portal

sedubois: @xdamman @piamancini

asood123:
Don't merge this. I'm breaking it up into smaller chunks.

asood123:
pulled of these changes in a different branch.

PR #445, opened 2016-06-15
return balance when you fetch groups

asood123:


server/controllers/users.js
```diff
@@ -1,14 +1,14 @@
...
-var _ = require('lodash');
-var Bluebird = require('bluebird');
-var async = require('async');
-var userlib = require('../lib/userlib');
-var generateURLSafeToken = require('../lib/utils').generateURLSafeToken;
-var imageUrlToAmazonUrl = require('../lib/imageUrlToAmazonUrl');
-var constants = require('../constants/activities');
-var sequelize = require('sequelize');
+const _ = require('lodash');
+const Bluebird = require('bluebird');
```
sedubois: Could be renamed to `Promise`

server/controllers/users.js
```diff
@@ -366,23 +367,23 @@ module.exports = (app) => {
...
-      var promise = withRoles ?
+      const promise = withRoles ?
         getGroupsFromUserWithRoles(req, options) :
         getGroupsFromUser(req, options);
 
-      promise
-        .then(out => res.send(out))
-        .catch(next);
+      promise.map((group) => {
+        return getGroupBalance(group.id)
```
sedubois: can omit braces and `return`

PR #449, opened 2016-06-16
added user.description and user.isOrganization

xdamman:


asood123:
Need to add a migration file or it'll croak in production

xdamman:
What do you mean by that? This is just adding new columns so there shouldn't be any migration involved.

asood123:
Given how so many people work on the code base, I think in general we
shouldn't let the model diverge from the underlying database. Is there a
reason not to do the migration at the same time?

xdamman:
Doh, I forgot to add the migration file. Done.

PR #451, opened 2016-06-17
Add 404 handler in API

sedubois:


asood123:
Could you add a test for it? In case we break it, we should know right away

sedubois:
I added a test

PR #456, opened 2016-06-20
Fix balance

asood123:


asood123:
These changes coming in a different branch.

PR #457, opened 2016-06-20
Groups refactor

asood123:
- Includes an exception to let nsp check skip the latest advisory (don't have a quick fix for it)
- refactor groups controller to move a few items to the model
- refactor users controller to send balance back for the app (fix the $NaN bug)


server/controllers/groups.js
```diff
@@ -552,70 +477,27 @@ module.exports = function(app) {
...
+    .then(values => {
+      group.stripeAccount = values[0] && _.pick(values[0], 'stripePublishableKey');
+      group.hasPaypal = values[1] && values[1].provider === 'paypal';
+      group.balance = values[2];
+      group.yearlyIncome = values[3];
+      group.donationTotal = values[4];
+      group.backersCount = values[5];
+      return group;
+    })
+    .then(group => res.send(group))
```
xdamman: `.then(res.send))`

server/controllers/users.js
```diff
@@ -18,20 +17,20 @@ module.exports = (app) => {
...
+  const errors = app.errors;
 
   /**
    *
    * Private methods.
    *
    */
 
-  var getUserGroups = (UserId) => {
+  const getUserGroups = (UserId) => {
```
xdamman: This should be in the model (for another refactoring)

server/controllers/groups.js
asood123: I tried that. For somer reason, it didn't work. Makes me wonder if there is something else wrong with this promise chain.

PR #461, opened 2016-06-20
fix  bug in the app

asood123:


server/controllers/users.js
```diff
@@ -353,7 +353,8 @@ module.exports = (app) => {
       .then(groups => groups.map(group => groupData.push(group.info)))
       .then(() => groupObjects.map(groupObject => {
         const group = _.find(groupData, {id: groupObject.id})
-        return _.extend(group, { balance: groupObject.getBalance() })
+        return groupObject.getBalance()
```
asood123: Another lesson for us in using promises. Previous line never returns.

xdamman: yes because you mixed a synchronous function `_.extend` with a promise `getBalance()`.

PR #465, opened 2016-06-22
enable temporal history on expense table

asood123:
With this, any changes in the Expense table done through Sequelize (not manually) will be recorded. So, whenever an expense gets approved, rejected, paid, etc, we'll know when it happened and what the state was before the row changed.

### DIFF
--- a/server/models/Expense.js
+++ b/server/models/Expense.js
@@ -1,4 +1,4 @@
-const Temporal = require('Sequelize-Temporal');
+const Temporal = require('sequelize-temporal');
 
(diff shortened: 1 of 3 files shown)


asood123:
Hmm I don't know why it can't find the module sequelize-temporal. It's in the package.json.

server/models/Expense.js
```diff
@@ -1,10 +1,12 @@
+const Temporal = require('Sequelize-Temporal');
```
xdamman: you may want to use lowercase here in the require.

PR #467, opened 2016-06-27
add support for githubRepo setting and count total stars across organ…

asood123:
…izations

### DIFF
--- a/scripts/update-contributions.js
+++ b/scripts/update-contributions.js
@@ -17,7 +17,8 @@ Group.findAll({
   attributes: [
     'id',
     'name',
-    'settings'
+    'settings',
+    'data'
   ]
 })
   .tap(groups => {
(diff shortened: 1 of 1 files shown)


scripts/update-contributions.js
```diff
@@ -17,37 +17,63 @@ Group.findAll({
...
-        data.repoData = _.mapValues(repos, repo => _.omit(repo, 'contributors'));
-        return data;
-      })
-      .then(data => {
-        group.settings = _.assign(group.settings || {}, {
-          githubOrg: org
+    var fetchPromise;
+
+    if (org) {
+      fetchPromise = client.contributorsInOrg({orgs: [org]})
```
xdamman: with such a method name `contributorsInOrg`, I would expect the argument to be the string of the organization.

asood123: That comes from the external library (that @boneskull wrote)

PR #469, opened 2016-06-28
added new /homepage route

xdamman:
`curl http://localhost:3060/homepage` returns 

    {
      collectives: {
        meetup: [{group1}, {}, ...],
        opensource: [{group1},{}, ...]
      },
      stats: {
        totalCollectives: 60,
        totalDonations: 30124,
        totalDonors: 293 
      }
    }

### DIFF
--- a/server/controllers/homepage.js
+++ b/server/controllers/homepage.js
@@ -1,0 +1,67 @@
+/**
+ * Controller.
+ */
+module.exports = function(app) {
+
+  const models = app.set('models');
+  const queries = require('../lib/queries')(app);
+
+  return (req, res, next) => {
+    const getGroupsByTag = (tag) => {
+      return models.Group.findAll({
+        where: {
+          isPublic: true,
+          tags: { $contains: [tag] }
+        },
+        order: [['updatedAt', 'DESC']],
+        limit: 6
+      })
+    };
+
+    /**
+     * get total number of active collectives
+     * (a collective is considered as active if it has ever received any funding from its host or through a donation)
+     */
+    const getTotalCollectives = () => {
+      return models.Transaction.aggregate('GroupId', 'count', {
+        distinct: true,
+        where: {
+          amount: { $gt: 0 }
+        }
+      })
+    };
+
+    const getTotalDonors = () => {
+      return models.Transaction.aggregate('UserId', 'count', {
+        distinct: true,
+        where: {
+          amount: { $gt: 0 },
+          PaymentMethodId: { $ne: null }
+        }
+      })
+    };
+
+    Promise.all([
+      getTotalCollectives(),
+      getTotalDonors(),
+      queries.getTotalDonations(),
+      getGroupsByTag('open source'),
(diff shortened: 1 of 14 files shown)


server/controllers/homepage.js
```diff
@@ -0,0 +1,44 @@
...
+    };
+
+    Promise.all([
+      queries.getTotalCollectives(), 
+      queries.getTotalDonors(), 
+      queries.getTotalDonations(),
+      getGroupsByTag('open source'),
+      getGroupsByTag('meetup')
+    ])
+    .then(r => {
```
asood123: In general, I think a more descriptive variable names are better. result > r

server/lib/queries.js
```diff
@@ -0,0 +1,137 @@
...
+    .then(res => res[0].totalDonors);
+  };
+
+  /**
+   * get total number of active collectives 
+   * (a collective is considered as active if it has ever received any funding from its host or through a donation)
+   */
+  const getTotalCollectives = () => {
+    return sequelize.query(`
+      SELECT COUNT(DISTINCT("GroupId")) as "totalCollectives" FROM "Transactions" WHERE amount > 0
```
asood123: This is going to include collectives that are deleted. Since it's a simple query, can we use Sequelize for it, instead of a raw query.

server/lib/queries.js
```diff
@@ -0,0 +1,137 @@
...
+      GROUP BY t."GroupId"
+      ORDER BY "amountInUSD" DESC`,
+    {
+      type: sequelize.QueryTypes.SELECT
+    });
+  };
+
+  const getTotalDonors = () => {
+    return sequelize.query(`
+      SELECT COUNT(DISTINCT("UserId")) as "totalDonors" FROM "Transactions" WHERE "PaymentMethodId" IS NOT NULL AND amount > 0
```
asood123: Same comment as below here. Either need to account for `deletedAt` or switch to using Sequelize directly (preferred).

asood123:
Can you also add a test for this new route with data? Since this will be on our homepage, let's make sure we'll know if it breaks.

server/lib/queries.js
xdamman: It's not a simple query given the COUNT and DISTINCT. I'll add the `deletedAt` filter.

server/lib/queries.js
```diff
@@ -0,0 +1,112 @@
...
+  const generateFXConversionSQL = (aggregate) => {
+    var currencyColumn = "currency";
+    var amountColumn = "amount";
+    
+    if(aggregate) {
+      currencyColumn = 'MAX(g.currency)';
+      amountColumn = 'SUM("amount")';
+    }
+
+    // FXRate as of 6/27/2016
```
asood123: 👍

asood123:
Waiting on a test to merge this. You can probably copy the leaderboard tests with some dummy data. Don't let perfect (real data for tests) prevent this from being merged, when good enough (dummy data) is fine. :)

PR #470, opened 2016-06-28
When requesting profile page send data for CollectiveCard:

carlosascari:
Get user groups and include yearlyIncome as well as backers for each group
OpenCollective/website repo has a branch with the same name, it depends on this branch:
https://github.com/OpenCollective/opencollective-website/pull/319

### DIFF
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -26,6 +26,7 @@ module.exports = (app) => {
   const UserGroup = models.UserGroup;
   const sendEmail = require('../lib/email')(app).send;
   const errors = app.errors;
+  const queries = require('../lib/queries')(app);
 
   /**
    *
@@ -327,7 +328,7 @@ module.exports = (app) => {
                   return backers;
                 }
 
-                User.getAllBelongingToGroupById(group.id)
+                queries.getUsersFromGroupWithTotalDonations(group.id)
                   .then(appendTier)
                   .then(resolve)
                   .catch(reject);
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -2,11 +2,13 @@
  * Dependencies.
  */
 const _ = require('lodash');
+const groupBy = require('lodash/collection/groupBy');
 const async = require('async');
 const userlib = require('../lib/userlib');
 const generateURLSafeToken = require('../lib/utils').generateURLSafeToken;
 const imageUrlToAmazonUrl = require('../lib/imageUrlToAmazonUrl');
 const constants = require('../constants/activities');
+const roles = require('../constants/roles');
 const sequelize = require('sequelize');
 
 /**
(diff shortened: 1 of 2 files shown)


server/controllers/users.js
```diff
@@ -309,22 +312,39 @@ module.exports = (app) => {
...
-          groups = results.map(g => Object.assign(g.minimal, { role: g.UserGroup.role, createdAt: g.UserGroup.createdAt }));
-          return groups;
+        var groupInfoArray = []
+        req.user.getGroups()
+        .then(groups => {
+          return Promise.all(groups.map(group => {
+            return Promise.all([
+              group.getYearlyIncome(),
+              new Promise((resolve, reject) => {
+                groupsController.getUsers({group: group}, {send: resolve}, reject);
```
asood123: We try not to link controllers to each other, especially by faking the req, res, next params. It might be better to move this in a library, since you only need backers (from what I can tell on the frontend).

server/controllers/users.js
```diff
@@ -309,22 +312,39 @@ module.exports = (app) => {
...
+              })
+            ])
+            .then(values => {
+              var groupInfo = group.info;
+              groupInfo.yearlyIncome = values[0];
+              var usersByRole = groupBy(values[1], 'role');
+              groupInfo.hosts = usersByRole[roles.HOST] || [];
```
asood123: Do you need the host on the frontend? Each group only has one host at any one time.

server/controllers/users.js
```diff
@@ -309,22 +312,39 @@ module.exports = (app) => {
...
+              groupInfo.members = usersByRole[roles.MEMBER] || [];
```
asood123: Same here. Is this needed on the frontend?

server/controllers/users.js
```diff
@@ -309,22 +312,39 @@ module.exports = (app) => {
...
+              return group;
+            })
+          }))
         })
         .then(groups => UserGroup.findAll({
-          where: { GroupId: { $in: groups.map(g => g.id) } },
-          attributes: ['GroupId', [ sequelize.fn('count', sequelize.col('GroupId')), 'members' ]],
-          group: ['GroupId']
-        }))
+            where: { GroupId: { $in: groups.map(g => g.id) } },
```
asood123: spacing is off. two space tabs.

server/controllers/users.js
```diff
@@ -309,22 +312,39 @@ module.exports = (app) => {
...
+            attributes: ['GroupId', [ sequelize.fn('count', sequelize.col('GroupId')), 'members' ]],
```
asood123: This code is also fetching the count of members?

server/controllers/users.js
carlosascari: No, hosts are not needed I will remove that line

You can find see the component here: https://github.com/OpenCollective/opencollective-website/pull/319

server/controllers/users.js
carlosascari: Yes I must of missed it, the only change is the spacing

PR #471, opened 2016-06-29
set a group's isPublic to true but mark deletedAt to hide it

asood123:
Without this fix, anyone who signs up via github will have a new group live instantly.


test/groups.routes.test.js
```diff
@@ -25,7 +25,7 @@ var stripeMock = require('./mocks/stripe');
 /**
  * Tests.
  */
-describe('groups.routes.test.js', () => {
+describe.only('groups.routes.test.js', () => {
```
xdamman: remove the .only

PR #473, opened 2016-07-05
fix query for yearlyincome

asood123:
Existing query counts each transaction separately for a recurring monthly donation. This removes the duplicates.


asood123:
Found a bug in the new query. Resubmitting it shortly.

PR #476, opened 2016-07-06
typo in the yearly income query

asood123:


asood123:
Merging this to unblock production pushes.

PR #478, opened 2016-07-12
Updated Collective data returned for open source and meetups on /homepage

carlosascari:
Includes backers list of groups
Includes yearlyIncome for each group

This is to complete the home page, the data here will fill the CollectiveCard component's footer.

### DIFF
--- a/server/controllers/homepage.js
+++ b/server/controllers/homepage.js
@@ -1,3 +1,7 @@
+const groupBy = require('lodash/collection/groupBy');
+const roles = require('../constants/roles');
+const utils = require('../lib/utils');
+
 /**
  * Controller.
  */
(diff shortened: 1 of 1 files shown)


server/controllers/homepage.js
```diff
@@ -45,8 +84,8 @@ module.exports = function(app) {
       getTotalCollectives(),
       getTotalDonors(),
       queries.getTotalDonations(),
-      getGroupsByTag('open source'),
-      getGroupsByTag('meetup')
+      getGroupsByTagForCollectiveCard('open source'),
+      getGroupsByTagForCollectiveCard('meetup')
```
xdamman: Why did you rename it? Just change the current implementation of `getGroupsByTag`. Otherwise we will have dead code.

carlosascari: I did not rename the function, I created another function that extends `getGroupsByTag`, specifically to supply the information that is displayed at the footer of the CollectiveCard component.

https://github.com/OpenCollective/opencollective-api/pull/478/files#diff-9c36a551a062a029a91b82d5299327ebR25

PR #480, opened 2016-07-12
Login

asood123:
Ready to be merged.

### DIFF
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -27,6 +27,7 @@ module.exports = (app) => {
   const sendEmail = require('../lib/email')(app).send;
   const errors = app.errors;
   const queries = require('../lib/queries')(app);
+  const Unauthorized = errors.Unauthorized;
 
   /**
    *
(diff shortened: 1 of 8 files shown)


server/controllers/users.js
```diff
@@ -400,6 +401,61 @@ module.exports = (app) => {
...
+      if (!req.application || !req.required.email) {
+        return next(new Unauthorized('Unauthorized'))
+      }
+      var redirect;
+      if (req.body.redirect) {
+        redirect = req.body.redirect;
+      } else {
+        redirect = '/';
+      }
+      return models.User.findOne({
```
xdamman: don't we already have `req.user` with the middleware?

server/controllers/users.js
```diff
@@ -400,6 +401,61 @@ module.exports = (app) => {
       .catch(next)
     },
 
+    /**
+     * Send an email with refreshed token
+     */
```
xdamman: "For the case when a user has submitted an expired token, we can automatically detect the email address and send a refreshed token"

PR #488, opened 2016-07-18
Migrate all subscriptions from old host to new

asood123:


server/gateways/stripe.js
```diff
@@ -20,7 +20,7 @@ const getOrCreatePlan = (stripeAccount, plan) => {
 
   return stripeClient.plans.retrieve(plan.id)
     .catch((err) => {
-      if (err.type === 'StripeInvalidRequest' && _.contains(err.message, 'No such plan')) {
+      if (err.type === 'StripeInvalidRequestError' && _.contains(err.message, 'No such plan')) {
```
xdamman: I'd add a `console.log` to make sure you log potential errors even if they don't match that specific error that you expect.

server/gateways/stripe.js
```diff
@@ -36,6 +36,30 @@ const createSubscription = (stripeAccount, customerId, subscription) => {
 };
 
 /**
+ * Retrieve stripe subscription
+ */
+const retrieveSubscription = (stripeAccount, customerId, stripeSubsriptionId) => {
+  return client(stripeAccount).customers.retreiveSubscription(customerId, stripeSubsriptionId);
```
sedubois: is the typo in `retreiveSubscription` intended?

asood123: done

asood123: nope. good catch. Ended up not using that code.

server/gateways/stripe.js
asood123: done.

scripts/migrate_subscriptions_in_stripe.js
```diff
@@ -0,0 +1,143 @@
...
+    // store the new stripeSubscription info in our table
+    .then(newStripeSubscription => {
+      const oldData = currentOCSubscription.data;
+
+      return currentOCSubscription.updateAttributes({
+        data: Object.assign({}, newStripeSubscription, { oldData }),
+        stripeSubscriptionId: newStripeSubscription.id
+      });
+    })
+    // delete the old subscription from stripe
```
xdamman: Before deleting, I'd add a `catch` and make sure there is no error before.

asood123: If there is an error, won't it jump to the catch on line 98?

PR #492, opened 2016-07-19
fix script error

asood123:


asood123:
Clock ticking on migration, so merging this myself.

PR #500, opened 2016-07-21
Related groups

asood123:
Modifies backend to return related groups data as part of any call that fetches group data for publicgroup on the frontend.

A few other comments:
- refactored `homepage` controller to move the group related call to `Groups` model as a `classMethod` 
- changed `queries` library to only need `Sequelize`, rather than `app`. It's easier to link it from various `Models` that way

### DIFF
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -1,7 +1,6 @@
-module.exports = function(app) {
+module.exports = function(sequelize) {
 
-  const models = app.set('models');
-  const sequelize = models.sequelize;
+  const models = sequelize.models;
 
   /*
   * Hacky way to do currency conversion on Leaderboard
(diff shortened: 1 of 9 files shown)


server/lib/queries.js
```diff
@@ -48,15 +47,19 @@ module.exports = function(app) {
...
+    }
     return sequelize.query(`
       WITH "totalDonations" AS (
         SELECT "GroupId", SUM(amount) as "totalDonations", MAX(currency) as currency, COUNT(DISTINCT "GroupId") as collectives FROM "Transactions" WHERE amount > 0 AND currency='USD' AND "PaymentMethodId" IS NOT NULL GROUP BY "GroupId"
       )
       SELECT g.id, g.name, g.slug, g.mission, g.logo, t."totalDonations", t.currency, t.collectives
       FROM "totalDonations" t LEFT JOIN "Groups" g ON t."GroupId" = g.id
-      WHERE t."totalDonations" > 100 AND g.tags @> $tag AND g."deletedAt" IS NULL
-      ORDER BY "totalDonations" DESC LIMIT 3
+      WHERE t."totalDonations" > 100 AND g.tags && $tag AND g."deletedAt" IS NULL ${excludeClause}
```
asood123: changed operator to `&&` as that will do an intersection of two sets, rather than subset check. Needed because `tags` can have multiple entries, so subset can fail easily.

server/models/Group.js
```diff
@@ -376,6 +379,11 @@ module.exports = function(Sequelize, DataTypes) {
         return settings.twitter;
       },
 
+      getRelatedGroups(number) {
```
xdamman: can you rename this to `limit` for consistency?

asood123:
Merging, since it's been reviewed.

PR #501, opened 2016-07-22
Fix related groups call

asood123:


server/models/Group.js
```diff
@@ -381,7 +383,7 @@ module.exports = function(Sequelize, DataTypes) {
 
       getRelatedGroups(limit) {
         limit = limit || 3
-        return Group.getGroupsSummaryByTag(this.tags, limit, [this.id]);
+        return Group.getGroupsSummaryByTag(this.tags, limit, [this.id])
```
xdamman: why no more semi colon? ;-)

PR #502, opened 2016-07-23
backend call to automatically generate supercollectives

asood123:
Note this only generates the collectives in a super collective. Will need to separately figure out sponsors, etc.

### DIFF
--- a/server/models/Group.js
+++ b/server/models/Group.js
@@ -382,8 +382,12 @@ module.exports = function(Sequelize, DataTypes) {
       },
 
       getRelatedGroups(limit) {
-        limit = limit || 3
-        return Group.getGroupsSummaryByTag(this.tags, limit, [this.id]);
+        // don't fetch related groups for supercollectives for now
+        if (!this.isSupercollective) {
+          limit = limit || 3
+          return Group.getGroupsSummaryByTag(this.tags, limit, [this.id]);
+        }
+        return Promise.resolve();
       },
 
       hasHost() {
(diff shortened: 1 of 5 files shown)


server/models/Group.js
```diff
@@ -382,8 +382,12 @@ module.exports = function(Sequelize, DataTypes) {
       },
 
       getRelatedGroups(limit) {
```
xdamman: For design sake, I'd rather have `getRelatedGroups(tags, limit)`
E.g.: the homepage could be considered has a super collective, and we want to show the related groups for "open source" and the ones for "meetup".

server/models/Group.js
```diff
@@ -382,8 +382,12 @@ module.exports = function(Sequelize, DataTypes) {
...
-        limit = limit || 3
-        return Group.getGroupsSummaryByTag(this.tags, limit, [this.id]);
+        // don't fetch related groups for supercollectives for now
+        if (!this.isSupercollective) {
```
xdamman: Ok. But why does it matter?
(don't need to change)

server/models/Group.js
```diff
@@ -394,15 +398,24 @@ module.exports = function(Sequelize, DataTypes) {
           }
         })
         .then(userGroup => Promise.resolve(!!userGroup));
+      },
+
+      getSuperCollectiveData() {
```
xdamman: I'm really questioning the need to introduce a new concept for "super collective". All collectives should be "super collectives" (because they are all super, ah ah)

E.g. MochaJS could link to other collectives that are dependencies.

xdamman: (to be clear, that shouldn't prevent us from merging this PR, it's definitely good enough for now)

server/models/Group.js
asood123: Otherwise, we'll end up repeating some of the groups below the page.

server/models/Group.js
asood123: yes, there's a better way to organize relatedGroups and supercollective. One way is to ignore the isSupercollective column and just use group.setting.superCollectiveTag. If that's set, we fetch related groups.

xdamman: I like that, but I'd rename it `settings.includeCollectivesByTag` (or withTags)
(so that a new reader doesn't have to understand the concept of super collective to understand what this does)

PR #503, opened 2016-07-25
New route to sync sponsors with meetup description

xdamman:
The new route is `GET /groups/:groupid/services/meetup/sync`

The group needs to have the right settings for meetup.com:

    group.settings = {
      meetup: {
        api_key: 'xxxxxxxxxxxx', // you can find it here: https://secure.meetup.com/meetup_api/key/
        url: 'http://www.meetup.com/meetup-group-NqJGKtSx'
      }
    }

I also took the opportunity to upgrade node from 5.11 to 6.2 so now we can use `class`, `let`, etc.


package.json
```diff
@@ -67,6 +67,7 @@
     "errorhandler": "1.4.3",
     "eslint": "3.1.1",
     "eslint-config-opencollective": "1.1.3",
+    "eslint-plugin-react": "^5.2.2",
```
sedubois: suggest to freeze version

server/lib/meetup.js
```diff
@@ -0,0 +1,65 @@
+const request = require('request');
+const filter = require('lodash/collection/filter');
+const values = require('lodash/object/values');
+const errors = require('../lib/errors');
+const Promise = require('bluebird');
+const requestPromise = Promise.promisify(request, {multiArgs: true});
```
sedubois: can normally just `require('request-promise')`

server/lib/meetup.js
```diff
@@ -0,0 +1,65 @@
...
+    const promises = [];
+    return requestPromise(reqopt).then(result => {
+      const meetups = result[1];
+      for (var i=0;i<meetups.length;i++) {
+        var meetup = meetups[i];
+        if (!meetup.description.match(new RegExp(`^${descriptionHeader.substr(0, 50)}`))) {
+          promises.push(this.updateMeetupDescription(meetup.id, `${descriptionHeader}\n ${meetup.description}`));
+        }
+      }
+      return Promise.all(promises).then(results => {
```
sedubois: I have a feeling you could just `return Promise.all(promises).map(r => r[1])`

server/routes.js
```diff
@@ -141,6 +141,8 @@ module.exports = (app) => {
   app.post('/groups/:groupid/payments', aN.authenticateUserOrApp(), required('payment'), mw.getOrCreateUser, donations.post); // Make a payment/donation.
   app.post('/groups/:groupid/payments/paypal', aN.authenticateUserOrApp(), required('payment'), donations.paypal); // Make a payment/donation.
 
+  app.get('/groups/:groupid/services/meetup/sync', mw.fetchUsers, controllers.meetup.sync);
```
sedubois: technically if it modifies data on the backend, it shouldn't be a `GET`

server/lib/meetup.js
xdamman: nice, it actually simplifies managing the result 👍

server/lib/meetup.js
xdamman: good point, but not necessary anymore with require('request-promise') :-)

server/routes.js
xdamman: it doesn't modify data on the backend :-)

asood123:
Should we be storing the meetup api_key in settings? those are often passed to the frontend. Better to store them in the ConnectedAccounts table?

xdamman:
Do we pass `group.settings` to the frontend? Why? We shouldn't do that.

Yes, ideally we should store that in `ConnectedAccounts` but I was waiting for when we would do a full oAuth. Maybe that's something that @sedubois can do this week? It's just another oAuth recipe for meetup.

In the meantime, isn't this good enough to get started?

asood123:
group.settings stores things like language and I'm sure it'll have other things that frontend needs. We should assume it'll be shared with the app. 

We can add a manual row to connectedAccounts while waiting for proper Oauth?

server/lib/meetup.js
```diff
@@ -1,9 +1,7 @@
-const request = require('request');
 const filter = require('lodash/collection/filter');
 const values = require('lodash/object/values');
 const errors = require('../lib/errors');
-const Promise = require('bluebird');
-const requestPromise = Promise.promisify(request, {multiArgs: true});
+const requestPromise = require('request-promise');
```
sedubois: (I'd just use var name `request` instead of `requestPromise`)

PR #504, opened 2016-07-25
randomly pick collectives to show on homepage and related groups

asood123:


server/controllers/homepage.js
```diff
@@ -34,8 +34,8 @@ module.exports = function(app) {
       getTotalCollectives(),
       getTotalDonors(),
       queries.getTotalDonations(),
-      models.Group.getGroupsSummaryByTag('open source'),
-      models.Group.getGroupsSummaryByTag('meetup'),
+      models.Group.getGroupsSummaryByTag('open source', 3, [], 100, true),
+      models.Group.getGroupsSummaryByTag('meetup', 3, [], 100, true),
```
xdamman: I wouldn't pick the top 100, we only have 67 collectives or so.
I'd pick randomly from the top 10 or 20 at best at the moment.

asood123: `3` is the limit, `100` is the minimum totalDonation needed to be in the set where we pick out ones at random.

xdamman: I know but still, don't we want to show the top?

I'll merge it for now, we can try and see the results.

asood123: Then, we should increase the minimum from $100 to let's say $500?

PR #505, opened 2016-07-26
upgrade pg to fix seuqelizeconnectionerror

asood123:


asood123:
blocking production, merging on my own.

PR #508, opened 2016-07-27
add more info to debug paypal

asood123:


server/lib/payExpense.js
```diff
@@ -36,7 +36,12 @@ module.exports = app => {
       return new Promise((resolve, reject) => {
         app.paypalAdaptive.pay(payload, (err, res) => {
           if (err) {
-            console.log("PayPal payment error", err, res);
+            console.log("PayPal payment error: ", err);
+            console.log("PayPal response: ", res);
+            if (res.error && res.error[0] && res.error[0].parameter) {
+              console.log("PayPal error.parameter: ", res.error[0].parameter); // this'll give us more details on the error
```
asood123: parameter is currently printed as '[Object]' and we need to see what's inside it.

asood123:
This isn't passing because of an issue on Paypal's sandbox: https://www.paypal-community.com/t5/About-Payments/Adaptive-payment-api-is-sending-Proxy-error/m-p/1091510. 

Merging this so we get info on the real issue of why some expenses aren't being paid.

PR #513, opened 2016-08-03
Try to fix twitter auth "Failed to find request token in session"

sedubois:
Inspired by:
- https://github.com/jaredhanson/passport-twitter/issues/45#issuecomment-44218220
- https://www.npmjs.com/package/express-session#resave
- http://stackoverflow.com/a/29624825/4623987

### DIFF
--- a/test/connectedAccounts.routes.test.js
+++ b/test/connectedAccounts.routes.test.js
@@ -42,7 +42,7 @@ describe('connectedAccounts.routes.test.js: GIVEN an application and group', () 
             expect(err).not.to.exist;
             const baseUrl = 'https://github.com/login/oauth/authorize';
             const apiKeyEnc = '.*';
-            const redirectUri = encodeURIComponent(`${config.host.api}/connected-accounts/github/callback?api_key_enc=${apiKeyEnc}&utm_source=mm&slug=`);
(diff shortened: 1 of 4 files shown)


test/connectedAccounts.routes.test.js
```diff
@@ -42,7 +42,7 @@ describe('connectedAccounts.routes.test.js: GIVEN an application and group', ()
             expect(err).not.to.exist;
             const baseUrl = 'https://github.com/login/oauth/authorize';
             const apiKeyEnc = '.*';
-            const redirectUri = encodeURIComponent(`${config.host.api}/connected-accounts/github/callback?api_key_enc=${apiKeyEnc}&utm_source=mm&slug=`);
+            const redirectUri = encodeURIComponent(`http://127.0.0.1:.*/connected-accounts/github/callback?api_key_enc=${apiKeyEnc}&utm_source=mm&slug=`);
```
asood123: Did you mean to check int 127.0.0.1?

sedubois: That's what allows the test to pass... The passport code seems to resolve to that in the unit tests. The port changes with every run.

sedubois:
@asood123 if the changes in this PR don't help, maybe we can try removing `session: false` here:

https://github.com/OpenCollective/opencollective-api/blob/b2750a0f37bf1b1a5af459234ff403a54a3dcb4d/server/lib/express.js#L34

(just replacing `Object.assign(...)` with `config.twitter`)

PR #516, opened 2016-08-05
monthly email report -- need to be tested but ready to be merged

xdamman:


cron/monthly/send-monthly-report.js
```diff
@@ -0,0 +1,102 @@
...
+    return Promise.all(promises);
+};
+
+const d = new Date;
+d.setMonth(d.getMonth() - 1);
+const month = moment(d).format('MMMM');
+
+const init = () => {
+
+  const query = {
```
asood123: spacing is off

cron/monthly/send-monthly-report.js
```diff
@@ -0,0 +1,102 @@
...
+
+const emailLib = require('../../server/lib/email')(app);
+const debug = require('debug')('monthlyreport');
+
+const Group = app.set('models').Group;
+const User = app.set('models').User;
+const Notification = app.set('models').Notification;
+
+const processGroups = (groups) => {
+    const promises = [];
```
asood123: fyi, you can use Promise.map and combine .push and .all into one step

server/models/Transaction.js
```diff
@@ -86,6 +86,22 @@ module.exports = function(Sequelize, DataTypes) {
   }, {
     paranoid: true,
 
+    classMethods: {
+      createMany: (transactions, defaultValues) => {
```
asood123: This really should be a Promise.map.

`return Promise.map(transactions, transaction => Transaction.create(_.merge(...)))`

test/group.model.test.js
```diff
@@ -0,0 +1,152 @@
...
+    });
+
+    group.getExpenses(null, startDate, endDate)
+      .then(expenses => {
+        expect(expenses.length).to.equal(totalExpenses);
+        done();
+      })
+      .catch(e => { console.error("error", e, e.stack) });
+  });
+});
```
asood123: Glad you added these tests!

cron/monthly/send-monthly-report.js
```diff
@@ -0,0 +1,102 @@
...
+
+  const query = {
+      attributes: [
+          'id',
+          'slug',
+          'name',
+          'currency'
+      ],
+      where: {
+        slug: { $in: ['railsgirlsatl'] } // for testing
```
asood123: Should this be in here? It would be better to do this through the notification table too. That way, we can easily enable for a group or two and then extend. Once it works, we can automatically add it for new groups to the table.

cron/monthly/send-monthly-report.js
xdamman: Good idea but I'll have to import the Bluebird promises. It's not built-in.

cron/monthly/send-monthly-report.js
xdamman: It is done through the notification table. 
But it is first fetching all the groups, then for each group looking for who to notify in the Notifications table.

xdamman: It's looking for notifications of type "group.monthlyreport", see https://github.com/OpenCollective/opencollective-api/blob/4649875921e583f29b1a670ea59193e305266882/cron/monthly/send-monthly-report.js#L81

asood123: What's the need for leaving that slug in there?

PR #520, opened 2016-08-08
bring back createMany that got squashed

asood123:


asood123:
Merging this so other tests can pass.

PR #524, opened 2016-08-09
Discover

carlosascari:
For https://github.com/OpenCollective/OpenCollective/issues/198
Related https://github.com/OpenCollective/opencollective-website/pull/393

##### New `/discover` endpoint
Accepts 3 url params: 
- show {String} 
Can be `all` to not filter by tag or a tag to filter the results by.

- sort {String} 
Either `newest` or `oldest`. Defaults to `newest`.

- offset {Numeric}
A number to the offset in results to start reading from. Provides basic pagination. Defaults to `0`.

Return is an array of collectives/groups and is limited to 12 results per request.
> 12 is evenly divisible by 1, 2, 3 & 4, which is the number of columns that there will be on most displays. This will make each response to render complete rows (no missing columns), unless there are no more results to discover.

##### New `/groups/tags` endpoint
Returns an array of strings where each string is a unique tag that exists.


asood123:
Very clean. Nice work. Can you add a test for the route and make sure it's returning something in that test? Homepage route is a good one that you can copy.

server/controllers/discover.js
```diff
@@ -0,0 +1,33 @@
...
+    const sort = req.query.sort;
+    const offset = req.query.offset;
+    models.Group.getGroupsSummaryByTag(
+      !show || show === 'all' ? '' : show, 
+      12,
+      [], 
+      100, 
+      false,
+      'g."createdAt"',
+      sort === 'newest' ? 'desc' : 'asc',
```
sedubois: could pass directly `sort = 'desc'` or 'asc' to avoid double vocabulary

server/controllers/discover.js
```diff
@@ -0,0 +1,33 @@
...
+      offset
+    )
+    .then(results => {
+      const di = {
+        show: show,
```
sedubois: can just use `show`, `sort` etc

server/lib/queries.js
```diff
@@ -48,12 +48,16 @@ module.exports = function(sequelize) {
...
-  const getGroupsByTag = (tag, limit, excludeList, minTotalDonation, randomOrder) => {
-    var excludeClause = ''
+  const getGroupsByTag = (tag, limit, excludeList, minTotalDonation, randomOrder, orderBy, orderDir, offset) => {
+    var tagClause = '';
+    var excludeClause = '';
     var minTotalDonationClause = '';
     var orderClause = 'BY t."totalDonations"';
-    if (randomOrder) {
-      orderClause = 'BY random()'
+    var orderDirection = (orderDir === 'asc') ? 'ASC' : 'DESC';
```
sedubois: SQL is case insensitive (AFAIK) so can use `orderDir` directly

server/lib/queries.js
```diff
@@ -64,14 +68,18 @@ module.exports = function(sequelize) {
...
     return sequelize.query(`
       WITH "totalDonations" AS (
         SELECT "GroupId", SUM(amount) as "totalDonations", MAX(currency) as currency, COUNT(DISTINCT "GroupId") as collectives FROM "Transactions" WHERE amount > 0 AND currency='USD' AND "PaymentMethodId" IS NOT NULL GROUP BY "GroupId"
       )
       SELECT g.id, g.name, g.slug, g.mission, g.logo, g."backgroundImage", t."totalDonations", t.currency, t.collectives
       FROM "Groups" g LEFT JOIN "totalDonations" t ON t."GroupId" = g.id
-      WHERE ${minTotalDonationClause} g.tags && $tag AND g."deletedAt" IS NULL ${excludeClause}
-      ORDER ${orderClause} DESC NULLS LAST LIMIT ${limit}
+      WHERE ${minTotalDonationClause} ${tagClause} g."deletedAt" IS NULL ${excludeClause}
+      ORDER ${orderClause} ${orderDirection} NULLS LAST LIMIT ${limit} OFFSET ${offset || 0}
```
sedubois: Collectives don't change frequently so we don't really care, but in general using `OFFSET` is discouraged, no guarantee that next query will start off from where previous one ended

sedubois: Similar comment for `orderBy` and `orderDirection`, we already have `mw.sorting()`

server/lib/queries.js
```diff
@@ -80,6 +88,16 @@ module.exports = function(sequelize) {
   };
 
   /**
+  * Get list of all unique tags for groups.
+  */
+  const getUniqueGroupTags = () => {
+    return sequelize.query('SELECT DISTINCT UNNEST(tags) FROM "Groups" WHERE ARRAY_LENGTH(tags, 1) > 0')
+    .then(results => {
```
sedubois: or `.then(results => results[0].map(x => x.unnest).sort())`

server/lib/queries.js
carlosascari: This was done with backward-compatibility in mind. 

It assumes that `orderDir` MAY be undefined, so instead we default to `desc`, which was the original sort direction. We only change direction to `asc` if it is explicitly passed.

This was done since the method is used elsewhere, and I wanted to preserve the method's usage as designed originally because I feel this mitigates the introduction of bugs.

I tippy-toed around existing/expected functionality

server/controllers/discover.js
carlosascari: You are right, directly passing `sort` will introduce a SQL injection vulnerability since it goes straight to a raw query. So, in order to pass sort directly, we must **escape** it before passing it to the `queries.getGroupsByTag` method.

So it can be either a ternary condition or a function call to the *escape* function

server/lib/queries.js
carlosascari: `mw.paginate` is hard coded to paginate using a table's `id` column, I don't think I can use it here.
https://github.com/OpenCollective/opencollective-api/blob/master/server/controllers/middlewares.js#L309

`mw.sorting` is usuable, however since we are using a raw query I don't believe I'll take full advantage of it, the middleware sets `req.paginate.dir`, which I would use the same way I use `req.query.sort` now, **plus** a function call to escape the input to prevent blind sql injection.

I believe these middlewares where intended to use alongside `_.merge()` and not-raw-sql: 
https://github.com/OpenCollective/opencollective-api/blob/master/server/controllers/groups.js#L140

Thanks for bringing these middlewares to light, I will make sure I take advantage of them from now on

PR #526, opened 2016-08-09
moving all scripts to cron/ directory + cleanup

xdamman:


cron/weekly/weekly-report.js
```diff
@@ -1,12 +1,12 @@
-const app = require('../index');
```
sedubois: You could rename to `weekly-slack-report.js` similarly to the `monthly-slack-report.js`. Also, as it's in directory `weekly`, could just be named `slack-report.js`.

PR #528, opened 2016-08-10
better styling of currency, fix bug for number of backers

xdamman:


cron/monthly/email-report.js
```diff
@@ -34,7 +34,10 @@ const init = () => {
           'name',
           'currency',
           'tags'
-      ]
+      ],
+      where: {
+        slug: 'railsgirlsatl'
```
asood123: Did you mean to check this in? Won't it require another PR once you want to take it live?

xdamman: argh no, good catch.

xdamman:
Merging so that I can send a last test email (tests are passing and @asood123 reviewed once)

PR #530, opened 2016-08-11
Bring back email notifications

asood123:
They were temporarily removed to refactor `transaction.create`. Now that emailLib is cleaner, this PR brings back the notification to all `MEMBER`s of a collective when there is a new transaction.

- Also, adds a `production` check on pushing out to all channels (gitter, slack, email) to avoid spamming

### DIFF
--- a/server/lib/notifications.js
+++ b/server/lib/notifications.js
@@ -5,6 +5,7 @@ const Promise = require('bluebird');
 const activitiesLib = require('../lib/activities');
 const slackLib = require('./slack');
 const twitter = require('./twitter');
+const emailLib = require('../lib/email');
 const activityType = require('../constants/activities');
 
 module.exports = (Sequelize, activity) => {
@@ -21,13 +22,17 @@ module.exports = (Sequelize, activity) => {
         return Promise.resolve([]);
       }
       return Sequelize.models.Notification.findAll({
+        include: {
+          model: Sequelize.models.User,
+          attributes: ['email']
+        },
         where: {
           type: [
             activityType.ACTIVITY_ALL,
             activity.type
           ],
           GroupId: activity.GroupId,
-          channel: ['gitter', 'slack', 'twitter'],
+          channel: ['gitter', 'slack', 'twitter', 'email'],
           active: true
         }
       })
@@ -40,6 +45,8 @@ module.exports = (Sequelize, activity) => {
           return publishToSlack(activity, notifConfig.webhookUrl, {});
         } else if (notifConfig.channel === 'twitter') {
           return twitter.tweetActivity(Sequelize, activity);
+        } else if (notifConfig.channel === 'email') {
+          return emailLib.sendMessageFromActivity(Sequelize, activity, notifConfig);
         } else {
           return Promise.resolve();
         }
--- a/server/lib/notifications.js
+++ b/server/lib/notifications.js
@@ -46,7 +46,7 @@ module.exports = (Sequelize, activity) => {
         } else if (notifConfig.channel === 'twitter') {
           return twitter.tweetActivity(Sequelize, activity);
         } else if (notifConfig.channel === 'email') {
-          return emailLib.sendMessageFromActivity(Sequelize, activity, notifConfig);
+          return emailLib.sendMessageFromActivity(activity, notifConfig);
         } else {
           return Promise.resolve();
         }
@@ -58,7 +58,7 @@ module.exports = (Sequelize, activity) => {
 
 function publishToGitter(activity, notifConfig) {
   const message = activitiesLib.formatMessageForPublicChannel(activity, 'markdown');
-  if (message) {
+  if (message && process.env.NODE_ENV === 'production') {
     return axios.post(notifConfig.webhookUrl, { message });
   } else {
     Promise.resolve();
--- a/server/lib/slack.js
+++ b/server/lib/slack.js
@@ -29,8 +29,8 @@ module.exports = {
     if (!options) {
       options = {};
     }
-    const message = activitiesLib.formatMessageForPublicChannel(activity, 'slack');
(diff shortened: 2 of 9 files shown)


test/notification.model.test.js
```diff
@@ -30,13 +32,14 @@ var Notification = models.Notification;
 /**
  * Tests.
  */
-describe("notification.model.test.js", () => {
+describe.only("notification.model.test.js", () => {
```
sedubois: intended?

asood123: i'll remove it before submitting it... was working with arnaud on the tests

arnaudbenard: Working on it now

arnaudbenard: @asood123 I have noticed multiple things in that code:
- The email should be sent with `sendMessageFromActivity` but this function does not exist
- The notifications lib expects `notifConfigs` to be able to send the email. In the tests, this array is empy, I think that's why it doesn't get called

Once those two things are fixed, I think that we will be able to test with a stub.

server/lib/utils.js
```diff
@@ -197,6 +197,22 @@ const defaultHostId = () => {
...
+ * Check if this is an internal email address.
+ * Useful for testing emails in localhost or staging
+ */
+const isEmailInternal = (email) => {
+  if (email.toLowerCase().indexOf('@opencollective.com') != -1 ||
+    email.toLowerCase().indexOf('@opencollective.org') != -1 ||
+    email.toLowerCase() === 'asood123@gmail.com' ||
+    email.toLowerCase() === 'piamancini@gmail.com' ||
+    email.toLowerCase() === 'xdamman@gmail.com' ||
+    email.toLowerCase() === 'sedubois@gmail.com') {
```
sedubois: Could we please avoid committing private emails in source code? If really needed, instead use opencollective.com emails?

asood123: was a bit torn about it. 

Actually, have an idea. I'll wipe the recipient if it's not internal and send the email to the bcc address. That way, we can remove our personal addresses.

server/lib/notifications.js
```diff
@@ -51,7 +58,7 @@ module.exports = (Sequelize, activity) => {
 
 function publishToGitter(activity, notifConfig) {
   const message = activitiesLib.formatMessageForPublicChannel(activity, 'markdown');
-  if (message) {
+  if (message && process.env.NODE_ENV === 'production') {
```
sedubois: Is this needed? Notify is normally already only called in prod:
https://github.com/OpenCollective/opencollective-api/blob/190521380f85ed80019fef4b2b78d3d84b223706/server/models/Activity.js#L19

server/lib/slack.js
```diff
@@ -29,8 +29,8 @@ module.exports = {
     if (!options) {
       options = {};
     }
-    const message = activitiesLib.formatMessageForPublicChannel(activity, 'slack');
-    return this.postMessage(message, webhookUrl, options);
+      const message = activitiesLib.formatMessageForPublicChannel(activity, 'slack');
+      return this.postMessage(message, webhookUrl, options);
```
sedubois: spurious whitespace?

server/lib/slack.js
```diff
@@ -53,6 +53,11 @@ module.exports = {
     }
 
     return new Promise((resolve, reject) => {
+      // production check
+      if (process.env.NODE_ENV !== 'production') {
```
sedubois: same remark as above

test/lib.notifications.test.js
```diff
@@ -0,0 +1,145 @@
...
+
+  // stub the transport
+  beforeEach(done => {
+    sinon.stub(nm, 'sendMail', (object, cb) => {
+      cb(null, object);
+    });
+    done();
+  });
+
+  afterEach(done => {
```
sedubois: I guess `afterEach(() => nm.sendMail.restore());` would be sufficient

test/lib.notifications.test.js
```diff
@@ -0,0 +1,145 @@
...
+      user: user,
+      group: group,
+      config: config
+    };
+
+    var subject, body;
+
+    templateData.transaction.id = 1;
+
+    if (templateData.transaction.link.match(/\.pdf$/))
```
sedubois: Missing braces

test/lib.notifications.test.js
```diff
@@ -0,0 +1,145 @@
...
+        .send({
+          transaction: transactionsData[0]
+        })
+        .expect(200))
+      .then(res => {
+        expect(res.body).to.have.property('GroupId', group.id);
+        expect(res.body).to.have.property('UserId', user.id); // ...
+      })
+      .then(() => models.Transaction.findAll())
+      .then(transactions => {
```
sedubois: This should be a `tap`

test/lib.notifications.test.js
```diff
@@ -0,0 +1,145 @@
...
+        body = emailLib.getBody(template);
+      })
+      .then(() => request(app)
+        .post('/groups/' + group.id + '/transactions')
+        .set('Authorization', 'Bearer ' + user.jwt(application))
+        .send({
+          transaction: transactionsData[0]
+        })
+        .expect(200))
+      .then(res => {
```
sedubois: This should be a `tap`

test/lib.notifications.test.js
```diff
@@ -0,0 +1,145 @@
...
+      })
+      .then(() => models.Transaction.findAll())
+      .then(transactions => {
+        expect(transactions.length).to.equal(1);
+      })
+      .then(() => models.Activity.findAll())
+      .then(activities => {
+        expect(activities.length).to.equal(1);
+      })
+      .then(() => {
```
sedubois: This should be a `tap`

PR #531, opened 2016-08-12
Transaction test

asood123:
Found a simple case to figure out tests. 

`Transaction` model (https://github.com/OpenCollective/opencollective-api/blob/master/server/models/Transaction.js) has two classMethods: `createFromPayload()`, which creates a new transaction and `createActivity()`. When a new `transaction` is created, an `afterCreate` hook calls `createActivity()`.

This PR adds a test for this flow. We want to test that `createFromPayload` leads to creating an `activity`. I tried stubbing `createActivity` to confirm it gets called and that didn't work. How would you test this? @arnaudbenard

### DIFF
--- a/test/transaction.model.test.js
+++ b/test/transaction.model.test.js
@@ -6,7 +6,9 @@ const _ = require('lodash');
 const app = require('../index');
 const config = require('config');
 const expect = require('chai').expect;
+const sinon = require('sinon');
 const utils = require('../test/utils.js')();
+var roles = require('../server/constants/roles');
 
 /**
  * Models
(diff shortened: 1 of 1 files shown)


test/transaction.model.test.js
```diff
@@ -98,4 +119,23 @@ describe('transaction model', () => {
     })
     .catch(done);
   });
+
+  it('createFromPayload() generates a new activity', done => {
+
+    beforeEach(() => sinon.spy(Transaction, 'createActivity'));
```
asood123: Tried using `Transaction.prototype` and `Transaction.constructor.prototype`.

test/transaction.model.test.js
```diff
@@ -98,4 +119,23 @@ describe('transaction model', () => {
...
+
+    afterEach(() => Transaction.createActivity.restore());
```
asood123: This errors out `TypeError: Transaction.createActivity.restore is not a function`

arnaudbenard:
@asood123 I fixed the tests. I got inspired by the sequelize unit tests themselves https://github.com/sequelize/sequelize/blob/master/test/unit/model/bulkcreate.test.js

PR #532, opened 2016-08-12
fix emailLib in the error handler

asood123:


asood123:
Merging this, since this is blocking any error emails from production. @xdamman pls review post-merge.

PR #536, opened 2016-08-17
Fix github callback which doesn't have access to application anymore

sedubois:
This PR fixes the Github e2e test currently blocking the website PRs.

This bug was introduced by this change:
https://github.com/OpenCollective/opencollective-api/pull/517/files#diff-33baa3b9cb45a92fc70513b61df48cd9L240

For meetup oauth, the encrypted API key couldn't be kept in the callback, because the callbackURL passed as argument to the two Meetup calls (oauth steps 1 and 2) need to be exactly the same. 

So instead, I keep track of the application through the session. But actually this introduces the same problem as encountered with oauth1 based on sessions, session isn't found during the API-website interaction (problem not observed on localhost).

- [ ] fix session not found


sedubois:
Bringing alternative fix through other PR (redirect through website instead of api to obtain api key)

PR #540, opened 2016-08-17
add meetup identifiers

asood123:


asood123:
merging this since api won't start without this.

PR #542, opened 2016-08-17
get users active in last 90 days

boneskull:
see OpenCollective/OpenCollective#212

@asood123 here's what I'm getting at.  I was having trouble with the test environment, so unfortunately I am going to have to abandon this for now.

### DIFF
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -126,6 +126,13 @@ module.exports = function(sequelize) {
         FROM "Donations" d
         WHERE d."GroupId" = :GroupId AND d.amount >= 0
         GROUP BY "UserId"
+      ), last_donation AS (
+        SELECT
+          max("UserId") as "UserId",
+          max("updatedAt") as "updatedAt"
+        FROM "Donations" d
+        WHERE d."GroupId" = :GroupId AND d.amount >= 0
+        GROUP BY "UserId"
       )
       SELECT
         ug."UserId" as id,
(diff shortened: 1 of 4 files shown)


server/lib/queries.js
```diff
@@ -126,6 +126,13 @@ module.exports = function(sequelize) {
         FROM "Donations" d
         WHERE d."GroupId" = :GroupId AND d.amount >= 0
         GROUP BY "UserId"
+      ), last_donation AS (
+        SELECT
+          max("UserId") as "UserId",
+          max("updatedAt") as "updatedAt"
+        FROM "Donations" d
+        WHERE d."GroupId" = :GroupId AND d.amount >= 0
+        GROUP BY "UserId"
```
asood123: This will miss recurring donations. Recurring donations are only listed once in the `Donations` table with a `SubscriptionId`, that indicates that it's recurring. After that recurring donations are only stored in `Transactions` table. 

You could deal with that by left join `subscriptionid` and checking that `isActive` is true.

asood123: Or in line 133, changing `Donations` to `Transactions` might do the trick also.

server/routes.js
```diff
@@ -138,7 +138,9 @@ module.exports = (app) => {
   app.get('/groups/tags', groups.getGroupTags); // List all unique tags on all groups
   app.get('/groups/:groupid', aZ.authorizeAccessToGroup({authIfPublic: true}), groups.getOne);
   app.get('/groups/:groupid/users', aZ.authorizeAccessToGroup({authIfPublic: true}), cache(60), groups.getUsers); // Get group users
+  app.get('/groups/:groupid/active_users', aZ.authorizeAccessToGroup({authIfPublic: true}), cache(60), groups.getActiveUsers); // Get group users
```
sedubois: How about adding a flag to the endpoint of line 140 instead of introducing new endpoint? Same comment 2 lines below

test/usergroup.routes.test.js
```diff
@@ -400,10 +400,11 @@ describe('usergroup.routes.test.js', () => {
           const users = res.text.split('\n').slice(1);
           expect(users.length).to.equal(4);
           users.sort((a,b) => (a.substr(22,1) < b.substr(22,1)) ? -1 : 1);
-          expect(users[0].split(",")[8]).to.equal('"contributor"');
-          expect(users[1].split(",")[8]).to.equal('"sponsor"');
-          expect(users[2].split(",")[8]).to.equal('"host"');
-          expect(users[3].split(",")[8]).to.equal('"backer"');
+          console.log(users)
```
sedubois: remove spurious line

server/routes.js
asood123: you are right. And we don't need a separate method for this, it should be a filter on the getUsers method in groups controller. I'll rework.

asood123:
merging this since seb reviewed it once already.

PR #543, opened 2016-08-17
[Don't merge] Split donations flow

asood123:
This splits the donation flow into two parts:

1) When the user hits `donate`, we store the user's intent with all the info (`stripeToken`, `amount`, `interval`, `currency`, `groupid`, `userid`, etc) in `Donations`. 

2) An `afterCreate` hook on `Donations` then processes that donation. It creates a customer, charge, subscription plan, etc on Stripe. It also then sends out the confirmation email.

@sedubois @xdamman can you review the general setup of the new flow? Writing tests for this is a fair amount of work. Want to make sure this makes sense before I invest that time.

One problem that I can already see: stripe errors like `card_declined` will need to be handled differently. We'll probably need to email the user and get them to reenter the info. We'd have to do that anyway in case of `pending_collective`.


xdamman:
there is no stripe call that you can do to verify the card without making a charge? I thought there was.

server/controllers/donations.js
```diff
@@ -58,253 +55,52 @@ module.exports = (app) => {
...
+      req.group.getStripeAccount(),
+      models.PaymentMethod.getOrCreate({
+        token: payment.stripeToken,
+        service: 'stripe',
+        UserId: user.id })
+      ])
+    .then(results => {
+      const stripeAccount = results[0];
+      if (!stripeAccount || !stripeAccount.accessToken) {
+        return Promise.reject(new errors.BadRequest(`The host for the collective id ${req.group.id} has no Stripe account set up`));
```
xdamman: why not logging the slug of the group? Easier to read.

server/controllers/donations.js
```diff
@@ -58,253 +55,52 @@ module.exports = (app) => {
...
+      } else if (process.env.NODE_ENV !== 'production' && _.contains(stripeAccount.accessToken, 'live')) {
+        return Promise.reject(new errors.BadRequest(`You can't use a Stripe live key on ${process.env.NODE_ENV}`));
```
xdamman: I was wondering about how to handle errors properly within promises. 
Is this the best way?
What about simply throwing the error: `throw new errors.BadRequest()`?

asood123:
Was thinking about that too. There must be a call to verify card validity. At some point, it'll be nice to make all stripe calls async though. Really fast to donate.

server/models/Group.js
```diff
@@ -223,6 +223,22 @@ module.exports = function(Sequelize, DataTypes) {
         });
       },
 
+      findOrAddUserWithRole(user, role) {
```
xdamman: if you say find or add user, then I expect this to return a User.
So you should add `include: [ {model: Sequelize.models.User } ]` and return `userGroup.User`

Where is this used?

I wonder if it wouldn't be simpler to just do `addUserWithRole` with the edge case of the user already being in the group for that role to simply return as if it wasn't already there.

server/lib/processDonation.js
```diff
@@ -0,0 +1,134 @@
+const gateways = require('../gateways');
+const constants = require('../constants/transactions');
+const roles = require('../constants/roles');
+const emailLib = require('./email');
+
+module.exports = (Sequelize, donation) => {
+
+  const services = {
+
+    stripe: (donation) => {
```
xdamman: This file is hard to read.
I wonder if it would be easier to create a new file `stripe.js` with all this and exposing a few methods that don't depend on the model. That will be easier to create other providers.

server/lib/processDonation.js
```diff
@@ -0,0 +1,134 @@
...
+      const group = donation.Group;
+      const user = donation.User;
+      const paymentMethod = donation.PaymentMethod;
+      const subscription = donation.Subscription;
+      var groupStripeAccount, charge;
+
+      return group.getStripeAccount()
```
xdamman: there is a lot of `then` in this method. Can we break it down?

It feels like anything within a `then` should be a one-three liner.

server/lib/processDonation.js
```diff
@@ -0,0 +1,134 @@
...
+          groupStripeAccount,
+          paymentMethod.token, {
+            email: user.email,
+            group
+          }))
+        .tap(customer => paymentMethod.customerId ? null : paymentMethod.update({customerId: customer.id}))
+
+        // create plan + subscription or a charge for one-time
+        .then(() => {
+          // create a plan + subscription
```
xdamman: those comments are the proof that this should be put aside as a separate function `createPlan`, `createOneTimeCharge`, etc.

xdamman:
Since you are decoupling the creation of a donation to the payment of it, we should add a `paidAt` timestamp to the `Donation` model.

server/controllers/donations.js
asood123: That might work too. I have generally using `.reject` and `.resolve` to be sure.

server/models/Group.js
asood123: It's used in the new flow. `findAndAddUser` seems better name and I think you are right, it should return a User. I'll make the change.

asood123:
That's a good idea. Instead of `paidAt`, I'll do a `processedAt` timestamp.

server/controllers/donations.js
asood123: changed

asood123:
Doesn't look like there is a stripe call to verify the credit card

asood123:
This one is ready to be reviewed @xdamman @sedubois. Changes stripe donation flow altogether. Only case that isn't handled well is `card_declined`. Since it's infrequent, we can email those people individually. And eventually create an email template to send to them.

Note that even today when someone has a recurring subscription and their card declines, we don't do anything about it.

server/controllers/donations.js
```diff
@@ -58,253 +55,52 @@ module.exports = (app) => {
...
+      req.group.getStripeAccount(),
+      models.PaymentMethod.getOrCreate({
+        token: payment.stripeToken,
+        service: 'stripe',
+        UserId: user.id })
+      ])
+    .then(results => {
+      const stripeAccount = results[0];
+      if (!stripeAccount || !stripeAccount.accessToken) {
+        return Promise.reject(new errors.BadRequest(`The host for the collective slug ${req.group.slug} has no Stripe account set up`));
```
sedubois: I think instead of `return Promise.reject(error)` can just `return error`

server/controllers/donations.js
```diff
@@ -58,253 +55,52 @@ module.exports = (app) => {
...
+      } else if (process.env.NODE_ENV !== 'production' && _.contains(stripeAccount.accessToken, 'live')) {
+        return Promise.reject(new errors.BadRequest(`You can't use a Stripe live key on ${process.env.NODE_ENV}`));
+      } else {
+        paymentMethod = results[1];
+        return Promise.resolve();
```
sedubois: can just `return null` too

server/controllers/donations.js
```diff
@@ -58,253 +55,52 @@ module.exports = (app) => {
...
+
+    // fetch Stripe Account and get or create Payment Method
+    return Promise.all([
+      req.group.getStripeAccount(),
+      models.PaymentMethod.getOrCreate({
+        token: payment.stripeToken,
+        service: 'stripe',
+        UserId: user.id })
+      ])
+    .then(results => {
```
sedubois: or `tap` as you don't return anything

server/controllers/donations.js
```diff
@@ -58,253 +55,52 @@ module.exports = (app) => {
...
-          paymentMethod
-        };
-
-        payload.transaction = {
-          type: constants.type.DONATION,
-          DonationId: results.createDonation.id,
+    var paymentMethod;
+
+    // fetch Stripe Account and get or create Payment Method
+    return Promise.all([
```
sedubois: `Promise.props` has a bit cleaner API (like used in `slack-report.js`)

migrations/20160822191515-processedAt-in-Donation-migration.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface, Sequelize) {
+    return queryInterface.addColumn(
+      'Donations',
+      'processedAt',
```
sedubois: what is the use of `processedAt`? There is already `isProcessed` to know if donation is processed. And there's already an `updatedAt` field. Also the `Expense` model uses a `status` (`PENDING`, `APPROVED`, etc). Should it be similar for donations (also for symmetry) instead of a boolean field?

server/controllers/donations.js
```diff
@@ -58,253 +55,52 @@ module.exports = (app) => {
...
+      } else if (process.env.NODE_ENV !== 'production' && _.contains(stripeAccount.accessToken, 'live')) {
+        return Promise.reject(new errors.BadRequest(`You can't use a Stripe live key on ${process.env.NODE_ENV}`));
+      } else {
+        paymentMethod = results[1];
+        return Promise.resolve();
+      }
+    })
+    // create a new subscription
+    // (this needs to happen first, because of hook on Donation model)
+    .then(() => {
```
sedubois: can `tap` then no need for `return Promise.resolve()`

server/controllers/donations.js
```diff
@@ -58,253 +55,52 @@ module.exports = (app) => {
...
+    .then(subscription => models.Donation.create({
+        UserId: user.id,
+        GroupId: group.id,
+        currency: currency,
+        amount: amountInt,
+        title: `Donation to ${group.name}`,
+        PaymentMethodId: paymentMethod.id,
+        SubscriptionId: subscription && subscription.id
+      }))
+    .then(() => res.send({success: true, user: req.user.info, hasFullAccount: hasFullAccount}))
```
sedubois: `success: true` probably not very useful

server/controllers/donations.js
```diff
@@ -333,13 +129,17 @@ module.exports = (app) => {
...
-          })
-        .then(subscription => cb(null, subscription))
-        .catch(cb)
+        if (isSubscription) {
+          models.Subscription.create({
+              amount: amountFloat,
+              currency,
+              interval
+            })
+          .then(subscription => cb(null, subscription))
```
sedubois: Can probably simplify with `asCallback` http://bluebirdjs.com/docs/api/ascallback.html

server/lib/donation.js
```diff
@@ -0,0 +1,156 @@
+const gateways = require('../gateways');
+const constants = require('../constants/transactions');
```
sedubois: (To be symmetrical with line below, could be called `transactions`?)

server/lib/donation.js
```diff
@@ -0,0 +1,156 @@
...
+              description: `OpenCollective: ${group.slug}`,
+              application_fee: parseInt(donation.amount*constants.OC_FEE_PERCENT/100, 10),
+              metadata: {
+                groupId: group.id,
+                groupName: group.name,
+                customerEmail: user.email,
+                paymentMethodId: paymentMethod.id
+              }
+            })
+          .tap(c => charge = c)
```
sedubois: you don't use `charge` except in next line, so line seems unnecessary

server/lib/donation.js
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+  return Sequelize.models.Donation.findById(donation.id, {
+    include: [{ model: Sequelize.models.User },
+              { model: Sequelize.models.Group },
+              { model: Sequelize.models.PaymentMethod },
+              { model: Sequelize.models.Subscription }]
+    })
+    .then(donation => {
+      if (!donation.PaymentMethod || (donation.PaymentMethod && donation.PaymentMethod.service === 'paypal')) {
```
sedubois: `donation.PaymentMethod && ` is always true, can remove

server/models/Donation.js
```diff
@@ -1,4 +1,5 @@
 const type = require('../constants/transactions').type.DONATION;
+const donationLib = require('../lib/donation');
```
sedubois: should it be `donations.js`? We seem to use both plural and singular in libs; controllers are all plural.

server/models/Group.js
sedubois: Do you still intend to do this rename to `findAndAddUser`?

test/donations.routes.test.js
```diff
@@ -27,11 +27,12 @@ const userData = utils.data('user3');
 const groupData = utils.data('group2');
 const models = app.set('models');
 const stripeMock = require('./mocks/stripe');
+const paypalNock = require('./mocks/paypal.nock');
 
 /**
  * Tests.
  */
-describe('donations.routes.test.js', () => {
+describe.only('donations.routes.test.js', () => {
```
sedubois: remove?

test/donations.routes.test.js
```diff
@@ -643,172 +327,98 @@ describe('donations.routes.test.js', () => {
...
-        amount: data.amount * 100
-      });
 
-      var plan = _.extend({}, stripeMock.plans.create, {
-        amount: data.amount,
-        interval: data.interval,
-        name: planId,
-        id: planId
-      });
+      beforeEach((done) => {
```
sedubois: no need for `done`, just `return request(app).post(...).send(...).expect(200)`.

test/donations.routes.test.js
```diff
@@ -643,172 +327,98 @@ describe('donations.routes.test.js', () => {
...
-              done();
-            });
-        });
-
-        it('creates a plan if it doesn\'t exist', () => {
-          expect(nocks['plans.retrieve'].isDone()).to.be.true;
-          expect(nocks['plans.create'].isDone()).to.be.true;
-        });
-
+      it('successfully creates a donation in the database', (done) => {
```
sedubois: similar comment as above, mocha supports promises

test/donations.routes.test.js
```diff
@@ -643,172 +327,98 @@ describe('donations.routes.test.js', () => {
...
-            .then((res) => {
-              expect(res.count).to.equal(0);
-              done();
-            })
-            .catch(done);
-        });
-
+      it('does not create a transaction', (done) => {
+        models.Transaction
+          .findAndCountAll({})
```
sedubois: I guess can just `.count()` ?

test/donations.routes.test.js
```diff
@@ -832,6 +442,7 @@ describe('donations.routes.test.js', () => {
         });
 
         it('creates a transaction and returns the links', (done) => {
+          console.log("\n\n\n", nock.pendingMocks());
```
sedubois: intended?

test/donations.routes.test.js
```diff
@@ -957,13 +567,14 @@ describe('donations.routes.test.js', () => {
...
               links = res.body.links;
+              console.log('links: ', links)
               done();
             });
         });
 
         it('creates a transaction and returns the links', (done) => {
           const redirect = _.find(links, { method: 'REDIRECT' });
-
+          console.log(redirect);
```
sedubois: intended?

test/lib.donation.test.js
```diff
@@ -0,0 +1,486 @@
...
+    beforeEach(() => {
+      return models.Donation.create({
+        amount: 100,
+        currency: 'USD',
+        SubscriptionId: null,
+        PaymentMethodId: null
+      })
+    });
+
+    it('isProcessed and processedAt should not be false and null', done => {
```
sedubois: no need for `done`, can return promise, and similar for other tests

migrations/20160822191515-processedAt-in-Donation-migration.js
asood123: processedAt is mostly another check to let us know when it happened. Since we are often messing with the data manually in Postico, I haven't been trusting `updatedAt` very much.

server/lib/donation.js
asood123: Used in line 79 as well.

asood123:
Going to deploy this on Monday, when I get to Brussels.

PR #545, opened 2016-08-19
Mailgun

xdamman:


xdamman:
So the tests fail because on Circle CI it doesn't have the Mailgun API KEY. That shouldn't be a problem because normally `nock` is supposed to intercept the call but for some reason it doesn't. I'm a bit stuck. If anyone can take a fresh look at this that would help.

asood123:
@xdamman nocks are only being imported once at the top and being cleaned after each test. So, any test after the first one won't be able to use them. Need to re-add them in beforeEach. That should fix it.

test/email.routes.test.js
```diff
@@ -0,0 +1,161 @@
...
+        const emailSentTo = [spy.args[0][3].bcc,spy.args[1][3].bcc];
+        expect(emailSentTo.indexOf(usersData[0].email) !== -1).to.be.true;
+        expect(emailSentTo.indexOf(usersData[1].email) !== -1).to.be.true;
+        done();
+      });
+  });
+
+  it("approves the email", (done) => {
+
+    const spy = sandbox.spy(emailLib, 'send', (template, recipient, data) => {
```
asood123: I'm not sure that the third variable (the function) is needed here. Spies can't replace the original method, afaik. I used `const spy = sandbox.spy(emailLib, 'send')`.

package.json
```diff
@@ -11,15 +11,14 @@
...
-    "babel": "5.8.38",
-    "babel-cli": "^6.11.4",
-    "babel-core": "6.7.7",
-    "babel-eslint": "6.1.2",
-    "babel-plugin-add-module-exports": "0.1.4",
-    "babel-plugin-lodash": "2.3.0",
-    "babel-preset-es2015": "6.6.0",
-    "babel-preset-stage-0": "6.5.0",
-    "babel-register": "6.7.2",
+    "babel": "^6.5.2",
```
asood123: Can you freeze all these dependencies?

scripts/subscribe_to_mailinglist.js
```diff
@@ -0,0 +1,39 @@
...
+  debug(`Subscribing UserId ${row.UserId} to ${type} of GroupId ${row.GroupId}`);
+  return Notification.create({
+    UserId: row.UserId,
+    GroupId: row.GroupId,
+    type
+  }).catch(() => console.error(`UserId ${row.UserId} already subscribed to ${type} of GroupId ${row.GroupId}`));
+};
+
+
+init();
```
asood123: When is this script run?

xdamman: manually, only needs to be run once.

package.json
xdamman: by defining the smallest number, it is frozen.

test/email.routes.test.js
xdamman: ok

server/controllers/services/email.js
```diff
@@ -0,0 +1,175 @@
...
+
+  const models = app.set('models');
+  const errors = app.errors;
+
+  // TODO: move to emailLib.js
+  const sendEmailToList = (to, email) => {
+    const tokens = to.match(/(.+)@(.+)\.opencollective\.com/i);
+    const list = tokens[1];
+    const slug = tokens[2];
+    email.from = email.from || `${slug} collective <info@${slug}.opencollective.com>`;
```
asood123: instead of `wwcodeaustin collective`, it's better to put the group's real name in there.

package.json
xdamman: We should remove all carets: http://stackoverflow.com/questions/22343224/difference-between-tilde-and-caret-in-package-json

asood123: that's right. @sedubois has been calling it out on code reviews for a while now.

scripts/subscribe_to_mailinglist.js
asood123: How will new backers will get added to mailing lists?

server/controllers/services/email.js
```diff
@@ -0,0 +1,175 @@
...
+      let approver, sender;
+      let email = {};
+
+      const fetchSenderAndApprover = (email) => {
+        const where = { '$or': [ {email: approverEmail}, { email: email.sender } ] };
+        sender = { name: email.From, email: email.sender }; // default value
+        return models.User.findAll({ where })
+                .then(users => { 
+                  users.map(user => {
+                    if (approverEmail == user.email) approver = user;
```
asood123: In general, I'm told that `===` is better than `==`?

server/controllers/services/email.js
```diff
@@ -0,0 +1,175 @@
...
+      if (users.length === 0) throw new errors.NotFound(`There is no user subscribed to ${email.recipient}`);
+    })
+    .then(results => results.map(r => r.User.email))
+    .then(recipients => {
+      console.log(`Sending email from ${email.from} to ${to} (${recipients.length} recipients)`);
+      if (email.template) {
+        console.log(`preview: http://localhost:3060/templates/email/${email.template}?data=${encodeURIComponent(JSON.stringify(email))}`);
+        return emailLib.send(email.template, to, email, { from: email.from, bcc: recipients.join(',') });
+      } else {
+        console.log("Subject: ", email.subject);
```
asood123: Do we need to leave all these console.log statements in there?

server/controllers/services/email.js
asood123: same comment on `==` vs `===`

server/controllers/services/email.js
```diff
@@ -0,0 +1,175 @@
...
+        auth: {
+          user: 'api',
+          pass: config.mailgun.api_key
+        }
+      };
+
+      request
+      .get(`https://so.api.mailgun.net/v3/domains/opencollective.com/messages/${messageId}`, requestOptions)
+      .then(json => {
+        email = json; return email;
```
asood123: two lines is better than one, even if short.

server/controllers/services/email.js
```diff
@@ -0,0 +1,175 @@
...
+      .then(fetchSenderAndApprover)
+      .then(() => {
+        const emailData = {
+          template: 'email.message',
+          subject: email.Subject,
+          body: email['body-html'] || email['body-plain'],
+          to: email.To,
+          sender: _.pick(sender, ['email', 'name', 'avatar'])
+        }
+        if ( approver && approver.email != sender.email )
```
asood123: same goes for `!=`. http://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons

server/controllers/services/email.js
```diff
@@ -0,0 +1,175 @@
...
+
+    // If an email is sent to info@:slug.opencollective.com,
+    // we simply forward it to members who subscribed to that list
+    if (list === 'info') {
+      return sendEmailToList(recipient, { 
+        subject: email.subject, 
+        body,
+        from: email.from 
+      })
+      .then(() => res.send('ok'))
```
asood123: Is it better to return json or a string? or a 200 status?

scripts/subscribe_to_mailinglist.js
xdamman: https://github.com/OpenCollective/opencollective-api/blob/6027f2e3950a60897fc27aeae4f3b3901c7d5f44/server/controllers/groups.js#L42-L53

server/controllers/services/email.js
xdamman: ok

server/controllers/services/email.js
xdamman: ok

server/controllers/services/email.js
xdamman: this returns a 200 status.

PR #547, opened 2016-08-24
Run esnext on lib directory

sedubois:
Auto upgrading to latest ES syntax using esnext. Experimented with server/libs (some extra fixes were necessary in non standard code):
`find server/lib/*.js -exec esnext -I {} \;`

Can run on other places if you like the principle:
`find . -name "*.js" -not -path "./node_modules/*" -exec esnext -I {} \;`

BTW wonder why eslint still allowed various `var` to be in the code

@asood123 @xdamman

### DIFF
--- a/server/lib/activities.js
+++ b/server/lib/activities.js
@@ -1,29 +1,29 @@
-const activities = require('../constants/activities');
-const flatten = require('flat');
-const currencies = require('../constants/currencies');
+import activities from '../constants/activities';
+import flatten from 'flat';
+import currencies from '../constants/currencies';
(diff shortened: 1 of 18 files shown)


asood123:
@sedubois It's because of this line that overrides the original config: https://github.com/OpenCollective/opencollective-api/blob/master/.eslintrc#L6. At the time, we had to upgrade node to use `let`. Shouldn't be a problem now, so we can remove it.

asood123:
Looks good to me, so feel free to run it everywhere

PR #550, opened 2016-08-24
Group settings update:  provide route + don't override pre-existing JSON data

sedubois:


server/controllers/groups.js
```diff
@@ -446,9 +446,21 @@ module.exports = function(app) {
...
+  const updateSettings = (req, res, next) => {
+    doUpdate(['settings'], req, res, next);
+  };
+
+  function doUpdate(whitelist, req, res, next) {
     whitelist.forEach((prop) => {
       if (req.required.group[prop]) {
-        req.group[prop] = req.required.group[prop];
+        if (typeof req.group[prop] === 'object') {
+          req.group[prop] = _.defaults(req.required.group[prop], req.group[prop]);
```
asood123: Is this flipped? According to this: https://lodash.com/docs#defaults, the destination object `req.group[prop]` is the first param.

sedubois: Right, I didn't see it mutates the first param, I'll fix.

server/controllers/groups.js
```diff
@@ -446,9 +446,21 @@ module.exports = function(app) {
...
+  const updateSettings = (req, res, next) => {
+    doUpdate(['settings'], req, res, next);
+  };
+
+  function doUpdate(whitelist, req, res, next) {
     whitelist.forEach((prop) => {
       if (req.required.group[prop]) {
-        req.group[prop] = req.required.group[prop];
+        if (typeof req.group[prop] === 'object') {
+          req.group[prop] = Object.assign(req.group[prop], req.required.group[prop]);
```
sedubois: Crazily, it's not enough to write `Object.assign(req.group[prop], req.required.group[prop])`, the result needs to be reapplied: `req.group[prop] = ...`. Otherwise sequelize doesn't detect the javascript field got updated, and doesn't effectively store it in DB, although the new data is returned by the `save()`!

PR #551, opened 2016-08-24
[in progress] Send an email to the core members of the collective to approve expense

xdamman:
Fixes https://github.com/OpenCollective/opencollective

It uses JWT token to automatically log in the website. 
So we need a new page on the website `/:slug/expenses/:expenseId/approve` that would approve the expense if logged in and then redirect to `/:slug/expenses/:expenseId`.

@carlosascari is that something that you can do?


carlosascari:
Sure can however I don't believe I'll get to it this week, is that alright?

asood123:
Already did this... closing.

PR #552, opened 2016-08-25
Mailinglist

xdamman:


server/controllers/services/email.js
```diff
@@ -207,7 +220,30 @@ module.exports = (app) => {
         return Promise.map(users, (user) => emailLib.send('email.approve', `members@${slug}.opencollective.com`, getData(user), {bcc:user.email}));
       })
       .then(() => res.send('Mailgun webhook processed successfully'))
-      .catch(next);
+      .catch(e => {
+        switch (e.message) {
+          case 'no_subscribers':
```
sedubois: these constants should be extracted

test/email.routes.test.js
```diff
@@ -127,6 +129,20 @@ describe("email.routes.test", () => {
       });
   });
 
+  it("rejects emails sent to unknown mailing list", (done) => {
+
+    const unknownMailingListWebhook = _.defaults({ recipient: 'unknown@testcollective.opencollective.com' }, webhookBody);
```
sedubois: actually I guess it's the same to use `Object.assign` (no need for external lib)?

test/email.routes.test.js
```diff
@@ -127,6 +129,20 @@ describe("email.routes.test", () => {
       });
   });
 
+  it("rejects emails sent to unknown mailing list", (done) => {
```
sedubois: no need for `done`, mocha handles promises:

```
it(..., () => request(app)
  .post(...)
  .send(...)
  .then(res => {
    expect(...);
    ...
  });
```

server/controllers/services/email.js
xdamman: there is no point if they are only used once.

test/email.routes.test.js
xdamman: You are right! I'll change it.

test/email.routes.test.js
xdamman: I didn't know. Ok, changed.

server/controllers/services/email.js
sedubois: But they are used twice, e.g lines 194 + 225. If you change one without the other, code breaks.

xdamman: but it's only in one file. Honestly, in the age of modern IDE and autocomplete, it doesn't really matter if it's in one single file.

PR #554, opened 2016-08-25
Upgrade models Javascript syntax with esnext + manual tweaks

sedubois:


server/models/Activity.js
```diff
@@ -15,7 +15,7 @@ module.exports = function(Sequelize, DataTypes) {
     updatedAt: false,
 
     hooks: {
-      afterCreate: function(activity) {
+      afterCreate(activity) {
```
asood123: Did you test this? Would have thought you would need to do `afterCreate: activity => {`

asood123: Looking below, I guess this should work

PR #555, opened 2016-08-25
Upgrade controllers Javascript syntax with esnext + manual tweaks

sedubois:


server/controllers/middlewares.js
```diff
@@ -77,20 +79,20 @@ module.exports = function(app) {
...
 
-      var name, email, paypalEmail;
+      let name, email, paypalEmail;
 
       // TODO remove #postmigration, replaced by req.body.expense
       if (req.body.transaction) {
-        email = req.body.transaction.email;
-        paypalEmail = req.body.transaction.paypalEmail;
-        name = req.body.transaction.name;
+        ({ email } = req.body.transaction);
```
asood123: didn't know you could do that. It treats all local vars as part of a single object? Are the parens on the outside needed?

asood123:
Also, is there a way to enforce this syntax? We should add it to the eslint rules.

server/controllers/middlewares.js
sedubois: Didn't know either ;) and yes, the parentheses are expected, otherwise syntax error

sedubois:
I'll have a look for eslint rules

PR #561, opened 2016-08-30
Upgrade linting rules to ES6 + check imports

sedubois:


server/controllers/services/email.js
```diff
@@ -71,7 +71,7 @@ export default (app) => {
 
     return fetchSubscribers(slug, type)
     .tap(subscribers => {
-      if (subscribers.length === 0) return mailinglistNotFound(recipient);
```
sedubois: FYI @xdamman linting picked up `mailinglistNotFound` as being undefined. Please provide PR if the way I changed isn't appropriate

PR #564, opened 2016-08-30
add build to postinstall step

asood123:


asood123:
Build passed. Had to cancel to kill the ssh session. Merging as it's already reviewed by @sedubois

PR #566, opened 2016-08-30
move errorhandler to production

asood123:


asood123:
1-line package.json change. self-merging :)

PR #568, opened 2016-08-31
ability to add a group if you are an internal user

asood123:


server/controllers/groups.js
```diff
@@ -291,35 +291,28 @@ export default function(app) {
...
-
-        const options = {
-          role,
-          remoteUser: req.remoteUser
-        };
-
-        return _addUserToGroup(group, req.remoteUser, options);
+      .tap(g => {
+        return Promise.map(users, user => {
+          return User.findOne({where: { email: user.email.toLowerCase() }})
```
sedubois: we have upgraded Sequelize, so now it should be possible to use `findOrCreate`

server/middleware/security/authentication.js
```diff
@@ -220,6 +221,35 @@ export default function (app) {
...
+              }
+              this._authenticateInternalUserById(req, res, next);
+            });
+          });
+        });
+      }
+    },
+
+    _authenticateInternalUserById: (req, res, next) => {
+      if (_.contains([1,2,8,30,40,212,772], req.jwtPayload.sub)) {
```
sedubois: could be extracted to constants if wished (IDs are also in weekly report script)

test/groups.routes.test.js
```diff
@@ -61,6 +66,17 @@ describe('groups.routes.test.js', () => {
         });
     });
 
+    it('fails creating a group if token but a different id', (done) => {
```
sedubois: no need for `done` and `expect(e).to.not.exist`, mocha handles promises: 

```
it('...', () => request(app)
  .post(...)
  .set(...)
  .expect(401));
```

PR #573, opened 2016-09-02
Put opt-in config for thank you tweet after donation into Notification table instead of Group.settings

sedubois:


server/models/Group.js
```diff
@@ -483,4 +486,13 @@ export default function(Sequelize, DataTypes) {
   });
 
   return Group;
+
+  function isThankDonationEnabled(GroupId) {
+    return models.Notification.findOne({where: {
+      channel: 'twitter',
+      type: activities.GROUP_TRANSACTION_CREATED,
```
asood123: this isn't a `group_transaction_created`. Can we create a new type of activity for it?

PR #574, opened 2016-09-04
few backend tweaks for create group flow

asood123:


server/lib/utils.js
```diff
@@ -193,6 +193,16 @@ export const defaultHostId = () => {
 };
 
 /**
+ * Demo host id, set this for new groups created through the flow
+ */
+export const demoHostId = () => {
+  if (process.env.NODE_ENV === 'production' || process.env.NODE_ENV  === 'staging') {
+    return 254;
```
sedubois: Just noting this kind of hard-coding will break the day another DB instance would be spun up

test/groups.routes.test.js
```diff
@@ -141,46 +141,6 @@ describe('groups.routes.test.js', () => {
         .end(done);
     });
 
-
-    it('successfully create a group without assigning a member', (done) => {
```
sedubois: why no tests?

asood123: Removed the test initially thinking it didn't apply. Changed flow later. Fixed.

PR #575, opened 2016-09-05
fix babel-node bug with weekly scripts

asood123:


asood123:
since it's very limited change in package.json and blocking weekly script, merging it myself. @sedubois pls review post-merge.

PR #576, opened 2016-09-06
Update lodash dependency to 4.15.0

datapimp:
This updates the project's version of lodash from a 3.x version to the latest 4.x version.

The main difference is that the lodash project has moved away from the lodash/object/module style of import to the more flattened lodash/module style. I updated all of our imports to be inline with this change.

There are also a couple of deprecations -- namely pluck, which is replaced by map; It has the same exact argument signature / behavior.

And the contains function which we use to test for the existence of a pattern in a string. I went with using a regex on the value, after ensuring it is a string, which should provide the same results.

Please note the babel plugins were also updated in this PR, as we use the lodash babel plugin I went ahead and updated them all to their latest versions


datapimp:
closing the PR from a fork

PR #578, opened 2016-09-07
Remove unused scripts

sedubois:


asood123:
@sedubois we still need some of these. Like `check_transactions_on_stripe` is one we run manually to match against stripe, `replace_stripe_accounts` is used when we copy databases (it's in package.json), etc.

I'd vote to leave these as they are. If they are creating work for you while cleaning up, leave a comment at the top of what needs to happen and whoever runs it next can fix it.

sedubois:
sure I'll decline the PR

PR #586, opened 2016-09-09
Propsal for a universal API client

datapimp:
This uses the [Axios](https://github.com/mzabriskie/axios) library as the basis
for a distributable / universal API client that can be loaded as a dependency 
by the `opencollective-website` project or any other projects in the future.

This will allow us to move the hydrating of data for server side rendering into 
the same module that is used in the client-side context, which will allow us to
eliminate the duplication of the routing tables that currently exists in that project.

[Current Status](https://github.com/OpenCollective/opencollective-website/blob/master/server/src/routes.js#L65)

**Note: This is just a rough draft of how something like this might look.**

This client would be used by Redux, Express, and other components of the application and would free
them from having to care about the details of the REST API and things like that. Just use the promises API.

The general idea is our client is a simple wrapper around axios that can be loaded with references to different `Services` for the lack of a better term.  These can be hard-coded or even dynamically generated for specific types of clients. For example we could distribute a public client which only implements the non-authenticated requests, or a private client which has more features for users who are logged in, or an admin only client, etc.

I would love to eventually have this client code be generated along side interactive API documentation that is produced by an automated integration test suite with it.

This provides an abstract localStorage API that works in all environments.
- [ ] Complete implementation of public services (groups, activeUsers, leaderboards)
- [ ] Replace `opencollective-website` express middleware implementations with an instance of the client
- [ ] Move data fetching responsibilities into the react containers in `opencollective-website`
- [ ] Develop test suite

### DIFF
--- a/client/src/index.js
+++ b/client/src/index.js
@@ -1,0 +1,72 @@
+import axios from 'axios'
+
+const defineProperty = Object.defineProperty
+
+const isBrowser = typeof window !== 'undefined' && typeof document !== 'undefined'
+
+export class ApiClient {
+  constructor(options = {}) {
+
+    defineProperty(this, 'options', {
+      enumerable: false,
+      value: ({ ...DefaultOptions, ...options })
+    })
+
+    const storage = isBrowser ? window.localStorage : require('./storage')()
+
+    defineProperty(this, 'storage', {
+      enumerable: false,
+      value: storage
+    })
+
+    defineProperty(this, 'api', {
+      enumerable: false,
+      value: axios.create({
+        baseURL: this.baseURL,
+        headers: this.authHeaders
+      })
+    })
+  }
+
+  setDefaults(values = {}) {
+    this.instance.defaults = {
+      ...this.instance.defaults,
+      ...values
+    }
+  }
+
+  get baseURL() {
+    const { protocol, host, prefix } = {
+      protocol: '',
+      prefix: '/api',
+      host: 'localhost',
+      ...this.options
+    }
+
+    return `${protocol}${host}${prefix}`
+  }
+
+  get headers() {
+    return {
+      ...this.authHeaders
+    }
+  }
+
+  get authHeaders() {
+    return this.isAuthenticated
+      ? { Authorization: `Bearer ${this.accessToken}` }
+      : {}
+  }
+
+  get accessToken() {
+    return this.storage.getItem(
+      this.options.accessTokenKey
+    )
+  }
+
+  get isAuthenticated() {
+    return !!this.accessToken
+  }
+}
+
+export default ApiClient
--- a/client/src/services/groups/index.js
+++ b/client/src/services/groups/index.js
@@ -1,0 +1,62 @@
+export function register (client, options = {}){
+  const get = client.api.get
+
+  client.register('groups', {
+    find: (slug) => (
+      get(builders.findBySlug(slug))
+    ),
+
+    details: (slug) => (
+      get(builders.findBySlug(slug)).then(user => presenters.details(user))
+    ),
+
+    activeUsers: (slug) => (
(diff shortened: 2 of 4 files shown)


datapimp:
This would make an implementation like 

```javascript

const addMeta = (req, res, next) => {
  const group = req.group;

  req.meta = {};
  if (!group.username) {
    req.meta = {
      url: group.publicUrl,
      title: `${group.name} is on Open Collective`,
      description: `${group.name} is on a mission to ${group.mission}`,
      image: group.image || group.logo,
      twitter: `@${group.twitterHandle}`,
    };
  } else {
    const user = req.group;
    let description = '';

    if (user.groups.length > 0) {
      const belongsTo = filterCollection(user.groups, { role: 'MEMBER' });
      const backing = filterCollection(user.groups, { role: 'BACKER' });

      if (belongsTo.length > 0) {
        description += `a member of ${belongsTo.length} collectives`;
      }
      if (backing.length > 0) {
        if (description.length > 0) description += ' and ';
        description += `supporting ${backing.length} collectives`;
      }

      user.groups.sort((a, b) => (b.members - a.members));
      description += ' including ';
      const includingGroups = user.groups.slice(0,3).map((g) => g.name);
      description += includingGroups.join(', ');
    }


    req.meta = {
      url: `${config.host.website}/${user.username}`,
      title: `${user.name} is on OpenCollective`,
      description: `${user.name} is ${description}`,
      twitter: `@${user.twitterHandle}`,
      image: user.avatar
    };
  }

  next();
};
```

look something like:

```javascript
import client from 'opencollective-api/client'

const groups = client.services.groups

const addMeta = (req, res, next) => {
  const groups = client.services.groups

  groups.details(req.params.slug)
  .then((details) => req.meta = details)
  .then(() => next())
  .catch((e) => next(e))
}
```

and the redux code would almost look identical / behave similarly in the SSR / Browser context, etc.

client/src/services/groups/index.js
```diff
@@ -0,0 +1,62 @@
+export function register (client, options = {}){
+  const get = client.api.get
```
sedubois: (just a detail, all the semicolons are missing)

sedubois:
I just wonder, if this is necessary right now during prototype phase. If we want to improve API etc, a broader discussion could be useful. (e.g I'm not even sure why we have separate web servers for API and website in the first place.)

client/src/index.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  setDefaults(values = {}) {
+    this.instance.defaults = {
+      ...this.instance.defaults,
+      ...values
+    }
+  }
+
+  get baseURL() {
```
datapimp: I am pleased then to introduce you to one of my favorite es6 language enhancements: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

datapimp:
@sedubois valid point -- I think having something like this would make us much more effective in the prototyping phase, though, because it reduces the amount of work and overhead involved in making changes by eliminating the need to dip through different layers of the stack. 

Plus Almost all of the code required to implement the pieces we need already exist and just need to be consolidated into this library, and getting this abstraction in place sooner than later will help keep up the speed.  

On the point of separate web servers -- the eventual goal of simplifying that setup in particular is what inspired this proposal; the API client extracts what is essential to what we have now

asood123:
@datapimp I'm not sure I follow the proposal. Let's chat live. 

Our codebase is a prototype. So, scaling and cleanups don't gain us much (since we don't know the final form of the product). Even stability isn't that important right now. Increasing iteration speed is key. Ability to hack new features quickly is paramount. That's why Webpack caught my eye. From what I saw in sample apps and demos, it can make UI iteration much faster.

PR #584, opened 2016-09-08
top related backers in monthly report

xdamman:
The goal is to show the top backers (sponsors) that have backed related collectives last month

PR #587, opened 2016-09-09
Remove interval transaction field

sedubois:


sedubois:
@asood123 @xdamman this also fails on master, so the branch is good to merge: https://circleci.com/gh/OpenCollective/opencollective-api/2716

Need to debug the error `expected 150 to equal 100` at `test/user.model.test.js:165:48` separately

asood123:
I had a lot of trouble with that error in this PR: https://github.com/OpenCollective/opencollective-api/pull/584. You can see the last few commits on how it appears and disappears. I was able to repro it as well.

server/lib/activities.js
```diff
@@ -181,15 +178,14 @@ export default {
...
+      const { interval } = activity.data.subscription;
       recurringAmount = amount + (interval ? `/${interval}` : '');
     }
 
     // get transaction data
     if (activity.data.transaction) {
       ({ amount } = activity.data.transaction);
-      ({ interval } = activity.data.transaction);
-      recurringAmount = amount + (interval ? `/${interval}` : '');
+      recurringAmount = amount;
```
asood123: Can you add `interval` line back in? Otherwise, we won't be able to tell on Slack whether it's a one-time or a recurring donation.

asood123:
Besides that one comment, rest lgtm.

server/lib/activities.js
sedubois: good catch

sedubois:
Fixed according to @asood123 comment but tests still fail due to unrelated https://github.com/OpenCollective/opencollective-api/pull/584

xdamman:
Problem is this: https://github.com/OpenCollective/opencollective-api/blob/7edc670549f7465981d8509f40c9cf0b1ec3b056/test/user.model.test.js#L98

we can't assume with `Promise.all` that those two groups will be created in a sequential way. So it works most of the time but not all the time. Let me see how I can fix that.

asood123:
@xdamman you can replace `promise.all` with `promise.each` in `createMany` function in `User.js`. It'll do sequential, instead of concurrent.

PR #595, opened 2016-09-15
Don't use Group.createMany

sedubois:


sedubois:
Merging to attempt to fix flaky tests on master

PR #598, opened 2016-09-16
Fix createmany test

asood123:


xdamman:
this is a nice refactoring 👍 

note though that this was not the cause of the failing tests. The error was coming from the createMany in the Group model. I'll make the change.

PR #599, opened 2016-09-16
exportCSV with or without email depending if logged in as admin

xdamman:
The `server/middleware/security/authorization.js` is a mess. We should refactor that at some point.
I found a few bugs because it's confusing.

Anyway, I'm running into the issue of "Can't set headers after they are sent". I've been looking into it for 30mn so I reached a point where I need a pair of fresh eyes.

Best way to reproduce quickly:

    mocha --require babel-core/register test/usergroup.routes.test.js -g "corresponding tier"


server/controllers/groups.js
```diff
@@ -87,6 +87,30 @@ const getUserData = (groupId, tiers) => {
 
 export const getUsers = (req, res, next) => {
   let promise = getUserData(req.group.id, req.group.tiers);
+
+  if (req.query.filter && req.query.filter === 'active') {
+    const now = moment();
+    promise = promise.filter(backer => now.diff(moment(backer.lastDonation), 'days') <= 90);
+  }
+
+  const isAdmin = !!(req.remoteUser && _.intersection(req.remoteUser.rolesByGroupId[req.group.id], ['HOST', 'MEMBER']).length > 0);
```
asood123: We are introducing another term. Let's stick to the ones we have. `isHostOrMember` is better.

xdamman: ok

xdamman: done

server/controllers/middlewares.js
```diff
@@ -152,6 +153,9 @@ export const apiKey = (req, res, next) => {
 
 /**
  * Authenticate.
+ * @PRE: Need to submit a login (username or email)/password as a POST request.
+ *       api_key is also required
+ * @POST: req.remoteUser and req.user are set to the logged in user
```
xdamman: Yes we should only have `req.remoteUser` but I didn't want to open that can of worms now.

test/usergroup.routes.test.js
```diff
@@ -437,8 +437,39 @@ describe('usergroup.routes.test.js', () => {
           expect(users[1].split(",")[9]).to.equal('"sponsor"');
           expect(users[2].split(",")[9]).to.equal('"host"');
           expect(users[3].split(",")[9]).to.equal('"backer"');
+          const headers = res.text.split('\n')[0];
+          expect(headers).to.not.contain('"email"');
         })
         .end(done);
     });
+
+    it('get the list of users in csv format without emails if logged in as backer', (done) => {
```
xdamman: agreed, not very important.

PR #601, opened 2016-09-16
Fix create many

xdamman:


server/models/Group.js
```diff
@@ -449,6 +449,10 @@ export default function(Sequelize, DataTypes) {
     },
 
     classMethods: {
+      createMany: (groups, defaultValues) => {
```
sedubois: I had bypassed this usage in the test, you can restore it if you want

PR #602, opened 2016-09-16
add sql file to scrub the database

asood123:


sedubois:
Should a check be added to ensure this isn't run in prod?

asood123:
@sedubois how would you limit to non-production database? It's not really a js script that can be run directly. Meant for us to copy/paste in postico for now. We'll find a better way later.

PR #603, opened 2016-09-18
First name

xdamman:
Splitting `User.name` into `User.firstName` and `User.lastName` as requested in https://github.com/OpenCollective/OpenCollective/issues/161


server/controllers/groups.js
```diff
@@ -87,6 +87,30 @@ const getUserData = (groupId, tiers) => {
 
 export const getUsers = (req, res, next) => {
   let promise = getUserData(req.group.id, req.group.tiers);
+
```
xdamman: it's because I've started this branch off of the `exportCSV` one. But since you already reviewed that one, as soon as it will be merged, you will only see the changed based on that.

PR #605, opened 2016-09-20
fix first name for github flow

xdamman:


xdamman:
Need this to be merged to make the frontend tests pass

PR #613, opened 2016-09-22
add a generic subscriptions link

asood123:


server/lib/donations.js
```diff
@@ -120,9 +121,10 @@ export const processDonation = (Sequelize, donation) => {
           { donation: donation.info,
             user,
             group,
-            interval: subscription && subscription.interval
+            interval: subscription && subscription.interval,
             // TODO: bring this back. Figure out how to pass the application link
```
xdamman: you can remove that todo no?

asood123: Leaving the todo because a better link would be one with a token, so you don't have to relogin.

PR #615, opened 2016-09-23
allow updating user info if logged in

xdamman:


server/middleware/security/authentication.js
```diff
@@ -53,7 +53,7 @@ export const parseJwtNoExpiryCheck = (req, res, next) => {
   } else {
     const header = req.headers && req.headers.authorization;
     if (!header) {
-      return next(new Unauthorized('Missing authorization header'));
```
xdamman: if the goal of the middleware is just to authenticate the user, it should never `return next(error);`, otherwise it stops the flow. The controller should simply check for `req.remoteUser`. Or use a `authorization` middleware.

server/routes.js
```diff
@@ -93,7 +93,7 @@ export default (app) => {
    */
   app.post('/users', aN.authenticateAppByApiKey, aZ.appAccess(0.5), required('user'), users.create); // Create a user.
   app.get('/users/:userid', aN.authenticateUserOrApp(), users.show); // Get a user.
-  app.put('/users/:userid', aN.authenticateAppByApiKey, required('user'), aZ.authorizeAccessToUserWithRecentDonation, users.updateUserWithoutLoggedIn); // Update a user.
```
xdamman: if a middleware is only used once, we probably don't need to make it a middleware.
That's why I moved the logic of `aZ.authorizeAccessToUserWithRecentDonation` to the `users` controller (see above)

PR #622, opened 2016-09-29
setting the minimum amount to show up to $10 instead of $100

xdamman:


xdamman:
It's in cents and not in dollars, so the minimum is already $10.

carlosascari:
@xdamman The minimum is currently $1 not $10
https://github.com/OpenCollective/opencollective-api/blob/master/server/controllers/discover.js#L11

PR #623, opened 2016-09-29
Update user description

xdamman:
allow the frontend to create a user with a name that we will automatically split in firstName and lastName

### DIFF
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -83,6 +83,8 @@ export const updateUser = (req, res, next) => {
       updateFields: (cb) => {
         ['firstName',
         'lastName',
+        'description',
+        'longDescription',
         'twitterHandle',
         'website',
         'avatar'
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -92,6 +92,15 @@ export const updateUser = (req, res, next) => {
           if (req.required.user[prop]) {
             req.user[prop] = req.required.user[prop];
           }
+
+          // Split user.name into user.firstName and user.lastName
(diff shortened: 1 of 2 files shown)


server/controllers/users.js
```diff
@@ -83,13 +83,24 @@ export const updateUser = (req, res, next) => {
...
         'website',
         'avatar'
         ].forEach((prop) => {
           if (req.required.user[prop]) {
             req.user[prop] = req.required.user[prop];
           }
+
+          // Split user.name into user.firstName and user.lastName
+          if (req.required.user.name) {
+            const nameTokens = req.required.user.name.split(' ');
```
carlosascari: I believe this is error prone. 

What if someone uses a middle initial, or a second last name? 
What if someone uses their full name, if the have a second/middle name it will become the last name.

Now, I suppose the possibilities I mentioned above, can be tucked under the rug for the time being, and all is fine.

However we should eliminate the case where two or more spaces are added by accident.
```
Xavier Damman => ['Xavier', 'Damman']
Xavier  Damman => ['Xavier', '',  'Damman']
```

By using regex: `/([\w-]+)/gi` You can forget about the spacing:

```
'     Xavier     Damman   '.match(/([\w-]+)/gi)  => ['Xavier', 'Damman']
```

In the future, **for scalability**, we will need to have separate inputs for first and last names
[It seems handling international names is not so intuitive](https://www.w3.org/International/questions/qa-personal-names)

xdamman:
The comment about extra white space is good.

Agreed on two inputs but we should then do the distinction between name for organizations and first name last name for people.

PR #624, opened 2016-10-02
Big refactoring: remove concept of application and group.isPublic

xdamman:
Better to review this commit by commit.

We should merge this ASAP otherwise we will never merge it.


server/controllers/expenses.js
```diff
@@ -35,7 +36,17 @@ export const create = (req, res, next) => {
 /**
  * Get an expense.
  */
-export const getOne = (req, res) => res.json(req.expense.info);
+export const getOne = (req, res) => {
+  auth.canEditExpense(req, res, (e, canEditExpense) => {
```
asood123: Seems weird to call auth from controller. We probably want to decorate req.remoteUser with this data?

asood123: Thinking more about it, it's hard to add that info to encompass all expenses. The right place is a lib function that both auth and controllers can use. `canEditExpense(user, expense)`.

server/controllers/groups.js
```diff
@@ -86,25 +86,25 @@ const getUserData = (groupId, tiers) => {
 };
 
 export const getUsers = (req, res, next) => {
-  let promise = getUserData(req.group.id, req.group.tiers);
-
-  if (req.query.filter && req.query.filter === 'active') {
-    const now = moment();
-    promise = promise.filter(backer => now.diff(moment(backer.lastDonation), 'days') <= 90);
-  }
+  auth.canEditGroup(req, res, (e, canEditGroup) => {
```
asood123: Same comment here. `canEditGroup(user, group)` should be a function in a lib to be used by both auth and controllers.

PR #627, opened 2016-10-03
Default logo

xdamman:


xdamman:
So what do you want me to do here?

Also, there is no web pack for API.

asood123:
I'm going to merge it, we'll deal with it later. 

Re: webpack for API. /static will still be served off of `website` repo, which has webpack and webpack uses `require('../assets/...png`). We'll find a way to hack this.

PR #628, opened 2016-10-04
bug in test controller

asood123:


asood123:
merging to unblock tests

PR #631, opened 2016-10-05
Fix e2e

xdamman:
We add a second expense that incurred before the first expense we created


xdamman:
Merging because I need this to continue the work on the e2e

PR #633, opened 2016-10-07
Group history

asood123:


migrations/20161007172819-lastEditedBy-in-Groups-migration.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface, Sequelize) {
+    return queryInterface.addColumn(
+      'Groups',
+      'lastEditedById',
```
xdamman: I think we should name this `lastEditedByUserId` to be very explicit and avoid confusion

migrations/20161007175650-add-grouphistories-migration.js
```diff
@@ -0,0 +1,52 @@
...
+      backgroundImage: DataTypes.STRING,
+      hostFeePercent: DataTypes.FLOAT,
+      settings: DataTypes.JSON,
+      whyJoin: DataTypes.TEXT,
+      data: DataTypes.JSON,
+      tags: DataTypes.ARRAY(DataTypes.STRING),
+      isSupercollective: DataTypes.BOOLEAN,
+      lastEditedById: DataTypes.INTEGER,
+
+      hid: {
```
xdamman: what is `hid`?

asood123: this is used by temporal. It's the id of the row, since the original id field is now recorded separately here.

PR #634, opened 2016-10-07
making sure we can reset the password without being logged in

xdamman:


server/routes.js
```diff
@@ -44,8 +44,11 @@ export default (app) => {
   app.use('/status', serverStatus(app));
 
   /**
-   * Routes without expiration validation
+   * User reset password or new token flow (no jwt verification)
    */
+  app.post('/users/password/forgot', required('email'), users.forgotPassword); // Send forgot password email
+  app.post('/users/password/reset/:userid_enc/:reset_token', required('password', 'passwordConfirmation'), users.resetPassword); // Reset password`
+  app.post('/users/new_login_token', required('email'), users.sendNewTokenByEmail);
```
asood123: shouldn't all of these still need an apikey? Otherwise, someone could keep sending email addresses to them?

xdamman: Good catch

PR #638, opened 2016-10-10
default tiers

xdamman:


server/controllers/groups.js
```diff
@@ -306,6 +306,12 @@ export const create = (req, res, next) => {
     emailLib.send('group.created', user.email, data);
   };
 
+  // Default tiers
```
asood123: fyi, need to add this to `createFromGithub` function as well.

xdamman: Good catch. 

I wish there was an easy way to override `Group.create()` so that we can define in one place the default values (without forcing `group.tier` to always return the default value as we may want to allow a NULL value)

PR #639, opened 2016-10-10
show transaction description - useful for Add Funds

asood123:


server/controllers/groups.js
```diff
@@ -181,7 +181,7 @@ export const getTransactions = (req, res, next) => {
         Link: getLinkHeader(getRequestedUrl(req), req.pagination)
       });
 
-      res.send(transactions.rows.map(transaction => Object.assign({}, transaction.info, {'description': transaction.Donation && transaction.Donation.title})));
+      res.send(transactions.rows.map(transaction => Object.assign({}, transaction.info, {'description': (transaction.Donation && transaction.Donation.title) || transaction.description })));
```
xdamman: cannot you add that directly in the `Transaction.info`?

PR #641, opened 2016-10-11
adding model for comments

xdamman:
TODO:
- [ ] add migration file


server/controllers/middlewares.js
```diff
@@ -59,6 +60,36 @@ export const format = (format) => {
...
+  required('user')(req, res, (e) => {
+    if (e) return next(e);
+    console.log("Lookign for ", req.required.user.email);
+    User.findOne({
+      where: {
+        email: req.required.user.email.toLowerCase()
+      }
+    })
+    .then(user => {
+      console.log("Found: ", user);
```
asood123: need to remove console.log

asood123: need to remove console.log

server/controllers/middlewares.js
```diff
@@ -59,6 +60,36 @@ export const format = (format) => {
...
+ */
+export const authOrCreateUser = (req, res, next) => {
+  // If already logged in, proceed
+  if (req.remoteUser) {
+    req.user = req.remoteUser;
+    return next();
+  }
+  required('user')(req, res, (e) => {
+    if (e) return next(e);
+    console.log("Lookign for ", req.required.user.email);
```
asood123: need to remove.

server/routes.js
```diff
@@ -179,6 +181,15 @@ export default (app) => {
   app.post('/groups/:groupid/expenses/:expenseid/pay', auth.mustHaveRole(roles.HOST), expenses.pay); // Pay an expense.
 
   /**
+   * Comments
+   */
+  app.get('/groups/:groupid/comments', mw.paginate(), mw.sorting({key: 'createdAt', dir: 'DESC'}), comments.list); // Get latest comments for a collective
+  app.get('/groups/:groupid/expenses/:expenseid/comments', mw.paginate(), mw.sorting({key: 'createdAt', dir: 'DESC'}), comments.list); // Get latest comments for an expense
```
asood123: Shouldn't we get all comments for an expense, instead of latest? And do it from earliest to latest? Frontend can then decide how to show it. Most expenses will likely have few comments, so that should be fine for now.

test/mocks/data.js
```diff
@@ -372,6 +372,18 @@ export default {
       }
   ],
 
+  "comments": [
+    {
+      "text": "Where did you buy this?"
+    },
+    {
+      "text": "I think we should go with a V-neck model"
```
asood123: 😄

PR #649, opened 2016-10-24
replace multiple spaces and tabs with a single space

asood123:
Should reduce log space usage

### DIFF
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -36,7 +36,7 @@ const getTotalDonations = () => {
     SELECT SUM(${generateFXConversionSQL()}) AS "totalDonationsInUSD"
     FROM "Transactions"
     WHERE amount > 0 AND "PaymentMethodId" IS NOT NULL
-  `.replace(/\n/g, ' '), // this is to remove the new lines and save log space.
+  `.replace(/\s\s+/g, ' '), // this is to remove the new lines and save log space.
   {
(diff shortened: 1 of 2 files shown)


server/lib/queries.js
```diff
@@ -36,7 +36,7 @@ const getTotalDonations = () => {
     SELECT SUM(${generateFXConversionSQL()}) AS "totalDonationsInUSD"
     FROM "Transactions"
     WHERE amount > 0 AND "PaymentMethodId" IS NOT NULL
-  `.replace(/\n/g, ' '), // this is to remove the new lines and save log space.
+  `.replace(/\s\s+/g, ' '), // this is to remove the new lines and save log space.
```
xdamman: Just \s+ means at least one occurrence or more. So no need for the second one which would mean at least two occurrences or more.

asood123: Won't that swap out a space for a space? By detecting two or more instances, we minimize unnecessary operations.

xdamman: It will. That won't kill the CPU ;-)
But your solution will swap 2 or 3 spaces with 2 spaces instead of one.

No big deal either way ;-)

asood123: Are you sure? I tested it and it always replaces multiple spaces with a single space. 
screenshot 2016-10-24 18 03 21

xdamman:
🤔

PR #650, opened 2016-10-25
Supercollective stats now include sub collectives

xdamman:
Though it's definitely slower than before. We should cache this better.

### DIFF
--- a/server/models/Group.js
+++ b/server/models/Group.js
@@ -204,7 +204,8 @@ export default function(Sequelize, DataTypes) {
           logo: this.logo,
           backgroundImage: this.backgroundImage,
           publicUrl: this.publicUrl,
-          mission: this.mission
+          mission: this.mission,
+          settings: this.settings
         }
       },
       minimal() {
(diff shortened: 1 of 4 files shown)


server/models/Group.js
```diff
@@ -220,6 +221,22 @@ export default function(Sequelize, DataTypes) {
     },
 
     instanceMethods: {
+      getSuperCollectiveGroupsIds() {
+        if (!this.isSupercollective) return Promise.resolve([this.id]);
+        if (this.superCollectiveGroupsIds) return Promise.resolve(this.superCollectiveGroupsIds);
```
asood123: What's this supposed to do?

asood123: is this like a poor man's cache? :)

asood123:
Did you test how much slower this is for opensource collective?

xdamman:
Keep in memory the list of ids so that we don't have to compute them again
(since we do multiple calls to get total backers, total donations, etc.)

PR #655, opened 2016-11-02
support unpaid param in expenses

asood123:


server/controllers/expenses.js
```diff
@@ -58,6 +58,10 @@ export const list = (req, res, next) => {
     order: [[req.sorting.key, req.sorting.dir]]
   }, req.pagination);
 
+  if (req.body.unpaid_only || req.query.unpaid_only) {
```
xdamman: I think it's better to keep it consistent with the structure, e.g. req.body.status = 'UNPAID'
And the status should go from 'PENDING' to 'UNPAID' to 'PAID'

PR #666, opened 2016-11-05
add currency in group card

asood123:


asood123:
merging to fix production

PR #673, opened 2016-11-07
Sending the email notification on new expense

xdamman:


server/models/Group.js
```diff
@@ -248,11 +248,27 @@ export default function(Sequelize, DataTypes) {
...
+        lists[roles.BACKER] = 'backers';
+        lists[roles.MEMBER] = 'members';
+        lists[roles.HOST] = 'host';
+
+        const notifications = [{type:`mailinglist.${lists[role]}`}];
+
+        switch (role) {
+          case roles.HOST:
+            notifications.push({type:activities.GROUP_TRANSACTION_CREATED});
+            notifications.push({type:activities.GROUP_EXPENSE_CREATED});
```
asood123: This should be under roles.MEMBER. MEMBER has to approve it. 

We can create a new type of activity. GROUP_EXPENSE_APPROVED and send that to the host for payment.

xdamman: That's a new behavior. I agree with it but I'd do another PR for it. And we will have to communicate about it. Also, not sure the members can approve expense in the current code (and need to make sure that one cannot approve own expense)

PR #676, opened 2016-11-15
upgrading selenium to fix test error

asood123:


asood123:
merging this to try to get frontend tests to pass.

PR #680, opened 2016-11-16
default blur to none

asood123:


asood123:
This is good to merge - but there is something funny going on with email.routes.test.js.

PR #682, opened 2016-11-16
get one transaction

xdamman:
used to generate the PDF on the website


xdamman:
@asood123 can I merge this? Should I just go ahead while you are on holidays?

xdamman:
Merging because Aseem is off.

PR #684, opened 2016-11-17
making sure we only return unique users across multiple collectives

xdamman:


xdamman:
Merging to fix production.

PR #687, opened 2016-11-25
Use transaction date instead of donation date for recurring emails

asood123:


asood123:
Tests passed, though tests are emitting a message frequently: ```Error while publishing activity type group.transaction.created for group 1 TypeError: Cannot convert undefined or null to object```

xdamman:
Do you know where this is coming from? Is that new? If so we should fix it now. If not new, then feel free to merge.

PR #692, opened 2016-12-01
measure each query time

asood123:


server/models/index.js
```diff
@@ -10,8 +10,8 @@ import { database as config } from 'config';
 console.log(`Connecting to postgres://${config.options.host}/${config.database}`);
 
 if (config.options.logging) {
-  config.options.logging = (query) => {
-    console.log('\n-------------------- <query> --------------------\n',query,'\n-------------------- </query> --------------------\n');
+  config.options.logging = (query, executionTime) => {
```
xdamman: let's only do this in production, in dev it's good to have the full query with easy separators.

PR #694, opened 2016-12-01
specify currency to be group currency by default in a transaction

asood123:
When WWCode adds funds currently, we always add them in dollars. Instead, we should do group's default currency.


xdamman:
Is the UI reflecting that properly when adding funds? Otherwise this could be very confusing/dangerous.

asood123:
Hard to say what's more confusing, since we don't show currency in the UI. Today, WWCode adds funds and then complains about donations showing up in USD and expenses in GBP because that's what the collective files expenses in. 

Fixing this properly involves touching the app - not worth it. I think it's cleaner to default to Group currency in all cases and really it's only for WWCode.

xdamman:
Ok

PR #695, opened 2016-12-01
fix for yahoo showing as a sponsor on MochaJS

xdamman:


server/controllers/groups.js
```diff
@@ -69,10 +69,13 @@ export const getUsers = (req, res, next) => {
...
 
   const _isActive = (backer) => {
-    if (tiers[backer.tier] && tiers[backer.tier].interval === 'monthly' && now.diff(moment(backer.lastDonation), 'days') > 31)
-      return false
-    else
-      return true;
+    if (tiers[backer.tier]) {
+      if (tiers[backer.tier][0].interval === 'monthly' && now.diff(moment(backer.lastDonation), 'days') > 31)
+        return false;
+      if (now.diff(moment(backer.lastDonation), 'days') > 365)
```
asood123: Doesn't this mean it's better for me to make a one-time donation to stay up there for a year?

I think we should cap one-time donations at 31 days as well to be fair to recurring donors.

xdamman: I tested it and Yahoo doesn't show up anymore.
It's not based on the donation but on the definition of the tier. Since the sponsor tier is defined as "monthly", Yahoo is not considered as a sponsor after 31 days.
For tiers defined as yearly or one-time, I consider a backer active for 365 days.

server/controllers/groups.js
```diff
@@ -69,10 +69,13 @@ export const getUsers = (req, res, next) => {
   const tiers = _.groupBy(req.group.tiers, 'name');
 
   const _isActive = (backer) => {
-    if (tiers[backer.tier] && tiers[backer.tier].interval === 'monthly' && now.diff(moment(backer.lastDonation), 'days') > 31)
-      return false
-    else
-      return true;
+    if (tiers[backer.tier]) {
```
boneskull: `_.get()` may be helpful here to avoid extra conditionals

server/controllers/groups.js
boneskull: (I would even cap one-time backers at much less than a year, but more than 31 days)

asood123: We are going to clean this up properly, merging this now to fix Yahoo issue.

PR #696, opened 2016-12-01
record paypalemail if it's updated when logged in

asood123:


server/controllers/middlewares.js
```diff
@@ -95,8 +95,15 @@ export const getOrCreateUser = (req, res, next) => {
 
   // If already logged in, proceed
   if (req.remoteUser) {
-    req.user = req.remoteUser;
-    return next();
+    if (req.body.expense) {
+      const {name, paypalEmail} = req.body.expense;
+      return req.remoteUser.updateWhiteListedAttributes({ name, paypalEmail })
```
xdamman: not a fan of adding this in the `getOrCreateUser`. It doesn't seem to be the right place.
Also, for more visibility would be good to add a check 
`if (paypalEmail !== req.remoteUser.paypalEmail)`

test/setup.js
```diff
@@ -4,7 +4,7 @@ if (!process.env.NODE_ENV) {
   process.env.NODE_ENV = 'development';
 }
 if (process.env.NODE_ENV === "development") {
-  console.log("Setting PG_DATABASE=opencollective_test");
+  console.log("Setting PG_DATABASE=opencollective_test")
```
xdamman: remove log

xdamman:
Works for me as a quick fix

xdamman:
Then just add a semicolon at the end :)

PR #701, opened 2016-12-05
only use one db connection for scripts

asood123:


xdamman:
wouldn't it be better to have an env variable `PG_CONNECTIONS=2`?

asood123:
Oh yea, that'll work too

xdamman:
Thank you, much cleaner this way! 👍

PR #702, opened 2016-12-05
forgot to pass right param :(

asood123:


xdamman:
ah good catch, I missed that

PR #707, opened 2016-12-08
getting rid of svn, making sure we cache npm install api

xdamman:


xdamman:
So basically `yarn` is a pain on circleci for the moment (issues for node-gyp, mocha, phantomjs), but I figured that we could already save 3-4 minutes by making sure we cache the result of `api> npm install` on circleci instead of rebuilding the api every time we want to run the e2e tests :-)

PR #709, opened 2016-12-12
Graphql

asood123:


asood123:
Pulled out of most of these changes in other branches

PR #710, opened 2016-12-15
fix bug where same user whose host and member doesn't show up as host…

asood123:
If same user is host and member of a collective, currently they won't show up as host and thus no donations can be accepted by collective. For example, see opencollective.com/golix.


xdamman:
But then they will show up twice in the list of users no?
shouldn't we instead return the host as part of returning the group? Since there is a 1:1 relationship here, it would make more sense rather that including that singleton as part of a large selection of users.

asood123:
Frontend already dedups - I added that as part of the new collective page. 

I could see a case for returning HOST as part of the group. I'd say, let's avoid that extra query now and get this out. Sending HOST differently will require changes on the frontend too, instead of relying on UsersByRole, so slightly bigger change.

PR #717, opened 2016-12-27
new notification type user.yearlyreport

xdamman:
so that people can unsubscribe from that very spammy once a year email 🙄


piamancini:
@xdamman you are a good man

PR #718, opened 2016-12-27
Making sure we can actually unsubscribe from yearly email

xdamman:


asood123:
would be nice to add a test for this

asood123:
merging now, since it's blocking production.

PR #720, opened 2016-12-28
script to subscribe all first members. Auto subscribe new members to group.expense.created

xdamman:


scripts/subscribe_to_group.expense.created.js
```diff
@@ -0,0 +1,56 @@
...
+import models from '../server/models';
+
+const debug = require('debug')('subscribe');
+
+const {
+  Notification,
+  UserGroup
+} = models;
+
+const processRows = (rows) => {
```
asood123: fyi, don't need processRows function. can add promise.map directly in line 31

scripts/subscribe_to_group.expense.created.js
```diff
@@ -0,0 +1,56 @@
...
+};
+
+const init = () => {
+
+  const query = {
+      where: {
+        role: 'MEMBER'
+      },
+      order: [['id', 'ASC']],
+      limit: 20
```
asood123: This is only going to fetch 20 rows. Don't we want to add this for all groups?

xdamman: argh of course that was for testing. good catch.

xdamman:
deployed and executed the script

PR #722, opened 2016-12-30
remove checkifenoughfunds for approves

asood123:
**Previously:**
- When approving an expense,
   - if it's manual, we'd check that there were enough funds in the group
   - if it's paypal, we'd ping paypal to make sure the host has enough preapproved funds (we never checked collective for enough funds)
- When paying an expense,
   - if it's manual, we'd pay it right away
   - if it's paypal, we'd try to pay via paypal and pass error back.

**Now:**
- When approving an expense,
   - we just approve (manual or non-manual)
- When paying an expense,
   - we check enough funds in the group
      - if it's manual, then pay (meaning create a txn)
      - if it's paypal, try to pay through paypal and pass error back

### DIFF
--- a/test/expenses.routes.test.js
+++ b/test/expenses.routes.test.js
@@ -5,7 +5,7 @@ import request from 'supertest-as-promised';
 import sinon from 'sinon';
 import * as utils from '../test/utils';
 import roles from '../server/constants/roles';
-import { badRequest, missingRequired } from './lib/expectHelpers';
+import { missingRequired } from './lib/expectHelpers';
 import paypalMock from './mocks/paypal';
 import paypalAdaptive from '../server/gateways/paypalAdaptive';
 import models from '../server/models';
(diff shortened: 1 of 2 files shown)


test/expenses.routes.test.js
```diff
@@ -632,68 +565,95 @@ describe('expenses.routes.test.js', () => {
...
+                        .tap(activity => expect(activity.data.paymentResponse).to.deep.equal(payMock)));
                   });
-              }
+                });
+
+                function expectTransactionCreated(expense, transaction) {
+                  expect(transaction).to.have.property('netAmountInGroupCurrency', -12000);
+                  expect(transaction).to.have.property('ExpenseId', expense.id);
+                  // TODO remove #postmigration, info redundant with joined tables?
+                  expect(transaction).to.have.property('amount', -expense.amount/100);
```
xdamman: isn't a good time to remove those?

test/expenses.routes.test.js
```diff
@@ -768,49 +720,71 @@ describe('expenses.routes.test.js', () => {
...
+                  it('THEN creates a transaction paid activity', () =>
+                    expectTransactionPaidActivity(group, host, transaction)
+                      .tap(activity => expect(activity.data.paymentResponse).to.be.undefined));
+                });
+
+                function expectTransactionCreated(expense, transaction) {
+                  expect(transaction).to.have.property('netAmountInGroupCurrency', -3737);
+                  expect(transaction).to.have.property('ExpenseId', expense.id);
+                  // TODO remove #postmigration, info redundant with joined tables?
+                  expect(transaction).to.have.property('amount', -expense.amount/100);
```
xdamman: same here

test/expenses.routes.test.js
asood123: I'd rather not remove the tests, until we remove the actual code that's populating them.

test/expenses.routes.test.js
asood123: Same as above. That way, when you remove the code, tests fail and you have confirmation that you removed the right fields and all of their instances.

PR #724, opened 2017-01-02
Update username algo

asood123:
- Cleans up username algorithm
- introduces `suggestedUsername` field on User, as a way to pass github usernames (or other ones in the future
- looks at group slug table for conflicts.

### DIFF
--- a/server/lib/userlib.js
+++ b/server/lib/userlib.js
@@ -1,6 +1,7 @@
 import url from 'url';
 import Promise from 'bluebird';
 import clearbit from '../gateways/clearbit';
+import { sequelize } from '../models';
 
 export default {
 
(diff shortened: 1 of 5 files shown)


server/lib/userlib.js
```diff
@@ -113,5 +114,55 @@ export default {
...
     });
-  }
+  },
+
+  /*
+   * If there is a username suggested, we'll check that it's valid or increase it's count
+   * Otherwise, we'll suggest something.
+   */
+
+  checkOrSuggestUsername(user) {
```
xdamman: It's a weird method name. Check + Or don't go well together (what are you checking before getting to the OR part?). I'd prefer in this case simply createUsername()

server/lib/userlib.js
```diff
@@ -113,5 +114,55 @@ export default {
...
+    // In theory, this should never happen because we already have an email
+    // TODO: add a random username to make sure that every user has a username
+    if (potentialUserNames.length === 0) {
+      console.error(`No potential username found for user: ${user.email}`);
+      return Promise.resolve()
+    }
+
+    // fetch any matching usernames or slugs for the top choice in the list above
+    return sequelize.query(`
+        SELECT username as username FROM "Users" where username like '${potentialUserNames[0]}%'
```
xdamman: Oh so you too are getting into raw queries? He he :-)

asood123: no way to do UNION ALL in sequelize that I can find :/

asood123:
fixed. merging now.

PR #726, opened 2017-01-03
fix user associated with expenses

asood123:
Currently, when a new expense is created, we set the userid in transactions table to be the host. This fixes it back to the person who filed the expense.

### DIFF
--- a/test/expenses.routes.test.js
+++ b/test/expenses.routes.test.js
@@ -27,7 +27,7 @@ const {
 } = models;
 
 describe('expenses.routes.test.js', () => {
-  let sandbox, host, member, otherUser, group, emailSendMessageSpy;
+  let sandbox, host, member, otherUser, expenseFiler, group, emailSendMessageSpy;
 
   before(() => {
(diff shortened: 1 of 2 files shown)


test/expenses.routes.test.js
```diff
@@ -829,8 +831,7 @@ describe('expenses.routes.test.js', () => {
     let group, user;
     return (g ? Promise.resolve(g) : models.Group.create(utils.data('group2')))
       .tap(g => group = g)
-      .then(() => u ? u : models.User.create(utils.data('user4'))
-        .tap(user => group.addUserWithRole(user, roles.HOST)))
+      .then(() => u ? u : expenseFiler)
```
xdamman: Not a fan of this function. `createExpense` should not be responsible to create a group. Also, this particular line is very hard to follow. Why is a user turning into `expenseFiler`?

asood123:
I don't think it's worth reworking the tests for this small change. Took me
over 2 hours to make these small changes to the tests.

PR #729, opened 2017-01-04
isPublic -> isActive

xdamman:


server/controllers/groups.js
```diff
@@ -311,10 +311,13 @@ export const create = (req, res, next) => {
     .tap(g => {
       // if there is a host id provided, we add the collective to the host
       if (group.HostId) {
-        return User.findOne({ where: { id: group.HostId }}).tap(host => _addUserToGroup(g, host, {role: roles.HOST, remoteUser: creator}))
+        return User.findOne({ where: { id: group.HostId }}).tap(h => {
+          host = h;
```
asood123: why do we need to store `host` separately?

xdamman:
Right now I did it on the front end. See pull request there.

server/controllers/groups.js
xdamman: We pass it to the activity data. I just reused the same pattern used in that segment for groupCreated. I think it's clearer and easier to read/follow.

asood123: 👍

PR #745, opened 2017-01-19
fix for approve/reject expense email

xdamman:
@asood123 looks like you broke this with your clean up of activities.
We need `group.slug` for generating the approve/reject action urls: https://github.com/OpenCollective/opencollective-api/blob/master/server/lib/email.js#L246-L247


asood123:
Good catch. Was wondering what happened with those emails.

PR #749, opened 2017-01-22
interval can be month, year or null

xdamman:
necessary for https://github.com/OpenCollective/opencollective-website/pull/627

### DIFF
--- a/server/controllers/groups.js
+++ b/server/controllers/groups.js
@@ -284,8 +284,8 @@ export const create = (req, res, next) => {
 
   // Default tiers
   group.tiers = group.tiers || [
-    {"name":"backer","range":[2,100000],"presets":[2,10,25],"interval":"monthly"},
-    {"name":"sponsor","range":[100,500000],"presets":[100,250,500],"interval":"monthly"}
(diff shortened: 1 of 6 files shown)


xdamman:
Actually, I just found out that everywhere in the database we use `{"interval":"one-time"}` or `{"interval":"monthly"}`.
So instead, I should rather standardize on that and not on `month / year / null`. So ignore this PR.

xdamman:
To be more accurate, the API expects the interval in the tier definition to be `monthly`, `yearly` or `one-time`, but the `Subscriptions` model, expects the interval to be `month`, `year` or `null` :-/

xdamman:
If we want to standardize everything to month/year/null, we would have to do a query in the DB to update all the tiers in the Groups table.

If we want to standardize everything to monthly/yearly/one-time, we would just have to run a migration on the `Subscriptions` table (I'd personally prefer the latter as it sounds more natural to have `monthly` imho)

xdamman:
@asood123 any strong feelings about this?

I think we can table this for now, no urgency. Just wanted to make sure you saw this.
Feel free to close this issue for now.

asood123:
Fine with tabling. We can come back to it later.

PR #750, opened 2017-01-22
Standardize subscription charge dates

asood123:
Details here: https://github.com/OpenCollective/OpenCollective/issues/288


asood123:
Lot of subtle things changed under the hood. I added a lot more tests to make sure all those cases were covered.

Easiest to review this PR commit by commit.

templates/partials/renewalnotice.hbs
```diff
@@ -0,0 +1 @@
+<p> Please note that your future payments will be charged on 1st of each month. <a href='https://opencollective.com/faq#renewal-notice'>Why we do this?</a></p>
```
asood123: Need this translated in French for `thankyou.fr.hbs` template

templates/partials/charge_date_notice.hbs
```diff
@@ -0,0 +1 @@
+<p> Please note that your future payments will be charged on 1st of each month. <a href='https://opencollective.com/faq#charge-date-notice'>Why we do this?</a></p>
```
asood123: Need this translated in French for `thankyou.fr.hbs` template

server/lib/donations.js
```diff
@@ -2,6 +2,7 @@ import * as stripe from '../gateways/stripe';
 import * as transactions from '../constants/transactions';
 import roles from '../constants/roles';
 import emailLib from './email';
+import { getSubscriptionTrialEndDate } from './utils';
```
asood123: moved.

server/lib/donations.js
```diff
@@ -26,14 +27,16 @@ export const processDonation = (Sequelize, donation) => {
               paymentMethod.customerId,
               { plan: plan.id,
                 application_fee_percent: transactions.OC_FEE_PERCENT,
+                trial_end: getSubscriptionTrialEndDate(donation.createdAt, subscription.interval),
                 metadata: {
                   groupId: group.id,
                   groupName: group.name,
                   paymentMethodId: paymentMethod.id,
                   description: `OpenCollective: ${group.slug}`
```
asood123: good idea, done

server/lib/donations.js
```diff
@@ -32,7 +31,7 @@ export const processDonation = (Sequelize, donation) => {
                   groupId: group.id,
                   groupName: group.name,
                   paymentMethodId: paymentMethod.id,
-                  description: `OpenCollective: ${group.slug}`
+                  description: `https://opencollective.com/${group.slug}`
```
xdamman: Do you have the interval passed to that function? If so it would be great to add it to the description like we did in our db ("Monthly donation to ..."). I'd also consider appending the description of the donation since we let people donate to a custom tier (with the request donation URL).

asood123: We do have that data but I'm not sure what we get out of adding that, since no one ever looks at it. This metadata is meant for the merchant internally to provide info about the payment - groupid, paymentMethodId doesn't mean anything to anyone else. User never sees it on credit card statement. If we were a typical merchant, this would be helpful for manual book keeping. Not clear how it helps us in our current setup?

xdamman: It doesn't cost us anything to add and one day we may very well be thankful that we did that. I could see a host like women who code make use of that info as they parse their stripe data (they already asked us to add the slug data there)

PR #753, opened 2017-01-24
Events

xdamman:


server/graphql/query.js
```diff
@@ -12,29 +12,44 @@ import models from '../models';
...
-   * Given a group slug and an event slug, returns the event
+   * Given a collective slug and an event slug, returns the event
    */
-  getEvents: {
-    type: new GraphQLList(EventType),
+  getEvent: {
+    type: EventType,
     args: {
       eventSlug: {
         type: GraphQLString,
```
asood123: Need to make this required. See line 25. Otherwise, it'll allow you to pass in just a `colletiveSlug` and return nothing.

PR #754, opened 2017-01-24
bring back subscription confirmed notification

asood123:
this was removed in the last PR from `stripeWebhook.js`, because we moved it to `lib/donations.js`


xdamman:
Would be good to have a test for that.
Feel free to merge.

PR #755, opened 2017-01-24
Expense Status fix

asood123:
Variable name confusion was still causing bugs.

### DIFF
--- a/server/controllers/expenses.js
+++ b/server/controllers/expenses.js
@@ -225,15 +225,15 @@ export const pay = (req, res, next) => {
   }
 };
 
-function assertExpenseStatus(expense, status) {
-  if (status === status.PENDING && expense.status === status.APPROVED) {
+function assertExpenseStatus(expense, statusToConfirm) {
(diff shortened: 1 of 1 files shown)


server/controllers/expenses.js
```diff
@@ -225,15 +225,15 @@ export const pay = (req, res, next) => {
   }
 };
 
-function assertExpenseStatus(expense, status) {
-  if (status === status.PENDING && expense.status === status.APPROVED) {
+function assertExpenseStatus(expense, statusToConfirm) {
```
xdamman: What about `currentStatus`?

asood123: Thought about that, `currentStatus` is technically what `expense.status` should be. We want to check that the current status of an expense matches a particular status. Open to better wording

asood123:
merging this for now to fix production

server/controllers/expenses.js
xdamman: Then 'expectedStatus'?

asood123: that's better. Let's swap it next time one of us touches that code.

PR #756, opened 2017-01-24
Event schema changes

asood123:
Changes based on https://github.com/OpenCollective/OpenCollective/issues/296

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,3 +8,5 @@ dropdb:
 database:
 	createdb $(DB_NAME)
 	createdb $(DB_TEST_NAME)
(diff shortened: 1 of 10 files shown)


asood123:
If we decide to add lat/long, still need to add that

asood123:
Proper lat/long support is in there. Note that you'll either need to manually run `CREATE EXTENSION POSTGIS;` in your local and test database or reset them using the `db:reset`.

We'll have to manually run that on staging and prod before we push as well.

asood123:
Hmm circleci test database is having trouble setting up extension...

PR #759, opened 2017-01-27
Always attach a new collective to a host, falls back to ours

xdamman:


xdamman:
This pull request is from Jan 26 (!!) :(

PR #761, opened 2017-01-27
download database before wiping local copy

asood123:


asood123:
If dumping from prod or staging fails, local db isn't wiped out and is still usable.

PR #762, opened 2017-01-27
pass it empty bcc for server errors, so ops@ doesn't get cc'd

asood123:


asood123:
Hmm tests aren't triggering

PR #764, opened 2017-01-27
Migrate remaining amounts in db to ints

asood123:


asood123:
Don't merge this yet, just approve if you are good with it. Need to sync with website (and maybe even app) and don't want to block pushes

server/controllers/donations.js
```diff
@@ -61,11 +43,11 @@ const stripeDonation = (req, res, next) => {
     return next(new errors.BadRequest('Stripe Token missing.'));
   }
 
-  if (!amountFloat) {
+  if (!amount) {
     return next(new errors.BadRequest('Payment Amount missing.'));
   }
 
-  if (amountInt < 50) {
+  if (amount < 50) {
```
xdamman: How do you feel about being super explicit and say "amountInCents"? #idea

migrations/20170127223729-change-amounts-to-ints-migration.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface, Sequelize) {
+    return queryInterface.sequelize.query('update "Transactions" SET amount = amount*100')
+      .then(() => queryInterface.sequelize.query('update "Subscriptions" SET amount = amount*100'))
```
xdamman: Love it, thanks for tackling this!

server/controllers/donations.js
```diff
@@ -123,11 +105,12 @@ export {stripeDonation as stripe};
 export {stripeDonation as post};
 
 
+/*
 const paypalDonation = (req, res, next) => {
   const { group } = req;
   const { payment } = req.required;
   const currency = payment.currency || group.currency;
-  const amountFloat = payment.amount; // TODO: clean this up when we switch all amounts to INTEGER
+  const amount = parseInt(payment.amount * 100, 10);
```
xdamman: that's where having a very explicit `amountInCents` would make this more readable.

Should the frontend keep using float?

server/controllers/stripeWebhook.js
```diff
@@ -219,8 +219,7 @@ export default function stripeWebhook(req, res, next) {
         platformFeeInTxnCurrency: fees.applicationFee,
         paymentProcessorFeeInTxnCurrency: fees.stripeFee,
         data: {charge, balanceTransaction},
-        description: 'Recurring subscription', // remove #postmigration
-        tags: ['Donation'], // remove #postmigration
+        description: 'Recurring subscription',
```
xdamman: Can you prepend `monthly` or `yearly`?

server/controllers/donations.js
asood123: By default, all amounts are in cents now. How about the opposite? In cases where it's not (like Paypal expects in dollars), we name it amountNotInCents or amountFloat?

server/controllers/donations.js
asood123: yes, I think the frontend should send this back in cents now as well. I'll flip it. Only time we ever turn it into float is to display or when it leaves our servers.

server/controllers/donations.js
xdamman: As the comment on our API suggests, people by default will assume that "amount" is in float, not in cents. That's why I'd rather be explicit. The more we can avoid "RTFM", the better.

server/controllers/donations.js
xdamman: ok

server/controllers/donations.js
asood123: I get it. It'd be better to do, it's a fair amount of work to rename amount everywhere (lots of tables have it, not to mention about 100s of mentions of it in the code - both backend and frontend. Maybe for another weekend...

server/controllers/stripeWebhook.js
asood123: done

PR #766, opened 2017-01-28
Graphql Mutations for Response

asood123:


asood123:
Note that this PR updates queries as well to be more consistent with what graph.cool and other graphql users recommended on our github issue.

server/graphql/mutations.js
```diff
@@ -0,0 +1,105 @@
...
+        }]
+      })
+      .then(t => {
+        if (!t) {
+          throw new Error(`No tier found with tierId:${args.tierId} for eventSlug:${args.eventSlug} in collectiveSlug:${args.collectiveSlug}`);
+        }
+        tier = t;
+      })
+      .then(() => tier.checkAvailableQuantity(args.quantity))
+      .tap(console.log)
```
xdamman: left a console.log

server/graphql/queries.js
```diff
@@ -36,9 +36,9 @@ const queries = {
     }
   },
   /*
-   * Given a collective slug, returns the events
+   * Given a collective slug, returns all events
    */
-  getEvents: {
+  allEventsForCollective: {
```
xdamman: do we really need "ForCollective"?
I think that this is a case where we can overload `allEvents` with or without a `CollectiveSlug`, because in both cases we expect an array of events.

server/models/Tier.js
```diff
@@ -71,6 +75,28 @@ export default function(Sequelize, DataTypes) {
...
+              TierId: this.id
+            }
+          })
+          .then(usedQuantity => {
+            if (this.maxQuantity && usedQuantity) {
+              return this.maxQuantity - usedQuantity;
+            } else if (this.maxQuantity) {
+              return this.maxQuantity;
+            } else {
+              return -1; // means there was no maxQuantity set, so infinite availability
```
xdamman: This might be confusing because for concurrency reasons, you may end up with a negative value from line 87.

so I would return `Infinity`;

PR #771, opened 2017-01-30
create a donation for manually added funds

asood123:


asood123:
First step to move over add funds from app to website. Currently, we create a transaction directly (very old code). Now, the frontend will create a donation first and it'll get processed through our normal flow.

asood123:
A better version of this PR: https://github.com/OpenCollective/opencollective-api/pull/836

PR #772, opened 2017-01-30
change amount formatting to reflect new transaction.amount and subscr…

asood123:
…iption.amount


asood123:
merging this, so people don't freak out seeing large numbers on their public channels

PR #775, opened 2017-02-02
disable clearbit

asood123:
Clearbit is the reason for all of our timeouts. We are using their streaming api which only guarantees that it'll return within 1 minute (and they only recommend using it in a queuing system).

We need to move clearbit to run separately like processDonations does right now. It's more work but in the meantime, this should remove timeouts from payments and new_login_token calls.


asood123:
Merging this to get it live in production asap. @xdamman pls review when you get back.

PR #776, opened 2017-02-02
bring back clearbit on model afterCreate hook

asood123:
Brings back clearbit as an async process.

### DIFF
--- a/test/users.routes.test.js
+++ b/test/users.routes.test.js
@@ -171,8 +171,6 @@ describe('users.routes.test.js', () => {
         });
     });
 
-    /*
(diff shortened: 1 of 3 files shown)


test/users.routes.test.js
```diff
@@ -537,7 +534,7 @@ describe('users.routes.test.js', () => {
 
     beforeEach(() =>
       models.User.create({
-        email: 'xdamman@gmail.com' // will have twitter avatar
+        email: 'xdamman1@gmail.com' // will have twitter avatar
```
xdamman: Can you use another email: e.g. `randomemail242@gmail.com`

PR #779, opened 2017-02-06
send email to the host on new group.created

xdamman:


server/controllers/groups.js
```diff
@@ -309,13 +309,10 @@ export const create = (req, res, next) => {
...
     })
     .tap(g => {
-      // if there is a host id provided, we add the collective to the host
-      if (group.HostId) {
-        return User.findOne({ where: { id: group.HostId }}).tap(h => {
-          host = h;
-          _addUserToGroup(g, host, {role: roles.HOST, remoteUser: creator})
-        })
-      }
+      return User.findOne({ where: { id: group.HostId || defaultHostId() }}).tap(h => {
```
asood123: That means we'll always be setting a host, regardless of the mission of the group?

server/controllers/groups.js
```diff
@@ -279,7 +279,7 @@ export const create = (req, res, next) => {
       group,
       confirmation_url: user.generateLoginLink(`/${group.slug}`)
     }
-    emailLib.send('group.created', user.email, data);
+    emailLib.send('group.confirm', user.email, data);
```
asood123: It would be better to move all email sending to notifications library. Non-blocking and easier to manage logic in one place. Let's try to work any new emails through that in the future. This one is fine for now.

server/controllers/groups.js
xdamman: correct, we can always change it before we activate the collective

xdamman:
PR moved here: https://github.com/OpenCollective/opencollective-api/pull/782

Issue #266, opened 2016-04-11
Documentation is not viewable


rhalff:
Even after registration the documentation is not viewable:

http://docs.opencollective.apiary.io/

![screenshot 25](https://cloud.githubusercontent.com/assets/274358/14435232/0fde3286-0017-11e6-9b74-39d01d0be7c1.png)

erlend-sh:
I'd also like to have access to OpenCollective's API docs. We're interested in working together on a plugin for Discourse, similar to what we're doing with Patreon:

https://meta.discourse.org/t/discourse-patreon-login/44366

balupton:
Hey all, I'm also super keen about this. Met with @piamancini this week to have a chat about it. So posting here as a follow up.

I'm building a reward fulfilment service that goes on top of different sponsorship sources and services (such as patreon, gratipay, open collective). Patreon currently is the only one that provides the API calls that we need - which is the history of donations with their patron details and donation amounts - such that my service can then ask the patrons for the extra details needed for reward fulfilment.

Gratipay have confirmed they are going to implement the necessary API calls. So that just leaves Open Collective. Naturally, my preference is to Gratipay and Open Collective, as they are both open-source, for open-source people, and have dramatically lower fees than Patreon.

gr2m:
Hey 👋 just in case you need any help with Apiary, I've some experience with it and am a big fan. We also use it at Hoodie and have it hooked up with a GitHub repository, so it allows the community to submit fixes and the documentation on apiary gets updated immediately: https://github.com/hoodiehq/account-json-api. It’s a great setup

znarf:
Closing. We're working on the next iteration of our API and a proper documentation is part of the effort.

PR #785, opened 2017-02-09
Remove link to private docs in README.md

erlend-sh:
Removed link to documentation to remove confusion, as per #266


asood123:
Thanks for submitting, we never got around to updating those.

PR #790, opened 2017-02-09
pass payment declined error back to user

asood123:


asood123:
This removes catching donation errors and users will start seeing them again.

PR #799, opened 2017-02-15
Event confirmation

asood123:


server/graphql/mutations.js
```diff
@@ -39,31 +38,59 @@ const mutations = {
...
-        }
-        tier = t;
-      })
+        })
+        .then(ev => {
+          if (!ev) {
+            throw new Error(`No event found with slug: ${response.event.slug} in collective: ${response.group.slug}`)
+          }
+          event = ev;
+        })
```
xdamman: this is hard to follow. I would split the case for "interested" in a different function. It would be easier to read; if status is interested, go this way and ignore everything that follows (finding a tier, check availability, pay, etc.)

server/graphql/mutations.js
```diff
@@ -80,32 +107,46 @@ const mutations = {
...
               amount: tier.amount * responseModel.quantity,
               currency: tier.currency,
-              description: `${tier.Event.name} - ${tier.name}`,
+              description: `${event.name} - ${tier.name}`,
             }
           })
           .then(() => responseModel);
         } else {
-          return Promise.resolve(responseModel);
+          if (response.status !== responseStatus.INTERESTED) {
```
xdamman: this would go away

server/graphql/mutations.js
```diff
@@ -80,32 +107,46 @@ const mutations = {
...
+        quantity: response.quantity || 0,
         status: response.status,
         description: response.description
       }))
 
       // record payment, if needed
       .then(responseModel => {
-        if (tier.amount > 0) {
+
+        if (tier && tier.amount > 0 && response.status === responseStatus.YES) {
```
xdamman: you wouldn't need to test for tier.

PR #802, opened 2017-02-17
Monthly Report for Backers Email Template

carlosascari:
For https://github.com/OpenCollective/OpenCollective/issues/312

- Updated design for template: **/templates/email/backer.monthlyreport.hbs**
   + Added Partial: `/partials/monthlyreport.subscription.hbs` as *mr-subscription*
   + Added Partial: `/partials/monthlyreport.collectivecard.hbs` as *mr-collectivecard*
   + Added Partial: `/partials/monthlyreport.footer.hbs` as *mr-footer*
- Added `compile:email` npm script to compile registered email templates for preview. It sends the compiled html to stdout allowing you to pipe it to a folder that is being served in order to see the resulting html.
I used  `npm run -s compile:email backer.monthlyreport > index.html` along with [http-server](https://www.npmjs.com/package/http-server) to preview the email.

----------------

NOTE: There are three logos that will not be found, since this repository does not hold any assets, they will be added to the website repository, or a cdn, and then the urls will be updated.

They images are:

In **backer.monthlyreport.hbs**
~~- `oc-symbol.svg` A svg of the OpenCollective logo of the circle symbol for the `.ContainerLogo` class.~~
Replaced image with existing `oc-logo-icon.svg`, mentioned [here](https://github.com/OpenCollective/opencollective-website/pull/655#issuecomment-280953787)
- `oc-logo-name.svg` A svg of the OpenCollective logo of the name (text only) for the `.ContainerLogoName` class.

In **backer.monthlyreport.hbs**
- `oc-logo.svg` A svg of the OpenCollective logo with the circle symbol and text for the `.FooterLogo`

The images where added to the website repo: https://github.com/OpenCollective/opencollective-website/pull/655


xdamman:
There is an issue with the logo. It shouldn't be repeated. I think we need to force the maxHeight but let the width auto because the logo is not always square. 



Also, this doesn't look good when there is no background image and logo:

xdamman:
See here for how we usually show the logo in the collective card: https://github.com/OpenCollective/opencollective-api/blob/master/templates/partials/collectivecard.hbs#L109

xdamman:
I made a few fixes but some margins can still be improved: 

I'll let you take it from here @carlosascari

carlosascari:
@xdamman 

- [x] "There is an issue with the logo. It shouldn't be repeated. I think we need to force the maxHeight but let the width auto because the logo is not always square"

**Done** `width` was set to auto with `brackground-repeat: no-repeat` for the *monthlyreport.subscription.hbs* partial

- [x] "Also, this doesn't look good when there is no background image and logo:"
"Anyway we can reuse the same partial? Or update that one with the changes that you need? Would love to keep this consistent across all email templates."

**Done** Re-used `collectivecard` partial instead of using `mr-collecticard` partial.

-----

There is an important issue to be aware of when **re-using** email partials. The issue comes in to play when styles are merged into the elements with the [juice]() module. I could not reliably replace the `.ShowCasedCard` with the `collectivecard` since the styles inside the `collectivecard` partial and the `mr-subscription` partial as well as the the styles in the `monthlyreport.backers` template, would not merge in a predictable manner.

Inline styles would look like:
```
margin-top: 24px;
margin: 0 auto;
```

Since one partial had a `margin: 0 auto` and another had the set `margin-top: 24px` on the same element, the juice module did not merge the two values into the correct rule: `margin: 24px auto 0 auto;`

Another issue comes from having the styles mixed with the partials themselves, determining the **specificity** of any css rule is not trivial, making re-use problematic in the long run.

> It is not a good idea to re-use partials that require their styles to be modified by the template or a parent partial attempting to reuse another partial. Re-use works best when you modify very little or nothing of a partial's original css.

BTW: I am sure the issue comes from the juice module since I compiled the email without it, and the css rendered as expected

xdamman:
We could avoid this problem by making sure that we only define in the partial the CSS of the elements in it. The margins around an element should not be defined there.

asood123:
Test passed. Github isn't updating for some reason. Merging.

xdamman:
@carlosascari fyi svg don't work in gmail

PR #805, opened 2017-02-23
return only confirmed responses and username bug fix

asood123:
I split out interested vs yes flow :).

### DIFF
--- a/server/models/User.js
+++ b/server/models/User.js
@@ -36,6 +36,21 @@ export default (Sequelize, DataTypes) => {
     firstName: DataTypes.STRING,
     lastName: DataTypes.STRING,
 
+    name: {
+      type: DataTypes.VIRTUAL(DataTypes.STRING, ['firstName','lastName']),
+      get() {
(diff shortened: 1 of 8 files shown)


server/models/User.js
```diff
@@ -36,6 +36,21 @@ export default (Sequelize, DataTypes) => {
...
+        const firstName = this.get('firstName');
+        const lastName = this.get('lastName');
+        if (firstName && lastName) {
+          return `${firstName} ${lastName}`;
+        } else if (firstName || lastName) {
+          return firstName || lastName;
+        } else {
+          return null;
+        }
+      }
```
xdamman: That works.

I'd have written this differently (more as fyi):

```javascript
const firstName = this.get('firstName') || '';
const lastName = this.get('lastName') || '';
return `${firstName} ${lastName}`.trim();
```

xdamman: or
```javascript
return `${firstName} ${lastName}`.trim() || null;
```
if you really want to return null instead of an empty string.

PR #806, opened 2017-02-24
add script to send one-time email and email template

asood123:


scripts/send_onetime_email_to_core_contributors.js
```diff
@@ -0,0 +1,74 @@
...
+
+const init = () => {
+  console.log("\nStarting script to send a one-time email...\n");
+
+  const startTime = new Date;
+
+  // find all active groups
+  const groupQuery = {
+    attributes: ['id', 'slug'],
+    include: [ { model: models.Transaction, required: true }]
```
xdamman: those would be all collectives that had at least one transaction.
If you want to only include "active collectives", I would consider adding a condition on the date of the last transaction (90 days?).

scripts/send_onetime_email_to_core_contributors.js
```diff
@@ -0,0 +1,74 @@
...
+  .tap(coreContributorsOfActiveGroups => console.log(`Core contributors found: ${JSON.stringify(coreContributorsOfActiveGroups)}`))
+  .then(coreContributorsOfActiveGroups => sendEmail(coreContributorsOfActiveGroups, {}))
+  .then(() => {
+    const timeLapsed = Math.round((new Date - startTime)/1000);
+    console.log(`Total run time: ${timeLapsed}s`);
+    process.exit(0);
+  });
+}
+
+const sendEmail = (recipients, data) => {
```
xdamman: why keeping `, data` if you only call this function with `{}` as data (line 54) ?

scripts/send_onetime_email_to_core_contributors.js
```diff
@@ -0,0 +1,74 @@
...
+  .then(groupIds => sequelize.query(`
+    SELECT u.email, u."firstName", u."lastName" from "UserGroups" ug
+    LEFT JOIN "Users" u on ug."UserId" = u.id
+    where ug.role = :role and ug."GroupId" IN (:groupIds)
+    `, {
+      type: sequelize.QueryTypes.SELECT,
+      replacements: { groupIds, role: roles.MEMBER }
+    }))
+  .tap(coreContributorsOfActiveGroups => console.log(`Core contributors found: ${JSON.stringify(coreContributorsOfActiveGroups)}`))
+  .then(coreContributorsOfActiveGroups => sendEmail(coreContributorsOfActiveGroups, {}))
```
xdamman: could be simplified to `.then(sendEmail)` (see next comment)

server/lib/emailTemplates.js
```diff
@@ -24,6 +24,7 @@ const templateNames = [
   'group.donation.created',
   'group.monthlyreport',
   'group.monthlyreport.text',
+  'onetime',
```
xdamman: Please add a `onetime.text` template with the text version.

Also isn't `announcement` better suited here as a name for the template? (not a strong opinion)

scripts/send_onetime_email_to_core_contributors.js
asood123: That was intentional, I think we should tell all active collectives since this impacts all of them.

scripts/send_onetime_email_to_core_contributors.js
asood123: ah yes, carry over from other script.

server/lib/emailTemplates.js
asood123: that's better. i'll change and merge.

scripts/send_onetime_email_to_core_contributors.js
asood123: done

PR #809, opened 2017-02-27
remove isExpense to clean up result from API

xdamman:


xdamman:
we were already returning `transaction.type = DONATION || EXPENSE` so no need to also return `isExpense` and `isDonation`

asood123:
Thanks for cleaning this up. You'll need to update website and App that depend on this. 

This is good to be merged once app and website PRs are ready.

Issue #769, opened 2017-01-28
💡 API Use Case: advertisement for project sponsors


gr2m:
I had a chat with @piamancini on this and she asked me to share it here.

One of the 3 big reasons why I think Open Collective is so revolutionary is that it’s about the exposure that sponsors get. And we could built upon that and go beyond just listing sponsors in our README, but actually have dedicated sponsor slots on our website etc.

I’m part of the maintainers team at Hoodie ([hood.ie](http://hood.ie/)). We had a sponsor program in the past which was quite successful, but we had to shut it down due to the overhead. If we could use [Hoodie’s collective](https://opencollective.com/hoodie) as a way to receive payments, then **half of the overhead** is already gone.

The other half would be

1. Show a banner on hood.ie for the week’s sponsor
2. Send out two tweets from [@HoodieHQ](https://twitter.com/hoodiehq)

For the 2nd half, we could build a simple "sponsor message management system" that would use Open Collective’s APIs. The API requirements would be

1. A webhook that calls a configurable URL every time someone makes a payment trough the Open Collective
2. Ideally we could embed a payment form into the external service itself in which a company could directly sign up for the pre-defined plan (or one-time payment), and the payment would directly go to our collective

I think the ideal flow from a sponsor’s perspective would be like this:

- Sign up as official sponsor at sponsor.hood.ie. We inform the sponsor that the price is $1,000 USD / week. 
- The sign up form lets the user put in their information about their company, select an available weekly spot, enter their company’s information, enter their sponsor message to be shown on the website and their messages to be tweeted by `@HoodieHQ`
- Then in the last step, Open Collective handles the payment. We don’t show all the options that we show on our Collective page, instead the price should be hard coded, but they payment goes to Hoodie’s Collective’s account and the transaction appears on Hoodie collective’s page. The payment should not be processed until the sponsor was accepted by the community, see below.

What would be special about this sponsor message solution is that it would not work like today’s advertisement networks. We would not inject a message with JavaScript from an external page, instead the service would send a pull request to update the Website’s source code (and maybe the project’s README). That would bring the added value that sponsors would receive some valuable backlinks which is good for their SEO. And the community can reject sponsor messages as they go trough the PR review process. In that way it’s very Open Sourcy :tm: There would also be no problem with tracking users etc which would help with acceptance from the Open Source community. 

[...]

PR #810, opened 2017-02-27
return transaction after donate

xdamman:
In progress.

Needed for https://github.com/OpenCollective/opencollective-api/issues/769

### DIFF
--- a/server/controllers/donations.js
+++ b/server/controllers/donations.js
@@ -46,7 +46,7 @@ const stripeDonation = (req, res, next) => {
       interval
     }
   })
-  .then((transaction) => res.send({success: true, user: req.user.info, transaction: transaction.info }))
+  .then((transaction) => res.send({success: true, user: req.user.info, transaction: transaction && transaction.info }))
   .catch(err => next(new errors.BadRequest(err.message)))
 };
 export {stripeDonation as stripe};
--- a/server/controllers/donations.js
+++ b/server/controllers/donations.js
@@ -46,7 +46,7 @@ const stripeDonation = (req, res, next) => {
       interval
     }
   })
-  .then(() => res.send({success: true, user: req.user.info }))
+  .then((transaction) => res.send({success: true, user: req.user.info, transaction }))
   .catch(err => next(new errors.BadRequest(err.message)))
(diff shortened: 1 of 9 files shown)


server/lib/payments.js
```diff
@@ -49,43 +49,44 @@ export const createPayment = (payload) => {
...
-    } else {
-      title = description || `Donation to ${group.name}`
-      return Promise.resolve();
-    }
-  })
-  // create a new donation
-  .then(subscription => models.Donation.create({
+    // create a new subscription
+    // (this needs to happen first, because of hook on Donation model)
+    .then(() => {
```
xdamman: in many places, your indentation is wrong @asood123. You should check your editor.
it makes it difficult to see in the diff what really changed.

asood123: thanks for fixing. Need to figure out how to reenable linting in editor.

server/controllers/donations.js
```diff
@@ -46,7 +46,7 @@ const stripeDonation = (req, res, next) => {
       interval
     }
   })
-  .then(() => res.send({success: true, user: req.user.info }))
+  .then((transaction) => res.send({success: true, user: req.user.info, transaction }))
```
asood123: Can you put a comment above this explaining why we are returning `transaction` and a TODO that it needs to be reworked when we split the flow? Or add a test that will break if we split the flow.

Longer term, since this will be async, an email with redirect link (that confirms the donation) would be better.

server/lib/payments.js
```diff
@@ -94,37 +95,39 @@ export const createPayment = (payload) => {
       PaymentMethodId: paymentMethod.id,
       SubscriptionId: subscription && subscription.id,
       ResponseId: response && response.id
-    }));
+    }))
+    .then(processPayment);
```
asood123: processPayment needs two parameters, see afterCreate hook.

xdamman: It used to be `processPayment(Sequelize, donation)` but there is no need to pass the `Sequelize` object.

It doesn't seem that the stub cares about the number of arguments, so not sure where it's failing

PR #811, opened 2017-02-27
using group.donation.created

xdamman:


xdamman:
Doh! +8 - 11 lines of code. Why is coveralls rejecting this? Can we only reject if coverage is reduced by more than x percents? @asood123

xdamman:
This one is from Feb 27

asood123:
is this still relevant?

PR #813, opened 2017-02-27
Simplify existing paypal calls and add support for new preapproval flow

asood123:
Under the new paypal setup, we don't need to worry about anything except expiry date of a preapproval key. So, this PR records start and end date of the keys separately in a table row.

### DIFF
--- a/migrations/20170228184811-add-fields-to-PaymentMethod-migration.js
+++ b/migrations/20170228184811-add-fields-to-PaymentMethod-migration.js
@@ -2,16 +2,12 @@
 
 module.exports = {
   up: function (queryInterface, Sequelize) {
-    return queryInterface.addColumn('PaymentMethods', 'startDate', {
+    return queryInterface.addColumn('PaymentMethods', 'expiryDate', {
       type: Sequelize.DATE
     })
-    .then(() => queryInterface.addColumn('PaymentMethods', 'endDate', {
-      type: Sequelize.DATE
-    }))
   },
 
   down: function (queryInterface, Sequelize) {
-    return queryInterface.removeColumn('PaymentMethods', 'startDate')
-    .then(() => queryInterface.addColumn('PaymentMethods', 'endDate'))
+    return queryInterface.removeColumn('PaymentMethods', 'expiryDate')
   }
 };
--- a/migrations/20170228184811-add-fields-to-PaymentMethod-migration.js
+++ b/migrations/20170228184811-add-fields-to-PaymentMethod-migration.js
@@ -1,0 +1,17 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface, Sequelize) {
+    return queryInterface.addColumn('PaymentMethods', 'startDate', {
+      type: Sequelize.DATE
+    })
+    .then(() => queryInterface.addColumn('PaymentMethods', 'endDate', {
+      type: Sequelize.DATE
+    }))
+  },
+
+  down: function (queryInterface, Sequelize) {
+    return queryInterface.removeColumn('PaymentMethods', 'startDate')
+    .then(() => queryInterface.addColumn('PaymentMethods', 'endDate'))
+  }
+};
--- a/server/controllers/paypal.js
+++ b/server/controllers/paypal.js
@@ -16,7 +16,7 @@ export const getDetails = function(req, res, next) {
   const preapprovalKey = req.params.preapprovalkey;
 
   return getPreapprovalDetailsAndUpdatePaymentMethod(preapprovalKey, req.remoteUser.id)
-    .then(res.json)
+    .then(response => res.json(response))
     .catch(next);
 };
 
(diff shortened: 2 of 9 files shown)


migrations/20170228184811-add-fields-to-PaymentMethod-migration.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface, Sequelize) {
+    return queryInterface.addColumn('PaymentMethods', 'startDate', {
```
xdamman: why a startDate? Isn't it always the same as `createdAt`?

migrations/20170228184811-add-fields-to-PaymentMethod-migration.js
```diff
@@ -0,0 +1,17 @@
...
+      type: Sequelize.DATE
+    })
+    .then(() => queryInterface.addColumn('PaymentMethods', 'endDate', {
```
xdamman: wouldn't it make more sense to call this `expiryDate`?

server/controllers/paypal.js
```diff
@@ -16,193 +15,121 @@ const {
 export const getDetails = function(req, res, next) {
   const preapprovalKey = req.params.preapprovalkey;
 
-  return paypalAdaptive.preapprovalDetails(preapprovalKey)
-    .then(paypalResponse => res.json(paypalResponse))
+  return getPreapprovalDetailsAndUpdatePaymentMethod(preapprovalKey, req.remoteUser.id)
+    .then(response => res.json(response))
```
xdamman: `.then(res.json)`

server/controllers/paypal.js
```diff
@@ -16,193 +15,121 @@ const {
...
+  
+  const payload = {
+    currencyCode: 'USD', // TODO: figure out if there is a reliable way to specify correct currency for a HOST.
+    startingDate: startDate.toISOString(),
+    endingDate: endDate.toISOString(),
+    returnUrl,
+    cancelUrl,
+    displayMaxTotalAmount: false,
+    feesPayer: 'SENDER',
+    maxAmountPerPayment: 2500.00, // PayPal claims this can go up to $10k without needing additional permissions from them.
```
xdamman: so should we raise this up to $4-5k?

server/controllers/paypal.js
```diff
@@ -16,193 +15,121 @@ const {
...
+    Promise.reject(new errors.NotFound('This preapprovalKey does not exist.')))
+
+  .then(() => getPreapprovalDetailsAndUpdatePaymentMethod(req.params.preapprovalkey, req.remoteUser.id, paymentMethod))
+
+  .then(() => Activity.create({
+    type: 'user.paymentMethod.created',
+    UserId: req.remoteUser.id,
+    data: {
+      user: req.remoteUser.minimal,
+      paymentMethod
```
xdamman: not sure it's safe to store the paymentMethod in the activity. We should sanitize it first.

server/controllers/paypal.js
```diff
@@ -16,193 +15,121 @@ const {
...
-            token: req.params.preapprovalkey
-          }
-        })
-        .then(paymentMethod => cb(null, paymentMethod))
-        .catch(cb);
-    }],
+/*
+ * Takes a Payment Response
+ */
+const getPreapprovalDetailsAndUpdatePaymentMethod = function(preapprovalKey, userId, paymentMethod = null) {
```
xdamman: why not splitting this in two? `getPreaprovalDetails` and `updatePaymentMethod` ?

server/controllers/paypal.js
asood123: Let's start lower and make sure it works first. We can easily disable old preapprovals and have them go through the flow again on website now.

server/controllers/paypal.js
asood123: Activity isn't used to share anything publicly - it's full of payments data, everything from stripe tokens to paypal keys. Though this should be .info, i'll change it.

server/controllers/paypal.js
asood123: The two always happen at the same time. Whenever we fetch details, we need to update the paymentMethod. If needed, we can split in the future.

server/controllers/paypal.js
asood123: changed

xdamman:
I'd recommend not having those sensitive data in more than one place for
security reasons.

migrations/20170228184811-add-fields-to-PaymentMethod-migration.js
asood123: removed

asood123:
Merging, tests passed on circleci.

server/controllers/paypal.js
asood123: fyi, this doesn't work

PR #820, opened 2017-03-02
add a field to info for email templates

asood123:


asood123:
Merging to fix NaN issue in emails on production.

PR #822, opened 2017-03-02
adding currency as well

asood123:


asood123:
merging to fix production error, fyi, @xdamman

PR #824, opened 2017-03-03
fix for returning stripePublishableKey

xdamman:


xdamman:
merging to fix bug in prod

PR #831, opened 2017-03-10
prevent xss attacks through field values

asood123:


package.json
```diff
@@ -119,6 +121,7 @@
     "nsp": "nsp check",
     "pretest": "npm run hint",
     "test": "npm run test:unit && npm run coverage",
+    "test:unitlocal": "mocha --compilers js:babel-core/register --require babel-polyfill --reporter-options mochaFile=$(bash scripts/test_output_dir.sh junit)/junit.xml test",
```
xdamman: can we use `test:unit:dev` to match `npm run db:migrate:dev` to apply to local dev environment?

server/middleware/sanitizer.js
```diff
@@ -0,0 +1,32 @@
...
+    if (req.body) {
+      req.body = sanitizeHelper(req.body);
+    }
+    return next();
+  };
+};
+
+const sanitizeHelper = (value) => {
+  if (typeof value === 'string') {
+    //value = value.replace(/<script>.*<\/script>/); // remove all script tags and anything in the middle.
```
xdamman: why is this commented out?

server/middleware/sanitizer.js
```diff
@@ -0,0 +1,32 @@
...
+    //value = value.replace(/<.*?>/gi, ''); // for other tags, only remove the tags
+    value = value.replace(/&gt;/gi, '>');
+    value = value.replace(/&lt;/gi, '<');
```
xdamman: why are we replacing html encoded with their equivalent?
isn't it still a problem?

xdamman: tests pass so I guess it's fine, I just want to understand.

asood123: value.replace replaces `&gt;` tags with `>` and then html sanitizer in the next step removes all html tags

server/middleware/sanitizer.js
asood123: argh will fix

PR #832, opened 2017-03-11
only return user email if canEditGroup

xdamman:


server/graphql/queries.js
```diff
@@ -5,12 +5,43 @@ import {
...
+   */
+  allUsers: {
+    type: new GraphQLList(UserType),
+    args: {
+      collectiveSlug: {
+        type: new GraphQLNonNull(GraphQLString)
+      }
+    },
+    resolve(_, args, req) {
+      console.log("Remote user: ", req.remoteUser.username);
```
asood123: remove console.log

server/models/Group.js
```diff
@@ -250,6 +250,15 @@ export default function(Sequelize, DataTypes) {
     },
 
     instanceMethods: {
+      getUsersForViewer(viewer) {
+        const promises = [this.getUsers()];
```
asood123: .getUsers() isn't going to be reliable for our setup, it's going to miss users/roles. Because we removed the unique index on UserGroup, which assumed a unique user and group combo. Our unique index is on user, group and role combo.

server/graphql/queries.js
```diff
@@ -5,12 +5,43 @@ import {
...
+    resolve(_, args) {
+      return models.Group.findOne({
+        where: { slug: args.collectiveSlug.toLowerCase() }
+      })
+    }
+  },
+  /*
+   * Given a collective slug, returns all users
+   */
+  allUsers: {
```
asood123: is this needed? Shouldn't there just be a single graphql query that can get the group and all its users?

```query collective {
    collective(slug: $groupSlug) {
        id,
        name,
        ...,
        users {
            id, 
            name,
            role,
            ...
        }
    }
}

File: server/graphql/types.js [hunk 4]
```diff
@@ -428,8 +434,8 @@ export const ResponseType = new GraphQLObjectType({
       },
       user: {
         type: UserType,
-        resolve(response) {
-          return response.getUser();
+        resolve(response, args, req) {
+          return response.getUserForViewer(req.remoteUser);
```
asood123: this still doesn't feel right. We'd have to implement this same logic on every type that can get a user... Is there no way to implement this once in the `email` field of `UserType`?

xdamman:
What do you suggest?

xdamman:
🤔

I guess the goal of graphql is to allow plenty of ways to access the same data.
I agree that we should also have a collective query.

Also I don't like that sometimes we say groupSlug and sometimes just slug. What could we do?

server/graphql/queries.js
xdamman: done

server/models/Group.js
xdamman: What do you suggest instead?

Not sure why you it's going to miss users/roles. The goal of this instance method on the Group model is to get all the users related to the current group. If a user appears multiple times because it has multiple roles it's fine.

server/graphql/types.js
xdamman: I see what you mean and I don't disagree but I also think that this current solution is good enough for our current use case.

If it was an easy change, I'd be fine with it but turns out it's not trivial because the `email` resolver only has access to the parent `User` and the context (which contains `req.remoteUser`). It doesn't have access to the group. Therefore, we don't know if the remoteUser canEditGroup.

xdamman:
March 10...

PR #836, opened 2017-03-14
Add Funds support

asood123:
Add funds now creates a donation, which automatically processes into a transaction


migrations/20170310165323-add-notes-to-donations-migration.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface, Sequelize) {
+
+    return queryInterface.addColumn('Donations', 'notes', {
```
xdamman: what's the rationale for notes in the context of a donation?
we used to use description.
we use notes as private instructions to the host to reimburse an expense

asood123: There was only a title field on description, so if someone wanted to add more explanation, we have no way of storing that info. This allows for more details to be attached to a donation. When displaying transaction, we can pass that info on.

asood123:
I'll merge this when frontend is ready.

asood123:
merging to get tests to pass on website.

PR #839, opened 2017-03-21
pass expense and donation info when returning transaction

asood123:
This PR populates a `canEditGroup` param on `remoteUser` to let us know if current user should get access to private info and then includes that info when returning a `transaction`.

### DIFF
--- a/server/controllers/middlewares.js
+++ b/server/controllers/middlewares.js
@@ -7,6 +7,7 @@ import queries from '../lib/queries';
 import models from '../models';
 import errors from '../lib/errors';
 import required from '../middleware/required_param';
+import roles from '../constants/roles';
 
 const {
   User
@@ -257,4 +258,17 @@ export const sorting = (options) => {
 
     next();
   };
+};
+
+/**
+ * Populate req with info on whether this user can edit colletive
+ */
+export const populateEditPermissions = (req, res, next) => {
+  if (req.remoteUser && req.group) {
+    return req.group.hasUserWithRole(req.remoteUser.id, [roles.HOST, roles.MEMBER])
+    .then(canEditGroup => req.remoteUser.canEditGroup = true)
+    .then(() => next()); // for some reason, .then(next) doesn't work here.
+  } else {
+    next();
+  }
 };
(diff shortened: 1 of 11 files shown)


server/models/Transaction.js
```diff
@@ -80,7 +99,12 @@ export default (Sequelize, DataTypes) => {
           paymentProcessorFee: this.paymentProcessorFee,
           netAmountInGroupCurrency: this.netAmountInGroupCurrency,
           amountInTxnCurrency: this.amountInTxnCurrency,
-          txnCurrency: this.txnCurrency
+          txnCurrency: this.txnCurrency,
+
+          // expense related fields
+          expenseCategory: this.expenseCategory,
+          expenseIncurredAt: this.expenseIncurredAt,
+          expensePayoutMethod: this.expensePayoutMethod
```
asood123: Same reason we don't send back the entire transaction object but send .public. You don't want to send entire expense object. It's unnecessary and creates parsing work on front end. 

Think of those fields as Transaction fields but happen to be on Expense table. They should be included with every Transaction request sent back. 

By including them here, we don't have to worry about populating them on every controller that returns Transaction. 

Open to other solutions.

xdamman: 🤔

I think the right pattern for this is the one we already use elsewhere: do a joint request and create a virtual getter on the Expense model to only get the attributes you need in that context (e.g. `Expense.dataForTransaction()` ).
That way, the client can get the category for example in a constant manner: `transaction.expense.category`. You don't want to introduce another way to write the variable `expense.category` like you do here with `expenseCategory`.

server/controllers/middlewares.js
```diff
@@ -257,4 +258,17 @@ export const sorting = (options) => {
...
   };
+};
+
+/**
+ * Populate req with info on whether this user can edit colletive
+ */
+export const populateEditPermissions = (req, res, next) => {
+  if (req.remoteUser && req.group) {
+    return req.group.hasUserWithRole(req.remoteUser.id, [roles.HOST, roles.MEMBER])
+    .then(canEditGroup => req.remoteUser.canEditGroup = canEditGroup)
```
xdamman: I wonder if it's not confusing to have `req.remoteUser.canEditGroup` because it's not very explicit on which group this is referring to. What about having instead `req.remoteUser.canEditGroup(groupId)` or `req.remoteUser.canEditGroup[groupId]` to avoid any confusion.

xdamman: I also wonder how this will work with graphql because `req.group` is not going to be populated.

asood123: We'll need something else for GraphQL. This applied to all groups-related routes makes it pretty easy to include private info on those controllers...

PR #840, opened 2017-03-22
update ticket confirmation sustainoss

piamancini:


templates/emails/ticket.confirmed.sustainoss.hbs
```diff
@@ -0,0 +1,40 @@
...
+<p>Hi!</p>
+{{/if}}
+
+<p>We are excited that you'll be joining us at {{ event.name }} on {{ event.startsAt }} at {{ event.locationName }}. </p>
+
+<p>This confirms your {{ response.quantity }} {{{pluralize "ticket" n=response.quantity}}}{{#if donation}} for a total of {{{currency donation.amount currency=donation.currency}}}{{/if}}. Thank you so much for being part of the first $ustainOSS Conversation!</p>
+
+<p>Next steps:
+<ul>
+  <li>You can follow our activities and how we will be spending the money on our open collective page:
```
asood123: should it be capital?

asood123:
Will need to make other code changes to load template and then call it for sustain group.

templates/emails/ticket.confirmed.sustainoss.hbs
piamancini: the proverbial semantics issue we have with our name, product and users. I think in this case is a noun and not a name, so it seems ok in lower case.

xdamman:
Added this to my PR with support for timezone: https://github.com/OpenCollective/opencollective-api/pull/842

PR #846, opened 2017-04-02
Allow users to update their username

xdamman:
Required for https://github.com/OpenCollective/opencollective-website/pull/700

### DIFF
--- a/server/lib/imageUrlToAmazonUrl.js
+++ b/server/lib/imageUrlToAmazonUrl.js
@@ -35,7 +35,7 @@ function imageUrlToAmazonUrl(knox_client, src, callback) {
         }
       }, (err, body) => err ? callback(err) : callback(null, body.Location));
     } else {
-      callback(new Error('Image not found'));
+      callback(new Error(`Image not found: ${src}`));
     }
   });
 }
--- a/server/lib/imageUrlToAmazonUrl.js
+++ b/server/lib/imageUrlToAmazonUrl.js
@@ -15,6 +15,10 @@ import MultiPartUpload from 'knox-mpu-alt';
 *     @param aws_src {String}
 */
 function imageUrlToAmazonUrl(knox_client, src, callback) {
+  // we skip this in development environment since we don't have the AWS credentials
+  if (!process.env.NODE_ENV || process.env.NODE_ENV === 'development') {
+    return callback(null, src);
+  }
   request.head(src, (error, response) => {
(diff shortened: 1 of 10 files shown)


piamancini:
@xdamman i'm gonna go away more often if you start sending me gifts like this when I'm overseas!

server/lib/imageUrlToAmazonUrl.js
```diff
@@ -15,6 +15,10 @@ import MultiPartUpload from 'knox-mpu-alt';
 *     @param aws_src {String}
 */
 function imageUrlToAmazonUrl(knox_client, src, callback) {
+  // we skip this in development environment since we don't have the AWS credentials
```
asood123: I'd vote to minimize or eliminate dev vs prod checks. Historically, this function has been somewhat unstable. And disabling it in development means less likely for us to experience the pain ourselves.

Maybe better to check that if aws keys exist, then attempt to upload? (My dev environment keeps aws keys so it'll keep working).

server/models/User.js
```diff
@@ -351,7 +351,8 @@ export default (Sequelize, DataTypes) => {
 
         let update = false;
         const allowedFields = 
-          [ 'firstName',
+          [ 'username',
```
asood123: What's going to happen if there's a duplicate slug in groups table that already exists?

xdamman: Don't know ¯\_(ツ)_/¯ 
Weren't you checking that when saving the user? Where did you put that logic?

If there is an error message that bubbles up it will be shown on the frontend.

server/lib/imageUrlToAmazonUrl.js
xdamman: Ok

server/models/User.js
asood123: If there's another username, it'll pop error message back to website. Problem is if there is a group slug, it won't error out but the profile will become inaccessible because we prioritize group slugs over usernames if there is a clash.

I check for it in userlib/suggestUsername but that only gets called if User doesn't have a username. hmm...

The right solution is to create a new table that stores slugs and usernames and has a unique index. 

One quick and dirty solution would be to pull out the code from userlib that checks both tables and call that in beforeCreate hook of User model?

xdamman: Updated. Have a look.

asood123: Niiiiice

PR #853, opened 2017-04-08
don't send emails for common errors

asood123:
Got bombarded this morning with a couple thousand emails


asood123:
merging this to at least stop sending thousands of emails... already sent 4k emails today.

PR #855, opened 2017-04-15
Mutations

xdamman:


xdamman:
This also solves the issue with /events not being always responsive by reducing the number of requests to postgres.

PR #856, opened 2017-04-15
get subscribers for an event (opt out)

xdamman:
New mailing list: `:eventSlug@:collectiveSlug.opencollective.com` (opt out)


server/models/Notification.js
```diff
@@ -40,6 +42,61 @@ export default function(Sequelize, DataTypes) {
...
+                type: `mailinglist.${mailinglist}`
+              },
+              include: [{model: models.Group, where: { slug: collectiveSlug } }],
+              paranoid: false
+            }
+          ).then(subscriptions => subscriptions.map(s => s.UserId ))
+          .then(excludeIds => {
+            return users.filter(u => excludeIds.indexOf(u.id) === -1)
+          })
+          .catch(console.error)
```
asood123: Don't we want to pass this error on to us, so we get an email about it?

server/models/Event.js
```diff
@@ -128,6 +128,12 @@ export default function(Sequelize, DataTypes) {
       }
     },
 
+    instanceMethods: {
+      getUsers() {
+        return this.getResponses({ include: [{model: models.User }]}).then(rows => rows.map(r => r.User));
```
asood123: dedup this or emails will go out twice?

server/models/Notification.js
```diff
@@ -40,6 +42,61 @@ export default function(Sequelize, DataTypes) {
...
+        const excludeUnsubscribed = (users) => {
+          return models.Notification.findAll(
+            {
+              where: {
+                channel: 'email',
+                active: false,
+                type: `mailinglist.${mailinglist}`
+              },
+              include: [{model: models.Group, where: { slug: collectiveSlug } }],
+              paranoid: false
```
asood123: why paranoid false?

server/models/Notification.js
```diff
@@ -40,6 +42,61 @@ export default function(Sequelize, DataTypes) {
...
       createMany: (notifications, defaultValues) => {
         return Promise.map(notifications, u => Notification.create(_.defaults({},u,defaultValues))).catch(console.error);
+      },
+
+      /**
+       * Get the list of subscribers to a mailing list (e.g. backers@:collectiveSlug.opencollective.com)
+       * For members, backers, and info: opt-in: we look for rows in the Notifications table
+       * For events, it's opt-out. We exclude people who have explicitly unsubscribed
+       */
+      getSubscribers: (collectiveSlug, mailinglist) => {
```
asood123: This whole function and sub functions took me a while to parse. Would it not be easier to separate into two separate calls? One for events and one for mailing lists?

server/models/Notification.js
xdamman: isn't this required to bypass the default scope that excludes all rows that have "isActive" false?

asood123:
That's only for deletedAt

server/models/Event.js
xdamman: good catch. done.

server/models/Notification.js
xdamman: That's exactly what I did with the sub functions `getSubscribersForEvent` and `getSubscribersForMailingList`.
It's just that I wanted to abstract that complexity from above. Because the only way to know that a given slug related to an event is to first look for an event for that slug.

server/models/Notification.js
xdamman: ok changing

server/models/Notification.js
xdamman: Latest version that I have on my machine is now returning a warning when we don't have a catch for a promise.
This will never lead to an error really. So it's more decorative.

PR #866, opened 2017-04-30
only bcc us once per hour per subject line

xdamman:


asood123:
I'd vote to redirect emails to another mailing list - helps in debugging, etc. 

And the login emails should always go to us - they are useful in logging in as someone else to debug issues.

asood123:
I can make the change ... otherwise, inbox unusable on 1st of the month.

asood123:
Made changes here: https://github.com/OpenCollective/opencollective-api/pull/867

PR #869, opened 2017-05-03
fix unsubscribe to user.monthlyreport

xdamman:


xdamman:
Need to run this query on the database after this:

```sql
UPDATE "Notifications" SET type='user.monthlyreport' WHERE type='backer.monthlyreport'
```

xdamman:
Just ran the query.

PR #873, opened 2017-05-07
fix login bcc

asood123:


asood123:
Merging to get fix out in production. @xdamman pls review

PR #875, opened 2017-05-08
allEvents query with optional collectiveSlug

xdamman:


asood123:
What's the need for fetching all events? Given how slow everything is, fetching all events and tiers could tie up the backend for several seconds at a time...

xdamman:
Just showing the next events for a given collective.
Something I worked on on the plane.
AllEvents was already there. Just merged it with allEventsForCollective. 

Also, given the number of events we have, this won't slow down anything.

PR #881, opened 2017-06-03
send backer monthly report fix

xdamman:
`babel-node cron/monthly/*.js` doesn't run all the scripts but only the first one.

When I tried to run it manually, it gets killed by heroku because it's a long running process. So it needs to be run as part of the cron.

So let's deploy this asap before 4pm CET.


xdamman:
Merging. @asood123 please review later (not urgent).

PR #883, opened 2017-06-04
Monthly host report

xdamman:
This includes quite a few changes that should be reviewed.

Most importantly: 

- Added `Transaction.HostId`
- Added `User.isHost`
- Added `User.currency` (for hosts)
- Using fixer.io to get the amount in the currency of the host for expenses in a different currency

I've also written an extensive migration script.

And I've included a small dataset from wwcode (sanitized) to make it easier to test the `hostlib` without having to create a gazillion of transactions and collectives.


xdamman:
Tests finally pass. @asood123 can you review?
Happy to hop on a call if that's useful.

asood123:
can you merge master into this branch? it has changes that are already merged.

server/lib/hostlib.js
```diff
@@ -0,0 +1,109 @@
...
+      res.byCurrency = amounts;
+    })
+    .then(amounts => Promise.map(amounts, s => convertToCurrency(s.amount, s.currency, hostCurrency || 'USD', date)))
+    .then(amounts => {
+      let total = 0;
+      amounts.map(a => total += a);
+      res.totalInHostCurrency = Math.round(total); // in cents
+      return res;
+    })
+    .catch(console.error);
```
asood123: remove catch

server/lib/hostlib.js
```diff
@@ -0,0 +1,109 @@
...
+
+export function getTotalHostFees(groupids, type, startDate = new Date('2015-01-01'), endDate = new Date, hostCurrency = 'USD') {
+  const where = {
+    GroupId: { $in: groupids },
+    createdAt: { $gte: startDate, $lt: endDate }
+  }; 
+  if (type) {
+    where.type = type;
+  }
+  return sumTransactions('hostFeeInTxnCurrency', where, hostCurrency).catch(console.error);
```
asood123: remove catch

server/lib/hostlib.js
```diff
@@ -0,0 +1,109 @@
...
+
+export function getTotalNetAmount(groupids, type, startDate = new Date('2015-01-01'), endDate = new Date, hostCurrency = 'USD') {
+  const where = {
+    GroupId: { $in: groupids },
+    createdAt: { $gte: startDate, $lt: endDate }
+  };
+  if (type) {
+    where.type = type;
+  }
+  return sumTransactions('netAmountInGroupCurrency', where, hostCurrency).catch(console.error);
```
asood123: catch

scripts/update_transaction_amountInTxnCurrency.js
```diff
@@ -0,0 +1,54 @@
+/**
+ * This scripts populates Transaction.netAmountInTxnCurrency based on historic fxrate
```
asood123: I'm hesitant to populate this data based on historic fxdata that doesn't match host's own books - their accounts won't match exactly and will create a headache for them to track differences per line item.

Do you know how many transactions fall in the category where host and group currencies differ? And for how many of them we aren't the host?

scripts/update_transaction_amountInTxnCurrency.js
```diff
@@ -0,0 +1,54 @@
...
+      return transaction.save({fields: ['txnCurrency','txnCurrencyFxRate','amountInTxnCurrency']});
+    })
+    .catch(e => {
+      console.error("Couldn't get a fxrate for ", transaction.id, e);
+    });
+}
+
+models.Transaction.findAll()
+.tap(transactions => console.log(`Processing ${transactions.length} transactions`))
+.then(transactions => Promise.map(transactions, transaction => {
```
asood123: fyi, can also do .map directly on a Promise returning an array.

scripts/update_transaction_amountInTxnCurrency.js
```diff
@@ -0,0 +1,54 @@
...
+      transaction.amountInTxnCurrency = -Math.round(fxrate * transaction.amount); // amountInTxnCurrency is an INTEGER (in cents)
+      console.log(`Updating transaction ${transaction.id}:\t${transaction.amount} ${transaction.currency}\ttxnCurrency: ${transaction.txnCurrency}\tamountInTxnCurrency: ${transaction.amountInTxnCurrency}\tfxrate: ${transaction.txnCurrencyFxRate}`);
+      return transaction.save({fields: ['txnCurrency','txnCurrencyFxRate','amountInTxnCurrency']});
+    })
+    .catch(e => {
+      console.error("Couldn't get a fxrate for ", transaction.id, e);
+    });
+}
+
+models.Transaction.findAll()
```
asood123: should exclude deletedAt transactions - no need to populate them.

scripts/update_transaction_amountInTxnCurrency.js
```diff
@@ -0,0 +1,54 @@
...
+  if (transaction.currency === transaction.txnCurrency) return;
+  if (transaction.txnCurrencyFxRate > 0 && transaction.txnCurrencyFxRate !== 1) return;
+
+  if (!transaction.txnCurrency) {
+    return transaction.getHost().then(host => {
+      if (!host.currency) {
+        return console.error("Host has no currency set", host.name, "HostId:", host.id);
+      } else {
+        transaction.txnCurrency = host.currency;
+        return updateTransaction(transaction);
```
asood123: only need to do this if txnCurrency is different than host currency. Otherwise, you can set fxRate to 1.

scripts/update_transaction_amountInTxnCurrency.js
```diff
@@ -0,0 +1,54 @@
...
+  if (err) console.log('err', err);
+  console.log('done!');
+  process.exit();
+};
+
+const updateTransaction = (transaction) => {
+  return getFxRate(transaction.currency, transaction.txnCurrency, transaction.createdAt)
+    .then(fxrate => {
+      transaction.txnCurrencyFxRate = fxrate;
+      transaction.amountInTxnCurrency = -Math.round(fxrate * transaction.amount); // amountInTxnCurrency is an INTEGER (in cents)
```
asood123: Is this amount always supposed to be negative?

package.json
```diff
@@ -44,6 +44,7 @@
     "graphql": "0.9.1",
     "handlebars": "4.0.5",
     "helmet": "3.5.0",
+    "isomorphic-fetch": "^2.2.1",
```
asood123: remove carat

scripts/migrate_wwcode_test_db.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
asood123: Can you add a line to create database if it doesn't exist? Getting an error for that now.

scripts/db_restore.sh
```diff
@@ -0,0 +1,72 @@
...
+  tables=`psql -qAt -c "select tablename from pg_tables where schemaname = 'public';" "${LOCALDBNAME}"`
+
+  for tbl in $tables ; do
+    psql "${LOCALDBNAME}" -c "alter table \"${tbl}\" owner to ${LOCALDBUSER};"
+  done
+
+  # Change ownership of the database
+  psql "${LOCALDBNAME}" -c "alter database ${LOCALDBNAME} owner to ${LOCALDBUSER};"
+
+  psql "${LOCALDBNAME}" -c "GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO ${LOCALDBUSER};"
```
asood123: Even after I created database manually, I can't access it. Geometry doesn't exist error and then permission denied when I open it.

migrations/201706030000-add-HostId.js
```diff
@@ -0,0 +1,108 @@
...
+ * And populates them
+ */
+const Promise = require('bluebird');
+
+const cache = { HostIdForGroupId: {}, updatedHosts: {} };
+
+const updateHosts = (sequelize) => {
+
+  const getHostCurrency = (ug) => {
+    const query = `SELECT "txnCurrency" FROM "Transactions" WHERE "GroupId"=${ug.GroupId} AND "txnCurrency" is NOT NULL LIMIT 1`
```
asood123: I'd limit this query to only consider donations that have a paymentMethod and aren't `deletedAt`. Expenses data for `txnCurrency` is unreliable.

Also, would be good to run this locally and spit out what it's setting for each host and make sure that anyone who isn't set to USD, we eyeball manually.

migrations/201706030000-add-HostId.js
```diff
@@ -0,0 +1,108 @@
...
+
+    return queryInterface.addColumn('Transactions', 'HostId', {
+        type: Sequelize.INTEGER,
+        references: { model: 'Users', key: 'id' },
+        onDelete: 'SET NULL',
+        onUpdate: 'CASCADE'
+      })
+      .then(() => queryInterface.addColumn('Users', 'currency', { type: Sequelize.STRING }))
+      .then(() => queryInterface.addColumn('Users', 'isHost', { type: Sequelize.BOOLEAN, defaultValue: false }))
+      .then(() => updateHosts(queryInterface.sequelize))
```
asood123: As nice as it is to run this seamlessly in a migration script, it might be to run this script separately. That way, if something fails, we aren't stuck halfway on a script's execution

migrations/201706030000-add-HostId.js
```diff
@@ -0,0 +1,108 @@
...
+      cache.HostIdForGroupId[groupid] = HostId;
+      return HostId;
+    });
+  }
+
+  // updateTransaction
+  // Add HostId to the transaction.
+  const updateTransaction = (transaction) => {
+    return getHostId(transaction.GroupId)
+      .then(hostid => {
```
asood123: should check to make sure that a hostId is returned. No guarantee that previous step will always populate a hostId

server/models/Transaction.js
```diff
@@ -27,7 +24,42 @@ export default (Sequelize, DataTypes) => {
...
+
+    GroupId: {
+      type: DataTypes.INTEGER,
+      references: {
+        model: 'Groups',
+        key: 'id'
+      },
+      onDelete: 'SET NULL',
+      onUpdate: 'CASCADE',
+      allowNull: false
```
xdamman: I ran the migration script locally and it worked fine.

Shouldn't a transaction always have a GroupId and HostId?
Maybe we can make GroupId optional to account for host fees in the future as a separate row?
The question is whether we want to make a host a collective in an on itself (which I think we should).

server/models/Transaction.js
```diff
@@ -108,22 +135,26 @@ export default (Sequelize, DataTypes) => {
...
-            transaction.netAmountInGroupCurrency =
-              Math.round((transaction.amountInTxnCurrency
-                - transaction.platformFeeInTxnCurrency
-                - transaction.hostFeeInTxnCurrency
-                - transaction.paymentProcessorFeeInTxnCurrency)
-              *transaction.txnCurrencyFxRate);
-        }
-        return Transaction.create(transaction);
+        return group.getHost().then(host => {
+          transaction.HostId = host.id;
```
asood123: It's being called directly from one of the Group controllers as well.

server/models/User.js
```diff
@@ -90,7 +90,18 @@ export default (Sequelize, DataTypes) => {
     description: DataTypes.STRING,
     longDescription: DataTypes.TEXT,
     isOrganization: DataTypes.BOOLEAN, // e.g. DigitalOcean, PubNub, ...
-
+    isHost: DataTypes.BOOLEAN, // e.g. WWCode
```
asood123: Personally, I think that creates more confusion over time - especially as we are often manually manipulating the database. So, now when Pia creates a host, she'll have to flip additional bits and forgetting it will create inconsistency.

server/lib/transactions.js
```diff
@@ -34,23 +35,29 @@ export function createFromPaidExpense(paymentMethod, expense, paymentResponses,
...
     description: expense.title,
     UserId,
     GroupId: expense.GroupId,
-  })
-  .tap(t => paymentMethod ? t.setPaymentMethod(paymentMethod) : null)
-  .then(t => createPaidExpenseActivity(t, paymentResponses, preapprovalDetails));
+    HostId: host.id
+  };
+
+  return getFxRate(expense.currency, txnCurrency, expense.incurredAt || expense.createdAt)
```
xdamman: Yep. Fixed. Still using this for add funds or paid manually (if expense.currency != host.currency)

server/models/Transaction.js
xdamman: There are only 3 transactions done by Pia more than a year ago and that already have a `deletedAt`.
I would recommend to delete them so that we can enforce the requirement of `GroupId` NOT NULL.

server/models/Transaction.js
xdamman: Added this

    if (!host) {
      throw new Error(`Cannot create a transaction: collective id ${group.id} doesn't have a host`);
    }

server/models/User.js
xdamman: ok

server/models/Transaction.js
asood123: makes sense, let's delete them.

PR #886, opened 2017-06-06
added tests for create/edit event

xdamman:


server/models/Group.js
```diff
@@ -260,6 +260,11 @@ export default function(Sequelize, DataTypes) {
         .then(results => results[0].map(user => results[1] ? user.info : user.public))
       },
 
+      getRoleForUser(user) {
+        if (!user) return null;
+        return models.UserGroup.findOne({ UserId: user.id}).then(ug => ug.role);
```
asood123: doesn't this need a filter for GroupId?

asood123: also, this can return multiple roles

PR #888, opened 2017-06-08
fix for unsubscribe by email

xdamman:


server/controllers/services/email.js
```diff
@@ -22,7 +22,9 @@ export const unsubscribe = (req, res, next) => {
     models.Group.findOne({ where: { slug }}),
     models.User.findOne({ where: { email }})
   ]).then(results => {
-      return results[1].unsubscribe(results[0].id, type, 'email')
+    if (!results[1]) throw new errors.NotFound(`Cannot find a user with email "${email}"`);
```
asood123: these bugs keep popping up. It'd be better experience to always tell the user that unsubscribe was successful and in case of errors, send us an email?

xdamman: never had a bug with no user, it was always no group (because no group is defined for the user-monthly-report.

PR #890, opened 2017-06-09
send host separately in group call

asood123:
Currently, we send a host of a collective at the same time as all the other users (members, backers, etc). 

In a supercollective, we send all the members, backers and hosts of *all* of it's subcollectives. This means the frontend receives multiple hosts and doesn't know which is the right one.

This PR sends Host separately for a collective in it's first group info.

### DIFF
--- a/server/models/Group.js
+++ b/server/models/Group.js
@@ -567,17 +567,16 @@ export default function(Sequelize, DataTypes) {
       },
 
       getHost() {
-        return Sequelize.models.UserGroup.find({
-          where: {
-            GroupId: this.id,
-            role: roles.HOST
-          }
+        return Sequelize.models.User.findOne({
+          include: [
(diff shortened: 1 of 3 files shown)


asood123:
Note this PR is backwards compatible and go live without waiting for frontend change.

server/models/Group.js
```diff
@@ -567,17 +567,16 @@ export default function(Sequelize, DataTypes) {
       },
 
       getHost() {
-        return Sequelize.models.UserGroup.find({
-          where: {
-            GroupId: this.id,
-            role: roles.HOST
-          }
+        return Sequelize.models.User.findOne({
```
asood123: we need to send the user because frontend needs info like the user email.

xdamman: Yeah I fixed that in my big PR

PR #892, opened 2017-06-11
Custom donation description

xdamman:
making sure we record the custom description in the transaction


server/lib/payments.js
```diff
@@ -192,6 +192,7 @@ const processPayment = (donation) => {
               hostFeeInTxnCurrency: parseInt(balanceTransaction.amount * hostFeePercent / 100, 10),
               platformFeeInTxnCurrency: fees.applicationFee,
               paymentProcessorFeeInTxnCurrency: fees.stripeFee,
+              description: donation.title,
```
asood123: This field is meant to be deprecated - it's better to keep info in donation or expense and pull it from there. That way, transaction remains a purely ledger entry and all info about the original intent is separate.

xdamman: Right now we need this to generate the PDF invoice.

xdamman: Don't have time to refactor this now.

PR #899, opened 2017-06-23
Monthly host report

xdamman:
new PDF attachment to the monthly report for hosts with all expenses with their receipt.


asood123:
Looks great!

PR #901, opened 2017-06-27
updated npm run git:clean  (╯°□°）╯︵ ┻━┻

xdamman:
This good old friend.

I think this time, it's the good one ;-)

### DIFF
--- a/scripts/git_clean.sh
+++ b/scripts/git_clean.sh
@@ -1,15 +1,20 @@
 #!/bin/bash
 if [[ `git checkout master` ]]; then 
   for BRANCH in `git branch | grep -v "*"`; do
-    if [[ `git diff master..$BRANCH` ]]; then
-      echo "- $BRANCH needs to be fast-forward";
-      git checkout $BRANCH;
-      if [[ `git merge master --ff-only` ]]; then
+    git checkout $BRANCH > /dev/null 2>&1;
+    if [[ `git merge master --no-edit 2> /dev/null` ]]; then
+      echo "> $BRANCH merged with master";
+      if [[ `git diff master` ]]; then
+	echo "> $BRANCH is different than master";
+        git reset --hard > /dev/null;
+      else
+	echo "> Removing branch $BRANCH";
         git checkout master;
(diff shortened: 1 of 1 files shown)


scripts/git_clean.sh
```diff
@@ -1,15 +1,20 @@
 #!/bin/bash
 if [[ `git checkout master` ]]; then 
   for BRANCH in `git branch | grep -v "*"`; do
-    if [[ `git diff master..$BRANCH` ]]; then
-      echo "- $BRANCH needs to be fast-forward";
-      git checkout $BRANCH;
-      if [[ `git merge master --ff-only` ]]; then
+    git checkout $BRANCH > /dev/null 2>&1;
+    if [[ `git merge master --no-edit 2> /dev/null` ]]; then
+      echo "> $BRANCH merged with master";
```
xdamman: in this case, we could merge the branch with master without any conflict

scripts/git_clean.sh
```diff
@@ -1,15 +1,20 @@
...
+      if [[ `git diff master` ]]; then
+	echo "> $BRANCH is different than master";
+        git reset --hard > /dev/null;
```
xdamman: if there are some differences, we make sure we clean any unmerged changes and we move on to the next branch

scripts/git_clean.sh
```diff
@@ -1,15 +1,20 @@
...
+      else
+	echo "> Removing branch $BRANCH";
         git checkout master;
-        git branch -d $BRANCH;
+        git branch -D $BRANCH;
```
xdamman: if there is not even a single line that is different with master, it is safe to hard remove the branch

scripts/git_clean.sh
```diff
@@ -1,15 +1,20 @@
...
       fi;
     else 
-      git branch -D $BRANCH;
+      git reset --hard > /dev/null;
+      echo "Unable to merge $BRANCH with master";
```
xdamman: case where we couldn't merge master branch without conflicts.
In this case we cancel the operation and continue

PR #924, opened 2017-10-06
fix for submit expense when logged in 

xdamman:
Fixes the Unhandled rejection TypeError: Cannot set property image of [object Object] which has only a getter at /app/dist/models/User.js:439:23


xdamman:
tests passed so merging @asood123

PR #935, opened 2017-10-08
remove dup transactions from subscription page

asood123:


xdamman:
shouldn't it always be at least a transaction?

asood123:
tests are written with expectation that a transaction isn't needed. Will take a while to update them and didn't want to block on this. Will come back to this.

PR #993, opened 2017-11-01
Invoice fix

asood123:


server/controllers/transactions.js
```diff
@@ -1,26 +1,59 @@
...
+    models.Collective.findOne({
+      where: { id: req.transaction.HostCollectiveId },
     }),
     req.transaction.getCollective(),
     req.transaction.getFromCollective(),
-    req.transaction.getCreatedByUser()
+    req.transaction.getCreatedByUser(),
+    
+    // TODO: we shouldn't need two paths to fetch a host
+    // User->Collective relationship needs to be stored in the Members table
```
xdamman: We don't need a relationship for that 1:1.
Instead, `Users` table should be eventually moved to `ConnectedAccounts` and use the `ConnectedAccounts.CollectiveId` to references the Collective it is associated to.

Likewise, we should introduce a new table `Locations` that should also references a `CollectiveId`. Location could be of type "residence", "billing", "shipping", "venue" (for events).

asood123: Agreed it should be moved - eventually. That eventually isn't going to be in next few months. It's a big migration and we are already behind on our goals for the quarter. And in the meantime, everywhere we need to fetch a User and it's Collective, requires two different pathways (one for USER and one for ORG). Creates more instability. Fixed two such pathways yesterday alone.

yep, addresses need to be separate table as well. That's easier to do.

PR #995, opened 2017-11-01
Fix orders

asood123:
For some reason v2 migration set `FromCollectiveId` incorrectly for various orders (mostly related to three of our accounts). Then, my migration didn't update Order table. This fixes both of those issues.


asood123:
FYI, don't merge this. I'll do it tmrw after testing on a fresh local prod copy - stripe webhooks can add a transaction with incorrect `FromCollectiveId` that will break this migration.

PR #1007, opened 2017-11-08
rely on order.interval in all cases

asood123:


asood123:
Pushing this to fix issue in production immediately, will check on tests afterwards

PR #1027, opened 2017-11-26
tweaks to ledger checker

asood123:


asood123:
Merging, since tests don't run on cron dir

PR #1037, opened 2017-11-30
only edit stripe payment methods

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/698


piamancini:
@xdamman can I let user know? do you close issues when you merge or when you push?

PR #1040, opened 2017-12-01
remove unused manual payment provider

asood123:


asood123:
@xdamman this is what's causing https://github.com/opencollective/opencollective/issues/710. 

this paymentprovider is written like a stripe-equivalent (money coming in). I don't see any reference to it anywhere in the code and it looks incomplete. Guessing it got left behind in v2. Let me know if it's used somewhere.

PR #1044, opened 2017-12-04
yearly user report

xdamman:


xdamman:
@piamancini @asood123 we should review together the text in this end of the year email.

xdamman:
How does this look?

PR #1045, opened 2017-12-04
proxy avatar for third party using api

xdamman:


phated:
This broke the gulp website since we were passing the avatar URL to the proxy URL.  You need a better way of communicating these changes.  cc @piamancini

piamancini:
@phated avatars load fine for me on https://gulpjs.com/
what url should we be looking at?

phated:
@piamancini I just fixed it but it's been broken for 9 days based on when this was merged.

PR #1054, opened 2017-12-09
new email notification to admins for new backers/followers/attendees

xdamman:


xdamman:
Notification on every new donation is very noisy -- and only the host gets them by default.
So here, the admins will only get an email notification on new backers (the first time they are added to their collective).

PR #1085, opened 2017-12-30
modify script to do partial reverse migration

asood123:


asood123:
tested this locally, merging to get live in production

PR #1111, opened 2018-01-25
Update pg module from 4.5.5 to 7.4.1

clarete:
Prompted by this https://nodesecurity.io/advisories/521


clarete:
This is looking good on the CI. But since it's such an important thing, I'd like to confirm if it's OK to proceed with this change, @asood123! Thanks.

PR #1112, opened 2018-01-25
Renew auth token on login

clarete:
This PR aggregates all the changes needed to update the token upon successful login.

This PR also removes a lot of code that seems to be dead. Tests still pass and the login function definitely works locally.


server/middleware/jwt.js
```diff
@@ -1,41 +0,0 @@
...
-        req.jwtPayload = decoded;
-      }
-
-      return next();
-    });
-  } else {
-    return next();
-  }
-
-};
```
asood123: 👍  for removing unused code!

PR #1122, opened 2018-02-06
don't break if null user

xdamman:


xdamman:
required for https://github.com/opencollective/frontend/pull/235

PR #1135, opened 2018-02-13
Don't get/create host stripe customer for pre-migration accounts

clarete:


server/paymentProviders/stripe/creditcard.js
```diff
@@ -54,6 +54,12 @@ export default {
      * @param {*} hostStripeAccount
      */
     const getOrCreateCustomerIdForHost = (hostStripeAccount) => {
+      // Customers pre-migration will have their stripe user connected
+      // to the platform stripe account, not to the host's stripe
+      // account. Since payment methods had no name before that
+      // migration, we're using it to test for pre-migration users;
+      if (!paymentMethod.name) return paymentMethod.custommerId;
```
asood123: `customerId` is misspelled. Worth adding a test for this use case?

clarete: Thanks for the catch! Yeah, certainly worth adding some tests! I'm currently fixing the ones that broke with this small change.

PR #1152, opened 2018-02-27
Properly populate transaction fees

clarete:
# TL;DR
Migration to fill in all the fee columns so the expression `netAmount = amount + hostFee + platformFee + paymentProcessorFee` is true for every single transaction in the database.

## Details
A single real transaction is currently represented by two rows in the `Transactions` table. One of the type `DEBIT` to represent the change in the **User ledger** and another one of the type `CREDIT` to represent changes in the **Collective Ledger**.

That's all good so far. The representation of the fees, however, is a little more confusing. Both rows (`CREDIT` & `DEBIT`) contain the columns to store the fees. But they're currently only populated on the `CREDIT` row. And their values are positive, so they don't really express exactly what is going on with that transaction. To make it a little easier to understand. Here's how a transaction is currently stored:

### How one transaction is stored

|            | User Ledger | Collective Ledger |
|------------|------------:|------------------:|
| Type       |       DEBIT |            CREDIT |
| Amount     |       -4075 |              5000 |
| Host Fee   |           0 |               500 |
| Plat Fee   |           0 |               250 |
| PP Fee     |           0 |               175 |
| Net Amount |       -5000 |              4075 |

### How it should look after the migration

|            | User Ledger | Collective Ledger |
|------------|------------:|------------------:|
| Type       |       DEBIT |            CREDIT |
| Amount     |       -4075 |              5000 |
| Host Fee   |        -500 |              -500 |
| Plat Fee   |        -250 |              -250 |
| PP Fee     |        -175 |              -175 |
| Net Amount |       -5000 |              4075 |

## Motivation

There are two features that we want to add that will greatly benefit from this change:

1. Refund transactions. There's an ongoing effort to automate refunding transactions. Although the feature is [already implemented and ready to be deployed](https://github.com/opencollective/opencollective-api/pull/1134), it greatly increases the complexity of managing transactions and fees. Especially when Payment Processing Fees are not refunded.

2. We're planning to unfold each transaction between User & Collective into more transactions between *User & Payment Processor*, *Payment Processor & Host* and *Host & Collective*. Having the fees correctly filled in will make this job a few orders of magnitude easier

## Challenges

1. **Currency conversion**. The formula `netAmount = amount + hostFee + platformFee + paymentProcessorFee` doesn't apply for transactions between hosts and collectives that use different currencies. That's OK for 93% of our transactions because they're between USD & USD but we need to improve the formula to account for the conversion.

2. **Manual changes**. Some database entries have the fees registered in both types of transactions (`DEBIT` and `CREDIT`) so some care has to be taken to make sure that we don't automate the replication of mistakes made by manual changes.

### DIFF
--- a/scripts/properly_populate_transaction_fees.js
+++ b/scripts/properly_populate_transaction_fees.js
@@ -25,17 +25,28 @@ class Migration {
     return transactions;
   }
 
-  /** Verify values of fees in a transaction */
-  verifyFees = (tr) => {
-    return tr.amountInHostCurrency +
+  /** net value of a transaction */
+  trValue = (tr) => Math.round(
+      tr.amountInHostCurrency +
       tr.hostFeeInHostCurrency +
       tr.platformFeeInHostCurrency +
-      tr.paymentProcessorFeeInHostCurrency === (tr.netAmountInCollectiveCurrency * tr.hostCurrencyFxRate);
-  }
+      tr.paymentProcessorFeeInHostCurrency) * tr.hostCurrencyFxRate;
+
+  /** Verify net value of a transaction */
+  verify = (tr) => this.trValue(tr) === tr.netAmountInCollectiveCurrency;
+
+  /** Difference between transaction net value & netAmountInCollectiveCurrency */
+  difference = (tr) => this.trValue(tr) - tr.netAmountInCollectiveCurrency;
 
   /** Convert `value` to negative if it's possitive */
   toNegative = (value) => {
     return value > 0 ? -value : value;
+  }
+
+  /** Ensure that `tr` has the `hostCurrencyFxRate` field filled in */
+  ensureHostCurrencyFxRate = (tr) => {
+    if (tr.amount === tr.amountInHostCurrency && !tr.hostCurrencyFxRate)
+      tr.hostCurrencyFxRate = 1;
   }
 
   /** Rewrite the values of the fees */
@@ -55,15 +66,17 @@ class Migration {
     const credit = tr1.type === 'CREDIT' ? tr1 : tr2;
     const debit =  tr1.type === 'DEBIT' ? tr1 : tr2;
     if (tr1.ExpenseId !== null && tr1.ExpenseId === tr2.ExpenseId) {
-      console.log('  Expense.:', this.verifyFees(tr1), this.verifyFees(tr2));
+      console.log('  Expense.:', this.verify(tr1), this.verify(tr2));
     } else if (tr1.OrderId !== null && tr1.OrderId === tr2.OrderId) {
-      console.log('  Order...:', this.verifyFees(tr1), this.verifyFees(tr2));
+      console.log('  Order...:', this.verify(tr1), this.verify(tr2));
+      this.ensureHostCurrencyFxRate(credit);
+      this.ensureHostCurrencyFxRate(debit);
       this.rewriteFees(credit, debit);
-      if (!this.verifyFees(credit)) {
-        console.log(`    Transaction CREDIT ${credit.TransactionGroup} doesn't add up`);
+      if (!this.verify(credit)) {
+        console.log(`    Transaction CREDIT ${credit.TransactionGroup} doesn't add up: ${this.difference(credit)}`);
       }
-      if (!this.verifyFees(debit)) {
-        console.log(`    Transaction DEBIT ${credit.TransactionGroup} doesn't add up`);
+      if (!this.verify(debit)) {
+        console.log(`    Transaction DEBIT ${credit.TransactionGroup} doesn't add up: ${this.difference(debit)}`);
       }
 
       // if (!credit.hostFeeInHostCurrency)
@@ -73,7 +86,7 @@ class Migration {
       // if (!credit.paymentProcessorFeeInHostCurrency)
       //   console.log('    * WARNING: C:ppFee.......: ', credit.paymentProcessorFeeInHostCurrency);
     } else {
-      console.log('  WAT.....:', this.verifyFees(tr1), this.verifyFees(tr2));
+      console.log('  WAT.....:', this.verify(tr1), this.verify(tr2));
     }
 
     console.log('    * C:amount......: ', credit.amountInHostCurrency);
--- a/scripts/properly_populate_transaction_fees.js
+++ b/scripts/properly_populate_transaction_fees.js
@@ -1,5 +1,23 @@
+/**
+ * These are some recommendations for using the script in
+ * *development* mode. We currently have ~70000 transactions in our
+ * database, running it with a large batch to process all the
+ * transactions at once is the fastest way to run it and get the
+ * output, however it also exhausts node's memory limits with batches
+ * larger than ~50000. Use --max-old-space=8192 to do it all in one
+ * batch and iterate faster.
+ *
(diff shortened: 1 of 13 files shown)


asood123:
@clarete let me know when you want to talk through currency mismatch cases... I can help you whittle them down.

clarete:
Just documenting what's going on with the off by one errors we found on the ledger. What I initially thought that was caused by manual mistakes ended up being a little deeper. The patterns that lead me to find it were:

* `netAmountICC` off by either `1` or `-1`
* all the transactions were of type `CREDIT`

That made me check the way we calculated the fees and I found out that we don't really round the fees. And then when we store the value we lose the precision because we're storing the currency related values as cents. One example that I have is the following one (all currencies involved are in `USD`):

### Example Input
* `amount 21650`
* `platform fee 0.05`
* `host fee 0.05`

### What `createFromPayload`/`createDoubleEntry` are currently returning:

#### CREDIT
* `amountIHC 21650`
* `netAmountICC 19485 `
#### DEBIT
* `amountIHC -19485`
* `netAmountICC -21650`

### What it should return instead

#### CREDIT
* `amountIHC 21650`
* `netAmountICC 19486 `

#### DEBIT
* `amountIHC -21650`
* `netAmountICC -19486 `

asood123:
Ah good catch!!

scripts/properly_populate_transaction_fees.js
```diff
@@ -0,0 +1,281 @@
...
+ * transactions at once is the fastest way to run it and get the
+ * output, however it also exhausts node's memory limits with batches
+ * larger than ~50000. Use --max-old-space=8192 to do it all in one
+ * batch and iterate faster.
+ *
+ * For production usage, It might be a good idea to break it down in
+ * multiple batches to leave some room for the database to process
+ * operations from the web clients. I've been running 7 batches of
+ * 10000 and didn't take more than ~20s in my machine (i7~2.2GHz),
+ * which is just a little bit slower than running it all in one batch.
```
asood123: Nice to have these perf numbers!

scripts/properly_populate_transaction_fees.js
```diff
@@ -0,0 +1,281 @@
...
+    return transactions;
+  }
+
+  /** Convert `value` to negative if it's possitive */
+  toNegative = (value) => value > 0 ? -value : value;
+
+  /** Ensure that `tr` has the `hostCurrencyFxRate` field filled in */
+  ensureHostCurrencyFxRate = (tr) => {
+    if (tr.amount === tr.amountInHostCurrency && !tr.hostCurrencyFxRate)
+      tr.hostCurrencyFxRate = 1;
```
asood123: This is a minor thing but I feel like, in general, it's better to make these type of functions as `get` functions, instead of modifying the underlying model. If it's `getHostCurrencyFXRate`, you don't need to think about whether model was saved or not. Otherwise, whoever calls this function will need to remember that it modifies the passed in object and they need to save it.

scripts/properly_populate_transaction_fees.js
```diff
@@ -0,0 +1,281 @@
...
+  hasHostFee = (tr) => tr.hostFeeInHostCurrency
+    && parseInt(tr.hostFeeInHostCurrency, 10) !== 0;
+  /** Return true if the transaction has any platform fees */
+  hasPlatformFee = (tr) => tr.paymentProcessorFeeInHostCurrency
+    && parseInt(tr.paymentProcessorFeeInHostCurrency, 10) !== 0;
+  /** Return false if there are no fees on a transaction */
+  hasFees = (tr) => this.hasHostFee(tr) || this.hasPlatformFee(tr)
+    || (tr.platformFeeInHostCurrency && parseInt(tr.platformFeeInHostCurrency, 10) !== 0);
+
+  rewriteFees = (credit, debit) => {
```
asood123: Same comment here. I think `getFees` would be cleaner. It returns the fees and the caller is clearly responsible for setting and storing them. 

Since this is just a script, it's fine to leave it as it is. More a comment for future reusable components.

scripts/properly_populate_transaction_fees.js
```diff
@@ -0,0 +1,281 @@
...
+        console.log('Expense.: true, true');
+        return;
+      }
+
+      // this.rewriteFees(credit, debit);
+
+      console.log('  Expense.:', transactionsLib.verify(tr1), transactionsLib.verify(tr2));
+
+      if (!transactionsLib.verify(credit)) {
+        console.log(`| EDAU | CREDIT | ${credit.id} | ${credit.TransactionGroup} | ${transactionsLib.difference(credit)} |`);
```
asood123: What's EDAU?

server/paymentProviders/stripe/bitcoin.js
```diff
@@ -84,8 +87,8 @@ export default {
             hostCurrency: balanceTransaction.currency,
             amountInHostCurrency: balanceTransaction.amount,
             hostCurrencyFxRate: order.totalAmount / balanceTransaction.amount,
-            hostFeeInHostCurrency: parseInt(balanceTransaction.amount * hostFeePercent / 100, 10),
-            platformFeeInHostCurrency: order.totalAmount*0.05,
+            hostFeeInHostCurrency,
+            platformFeeInHostCurrency: Math.round(order.totalAmount * 0.05),  /* Why not dividing by 100? Is it in a different unit? */
```
asood123: That's a good question on why it's not divided by 100. Can you look at a couple of bitcoin transactions in the database and see if fees look reasonable?

scripts/properly_populate_transaction_fees.js
clarete: Yeah, in that case what I usually like is to create two layers.

 1. Function for the calculation of the new value (the `get` function you mentioned);
 2. Function that actually performs the mutation in the field (the one that I did);

I started from the second layer because I really wanted to test the actual side effect since these objects are going to be instances of database rows and whatever change to them will actually reflect in the database.

Although this is a script, I actually ended up testing most of the delicate functions because I was thinking about keeping that script around since it became sort of a cleanup tool. But I totally agree with you that we should also have the `get*` functions within a library (thinking of `lib/transactions`). I'll do that in small chunks for things that I need to reuse during the changes in the application code.

scripts/properly_populate_transaction_fees.js
clarete: I wrote an answer to the previous question that might address this too.

scripts/properly_populate_transaction_fees.js
clarete: EDAU: Expense Don't Add Up
ODAU: Order Don't Add Up

scripts/properly_populate_transaction_fees.js
clarete: I started measuring because it exhausted the v8 memory when I joined the Collectives table to read the `hostFeePercent` but I think it also helps knowing what to expect when we run it on prod.

server/paymentProviders/stripe/bitcoin.js
clarete: Good call. I'll check that and I'll also double check if there's any other place that I should be rounding it up.

clarete:
The failure in the tests is related to the decrease in the test coverage. I'll come back to it as soon as the migration is rolled out.

PR #1154, opened 2018-02-28
WIP: Docker support

miguelpeixe:
I started working with a docker support for the API, currently the only change needed is skipping migration for the image build. Since docker builds are environment agnostic, we can't run the db migration on it.

As a proposal I added a `SKIP_MIGRATE` env var and manually execute `npm run db:migrate` on the docker entrypoint.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,39 @@
+FROM node:9
+
+ARG UID=991
+ARG GID=991
+
+ENV NODE_ENV=production \
+    SKIP_MIGRATE=true
+
+EXPOSE 3060
+
+RUN apt-get update \
+  && apt-get install -y --no-install-recommends \
+    apt-transport-https \
+    build-essential \
+    git \
+    postgresql-client \
+    unicode-data \
+  && rm -rf /tmp/* /var/lib/apt/lists/*
+
+RUN addgroup --gid ${GID} opencollective \
+  && useradd \
+    --home-dir /opencollective \
+    --shell /bin/bash \
+    --gid ${GID} \
+    --uid ${UID} \
+    --create-home \
+    opencollective
+
+COPY --chown=991:991 . /opencollective
+
+USER opencollective
+
+WORKDIR /opencollective
+
+RUN npm install
+
+ENTRYPOINT ["/opencollective/docker_entrypoint.sh"]
+
+CMD ["node", "/opencollective/dist/index.js"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,9 +23,8 @@ services:
       SESSION_SECRET: i&j@/V6Wx.`g>Aq?qQQF(>u)/Erm;3A=
   postgres:
     image: kartoza/postgis:9.6-2.4
-    # Uncomment below for production (persist database)
     # volumes:
-    # - ./.data/database:/data/db
+    # - /path/to/persistent/db:/var/lib/postgresql
     ports:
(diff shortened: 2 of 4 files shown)


clarete:
Wooot! Thank you so much for that @miguelpeixe!

Do you think this setup would replace the current "out of docker" development environment or you think this could be a step into packing it so people can run an OC API instance on their own servers? The answer for this would probably change the configuration a little bit.

Also, what do you think about maybe moving the variables to a file under the `config` directory?

miguelpeixe:
Cheers @clarete !

It could turn into a dev environment but the current setup fits best for production use.

Correct me if I'm wrong, but I think there are a few bigger changes needed for dev environment, since the current setup assumes localhost for database setup scripts. I don't know much of OC API dev setup yet.

I'm all for docker dev setup, makes it a lot easier to get someone coding.

miguelpeixe:
First I would add a docker-compose.dev.yml with the proper env vars, codebase volume and nodemon as the command to see what happens.

clarete:
Hi @miguelpeixe!!!

> It could turn into a dev environment but the current setup fits best for production use.
>
> Correct me if I'm wrong, but I think there are a few bigger changes needed for dev environment, since the > current setup assumes localhost for database setup scripts. I don't know much of OC API dev setup yet.
>
> I'm all for docker dev setup, makes it a lot easier to get someone coding.

You are right about the scripts for setting up the database. They're all hardcoded to use `localhost`. I also had some trouble when I tried to use those scripts on a fresh environment. From what I can remember, the permissions were a little messed up because the bash scripts call `pg_*` programs with the current user using the terminal and not really the user in the database. So in my case, all the tables and database objects created by the bash scripts are owned by `lincoln` (my shell user) and then I have to manually grant rights to the `opencollective` postgres user, which kinda sucks.

But I think that focusing on getting a docker development environment that sets up everything besides the database might be a very good first step to get this mess figured out and fully automated. Another point that makes me feel like moving towards docker on dev first is that we won't really be able to use docker on prod for a while since it would have to do the clean up on the database scripts anyway.

> About the variables, do you mean the docker-compose env? It actually could be the .env file, since docker supports it. I tend to place it all in the compose because I use rancher to orchestrate production and its easier to handle the vars right in the compose. Its probably best to integrate with the already functional .env file.

Yeah, our system currently has multiple places for variables. but the `.env` file does override the other ones (declared [here](https://github.com/opencollective/opencollective-api/tree/master/config)). It could be nice to centralize everything in the `.env` file. What do you think? Would you just create a new `.env` file with all the vars for the docker environment?

Again, thanks a lot for looking into this. I truly believe that it would make our development experience so much better to have a nice `docker-composer up` doing all the lifting!! That's very exciting!

clarete:
I've been thinking about taking some time to work on this docker setup for development. It's currently very hard to setup the database & connect it to our application in a reproducible way and using docker would make it beautifully simpler for new comers.

I found a very good example of a docker setting while reading the `fixer.io` repository[0]. I might take a stab at this task today. Let me know if you have any further thoughts on it @miguelpeixe!

[0] https://github.com/fixerAPI/fixer

clarete:
Thank you so much for kicking off the conversation and the code to support docker! I just merged the PR #1204 and that implemented the scripts for setting up the database that were missing in this PR. So I'll be closing this PR.

Please feel free to file issues or PRs against the new setup, I'm pretty sure we'll need to polish some edges before it's great for everyone!

PR #1157, opened 2018-03-01
graphql: Don't return memberships with deleted collectives

clarete:
This fixes the issue https://github.com/opencollective/opencollective/issues/936


clarete:
Test's are broken because of an unrelated issue. So I'll push this to production and then I'll figure out what's up with the tests!

PR #1169, opened 2018-03-12
fix slack report query

asood123:


asood123:
Merging this to get data before weekly dhat

PR #1171, opened 2018-03-12
Rework db_restore.sh script & loadDB helper

clarete:
script:
 1. Don't do any retries after silent failures. The script either
    works or properly fails printing out the errors to stderr.

 2. Kill queries that are still running after the tests that started
    them are done. I was initially trying to figure out why the
    queries are still running after their test is done. That's still
    possible. But because of time constraints, I'm happy with just
    killing the queries prior to renewing the database.

loadDB helper:
 1. Remove retry mechanism as well. It should just not fail. If it
    does fail, the underlying problem should be fixed.
 2. Rewrite with async/await.

### DIFF
--- a/scripts/db_restore.sh
+++ b/scripts/db_restore.sh
@@ -7,81 +7,76 @@ usage() {
   exit 0;
 }
 
+# Parse command line arguments
 while [[ $# -gt 1 ]]
 do
-key="$1"
-
-case $key in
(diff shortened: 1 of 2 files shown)


clarete:
Don't think this change is needed anymore. I couldn't get the migration applied in the dump files because of a version mismatch between my postgres version & the CI's postgres version.

PR #1174, opened 2018-03-15
Prepend http to website addresse that might be missing a scheme

Munter:
Some users `website` property end up being sent without a scheme, which means sites embedding opencollective links to users will have broken links on their pages.

Example in the wild where this error occurs:

https://webpack.js.org/
```js
// Select all opencollective links where website is missing a scheme
Array.from(document.querySelectorAll('.support__item'))
  .map(a => a.getAttribute('href'))
  .filter(href => !href.startsWith('http'))
```

I haven't been able to run the tests on my windows machine, so this PR is untested.
The diff on `package-lock.json` seemed quite massive, leading me to think it's not being continuously updated on `master`. I left it out of the commit

Closes https://github.com/webpack/webpack.js.org/pull/1911

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "paypal-rest-sdk": "1.6.8",
     "pg": "4.5.5",
     "pg-hstore": "2.3.2",
+    "prepend-http": "^2.0.0",
     "querystring": "0.2.0",
     "request-as-curl": "0.1.0",
     "request-promise": "4.1.1",
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -5,6 +5,7 @@ import _, { get } from 'lodash';
 import Temporal from 'sequelize-temporal';
 import config from 'config';
 import deepmerge from 'deepmerge';
+import prependHttp from 'prepend-http';
 import queries from '../lib/queries';
 import { difference, uniq, pick } from 'lodash';
 import { types } from '../constants/collectives';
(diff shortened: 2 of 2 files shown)


xdamman:
Thank you!

Munter:
When will this change be deployed? We are still seeing invalid urls for opencollective backers in the API responses

xdamman:
Mmmm... this has already been merged and deployed. Can you show me an example of query/url and expected result? Bonus point if you can write a test like this one: https://github.com/opencollective/frontend/blob/master/test/server/api.json.test.js#L37

Munter:
`curl https://opencollective.com/api/groups/webpack/backers | jq '.[].website | select(.)'` produces a list where lots of urls don't have `http://` or `https://` prepended. 

I can certainly run a test, but it looks like the ones you are linking to are live data? How can I guaranteee correctness of a test when the dataset might change at any time?

xdamman:
We have a small export of the database with some sanitized data: https://github.com/opencollective/opencollective-api/blob/master/test/dbdumps/opencollective_dvl.pgsql

Anyway, I have a fix: https://github.com/opencollective/opencollective-api/pull/1230 and here is the test: https://github.com/opencollective/frontend/pull/316
Apparently, sequelize doesn't use the getter of the model when using a raw query despite being specified https://github.com/opencollective/opencollective-api/blob/master/server/lib/queries.js#L493

Should be deployed later today.

Munter:
Thank you! :D

piamancini:
@xdamman two more sponsors complained about this.

xdamman:
@piamancini this should have been fixed 2 days ago (but maybe Webpack has to redeploy on their side for the change to take effect there). Were those complains more recent? This should be fixed for everyone. Please double check and close the issue if it's all good. Thanks! :)

Munter:
In all the cases I've tested the issue is fixed. I've just triggered a webpack docs rebuild from `mater`. This should also fix the published version

PR #1175, opened 2018-03-15
Fix linting error introduced in a4a38495

Munter:


clarete:
Sorry, I didn't see this PR and ended up fixing it in another commit. Thanks a lot though @Munter!

PR #1176, opened 2018-03-16
Kick off of some design for the new ledger system

clarete:
Besides your general opinion on this PR, I'd love to hear feedback on a few specific points:

1. *Format*: Is this a good place/format for keeping this sort of documentation?
2. *Organization of the content*: I created initial files: rfc-001 as the umbrella for the design of a new ledger and created two more specific ones 002 (to document what we've done with the fees so far) and 003 with more planning for the double entry ledger
3. *Proposed Design*: This initial set of documents aims to be the design of the new ledger system so it'd be amazing to polish this proposal before we start with more drastic changes. It's currently fairly incomplete but the start here!

## Summary of the result

The new ledger needs to provide double entry ledger between collective & payment providers; collective & platform; and collective & host (for fees).

* The balance of the collective doesn't change, it will come from the sum of the collective ledger;

* The balance of the platform collective now sums up all the fees received from other collectives;

* The balance of the host will need a different operation (`accumulatedHostBalance()`).

### DIFF
--- a/rfcs/001-redesign-ledger.md
+++ b/rfcs/001-redesign-ledger.md
@@ -9,10 +9,14 @@ manage the transactions.
 ## Goals
 
 1. Increase accuracy of the ledger
-   1. [-] Store Fees on the User & Collective ledgers [rfc-002]
+   1. Store Fees on the User & Collective ledgers [rfc-002]
       1. [X] Orders
-      2. [ ] Expenses
-      3. [ ] Neither
-   2. [ ] Double Entry Ledger [rfc-003]
+      2. [X] Expenses
+      3. [X] Neither
+   2. Double Entry Ledger [rfc-003]
       1. [ ] Create collective type "Payment Provider"
       2. [ ] Create separate transactions for fees
+   3. Multi Currency v2.0 [rfc-004]
+      1. [ ] Store amounts in Host Currency
+      2. [ ] Where currency conversion happens
+      3. [ ] Multi Wallets
--- a/rfcs/001-redesign-ledger.md
+++ b/rfcs/001-redesign-ledger.md
@@ -14,5 +14,5 @@ manage the transactions.
       2. [ ] Expenses
       3. [ ] Neither
    2. [ ] Double Entry Ledger [rfc-003]
-      1. [ ] Create collective type "Payment Method"
+      1. [ ] Create collective type "Payment Provider"
       2. [ ] Create separate transactions for fees
--- a/rfcs/001-redesign-ledger.md
+++ b/rfcs/001-redesign-ledger.md
@@ -1,0 +1,18 @@
+# REQUEST FOR COMMENTS 001 - Redesign Ledger
+
+## Overview
+
+The current implementation of the ledger can be redesigned to provide
+more accuracy and make it easier to understand the model used to
+manage the transactions.
+
+## Goals
+
+1. Increase accuracy of the ledger
+   1. [-] Store Fees on the User & Collective ledgers [rfc-002]
+      1. [X] Orders
+      2. [ ] Expenses
+      3. [ ] Neither
+   2. [ ] Double Entry Ledger [rfc-003]
+      1. [ ] Create collective type "Payment Method"
+      2. [ ] Create separate transactions for fees
--- a/rfcs/001-redesign-ledger.md
+++ b/rfcs/001-redesign-ledger.md
@@ -20,3 +20,4 @@ manage the transactions.
       1. [ ] Store amounts in Host Currency
       2. [ ] Where currency conversion happens
       3. [ ] Multi Wallets
+   4. `libledger` API [rfc-005]
--- a/rfcs/002-store-fees-on-user-and-collective-ledgers.md
+++ b/rfcs/002-store-fees-on-user-and-collective-ledgers.md
@@ -40,7 +40,7 @@ amountPlusFees = (
 
 ## Use case example
 
-Given a transaction that represents moving $5 dollars from User to
+Given a transaction that represents moving $50 dollars from User to
 Collective
 
 ### How it was stored
--- a/rfcs/002-store-fees-on-user-and-collective-ledgers.md
+++ b/rfcs/002-store-fees-on-user-and-collective-ledgers.md
@@ -1,0 +1,72 @@
+# REQUEST FOR COMMENTS 001 - Store fees on the User & Collective ledgers
+
+## Overview
+
+Every operation of moving funds from one ledger to another is
+currently represented as two rows in the `Transactions` table. One to
+add up to the Collective ledger (CREDIT) and the other to subtract
+from the User's ledger (DEBIT).
+
+Before this change, the fees were only stored in the CREDIT
+transaction. Now the fees are stored in both transaction rows as
+negative numbers.
+
+The change was applied to all the types of fees:
+
+ * Host Fees (`Transaction.hostFeeInHostCurrency` field)
+ * Platform Fees (`Transactions.platformFeeInHostCurrency` field)
+ * Payment Processor Fees
+   (`Transactions.paymentProcessorFeeInHostCurrency` field)
+
+## Verification
+
+The following operation is being used to validate each transaction
+row:
+
+```
+netAmountInCollectiveCurrency === Math.round(amountPlusFees * hostCurrencyFxRate)
+```
+
+And `amountPlusFees` is:
+
+```
+amountPlusFees = (
+  amountInHostCurrency +
+  hostFeeInHostCurrency +
+  platformFeeInHostCurrency +
+  paymentProcessorFeeInHostCurrency
+);
+```
+
+## Use case example
+
+Given a transaction that represents moving $5 dollars from User to
+Collective
+
+### How it was stored
+|            | User Ledger | Collective Ledger |
+|------------|------------:|------------------:|
+| Type       |       DEBIT |            CREDIT |
+| Amount     |       -4075 |              5000 |
+| Host Fee   |           0 |               500 |
+| Plat Fee   |           0 |               250 |
+| PP Fee     |           0 |               175 |
+| Net Amount |       -5000 |              4075 |
+
+### How it is stored now
+
+|            | User Ledger | Collective Ledger |
+|------------|------------:|------------------:|
+| Type       |       DEBIT |            CREDIT |
+| Amount     |       -4075 |              5000 |
+| Host Fee   |        -500 |              -500 |
+| Plat Fee   |        -250 |              -250 |
+| PP Fee     |        -175 |              -175 |
+| Net Amount |       -5000 |              4075 |
+
+## Overview of the next step
+
+This was done because fees represent debit from the transaction pair
+and a credit in a yet to be created Payment Method ledger, which is a
+dependency to create a double entry ledger system that includes User,
+Collective, Host, and Payment Method.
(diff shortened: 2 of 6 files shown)


rfcs/001-redesign-ledger.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Goals
+
+ 1. Increase accuracy of the ledger
+    1. [-] Store Fees on the User & Collective ledgers [rfc-002]
+       1. [X] Orders
+       2. [ ] Expenses
+       3. [ ] Neither
+    2. [ ] Double Entry Ledger [rfc-003]
+       1. [ ] Create collective type "Payment Method"
```
asood123: Did you mean "Payment Provider"? Was thinking we'd have a collective for "Paypal", "Stripe", etc.

rfcs/001-redesign-ledger.md
```diff
@@ -0,0 +1,18 @@
...
+more accuracy and make it easier to understand the model used to
+manage the transactions.
+
+## Goals
+
+ 1. Increase accuracy of the ledger
+    1. [-] Store Fees on the User & Collective ledgers [rfc-002]
+       1. [X] Orders
+       2. [ ] Expenses
+       3. [ ] Neither
```
asood123: Where should we track currency conversion migrations?

rfcs/001-redesign-ledger.md
```diff
@@ -0,0 +1,18 @@
...
+more accuracy and make it easier to understand the model used to
+manage the transactions.
+
+## Goals
+
+1. Increase accuracy of the ledger
+   1. [-] Store Fees on the User & Collective ledgers [rfc-002]
+      1. [X] Orders
+      2. [ ] Expenses
+      3. [ ] Neither
```
asood123: Where should we track multiple currencies?

rfcs/001-redesign-ledger.md
```diff
@@ -0,0 +1,18 @@
...
+   2. [ ] Double Entry Ledger [rfc-003]
+      1. [ ] Create collective type "Payment Method"
```
asood123: Did you mean "Payment Provider"? I thought we needed a collective for "Stripe", "Paypal", etc. (vs. "Credit Card", "Bitcoin", etc).

rfcs/002-store-fees-on-user-and-collective-ledgers.md
```diff
@@ -0,0 +1,72 @@
+# REQUEST FOR COMMENTS 001 - Store fees on the User & Collective ledgers
+
+## Overview
+
+Every operation of moving funds from one ledger to another is
+currently represented as two rows in the `Transactions` table. One to
+add up to the Collective ledger (CREDIT) and the other to subtract
```
asood123: For completeness, I'd rephrase this as: Each row represents a change in one ledger only. Currently, we only represent two ledgers in the system: Collective and User (also a Collective) and fees awkwardly hang out on both.

rfcs/002-store-fees-on-user-and-collective-ledgers.md
```diff
@@ -0,0 +1,72 @@
...
+  amountInHostCurrency +
+  hostFeeInHostCurrency +
+  platformFeeInHostCurrency +
+  paymentProcessorFeeInHostCurrency
+);
+```
+
+## Use case example
+
+Given a transaction that represents moving $5 dollars from User to
```
asood123: Fyi, transaction below is for $50

rfcs/002-store-fees-on-user-and-collective-ledgers.md
```diff
@@ -0,0 +1,72 @@
...
+| Type       |       DEBIT |            CREDIT |
+| Amount     |       -4075 |              5000 |
+| Host Fee   |        -500 |              -500 |
+| Plat Fee   |        -250 |              -250 |
+| PP Fee     |        -175 |              -175 |
+| Net Amount |       -5000 |              4075 |
+
+## Overview of the next step
+
+This was done because fees represent debit from the transaction pair
```
asood123: 👍

rfcs/002-store-fees-on-user-and-collective-ledgers.md
asood123: Ignore this comment, it's covered in next RFC

rfcs/003-double-entry-ledger.md
```diff
@@ -0,0 +1,122 @@
...
+
+This change will increase accuracy because it will simplify the way we
+store the transactions. It will also decrease duplicated code (likely
+reducing bugs) because it will provide a library that implements all
+the basic operations related to the ledger.
+
+## Goals
+
+1. Increase accuracy of the ledger
+   1. [ ] Create collective type "Payment Method"
```
asood123: s/Payment Method/Payment Provider?

rfcs/003-double-entry-ledger.md
```diff
@@ -0,0 +1,122 @@
...
+
+This is different from how transactions are tracked between User &
+Collective. To address this difference, the Payment Method needs to
+have a ledger as well.
+
+### Create separate transactions for fees
+
+#### Use case example
+
+Given a transaction that represents moving $5 dollars from User to
```
asood123: s/$5/$50

rfcs/003-double-entry-ledger.md
```diff
@@ -0,0 +1,122 @@
...
+| Type       |       DEBIT |            CREDIT |
+| Amount     |       -4075 |              5000 |
+| Host Fee   |        -500 |              -500 |
+| Plat Fee   |        -250 |              -250 |
+| PP Fee     |        -175 |              -175 |
+| Net Amount |       -5000 |              4075 |
+
+##### How it should look like after this change
+
+|        |  User | Collective | Collective | Platform | Collective | Host | Collective | Payment Method |
```
asood123: Might be helpful to represent each of these as rows with "fromCollective" and "toCollective". And then showing the net for each entity at the end of it.

rfcs/003-double-entry-ledger.md
```diff
@@ -0,0 +1,122 @@
...
+|--------|------:|-----------:|-----------:|---------:|-----------:|-----:|-----------:|---------------:|
+| Amount | -5000 |       5000 |       -250 |      250 |       -500 |  500 |       -175 |            175 |
+
+##### Where did the net amount go
+
+The net amount won't be a static value in the database anymore. The
+next section introduces an API that provides core operations to the
+ledger including calculate (& cache) the balance, getting fees etc.
```
asood123: Generally like the idea that we don't need to store netAmount anywhere. Curious how you would want to implement it to make it easy to have this value everywhere we need it. 

Also, how would we validate that rows add up without a netAmount?

rfcs/003-double-entry-ledger.md
```diff
@@ -0,0 +1,122 @@
...
+   transaction.
+
+   ```javascript
+   > libledger.rows('d3e36baa-69a2-4f6e-ac44-f93981e51c97')
+   [
+     { fromCollective: 10165, collective: 58, amount: -5000 }, // from User to Collective
+     { fromCollective: 58, collective: 1, amount: -250 },      // from Collective to Platform
+     { fromCollective: 58, collective: 11004, amount: -250 },  // from Collective to Host
+     { fromCollective: 58, collective: 90000, amount: -175 },  // from Collective to Payment Processor
+   ]
```
asood123: 👍 for creating a library to handle this. Would be nice to see the library handle multiple currencies as well.

asood123:
Re: format. I think it's a good start to document ledger changes. Only thing missing is that comments might be lost. So, either someone will have to change the RFC to record comments or dig up this PR. 

Overall, am happy that we are documenting these changes for posterity!

rfcs/001-redesign-ledger.md
clarete: That's an interesting question. I feel like it'd be either before or during the implementation of the double entry ledger. I'm going to do some more reading and suggest a more specific place.

rfcs/001-redesign-ledger.md
clarete: That's an interesting question. I feel like it'd be either before or during the implementation of the double entry ledger. I'm going to do some more reading and suggest a more specific place

rfcs/003-double-entry-ledger.md
clarete: Do you think that's an explanation that should go into the documentation or you're talking about the data that will be saved in the table? If it's related to the data in the table, I'd actually answer with another question. Is it useful to separately list all the fields that we'd keep in that table?

asood123: Not worth listing table fields here. More the idea of double entry. It wasn't clear to me that, for example, `Collective` columns above should be summed to get the netAmount for that Collective.

rfcs/003-double-entry-ledger.md
clarete: That's something I started thinking about but again I'd certainly need more reading & thinking.

My initial thought was to implement that as a JavaScript function that retrieves all the transactions from a given `TransactionGroup` and a collective and returned the net amount. Something like this:

(I'm using the values present in the example of the API call `libledger.rows`)

```javascript
> libledger.netAmount('d3e36baa-69a2-4f6e-ac44-f93981e51c97', 58)
4325
> libledger.netAmount('d3e36baa-69a2-4f6e-ac44-f93981e51c97', 58)
-5000
```

This function would query the database and retrieve all the rows related to a given transaction. Since that can get expensive very quickly with a larger amount of transactions, I'd suggest creating a caching layer on top of these calls (initially thought of `redis`).

rfcs/003-double-entry-ledger.md
clarete: Yeah, I'll give some more thought to how the currencies should be handled in general and then think of which methods could wrap up these currency operations!

rfcs/002-store-fees-on-user-and-collective-ledgers.md
clarete: OK Thanks!

clarete:
I made a few more fixes and added a very initial draft for multi currency. I'll keep iterating over it! Feedback is more than welcome!!

rfcs/003-double-entry-ledger.md
```diff
@@ -0,0 +1,149 @@
...
+| Amount     |      -5000 |   5000 |
+| PP Fee     |       -175 |   -175 |
+| Net Amount |      -5175 |   5000 |
+
+##### How it should look like after this change
+
+| Collective |  Host | Payment Provider | User |
+|-----------:|------:|-----------------:|-----:|
+|      -5000 |  5000 |                  |      |
+|            | -5175 |             5175 |      |
```
asood123: FYI, the payment provider fees comes from the collective, not the host.

clarete: Oh Thanks! This brings me to a question though. If each row in that table represents a Transaction row, what would you think that is more appropriate?

### To create an explicit transaction for the fee?
```
| Collective |  Host | Payment Provider | User |
|-----------:|------:|-----------------:|-----:|
|      -5000 |  5000 |                  |      |
|            | -5000 |             5000 |      |
|       -175 |       |              175 |      |
|            |       |            -5000 | 5000 |
```

### To assume the fee is whatever didn't get sent to the user:
```
| Collective |  Host | Payment Provider | User |
|-----------:|------:|-----------------:|-----:|
|      -5175 |  5175 |                  |      |
|            | -5175 |             5175 |      |
|            |       |            -5000 | 5000 |
```

PR #1184, opened 2018-03-19
Update postgres default config

znarf:
- replace `opencollective_localhost` by `opencollective_dvl`

I'm not sure what was `opencollective_localhost` historically but looks like it should now be 'opencollective_dvl'.

- pass username to `sequelize-cli`

No username is currently passed to `sequelize-cli` in `db:migrate` which is leading to permission errors on a default setup.


clarete:
Hello @znarf! Thank you so much for taking the time to look into these scripts, they really need some TLC. And I think the change to run the migrations with a specific user is pretty great! It definitely fixes some permission annoyance I've been through before.

When it comes to defaulting to use the `opencollective_dvl` database, I just have one concern. That's the database used with some tests and the tests will break if the values on it change. Here's an example: https://github.com/opencollective/opencollective-api/blob/7e0c1a785bdde2f29b1e6ef5eb60017b8b525aa0/test/graphql.invoices.test.js#L11

What that means is that if you run the test above, it will load a dump file called `opencollective_dvl` into `opencollective_test`. And when you add a new migration to the code, you have to run this script:

https://github.com/opencollective/opencollective-api/blob/master/scripts/db_migrate_opencollective_dvl.sh and it will use the `opencollective_dvl` script. I'm still wrapping my head about it but I want to make sure that we won't break those tests if we start using this database as the development database.

@xdamman do you have any suggestions?

xdamman:
There shouldn’t be any conflict. So I have no problem with that change. The opencollective_dvl.pgsql can also be considered as a seed file to get some data when you set up the environment.

PR #1193, opened 2018-03-23
Fix tests with default repository

znarf:
I currently have 6 tests failing with the default repository and these few changes are attempting to fix that.

- default `NODE_ENV` to 'test' when testing

It's a good practice and it's needed to make sure the API key is required ([`process.env.NODE_ENV !== 'development'` in server/routes.js](https://github.com/opencollective/opencollective-api/blob/935b65c196d94c120f7b1cc1489f29b28fb46b8c/server/routes.js#L48)). If the API key is not required, some test are failing.

- load configuration from config/test.json

Now that we have a 'test' environment, we can define its own configuration there. We can then remove the environment tweaks from test/setup.js.

- don't load dotEnv anymore

The current default `.env` would overload `NODE_ENV` and we don't want that (alternative would be to stop overloading `NODE_ENV` there by default).

- rely on mocha.opts for configuration

That let us remove these configuration parameters from the `scripts` section in `package.json`. We can now simply write `npx mocha test/my.test.js` to run an individual test.

- remove `eslint-plugin-mocha` from dependencies

It's already in devDependencies and npm is warning about that.

I'm not sure about `api_key` and `secret` in test.json, should we create new values?


znarf:
That's not working well with `nyc` for now, looking at it.

znarf:
Tests passing: https://travis-ci.org/znarf/opencollective-api/builds/357427912

PR #1208, opened 2018-04-02
Remove deprecated node-uuid and update uuid

znarf:
npm is warning about it

```
npm WARN deprecated node-uuid@1.4.8: Use uuid module instead
```

We can simply replace `node-uuid` by `uuid` and take the opportunity to upgrade the package.


asood123:
From what I remember, doesn't `uuid` module use a different way of calling (like currently `node-uuid.v1()` works. Didn't think that works for the `uuid` module).

One annoying thing about our setup is that circleci tests don't run for forked builds, so the tests won't catch this issue.

znarf:
The history of these packages looks complicated. But in the end, looks like they peacefully merged, see:
 - https://github.com/kelektiv/node-uuid/issues/116
 - https://github.com/kelektiv/node-uuid/issues/142

I double checked and all the modified files in this PR are calling either `uuid.v1()` or `uuid.v4()` and these are properly supported in the `uuid` module. I also did setup Travis on my fork, tests are passing: https://travis-ci.org/znarf/opencollective-api/jobs/361360858

However, this is currently written in the Readme: `[Deprecation warning: The use of require('uuid') is deprecated and will not be supported after version 3.x of this module. Instead, use require('uuid/[v1|v3|v4|v5]')`

So, in the future, that code:

```
const uuid = require('uuid');
const myuuidv1 = uuid.v1()
```

Will need to be replaced by:

```
const uuidv1 = require('uuid/v1');
const myuuidv1 = uuidv1();
```

Maybe that should be done now in this PR?

clarete:
Nice, thank you for running the tests against your fork! I think it's OK to merge this PR & leave the second API update for another PR, but it's up to you!

znarf:
Remaining issue:
https://github.com/opencollective/opencollective-api/blob/master/test/lib.imageUrlToAmazonUrl.js#L22

uuid.v1 is currently stubbed but that is not possible to do if we use it as a function. See: https://github.com/sinonjs/sinon/issues/664

asood123:
Now I remember, I ran into same issue last time when I upgraded it and saved that battle for another day :)

clarete:
hmmm, if you don't find a better way to do it, I'd suggest to create a file called `lib/uuid.js` and then export a function that wraps the uuid module that exposes a function and then use the `lib/uuid.js` module within our codebase. Than we can mock that guy in the test.

xdamman:
What would it take to make sure the tests run on this type of PR?

znarf:
@clarete you're right, it's the way to do it. However, I'm afraid this is also adding complexity. I would personally tend to skip that for now and leave the test like this, but also happy to do it if everyone thinks it's the way to go.

@xdamman There is some info here: https://circleci.com/docs/1.0/fork-pr-builds/

clarete:
I'm +1 on simplifying for the version switch.

PR #1224, opened 2018-04-11
Change amount feature

znarf:
Would be great to have some feedback.

Also, need more tests.

### DIFF
--- a/server/graphql/mutations/orders.js
+++ b/server/graphql/mutations/orders.js
@@ -359,7 +359,6 @@ export async function updateSubscription(remoteUser, args) {
     }
 
     order.Subscription.deactivate();
-    order = await order.destroy();
 
     const newSubscriptionDataValues = Object.assign(omit(order.Subscription.dataValues, [
       'id',
@@ -375,7 +374,6 @@ export async function updateSubscription(remoteUser, args) {
 
     const newOrderDataValues = Object.assign(omit(order.dataValues, [
       'id',
-      'deletedAt',
     ]), {
       totalAmount: amount,
       SubscriptionId: newSubscription.id,
--- a/server/graphql/mutations/orders.js
+++ b/server/graphql/mutations/orders.js
@@ -1,4 +1,4 @@
-import { pick } from 'lodash';
+import { pick, omit } from 'lodash';
 import Promise from 'bluebird';
 
 import models from '../../models';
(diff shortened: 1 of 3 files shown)


server/graphql/mutations/orders.js
```diff
@@ -311,63 +309,71 @@ export function updateSubscription(remoteUser, args) {
...
+    ]), {
+      amount: amount,
+      isActive: true,
+      updatedAt: new Date,
+      activatedAt: new Date,
+    });
+
+    const newSubscription = await models.Subscription.create(newSubscriptionDataValues);
+
+    order = await order.update({ totalAmount: amount, SubscriptionId: newSubscription.id });
```
asood123: Any reason not to create a new `order` as well? Some tradeoffs of the top of my head:

- updating old order will create transactions with two different amounts for that order
- updating old order won't maintain history. User will only see the new subscription in their UI with no record of an old subscription with previous amount.
- One downside of new order is that it'll appear that we just activated that order. We could cheat by copying dates from previous order?

In general, `subscription` and `order` are 1:1 relationship and really don't even need to be separate tables (legacy reasons for their separation). My instinct is to lean towards maintaining that 1:1 relationship...

Thoughts? @clarete @xdamman

server/graphql/mutations/orders.js
```diff
@@ -292,16 +292,14 @@ export function cancelSubscription(remoteUser, orderId) {
   .then(() => models.Order.findOne(query)) // need to fetch it second time to get updated data.
 }
 
-export function updateSubscription(remoteUser, args) {
-
+export async function updateSubscription(remoteUser, args) {
```
asood123: Thanks for converting to `async`. Way cleaner to read.

server/graphql/mutations/orders.js
```diff
@@ -311,63 +309,71 @@ export function updateSubscription(remoteUser, args) {
...
+  }
+  if (!order.Subscription.isActive) {
+    throw new Error('Subscription must be active to be updated');
+  }
 
-    // determine if this order was pastdue
-    if (order.Subscription.chargeRetryCount > 0) {
-      const updatedDates = getNextChargeAndPeriodStartDates('updated', order);
-      const chargeRetryCount = getChargeRetryCount('updated', order);
+  if (amount) {
```
asood123: Need to check that `amount` has a minimum. The order creation flow has an amount - I think `50` or `100` cents. Otherwise, people maybe able to set really low values that will be problematic with Stripe.

server/graphql/mutations/orders.js
znarf: @asood123 I can change that.

How would you approach testing? Having in mind that this we would now have new order and subscription ids. How do you check that they're properly "updated"? Retrieve the new order by collectiveId + userId?

server/graphql/mutations/orders.js
znarf: @asood123 On the backend side, I haven't found the minimal amount handling anywhere. Any pointer? On the frontend side, I only see some 'min' that are matching the Tier.

server/graphql/mutations/orders.js
asood123: Yes, exactly. I'd search for all orders in the database by that user. Should return 2. One's subscription should be cancelled. Other is active and all the fields that were copied over should match.

server/graphql/mutations/orders.js
asood123: It was in our rest end point and looks like disappeared in the graphql, which is checking only for amount > 0. I'd suggest putting a min of 100 cents (regardless of currency). It's not perfect, but good enough to prevent users from setting too low amounts that will mess with Stripe and our system.

asood123: Scratch that, saw you already put that in place. :)

PR #1239, opened 2018-04-19
Subscription end date

clarete:


clarete:
@asood123 Thanks for the comment! I think it might definitely be a good idea! I'll change the patch to not exclude them from the select and then mark them as deactivated instead of charging them!

PR #1103, opened 2018-01-12
populate search index

asood123:
Info at https://github.com/opencollective/opencollective/issues/828

PR #1247, opened 2018-04-24
Search Index and Query

HipsterBrown:
Backend work for https://github.com/opencollective/opencollective/issues/828

Includes @asood123's work from #1103, just updated. The index for Algolia is set using the `ALGOLIA_INDEX` environment variable, so we can have a `dvl_collectives` index and a `collectives` index for staging and production. 

Query:
screen shot 2018-04-24 at 4 33 00 pm

### DIFF
--- a/cron/daily/populate_search_index.js
+++ b/cron/daily/populate_search_index.js
@@ -1,0 +1,90 @@
+import config from 'config';
+import algoliasearch from 'algoliasearch';
+import Promise from 'bluebird';
+
+import models from '../../server/models';
+import { types as collectiveTypes } from '../../server/constants/collectives';
+import { chunkArray } from '../../server/lib/utils';
+import emailLib from '../../server/lib/email';
+
+const ALGOLIA_APP_ID = config.algolia.appId;
+const ALGOLIA_KEY = config.algolia.appKey;
+const ALGOLIA_INDEX = 'collectives';
+const chunkSize = 10; // number of collectives to send at once
+
+const done = (err) => {
+  if (err) {
+    console.log('err', err)
+    return emailLib.sendMessage(
+    'ops@opencollective.com', 
+    'Error in updating index', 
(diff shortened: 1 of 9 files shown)


cron/daily/populate_search_index.js
```diff
@@ -0,0 +1,107 @@
...
+    .catch(console.error)
+  }
+  console.log('Finished updating search records');
+  process.exit();
+}
+
+
+// TODO: clean index once a week
+const populateIndex = () => {
+  return models.Collective.findAll({
```
xdamman: this works, but FYI, you can also use `async` `await` which makes up for a cleaner code imho.

server/graphql/queries.js
```diff
@@ -640,6 +650,44 @@ const queries = {
         }
       });
     }
+  },
+
+  /*
+   * Given a search term, return a list of related Collectives
+   */
+  search: {
+    type: new GraphQLList(CollectiveSearchType),
```
xdamman: why not returning a list of `CollectiveInterfaceType`?
could use algolia to return the `ids` and then fetch the corresponding collectives. That way we keep the number of GraphQL types small.

server/lib/utils.js
```diff
@@ -579,31 +579,31 @@ export function hashCode(str) {
...
 
-  return chunkArray(arr, consecutive).reduce((prom, items, ix) => {
+// This generates promises of n-length at a time 
+// Useful so we don't go over api quota limit on Stripe
+export function promiseSeq(arr, predicate, consecutive =100) {  
+  return chunkArray(arr, consecutive).reduce(( prom, items, ix ) => {
     // wait for the previous Promise.all() to resolve
     return prom.then(() => {
-      //console.log("processing batch", ix);
+      console.log("processing batch", ix);
```
xdamman: remove console.log

PR #1251, opened 2018-04-25
Monthly revenue query: Add filter for refund transactions

clarete:
These changes aim to solve the issue https://github.com/opencollective/opencollective/issues/1026

### New filters to the sub-query

The clause `t."RefundTransactionId" IS NULL` from the `WHERE` statement of the sub-query `conversions` and the following *mutually exclusive* new filters were created:

  * `isRefund`: A refund is true when the transaction represents moving funds from Collective to User after a refund.
  *  `hasBeenRefunded`: A refunded transaction represents the original donation from User to Collective
  * `isNotRefund`: The transaction isn't either a refund or refunded.

## Updates to the columns that aggregate transactions

All aggregations that depend on the following fields in the virtual table required changes:  `recurringMonthlyTotal`, `recurringAnnuallyTotal`, `oneTimeDonations`, `addedFunds`.

The above aggregations only sum up a given transaction if it's either not related to a refund at all or if it's the refund transaction (decreasing collective ledger and increasing user ledger).

### DIFF
--- a/sql/all-sponsors.sql
+++ b/sql/all-sponsors.sql
@@ -17,6 +17,9 @@ where
 	((t."OrderId" IS NOT NULL AND t.type LIKE 'CREDIT')
 	OR (t."ExpenseId" IS NOT NULL AND t.type LIKE 'DEBIT'))
 	AND c.type ilike 'organization'
+    AND ((t."RefundTransactionId" IS NOT NULL AND
+          t."data"->'refund' IS NULL AND
+          t.type = 'CREDIT') OR t."RefundTransactionId" IS NULL)
 
 group by c.slug, "month"
 order by c.slug
--- a/sql/monthly-revenue-query.sql
+++ b/sql/monthly-revenue-query.sql
@@ -161,14 +161,39 @@ with conversions as (select
     CASE
         WHEN (fc.type ilike 'organization') THEN 1
         ELSE 0
-    END as "isOrg"
+    END as "isOrg",
+
+    /** isNotRefund: The transaction isn't either a refund or
+        refunded. */
+    CASE
+        WHEN (t."RefundTransactionId" IS NULL) THEN 1
+        ELSE 0
+    END as isNotRefund,
+    /** hasBeenRefunded: A refunded transaction represents the
+        original donation from User to Collective */
+    CASE
+        WHEN (t."RefundTransactionId" IS NOT NULL AND
+              t."data"->'refund' IS NULL AND
+              t.type = 'CREDIT')
+        THEN 1
+        ELSE 0
+    END as hasBeenRefunded,
+    /** isRefund: A refund is true when the transaction represents
+        moving funds from Collective to User after a refund. */
+    CASE
+        WHEN (t."RefundTransactionId" IS NOT NULL AND
+              t."data"->'refund' IS NOT NULL AND
+              t."type" = 'DEBIT')
+        THEN 1
+        ELSE 0
+    END as isRefund
 
     FROM "Transactions" t
     LEFT JOIN "Orders" d on t."OrderId" = d.id
     LEFT JOIN "Subscriptions" s on d."SubscriptionId" = s.id
     LEFT JOIN "Collectives" fc on t."FromCollectiveId" = fc.id
     WHERE
-        t."deletedAt" IS NULL AND t."RefundTransactionId" IS NULL AND
+        t."deletedAt" IS NULL AND
         t."createdAt" BETWEEN '2016/01/01' AND '2020/01/01' AND
         d."deletedAt" IS NULL AND
         s."deletedAt" IS NULL)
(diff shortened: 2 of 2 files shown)


sql/monthly-revenue-query.sql
```diff
@@ -161,14 +161,23 @@ with conversions as (select
     CASE
         WHEN (fc.type ilike 'organization') THEN 1
         ELSE 0
-    END as "isOrg"
+    END as "isOrg",
+
+    CASE
+        WHEN (t."RefundTransactionId" is not null) THEN 1
```
asood123: Won't this detect every refund transaction instead of only the new ones? So past months will change...

clarete: Yeah, this was just an initial experiment to show a column with something. I've pushed another change with some more well defined filters.

sql/all-sponsors.sql
```diff
@@ -17,6 +17,9 @@ where
 	((t."OrderId" IS NOT NULL AND t.type LIKE 'CREDIT')
 	OR (t."ExpenseId" IS NOT NULL AND t.type LIKE 'DEBIT'))
 	AND c.type ilike 'organization'
+    AND ((t."RefundTransactionId" IS NOT NULL AND
```
asood123: Yes I think that should work. @piamancini can try out the query next week on her data set and see if it reflects coinbase properly

sql/monthly-revenue-query.sql
```diff
@@ -179,104 +204,105 @@ SELECT
...
-        SUM(oneTimeDonations * "isUser") as "numOneTimeDonationsFromUsers",
-        SUM(oneTimeDonations * "isOrg") as "numOneTimeDonationsFromOrgs",
+        SUM(oneTimeDonations * (isNotRefund + isRefund)) as "numOneTimeDonations",
+        SUM(oneTimeDonations * (isNotRefund + isRefund) * "isUser") as "numOneTimeDonationsFromUsers",
+        SUM(oneTimeDonations * (isNotRefund + isRefund) * "isOrg") as "numOneTimeDonationsFromOrgs",
 
     SUM(addedFunds) as "numAddedFunds",
-    SUM(totalExpensesRecorded) as "numExpensesPaid"
+    SUM(totalExpensesRecorded) as "numExpensesPaid",
+    SUM("amountInUSD" * (isRefund + isRefunded)) as "refundTransactions"
```
asood123: Can you move this up above where all the amounts are and follow the decimal conversion? Maybe after `totalDonationsMadeOnPlatform`. That way it's easier to process when we copy paste on spreadsheet

sql/monthly-revenue-query.sql
```diff
@@ -161,14 +161,39 @@ with conversions as (select
...
         ELSE 0
-    END as "isOrg"
+    END as "isOrg",
+
+    /** isNotRefund: The transaction isn't either a refund or
+        refunded. */
+    CASE
+        WHEN (t."RefundTransactionId" IS NULL) THEN 1
+        ELSE 0
+    END as isNotRefund,
```
asood123: Do we need `isNotRefund` at all?

asood123:
Running revenue query gives me positive in some months and negative in others in the last column. Shouldn't it only give me negative numbers? We only care about the actual refunds (not the original)

sql/monthly-revenue-query.sql
```diff
@@ -179,104 +204,105 @@ SELECT
     to_char("givenMonth", 'YYYY-mm') as "month",
 
     /* donations */
-    (SUM("amountInUSD" * recurringMonthlyTotal +
-        "amountInUSD" * recurringAnnuallyTotal +
-        "amountInUSD" * oneTimeDonations +
+    (SUM("amountInUSD" * recurringMonthlyTotal * (isNotRefund + isRefund) +
```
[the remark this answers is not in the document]
clarete: This is the way I found to list both transactions that have nothing to do with refunds *and* transactions that subtract from the collective's ledger (`isRefund`) and not include the original refunded transactions that would add to the collective's ledger.

If we keep only `isNotRefund` it becomes the same thing as it was before. We're just filtering out everything that has anything to do with refunds.

sql/monthly-revenue-query.sql
clarete: I kinda described the `isNotRefund` filter in the comment above, let me know if that doesn't make sense.

sql/monthly-revenue-query.sql
```diff
@@ -161,14 +161,39 @@ with conversions as (select
...
+    END as isNotRefund,
+    /** isRefund: A refund is true when the transaction represents
+        moving funds from Collective to User after a refund. */
+    CASE
+        WHEN (t."RefundTransactionId" IS NOT NULL AND
+              t."data"->'refund' IS NULL AND
+              t.type = 'CREDIT')
+        THEN 1
+        ELSE 0
+    END as isRefunded,
```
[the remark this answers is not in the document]
clarete: Agreed. The naming is very confusing :( I'll elaborate a lil more in your next comment!

sql/monthly-revenue-query.sql
```diff
@@ -161,14 +161,39 @@ with conversions as (select
...
+    /** isRefunded: A refunded transaction represents the original
+        donation from User to Collective */
+    CASE
+        WHEN (t."RefundTransactionId" IS NOT NULL AND
+              t."data"->'refund' IS NOT NULL AND
+              t."type" = 'DEBIT')
+        THEN 1
+        ELSE 0
+    END as isRefund
```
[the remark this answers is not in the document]
clarete: This is actually a great name!!! Thanks a lot for the suggestion!!! After incorporating that suggestion things would look like this:

 *  `isRefund`: A refund is true when the transaction represents moving funds from Collective to User after a refund.
 * `hasBeenRefunded`: A refunded transaction represents the original donation from User to Collective
 * `isNotRefund`: The transaction isn't either a refund or refunded.

clarete:
> Running revenue query gives me positive in some months and negative in others in the last column. Shouldn't it only give me negative numbers? We only care about the actual refunds (not the original)

Although it doesn't really influence the filtering of the actual transactions, I kept the negative number in that column just to help debugging. I wanted to see the impact of the original transactions in the months' values. We just need to use another filter to not include them.

sql/monthly-revenue-query.sql
clarete: Sure thing! I'll do it in a min!

clarete:
I'm going to merge PR so the links in our Queries wiki page will direct to the new version. Please let me know if they're working correctly or if there should be any other changes.

/cc @piamancini

PR #1272, opened 2018-05-04
Fix broken test in collective.model

znarf:
Actually, maybe the best thing to do would have been to delete it and delete `getTiersWithUsers` from the model as it appears not used anywhere.


znarf:
@xdamman should we actually delete this test and the unused model function?

xdamman:
If it's not used anywhere, sure.

PR #1274, opened 2018-05-07
Stop using Sequelize operators aliases

znarf:


znarf:
@clarete Ok, hope this doesn't break anything. Some issues were properly catched by the tests, which is great!

xdamman:
Do you know why it was considered unsafe to use $ aliases?

znarf:
They write about it here:
http://docs.sequelizejs.com/manual/tutorial/querying.html#operators-security

Doesn't sound a high liability to me. But they're gonna remove it from future major releases.

PR #1285, opened 2018-05-10
GraphQL: add 'transactions' query to return all transactions with paginated results

HipsterBrown:
To support the new homepage design, I'm working on creating and updating graphql queries to provide the necessary data. Most of the API design decisions are inspired by the [GitHub GraphQL API](https://developer.github.com/v4/explorer/) as it seems to be one of the more mature, real-world examples. 

This PR adds a `transactions` query to eventually replace the `allTransactions` query to support a paginated return value that includes `total, limit, offset, transactions` fields. `transactions` also does not require `CollectiveId` or `CollectiveSlug` arguments. In reality, the use of `allTransactions` should be replaced by using the `Collective` query and selecting the `transactions` field off that relationship. 

This PR is also using snapshot testing for verifying the result of graphql queries to simplify testing. Interested in feedback on this pattern. cc: @znarf @clarete

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha1-mlldxgJlS3FxiKKUUHCHpR55+6M=",
       "optional": true,
       "requires": {
-        "nan": "2.8.0"
+        "nan": "^2.4.0"
       }
     },
     "@types/debug": {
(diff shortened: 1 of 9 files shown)


znarf:
@HipsterBrown Good one.  Thank your for updating Mocha and introducing this new pattern for testing. That would be great to have a quick guide when someone wants also to do that in the future.

The GraphQL updates are looking pretty clean. Unfortunately, I haven't touched much this part yet and can't give you much feedback.

HipsterBrown:
@znarf I'll add a section to the README about the testing setup and usage. :+1:

PR #1298, opened 2018-05-21
Search: daily cron job for removing deleted collectives from index

HipsterBrown:
In order to keep the Algolia search index accurate, we need to remove deleted collectives from the service. I chose to use a daily cron job instead of an `afterDestroy` hook because the update does not need to be immediate.

This script will get all COLLECTIVE and ORGANIZATION type collectives deleted the day before and send a `deleteObjects` request to the Algolia index. If no collectives were deleted in the past day, no request is sent.

### DIFF
--- a/cron/daily/clean_search_index.js
+++ b/cron/daily/clean_search_index.js
@@ -1,0 +1,77 @@
+import config from 'config';
+import algoliasearch from 'algoliasearch';
+import { Op } from 'sequelize';
+import moment from 'moment-timezone';
+
+import models from '../../server/models';
+import { types as collectiveTypes } from '../../server/constants/collectives';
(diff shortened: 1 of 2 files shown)


cron/daily/clean_search_index.js
```diff
@@ -0,0 +1,77 @@
+import config from 'config';
+import algoliasearch from 'algoliasearch';
+import { Op } from 'sequelize';
+import moment from 'moment-timezone';
+
+import models from '../../server/models';
+import { types as collectiveTypes } from '../../server/constants/collectives';
+import emailLib from '../../server/lib/email';
+import debugLib from 'debug';
```
clarete: I usually group the absolute imports on top.

PR #1299, opened 2018-05-23
feat(graphql): Add pagination params to query allCollectives

clarete:


server/graphql/queries.js
```diff
@@ -540,14 +540,28 @@ const queries = {
       if (args.tags) query.where.tags = { [Op.overlap]: args.tags };
 
       if (args.orderBy === 'balance' && (args.ParentCollectiveId || args.HostCollectiveId || args.tags)) {
-        return rawQueries.getCollectivesWithBalance(query.where, args);
+        const queryResult = await rawQueries.getCollectivesWithBalance(query.where, args);
+        console.log(queryResult);
```
HipsterBrown: Can we remove the `console.log`?

clarete: Done in another PR! Thanks for this and the limit/offset catch!

PR #1300, opened 2018-05-23
fix(getCollectivesWithBalance): Limit/offset doesn't apply to total

clarete:


server/lib/queries.js
```diff
@@ -272,18 +280,19 @@ const getCollectivesWithBalance = (where = {}, options) => {
...
-  {
-    bind: where,
-    model: models.Collective
-  });
+    ORDER BY ${orderBy} ${orderDirection} NULLS LAST
+  `.replace(/\s\s+/g, ' '); // remove the new lines and save log space
+
+  const totalQuery = await sequelize.query(sql('COUNT(*) as "total"'), params);
+  const [{ dataValues: { total } }] = totalQuery
+  const collectives = await sequelize.query(`${sql(allFields)} LIMIT ${limit} OFFSET ${offset}`, params);
```
HipsterBrown: You can parallelize the work for the two queries:

```es6
const [
  totalQuery,
  collectives,
] = await Promise.all([
  sequelize.query(sql('COUNT(*) as "total"'), params),
  sequelize.query(`${sql(allFields)} LIMIT ${limit} OFFSET ${offset}`, params),
]);
```

server/graphql/queries.js
```diff
@@ -540,14 +540,10 @@ const queries = {
...
-        console.log(queryResult);
-        return {
-          total: queryResult.total,
-          collectives: queryResult,
-          limit: args.limit,
-          offset: args.offset,
-        };
+        const { total, collectives } = await rawQueries.getCollectivesWithBalance(query.where, args);
+        console.log('TOTAL', total);
+        console.log('COLLECTIVES', collectives);
```
HipsterBrown: Don't forget to remove these 😉

PR #1304, opened 2018-05-29
Enable email debugging in local environments

sloria:
- Log email text to console in dev environment
- Also fixes incorrect port in .env.default

### DIFF
--- a/.env.default
+++ b/.env.default
@@ -2,5 +2,6 @@ NODE_ENV=development
 API_KEY=dvl-1510egmf4a23d80342403fb599qd
 PG_DATABASE=opencollective_dvl
 PORT=3060
-WEBSITE_URL=http://localhost:3000
+WEBSITE_URL=http://localhost:13000
 SESSION_SECRET=i&j@/V6Wx.`g>Aq?qQQF(>u)/Erm;3A=
+DEBUG=email
--- a/server/lib/email.js
+++ b/server/lib/email.js
@@ -156,6 +156,7 @@ const sendMessage = (recipients, subject, html, options = {}) => {
     });
   } else {
     debug(">>> mailgun not configured");
+    debug(options.text);
     return Promise.resolve();
   }
 };


clarete:
Nice! Thanks for both fixes! \o/

.env.default
```diff
@@ -2,5 +2,6 @@ NODE_ENV=development
 API_KEY=dvl-1510egmf4a23d80342403fb599qd
 PG_DATABASE=opencollective_dvl
 PORT=3060
-WEBSITE_URL=http://localhost:3000
+WEBSITE_URL=http://localhost:13000
```
znarf: Note that this is wrong when not using Docker, and that this will be overhidden by the Docker environment variables anyway, so unnecessary.

See:
https://github.com/opencollective/opencollective-api/blob/master/docker/docker-compose.yml

sloria: Ah ok. The reason I noticed this is that emails use this envvar when building URLS, so links in emails were broken when using Docker.

znarf: @sloria Thank you for the feedback. We will need to check what are the actual priorities between the rules set in the Docker environment and the one set in `.env`.

clarete: I could be wrong, but I think the file overwrites the values for the application because it's loaded within docker.

I added a little bit more of context on the PR that you referenced @znarf! Thanks a lot! We'll get that running smoothly soon!

patcon:
I think there's already DEBUG logic for when the envvar is set to "preview", so could we just add this:

```diff
diff --git a/docker/docker-compose.yml b/docker/docker-compose.yml
index b6a3227..8c9e126 100644
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -6,6 +6,7 @@ services:
     environment:
       NODE_ENV: "docker"
       SEQUELIZE_ENV: "docker"
+      DEBUG: "preview"
     volumes:
       - ..:/usr/src/api
     ports:
diff --git a/server/lib/email.js b/server/lib/email.js
index 94e63f9..e382edc 100644
--- a/server/lib/email.js
+++ b/server/lib/email.js
@@ -33,6 +33,7 @@ const render = (template, data) => {
 
   // When in preview mode, we export an HTML version of the email in `/tmp/:template.:slug.html`
   if (process.env.DEBUG && process.env.DEBUG.match(/preview/)) {
+    console.log(text);
     filepath = `/tmp/${template}.${slug}.html`;
     const script = `<script>data=${JSON.stringify(data)};</script>`;
     fs.writeFileSync(filepath, `${html}\n\n${script}`);
```

fwiw the PR as merged didn't actually write anything to logs, for whatever reason (I'm working in Docker too)

PR #1309, opened 2018-05-30
Docker: add hostnames, allow frontend to reach api

znarf:


clarete:
That's great. I didn't know about the existence of the `hostname` parameter. The problem that we were having was that the frontend needed to access the API both from its little server (docker to docker) and also from the browser (host machine to docker). \o/

znarf:
My problem now is that `http://api:3060` works from the frontend (docker to docker) but not from my browser (host machine to docker). Would be great if it could work in both cases.

But today it's not mandatory for the frontend to work properly, because all the client side API requests are now going to the frontend API proxy.

PR #1310, opened 2018-05-30
Whitelist anchors for updates creation

Betree:
Whitelisted anchors:
- a
- h1
- h2
- strong
- img
- ul
- li

Fix for https://github.com/opencollective/frontend/issues/462

### DIFF
--- a/server/graphql/mutations/updates.js
+++ b/server/graphql/mutations/updates.js
@@ -10,11 +10,12 @@ function require(args, path) {
 
 export async function createUpdate(_, args, req) {
   const CollectiveId = get(args, 'update.collective.id');
+  const allowedExtensions = ['a', 'h1', 'h2', 'strong', 'img', 'ul', 'li'];
   mustHaveRole(req.remoteUser, 'ADMIN', CollectiveId, "create an update");
   require(args, 'update.title');
 
   const markdown = args.update.markdown
-        ? strip_tags(args.update.markdown)
+        ? strip_tags(args.update.markdown, allowedExtensions)
         : '';
 
   const update = await models.Update.create({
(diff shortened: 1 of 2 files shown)


HipsterBrown:
Thanks for creating a PR! :clap: The code looks good. I'll test it out locally and see if we have any test coverage we can add to that function.

znarf:
@Betree Thank you, I'll check the DEBUG=email issue.

HipsterBrown:
@Betree `strip_tags` appears to called in the `UpdateType` (`/server/graphql/types.js`) and parses the markdown text before being handled by the `createUpdate` function, so the `h1` and `h2` tags are stripped before your change. 

To fix that issue, I think it's fine to modify the `strip_tags` function to add `h1` and `h2` as default allowed tags. That change should also allow those tags after editing an update as well.

HipsterBrown:
@Betree sorry about the delayed response. Those _should_ allowed by default in for the `sanitize-html` package (https://www.npmjs.com/package/sanitize-html#what-are-the-default-options).

HipsterBrown:
@Betree any updates on this? Can you help with anything?

Betree:
Hi @HipsterBrown, I'm still unable to connect with an account in dev - event with `DEBUG=email`. Maybe I missed something. I'd like to visually confirm that it works before pushing :wink:

HipsterBrown:
@Betree Makes sense. 😄 Did you set the `DEBUG` environment variable in the `docker-compose.yml`, or somewhere else?

Betree:
I just retried with setting DEBUG in `the docker-compose.yml` and it worked, I probably missed something the other day :wink: 

It now works as intended :
selection_495

Anyway the short resume from collective's homepage doesn't look very nice. Should we open another issue for this?

selection_496

HipsterBrown:
@Betree Great work! 

> Anyway the short resume from collective's homepage doesn't look very nice. Should we open another issue for this?

Yes, can you create a separate issue for this bug? Then we can merge your work. :smile:

Do you have a Twitter we can mention when we thank you for this fix?

Betree:
New issue in #1327. My twitter handle is @Betree83 :wink:

PR #1311, opened 2018-05-30
Install GraphicsMagick on the frontend docker container

patcon:
In the frontend README, installing GraphicsMagick is instructed, but then it's not happening in the docker container.

(Seems this was added 5 months ago, so seems that perhaps no one else is using docker, but i really appreciate it being supported!)

Thanks!

### DIFF
--- a/docker/frontend/Dockerfile
+++ b/docker/frontend/Dockerfile
@@ -3,3 +3,6 @@ FROM node:8.4.0
 VOLUME /usr/src/frontend
 
 WORKDIR /usr/src/frontend
+
+RUN apt-get update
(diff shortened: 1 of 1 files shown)


znarf:
Sounds good, thank you.

clarete:
Hi @patcon thanks a lot for the contribution. That will save us some time every time we run that container! Do you think it'd be OK if you just changed the apt command to be in one line? Something like this?

```
RUN apt-get update -y && apt-get install -y graphicsmagick 
```

That way the install won't ever be executed against an old apt index.

clarete:
Thank you for the updates @patcon !!!

PR #1312, opened 2018-05-31
Add INTERNAL_API_URL envvar for docker frontend

patcon:
Builds on https://github.com/opencollective/opencollective-api/pull/1311

Related: https://github.com/opencollective/frontend/issues/467


znarf:
@patcon INTERNAL_API_URL is not needed in that case, we should just use API_URL.

There is actually a PR to fix this issue:
https://github.com/opencollective/opencollective-api/pull/1309

Im gonna merge it.

PR #1313, opened 2018-05-31
Add explanation of how to run scripts

patcon:


clarete:
Thanks and sorry for taking so long to apply this helpful improvement!

patcon:
No prob at all! I'm not even sure if it's how y'all run them, but worked for me :)

clarete:
I happen to have a tiny script that I call `dkr` with the following contents:

```sh
#!/bin/sh
[ -z "$1" ] && echo "Usage: $0 {docker-compose command} [params...]" && exit 1
COMMAND=$1; shift;
COMMAND_PARAMS=$([ "$COMMAND" = "up" ] && echo "--build")
PROJECT=oc
exec docker-compose \
     --file docker/docker-compose.yml \
     --project-name=${PROJECT} \
     ${COMMAND} ${COMMAND_PARAMS} $@
```

And then I use it like this:

```sh
 $ ./dkr up
```

If you think it's useful, I can push it to the repo. Also it corrects the annoyance of creating the docker containers prefixed with the name of the directory (docker) instead of using the name of the project.

PR #1338, opened 2018-06-19
Self serve flow

xdamman:
- don't automatically assign a host when creating a new collective
- filter memberOf by type of collective
- allow changing host of collective if empty balance
- custom redirect and post action after connecting stripe account
- added isPublicHost as a parameter to allCollectives


server/lib/forest.js
```diff
@@ -281,7 +281,7 @@ function getForestModels () {
   // Hidden in Forest
   delete m.Collective.associations.Activities;
   delete m.Collective.associations.Notifications;
-  delete m.Collective.associations.ConnectedAccounts;
+  // delete m.Collective.associations.ConnectedAccounts; // xdamman: this is breaking the `allCollectives` graphql query: error: ConnectedAccount is not associated to Collective!
```
xdamman: @znarf can you look into this?

znarf: @xdamman we're doing a cloneDeep on top of the function, so why a change here would impact the application?

PR #1342, opened 2018-06-21
fix monthly report for collectives + new and lost backers

xdamman:


xdamman:
I've added goals and latest updates:

PR #1369, opened 2018-07-16
add minBackerCount to allCollectives query w/ getCollectivesWithMinBackers, memoized

HipsterBrown:
To help with the optimization of the home page load on the GraphQL API, I've memoized some of the heavier queries, including a new query for showing collectives with a minimum number of backers.

### DIFF
--- a/server/graphql/queries.js
+++ b/server/graphql/queries.js
@@ -562,9 +562,17 @@ const queries = {
         type: GraphQLString,
         description: "Fetch all collectives hosted by hostCollectiveSlug"
       },
+      isActive: {
+        description: 'Only return active collectives',
+        type: GraphQLBoolean,
+      },
       memberOfCollectiveSlug: {
         type: GraphQLString,
         description: "Fetch all collectives that `memberOfCollectiveSlug` is a member of"
+      },
+      minBackerCount: {
+        description: 'Filter collectives with this minimum number of backers',
+        type: GraphQLInt,
       },
       role: {
         type: GraphQLString,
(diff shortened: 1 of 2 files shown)


server/graphql/queries.js
```diff
@@ -915,6 +915,46 @@ const queries = {
         total,
       };
     }
+  },
+
+  activeCollectives: {
+    type: CollectiveSearchResultsType,
+    description: 'An opinionated list of Collective with a minimum number of backers',
+    args: {
+      backerCount: {
```
xdamman: is that the only difference with the `allCollectives` query?
If so, cannot we merge the two to reduce the query surface?
Maybe also rename `backerCount` to `minBackerCount`? (the former makes conveys that you will be returning collectives with that exact number of backers)

server/lib/queries.js
```diff
@@ -667,6 +670,51 @@ const getTotalNumberOfDonors = () => {
...
+  };
+
+  const sql = (fields) => `
+    with "actives" as (
+      SELECT c.id
+      FROM "Collectives" c
+      LEFT JOIN "Members" m ON m."CollectiveId" = c.id
+      WHERE
+        c.type = 'COLLECTIVE'
+        AND m.role = 'BACKER'
```
xdamman: need to add `AND m."deletedAt" IS NULL`

server/lib/queries.js
```diff
@@ -667,6 +670,51 @@ const getTotalNumberOfDonors = () => {
...
 
+const getCollectivesWithMinBackers = memoize(async ({ backerCount = 0, orderBy = 'createdAt', orderDirection = 'ASC', limit = 0, offset = 0 }) => {
+  const params = {
+    model: models.Collective,
+  };
+
+  const sql = (fields) => `
+    with "actives" as (
+      SELECT c.id
+      FROM "Collectives" c
```
xdamman: Better to add the conditions on `Collectives` here to reduce the selectivity before doing the `JOIN` (i.e. `WHERE c."isActive" IS TRUE and c."deletedAt" IS NULL`

server/lib/queries.js
```diff
@@ -667,6 +670,51 @@ const getTotalNumberOfDonors = () => {
...
+      WHERE
+        c.type = 'COLLECTIVE'
+        AND m.role = 'BACKER'
+        GROUP BY c.id
+        HAVING count(m."MemberCollectiveId") >= ${backerCount}
+    )
+    SELECT ${fields} from "Collectives" c
+    INNER JOIN "actives" a on a.id = c.id
+    WHERE c."isActive" IS TRUE
+    AND c."deletedAt" IS NULL
```
xdamman: those two lines can go if used above

PR #1373, opened 2018-07-18
Collective.stats.burnrate + simplified raw query for getSponsors

xdamman:
Branched off from https://github.com/opencollective/opencollective-api/pull/1371
Required for https://github.com/opencollective/opencollective/issues/1176

### DIFF
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -1,5 +1,6 @@
 import models, { sequelize, Op } from '../models';
 import currencies from '../constants/currencies'
+import Promise from 'bluebird';
 import config from 'config';
 import { memoize, pick } from 'lodash';
 memoize.Cache = Map;
@@ -405,7 +406,6 @@ const getTopSponsors = () => {
  * - id 51 and 9804 (wwcode host)
  */
 const getCollectivesOrderedByMonthlySpendingQuery = async ({ where = {}, orderDirection = "ASC", limit = 0, offset = 0 }) => {
-  where.type = where.type || 'ORGANIZATION'; // setting the default type
   const whereStatement = Object.keys(where).reduce((statement, key) => `${statement} AND c."${key}"=:${key}`, '');
 
   const d = new Date;
@@ -437,8 +437,13 @@ const getCollectivesOrderedByMonthlySpendingQuery = async ({ where = {}, orderDi
     ORDER BY "monthlySpending" ${orderDirection} NULLS LAST
   `.replace(/\s\s+/g, ' ');
 
-  const [ [ { dataValues: { total } } ], collectives ] = await Promise.all([
-    sequelize.query(`${sql('COUNT(c.*) OVER() as "total"')} LIMIT 1`, params),
+  // If we use this query to get the monthlySpending of one single collective, we don't need to perform a count query
+  const countPromise = (typeof where.id === 'number')
+    ? Promise.resolve(1)
+    : sequelize.query(`${sql('COUNT(c.*) OVER() as "total"')} LIMIT 1`, params).then(res => res.length === 1 && res[0].dataValues.total);
+
+  const [ total, collectives ] = await Promise.all([
+    countPromise,
     sequelize.query(`${sql('c.*')} LIMIT ${limit} OFFSET ${offset}`, params),
   ]);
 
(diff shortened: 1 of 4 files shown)


server/lib/queries.js
```diff
@@ -397,39 +399,41 @@ const getTopSponsors = () => {
...
- * Returns sponsors ordered by total amount donated
- * (excluding open source collective id 9805)
+ * Returns collectives ordered by burn rate (total amount spent last month)
+ * excluding:
+ * - id 9805 (open source collective host)
+ * - id 1 (opencollective-company)
+ * - id 9804 (wwcode host)
  */
-const getSponsors = memoize(async (where = {}, { orderDirection = 'ASC', limit = 0, offset = 0 }) => {
+const getCollectivesOrderedByTotalAmountSpentLastMonthQuery = async (query) => {
```
HipsterBrown: Can we continue to use the default arguments?

```es6
const getCollectivesOrderedByTotalAmountSpentLastMonthQuery = async ({
  where = {},
  orderDirection = 'ASC',
  limit = 0,
  offset = 0,
}) => {
  // rest of code
};
```

This makes the `query` argument more transparent and cleans up the body of the function.

server/lib/queries.js
```diff
@@ -397,39 +399,41 @@ const getTopSponsors = () => {
...
   const whereStatement = Object.keys(where).reduce((statement, key) => `${statement} AND c."${key}"=$${key}`, '');
   const params = {
     bind: where,
-    model: models.Collective,
+    model: models.Collective
   };
 
+  const d = new Date;
+  d.setMonth(d.getMonth() - 1);
+  const since = moment(d).format('YYYY-MM-DD');
```
HipsterBrown: If we're using moment, let's use it for the Date math too. 

```es6
const since = moment().subtract(1, 'months').format('YYYY-MM-DD');
```

server/lib/queries.js
```diff
@@ -397,39 +399,41 @@ const getTopSponsors = () => {
...
-    LEFT JOIN "sponsors" s on s.id = c.id
+    SELECT c.id, -SUM(amount) as "burnrate", ${fields} from "Collectives" c
+    LEFT JOIN "Transactions" t on t."CollectiveId" = c.id
     WHERE c."isActive" IS TRUE
-    ${whereStatement}
-    AND c."deletedAt" IS NULL
-    GROUP BY c.id, s.id, s."amountSent"
-    ORDER BY s."amountSent" ${orderDirection} NULLS LAST
+      AND c."deletedAt" IS NULL
+      AND c.id NOT IN (1, 51, 9804, 9805)
```
HipsterBrown: What is id 51?

server/lib/queries.js
xdamman: Ah, I thought that

```
const func = ({ limit: 10, offset: 0 }) => {
console.log(limit);
}
```

Would return `undefined` when we call `func({offset: 1})`, but apparently, it's not. So cool, will change.

server/models/Collective.js
```diff
@@ -1270,11 +1270,29 @@ export default function(Sequelize, DataTypes) {
     .then(result => Promise.resolve(-parseInt(result.toJSON().total, 10)));
   };
 
-  // Get the total amount spent last month (burn rate)
-  Collective.prototype.getTotalAmountSpentLastMonth = function() {
-    const lastMonth = new Date;
-    lastMonth.setMonth(lastMonth.getMonth() - 1);
-    return this.getTotalAmountSpent(lastMonth);
+  // Get the average monthly spending based on last 90 days
+  Collective.prototype.getMonthlySpending = function() {
```
HipsterBrown: Can this reuse the raw query so we don't have repeated raw sql around the codebase? 

```es6
import { getCollectivesOrderedByMonthlySpending } from '../lib/queries';

// in Collective.prototype.getMonthlySpending
const { collectives } = getCollectivesOrderedByMonthlySpending({ where: { id: this.id }});
return collectives[0].monthlySpending;
```

xdamman: Mmmm... I see your point. But the problem is that one query is supposed to return a list of collectives whereas the other is supposed to return the actual average monthly spending for a given collective. I could consolidate both queries and make a condition that if `limit===1` we return directly the number? And if limit > 1 we return the total numbers of results and the current page of results? Wdyt?

HipsterBrown: Why make a condition in the raw query if you can pass limit 1 and where the id = the collective id? You're using the raw query inside the `getMonthlySpending` method of the Collective model, so the raw query doesn't need to be concerned with how the returned result is used.

xdamman: Because it seems like a waste to make two queries to get number of results, then the result for that second case.

HipsterBrown: The queries are performed parallel, so it feels like a premature optimization to avoid a single `count` query in favor of changing the return value of a reusable query.

xdamman: Except there is no loaders for computing that `monthlySpending` yet so the query is run for every result of a list. So when when we fetch 10 collectives, we are doing 20 queries instead of 10 to get that value for each.

(that said, I still think that you have a valid point. I'll adapt the query accordingly. One quick small optimization is to not perform the count query if there is a condition on `Collective.id`)

Thanks for the constructive exchange here. It feels like we are improving the code! :-)

PR #1375, opened 2018-07-19
Update amount precision in expense approval email

luceos:
fixes opencollective/opencollective#1177


xdamman:
Thanks! :-)

luceos:
Is leaving the precision to 0 in the subject intentional @xdamman ?

I noticed that it is rounded for all subject in mails about expenses.

PR #1383, opened 2018-07-26
W9 bot

marcogbarcellos:
This PR is trying to close the issue [1155](https://github.com/opencollective/opencollective/issues/1155), here are some highlights:

- Creates a **BOT** type for collectives.
- Creates a **bot** for open collective so this bot will be initially responsible to communicate with users through the expenses comments.
- Triggering to check if user expenses reached 600USD or more to then create a comment on the expense.
- Tests to validate notifications business logic.


migrations/20180725202700-populate-collective-with-opencollective-bot.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const Promise = require('bluebird');
+
+const insert = (sequelize, table, entry) => {
+  console.log(sequelize);
+  return sequelize.query(`
+    INSERT INTO "${table}" ("${Object.keys(entry).join('","')}") VALUES (:${Object.keys(entry).join(",:")})
+  `, { replacements: entry });
```
znarf: Can we figure a better syntax, using sequelize insert?

migrations/20180725202700-populate-collective-with-opencollective-bot.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const Promise = require('bluebird');
```
znarf: Any reason not to simply rely on the native Promise?

server/constants/collectives.js
```diff
@@ -4,9 +4,12 @@ export const types = {
   COLLECTIVE: 'COLLECTIVE',
   EVENT: 'EVENT',
   USER: 'USER',
-  ORGANIZATION: 'ORGANIZATION'
+  ORGANIZATION: 'ORGANIZATION',
+  BOT: 'BOT'
```
znarf: I think this is a good idea and will be reused in the future.

znarf:
Cool, you might wand to add `eslint` to your editor to get some extra warnings. Maybe automate, things like `trim_trailing_whitespace` and `insert_final_newline`.

From the CLI, you can also run ```npx eslint '**/*.js'```. We should actually add a `npm` script for that like we have on the frontend.

migrations/20180725202700-populate-collective-with-opencollective-bot.js
marcogbarcellos: I think we can, tbh I kinda copied the way the insertions were done from past [migration files](https://github.com/opencollective/opencollective-api/tree/master/migrations)(like [here](https://github.com/opencollective/opencollective-api/blob/master/migrations/201709110000-DoubleEntry.js#L18-L27) and [here](https://github.com/opencollective/opencollective-api/blob/master/migrations/201707140000-GroupToCollective.js#L99-L103) for example) as most of them use the same function(which actually could be an util function or something like that as the same code is there for several times). But yeah I will definitely find an easier way to do it. thanks

migrations/20180725202700-populate-collective-with-opencollective-bot.js
marcogbarcellos: I agree with you, but I have seen bluebird in some files of the project(but for the migration folder it's there most of the times) so I was wondering if we have a standard for it or today we actually use both bluebird and native promises. Anyways, the native one will do the job here so I'll remove.

server/constants/collectives.js
marcogbarcellos: 👍

migrations/20180725202700-populate-collective-with-opencollective-bot.js
znarf: Yes, let's skip bluebird for new code unless there is good reasons to use it.

server/lib/notifications.js
```diff
@@ -137,6 +141,61 @@ export async function notifyAdminsOfCollective(CollectiveId, activity, options =
   return notifySubscribers(adminUsers, activity, options);
 }
 
+async function notifyAdminsOfCollectiveAndCheckUserExpensesAmount(activity) {
+  try {
+    await notifyAdminsOfCollective(activity.data.collective.id, activity);
+    await notifyUserIfExpensesExceededMinimumFormAmount(activity);  
+  } catch (error) {
+    console.log(error);
```
HipsterBrown: Is there a plan to handle this error somewhere? Is there a way to notify the developers of an error?

server/constants/collectives.js
xdamman: Just need to watch out for potential issues in the frontend code that may expect `ORGANIZATION` or `USER` (a quick check feels like it should be ok but we need to test).

I also like this pattern to create 3rd party bots or apps. One could add such a bot as a member of the collective or host organization to activate the bot. And the bot could perform tasks that any other member/admin (depending on the permissions given to the bot when added as a member of a collective) could do.

marcogbarcellos:
@znarf I once more agree with you on `eslint` but as far as I know if you run `npm run test` it will triggers `eslint`(@xdamman can confirm or unconfirm it).  Perhaps we can try to improve the existing rules and make it standard not only for this project. 

From the CLI, you can also run ```npx eslint '**/*.js'```. We should actually add a `npm` script for that like we have on the frontend.

server/lib/notifications.js
```diff
@@ -137,6 +141,61 @@ export async function notifyAdminsOfCollective(CollectiveId, activity, options =
...
+        CollectiveId: activity.data.collective.id,
+        ExpenseId: activity.data.expense.id,
+        html: expensesFormData.FORM_EXPENSE_COMMENT_HTML,
+        FromCollectiveId: (fromCollectiveIdData ? fromCollectiveIdData.id : null)
+      };
+      return models.Comment.create(commentData);
+    }
+    return Promise.resolve(true);
+  } catch (error) {
+    throw error;
```
HipsterBrown: In a `try...catch` block, the error is caught if it is thrown in the `try` block so throwing it again after catching it is redundant. Is there a need for this?

server/lib/notifications.js
marcogbarcellos: I consoled log instead of throwing an error in this case because this method is executed asynchronously. But that's something I noted we need to improve: error handling, we need to do `trycatch` is most async await functions and then either throw an error or at least log it properly( using `debug` or a log library). For now I can use the `debug` library instead of console.log as we're using in other places as well, but that's definitely something we can agree on always take the same actions.

server/lib/notifications.js
```diff
@@ -137,6 +141,61 @@ export async function notifyAdminsOfCollective(CollectiveId, activity, options =
...
+      });
+      const commentData = {
+        CollectiveId: activity.data.collective.id,
+        ExpenseId: activity.data.expense.id,
+        html: expensesFormData.FORM_EXPENSE_COMMENT_HTML,
+        FromCollectiveId: (fromCollectiveIdData ? fromCollectiveIdData.id : null)
+      };
+      return models.Comment.create(commentData);
+    }
+    return Promise.resolve(true);
```
HipsterBrown: Just returning `true` here _should_ do the same thing.

> An asynchronous function is a function which operates asynchronously via the event loop, using an implicit Promise to return its result.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

migrations/20180725202700-populate-collective-with-opencollective-bot.js
xdamman: that's the helper function that I wrote in the past. It would be indeed better to use the `sequelize.insert`. I'm not sure anymore why I did it this way. It was a long time ago and then I kept on copy pasting what worked. I think it might have to do with the ability to do a `DRY_RUN` and stack all queries before running them and/or with more control on the actual data being recorded in the database. Can't remember for sure. But since migration files are often about raw queries, this helper function has been quite helpful (and might be worth pulling it into a migration lib)

migrations/20180725202700-populate-collective-with-opencollective-bot.js
xdamman: Really useful if you need `Promise.map` or `Promise.props`

server/lib/notifications.js
xdamman: there are plenty of places throughout the code where we don't do much to handle those errors.
What would you recommend @HipsterBrown?

We have emails being sent to the server-errors email group but it's a bit overwhelming and it hasn't been super helpful in my view. What would be better here?

znarf:
@marcogbarcellos We have common eslint rules for all our projects, see: https://github.com/opencollective/eslint-config-opencollective. They are setup on all our major repositories.

In API, `npm run test` is effectively triggering eslint but `eslint '**/*.js' --quiet`, this mean it's only failing on errors, not on warnings. On purpose.

As our eslint configuration is also generating warnings, It's great to fix them but not mandatory. Yet :-)

server/constants/collectives.js
```diff
@@ -4,9 +4,12 @@ export const types = {
...
 };
 
+export const OPENCOLLECTIVE_BOT_NAME = 'Open Collective bot';
```
xdamman: Let's do `Open Collective Bot`. Since you are creating a collective for it, I wouldn't have a constant for the name. Just save it in the database as `collective.name`.

xdamman: Now that I think of it, maybe we should call this bot "IRS Bot" (Internal Revue Service), since this particular feature is really about the IRS.

server/lib/notifications.js
marcogbarcellos: What I used to do in the past is one or 2 actions:

- log the error using a log library(love [pino](https://github.com/pinojs/pino) )
- Warn devs through some third party modules (Only if it was necessary)

server/lib/notifications.js
marcogbarcellos: You're right, as said in my comment I believe we should either just log it or log and warn devs if it feels like it's important, but at the very least log it. Will fix.

marcogbarcellos:
@znarf @xdamman @HipsterBrown I feel like I can update my PR using a log library, do you already have a standard log library or can I just use one? If I were to choose I'd use [pino](https://github.com/pinojs/pino) if you guys are okay with that.

server/constants/comments.js
```diff
@@ -0,0 +1,8 @@
+
+export const MAXIMUM_AMOUNT_TO_SEND_FORM_COMMENT = 60000; // U$ 600.00 total amount allowed without form as of July 2018
```
xdamman: For the same reason than above, I'd keep this in `collective.settings`.
Maybe a better name for this variable (naming is np-hard!) would be simply `settings.thresholdW9` now that this would be part of the `IRS Bot`. Makes things self explanatory.

server/constants/comments.js
```diff
@@ -0,0 +1,8 @@
...
+export const expensesFormData = {
+  FORM_EXPENSE_COMMENT_HTML: '<p>You have now been paid $600 or more through Open Collective, which' +
```
xdamman: same here, I'd make this part of the `collective.settings`.

server/lib/notifications.js
marcogbarcellos: Good catch.

server/lib/notifications.js
```diff
@@ -137,6 +141,61 @@ export async function notifyAdminsOfCollective(CollectiveId, activity, options =
...
+  try {
+    const host = await models.Collective.findById(activity.data.host.id);
+    if (host.currency && host.currency !== 'USD') {
+      return;
+    }
+    // ideally change to consider the current fiscal year instead(Example: might be from
+    // march to february instead of from the beginning from january)
+    const beginningOfCurrentYear = new Date(new Date().getFullYear(), 0, 1);
+    const userExpenses = await models.Expense.findAll({
+      where: {
```
xdamman: better to also specify the `attributes` array to only fetch what you really need (in this case, it looks like you only need `['currency', 'amount', 'status']`

znarf:
@marcogbarcellos I would skip the log library and open a ticket about the general problem (lack of log library in api project). We have to think together about what we want, let's take some time. Fyi, n frontend world, we're using a mix of `debug` and `winston`.

server/lib/notifications.js
```diff
@@ -137,6 +141,61 @@ export async function notifyAdminsOfCollective(CollectiveId, activity, options =
...
+        const expenseDate = (expense.status === expenseStatuses.PAID) ? expense.updatedAt : 'latest';
+        expenseAmountInHostCurrency = await convertToCurrency(expense.amount, expense.currency, host.currency, expenseDate);
+      }
+      totalAmount += expenseAmountInHostCurrency;
+    }
+    // U$ 600.00 total amount allowed without form as of July 2018
+    if (totalAmount > MAXIMUM_AMOUNT_TO_SEND_FORM_COMMENT) {
+      const fromCollectiveIdData = await models.Collective.findOne({ 
+        where: { 
+          name: OPENCOLLECTIVE_BOT_NAME
```
xdamman: I see why you wanted to have a constant with the name, but in that case I'd rather use `slug` or `id` which are unique (name is not); i.e. `IRS_BOT_COLLECTIVE_ID` or `IRS_BOT_COLLECTIVE_SLUG`

server/lib/notifications.js
```diff
@@ -172,7 +231,7 @@ async function notifyByEmail(activity) {
       break;
 
     case activityType.COLLECTIVE_EXPENSE_CREATED:
-      notifyAdminsOfCollective(activity.data.collective.id, activity);
+      notifyAdminsOfCollectiveAndCheckUserExpensesAmount(activity);
```
xdamman: this looks like overloading a function. I'd rather see two separate functions.
`notifyAdminsOfCollective` can be performed independently of the action of the bot.

test/graphql.expenses.test.js
```diff
@@ -328,6 +335,157 @@ describe('GraphQL Expenses API', () => {
       expect(res.errors).to.not.exist;
     }); /* End of "creates a new expense logged in and send email to collective admin for approval" */
 
+    it('creates a new expense greater than 600 USD but DO NOT create Comment because host is not USD based', async () => {
```
xdamman: I'd create a new test file `bot.irs.test.js` to put this -- that will also make it easier for people to create other bots.

test/graphql.expenses.test.js
```diff
@@ -328,6 +335,157 @@ describe('GraphQL Expenses API', () => {
...
+      // Given a collective
```
xdamman: // Given a collective in USD with a host in EUR

test/graphql.expenses.test.js
```diff
@@ -328,6 +335,157 @@ describe('GraphQL Expenses API', () => {
...
+      await utils.waitForCondition(() => emailSendMessageSpy.callCount > 0);
+      expect(emailSendMessageSpy.callCount).to.equal(1);
+      expect(emailSendMessageSpy.firstCall.args[0]).to.equal(admin.email);
+      expect(emailSendMessageSpy.firstCall.args[1]).to.equal("New expense on Test Collective: $700 for Test expense for pizza");
+      expect(emailSendMessageSpy.firstCall.args[2]).to.contain("/test-collective/expenses/1/approve");
+
+    }); /* End of "creates a new expense greater than 600 USD but DO NOT create Comment because host is not USD based" */
+
+    it('creates a new expense greater than 600 USD and create Comment expense forms email', async () => {
+      // Given a collective
```
xdamman: // Given a collective in USD with a host in USD

test/graphql.expenses.test.js
```diff
@@ -328,6 +335,157 @@ describe('GraphQL Expenses API', () => {
...
+      expect(emailSendMessageSpy.firstCall.args[0]).to.equal(admin.email);
+      expect(emailSendMessageSpy.firstCall.args[1]).to.equal("New expense on Test Collective: $700 for Test expense for pizza");
+      expect(emailSendMessageSpy.firstCall.args[2]).to.contain("/test-collective/expenses/1/approve");
+      expect(emailSendMessageSpy.secondCall.args[0]).to.equal(user.email);
+      expect(emailSendMessageSpy.secondCall.args[1]).to.contain("New comment on your expense");
+      
+    }); /* End of "creates a new expense greater than 600 USD and create Comment expense forms email" */  
+    
+    it('creates 2 new expenses that adds up more than 600 USD and create Comment expense forms email', async () => {
+      // Given a collective
```
xdamman: // Given a collective in EUR with a host in USD

migrations/20180725202700-populate-collective-with-opencollective-bot.js
znarf: If I was unclear, I generally prefer that we stick to native Promises. If we need one of the feature that are only available with Bluebird, we can just import it where needed.

server/constants/collectives.js
marcogbarcellos: sure. IRS coming along.

server/lib/notifications.js
marcogbarcellos: Humm Ok, I wrapped them up because i thought the order was important, first sending an email to the `admin` the creating the `comment` to trigger the `new comment` email. Isn't it?

test/graphql.expenses.test.js
marcogbarcellos: I agree with you somehow but at the same time it's also testing the Graphql expenses creation(and it's additional feature) so I'm kinda not sure if moving this to a new file would make it somehow confusing. Do you disagree?

server/lib/notifications.js
xdamman: Keeping functions small and decoupling them is generally a good idea.
If something fails, it doesn't fail everything. Also easier to maintain.

The order here doesn't matter. Basically, whenever a new comment is created another service should pick up this event and do a bunch of things. We could even imagine different micro services, one for notifying followers of that thread (admins), one for each bot.
Even though we don't have those micro services them today, it's good to have that in mind and build our code in a way that will be easier to migrate to such architecture in the future.

znarf:
@xdamman @marcogbarcellos I'm afraid that having the bot named "IRS bot" will be confusing, having users think it's coming directly from the IRS. For that reason, I would avoid the "IRS" branding. What do you think?

marcogbarcellos:
@znarf that's a fair point. In my opinion I believe most people will get it right but we know that there's always someone(s) that can interpret it on the wrong way. Let's see what @xdamman has to say

xdamman:
@znarf that's a good point. Let's call it "W9 Bot". So that it's clear what it does. It asks for W9 (or W8-BEN which is the W9 equivalent for foreigners).

migrations/20180725202700-populate-collective-with-IRS-Bot.js
```diff
@@ -0,0 +1,38 @@
...
+};
+
+module.exports = {
+  up: (queryInterface, sequelize) => {
+    
+    // Inserting OpenCollective Bot Collective
+    const botData = {
+      name: 'IRS Bot',
+      mission: 'Support users through the platform',
+      description: 'IRS bot that support users regarding IRS issues through the platform',
```
xdamman:
```js
const botData = {
name: "W9 bot",
slug: "w9bot",
mission: "Help hosts by automating requesting users to submit their W9 or W8-BEN form when needed",
description: "Help hosts by automating requesting users to submit their W9 or W8-BEN form when needed",
longDescription: "Whenever someone files an expense to a host that has USD as its base currency, this bot will look at the sum of all past expenses of that user made during the year. If the sum exceeds $600, it will create a comment on the expense to ask to submit the W9 or W8-BEN form to the host",
...
}
```

migrations/20180725202700-populate-collective-with-IRS-Bot.js
```diff
@@ -0,0 +1,38 @@
...
+      longDescription: 'IRS bot that support users regarding IRS issues  through the platform',
+      currency: 'USD',
+      image: 'https://cldup.com/rdmBCmH20l.png',
+      isActive: true,
+      slug: 'irs-bot',
+      website: 'https://opencollective.com',
+      type: 'BOT',
+      settings: JSON.stringify({
+        thresholdW9: 60000,
+        thresholdW9HtmlComment: '<p>You have now been paid $600 or more through Open Collective, which' +
```
xdamman: They have not been paid $600 or more yet.
You can say instead: 
> The total amount of the expenses that you have submitted this year to this host exceeds $600. To comply with the IRS, we need you to send us a <a href="https://www.irs.gov/pub/irs-pdf/fw9.pdf">W9 form</a> if you are a US resident (or a <a href="https://www.irs.gov/pub/irs-pdf/fw8ben.pdf">W9-BEN form</a> if you are not a US resident) before we can proceed with this payment. <a href="https://github.com/opencollective/opencollective/wiki/Submitting-Expenses#taxes">More info on our wiki</a>.

server/constants/collectives.js
```diff
@@ -4,9 +4,12 @@ export const types = {
   COLLECTIVE: 'COLLECTIVE',
   EVENT: 'EVENT',
   USER: 'USER',
-  ORGANIZATION: 'ORGANIZATION'
+  ORGANIZATION: 'ORGANIZATION',
+  BOT: 'BOT'
 };
 
+export const IRS_BOT_SLUG = 'irs-bot';
```
xdamman: W9_BOT_SLUG = 'w9bot';

server/lib/notifications.js
```diff
@@ -8,7 +8,10 @@ import slackLib from './slack';
 import twitter from './twitter';
 import emailLib from '../lib/email';
 import activityType from '../constants/activities';
-import models from '../models';
+import expenseStatuses from '../constants/expense_status';
+import {IRS_BOT_SLUG} from '../constants/collectives';
```
xdamman: W9_BOT_SLUG

server/lib/notifications.js
```diff
@@ -137,6 +140,49 @@ export async function notifyAdminsOfCollective(CollectiveId, activity, options =
   return notifySubscribers(adminUsers, activity, options);
 }
 
+async function notifyUserIfExpensesExceededMinimumFormAmount(activity) {
```
xdamman: this is a very long function. I'd split it by creating a helper function attached to the `Expense` model:

```js
models.Expense.getTotalExpensesFromUserId = async function(UserId, since, until = new Date) {
  // ...
}
```

```js
async function w9bot(activity) {
  const hostCollective = await models.Collective.findById(activity.data.host.id);
  if (hostCollective.currency !== 'USD') return;
  // ideally change to consider the current fiscal year instead(Example: might be from
  // March to February instead of from the beginning of January)
  const beginningOfCurrentYear = new Date(new Date().getFullYear(), 0, 1);
  const totalAmountThisYear = await models.Expense.getTotalExpensesFromUserId(UserId, beginningOfCurrentYear);
  if (totalAmountThisYear > 60000) {
    // create comment
  }
}
```

server/lib/notifications.js
```diff
@@ -173,6 +219,7 @@ async function notifyByEmail(activity) {
 
     case activityType.COLLECTIVE_EXPENSE_CREATED:
       notifyAdminsOfCollective(activity.data.collective.id, activity);
+      notifyUserIfExpensesExceededMinimumFormAmount(activity);
```
xdamman: `w9bot(activity);`

migrations/20180725202700-populate-collective-with-IRS-Bot.js
marcogbarcellos: There is also the [Form W-8BEN-E](https://www.irs.gov/pub/irs-pdf/fw8bene.pdf) for Non-US companies, right? so I believe we should do something like:

> The total amount of the expenses that you have submitted this year to this host exceeds $600. To comply with the IRS, we need you to send us a <a href="https://www.irs.gov/pub/irs-pdf/fw9.pdf">W9 form</a> if you are a US resident (if you are not a US resident send <a href="https://www.irs.gov/pub/irs-pdf/fw8ben.pdf">W-8BEN form</a> for individuals or <a href="https://www.irs.gov/pub/irs-pdf/fw8bene.pdf">W-8BEN-E form</a> for companies ) before we can proceed with this payment. <a href="https://github.com/opencollective/opencollective/wiki/Submitting-Expenses#taxes">More info on our wiki</a>.

Issue #1386, opened 2018-07-27
AddMissingBackers migration failing 


znarf:
`ERROR: null value in column "CreatedByUserId" violates not-null constraint`

To reproduce:

- `dropdb opencollective_dvl`
- ` ./scripts/postinstall.sh` (restore the database)
- ` ./scripts/postinstall.sh` (run the migration)

xdamman:
Ah, I didn't see this issue (I should double check my notification settings)
I fixed this already last week in https://github.com/opencollective/opencollective-api/pull/1388

PR #1388, opened 2018-07-27
fix error on circleci with migration file + improve scripts/postinstall.sh readibility

xdamman:
This is an attempt to fix tests on https://github.com/opencollective/opencollective-api/pull/1383


xdamman:
This also fixes #1386

PR #1391, opened 2018-07-31
Add status field to Orders

HipsterBrown:
To facilitate creating pledges for collectives (see opencollective/opencollective#1168), this PR adds a `status` field to Orders:

- PENDING (default)
- PAID (one-time completed orders)
- ACTIVE (subscriptions)
- CANCELLED
- REJECTED
- ERROR

The migration also includes code to update existing Orders.

### DIFF
--- a/migrations/20180727173834-add-status-to-order.js
+++ b/migrations/20180727173834-add-status-to-order.js
@@ -1,0 +1,134 @@
+'use strict';
+import status from '../server/constants/order_status';
+import models, { sequelize } from '../server/models';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    const { Op } = Sequelize;
+
+    return queryInterface.addColumn('Orders', 'status', {
+      type: Sequelize.STRING,
+      defaultValue: status.PENDING,
+      allowNull: false,
+    })
+    .then(() => queryInterface.addColumn('OrderHistories', 'status', {
+      type: Sequelize.STRING,
+      defaultValue: status.PENDING,
+      allowNull: false,
+    }))
+    .then(() => sequelize.sync())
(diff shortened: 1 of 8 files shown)


server/graphql/mutations/orders.js
```diff
@@ -238,6 +239,10 @@ export function createOrder(order, loaders, remoteUser) {
     })
     .catch(e => {
       debugOrder("createOrder mutation error: ", e);
+      if (!orderCreated.processedAt) {
+        // TODO: Order should be updated with data JSON field to store the error to review later
+        orderCreated.update({ status: status.ERROR });
+      }
       throw e;
```
marcogbarcellos: Do you need to throw the error as you're already inside the catch?

HipsterBrown: I believe it was thrown again to bubble it back up to the graphql response, otherwise the frontend never sees the error.

migrations/20180727173834-add-status-to-order.js
```diff
@@ -0,0 +1,125 @@
+'use strict';
+import status from '../server/constants/order_status';
```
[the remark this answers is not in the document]
HipsterBrown: I agree. I used this convention to match the `expense_status.js` file. I'll create an issue to standardize this.

server/graphql/mutations/orders.js
marcogbarcellos: got it! LGTM then..

migrations/20180727173834-add-status-to-order.js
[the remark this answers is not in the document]
znarf: After having looked at several repositories, looks like the current common practice is ... hyphen case! So that would be `order-status`. Anyway, let's take care of that later.

PR #1393, opened 2018-08-01
Update ESLint configuration and fix style

marcogbarcellos:
This PR is regarding an update on the [eslint-config-opencollective](https://github.com/opencollective/eslint-config-opencollective) library, I also updated the other eslint related libraries and then got a bunch of errors and warnings, that's why this PR is big but it's all about eslint tweaks.


znarf:
I'm good with it.

Some of the rules (like the quotes or trailingComma rule) are not consensual yet within the team. This is why they were not automatically enforced so far.

.eslintignore
```diff
@@ -4,3 +4,4 @@ seeders/
 migrations/
 coverage/
 dist/
+test/
```
znarf: I don't think test should be ignored, it's code.

marcogbarcellos: humm ok, will remove it.

marcogbarcellos:
@znarf to be honest I have just followed the list `npx eslint '**/*.js'` has generated and removed the warns and errors, I don't have any strong opinion on changing or keeping the way it is(I think it's good the way it is) but if the team agrees to change the lint later we can just update it make the code to be compliant with the `npx eslint '**/*.js'` once more.

marcogbarcellos:
@znarf ping me if you think we can merge!! Thanks in advance!

znarf:
@marcogbarcellos there is now conflicts to be fixed

marcogbarcellos:
@znarf yes I think this will conflict with almost any new PR that's merged to master so I'll update the eslint file to included the space comment rule and will fix the conflicts to merge it.

znarf:
@marcogbarcellos also don't forget publishing the package on npm (ask me or xdamman).

marcogbarcellos:
@znarf created a PR on the eslint project( opencollective/eslint-config-opencollective#9 ) so whenever you or xavier publish the new i will updated the package.json here, resolve the conflicts and let you know it's ready to merge.

znarf:
Updated libraries in https://github.com/opencollective/opencollective-api/pull/1437

Skipping the formatting for now.

Issue #1396, opened 2018-08-02
totalDonations doubled in allMembers GraphQL query


znarf:
```
query Members($CollectiveId: Int!, $TierId: Int, $role: String, $type: String, $limit: Int, $offset: Int, $orderBy: String) {
  allMembers(CollectiveId: $CollectiveId, TierId: $TierId, role: $role, type: $type, limit: $limit, offset: $offset, orderBy: $orderBy) {
    id
    role
    collective {
      name
    }
    stats {
      totalDonations
    }
    tier {
      id
      name
    }
    member {
      id
      type
      name
      company
      description
      slug
      image
      backgroundImage
      website
    }
  }
}
```

With query variables:

```
{
  "CollectiveId":43,
  "offset":0,
  "type":"ORGANIZATION",
  "role":"BACKER",
  "orderBy":"totalDonations",
  "limit":100
}
````

We get:

```
{
  "data": {
    "allMembers": [
      {
        "id": 314,
        "role": "BACKER",
        "createdAt": "Tue Apr 12 2016 17:19:41 GMT+0000 (GMT)",
        "collective": {
          "name": "APEX"
        },
        "stats": {
          "totalDonations": 340000
        },
        "tier": {
          "id": 470,
          "name": "Sponsors"
        },
        "member": {
          "id": 2428,
          "type": "ORGANIZATION",
          "name": "PubNub",
          "company": null,
          "description": null,
          "slug": "pubnub",
          "image": "https://opencollective-production.s3-us-west-1.amazonaws.com/pubnublogopng_38ab9250-d2c4-11e6-8ba3-b7985935397d.png",
          "backgroundImage": null,
          "website": "https://www.pubnub.com/"
        }
      }
    ]
  }
}
```

Without the orderBy:

```
{
  "CollectiveId":43,
  "offset":0,
  "type":"ORGANIZATION",
  "role":"BACKER",
  "limit":100
}
````

We get:

```
{
  "data": {
    "allMembers": [
      {
        "id": 314,
        "role": "BACKER",
        "collective": {
          "name": "APEX"
        },
        "stats": {
          "totalDonations": 170000
        },
        "tier": {
          "id": 470,
          "name": "Sponsors"
        },
        "member": {
          "id": 2428,
          "type": "ORGANIZATION",
          "name": "PubNub",
          "company": null,
          "description": null,
          "slug": "pubnub",
[...]

PR #1397, opened 2018-08-02
Ignore deleted Transactions in queries

znarf:


znarf:
Fixes: https://github.com/opencollective/opencollective-api/issues/1396

PR #1404, opened 2018-08-06
Add Collective to Collective donation across different hosts(one time and recurring payments) 

marcogbarcellos:
Adding functionality so Collectives can pay other Collectives through different hosts. Also adding the feature `recurring` payments to the payment Method `opencollective`.

The way it's implemented is based on the issue [1196](https://github.com/opencollective/opencollective/issues/1196):

1. Collective pays other collective through `opencollective` method(both one time and recurring)
2. Method will look if both collectives have Stripe accounts setup, if so it will proceed with the payment through the credit card setup on the collective(that's paying) Stripe account.

Closes opencollective/opencollective#1196


xdamman:
You need to write "Fixes" or "Closes" followed by the issue to close it automatically when the PR is merged: https://help.github.com/articles/closing-issues-using-keywords/

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+    before('creates User 2', () => models.User.createUserWithCollective({ name: "User 2" }).then(u => user2 = u));
+    before('create Host 1', () => models.Collective.create({ name: "Host 1", currency: "USD", isActive: true }).then(c => host1 = c));
+    before('create Host 2', () => models.Collective.create({ name: "Host 2", currency: "USD", isActive: true }).then(c => host2 = c));
+    before('create Collective 1', () => models.Collective.create({ name: "Collective 1", currency: "USD", HostCollectiveId: host1.id, isActive: true }).then(c => collective1 = c));
+    before('create Collective 2', () => models.Collective.create({ name: "Collective 2", currency: "USD", HostCollectiveId: host1.id, isActive: true }).then(c => collective2 = c));
+    before('create Collective 3', () => models.Collective.create({ name: "Collective 3", currency: "USD", HostCollectiveId: host2.id, isActive: true }).then(c => collective3 = c));
+    // before('Add User 1 as Admin of Collective 1', () => collective1.editMembers([{ id: user1.id, MemberCollectiveId: user1.CollectiveId, role: 'ADMIN' }]));
```
xdamman: remove if not needed

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+      // the `totalAmount' field doesn't change throught the tests.
+      utils.stubStripeBalance(sandbox, ORDER_TOTAL_AMOUNT, 'usd',
+                              0,
+                              STRIPE_FEE_STUBBED_VALUE); // This is the payment processor fee.
+    });
+
+    afterEach(() => sandbox.restore());
+
+
+    it('collective balance for Admin Member matches sum of transactions amounts', async () => {
```
xdamman: I had to read this twice to understand. What about: `it('the available balance of the payment method of the collective should equal the balance of the collective', ...`

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+    before('create an organization', () => models.Collective.create({ name: "pubnub", currency: "USD" }).then(o => organization = o));
+    before('create a payment method', async () => models.PaymentMethod.create({
+      name: '4242',
+      service: 'stripe',
+      type: 'creditcard',
+      token: 'tok_123456781234567812345678',
+      CollectiveId: organization.id,
+      monthlyLimitPerMember: 10000
+    }).then(pm => stripePaymentMethod = pm));
+    beforeEach('create many transactions', async () => {
```
xdamman: I'd be more specific: "create transactions for 3 donations from organization to collective1"

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+      transactions = [
+        { amount: -500, netAmountInCollectiveCurrency: -500 },
+        { amount: -200, netAmountInCollectiveCurrency: -200 },
+        { amount: -1000,netAmountInCollectiveCurrency: -1000 }
+      ];
```
xdamman: it makes it hard to follow that those amounts are negative amounts (it looked like you were creating expenses).
I'd instead make those positive amounts and flip the `transactionsDefaultValue` `FromCollectiveId` and `CollectiveId`.

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+
+    afterEach(() => sandbox.restore());
+
+
+    it('collective balance for Admin Member matches sum of transactions amounts', async () => {
+      // getting balance of transactions that were just created
+      const reducer = (accumulator, currentValue) => accumulator + currentValue;
+
+      // multiplying by -1 to get the Collective 1 Credit values
+      const balance = (-1) * transactions.map(t => t.netAmountInCollectiveCurrency).reduce(reducer, 0);
```
xdamman: If the transactions above are now positive, you won't need this `-1` here (which was also hard to understand at first)

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+
+      //finding opencollective payment method for Collective 1
```
xdamman: need a space after //

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+      openCollectivePaymentMethod = await models.PaymentMethod.findOne({ where: {type: 'collective', CollectiveId: collective1.id}});
+
+      // get Balance given the created user
+      const ocPaymentMethodBalance = await openCollectivePaymentMethod.getBalanceForUser(user1);
+
+      expect(balance).to.equal(ocPaymentMethodBalance.amount);
+      expect(collective1.currency).to.equal(ocPaymentMethodBalance.currency);
+    });/** END OF "check if collective balance for Admin Member matches sum of transactions amounts" */
+
+    it('Collective 1 fails to pay Collective 2 because it is not logged in', async () => {
```
xdamman: `it('Non admin members can't use the payment method of the collective', ...`

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+        fromCollective: { id: collective1.id },
+        collective: { id: collective2.id },
+        paymentMethod: { uuid: openCollectivePaymentMethod.uuid },
+        totalAmount: ocPaymentMethodBalance.amount,
+      }
+      // Executing queries
+      const res = await utils.graphqlQuery(createOrderQuery, { order });
+
+      // Then there should be Errors
+      expect(res.errors).to.exist;
```
xdamman: would be good to double check that it's the right error message that we are getting here.

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+
+    });/** END OF "Collective 1 fails to pay Collective 2 because it is not logged in" */
+
+    it('Collective 1 fails to pay Collective 2 because user is not Admin', async () => {
```
xdamman: Can probably be consolidated with the previous test ("Non admin users can't use the payment method of a collective")

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+    it('Collective 1 sends money to Collective 2 with no fees(same host)', async () => {
+      // Add User 1 as Collective 1 Admin
+      await models.Member.create({
+        CreatedByUserId: user1.id,
+        MemberCollectiveId: user1.CollectiveId,
+        CollectiveId: collective1.id,
+        role: 'ADMIN'
+      });
+
+      //finding opencollective payment method for Collective 1
```
xdamman: space after //

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+      expect(orderCollective.id).to.equal(collective2.id);
+      expect(transaction.CollectiveId).to.equal(collective2.id);
+      expect(transaction.currency).to.equal(collective2.currency);
+      expect(transaction.platformFeeInHostCurrency).to.equal(0);
+      expect(transaction.hostFeeInHostCurrency).to.equal(0);
+      expect(transaction.paymentProcessorFeeInHostCurrency).to.equal(0);
+
+    });/** END OF "Collective 1 sends money to Collective 2 with no fees(same host)" */
+
+    it('Collective 1 fails to send money to Collective 3 because Collective 1\'s Host does not have a credit card', async () => {
```
xdamman: `it('Cannot send money to a different host if the host of the collective doesn\'t have a credit card')`

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+        fromCollective: { id: collective1.id },
+        collective: { id: collective3.id },
+        paymentMethod: { uuid: openCollectivePaymentMethod.uuid },
+        totalAmount: ocPaymentMethodBalance.amount/10,
+      }
+      // Executing queries
+      const res = await utils.graphqlQuery(createOrderQuery, { order }, user1);
+
+      // Then there should be Errors
+      expect(res.errors).to.exist;
```
xdamman: always check that the error is the right one (what if there is another error?)

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+    });/** END OF "Collective 1 fails to send money to Collective 3 because Collective 1\'s Host does not have a credit card" */
+
+    it('Collective 1 fails to send money to Collective 3 because Collective 3\'s Host does not have a credit card', async () => {
```
xdamman: that shouldn't fail. The only requirement on the host of the recipient is to have a connected Stripe account (which they should as a host , but good to double check)

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+      await models.PaymentMethod.create({
+        name: '4343',
+        service: 'stripe',
+        type: 'creditcard',
+        token: 'tok_123456781234567812345678',
+        CollectiveId: collective3.HostCollectiveId,
+        monthlyLimitPerMember: 10000
+      });
+
+      //finding opencollective payment method for Collective 1
```
xdamman: space after //
(would be good to add a eslint rule for it)

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+      }
+      // Executing queries
+      const res = await utils.graphqlQuery(createOrderQuery, { order }, user1);
+
+      // Then there should be Errors
+      expect(res.errors).to.exist;
+    });/** END OF "Collective 1 fails to send money to Collective 3 because Collective 3\'s Host does not have a credit card" */
+
+    it('Collective 1 sends money to Collective 3 through different hosts(no platform fees, but still has stripe and host fees)', async () => {
+      // Add User 1 as Collective 1 Admin
```
xdamman: Respect the case to be able to follow the variable you are referencing to:
`// Add user1 as an ADMIN of collective1`
(double check all your comments for this as I haven't pointed those discrepancies every time)

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+        type: 'creditcard',
+        token: 'tok_123456781234567812345678',
+        CollectiveId: collective3.HostCollectiveId,
+        monthlyLimitPerMember: 10000
+      });
+
+      //finding opencollective payment method for Collective 1
+      openCollectivePaymentMethod = await models.PaymentMethod.findOne({ where: {type: 'collective', CollectiveId: collective1.id}});
+
+      // Setting up order with amount less than the credit card monthly limit
```
xdamman: What happens when the amount is higher than that number? 
We should check for the error message in that case.
We should also make sure that if the error happens as part of the cron job (for recurring donations), the proper error message is sent by email (and cc us?)

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+          CollectiveId: orderCreatedCollective.id,
+          FromCollectiveId: orderCreatedFromCollective.id,
+          amount: order.totalAmount,
+        }
+      });
+      // make sure the transaction has been recorded
+      expect(transaction.FromCollectiveId).to.equal(collective1.id);
+      expect(transaction.CollectiveId).to.equal(collective3.id);
+      expect(transaction.currency).to.equal(collective1.currency);
+    });
```
xdamman: This is only testing the first transaction of the subscription. 
Can you also add a test (maybe in the subscription cron job tests?) to make sure that orders with an active subscription using the opencollective payment method are working across hosts?

marcogbarcellos:
@xdamman thanks a lot for the review. Really appreciate it and already started working on fixing it. One thing: I tried using Fixes #1196  and Closes #1196  to close the issue #1196 but it's not redirecting to the other repo, instead this link it's going to a previous [Pull Request](https://github.com/opencollective/opencollective-api/pull/1196) of this project(opencollective-api).

marcogbarcellos:
Thanks for the awesome review, working on it!

test/paymentmethods.collective.to.collective.test.js
znarf: https://eslint.org/docs/rules/spaced-comment

test/paymentmethods.collective.to.collective.test.js
```diff
@@ -0,0 +1,438 @@
...
+      }
+      totalAmount
+      currency
+      description
+    }
+  }
+`;
+
+
+describe("payments.collectiveToCollective.test.js", () => {
```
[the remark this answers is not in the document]
marcogbarcellos: yeah I was really confused because today there is no standard so i tried to follow at least the same standard as the other paymentMethods test files( example: `paymentmethods.paypal.payment.test.js`). anyways, i'll change here to camelCase at least in this case for now and will create an issue to standardize files test to camelCase. thanks.

test/paymentmethods.collective.to.collective.test.js
marcogbarcellos: `it('the available balance of the payment method of the collective should **be** equal **to the** balance of the collective', ... ` perhaps?

xdamman:
This also fixes https://github.com/opencollective/opencollective/issues/1211

server/paymentProviders/opencollective/collective.js
```diff
@@ -99,12 +99,13 @@ paymentMethodProvider.processOrder = async (order, options = {}) => {
       }
     });
     if (!fromCollectiveHostPaymentMethod) {
-      throw new Error(`Host ${fromCollectiveHost.name} needs to add a credit card to send money to different Hosts`);
+      throw new Error(`Host ${fromCollectiveHost.name} needs to add a credit card to send money`,
```
xdamman: unlike `console.log`, you can't append strings by adding more parameters to `new Error()`. Also, usually, when using ``, we concatenate everything in one string.

marcogbarcellos: minor thing, did this because the string was being break along with breaking the line and eslint rule(`[eslint] Unexpected string concatenation. (prefer-template)`) doesn't allow me to use quotes. anyways, will fix it. thanks.

marcogbarcellos: don't like it, but will do this(to keep less than 120 character columns per line):
https://stackoverflow.com/questions/38000659/template-strings-es6-prevent-line-breaks

xdamman:
yes, that works, but this doesn't: `throw new Error(`does it work? ${1 ===
1}`, `not this`)`;

marcogbarcellos:
@xdamman my bad, I reread what you've written and removed my comment that didn't make sense! thanks for the heads up!

PR #1407, opened 2018-08-08
Add User To Host W9 Received forms List

marcogbarcellos:
Today when we create an expense that exceeds the **W9 Bot Threshold**(600USD as of Today), we create a comment to warn the user that he needs to send the form to the Host and also add that user to the list of users that already received the request(`host.data.W9.requestSentToUserIds`). 

Through this PR we will consider that after the request is done, whenever the **Host** approves that expense(or any other after the warn) it means that the Host already received the form. So we add the user to the List of Users that already sent the form on the **Host** data(`host.data.W9.receivedFromUserIds`).

Extra: Refactored some Tests to look for the W9 Threshold in the database instead of hardcoding a 600USD value(today we have this value, but if we change it tomorrow we will change the database value).

Closes opencollective/opencollective#1200

### DIFF
--- a/server/graphql/mutations/expenses.js
+++ b/server/graphql/mutations/expenses.js
@@ -1,4 +1,4 @@
-import { get } from 'lodash';
+import { get, set } from 'lodash';
 import errors from '../../lib/errors';
 import roles from '../../constants/roles';
 import statuses from '../../constants/expense_status';
@@ -31,6 +31,24 @@ async function _createActivity(expense, type) {
       transaction: transaction.info
     }
   });
+}
+
+async function _addUserIdToHostW9ReceivedList(expense) {
+  const host = await expense.collective.getHostCollective();
+  // If user is already included in Host data List, don't do anything
+  if (get(host, 'data.W9.receivedFromUserIds') && host.data.W9.receivedFromUserIds.includes(expense.UserId)) {
+    return false;
+  }
+  // Only Inserts User in W9 Received list if he is already present on
+  // the W9.requestSentToUserIds (which means this user already overstepped the W9 Threshold)
+  if (get(host, 'data.W9.requestSentToUserIds') && host.data.W9.requestSentToUserIds.includes(expense.UserId)) {
+    const receivedFromUserIds = !get(host, 'data.W9.receivedFromUserIds')
+      ? [expense.UserId]
+      : get(host, 'data.W9.receivedFromUserIds').concat([expense.UserId]);
+    set(host, 'data.W9.receivedFromUserIds', receivedFromUserIds);
+    return host.update({data:host.data});
+  }
+  return false;
 }
 
 /**
@@ -97,6 +115,7 @@ export async function updateExpenseStatus(remoteUser, expenseId, status) {
   const res = await expense.update({ status, lastEditedById: remoteUser.id });
   if (status === statuses.APPROVED) {
     await _createActivity(expense, activities.COLLECTIVE_EXPENSE_APPROVED);
+    await _addUserIdToHostW9ReceivedList(expense);
   }
 
   return res;
--- a/test/w9.bot.test.js
+++ b/test/w9.bot.test.js
@@ -5,6 +5,7 @@
 /* Test libraries */
 import sinon from 'sinon';
 import { expect } from 'chai';
+import { get } from 'lodash';
 
 /* Test utilities */
 import * as utils from './utils';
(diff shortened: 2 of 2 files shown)


server/graphql/mutations/expenses.js
```diff
@@ -33,6 +33,24 @@ async function _createActivity(expense, type) {
...
+  // If user is already included in Host data List, don't do anything
+  if (get(host, 'data.W9.receivedFromUserIds') && host.data.W9.receivedFromUserIds.includes(expense.UserId)) {
+    return false;
+  }
+  // Only Inserts User in W9 Received list if he is already present on
+  // the W9.requestSentToUserIds (which means this user already overstepped the W9 Threshold)
+  if (get(host, 'data.W9.requestSentToUserIds') && host.data.W9.requestSentToUserIds.includes(expense.UserId)) {
+    const receivedFromUserIds = !get(host, 'data.W9.receivedFromUserIds')
+      ? [expense.UserId]
+      : get(host, 'data.W9.receivedFromUserIds').concat([expense.UserId]);
```
xdamman: why `concat` and not `push`?

server/graphql/mutations/expenses.js
```diff
@@ -97,6 +115,7 @@ export async function updateExpenseStatus(remoteUser, expenseId, status) {
   const res = await expense.update({ status, lastEditedById: remoteUser.id });
   if (status === statuses.APPROVED) {
     await _createActivity(expense, activities.COLLECTIVE_EXPENSE_APPROVED);
+    await _addUserIdToHostW9ReceivedList(expense);
```
xdamman: this doesn't work because any admin of the collective can approve the expense but they have no way to know if the host has received the W9.

Ideally, we would want the host to also approve expenses, but we don't support that at the moment.

So right now, the usual process is that one of the admins of the collective approves pending expenses then one of the admin of the host pays approved expenses. 

So I'd only run `_addUserIdToHostW9ReceivedList(expense);` when `status === statuses.PAID`

test/w9.bot.test.js
```diff
@@ -306,52 +346,220 @@ describe('w9.bot.test.js', () => {
...
-      expect(emailSendMessageSpy.secondCall.args[0]).to.equal(admin.email);
-      expect(emailSendMessageSpy.secondCall.args[1]).to.contain("New expense on");
-      expect(emailSendMessageSpy.secondCall.args[2]).to.contain("/test-collective/expenses/2/approve");
-      expect(emailSendMessageSpy.thirdCall.args[0]).to.equal(user.email);
-      expect(emailSendMessageSpy.thirdCall.args[1]).to.contain("New comment on your expense");
+      expect(emailSendMessageSpy.callCount).to.be.equal(3);
+
+    });/* End of "creates 2 new expenses that adds up more than the W9 Bot Threshold and create Comment expense forms email" */
+
+    it('Host Data must NOT include user in W9 Received List if he didn\'t spend more than W9 Threshold after Expense is Approved', async () => {
```
xdamman: PAID

test/w9.bot.test.js
```diff
@@ -306,52 +346,220 @@ describe('w9.bot.test.js', () => {
...
+      // Approved expense triggers one email as well
+      await utils.waitForCondition(() => emailSendMessageSpy.callCount > 1);
+      expect(emailSendMessageSpy.callCount).to.be.equal(2);
+
+      // And then the host data has to be null
+      const updatedHost = await models.Collective.findById(hostCollective.id);
+      expect(updatedHost.data).to.be.null;
+    });/* End of "Host Data must NOT include user in W9 Received List if he didn\'t spend more than W9 Threshold after Expense is Approved" */
+
+    it('After User Expense is approved and User Expenses exceed W9 threshold, Host Data must include user in W9 Received List', async () => {
```
xdamman: PAID

server/graphql/mutations/expenses.js
marcogbarcellos: Got it, thanks.

I would rather move both `_createActivity` and  `_addUserIdToHostW9ReceivedList ` to a hook on the expense model than creating one more `if` in this method. anyways, will do it.

server/graphql/mutations/expenses.js
marcogbarcellos: This way is a "one line" solution. using push would require me at least 2 lines:

```
const receivedFromUserIds = !get(host, 'data.W9.receivedFromUserIds')
      ? []
      : get(host, 'data.W9.receivedFromUserIds');
receivedFromUserIds.push(expense.UserId);
```

But it's not a big deal, can do this way as well.

PR #1408, opened 2018-08-10
Application model and authorization middleware

znarf:
# To test it

## Create a new App

You need to be authenticated! This will be usually done with the web UI. Because, we don't have the UI yet, we will do everything from the CLI.

### Use the existing API_KEY

`export API_KEY=dvl-1510egmf4a23d80342403fb599qd`

###  Get Login Token

`curl "http://0.0.0.0:3060/users/signin?api_key=${API_KEY}" -H "Content-Type: application/json" -X POST -d '{"user":{"email":"francois@opencollective.com"}}' `

`export LOGIN_TOKEN=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzY29wZSI6ImxvZ2luIiwiaWQiOjk0NzUsImVtYWlsIjoiZnJhbmNvaXNAaG9kaWVybmUubmV0IiwiaWF0IjoxNTMzODkzMzg2LCJleHAiOjE1MzM5Nzk3ODYsImlzcyI6Imh0dHA6Ly9sb2NhbGhvc3Q6MzA2MCIsInN1YiI6OTQ3NX0.grpRXdUUyZgnnItfOXAJIsyLuNF4CSLaSOvXqhQoDO0`

### Exchange Login Token with Access Token

`curl "http://0.0.0.0:3060/users/update-token?api_key=${API_KEY}" -H "Authorization: Bearer ${LOGIN_TOKEN}" -X POST`

### Create an App with the token

`export ACCESS_TOKEN=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzY29wZSI6ImxvZ2luIiwiaWQiOjk0NzUsImVtYWlsIjoiZnJhbmNvaXNAaG9kaWVybmUubmV0IiwiaWF0IjoxNTMzODkzMzg2LCJleHAiOjE1MzM5Nzk3ODYsImlzcyI6Imh0dHA6Ly9sb2NhbGhvc3Q6MzA2MCIsInN1YiI6OTQ3NX0.grpRXdUUyZgnnItfOXAJIsyLuNF4CSLaSOvXqhQoDO0`

`curl "http://0.0.0.0:3060/applications/create?api_key=${API_KEY}" -H "Content-Type: application/json" -H "Authorization: Bearer ${ACCESS_TOKEN}" -X POST -d '{ "name":"My Open Collective App", "description":"This is my Open Collective App", "callbackUrl":"http://0.0.0.0:3000/callback" }' `

### Note the Client Id and Client Secret

`export CLIENT_ID=52e309799aee23622a5fac8f67adc6d6e47b57f6`

### Request Sign In

`curl "http://0.0.0.0:3060/users/signin" -H "Client-Id: ${CLIENT_ID}" -H "Content-Type: application/json" -X POST -d '{"user":{"email":"francois@opencollective.com"}}'`

### Note the temporary login token

`export APP_LOGIN_TOKEN=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzY29wZSI6ImxvZ2luIiwiYXBwIjoxMCwiaWF0IjoxNTMzODk2MzYyLCJleHAiOjE1MzM5ODI3NjIsImlzcyI6Imh0dHA6Ly9sb2NhbGhvc3Q6MzA2MCIsInN1YiI6OTQ3NX0.RQbCneF9Mvx0-2dPSDBw532S9QCL0-Gt84jNpUdlteM`

### Exchange it with a long-lived access token

`curl "http://0.0.0.0:3060/users/token" -H "Client-Id: ${CLIENT_ID}" -H "Authorization: Bearer ${APP_LOGIN_TOKEN}" -X POST`

### Note the long-lived access token

`export APP_ACCESS_TOKEN=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpZCI6OTQ3NSwiZW1haWwiOiJmcmFuY29pc0Bob2RpZXJuZS5uZXQiLCJpYXQiOjE1MzM4OTY2OTYsImV4cCI6MTU0MTY3MjY5NiwiaXNzIjoiaHR0cDovL2xvY2FsaG9zdDozMDYwIiwic3ViIjo5NDc1fQ.BXWDeDywxyvWpMl-OTJh2vIon_PmtiRcQyD0TFzgY-o`

### Inspect the Token payload

```
{
  "app": 10, <- App Id
  "scope": "session", <- Token Type
  "iat": 1533897300, <- Created timestamp
  "exp": 1541673300, <- Expiration timestamp
  "sub": 9475  <- User Id
}
```


xdamman:
The looks good so far. I like that our own frontend would follow the exact same pattern.
We just need to make sure that the temporary login token is never returned by the API. It can only be sent by email to the user.

How is this different than oAuth? Are we taking risks here by not follow such standard?

It feels to me that the client id and secret could be merged into one `api_key` no? Why having two separate strings?

If possible, it would be better to follow our convention for naming attributes, e.g. "iat" => "createdAt", "app" => "AppId", etc.

It would be useful to see the end result in terms of developer experience.
What about having a different repo for developer.opencollective.com (related: https://github.com/opencollective/opencollective/issues/1055)
Maybe you can start one with the documentation the way it should be? (DDD -- Documentation Driven Development -- I just made it up ;-) )

The main things that people should be able to do (probably in that order):
See also: https://github.com/opencollective/opencollective/issues/464

- Make simple READ calls (rate limited without `api_key`)
- Define webhooks
- Create an app and an `api_key` (I'd make a collective for the app, like we did for the `W9Bot` so that we can make it easy to "add an app" to a collective by adding it as an ADMIN or MEMBER which would automatically give it permissions to do certain things in the collective)
- Connect to GraphQL with that `api_key`, make authenticated calls on behalf of a user with a token that is related to their `api_key`

znarf:
> How is this different than oAuth? Are we taking risks here by not follow such standard?

In that implementation, we're shortcuting oAuth by sending the login token in an authorization email directly.

I do think there is value in supporting traditional oAuth, but it's usually involving a login page with a password. To bring come clarity to the conversation, I will later summarize our options there.

> It feels to me that the client id and secret could be merged into one api_key no? Why having two separate strings?

It's currently mimicking the oAuth implementations that I'm aware of. It's effectively not required if we're not following the oAuth protocol.

> If possible, it would be better to follow our convention for naming attributes, e.g. "iat" => "createdAt", "app" => "AppId", etc.

Here, we're following the JSON Web Token existing semantics. Also, we're incentivized to keep the keys short because the longer they are, the longer the token will be.

znarf:
It's now ready, it's compatible with existing cases and existing tests.

The login by email is abandoned, we're skipping authentication for now, and we will likely be using oAuth for that.

PR #1409, opened 2018-08-14
Allow createOrder mutation to create a Pledge

HipsterBrown:
A Pledge is defined as a pending order for a collective that is inactive. If the collective is inactive, we don't charge the order.

### DIFF
--- a/server/graphql/mutations/orders.js
+++ b/server/graphql/mutations/orders.js
@@ -17,17 +17,14 @@ import { getNextChargeAndPeriodStartDates, getChargeRetryCount} from '../../lib/
 
 const debugOrder = debug('order');
 
-// TODO:
-// - refactor to use await
-// - add logic for creating a pending collective if a website is provided instead of an id
 export async function createOrder(order, loaders, remoteUser) {
   try {
     if (order.paymentMethod && order.paymentMethod.service === 'stripe' && order.paymentMethod.uuid && !remoteUser) {
       throw new Error("You need to be logged in to be able to use a payment method on file");
     }
 
-    if (!order.collective || !order.collective.id) {
-      throw new Error("No collective id provided");
+    if (!order.collective || (!order.collective.id && !order.collective.website)) {
+      throw new Error("No collective id or website provided");
     }
 
     if (order.platformFeePercent && !remoteUser.isRoot()) {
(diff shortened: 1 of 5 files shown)


server/graphql/mutations/orders.js
```diff
@@ -14,240 +14,230 @@ import status from '../../constants/order_status';
...
+    } else if (order.collective.website) {
+      collective = (await models.Collective.findOrCreate({ where: {website: order.collective.website}, defaults: order.collective }))[0];
+      collective.tiers = defaultTiers(collective.HostCollectiveId, collective.currency);
     }
-    order.collective = collective = c;
 
-    if (!collective.isActive) {
-      throw new Error(`This collective is not active`);
+    if (!collective) {
+      throw new Error(`No collective found: ${order.collective.id || order.collective.website}`);
```
xdamman: Not sure about this error message. Since you do a `findOrCreate`, the error should be "can't create a collective" no?

test/graphql.createOrder.test.js
```diff
@@ -381,8 +392,10 @@ describe('createOrder', () => {
       expect(res.errors).to.not.exist;
       // Then the created transaction should match the requested data
       const orderCreated = res.data.createOrder;
+      expect(orderCreated.status).to.equal('ACTIVE');
```
xdamman: can you also add tests for the other possible statuses (CANCELED, REJECTED, ERROR)?

server/graphql/mutations/orders.js
```diff
@@ -14,240 +14,230 @@ import status from '../../constants/order_status';
...
-  .then(c => {
-    if (!c) {
-      throw new Error(`No collective found with id: ${order.collective.id}`);
+    // Check the existence of the recipient Collective
+    let collective;
+    if (order.collective.id) {
+      collective = await loaders.collective.findById.load(order.collective.id);
+    } else if (order.collective.website) {
+      collective = (await models.Collective.findOrCreate({ where: {website: order.collective.website}, defaults: order.collective }))[0];
+      collective.tiers = defaultTiers(collective.HostCollectiveId, collective.currency);
```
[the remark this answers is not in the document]
HipsterBrown: Got it. I'll remove that part.

server/graphql/mutations/orders.js
HipsterBrown: If we get to this point in the code, it's because no collective was found if `order.collective.id` is passed to the mutation. 

You are correct that it will default to creating a new collective if `order.collective.website` is passed to the mutation and not found in the database.

PR #1413, opened 2018-08-20
refactor to not rely on PaymentMethod.customerId 

xdamman:
...while still making sure the logged in user has enough permissions to use the payment method

fixes https://github.com/opencollective/opencollective/issues/1229


xdamman:
@marcogbarcellos I realized that moving the check to `PaymentMethod.canBeUsedForOrder` won't work because `this.CollectiveId` might be `null` within the `PaymentMethod` model if the user decided not to save the credit card (i.e. not attach it to their user collective).

xdamman:
I found a way to fix this by making sure that if there is no `PaymentMethod.CollectiveId`, the `remoteUser` needs to be the creator of the `PaymentMethod`.

PR #1426, opened 2018-08-30
Enable Organizations/Hosts to have multiple prepaid card Payment method

marcogbarcellos:
Today, we are always adding funds(or if it's the first time create a prepaid card to then add those funds) to one prepaid card. It makes sense to have multiple prepaid cards and this PR enables it.

Closes opencollective/opencollective#1252

### DIFF
--- a/server/graphql/mutations.js
+++ b/server/graphql/mutations.js
@@ -320,6 +320,7 @@ const mutations = {
       CollectiveId: { type: new GraphQLNonNull(GraphQLInt) },
       HostCollectiveId: { type: new GraphQLNonNull(GraphQLInt) },
       description: { type: GraphQLString },
+      PaymentMethodId: { type: GraphQLInt },
     },
     resolve: async (_, args, req) => addFundsToOrg(args, req.remoteUser),
   }
--- a/server/graphql/mutations/orders.js
+++ b/server/graphql/mutations/orders.js
@@ -455,30 +455,36 @@ export async function addFundsToOrg(args, remoteUser) {
     models.Collective.findById(args.CollectiveId),
     models.Collective.findById(args.HostCollectiveId)
   ]);
-  let paymentMethod = await models.PaymentMethod.findOne({ where: {
+  // if a PaymentMethodId is defined it means we are going to try to
+  // increase the balance of an existing Payment method
+  if (args.PaymentMethodId) {
+    const paymentMethod = await models.PaymentMethod.findOne({ where: {
+      id: args.PaymentMethodId,
+    } });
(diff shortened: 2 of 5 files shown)


marcogbarcellos:
PS: This PR must not be merged until there's a new implementation on the frontend to support this feature: When the user selects the hosts, the modal should show a Picker with the existing prepaid payment methods(and also a `create new` option) so the user can specify whether it should use an existing one or not.

xdamman:
shouldn't we say that we can't increase the balance of a card? And therefore we should always create a new one?
If that's the case then the frontend can remain as is.

server/graphql/mutations/orders.js
```diff
@@ -455,30 +455,36 @@ export async function addFundsToOrg(args, remoteUser) {
...
+        initialBalance: paymentMethod.initialBalance + args.totalAmount,
+      });
+      return paymentMethod;
+    }
+  }
+  // otherwise, creates a new Payment method
+  const paymentMethod = await models.PaymentMethod.create({
+    name: args.description || 'Host funds',
+    initialBalance: args.totalAmount,
+    monthlyLimitPerMember: args.totalAmount,
```
xdamman: no reason to define the monthlyLimitPerMember to this value. I'd keep it to null.
(that's what created the issue for Pia if I understand correctly).

server/graphql/mutations.js
```diff
@@ -320,6 +320,7 @@ const mutations = {
       CollectiveId: { type: new GraphQLNonNull(GraphQLInt) },
       HostCollectiveId: { type: new GraphQLNonNull(GraphQLInt) },
       description: { type: GraphQLString },
+      PaymentMethodId: { type: GraphQLInt },
```
xdamman: No need for this if we decide that we don't allow to increase the balance of a prepaid card. 
I think this would simplify things.

server/graphql/mutations/orders.js
```diff
@@ -455,30 +455,36 @@ export async function addFundsToOrg(args, remoteUser) {
     models.Collective.findById(args.CollectiveId),
     models.Collective.findById(args.HostCollectiveId)
   ]);
-  let paymentMethod = await models.PaymentMethod.findOne({ where: {
+  // if a PaymentMethodId is defined it means we are going to try to
+  // increase the balance of an existing Payment method
```
xdamman: same comment as above.

marcogbarcellos:
Sounds good then @xdamman , in this case we wouldn't need anything on the frontend. here it is!! :)

PR #1429, opened 2018-08-31
Isolate background jobs, only run with the server

znarf:
That's a fix for 
https://github.com/opencollective/opencollective/issues/1254

### DIFF
--- a/server/background-jobs.js
+++ b/server/background-jobs.js
@@ -1,0 +1,31 @@
+import Promise from 'bluebird';
+
+import queries from './lib/queries';
+
+// const useCache = ['production', 'staging'].includes(process.env.NODE_ENV);
+const useCache = true;
+const CACHE_REFRESH_INTERVAL = process.env.CACHE_REFRESH_INTERVAL || 1000 * 60 * 60;
+
+// warming up the cache with the homepage queries
+const cacheEntries = [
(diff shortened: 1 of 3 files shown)


server/background-jobs.js
```diff
@@ -0,0 +1,31 @@
+import Promise from 'bluebird';
+
+import queries from './lib/queries';
+
+// const useCache = ['production', 'staging'].includes(process.env.NODE_ENV);
```
xdamman: comment it  out?

xdamman:
I don't see the link with that issue.
You are basically just moving the process to update the cache for the complex queries for the homepage into its own background process file. How is that related to `contributors.svg`? What am I missing?

xdamman:
Ok I understand better with your comment on slack:

> According to my inbox, the daily cronjob is not completing since July 17th. The last day we received the ‘Daily ledger health report’
> But “Daily Subscription Report” is still there, so there is a problem in between.
> Ok, what happened on the 17th of July? We got a deploy https://github.com/opencollective/opencollective-api/compare/19c0fb50...d05e2954 The highlight is:
> ```// refresh every hour
>   setInterval(refreshCache, 1000 * 60 * 60);
> ```
> -> the setInterval is preventing the script to finish

server/background-jobs.js
znarf: Should be un-commented, yes. I will fix it.

xdamman:
I've removed the comment and I'll merge and deploy to make sure that the cron job doesn't fail tomorrow morning for the monthly reports.

PR #1436, opened 2018-09-03
README local set-up instructions

bartvanremortele:
I have updated the README to include local set-up instructions. 

I have also attempted to make db_restore work when running it for the first time. I'm not sure what the impact of this is but it solves 2 issues:
- script failing when the database has not been created before
- createdb trying to use a role which hasn't been created yet

`createdb $LOCALDBNAME 2> /dev/null` the pipe to /dev/null also makes this silently fail

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,37 @@ $ docker-compose -f docker/docker-compose.yml up --build
 Then you'll be able to access the UI from http://localhost:13000 and
 the API from the address http://localhost:13060.
 
+### Local setup
+
+#### Requirements
+- A PostgreSQL installation (10.3, 9.6.8, 9.5.12, 9.4.17, 9.3.22 or newer)
+
+#### Steps
+
+Make sure you have python 2 as your active version. You can use `pyenv` to switch python versions.
+
+Copy `env.default` and rename it to `.env`
+ 
+Install the dependencies by running: 
+
+```
+npm install
+```
+
+If you're running into node-gyp issues related to Python 3 vs Python 2 you can run `npm rebuild` or run `npm install` again.
+
+Setup your database and restore the seed dump by running:
+
+```
+npm run db:setup && ./node_modules/.bin/babel-node ./scripts/db_restore.js opencollective_dvl
+```
+
+Start the API by running:
+
+```
+npm run dev
+```
+
 ### Once it's running
 
 This new environment is created with the `opencollective_dvl`
--- a/README.md
+++ b/README.md
@@ -67,13 +67,9 @@ Install the dependencies by running:
 npm install
 ```
 
-If you're running into node-gyp issues related to Python 3 vs Python 2 you can run `npm rebuild` or run `npm install` again.
-
-Setup your database and restore the seed dump by running:
-
-```
-npm run db:setup && ./node_modules/.bin/babel-node ./scripts/db_restore.js opencollective_dvl
-```
+Troubleshooting:
+- If you're running into node-gyp issues related to Python 3 vs Python 2 you can run `npm rebuild` or run `npm install` again.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -50,6 +50,37 @@ $ docker-compose -f docker/docker-compose.yml up --build
 Then you'll be able to access the UI from http://localhost:13000 and
 the API from the address http://localhost:13060.
 
+### Local setup
+
+#### Requirements
+- A PostgreSQL installation (10.3, 9.6.8, 9.5.12, 9.4.17, 9.3.22 or newer)
```
znarf: Have you checked it's working with 10.3 ? Last time I checked it was only working with 9.x

README.md
```diff
@@ -50,6 +50,37 @@ $ docker-compose -f docker/docker-compose.yml up --build
...
+
+Copy `env.default` and rename it to `.env`
+ 
+Install the dependencies by running: 
+
+```
+npm install
+```
+
+If you're running into node-gyp issues related to Python 3 vs Python 2 you can run `npm rebuild` or run `npm install` again.
```
znarf: Do you remember for which packages we need Python? It might be something we don't need anymore.

README.md
```diff
@@ -50,6 +50,37 @@ $ docker-compose -f docker/docker-compose.yml up --build
...
+### Local setup
+
+#### Requirements
+- A PostgreSQL installation (10.3, 9.6.8, 9.5.12, 9.4.17, 9.3.22 or newer)
+
+#### Steps
+
+Make sure you have python 2 as your active version. You can use `pyenv` to switch python versions.
+
+Copy `env.default` and rename it to `.env`
```
znarf: `npm install` is triggering `./scripts/postinstall.sh` which is already doing `cp .env.default .env`

README.md
```diff
@@ -50,6 +50,37 @@ $ docker-compose -f docker/docker-compose.yml up --build
...
+```
+npm install
+```
+
+If you're running into node-gyp issues related to Python 3 vs Python 2 you can run `npm rebuild` or run `npm install` again.
+
+Setup your database and restore the seed dump by running:
+
+```
+npm run db:setup && ./node_modules/.bin/babel-node ./scripts/db_restore.js opencollective_dvl
```
znarf: This should also be triggered in `postinstall`, see `./scripts/db_restore.sh -d opencollective_dvl -f test/dbdumps/opencollective_dvl.pgsql`

README.md
bartvanremortele: Python is required to use node-gyp (https://github.com/nodejs/node-gyp) which is a requirement to build newrelic's native-metrics (https://github.com/newrelic/node-native-metrics#readme)

README.md
bartvanremortele: I'll verify to be sure

README.md
bartvanremortele: It seems like something went wrong for me then. I didn't have the .env file and the database setup after npm installing. I'll try to reproduce.

README.md
znarf: So, let's just say 9.x, 10.x or newer ?

Maybe add Python there as a requirement?

bartvanremortele: I explicitly added these versions because the seed db can't be loaded with lower versions than these because it was dumped with a version that had this security patch included. Lower PG versions won't be able to load the dump.
I'll add Python 2.7 as a requirement.

znarf: Ok, thank you, get it.

znarf:
@bartvanremortele Thank you for your contribution. Merging now.

I think we need to follow up on that because the steps on setting up the database are still potentially problematic. We definitely need to simplify it. And also enhance the documentation, we already have that document that we can iterate on https://github.com/opencollective/opencollective-api/blob/master/docs/postgres.md

PR #1447, opened 2018-09-16
Fix/1300 virtual card

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/1300


server/graphql/mutations/paymentMethods.js
```diff
@@ -54,18 +54,23 @@ async function createVirtualPaymentMethod(args, remoteUser) {
...
   const emitter = await models.Collective.findById(paymentMethod.sourcePaymentMethod.CollectiveId);
   const userCollective = await user.getCollective();
-  const name = userCollective.name && userCollective.name.match(/anonymous/i) ? '' : user.name;
+  const userName = userCollective.name && userCollective.name.match(/anonymous/i) ? '' : userCollective.name;
 
-  emailLib.send('user.card.claimed', args.email, {
-    loginLink: user.generateLoginLink(`/redeemed?name=${name}&amount=${amount}&currency=${currency}&emitterSlug=${emitter.slug}&emitterName=${emitter.name}`),
-    amount,
+  emailLib.send('user.card.claimed', user.email, {
+    loginLink: user.generateLoginLink(`/redeemed?name=${encodeURIComponent(userName)}&amount=${initialBalance}&currency=${currency}&emitterSlug=${emitter.slug}&emitterName=${encodeURIComponent(emitter.name)}`),
```
znarf: There are some utilities creating a queryString from an object, way to go there.

znarf: I'm still concerned about passing display values in the URL, let's pass an id and load the data from API.

server/graphql/mutations.js
```diff
@@ -379,13 +380,13 @@ const mutations = {
     },
     resolve: async (_, args, req) => createPaymentMethod(args, req.remoteUser),
   },
-  claimVirtualCard: {
+  claimPaymentMethod: {
```
marcogbarcellos: Does it make sense creating a "more generic" name in a case where only virtual cards can be claimed today? I know we can have other kinds of payment methods later on that can be claimed, but this may sound confusing if we consider the other payment methods that cannot be claimed. not a big deal though.

server/graphql/mutations/paymentMethods.js
marcogbarcellos: I kind of agree with you, having 100% static page makes it really fast but I don't see this approach best practice. It's not that bad as well as the email goes straight to that user and the data is no that sensitive, but at the very least we should avoid doing something like this even if we won't have a complete static page.

server/graphql/mutations.js
xdamman: I see what you mean but I'd rather limit the number of those graphql mutations and therefore keep them as generic as possible. Less surface to test / maintain.

PR #1448, opened 2018-09-16
Allow for manual donations

xdamman:
This PR enables orders to specify a "manual" payment method, in which case the order's status will be PENDING and will have to be manually confirmed as PAID.

Related issue: https://github.com/opencollective/opencollective/issues/1281

TODO:
- [x] New manual payment provider
- [x] Set default order status to PENDING
- [x] Email notification for pending orders

For another PR to avoid growing this PR even more:
- Add a mutation to mark a donation as PAID

### DIFF
--- a/server/models/PaymentMethod.js
+++ b/server/models/PaymentMethod.js
@@ -337,7 +337,7 @@ export default function(Sequelize, DataTypes) {
         this.type !== 'manual'
       ) {
         throw new Error(
-          `You don't have enough permissions to use this payment method (you need to be an admin of the collective that owns this payment method)`,
+          "You don't have enough permissions to use this payment method (you need to be an admin of the collective that owns this payment method)",
         );
       }
     }
--- a/server/models/PaymentMethod.js
+++ b/server/models/PaymentMethod.js
@@ -253,33 +253,56 @@ export default function(Sequelize, DataTypes) {
     }
 
     if (this.limitedToTags) {
-      const collective = order.collective || await order.getCollective();
+      const collective = order.collective || (await order.getCollective());
       if (intersection(collective.tags, this.limitedToTags).length === 0) {
-        throw new Error(`This payment method can only be used for collectives in ${formatArrayToString(this.limitedToTags)}`);
+        throw new Error(
+          `This payment method can only be used for collectives in ${formatArrayToString(
+            this.limitedToTags,
+          )}`,
+        );
       }
     }
 
     // quick helper to get the name of a collective given its id to format better error messages
-    const fetchCollectiveName = (CollectiveId) => {
-      return CollectiveId && models.Collective.findOne({
-        attributes: ['name'],
-        where: { id: CollectiveId },
-      }).then(r => r && r.name);
+    const fetchCollectiveName = CollectiveId => {
+      return (
+        CollectiveId &&
+        models.Collective.findOne({
+          attributes: ['name'],
+          where: { id: CollectiveId },
+        }).then(r => r && r.name)
+      );
     };
 
     if (this.limitedToCollectiveIds) {
-      const collective = order.collective || await order.getCollective();
+      const collective = order.collective || (await order.getCollective());
       if (!this.limitedToCollectiveIds.includes(collective.HostCollectiveId)) {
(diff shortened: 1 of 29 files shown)


marcogbarcellos:
any status here?

test/utils.js
```diff
@@ -172,6 +173,19 @@ export const readFee = (amount, feeStr) => {
   }
 };
 
+export const separator = length => {
+  const terminalCols =
+    process.platform === 'win32'
+      ? length || 40
+      : length || parseInt(execSync(`tput cols`).toString());
```
Betree: Is this `separator` function used somewhere?

A user recently had a problem with `tput` not installed on an alpine linux (see https://github.com/opencollective/opencollective/issues/1443). Maybe we should encapsulate it in a separate `getTerminalCols` which would returns the default `40` if the call fails.

Betree: Could happen on Windows machine (but this is tested here) or on Alpine linux if we ever run our tests in a such container (for example in CI). We could just put that in a `try..catch` block and return `40` if we catch.

xdamman: Good idea. Done in this commit: https://github.com/opencollective/opencollective-api/pull/1448/commits/329526f32e1fee46de7081c103572ea06052065f

server/models/PaymentMethod.js
```diff
@@ -308,9 +331,13 @@ export default function(Sequelize, DataTypes) {
...
       // If there is no monthly limit, the user needs to be an admin of the collective that owns the payment method
-      if (!this.monthlyLimitPerMember && !user.isAdmin(this.CollectiveId)) {
+      if (
+        !this.monthlyLimitPerMember &&
+        !user.isAdmin(this.CollectiveId) &&
+        this.type !== 'manual'
+      ) {
         throw new Error(
-          "You don't have enough permissions to use this payment method (you need to be an admin of the collective that owns this payment method)",
+          `You don't have enough permissions to use this payment method (you need to be an admin of the collective that owns this payment method)`,
```
marcogbarcellos: we better use single quotes when there's no variable to be inserted in the string template.

server/paymentProviders/opencollective/manual.js
```diff
@@ -0,0 +1,65 @@
...
+    netAmountInCollectiveCurrency: order.totalAmount - hostFeeInHostCurrency - platformFeeInHostCurrency,
+    amountInHostCurrency: order.totalAmount,
+    hostFeeInHostCurrency,
+    platformFeeInHostCurrency,
+    paymentProcessorFeeInHostCurrency,
+    description: order.description,
+  };
+
+  const creditTransaction = await models.Transaction.createFromPayload(payload);
+  console.log(">> creditTransaction", creditTransaction.dataValues);
```
marcogbarcellos: no need for console log here, right?

test/utils.js
```diff
@@ -172,6 +173,28 @@ export const readFee = (amount, feeStr) => {
...
+};
+
+export const separator = length => {
+  const terminalCols = length || getTerminalCols();
+
+  let separator = '';
+  for (let i = 0; i < terminalCols; i++) {
+    separator += '-';
+  }
+  console.log(`\n${separator}\n`);
```
marcogbarcellos: perhaps remove this console log as well...

marcogbarcellos:
other than the minor things on my comments, looks really good to me!

server/paymentProviders/opencollective/manual.js
xdamman: right (that's 2 month old but I'll take the opportunity to remove this)

test/utils.js
xdamman: well no, that's the whole point of this helper function: make it easier to debug in the console by spacing out and separating sections.

server/models/PaymentMethod.js
xdamman: good catch. I'll use double quotes here because of the `don't`.

test/utils.js
marcogbarcellos: I see, makes sense! missed that

znarf:
@xdamman for the "algorithm", I'm afraid the only thing that make sense is discussions between nice humans :-)

xdamman:
@znarf ok, so based on this conversation, I'm merging this :-)

PR #1470, opened 2018-10-05
Running tests with proper database

marcogbarcellos:
Using NODE_ENV=test will make it use default config to point to the test database instead of pointing to the development database as it is doing today.

fixes https://github.com/opencollective/opencollective/issues/1330

### DIFF
--- a/test/setup.js
+++ b/test/setup.js
@@ -2,13 +2,15 @@ import chai from 'chai';
 import chaiJestSnapshot from 'chai-jest-snapshot';
 import chaiAsPromised from 'chai-as-promised';
 
+// setting up NODE_ENV to test when running the tests.
+if (!process.env.NODE_ENV) {
+  process.env.NODE_ENV = 'test';
+}
+
 chai.use(chaiAsPromised);
 chai.use(chaiJestSnapshot);
 
 before(() => {
-  if (!process.env.NODE_ENV) {
-    process.env.NODE_ENV = 'test';
-  }
   chaiJestSnapshot.resetSnapshotRegistry();
 });
 


server/env.js
```diff
@@ -6,8 +6,7 @@ if (!process.env.NODE_ENV) {
 
 if (['production', 'circleci'].indexOf(process.env.NODE_ENV) === -1) {
   const dotenv = require('dotenv');
-  dotenv.config(); // this loads .env with real values. It needs to be first because dotenv doesn't overwrite any values
-  dotenv.config({ path: 'default.env' }); // this loads the default file in github with dummy values
```
znarf: If we remove 'default.env' should we delete the file also?

server/env.js
```diff
@@ -6,8 +6,7 @@ if (!process.env.NODE_ENV) {
...
+  dotenv.config({ path: `.env.${process.env.NODE_ENV}` }); // this loads the default file in github with dummy values
```
znarf: Does the comment `// this loads the default file in github with dummy values` still stand?

znarf:
What about https://github.com/opencollective/opencollective-api/blob/master/config/test.json Isn't that used?

server/env.js
marcogbarcellos: i replaced the `default.env` for either `.env.development` or `.env.test` to make it more generic.

marcogbarcellos: Ah, I understood your question now @znarf , well as far as I understood this comment still makes sense. I understood it's just saying it's going to load the `.env` file in the root of the project(that's what I think they meant with *github*).

marcogbarcellos:
I think we can remove this `test.json` file @znarf . at least afaik, it's not being used.

Looking at the code, they were loading the `./test/setup.js` and then setting the `process.env.NODE_ENV` on a mocha *before* function which is not the proper way to set the process.env 'cause it'd be too late.

So I believe we can remove this test.json. There may be another way to use it but at the very least is not through the script `npm run test`.

server/env.js
znarf: @marcogbarcellos 

- [ ] delete the 'default.env' if it's not used
- [ ] the dummy comment was valid for `default.env`, it's not anymore with the new behavior
- [ ] seems we're not loading `.env` anymore, that's a concern

znarf:
@marcogbarcellos `test.json` is the default and is used to define environment variables when they are missing. It's likely what is used on `circleci`.

server/env.js
marcogbarcellos: thanks for pointing that out @znarf . you're right, I got confused about both files but I believe they're meant to run either the project or tests locally and, in my opinion I can just merge the `default.env` file content into `.env.development`.We can see this looking the [README](https://github.com/opencollective/opencollective-api#steps) steps to run the project locally. So I'll create another PR to update the README and remove the `default.env`, This PR will have the following(if you agree with them): 
- [ ] delete the 'default.env' file
- [ ] remove the dummy comment that was previously valid for `default.env`
- [ ] Update the README with new instructions 

regarding your following concern: 
- `removing `dotenv.config();`, seems we're not loading `.env` anymore, that's a concern` - This would be necessary if we were to follow the README process of renaming the `default.env` to `.env` locally. As we will just use the `.env.development` or `.env.test`locally we won't need to load 2 dotenv files.

Can I go ahead and create another PR with the fixes? once more, thanks for pointing it out.

marcogbarcellos:
Hum I see @znarf , I still don't know much about the `circleci` config. thanks, will try to check that out.

server/env.js
znarf: @marcogbarcellos Ok, you can proceed, we will later see if we can improve further.

PR #1477, opened 2018-10-10
Fix permissions for deleting and editing updates

Betree:
Permission check on `deleteUpdate` was calling `remoteUser.isAdmin` with an update ID as param. However `isAdmin` takes a CollectiveID as parameter, not an update.id:
```es6
  User.prototype.isAdmin = function(CollectiveId) {
    const result =
      this.CollectiveId === Number(CollectiveId) ||
      this.hasRole([roles.HOST, roles.ADMIN], CollectiveId);
    debug('isAdmin of CollectiveId', CollectiveId, '?', result);
    return result;
  };
```

Test `deletes an update` was successful because the ID we were giving was matching an existing collective for which user had the authorization.

### DIFF
--- a/server/models/Update.js
+++ b/server/models/Update.js
@@ -6,7 +6,7 @@ import _, { pick } from 'lodash';
 import Temporal from 'sequelize-temporal';
 import slugify from 'slug';
 import activities from '../constants/activities';
-import { mustBeLoggedInTo, mustHaveRole } from '../lib/auth';
+import { mustHaveRole } from '../lib/auth';
 import Promise from 'bluebird';
 import showdown from 'showdown';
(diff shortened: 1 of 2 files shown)


Betree:
New check has been copied from `editUpdate` permissions.

 However I see a **security issue** here: if a contributor creates an update then gets removed from the collective, he will still be able to edit and delete updates.

I think we should change the permissions check on both `editUpdate` and `deleteUpdate` to only check if user has permissions to edit this collective.

znarf:
@Betree I agree, who created the update should not matter. +1 for a change.

Betree:
@znarf I've amended my commit to limit updates edition and deletion to collective admins only.

PR #1480, opened 2018-10-13
Handle edge cases in collective slug generation

Betree:
Fix https://github.com/opencollective/opencollective/issues/1359
Fix https://github.com/opencollective/opencollective/issues/1360

This PR implements three changes:

1. A blacklist of slugs for collective to avoid conflicting routes (see https://github.com/opencollective/opencollective/issues/1359)
2. If username cannot be slugified (ex: firstName = `"..."`) then collective is created with `user` slug.
3. Use [limax](https://github.com/lovell/limax) to generate slugs. This library handle Chinese characters and is actively maintained.

### DIFF
--- a/test/user.model.test.js
+++ b/test/user.model.test.js
@@ -102,6 +102,18 @@ describe('user.models.test.js', () => {
         expect(user.collective.slug.startsWith('user')).to.equal(true);
       });
     });
+
+    it('knows how to deal with special characters', () => {
+      return User.createUserWithCollective({
+        email: randEmail('user@domain.com'),
+        firstName: '很棒的用户',
+        lastName: 'awesome',
+      }).then(user => {
+        expect(user.collective.slug).to.equal(
+          'hen3-bang4-de-yong4-hu4-awesome',
+        );
+      });
+    });
   });
 
   /**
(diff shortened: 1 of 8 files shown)


test/user.model.test.js
```diff
@@ -84,6 +85,37 @@ describe('user.models.test.js', () => {
...
+    });
+
+    it('knows how to deal with special characters', () => {
+      return User.createUserWithCollective({
+        email: randEmail('user@domain.com'),
+        firstName: '很棒的用户',
+        lastName: 'awesome',
+      }).then(user => {
+        expect(user.collective.slug).to.equal(
+          'hen3-bang4-de-yong4-hu4-awesome',
```
znarf: why do we have numbers in there?

Betree: Apparently it is the standard way to represent these (mandarin) characters with Latin characters. Example: https://en.wiktionary.org/wiki/hen3

marcogbarcellos:
I'd ask the same question @znarf did, but given your answer. everything looks good to me as well. Also, limax looks more useful! thanks once more!!

znarf:
I pushed the branch here to run tests on Circle CI.

PR #1493, opened 2018-11-02
Fix typo in complete pledge email

sloria:


HipsterBrown:
@sloria Good catch. Thanks for creating a pull request. :clap:

PR #1499, opened 2018-11-05
Package security fixes

HipsterBrown:
This PR addresses most the security issues shown by GitHub and `npm audit`, not including the ones related to the `maildev` package which is only used for local development.


znarf:
@HipsterBrown Let's move that forward. What's breaking the tests?

HipsterBrown:
@znarf Looking into those failures today so we can merge this.

package.json
```diff
@@ -145,9 +145,9 @@
   "scripts": {
     "start": "node ./dist/index.js",
     "doc:gen": "jsdoc --configure .jsdoc.json --verbose",
-    "predocker:dev": "npm install && npm run db:setup && babel-node ./scripts/db_restore.js opencollective_dvl",
+    "predocker:dev": "npm install && npm run db:setup && babel-node --presets es2015,stage-0 --plugins add-module-exports,lodash ./scripts/db_restore.js opencollective_dvl",
```
znarf: Is there a way to refactor this `babel-node . --presets es2015,stage-0 --plugins add-module-exports,lodash` so that we just have these parameters in one place?

HipsterBrown: I'll double check and see if that's possible

HipsterBrown: Good to merge this then?

znarf: @HipsterBrown Have you created the issue?

HipsterBrown: I'm working on the update today

PR #1500, opened 2018-11-05
Limit user orders on 5 orders per hour

marcogbarcellos:
Temporary solution:

Block Order If the user tries to donate to the specific collective more than twice an hour.
Block Order If the user tries to make more than 10 donations an hour.

The permanent solution would be using cache and also check Subscriptions(mainly on 1st of the month).

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -156,21 +156,34 @@ export async function createOrder(order, loaders, remoteUser) {
         );
       }
 
-      // if there are more than 5 transactions in the last hour, blocks new orders
-      const end = new Date();
-      const start = new Date();
-      start.setHours(start.getHours() - 1);
-      const where = {
+      const nowLessOneHour = new Date();
+      nowLessOneHour.setHours(nowLessOneHour.getHours() - 1);
+      // Checking hourly limit of donations per fromCollective
+      const fromCollectiveOrdersCountLastHourQuery = {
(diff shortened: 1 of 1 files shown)


server/graphql/v1/mutations/orders.js
```diff
@@ -156,6 +156,22 @@ export async function createOrder(order, loaders, remoteUser) {
         );
       }
 
+      // if there are more than 5 transactions in the last hour, blocks new orders
+      const end = new Date();
+      const start = new Date();
+      start.setHours(start.getHours() - 1);
+      const where = {
+        FromCollectiveId: order.fromCollective.id,
+        createdAt: { [Op.between]: [start, end] },
```
xdamman: just greater than an hour ago

server/graphql/v1/mutations/orders.js
```diff
@@ -156,6 +156,22 @@ export async function createOrder(order, loaders, remoteUser) {
...
+        type: 'CREDIT',
+      };
+      const lastHourTransactionsCount = await models.Transaction.count({ where });
+      if ( lastHourTransactionsCount >= 5 ) {
```
xdamman: >=3

No reason to have more than 3 transactions from and to the same collective.

xdamman: ah I didn't realize it was not to the same collective.
Maybe add `CollectiveId` in the `where`?

PR #1503, opened 2018-11-06
Using cache to limit orders.

marcogbarcellos:


server/graphql/v1/mutations/orders.js
```diff
@@ -17,6 +17,18 @@ import {
...
+// Create Orders Cache Object to limit FromCollective orders to up to 10 orders/hour
+const ordersFromCollectiveCache = LRU({
+  max: 1000,
+  maxAge: 1000 * 60 * 60, // we keep it max 1 hour
+});
+// Create Orders Cache Object to limit pair (FromCollective,Collective) orders to up to 2 orders/hour
+const ordersFromCollectiveAndCollectiveCache = LRU({
+  max: 1000,
+  maxAge: 1000 * 60 * 60, // we keep it max 1 hour
+});
```
znarf: Why do we need to instanciate 2 cache instances?

marcogbarcellos: you're right! we don't. fixing it.

server/graphql/v1/mutations/orders.js
```diff
@@ -68,6 +80,30 @@ export async function createOrder(order, loaders, remoteUser) {
       );
     }
 
+    // Check if FromCollective reached max limit of orders per hour
+    if (ordersFromCollectiveCache.get(order.fromCollective.id)) {
+      if (ordersFromCollectiveCache.get(order.fromCollective.id) > 10) {
+        debug('Orders Cache: You have reached the hourly limit of donations', order);
```
znarf: Shouldn't that be debugOrder?

When will that appear in our logs?

server/graphql/v1/mutations/orders.js
```diff
@@ -68,6 +80,30 @@ export async function createOrder(order, loaders, remoteUser) {
       );
     }
 
+    // Check if FromCollective reached max limit of orders per hour
+    if (ordersFromCollectiveCache.get(order.fromCollective.id)) {
```
znarf: Why not get the value once and use it after?

server/graphql/v1/mutations/orders.js
```diff
@@ -68,6 +80,30 @@ export async function createOrder(order, loaders, remoteUser) {
...
+    } else {
+      ordersFromCollectiveCache.set(order.fromCollective.id, 1);
+    }
+    // Check if pair (FromCollective, Collective) reached max limit of orders per hour
+    if (ordersFromCollectiveAndCollectiveCache.get(`${order.fromCollective.id}_${order.collective.id}`)) {
+      if (ordersFromCollectiveAndCollectiveCache.get(`${order.fromCollective.id}_${order.collective.id}`) > 2) {
+        debug('Orders Cache: You have reached the hourly limit of donations for this Collective', order);
+        throw new Error('You have reached the hourly limit of donations for this Collective');
+      }
+      ordersFromCollectiveAndCollectiveCache.set(
```
znarf: I recommend to put the cacheKey in a variable so that you can use it in the get and in the set without repeating it.

server/graphql/v1/mutations/orders.js
```diff
@@ -68,6 +80,30 @@ export async function createOrder(order, loaders, remoteUser) {
       );
     }
 
+    // Check if FromCollective reached max limit of orders per hour
+    if (ordersFromCollectiveCache.get(order.fromCollective.id)) {
+      if (ordersFromCollectiveCache.get(order.fromCollective.id) > 10) {
+        debug('Orders Cache: You have reached the hourly limit of donations', order);
+        throw new Error('You have reached the hourly limit of donations');
+      }
+      ordersFromCollectiveCache.set(order.fromCollective.id, ordersFromCollectiveCache.get(order.fromCollective.id) + 1);
```
znarf: What happens if the value is undefined? I would suggest to explicitly set it to 0.

marcogbarcellos: I agree with all your other comments, but in this one I disagree as:
1) We are already creating the order so I'd set to 1 and not 0
2) I check whether the value for that key exists, if it doesnt I'm setting the new key to one(look the if else).

server/graphql/v1/mutations/orders.js
marcogbarcellos: It will appear when the user reaches the limit and the error is thrown. do you think it's better showing this everytime? I think showing it everytime someone creates an order may pollute the logs too much and imo kinda unnecessarily. Not a big deal though, lmk what you think works best...

PR #1505, opened 2018-11-07
Ensure user is logged in when making a donation with an existing email

Betree:
Some of the test were failing because an `user` field was added in global `order` object in another test.

To avoid this kind of errors in the future I've added an `Object.freeze` on it to make it immutable and added `const order = cloneDeep(baseOrder);` in tests to make local copies.


xdamman:
Looks good but need to make the tests pass.

Betree:
I went a little too fast on this one, tests needed more attention. I've fixed them + added new.

PR message edited to explain the `Object.freeze` on `order` in `test/graphql.user.test.js`

PR #1506, opened 2018-11-07
Link VirtualCard origin collective to Order and PaymentMethod

Betree:
- Linked virtual card origin collective to `PaymentMethod` (`CreatedByCollectiveId` field)
- Linked virtual card origin collective to `Order` (`UsingVirtualCardFromCollectiveId` field)
- [Fixed an issue allowing to create VirtualCards with another collective payment method](https://github.com/opencollective/opencollective-api/commit/e3bccc3b2abe5add27ab8f54879757d2e63948db)
- [Fixed tests](https://github.com/opencollective/opencollective-api/commit/d03a9fb2b7f30cf4df9830896bf4d1fbe8b57dbe)


Betree:
@znarf @marcogbarcellos I think this is now ready to be merged to `feature/virtual-cards-release`. I simplified the model and only store `UsingVirtualCardFromCollectiveId` in `Transactions`.

The major change here is in Collective `getTotalAmountSpent` function that now integrates transactions with `CollectiveId` set to collective (as before) **but also** the one that matches `UsingVirtualCardFromCollectiveId` in the total.

https://github.com/opencollective/opencollective-api/blob/65f78b5f2216268697d11dab21f66b16bb6f90df/server/models/Collective.js#L1771-L1802

**So for example with:**
SQL Query results

**You'll get:**
Organization total amount spent

Betree:
In last commit (https://github.com/opencollective/opencollective-api/pull/1506/commits/b9b3560f3d2402fcb9c99e84f07098fa52dcf86f) I've plugged VirtualCard emitter in members `getTotalAmountSpent` calculation.

It work as intended, this data (truncated)

|     type   |       name   |  directDonations      |  donationsThroughEmittedVirtualCards      |     totalDonations   |  
|----------------|--------------------|--------|--------|--------|
| "ORGANIZATION" | "OrganizationTest" | 11000  | 139398 | 150398 |
| "USER"         | "SuperUser"        | 133398 | 0      | 133398 |
| "USER"         | "anonymous"        | 100000 | 0      | 100000 |

**renders**
CollectiveBannerPreview

### However

This could be confusing for the user, as the total amount show below (`€2,395`) will not match what he may see above by hovering the different donators. **So** we need more data, I've added the `directDonationsFromTo` and `donationsThroughEmittedVirtualCardsFromTo` to the `AllMembers` graphql endpoint.

It will allow us to be more explicit on frontend, to show things like `Financial contributions: xxx€ (yyy€ through gift cards)`

xdamman:
I'm late on this but I'm not a fan of adding `UsingVirtualCardFromCollectiveId` to the `Transaction` model. It won't play nicely with the new Ledger that @marcogbarcellos is working on.
We can keep it that way but once we have the new ledger, this will have to be rethought.

Why not simply having a loader to fetch the payment method for each transaction?

xdamman:
Payment methods should never be deleted. They could be de-activated, but never deleted.
The reason being that we should be able to show any past transaction and fetch the details of the payment method used at the time.

Betree:
Ok thanks for the information @xdamman. Migrating to a different solution once the new ledger will be there should be pretty easy as data is now available in two places. I'll keep this in mind.

PR #1507, opened 2018-11-08
Virtual Cards 2nd release

znarf:


server/graphql/v1/mutations/orders.js
```diff
@@ -28,34 +28,58 @@ const ordersLimitCache = LRU({
...
   // temporarily for "test" or "circleci" ENVs(both test environments) as most tests create more than 2 orders per (fromcollective,collective) and it fails
-  if (process.env.NODE_ENV !== 'test' && process.env.NODE_ENV !== 'circleci' &&
-      fromCollectiveIdOrUserEmailKey && collectiveIdKey) {
-    const fromCollectiveValue = ordersLimitCache.get(fromCollectiveIdOrUserEmailKey);
+  if (
+    process.env.NODE_ENV !== 'test' &&
+    process.env.NODE_ENV !== 'circleci' &&
+    fromCollectiveIdOrUserEmailKey &&
+    collectiveIdKey
+  ) {
```
marcogbarcellos: out of curiosity: are we following the up to 120 chars in one line, right? I personally prefer having less lines and filling out those lines with something closer to the 120 chars limi

Betree:
@znarf Re-based on master. I tested that organization is still credited correctly, it works.

PR #1508, opened 2018-11-08
Script to refund transactions and Transaction.data update

marcogbarcellos:
Adding Script to refund transactions either from stripe or only in the opencollective database given an array of `FromCollectiveId`s, considering this Collective ids on the following query:
```
where: {
      FromCollectiveId: fromCollectiveIds,
      type: 'CREDIT',
      RefundTransactionId: null
    }
```
where `fromCollectiveIds` is the array. 

I'm also updating the `Transaction.data` payload field: 
- after a refund, because the charge information is updated on the "ordinary"(non-refund) transactions regarding stripe.
- the refund transactions payload data was inconsistent if compared with the "ordinary" transactions. Standardizing it as well..

Also other fixes:
fixes https://github.com/opencollective/opencollective/issues/1416
fixes https://github.com/opencollective/opencollective/issues/1417

### DIFF
--- a/scripts/add_refund_transactions_from_collective.js
+++ b/scripts/add_refund_transactions_from_collective.js
@@ -5,8 +5,7 @@
  * from status 'ACTIVE' to status 'CANCELLED'
  *  1) Refund transactions
  *  2) Mark Orders as cancelled
- *  2) Get transactions ORders and Mark the order Subscriptions of orders as isActive False and deactivedAt now()
- * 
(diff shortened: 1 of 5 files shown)


znarf:
@marcogbarcellos Fix the tests? Is there anything that was modified when using it live?

marcogbarcellos:
no @znarf , that's exactly what we ran in prod!

marcogbarcellos:
just re-ran the tests in circleci and it passed as expected. lmk if you have something else otherwise I think we're good to merge.

PR #1509, opened 2018-11-08
Upgrade to babel 7

HipsterBrown:
Used `babel-upgrde` to migrate this repo to use Babel 7. This includes improvements to `babel-node` to read the local `.babelrc`. Tested running locally.


znarf:
@HipsterBrown In the latest version, what is the rationale using Bluebird? Could that be avoided?

.babelrc
```diff
@@ -1,7 +1,21 @@
...
+  ],
   "plugins": [
     "add-module-exports",
-    "lodash"
+    "lodash",
+    "@babel/plugin-proposal-class-properties",
+    ["@babel/plugin-transform-async-to-generator", {
+      "module": "bluebird",
+      "method": "coroutine"
+    }]
```
znarf: More specifically about Bluebird, why is that needed? What is happening without that plugin and this configuration?

server/env.js
```diff
@@ -1,4 +1,6 @@
-import 'babel-polyfill';
+import Promise from 'bluebird';
+
+global.Promise = Promise;
```
znarf: More specifically about Bluebird.

Why do we need to declare it as a global? How was that handled before this PR?

Where are we relying on a Bluebird as the global Promise object?

HipsterBrown: Before `babel-polyfill` was replacing Promise with Bluebird and `async` functions with the `regenerator-runtime` (https://babeljs.io/docs/en/v7-migration#remove-proposal-polyfills-in-babel-polyfill-https-githubcom-babel-babel-issues-8416). We should get to a place where we use native Promises but that is a larger project.

.babelrc
znarf: What happens if we use `@babel/plugin-transform-async-to-generator` without this configuration, with just the default?

What happens if we skip `@babel/plugin-transform-async-to-generator`? Isn't async/await supported natively in node since version 7?

HipsterBrown:
@znarf I believe this PR is now using only the minimum babel-plugins needed for the current codebase to work without regression. The changes have also caught a few errors in our code as well. 

Thanks for the thorough review. :+1:

PR #1512, opened 2018-11-11
Limit creation of user in certain conditions

marcogbarcellos:


xdamman:
This is quite arbitrary to limit the number of anonymous users in the past 60 minutes.
As you can see in this query,  4 times we already had more than 6 anonymous users created in a given hour: 
```sql
SELECT count(*) as count, substring("createdAt"::text from 1 for 13) as "yyyy-mm-dd hh" 
FROM "Collectives" 
WHERE type='USER'
  AND slug LIKE 'anonymous%' GROUP BY "yyyy-mm-dd hh"
ORDER BY count DESC
```


That said, I think it's a good first mitigation for the problem right now so I see no problem to deploy that. But we need to know that this will create issues in the future.

Ideally, we should limit this to 3 (or 2 or 4?) anonymous accounts per hour per IP address hash (because we don't want to store IP addresses to protect our users' privacy). Please note that the IP address from the perspective of the API will always be the IP address of the frontend server, so it has to be passed by the frontend proxy `/api/*` as a special header.

PR #1515, opened 2018-11-12
Setup to read client IP address

znarf:


HipsterBrown:
@znarf Can you please add a description as to what this PR is doing and why?

znarf:
@HipsterBrown done!

PR #1516, opened 2018-11-12
IP limit on orders

znarf:


server/graphql/v1/mutations/orders.js
```diff
@@ -19,51 +19,57 @@ import {
...
+  if (reqIp && !has(order, 'fromCollective.id')) {
+    limits.push({ key: reqIp, value: 10 });
+  }
+
+  for (const limit of limits) {
+    const count = ordersLimitStore.get(limit.key);
+    if (count) {
+      debugOrder(`${count} orders for limit '${limit.key}'`);
+      const limitReached = count > limit.value;
+      ordersLimitStore.set(limit.key, count + 1);
```
Betree: We will increment the count even if the order fails, which means a user constantly retrying to order will never be able to do so. Is it intentional?

znarf: Incrementing or not is not changing much. What matters is the 1 hour TTL in the end. The user hitting the limit will only be able to something after 1 hour, when the key expires.

Betree: I didn't properly understand how the LRU cache worked, thanks for the explanation. Sounds good to me :+1:

marcogbarcellos: I think  there is a ttl that will solve(clean up after 1h) the problem i understood you're trying to say here. isn't it?

PR #1518, opened 2018-11-13
Recaptcha integration

znarf:


xdamman:
Is there an issue linked to it?
What's the behavior of this?

znarf:
@xdamman https://github.com/opencollective/opencollective/issues/1457

migrations/201811160000-add-data-to-orders.js
```diff
@@ -0,0 +1,27 @@
...
+        console.log('>>> done');
+      });
+  },
+
+  down: (queryInterface, Sequelize) => {
+    return queryInterface
+      .removeColumn('Orders', 'data')
+      .then(() => queryInterface.removeColumn('OrderHistories', 'data'))
+      .then(() => {
+        console.log('>>> done');
```
marcogbarcellos: not a big deal but we could remove this console log later!

znarf: @marcogbarcellos It's a migration

marcogbarcellos: you could then improve the message on the console log as when you run the sequelize `db:migrate:undo` it'll run all files and will perhaps not be that clear it's regarding this file.

znarf: I just did this migration like multiple before. If these `'>>> done'` messages are a problem, then we should remove all of them.

PR #1520, opened 2018-11-14
Background Jobs: do not manually set cache for mem queries

HipsterBrown:
Fixes opencollective/opencollective#1435

The `mem` package has an internal timer for removing old cached entries based on the `maxAge` option. This updates the background job to just call those cached queries after the `maxAge` to refresh the internal cache.


znarf:
The problem with this implementation is that the timings can become out of sync, leading to the `refreshCache` being ineffective (if called before the expiration) or too late (after the expiration).

The concern is that there is no guarantee that the cache will be hot anymore.

Betree:
I agree that current implementation could end up to a cold cache. Not sure if we can work this out using `mem` as it does not allow to manually store a result.

HipsterBrown:
Thanks for the feedback @znarf @Betree 

So when reviewing this issue again, I realized original errors fixed in (https://github.com/opencollective/opencollective-api/pull/1517) was because of the refresh cache changing the expected values from the queries. 

I reverted the old use of `mem` and changed the refreshCache to set the cache with the Promise returned by the query.

znarf:
@HipsterBrown Can you update the title of the PR?

I checked the code and this is great to fix the regression.

However, now that I had a look at it, I don't like the idea of storing Promises in cache, it might work but I would prefer in the future to only use caching as simple key/value store.

PR #1521, opened 2018-11-15
Internal Cache API

znarf:
We're using more and more caching (through `memoize` or `lru-cache`) but this is currently lacking a coherent API.

This PR introduces:
 - a cache library (key/value) that can be used everywere it's needed
 - support for a memcache backend to scale that in production


server/controllers/users.js
```diff
@@ -71,7 +68,7 @@ export const exists = (req, res) => {
   } else {
     return models.User.findOne({ attributes: ['id'], where: { email } }).then(
       user => {
-        cache.set(email, Boolean(user));
+        cache.set(email, Boolean(user), tenMinutesInSeconds);
```
Betree: I guess this is the one we should disable as it may cause issues with newly registered users. 

In all cases I don't really see the point of having this cache as `email` parameter will almost always be different.

Also this is a very light query and the (potential) gain of caching this could be overthrown by the fact it is easy to fill it up with trashy data and consume RAM.

znarf: @Betree Yeah, let's do that.

server/lib/cache.js
```diff
@@ -1,32 +1,113 @@
...
+  } else if (lru) {
+    return lru.del(key);
+  }
+}
+
+async function cacheClear() {
+  debugCache('clear');
+  if (lru) {
+    return lru.reset();
+  }
```
Betree: Why not clearing the `memcached`cache too?

znarf: Good, question! The `memcached-promisify` library I decide to use is not exposing the `flush` method. Sad!

We should use something else.

znarf:
@Betree About global cache, the use case is to use 1 memcached server or 1 memcached cluster. I had only but good experience with that. So, it's global.

I think what we might do to go in your direction is turning the cache API in an object that needs to be instantiated, and making it automatically prefix the keys.

Granular flushing with memcached is an issue. It's doable but needs to be handled from the client and that might increase complexity. We might skip that feature for now.

Betree:
Granular flushing is indeed not a big requirement.

Prefixing the keys with a namespace is a good idea, the alternative is to be careful with naming.

PR #1535, opened 2018-11-22
Do not use collective as referral when using one of its virtual cards

Betree:
This will ensure we don't count collectives as referral when the transaction is paid with one of its virtual cards. Following queries will need to be executed to remove the incorrect referral attributions.

In production this represents a total of 226 orders **all** made using virtual cards from Triplebyte for a total around $14k (ping @piamancini). 

- To show the entries that need to be fixed

```sql
SELECT 	*
FROM 		"Orders" o
INNER JOIN 	"PaymentMethods" pm ON o."PaymentMethodId" = pm.id
INNER JOIN 	"PaymentMethods" spm ON pm."SourcePaymentMethodId" = spm.id
WHERE 		o."ReferralCollectiveId" = spm."CollectiveId";
```

- To fix them

```sql
UPDATE	"Orders" o
SET	"ReferralCollectiveId" = NULL
FROM	"PaymentMethods" pm, "PaymentMethods" spm
WHERE	o."PaymentMethodId" = pm.id
AND	pm."SourcePaymentMethodId" = spm.id
AND 	o."ReferralCollectiveId" = spm."CollectiveId";
```


xdamman:
Also need to remove the role "FUNDRAISER" in the `Members` table for those past transactions.

znarf:
Should we have the update as a migration?

znarf:
@Betree The migration is helping document what we're doing to the database, I think it's the good practice here and in the future.

Betree:
@xdamman I wanted to use the results of following query to remove invalid fundraisers from members, but found out it return (11) results before fixing the referrers with above query. Any idea what those could be? 

```sql
/* Delete fundraiser members when no order has them as referral */

DELETE FROM "Members"
WHERE id in (
	SELECT *
	FROM "Members" m
	INNER JOIN "Collectives" c ON m."MemberCollectiveId" = c.id
	LEFT JOIN "Orders" o ON m."MemberCollectiveId" = o."ReferralCollectiveId"
	WHERE o.id IS NULL
	AND c.type = 'ORGANIZATION'
	AND m.role = 'FUNDRAISER';
)
```

More details:
Query results

<details title="xx">
 <summary>Corresponding query</summary>

```sql
SELECT c.id AS "MemberCollectiveId", c.slug AS "MemberSlug", m."CollectiveId" as "CollectiveId", tc.slug as "CollectiveSlug"
FROM "Members" m
INNER JOIN "Collectives" c ON m."MemberCollectiveId" = c.id
INNER JOIN "Collectives" tc ON m."CollectiveId" = tc.id
LEFT JOIN "Orders" o ON m."MemberCollectiveId" = o."ReferralCollectiveId"
WHERE o.id IS NULL
AND c.type = 'ORGANIZATION'
AND m.role = 'FUNDRAISER';
```

</details>

xdamman:
No idea but I'm also not sure I'm following your query. Why not reusing the first query in this issue as the selector? From it you can extract the `FromCollectiveId` and `CollectiveId` and delete all `Members` entries that match this combo + `FUNDRAISER` role.

Betree:
I thought about that but I'm afraid we could delete legit Members entries if collective is also a valid fundraisers of given collectives. But this is just a guess, I haven't actually tried to run queries in production to ensure we have such cases.

PR #1538, opened 2018-11-26
More permissive checkOrdersLimit in dev

Betree:
During development I got stuck a few times by this limit. As the message is (voluntarily) unclear, a typical open source developer will have to search for the root cause of this error. My fix here is a quick one, a better approach would be:
1. To log a different message in dev that inform the developer on what can be done to overthrow the limitation
2. To use an ENV variable to configure the orders limit per hour, set to `MAX_ORDERS_PER_HOUR=10000` in dev

My favorite approach being 2.

@znarf do you have an opinion about this?

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -4,6 +4,7 @@ import debug from 'debug';
 import md5 from 'md5';
 import Promise from 'bluebird';
 import { pick, omit, get } from 'lodash';
+import config from 'config';
 
 import models from '../../../models';
 import * as errors from '../../errors';
@@ -24,10 +25,11 @@ const oneHourInSeconds = 60 * 60;
 const debugOrder = debug('order');
 
 function checkOrdersLimit(order, remoteUser, reqIp) {
-  if (process.env.NODE_ENV === 'test' || process.env.NODE_ENV === 'circleci') {
+  if (['circleci', 'test'].includes(process.env.NODE_ENV)) {
     return;
   }
 
+  const ordersLimits = config.limits.ordersPerHour;
   const collectiveId = get(order, 'collective.id');
   const fromCollectiveId = get(order, 'fromCollective.id');
   const userEmail = get(order, 'user.email');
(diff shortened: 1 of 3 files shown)


znarf:
It's great to still execute the code in dev environment, might help spot regressions earlier than staging.

We can have a higher limitation AND a better message in dev.

Betree:
+ an env variable could allow to dynamically configure the limit in case of attack. I'll add this to this PR

.env.development
```diff
@@ -5,3 +5,6 @@ PORT=3060
 WEBSITE_URL=http://localhost:3000
 SESSION_SECRET=i&j@/V6Wx.`g>Aq?qQQF(>u)/Erm;3A=
 DEBUG=email
+ORDER_LIMIT_ON_ACCOUNT_PER_HOUR=100
+ORDER_LIMIT_ON_ACCOUNT_PER_HOUR_PER_COLLECTIVE=20
+ORDER_LIMIT_ON_IP_PER_HOUR=20
```
znarf: Let's use `config/default.json` for that, this file is gonna be deprecated soon.

server/graphql/v1/mutations/orders.js
```diff
@@ -27,10 +27,16 @@ const oneHourInSeconds = 60 * 60;
...
-  if (process.env.NODE_ENV === 'test' || process.env.NODE_ENV === 'circleci') {
+  if (['circleci', 'test'].includes(process.env.NODE_ENV)) {
     return;
   }
 
+  const {
+    ORDER_LIMIT_ON_ACCOUNT_PER_HOUR = 10,
+    ORDER_LIMIT_ON_ACCOUNT_PER_HOUR_PER_COLLECTIVE = 2,
+    ORDER_LIMIT_ON_IP_PER_HOUR = 2,
+  } = process.env;
```
znarf: Let's use the `config` module instead of environment variables.

xdamman:
Great, I was also hitting that issue in my dev environment and that was quite annoying. Thanks for picking this up.

Betree:
@znarf Changed code to use `config`, it looks way better like that :+1:

PR #1555, opened 2018-11-30
Introduce logger

znarf:


Betree:
Any clues about when we should use `logger` vs `debug`?

znarf:
@Betree This is not articulated yet. 

- Use `logger` by default, instead of using console.log. Use properly the different levels (verbose, info, warn, error). Default level is 'info', and 'warn' when testing.
- Use `debug` when you want detailed logging that is scoped to one feature or component, when  what you want is "opt in logging", and when you're fine that it's disabled by default in all environments (development, test, staging, production)

PR #1561, opened 2018-11-30
Update async to version 2.6.1

znarf:


eloyekunle:
@znarf Tests pass now.

znarf:
@eloyekunle we also use `async` in `collective.js` `deleteTransaction`.

Does that usage looks correct to you with the new version?

eloyekunle:
@znarf I've fixed the affected usage in `collectives.js`.

PR #1563, opened 2018-12-03
Correct "total donated" amount in user's monthly report

marcogbarcellos:
This PR properly filters orders to return orders only when their subscriptions are active.

references https://github.com/opencollective/opencollective/issues/1504
fixes https://github.com/opencollective/opencollective/issues/1505

### DIFF
--- a/cron/monthly/user-report.js
+++ b/cron/monthly/user-report.js
@@ -112,7 +112,10 @@ const processBacker = async FromCollectiveId => {
   const query = {
     attributes: ['CollectiveId', 'HostCollectiveId'],
     where: {
-      FromCollectiveId,
+      [Op.or]: {
+        FromCollectiveId,
+        UsingVirtualCardFromCollectiveId: FromCollectiveId,
+      },
       type: 'CREDIT',
       createdAt: { [Op.gte]: startDate, [Op.lt]: endDate },
     },
@@ -198,17 +201,10 @@ const processBacker = async FromCollectiveId => {
       },
       deletedAt: null,
     },
-    include: [
-      {
-        model: models.Subscription,
-        where: {
-          isActive: true,
-        },
-      },
-    ],
+    include: [{ model: models.Subscription }],
   });
-
-  const ordersByCollectiveId = groupBy(orders, 'CollectiveId');
+  // group orders(by collective) that either don't have subscription or have active subscription
+  const ordersByCollectiveId = groupBy(orders.filter(o => !o.Subscription || o.Subscription.isActive), 'CollectiveId');
   const collectivesWithOrders = [];
   collectives.map(collective => {
     // It's only possible to have one order with subscription active Per Collective
--- a/cron/monthly/user-report.js
+++ b/cron/monthly/user-report.js
@@ -18,6 +18,7 @@ import debugLib from 'debug';
 import models, { Op } from '../../server/models';
 import emailLib from '../../server/lib/email';
 import roles from '../../server/constants/roles';
+import ORDER_STATUS from '../../server/constants/order_status';
 import { formatCurrencyObject, formatArrayToString } from '../../server/lib/utils';
 import { convertToCurrency } from '../../server/lib/currency';
 import path from 'path';
@@ -98,7 +99,6 @@ const init = async () => {
   }
 
   console.log(`Preparing the ${month} report for ${FromCollectiveIds.length} backers`);
-
   await Promise.each(FromCollectiveIds, processBacker);
 
   const timeLapsed = Math.round((new Date() - startTime) / 1000);
@@ -196,6 +196,12 @@ const processBacker = async FromCollectiveId => {
         createdAt: { [Op.gte]: startDate, [Op.lt]: endDate },
         SubscriptionId: { [Op.ne]: null },
       },
+      status: {
+        [Op.and]: {
+          [Op.ne]: ORDER_STATUS.ERROR,
+          [Op.ne]: ORDER_STATUS.CANCELLED,
+        },
+      },
       deletedAt: null,
     },
     include: [{ model: models.Subscription }],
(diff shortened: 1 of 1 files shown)


cron/monthly/user-report.js
```diff
@@ -198,25 +196,32 @@ const processBacker = async FromCollectiveId => {
...
     },
-    include: [{ model: models.Subscription }],
-  });
-  const ordersByCollectiveId = {};
-  orders.map(o => {
-    ordersByCollectiveId[o.CollectiveId] = o;
+    include: [
+      {
+        model: models.Subscription,
+        where: {
```
Betree: Adding a where clause to the include automatically makes it required. I think we may miss orders that don't have a subscription here.

marcogbarcellos: Hum, I'll better use just a filter then. thanks for the good catch.

Betree:
@marcogbarcellos Regarding by previous comment for VirtualCards, I just merged https://github.com/opencollective/opencollective-api/pull/1562 so if you rebase on master you should be able to replace the whole query line `112` to include virtual cards with:

```es6
const transactions = await backerCollective.getTransactions({
  attributes: ['CollectiveId', 'HostCollectiveId'],
  includeUsedVirtualCardsEmittedByOthers: false,
  type: 'CREDIT',
  startDate,
  endDate
})
```

Please tell me if this is unclear.

cron/monthly/user-report.js
```diff
@@ -112,7 +112,10 @@ const processBacker = async FromCollectiveId => {
   const query = {
     attributes: ['CollectiveId', 'HostCollectiveId'],
     where: {
-      FromCollectiveId,
+      [Op.or]: {
+        FromCollectiveId,
+        UsingVirtualCardFromCollectiveId: FromCollectiveId,
+      },
```
znarf: No need to support Virtual Card here at the moment.

Betree: Yes sorry @marcogbarcellos, I didn't realized this was user report and user cannot emit virtualcards, my first comment was misleading.

cron/monthly/user-report.js
```diff
@@ -198,25 +199,25 @@ const processBacker = async FromCollectiveId => {
...
-  const ordersByCollectiveId = {};
-  orders.map(o => {
-    ordersByCollectiveId[o.CollectiveId] = o;
-  });
+  // group orders(by collective) that either don't have subscription or have active subscription
+  const ordersByCollectiveId = groupBy(orders.filter(o => !o.Subscription || o.Subscription.isActive), 'CollectiveId');
   const collectivesWithOrders = [];
   collectives.map(collective => {
+    // It's only possible to have one order with subscription active Per Collective
+    const collectiveOrder = ordersByCollectiveId[collective.id] && ordersByCollectiveId[collective.id][0];
```
znarf: What happens if `collectiveOrder` is null?

cron/monthly/user-report.js
```diff
@@ -198,25 +199,25 @@ const processBacker = async FromCollectiveId => {
...
+      deletedAt: null,
     },
     include: [{ model: models.Subscription }],
   });
-  const ordersByCollectiveId = {};
-  orders.map(o => {
-    ordersByCollectiveId[o.CollectiveId] = o;
-  });
+  // group orders(by collective) that either don't have subscription or have active subscription
+  const ordersByCollectiveId = groupBy(orders.filter(o => !o.Subscription || o.Subscription.isActive), 'CollectiveId');
```
znarf: It might be possible that we have collectives with multiple donations each month (multiple single time donation,  mix of subscription + single time, and even multiple subscriptions).

How do we handle that?

marcogbarcellos: This script was considering only subscription orders. At first I thought it was intentional but It makes senses to cover all orders regardless. I'll update the code to reflect it.

cron/monthly/user-report.js
marcogbarcellos: if you look at the `computeStats` method it checks whether the collective has an Order or an expense so we won't have validation problems.

cron/monthly/user-report.js
marcogbarcellos: Sorry, i explained wrong. I meant it was only getting the first order of each collective which would ONLY make sense if it was considering subscriptions only(afaik a contributor is supposed to have only one subscription per collective). The following part of the master user-report is wrong because it's basically iterating over the `Order.findAll` query and if there is more than one order of the same collective it would consider only the latest one(also there was not filter to check whether the subscription was active anymore) :
https://github.com/opencollective/opencollective-api/blob/master/cron/monthly/user-report.js#L204-L207

marcogbarcellos:
all good to merge here @znarf @Betree >

cron/monthly/user-report.js
znarf: @marcogbarcellos ?

cron/monthly/user-report.js
```diff
@@ -198,25 +196,28 @@ const processBacker = async FromCollectiveId => {
...
-      ...collective,
-      order: ordersByCollectiveId[collective.id],
-    });
-  });
-  collectivesWithOrders.sort((a, b) => {
-    if (get(a, 'order.totalAmount') > get(b, 'order.totalAmount')) return -1;
-    else return 1;
+    // It's only possible to have one order with subscription active Per Collective
+    const collectiveOrders = ordersByCollectiveId[collective.id] && ordersByCollectiveId[collective.id];
+    collectiveOrders &&
```
marcogbarcellos: @znarf that commented you've just marked me is outdated. I'm just pushing to `collectivesWithOrders` after validating`collectiveOrders`.

cron/monthly/user-report.js
```diff
@@ -198,25 +196,28 @@ const processBacker = async FromCollectiveId => {
...
-    collectivesWithOrders.push({
-      ...collective,
-      order: ordersByCollectiveId[collective.id],
-    });
-  });
-  collectivesWithOrders.sort((a, b) => {
-    if (get(a, 'order.totalAmount') > get(b, 'order.totalAmount')) return -1;
-    else return 1;
+    // It's only possible to have one order with subscription active Per Collective
+    const collectiveOrders = ordersByCollectiveId[collective.id] && ordersByCollectiveId[collective.id];
```
znarf: Doesn't looks good.

cron/monthly/user-report.js
znarf: In that case, avoid the `&&` syntax, `if` will be easier to read.

znarf:
@marcogbarcellos What about line 214 comment?

marcogbarcellos:
@znarf that 214 comment doesn't make sense anymore as we are getting all orders given a `FromCollective` and a `Collective`.

znarf:
@marcogbarcellos Ok, my 2nd comment was made invalid by: https://github.com/opencollective/opencollective-api/compare/e461b0a7121364cbf8191262bab8652ee23cc8dc..a9b9bc88908119b5f679fc120f5946cb6d511a32

cron/monthly/user-report.js
```diff
@@ -451,6 +452,7 @@ const computeStats = async (collectives, currency = 'USD') => {
   stats.expensesBreakdownString = `${Object.keys(categories).length > 3 ? ', mostly in' : ' in'} ${formatArrayToString(
     ar,
   )}`;
+  console.log(`>>> Stats: ${JSON.stringify(stats, null, 2)}`);
```
znarf: Remove console.log ?

marcogbarcellos: I believe this `console.log` may be really helpful to check the logs after running these scripts. in this case it'll show the consolidated stats per user. I think it worths keeping this console.log as we already have other logs in this report as well. But let me know if you think it's too much.

marcogbarcellos: afaik the same way we can log the app but instead of using `--app` we'd use the scheduler job id. example: `heroku logs -d scheduler.5255`

PR #1568, opened 2018-12-05
Fix broken link on user monthly report

marcogbarcellos:
Fixing broken link on user monthly report due to wrong path.

Testing locally added 2 duplicated <li>, difference was only the link.
screen shot 2018-12-05 at 7 50 20 am

Here is the result(Sent TEST email locally):
screen shot 2018-12-05 at 7 50 06 am

fixes https://github.com/opencollective/opencollective/issues/1503


znarf:
@marcogbarcellos Two links are mentioned in the issue, it seems this is only fixing one?

marcogbarcellos:
@znarf sorry i missed the second broken link on the issue! fixed now as well!

PR #1572, opened 2018-12-06
Simplify Collective getTransactions method's query

marcogbarcellos:
Small improvement to make query simpler.

references https://github.com/opencollective/opencollective/issues/1519

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1658,32 +1658,19 @@ export default function(Sequelize, DataTypes) {
    *  cards from other collectives when set to false.
    */
   Collective.prototype.transactionsWhereQuery = function(includeUsedVirtualCardsEmittedByOthers = true) {
-    const debitTransactionOrQuery = includeUsedVirtualCardsEmittedByOthers
-      ? // Include all transactions made by this collective or using one of its
-        // virtual cards
-        { CollectiveId: this.id, UsingVirtualCardFromCollectiveId: this.id }
-      : // Either Collective made the transaction without using a virtual card,
-        // or a transaction was made using one of its virtual cards - but don't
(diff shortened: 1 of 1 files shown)


Betree:
I don't think this works.

### If `includeUsedVirtualCardsEmittedByOthers` is false

- It will wrongly include the transactions made using gift cards emitted by others

Admitting we're interested in getting user transactions, the new query will return entries that match:

```es6
{CollectiveId: UserCollectiveId, UsingVirtualCardFromCollectiveId: RandomOrganizationCollectiveId}
```

We don't want them as we've set `includeUsedVirtualCardsEmittedByOthers` to false.

This was the reason for the:
```es6
[
  { CollectiveId: this.id, UsingVirtualCardFromCollectiveId: null },
  { UsingVirtualCardFromCollectiveId: this.id },
];
```

Aka get the transactions I've made without using a virtual card (UsingVirtualCardFromCollectiveId=null), or the ones made using one of my virtual cards.

- It will not include the transactions made using own virtual cards

When making a transaction using a virtualcard, `CollectiveId` is referencing the user that made the transaction while the organization that emitted the virtual card is referenced by `UsingVirtualCardFromCollectiveId`. We must use this field to return organization the transactions that was made using one of their virtualcard.

---

I get that this can be confusing and I've started writing some documentation to make it clearer. I'll need to reference this in function's comments. Do you know what would be the best place for me to start uploading these (markdown) documentation? There is the `docs/` folder, but i'd love to use the github wiki so we don't have to commit to update it.

marcogbarcellos:
Hum I think now it's clearer for me @Betree . Now I understand the importance of `UsingVirtualCardFromCollectiveId: null`. thanks for clarifying. Closing here..

PR #1574, opened 2018-12-10
VirtualCards admin

Betree:
- Added an endpoint to list all virtual cards created by a collective
- Added a `canCreateVirtualCards` field, only used on frontend side as we don't want to break third parties using this API.
- Updated routes to `editCollective` in emails to match the new scheme (see https://github.com/opencollective/opencollective-frontend/pull/1156)


migrations/20181204131511-add_canCreateVirtualCards_column_on_collectives.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+const canCreateVirtualCardsColName = 'canCreateVirtualCards';
```
znarf: I don't like the idea of creating a column for that. I will prefer to use `data` or `settings` for that.

Betree: I wanted this subject to be discussed during the demo, regarding the fact we're planning to make gift cards public should we really keep that feature?

Betree: Added `canCreateVirtualCards` to settings

PR #1577, opened 2018-12-10
Post transactions to ledger service(asynchronously)

marcogbarcellos:
After creating a transaction on the api, we should send it to the ledger service so it can also persist it. This PR enables 2 ways to do it:
- Queue which will be the default way to do it. For this to work we'll need to set the Environment variable `QUEUE_ENABLED ` to true.
- HTTP Post endpoint.

### DIFF
--- a/server/models/Transaction.js
+++ b/server/models/Transaction.js
@@ -1,18 +1,16 @@
 /** @module models/Transaction */
 
 import Promise from 'bluebird';
-import amqp from 'amqplib';
-import axios from 'axios';
 import activities from '../constants/activities';
 import { TransactionTypes } from '../constants/transactions';
 import CustomDataTypes from './DataTypes';
 import uuidv4 from 'uuid/v4';
-import { ledger, ledgerQueue } from 'config';
 import debugLib from 'debug';
 import { toNegative } from '../lib/math';
 import { exportToCSV } from '../lib/utils';
 import { get } from 'lodash';
 import moment from 'moment';
+import { postTransactionToLedger } from '../lib/ledger';
 
 const debug = debugLib('transaction');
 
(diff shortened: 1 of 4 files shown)


server/models/Transaction.js
```diff
@@ -506,5 +510,94 @@ export default (Sequelize, DataTypes) => {
...
+    try {
+      const ledgerTransaction = await getTransactionWithNestedProperties(transaction);
+      if (!process.env.QUEUE_ENABLED) {
+        return axios.post(ledger.transactionUrl, ledgerTransaction);
+      }
+      const conn = await amqp.connect(ledgerQueue.url);
+      const channel = await conn.createChannel();
+      await channel.assertQueue(ledgerQueue.transactionQueue, { exclusive: false });
+      channel.sendToQueue(ledgerQueue.transactionQueue, Buffer.from(JSON.stringify([ledgerTransaction]), 'utf8'));
+    } catch (error) {
```
znarf: What's the point of catching the error to just throw it?

server/models/Transaction.js
```diff
@@ -506,5 +510,94 @@ export default (Sequelize, DataTypes) => {
...
+      }
+    }
+    return transaction;
+  };
+
+  Transaction.postTransactionToLedger = async transaction => {
+    if (transaction.deletedAt || transaction.type !== 'CREDIT') {
+      return Promise.resolve();
+    }
+    try {
```
znarf: Can we move that, maybe not the `postTransactionToLedger` itself, but what's inside into `lib/ledger`? Also add `getTransactionWithNestedProperties`.

server/models/Transaction.js
marcogbarcellos: removing it, was logging before but I don't think it's necessary anymore, at least here.

server/models/Transaction.js
marcogbarcellos: Sure, makes sense. I've created the `lib/ledger` on the other branch. Would you mind if we then merge the [*read*](https://github.com/opencollective/opencollective-api/pull/1578) branch before so I can rebase from master here and do this change?

server/models/Transaction.js
znarf: @marcogbarcellos I think it's best to keep the `try/catch`, we don't want to throw errors if we have a problem there.

I'm not sure what's the behavior with Sequelize hooks, what happen if they throw an error?

marcogbarcellos: Okay, so I'll just log the error. As it's an async method All it did locally was log the error as there was no other method waiting on any answer. But I could instead of throw just `console.error` or `debug` the error if so.

PR #1578, opened 2018-12-10
Get Transactions from Ledger

marcogbarcellos:
This PR changes the `allTransactions` query to support reading data either from the api OR from the ledger through either a boolean parameter on the query or an Environment variable `GET_TRANSACTIONS_FROM_LEDGER `.

### DIFF
--- a/config/development.json
+++ b/config/development.json
@@ -30,5 +30,13 @@
   "github": {
     "clientID": "cdc163e5f5695e16787b",
     "clientSecret": "1cc8e3a615af3a8ca47cd1f15adcfcdd1e880db4"
+  },  
+  "ledger": {
+    "url": "http://localhost:3070",
+    "transactionUrl": "http://localhost:3070/transactions"
+  },
+  "ledgerQueue": {
+    "url": "amqp://localhost",
+    "transactionQueue": "transactions"
   }
 }
--- a/server/graphql/v1/TransactionInterface.js
+++ b/server/graphql/v1/TransactionInterface.js
@@ -1,3 +1,5 @@
+import { get } from 'lodash';
+import models from '../../../server/models';
 import {
   GraphQLInt,
   GraphQLFloat,
@@ -65,7 +67,10 @@ const TransactionFields = () => {
     refundTransaction: {
       type: TransactionInterfaceType,
       resolve(transaction) {
-        return transaction.getRefundTransaction();
+        if (transaction && transaction.getRefundTransaction) {
+          return transaction.getRefundTransaction();
+        }
+        return null;
       },
     },
     uuid: {
@@ -74,7 +79,10 @@ const TransactionFields = () => {
         if (!req.remoteUser) {
           return null;
         }
-        return transaction.getDetailsForUser(req.remoteUser);
+        if (transaction && transaction.getDetailsForUser) {
+          return transaction.getDetailsForUser(req.remoteUser);
+        }
+        return null;
       },
     },
     type: {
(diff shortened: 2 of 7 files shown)


config/development.json
```diff
@@ -26,5 +26,23 @@
...
+    "url": "amqp://localhost",
+    "transactionQueue": "transactions"
+  },
+  "keys": {
+    "opencollective": {
+      "api_key": "dvl-1510egmf4a23d80342403fb599qd",
+      "secret": "dvl-7749ba6ad8e62b53b18b4c30295f45d849c1",
+      "session_secret": "i&j@/V6Wx.`g>Aq?qQQF(>u)/Erm;3A="
+    }
+  },
```
znarf: `keys` added there are outdated and unnecessary.

config/development.json
```diff
@@ -26,5 +26,23 @@
...
+  "stripe": {
+    "platformPublishableKey": "pk_test_VgSB4VSg2wb5LdAkz7p38Gw8"
```
znarf: `stripe` added there is outdated and unnecessary.

server/graphql/v1/queries.js
```diff
@@ -327,14 +343,57 @@ const queries = {
...
+      const allTransactions = await collective.getTransactions({
+        attributes: attributes,
         order: [['createdAt', 'DESC']],
         type: args.type,
         limit: args.limit,
         offset: args.offset,
         startDate: args.dateFrom,
         endDate: args.dateTo,
       });
+      if (!fetchDataFromLedger) return allTransactions;
```
znarf: Why execute and wait for `allTransactions` if `fetchDataFromLedger` is true?

server/graphql/v1/queries.js
```diff
@@ -327,14 +343,57 @@ const queries = {
...
+          parseLedgerTransactionToApiFormat(key, ledgerTransactions[key], {
+            AccountId: args.CollectiveId,
+            legacyUuid: uuid,
+            VirtualCardCollectiveId,
+            HostCollectiveId,
+            RefundTransactionId,
+          }),
+        );
+      }
+      return ledgerFormattedTransactions;
```
znarf: Extract all this block in a separated `ledger` file, which should expose a `getTransactions` method or similar.

server/lib/transactions.js
```diff
@@ -161,6 +170,113 @@ export async function createTransactionFromInKindDonation(expenseTransaction) {
...
+  }
+  // setting paymentMethod
+  if (get(creditTransaction[0], 'fromWallet.PaymentMethodId')) {
+    parsedTransaction.paymentMethod = {
+      id: parseInt(get(creditTransaction[0], 'fromWallet.PaymentMethodId')),
+    };
+  }
+  return parsedTransaction;
+}
+
```
znarf: That can also be added to the `ledger` lib.

server/graphql/v1/queries.js
```diff
@@ -317,8 +321,20 @@ const queries = {
       dateTo: { type: GraphQLString },
       /** @deprecated since 2018-11-29: Virtual cards now included by default when necessary */
       includeVirtualCards: { type: GraphQLBoolean },
+      fetchDataFromLedger: { type: GraphQLBoolean },
     },
     async resolve(_, args) {
+      const fetchDataFromLedger = args.fetchDataFromLedger || process.env.GET_TRANSACTIONS_FROM_LEDGER || false;
```
znarf: If `args.fetchDataFromLedger` is explicitly false, it will use the value of `process.env.GET_TRANSACTIONS_FROM_LEDGER`. This is not what we want.

server/graphql/v1/queries.js
marcogbarcellos: Humm I get your point, perhaps something fails when fetching data from the ledger and then makes it useless. you're right, I'll fix that.

server/graphql/v1/queries.js
```diff
@@ -317,24 +317,50 @@ const queries = {
...
       includeVirtualCards: { type: GraphQLBoolean },
+      fetchDataFromLedger: { type: GraphQLBoolean }, // flag to go with either api or ledger transactions
+      includeHostedCollectivesTransactions: {
+        type: GraphQLBoolean,
+      } /** flag to determine
+        whether we should include the transactions of the collectives of that host(if it's a host collective) */,
     },
     async resolve(_, args) {
+      let fetchDataFromLedger = process.env.GET_TRANSACTIONS_FROM_LEDGER || false;
+      if (args.fetchDataFromLedger) {
```
znarf: This won't go in the if when `args.fetchDataFromLedger` is false.

server/graphql/v1/TransactionInterface.js
```diff
@@ -235,32 +290,34 @@ export const TransactionOrderType = new GraphQLObjectType({
...
+          const getOrder = transaction.getOrder
+            ? transaction.getOrder().then(order => order && order.description)
+            : null;
+          return transaction.description || getOrder;
         },
       },
       privateMessage: {
         type: GraphQLString,
         resolve(transaction) {
-          // TODO: Put behind a login check
```
znarf: Why do we add all this defensive code to check if methods are available?

server/graphql/v1/TransactionInterface.js
```diff
@@ -235,32 +290,34 @@ export const TransactionOrderType = new GraphQLObjectType({
...
-          return transaction.getOrder().then(order => order && order.privateMessage);
+          return transaction.getOrder ? transaction.getOrder().then(order => order && order.privateMessage) : null;
```
znarf: Why do we add all this defensive code to check if methods are available?

server/graphql/v1/queries.js
marcogbarcellos: the default will be false which makes it ok as it'll only catch the `if` if the args are true, otherwise it will already be false **OR** defined by `process.env.GET_TRANSACTIONS_FROM_LEDGER`.

The way it'll be is:
1. the default is false
2. if there is `process.env.GET_TRANSACTIONS_FROM_LEDGER` it will go with that.
3. if there is `args.fetchDataFromLedger` it'll go with that regardless the default and `process.env.GET_TRANSACTIONS_FROM_LEDGER`.

znarf: No, `if (args.fetchDataFromLedger) {`  is not checking if `args.fetchDataFromLedger` is `undefined` or not. It's checking if `args.fetchDataFromLedger` is truthy or falsy. If the value is `false`, it's not going in the condition.

If `GET_TRANSACTIONS_FROM_LEDGER` is true and `args.fetchDataFromLedger` is false, we're not having the expected behavior.

server/graphql/v1/TransactionInterface.js
marcogbarcellos: TransactionInterface is supposed to be expecting a Sequelize Model but when we parse the transactions coming from the ledger, those transactions are just object(with some fields like the Sequelize model) but they don't have their methods. That's why the validation is necessary.

znarf: Can you explain me why it's ok to send `null` there when there is no method? Aren't we expecting a value?

marcogbarcellos: I've checked every `return null` case for those validations. I don't think we're waiting these specific values where we return null/undefined. But also in all other scenarios we have, we are using the sequelize models and the validations I added won't never be catched other than the `transactionsFromLedger`

marcogbarcellos:
@znarf after the comments i'll be moving forward in a couple minutes if there's nothing else! thanks for the review.

PR #1583, opened 2018-12-17
Packages: update deepmerge

znarf:
Upgrading to `deepmerge` 3.0.0 was breaking.

That was fixable but better not to use it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5685,9 +5685,9 @@
       "dev": true
     },
     "deepmerge": {
-      "version": "1.3.0",
-      "resolved": "http://registry.npmjs.org/deepmerge/-/deepmerge-1.3.0.tgz",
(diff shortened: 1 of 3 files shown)


xdamman:
Object.assign is not doing a deep merge. This could create issues with options is more than one level deep.

znarf:
@xdamman I found it funny that it was the only place where we were requiring deepmerge. That's why I tried to remove it. By reversing the Object.assign, we might support any `options` in the future. And add a comment about that. `Object.assign(options, { where: { CollectiveId: this.id } }`.

Or we keep `deepmerge`, update to the latest version, and upgrade the syntax with `{clone: false}`. But that option is already deprecated so, it's a bit sad.

xdamman:
If I remember well, object.assign will break here because options can define new arguments in options.where that would be overridden. Hence the deepmerge

PR #1584, opened 2018-12-17
Fix ics import

znarf:
Better use the `import` syntax than `require`.

No need to have an `eslint` exception anymore.

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1,8 +1,6 @@
 /**
  * Dependencies.
  */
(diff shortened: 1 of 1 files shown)


xdamman:
Have you tested it?
In my memories it wasn’t working for some reason when I imported it that way. Maybe they fixed that in a newer version of the package?

znarf:
@xdamman There is a test covering it in `collective.model.test.js`. It's passing.

Maybe you tried `import ics from 'ics'` and not `import * as ics from 'ics'`?

PR #1594, opened 2018-12-18
Update multer to version 1.4.1

znarf:


znarf:
@Betree how did you test with `InputTypeDropzone`?

Betree:
@znarf I go to `src/components/InputTypeDropzone.js` and comment the following block:

```es6
    // for e2e testing purposes
    if (window.location.hostname === 'localhost') {
      const fileUrl = 'https://d.pr/free/i/OlQVIb+';
      return setTimeout(() => {
        this.setState({ value: fileUrl, url: fileUrl, loading: false });
        return this.props.onChange(fileUrl);
      }, 2500);
    }
```

znarf:
@Betree I think your 400 is unrelated to this change. Check that `AWS_KEY`, `AWS_S3_BUCKET` and `AWS_SECRET` are set and try again?

znarf:
Ok, great to hear that it's working.

Fyi, I added a commit since you initially reported the issue, but that was not fixing a 400.

PR #1597, opened 2018-12-18
Fix yearly email report to backers

xdamman:
Part of https://github.com/opencollective/opencollective/issues/1559

### DIFF
--- a/templates/emails/user.yearlyreport.hbs
+++ b/templates/emails/user.yearlyreport.hbs
@@ -50,6 +50,9 @@ th a {
   margin: 15px;
   font-size: 64px;
 }
+  .btn {
+    height: auto;
+  }
 </style>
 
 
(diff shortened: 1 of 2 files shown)


templates/emails/user.yearlyreport.hbs
```diff
@@ -3,77 +3,87 @@ Subject: 🏅Your year on Open Collective
...
+    margin: 15px;
+    font-size: 64px;
+  }
 </style>
 
 
 <div class="stats">
   <div class="largeEmoji">🎁</div>
   <p>{{tweet.text}}<br />
-  <a href="https://opencollective.com/{{collective.slug}}?{{utm}}">https://opencollective.com/{{collective.slug}}</a></p>
```
znarf: Can we start removing `utm` parameters?

xdamman: yes good idea, it was for back in the days when we had for a short time Google analytics.
But wouldn't this utm tags be understood by the other open source analytics service that you wanted to use? In this case we should keep it no?

znarf: Ok, let's not do that until we have more visibility then.

PR #1598, opened 2018-12-19
Update paypal-rest-sdk to version 1.8.1

znarf:


znarf:
Should be safe looking at the changelog https://github.com/paypal/PayPal-node-SDK/blob/master/CHANGELOG.md

PR #1599, opened 2018-12-19
Update json2csv to version 4.3.2

znarf:


znarf:
Following up in https://github.com/opencollective/opencollective-api/pull/1661

PR #1602, opened 2018-12-19
Update nodemailer to version 4.7.0

znarf:


znarf:
Let's follow up in https://github.com/opencollective/opencollective-api/pull/1629

PR #1633, opened 2019-01-02
Remove deprecated endpoints (and related)

znarf:
@xdamman do you think that's still needed for what's left of `opencollective-website`?


xdamman:
Should be ok. But I'd definitely deploy this asap after merging and stay ready to undo the PR if needed.

PR #1637, opened 2019-01-03
Create env var to set CreatedByUserId in refund transactions

marcogbarcellos:
Today we have the refund transactions field `CreatedByUserId`  hardcoded as 18520( collective id of marcogbarcellos).

This PR creates the environment variable `UPDATER_COLLECTIVE_ID` and defaults it to @znarf collective id.


znarf:
Please don't put me as default. Better nothing (and fail in that case).

Why not use command line arguments instead of environment variables?

Are you sure it's correct to use the Collective.id? I think createdByUserId should be the User.id

marcogbarcellos:
ok, will fix it.

marcogbarcellos:
we could go with args as well, no big deal but i personally don't see a huge advantage if we compare to env vars so I'd keep the way it is. I may be wrong though, your call. let me know if you want me to update it.

znarf:
Can you throw an error if UPDATER_USER_ID is undefined/empty/invalid?

PR #1639, opened 2019-01-07
Refactor Postgres configuration

znarf:
- read first from `PG_URL`
- if not available, read from the variable defined by `PG_URL_ENVIRONMENT_VARIABLE`
- if not available, read from `DATABASE_URL`
- if not available, use the default from the environment:
  - circleci: `postgres://ubuntu@localhost:5432/opencollective_dvl`
  - docker: `postgres://opencollective@postgres:5432/opencollective_dvl`
  - test: `postgres://opencollective@127.0.0.1:5432/opencollective_test`
- if other environment: `postgres://opencollective@127.0.0.1:5432/opencollective_dvl` 
- after that, ability to override individual values with `PG_DATABASE`, `PG_USERNAME`, `PG_PASSWORD`, `PG_HOST`, `PG_PORT`


xdamman:
ENVIRONMENT_VARIABLE Is redundant in an all cap variable name (and it’s long)

I think that env variables should always take precedence. What are in files are fallback or default values.

znarf:
@xdamman `PG_URL_ENVIRONMENT_VARIABLE` is the environment variable to define the environment variable where to read the credentials.

Use case, we have 2 databases, and Heroku is populating automatically `HEROKU_POSTGRESQL_DOG_URL` and `HEROKU_POSTGRESQL_CAT_URL`. If we want to use the dog one, we would set `PG_URL_ENVIRONMENT_VARIABLE=HEROKU_POSTGRESQL_DOG_URL`, if we want to use the cat one, we would set `PG_URL_ENVIRONMENT_VARIABLE=HEROKU_POSTGRESQL_CAT_URL`.

I think that it is effectively verbose but properly named, feel free to propose alternatives!

Betree:
Looks ok to me. Is this `override` really necessary?

HipsterBrown:
So it appears we could also use `PG_URL` to get the url for the database the heroku app, either `opencollective-prod-api` or `opencollective-staging-api`, to do the same thing as the value for `PG_URL_ENVIRONMENT_VARIABLE`. 

Is `PG_URL_ENVIRONMENT_VARIABLE` used for the case of one app being able to switch between databases? 

It also appears we would never set `PG_URL_ENVIRONMENT_VARIABLE` ourselves in development, it's purely a heroku-specific name, so I think it is fine to keep as is.

HipsterBrown:
Before merging, I think it would be good to document how to use the `PG_URL` env variable to override the default settings, as well as how it used to copy staging or production databases locally. We should also mention always copying those databases rather than connecting to the live version for local development.

znarf:
@HipsterBrown Is `PG_URL` automatically populated by Heroku? If that's the case, I would simplify the whole thing.

znarf:
@Betree we need `override` to keep compatibility with all sort of scripts that are using it, have a look at `package.json` for a sample.

HipsterBrown:
@znarf The command `heroku config:get PG_URL -a opencollective-staging-api` appears to reutrn the same value as `heroku config:get ${PG_URL_ENVIRONMENT_VARIABLE} -a opencollective-staging-api`, so I believe it's auto-populated.

znarf:
`PG_URL` is not mentioned in the docs
https://devcenter.heroku.com/articles/heroku-postgresql

After some reconsideration, I still think the proposal in this PR is good.

PR #1652, opened 2019-01-09
Update Sequelize findById -> findByPk

znarf:
The latest Sequelize version deprecated `findById` and advise to use `findByPk` instead.


HipsterBrown:
Is this because the method is being deprecated?

PR #1653, opened 2019-01-09
Update Sequelize find -> findOne

znarf:
The latest Sequelize version deprecated `find` and advised to use `findOne` instead.


HipsterBrown:
Is this because the method is being deprecated?

PR #1662, opened 2019-01-12
fix for email.js

jollyburnz:
**Issue**
duplicate imports, and missing lodash.

**Solution**
removed duplicate, and added lodash.


znarf:
Thank you!

Currently `scripts` are not covered by `eslint` or test and this was introduced by a refactoring.

PR #1671, opened 2019-01-17
Add checks on order amount

Betree:
- Ensure total amount is >= 0
- Add checks on total amount when using a tier (see https://github.com/opencollective/opencollective/issues/1637)

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -318,6 +318,11 @@ export async function createOrder(order, loaders, remoteUser, reqIp) {
       throw new Error(`Invalid currency. Expected ${currency}.`);
     }
 
+    // ---- Checks on totalAmount ----
+    if (order.totalAmount < 0) {
+      throw new Error('Total amount cannot be a negative value');
+    }
+
     const quantity = order.quantity || 1;
 
     let totalAmount;
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -3,7 +3,7 @@ import uuidv4 from 'uuid/v4';
 import debugLib from 'debug';
 import md5 from 'md5';
 import Promise from 'bluebird';
-import { pick, omit, get } from 'lodash';
+import { pick, omit, get, min, isNil } from 'lodash';
 import config from 'config';
 
 import models from '../../../models';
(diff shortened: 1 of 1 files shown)


server/graphql/v1/mutations/orders.js
```diff
@@ -318,14 +321,27 @@ export async function createOrder(order, loaders, remoteUser, reqIp) {
...
+      throw new Error('Total amount cannot be a negative value');
+    }
 
-    let totalAmount;
-    if (tier && tier.amount && !tier.presets) {
-      // if the tier has presets, we can't enforce tier.amount
-      totalAmount = tier.amount * quantity;
-    } else {
-      totalAmount = order.totalAmount; // e.g. the donor tier doesn't set an amount
+    // Don't allow custom values if using a tier with fixed amount
```
znarf: @xdamman can you confirm we want that?

server/graphql/v1/mutations/orders.js
```diff
@@ -318,14 +321,27 @@ export async function createOrder(order, loaders, remoteUser, reqIp) {
...
+    if (tier && tier.amount && !tier.presets && tier.amount * order.quantity !== order.totalAmount) {
+      const prettyTotalAmount = formatCurrency(order.totalAmount, currency);
+      const prettyExpectedAmount = formatCurrency(tier.amount * order.quantity, currency);
+      throw new Error(
+        `This tier uses a fixed amount. Order total must be ${prettyExpectedAmount}. You set: ${prettyTotalAmount}`,
+      );
+    }
+
+    // If using a tier, amount can never be less than the minimum amount
```
znarf: @xdamman can you confirm we want that?

server/graphql/v1/mutations/orders.js
Betree: @xdamman This behaviour is better explained in https://github.com/opencollective/opencollective/issues/1637

Betree: Something to note on this one: we were already forcing the amount. The change is to throw an error instead of forcing it on the API, this should be enforced on frontend, not on the API.

The idea is that if the user send a total amount of $40, we should never bill more or less than this amount. If it doesn't match, then it's an error.

Betree: So this was breaking some tests, apparently parts of the code don't send the `totalAmount` if using a fixed tier. I've made a change to only throw an error if `totalAmount` is not null or undefined.

See https://github.com/opencollective/opencollective-api/pull/1671/commits/59a176d9d41e486c149bc7f06381e2577e87645b

xdamman: Yes I confirm.

PR #1676, opened 2019-01-18
Ability to precisely count active unique donators in Tiers

znarf:
Related Frontend issue: https://github.com/opencollective/opencollective-frontend/pull/1313

We would like to only display the active members for Tiers that are a subscription.

The displayed total should also be consistent with that.

### DIFF
--- a/server/graphql/loaders.js
+++ b/server/graphql/loaders.js
@@ -271,6 +271,25 @@ export const loaders = req => {
           .then(results => sortResults(ids, results, 'TierId'))
           .map(result => get(result, 'dataValues.count') || 0),
       ),
+      totalActiveDistinctOrders: new DataLoader(ids =>
+        models.Order.findAll({
+          attributes: [
+            'TierId',
+            [
+              sequelize.fn(
+                'COALESCE',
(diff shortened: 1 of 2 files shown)


server/graphql/loaders.js
```diff
@@ -271,6 +271,25 @@ export const loaders = req => {
           .then(results => sortResults(ids, results, 'TierId'))
           .map(result => get(result, 'dataValues.count') || 0),
       ),
+      totalActiveDistinctOrders: new DataLoader(ids =>
```
xdamman: `totalActiveDistinctOrders` is misleading. Not clear what a distinct order relates to. Every order is different.
I guess you meant `totalActiveOrdersFromDistinctCollectives` (or `countDistinctFromCollectives`)

(it's annoying to rename variables I know and picking the right variable name is the hardest thing in programming ¯\_(ツ)_/¯ )

znarf: I tried to make it consistent.

`totalActiveDistinctOrders` is inspired by `totalDistinctOrders`, which was itself introduce there: https://github.com/opencollective/opencollective-api/commit/6adfb4dc4180c55efc05562f8632d316853b6463

PR #1687, opened 2019-01-25
Update expense invalid link in collective expense approved email template

flickz:
This PR fix the invalid expense link pointed out in [#1675 ](https://github.com/opencollective/opencollective/issues/1675)

cc @Betree

### DIFF
--- a/templates/emails/collective.expense.approved.hbs
+++ b/templates/emails/collective.expense.approved.hbs
@@ -7,7 +7,7 @@ Subject: Your expense to {{collective.name}} for {{{currency expense.amount curr
 
 <center>
   <h1>{{{currency expense.amount currency=expense.currency precision=2}}}</h1>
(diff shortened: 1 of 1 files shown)


Betree:
Looks good to me too, there's a second link that needs to be fixed in the same way at line 10:

https://github.com/opencollective/opencollective-api/pull/1687/files#diff-c831a643d5338fda0d93a367bc5b8f43R10

flickz:
Fixed!

znarf:
@Betree Just wondering, why do we need that fix when the route is the following:
`/:parentCollectiveSlug?/:collectiveType(events)?/:collectiveSlug/expenses/:filter(categories|recipients)?/:value?`

Betree:
@znarf Previous links were pointing to `/expenses/____`. In the route you mention above, having the collective slug before `/expenses` is mandatory.

PR #1689, opened 2019-01-29
Add collective country ISO code

flickz:
This Fix [#1654](https://github.com/opencollective/opencollective/issues/1654).

- Adds new migration [file](https://github.com/flickz/opencollective-api/blob/d0817367822d4e942cbe7e2d21e3aba7ea1f3d14/migrations/20190129083947-add-collective-country_iso.js).
- Updates [Collective model](https://github.com/flickz/opencollective-api/blob/d0817367822d4e942cbe7e2d21e3aba7ea1f3d14/server/models/Collective.js#L228).
- Updates the [`CollectiveInputType`](https://github.com/flickz/opencollective-api/blob/d0817367822d4e942cbe7e2d21e3aba7ea1f3d14/server/graphql/v1/inputTypes.js#L102) and [`CollectiveAttributesInputType`](https://github.com/flickz/opencollective-api/blob/d0817367822d4e942cbe7e2d21e3aba7ea1f3d14/server/graphql/v1/inputTypes.js#L151).


migrations/20190129083947-add-collective-country_iso.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.addColumn('Collectives', 'countryISO', Sequelize.STRING).then(() => {
```
Betree: It's a detail but a good practice to only use the size needed. As we know that ISO 3166 defines country codes as two characters long, we could go with `Sequelize.STRING(32) `

server/models/Collective.js
```diff
@@ -225,6 +225,8 @@ export default function(Sequelize, DataTypes) {
 
       address: DataTypes.STRING,
 
+      countryISO: DataTypes.STRING,
```
Betree: Would be nice to have validations for `len: 2` and `isAlpha: true` on this one (or with a regex or any other trick). See http://docs.sequelizejs.com/manual/tutorial/models-definition.html#validations

server/models/Collective.js
```diff
@@ -225,6 +225,14 @@ export default function(Sequelize, DataTypes) {
 
       address: DataTypes.STRING,
 
+      countryISO: {
+        type: DataTypes.STRING,
+        validate: {
+          len: 2,
+          isISO31661Alpha2: true,
```
Betree: I can't find this one in the docs, any pointer to how it works?

migrations/20190129083947-add-collective-country_iso.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const size = 32;
```
Betree: Why `32`?

Betree: Oh sorry, I just saw that I put `32` in the example I sent you, my bad. 

The correct value is `Sequelize.STRING(2)` as we only need to chars.

server/models/Collective.js
flickz: 😆  Omg! Added this late last night, sequelize uses [validator](https://www.npmjs.com/package/validator) internally which has  `isISO31661Alpha2()` implemented I interestingly added it hoping to copy the [function](https://github.com/chriso/validator.js/blob/master/lib/isISO31661Alpha2.js) to a `util` file since it wouldn't make sense to install a new a package just to verify a single string.  I tested it and somehow worked. Good sleep is good lol, thought I copied it already. I will do just that asap.

Betree: I'm personally ok if you want to add `validator` as a dependency, it's a good library that we could use for many other validations. Also you want to keep this kind of things up-to-date: country codes can appear and disappear, it's good to rely on something that updates it  for us :slightly_smiling_face: 

Just be careful, we use fixed dependencies versions. You can do `npm install validator --save-exact` to achieve that.

flickz:
@Betree Good?

server/models/Collective.js
```diff
@@ -225,6 +226,16 @@ export default function(Sequelize, DataTypes) {
 
       address: DataTypes.STRING,
 
+      countryISO: {
+        type: DataTypes.STRING,
+        validate: {
+          len: 2,
+          isCountryISO(value) {
+            return isISO31661Alpha2(value);
```
Betree: I'm surprised by this: documentation state that the validate function is supposed to throw, not to return a boolean. Did you check that this was working properly?

znarf: If we have a doubt about it, would be great to back that with tests.

flickz: Apologies! :slightly_smiling_face:

PR #1697, opened 2019-02-06
Refactor: Abstract cache provider and add Redis support in caching layer

kewitz:
Fixes: https://github.com/opencollective/opencollective/issues/1715


kewitz:
Fixed the CircleCI bug.

server/lib/cacheProvider.js
```diff
@@ -0,0 +1,130 @@
...
+      return value !== undefined;
+    },
+    set: async (key, value, expirationInSeconds = 0) => {
+      if (value !== undefined) return client.set(key, serialize(value), { expires: expirationInSeconds });
+    },
+  };
+};
+
+const makeRedisProvider = ({ serverUrl, opts }, { serialize, unserialize }) => {
+  const client = redis.createClient(serverUrl, opts);
```
znarf: Because Bluebird is available, we'll better use `promisifyAll` for that.

package.json
```diff
@@ -95,6 +95,7 @@
     "prepend-http": "2.0.0",
     "query-string": "6.2.0",
     "querystring": "0.2.0",
+    "redis": "^2.8.0",
```
znarf: We use fixed version for core dependencies.

server/lib/cacheProvider.js
```diff
@@ -0,0 +1,130 @@
...
+            password: get(config, 'memcache.password'),
+            servers: get(config, 'memcache.servers'),
+            username: get(config, 'memcache.username'),
+          },
+          { serialize, unserialize },
+        ),
+      );
+    case PROVIDER_TYPES.REDIS:
+      debugCache('Redis configuration detected, using redis as cache backend.');
+      return unfailable(makeRedisProvider(has(config, 'redis.serverUrl'), { serialize, unserialize }));
```
znarf: `has(config, 'redis.serverUrl')` seems wrong. `get(config, 'redis')` ?

server/lib/cacheProvider.js
```diff
@@ -0,0 +1,130 @@
...
+  return {
+    clear: async () => lruCache.reset(),
+    del: async key => lruCache.del(key),
+    get: async key => lruCache.get(key),
+    has: async key => lruCache.has(key),
+    set: async (key, value, expirationInSeconds) => lruCache.set(key, value, expirationInSeconds * 1000),
+  };
+};
+
+const unfailable = provider =>
```
znarf: While we get it, we usually prefer simpler code.

server/lib/cacheProvider.js
```diff
@@ -0,0 +1,130 @@
...
+            logger.warn(`Error on cache.${key}: ${err.message}`);
+          }
+        };
+      } else return target[key];
+    },
+  });
+
+export const getProvider = (
+  provider = defaultProvider,
+  { serialize, unserialize } = { serialize: JSON.stringify, unserialize: JSON.parse },
```
znarf: We're using `serialize` / `unserialize` when using `get` and `set`, not at configuration time.

See the use case in `memoize`.

znarf:
Regarding code organization, it would be even clearer to split `memory`, `memcache` and `redis` in specific files.

znarf:
Closing. Following up in https://github.com/opencollective/opencollective-api/pull/1711

server/lib/cacheProvider.js
kewitz: Yes, you're absolutely right.
Since tests were passing, I imagine redis defaults to localhost with no passwords.

server/lib/cacheProvider.js
kewitz: Yeah, I agree 100%.
Proxy is the kind of thing you should only use in libraries for very specific reasons.

PR #1704, opened 2019-02-08
Add createCollectiveFromGithub to GraphQL mutation resolver

flickz:
This PR is needed for the Revamping of creating opensource collective from Gihub. [#1699](https://github.com/opencollective/opencollective/issues/1699).

It's basically refactors [`createFromGithub`](https://github.com/opencollective/opencollective-api/blob/master/server/controllers/collectives.js#L274) available in REST API to GraphQL.

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -13,6 +13,7 @@ import activities from '../../../constants/activities';
 import { types } from '../../../constants/collectives';
 
 const debugClaim = debug('claim');
+const debugGithub = debug('github');
 
 export async function createCollective(_, args, req) {
   if (!req.remoteUser) {
@@ -174,6 +175,74 @@ export async function createCollective(_, args, req) {
       },
     },
   });
+  return collective;
+}
+
+export async function createCollectiveFromGithub(_, args, req) {
+  if (!req.remoteUser) {
+    throw new errors.Unauthorized({
+      message: 'You need to be logged in to create a collective',
+    });
+  }
+
+  if (!args.collective.name) {
+    throw new errors.ValidationFailed({ message: 'collective.name required' });
+  }
+
+  let collective;
+  const collectiveData = { ...args.collective };
+  const user = await models.User.findByPk(req.remoteUser.id);
+  if (!user) {
+    return Promise.reject(new Error(`No user with Id ${req.remoteUser.id} found`));
+  }
+
+  const existingCollective = await models.Collective.findOne({
+    where: { slug: collectiveData.slug.toLowerCase() },
+  });
+
+  if (existingCollective) {
+    collectiveData.slug = `${collectiveData.slug}-${Math.floor(Math.random() * 1000 + 1)}`;
+  }
+
+  collectiveData.ParentCollectiveId = defaultHostCollective('opensource').ParentCollectiveId;
+  collectiveData.currency = 'USD';
+  collectiveData.CreatedByUserId = user.id;
+  collectiveData.LastEditedByUserId = user.id;
+
+  try {
+    collective = await models.Collective.create(collectiveData);
+  } catch (err) {
+    throw new Error(err.message);
+  }
+
+  debugGithub('createdCollective', collective && collective.dataValues);
+  const host = await models.Collective.findByPk(defaultHostCollective('opensource').CollectiveId);
+  const promises = [
+    collective.addUserWithRole(user, roles.ADMIN),
+    collective.addHost(host, user),
+    collective.update({ isActive: true }),
+  ];
+
+  await Promise.all(promises);
+
+  const data = {
+    firstName: user.firstName,
+    lastName: user.lastName,
+    collective: collective.info,
+  };
+  debug('sending github.signup to', user.email, 'with data', data);
+  await emailLib.send('github.signup', user.email, data);
+  models.Activity.create({
+    type: activities.COLLECTIVE_CREATED,
+    UserId: user.id,
+    CollectiveId: collective.id,
+    data: {
+      collective: collective.info,
+      host: host.info,
+      user: user.info,
+    },
+  });
+
   return collective;
 }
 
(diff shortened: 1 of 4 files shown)


server/graphql/v1/mutations/collectives.js
```diff
@@ -177,6 +178,74 @@ export async function createCollective(_, args, req) {
...
+    });
+  }
+
+  if (!args.collective.name) {
+    throw new errors.ValidationFailed({ message: 'collective.name required' });
+  }
+
+  let collective;
+  const collectiveData = { ...args.collective };
+  const user = await models.User.findByPk(req.remoteUser.id);
```
znarf: Isn't `req.remoteUser` already an user?

server/graphql/v1/mutations/collectives.js
```diff
@@ -177,6 +178,74 @@ export async function createCollective(_, args, req) {
...
+  ];
+
+  await Promise.all(promises);
+
+  const data = {
+    firstName: user.firstName,
+    lastName: user.lastName,
+    collective: collective.info,
+  };
+  debug('sending github.signup to', user.email, 'with data', data);
```
znarf: `debug` is not bound to a namespace.

server/graphql/v1/mutations/collectives.js
```diff
@@ -177,6 +178,74 @@ export async function createCollective(_, args, req) {
...
+  if (!req.remoteUser) {
+    throw new errors.Unauthorized({
+      message: 'You need to be logged in to create a collective',
+    });
+  }
+
+  if (!args.collective.name) {
+    throw new errors.ValidationFailed({ message: 'collective.name required' });
+  }
+
```
znarf: TODO: we should do some extra validation of the GitHub repo/organization.

- check 100 stars
- validate that the user is an admin of the repo/organization
- check if the repo/organization is already registered as a collective

server/graphql/v1/mutations/collectives.js
```diff
@@ -177,6 +178,74 @@ export async function createCollective(_, args, req) {
...
+  const user = await models.User.findByPk(req.remoteUser.id);
+  if (!user) {
+    return Promise.reject(new Error(`No user with Id ${req.remoteUser.id} found`));
+  }
+
+  const existingCollective = await models.Collective.findOne({
+    where: { slug: collectiveData.slug.toLowerCase() },
+  });
+
+  if (existingCollective) {
```
znarf: Maybe we should just fail and say that the collective slug is already taken. Let the user try again with another. Later, we could add some early validation in the frontend to avoid failing so late.

server/graphql/v1/mutations/collectives.js
```diff
@@ -177,6 +178,74 @@ export async function createCollective(_, args, req) {
...
+      message: 'You need to be logged in to create a collective',
+    });
+  }
+
+  if (!args.collective.name) {
+    throw new errors.ValidationFailed({ message: 'collective.name required' });
+  }
+
+  let collective;
+  const collectiveData = { ...args.collective };
```
znarf: There is a lot of extra functionnality that needs to be ported from here:
https://github.com/opencollective/opencollective-website/blob/04d62f7047cb4aa25946ab226840b6203a16a027/frontend/src/containers/OnBoarding.js#L108

- default tiers
- repository settings
- repository website
- repository tags

server/graphql/v1/mutations/collectives.js
flickz: Yeah it is the same thing, just trying too be a little defensive.

server/graphql/v1/mutations/collectives.js
flickz: Fixed that, thanks.

server/graphql/v1/mutations/collectives.js
znarf: We don't have to here. We should be confident in `req.remoteUser`.

server/graphql/v1/mutations/collectives.js
flickz: - Checking stars already happens through filtering the repositories returned by GitHub.

- Currently the best way to know if a user is an admin is by fetching individual repo information, Github does not provide that information in list of public repos API currently being used. I'm still thinking about the best way to do this. While minding network round trips. 

- Yes this is important because it is included in the design, I will work on that.

server/graphql/v1/mutations/collectives.js
flickz: Yeah, that is how it is currently done in `createCollective` resolver. Adding that to TODO.

server/graphql/v1/mutations/collectives.js
znarf: Technically the "100 stars" is currently checked on the frontend, so we can't trust it. That, and checking if the user is an admin of the repo/organization should be done again in the backend.

Good news, we have similar behavior elsewhere that we can copy!

Here, we check that the user is an admin:
https://github.com/opencollective/opencollective-api/blob/master/server/graphql/v1/mutations/collectives.js#L406

Here, we check the stars:
https://github.com/opencollective/opencollective-api/blob/master/server/graphql/v1/mutations/orders.js#L153

flickz: > Here, we check that the user is an admin:

Works well for a single repository fetch, it has `{permission: {admin: true} }` included in the repository data returned, in our case we are fetching all public repositories the data returned has no `permission` object included to validate `isAdmin`, the question is how do we do this for public repositories's list?:
Suggestions I'm thinking:
- Fetch all repositories, filter based on their stars and for each filtered repository fetch their `permission` data individually. The only problem is the user will have to wait for each repo in the array, before the `GithubRepositories` component is rendered. Time here is directly proportional to numbers of filtered repositories.

- My second suggestion is fetching the permission data only when a user selects a repo in `GithubRepositories` component and notify them, that, they don't have permission to create collective with the currently selected repo.

server/graphql/v1/mutations/collectives.js
flickz: I will add this tonight.

server/graphql/v1/mutations/collectives.js
znarf: If we’re checking that in the GraphQL mutation, we can do a single repository fetch, like in the examples I shared.

PR #1711, opened 2019-02-11
Abstract cache provider and add Redis support in caching layer

znarf:
Follow up on https://github.com/opencollective/opencollective-api/pull/1697

- split providers in separated files
- use Bluebird to promisify redis
- fix implementation of serialize / unserialize
- fix Redis configuration
- default configuration in REDIS_URL

Thank you @kewitz!


server/lib/cache/index.js
```diff
@@ -0,0 +1,139 @@
...
+    return PROVIDER_TYPES.MEMCACHE;
+  } else {
+    return PROVIDER_TYPES.MEMORY;
+  }
+};
+
+let defaultProvider;
+
+const getDefaultProvider = () => {
+  const defaultProviderType = getDefaultProviderType();
```
kewitz: I would put this inside the conditional.

PR #1717, opened 2019-02-12
Extend Github handle and allow URLs

Betree:
Fix https://github.com/opencollective/opencollective/issues/1700

---

- Extends the `githubHandle` column size to 255 characters
- Allow for repositories: something like `opencollective/opencollective-api` is now considered valid
- Allow for an URL to be given. It will be formatted before insert to the DB.

I've also applied the URL feature to `twitterHandle`, so pasting an URL like https://twitter.com/opencollective in the form will now result in the value `opencollective` being stored for Twitter handle.
Before that, we would have stored the entire URL and ended up with an incorrect link on the frontend.

### DIFF
--- a/migrations/20190212154818-extend-github-handle.js
+++ b/migrations/20190212154818-extend-github-handle.js
@@ -1,0 +1,28 @@
+'use strict';
+
+const githubHandleColumn = 'githubHandle';
+
+module.exports = {
+  /**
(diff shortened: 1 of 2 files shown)


znarf:
Can we add in the migration an automated conversion of URLs in the `twitterHandle` or `githubHandle` column?

Betree:
@znarf Done

PR #1736, opened 2019-02-20
CircleCI: Split tests in separate jobs

Betree:


.circleci/config.yml
```diff
@@ -1,41 +1,93 @@
 version: 2
+
+aliases:
+  restore_cache: &restore_cache
+    restore_cache:
+      name: Restore node_modules cache
+      keys:
+        - npm-cache-{{ checksum "package-lock.json" }}
```
znarf: If you add `-npm-cache-` this can boost results even if package-lock.json is the same.

.circleci/config.yml
```diff
@@ -1,41 +1,93 @@
...
+
+  persist_cache: &persist_cache
+    save_cache:
+      name: Save node modules cache
+      key: npm-cache-{{ checksum "package-lock.json" }}
+      paths:
+        - node_modules
```
znarf: My understanding is that there is no need to cache node_modules when using `npm ci`

.circleci/config.yml
```diff
@@ -1,41 +1,93 @@
...
-          key: dependency-cache-{{ checksum "package.json" }}
-          paths:
-            - ./node_modules
-
-      - run: npm run test
+      - <<: *persist_cache
+      - run: npm run test:unit
+      - run: npm run coverage
       - store_test_results:
           path: /tmp/circleci-test-results
```
znarf: Is `store_test_results:` doing anything useful?

.circleci/config.yml
```diff
@@ -1,41 +1,93 @@
...
+workflows:
+  version: 2
+  main:
+    jobs:
+      - lint
+      - prettier
+      - test-bdd
+      - test-unit:
+          filters:
+            branches:
```
znarf: Like that, the filters are only on the test-unit job.

.circleci/config.yml
Betree: It is used to generate the "Insights" view on CircleCI that gives details and statistics about tests: https://circleci.com/build-insights/gh/opencollective/opencollective-api/master

image

.circleci/config.yml
Betree: You're right, documentation states:

> If a node_modules is already present, it will be automatically removed before npm ci begins its install.

.circleci/config.yml
Betree: See https://circleci.com/docs/2.0/collect-test-data/

PR #1754, opened 2019-02-28
Remove support for sessions

znarf:
I had a look at that and figured this was being unused. If that's really the case, great to remove it.

The session table is absolutely empty, which is a good validation.


xdamman:
This will also help with caching as it won't add a useless cookie to track a session 👌

PR #1758, opened 2019-03-01
Update Stripe

kewitz:
Stripe is no longer using Bluebird's promises, which means we no longer have `tap` in available in those chains.


znarf:
Thank you @kewitz. Followed up in https://github.com/opencollective/opencollective-api/pull/1760 with an extra commit.

PR #1760, opened 2019-03-02
Update Stripe (updated with nock update)

znarf:
Follow up on https://github.com/opencollective/opencollective-api/pull/1758


test/graphql.matchingFund.nock.js
```diff
@@ -3,17 +3,14 @@ import nock from 'nock';
 
 export default function() {
   nock('https://api.stripe.com:443', { encodedQueryParams: true })
-    .post(
-      '/v1/tokens',
-      'card%5Bnumber%5D=4242424242424242&card%5Bexp_month%5D=12&card%5Bexp_year%5D=2028&card%5Bcvc%5D=222',
-    )
+    .post('/v1/tokens', 'card[number]=4242424242424242&card[exp_month]=12&card[exp_year]=2028&card[cvc]=222')
```
kewitz: Nice!

test/graphql.matchingFund.nock.js
```diff
@@ -118,84 +176,90 @@ export default function() {
...
+        'req_bRBIfM5O0cLdlB',
         'Stripe-Version',
         '2015-04-07',
         'Strict-Transport-Security',
         'max-age=31556926; includeSubDomains; preload',
       ],
     );
 
-  nock('http://data.fixer.io', { encodedQueryParams: true })
+  nock('http://data.fixer.io:80', { encodedQueryParams: true })
```
kewitz: Does this make sense? Which library is calling this?

test/graphql.matchingFund.test.js
```diff
@@ -174,7 +174,7 @@ describe('graphql.matchingFund.test.js', () => {
     expect(transactions[3].amount).to.equal(user1.paymentMethod.matching * order.totalAmount);
 
     const balance = await user1.paymentMethod.getBalanceForUser(user2);
-    expect(balance.amount).to.equal(141621); // €1,500 - $100
+    expect(balance.amount).to.equal(141226); // €1,500 - $100
```
kewitz: Why have these changed? Different trade rate from the time the tests were first written?

znarf: Yes, I think so.

test/graphql.matchingFund.nock.js
znarf: This is used in `server/lib/currency.js`, looks like some internal changed, but the semantic is the same.

kewitz:
:shipit:

PR #1767, opened 2019-03-05
Add unsubscribe link to "thankyou" email

flickz:
While we sort for the best possible way to test this fix, reviewing the code might help decide how to go about testing it properly.

Ref [#1502](https://github.com/opencollective/opencollective/issues/1502)

### DIFF
--- a/server/lib/email.js
+++ b/server/lib/email.js
@@ -312,12 +312,10 @@ const generateEmailFromTemplate = (template, recipient, data = {}, options = {})
 };
 
 const isNotificationActive = async (template, data) => {
-  // Skip active email notification check for unit test
-  if (config.env === 'test' || config.env === 'circleci') {
-    logger.info('Skips active email notification checks for unit test');
-    return true;
+  if (data.user && data.user.id) {
+    return models.Notification.isActive(template, data.user.id, data.collective);
   } else {
-    return models.Notification.isActive(template, data.collective.id, data.user.id);
+    return true;
   }
 };
 
--- a/server/lib/email.js
+++ b/server/lib/email.js
@@ -320,17 +320,18 @@ const generateEmailFromTemplateAndSend = async (template, recipient, data, optio
     return;
   }
 
-  // Unit test like collective.model.test.js don't care about sending
-  // the email only making sure the function was called
-  // hence data.collective or data.user are usually an empty object
-  // we skip the notification active check in this case.
-  if (process.env.NODE_ENV === 'production' && data.collective && data.user) {
-    const notificationIsActive = await models.Notification.isActive(template, data.collective.id, data.user.id);
+  // A temporary solution to enable unit test that only cares about the function being
+  // called to pass.
+  if (config.env === 'test' || config.env === 'circleci') {
+    const user = { id: 1 };
+    data.user = user;
+    data.collective.id = 1;
+  }
 
-    if (!notificationIsActive) {
-      logger.info(`Email with template '${template}' not sent. Recipient email notification not active.`);
-      return;
-    }
+  const notificationIsActive = await models.Notification.isActive(template, data.collective.id, data.user.id);
+  if (!notificationIsActive) {
+    logger.info(`Email with template '${template}' not sent. Recipient email notification not active.`);
+    return;
   }
 
   return generateEmailFromTemplate(template, recipient, data, options).then(renderedTemplate => {
--- a/server/lib/email.js
+++ b/server/lib/email.js
@@ -317,11 +317,7 @@ const isNotificationActive = async (template, data) => {
     logger.info('Skips active email notification checks for unit test');
     return true;
   } else {
-    const notificationIsActive = await models.Notification.isActive(template, data.collective.id, data.user.id);
-    if (!notificationIsActive) {
-      return false;
-    }
-    return true;
+    return models.Notification.isActive(template, data.collective.id, data.user.id);
   }
 };
 
--- a/server/lib/email.js
+++ b/server/lib/email.js
@@ -320,11 +320,17 @@ const generateEmailFromTemplateAndSend = async (template, recipient, data, optio
     return;
   }
 
-  const notificationIsActive = await models.Notification.isActive(template, data.collective.id, data.user.id);
+  // Unit test like collective.model.test.js don't care about sending
+  // the email only making sure the function was called
+  // hence data.collective or data.user are usually an empty object
+  // we skip the notification active check in this case.
+  if (process.env.NODE_ENV === 'production' && data.collective && data.user) {
+    const notificationIsActive = await models.Notification.isActive(template, data.collective.id, data.user.id);
 
-  if (!notificationIsActive) {
-    logger.info(`Email with template '${template}' not sent. Recipient email notification not active.`);
-    return;
+    if (!notificationIsActive) {
+      logger.info(`Email with template '${template}' not sent. Recipient email notification not active.`);
+      return;
+    }
   }
 
   return generateEmailFromTemplate(template, recipient, data, options).then(renderedTemplate => {
--- a/server/lib/email.js
+++ b/server/lib/email.js
@@ -223,6 +223,7 @@ const getNotificationLabel = (template, recipients) => {
     onboarding: 'onboarding emails',
     'user.monthlyreport': 'monthly reports for backers',
     'user.yearlyreport': 'yearly reports',
+    thankyou: 'thank you for your donation',
   };
 
   return notificationTypeLabels[template];
--- a/server/lib/email.js
+++ b/server/lib/email.js
@@ -320,18 +320,15 @@ const generateEmailFromTemplateAndSend = async (template, recipient, data, optio
     return;
   }
 
-  // A temporary solution to enable unit test that only cares about the function being
-  // called to pass.
+  // Skip active email notification check for unit test
   if (config.env === 'test' || config.env === 'circleci') {
-    const user = { id: 1 };
-    data.user = user;
-    data.collective.id = 1;
-  }
-
-  const notificationIsActive = await models.Notification.isActive(template, data.collective.id, data.user.id);
-  if (!notificationIsActive) {
-    logger.info(`Email with template '${template}' not sent. Recipient email notification not active.`);
-    return;
+    logger.info('Skips active email notification checks for unit test');
+  } else {
+    const notificationIsActive = await models.Notification.isActive(template, data.collective.id, data.user.id);
+    if (!notificationIsActive) {
+      logger.info(`Email with template '${template}' not sent. Recipient email notification not active.`);
+      return;
+    }
   }
 
   return generateEmailFromTemplate(template, recipient, data, options).then(renderedTemplate => {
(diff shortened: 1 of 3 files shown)


server/models/Notification.js
```diff
@@ -132,6 +132,28 @@ export default function(Sequelize, DataTypes) {
...
+  Notification.isActive = (notificationType, CollectiveId, UserId) => {
+    debug('isActive', notificationType, CollectiveId, UserId);
+    return models.Notification.findOne({
+      where: {
+        CollectiveId,
+        type: notificationType,
+        UserId,
+      },
+    }).then(notification => {
+      if (notification && notification.active) {
```
znarf: If I'm not mistaken, an absence of notification entry in the table should be considered active. Right?

server/lib/subscriptions.js
```diff
@@ -277,24 +277,30 @@ export async function sendThankYouEmail(order, transaction) {
...
-      subscriptionsLink: user.generateLoginLink(`/${order.fromCollective.slug}/subscriptions`),
-    },
-    {
-      from: `${order.collective.name} <hello@${order.collective.slug}.opencollective.com>`,
-    },
-  );
+  const notificationIsActive = await models.Notification.isActive('thankyou', order.collective.id, user.id);
+
+  // check if the user email notification is active before sending thankyou email.
+  if (notificationIsActive) {
```
znarf: I would personally write that ```if (!notificationIsActive) {
  return;
```

Keep the current indentation, minimizing changes. But it's up to you.

File: server/lib/subscriptions.js [hunk 3]
```diff
@@ -277,24 +277,30 @@ export async function sendThankYouEmail(order, transaction) {
...
-      recommendedCollectives,
-      config: { host: config.host },
-      interval: order.Subscription.interval,
-      subscriptionsLink: user.generateLoginLink(`/${order.fromCollective.slug}/subscriptions`),
-    },
-    {
-      from: `${order.collective.name} <hello@${order.collective.slug}.opencollective.com>`,
-    },
-  );
+  const notificationIsActive = await models.Notification.isActive('thankyou', order.collective.id, user.id);
```
znarf: Have you thought about doing that in the email library (server/lib/email.js) directly, so that we have it available for other templates?

server/models/Notification.js
flickz: No, not active.

znarf: It should be active by default I think.

server/lib/subscriptions.js
flickz: Cool but I most times prefer handling the common condition first.

server/lib/subscriptions.js
flickz: Not really but I think It's important to do that now to fix the failing unit test in `lib.subscriptions.test.js`

znarf: Let's do that.

server/models/Notification.js
```diff
@@ -132,6 +132,28 @@ export default function(Sequelize, DataTypes) {
...
+    debug('isActive', notificationType, CollectiveId, UserId);
+    return models.Notification.findOne({
+      where: {
+        CollectiveId,
+        type: notificationType,
+        UserId,
+      },
+    }).then(notification => {
+      if (notification) {
+        return notification.active;
```
znarf: Is that properly returned as a boolean by the model?

znarf:
@flickz Can you ensure tests are passing?

server/models/Notification.js
flickz: Yeah, I think so, the active field  `sequelize.Type` is set to boolean.

server/lib/email.js
```diff
@@ -320,11 +320,17 @@ const generateEmailFromTemplateAndSend = async (template, recipient, data, optio
...
-  const notificationIsActive = await models.Notification.isActive(template, data.collective.id, data.user.id);
-
-  if (!notificationIsActive) {
-    logger.info(`Email with template '${template}' not sent. Recipient email notification not active.`);
-    return;
+  // Unit test like collective.model.test.js don't care about sending
+  // the email only making sure the function was called
+  // hence data.collective or data.user are usually an empty object
+  // we skip the notification active check in this case.
+  if (process.env.NODE_ENV === 'production' && data.collective && data.user) {
```
znarf: I would not rely on `data.collective`  and `data.user`.

I would just check `config.env === 'circleci` || config.env === 'test'. If it's not precise enough, I would make sure unit tests can be properly detected.

`config.env` is preferred to `process.env.NODE_ENV` but is usually the same value.

flickz: Thinking of using this condition, only a single test is failing the notification active check because of the incomplete data structure.
```js
if (config.env === 'test' || config.env === 'circleci') {
    const user = { id: 1 };
    data.user = user;
    data.collective.id = 1;
  }
```
But I'm not completely comfortable with it.

znarf: @flickz please commit what you think is best now, and we'll discuss it from there.

flickz: @znarf The best possible way I think right now is to skip the unit test for active notification check.

server/lib/email.js
```diff
@@ -312,11 +314,23 @@ const generateEmailFromTemplate = (template, recipient, data = {}, options = {})
...
  * Given a template, recipient and data, generates email and sends it.
  */
-const generateEmailFromTemplateAndSend = (template, recipient, data, options = {}) => {
+const generateEmailFromTemplateAndSend = async (template, recipient, data, options = {}) => {
   if (!recipient) {
     logger.info(`Email with template '${template}' not sent. No recipient.`);
     return;
   }
+
+  // Skip active email notification check for unit test
```
znarf: Let's maybe extract that in a function (ie: `isNotificationActive(template, data.collective.id, data.user.id)` returning a boolean.

flickz:
We still need to test this fix manually, probably in staging to make sure:
- "Thankyou email" has an unsubscribe link.
- Users without `active` email notification field as `true` don't get email.

server/lib/email.js
```diff
@@ -311,6 +311,20 @@ const generateEmailFromTemplate = (template, recipient, data = {}, options = {})
   return Promise.resolve(renderedTemplate);
 };
 
+const isNotificationActive = async (template, data) => {
+  // Skip active email notification check for unit test
+  if (config.env === 'test' || config.env === 'circleci') {
+    logger.info('Skips active email notification checks for unit test');
+    return true;
+  } else {
+    const notificationIsActive = await models.Notification.isActive(template, data.collective.id, data.user.id);
```
znarf: Why not just return that?

flickz: Oh my bad.

server/lib/email.js
```diff
@@ -309,14 +311,31 @@ const generateEmailFromTemplate = (template, recipient, data = {}, options = {})
   return Promise.resolve(renderedTemplate);
 };
 
+const isNotificationActive = async (template, data) => {
+  // Skip active email notification check for unit test
+  if (config.env === 'test' || config.env === 'circleci') {
+    logger.info('Skips active email notification checks for unit test');
+    return true;
+  } else {
+    return models.Notification.isActive(template, data.collective.id, data.user.id);
```
znarf: What happens if `data.collective` or `data.user` are null/undefined, can that happen?

server/models/Notification.js
```diff
@@ -132,6 +132,28 @@ export default function(Sequelize, DataTypes) {
...
+   * @param {*} CollectiveId
+   * @param {*} UserId
+   */
+  Notification.isActive = (notificationType, CollectiveId, UserId) => {
+    debug('isActive', notificationType, CollectiveId, UserId);
+    return models.Notification.findOne({
+      where: {
+        CollectiveId,
+        type: notificationType,
+        UserId,
```
znarf: I don't think `UserId` is necessary here, `CollectiveId` is enough.

server/lib/email.js
flickz: I don't expect that to happen but we probably need to be defensive with that, it happened with unit test but that was with a mocked data. I need suggestion on how to handle that, in case that happens, what do you think?

server/models/Notification.js
flickz: I think it is necessary to identify if the recipient of the email wants it, remember it's a "thankyou" message to a particular user.

znarf: I'll double check.

server/models/Notification.js
```diff
@@ -171,7 +193,7 @@ Types:
       data: collective, user (caller), target (the updated user), collectiveuser (updated values)
       2* Userid: the updated user + the caller
   - collective.user.deleted
-      data: collective, user (caller), target (the deleted user)
+      data: collective, user (caller), target (the  deleted user)
       2* Userid: the deleted user + the caller
```
Betree: Little unnecessary space here

server/models/Notification.js
znarf: Ok, so the un-subscription is hyper granular, it will only unsubscribe from future "thankyou" emails related to a given collective.

Example:  if I get a "Thank you for giving to Babel",  and click the unsubscribe link in the email, I will be unsubscribed from future  "Thank you for giving to Babel", but not "Thank you for giving to Webpack".

In that regard, `UserId` and `CollectiveId` usage is effectively fine.

server/lib/email.js
znarf: @flickz There are notifications where `data.collective` is empty such as `user.monthlyreport`. This needs to be handled.

Let's be defensive on `data.user` and send if it's missing, without checking in the database.

flickz: Alright noted.

server/lib/email.js
```diff
@@ -312,12 +312,10 @@ const generateEmailFromTemplate = (template, recipient, data = {}, options = {})
 };
 
 const isNotificationActive = async (template, data) => {
-  // Skip active email notification check for unit test
-  if (config.env === 'test' || config.env === 'circleci') {
-    logger.info('Skips active email notification checks for unit test');
-    return true;
-  } else {
+  if (data.collective && data.collective.id && (data.user && data.user.id)) {
```
znarf: The parenthesis looks unnecessary.

znarf: That's not correct, `data.collective` is optional.

server/models/Notification.js
```diff
@@ -132,6 +132,31 @@ export default function(Sequelize, DataTypes) {
       },
     }).then(us => us.map(us => us.UserId));
   };
+  /**
+   * Check if notification with `NotificationType` and `UserId` is active.
+   * @param {*} CotificationType
+   * @param {*} UserId
+   * @param {*} [Collective]
+   */
+  Notification.isActive = (NotificationType, UserId, Collective) => {
```
znarf: It's strange to have `UserId` (an id) in one case and `Collective` (an object) in the other. Can you reconsider that?

flickz: Yeah, make both object.

server/models/Notification.js
xdamman: Does that mean that we don't send a notification unless the user has an active `Notification`?

We did that at the beginning but then we realized that it was much better to do the opposite.
Always send a notification *unless* the user explicitly asked not to.

Since you are returning true if there is no row, it might still be the case. So I think this is good. Just wanted to make sure the purpose of the `Notifications` table is well understood (the `active` column should in fact always be `false`).

znarf: @xdamman I think we got it right and are on the same page.

PR #1776, opened 2019-03-08
Enforce models validations

Betree:
This PR is linked to the recent issue we encountered with react-hooks https://github.com/opencollective/opencollective/pull/1790

Proper validations on model would have limited the impact and prevented the insertion of corrupted data into the database.

I've added the validations while looking closely to production's data to limit regressions as much as possible.

I found two users, `jessIenewell` and `AB`, who had uppercase letters in their names. This is not currently supported and users profiles used to return 404s.
I have migrated them manually.


test/paymentMethods.opencollective.prepaid.test.js
```diff
@@ -16,7 +16,7 @@ describe('paymentMethods.opencollective.prepaid', () => {
 
   before(async () => {
     hostCollective = await models.Collective.create({
-      type: 'HOST',
+      type: 'ORGANIZATION',
```
Betree: The type we were using here, `HOST`, is not a valid collective type

test/graphql.createOrder.test.js
```diff
@@ -184,7 +184,7 @@ describe('createOrder', () => {
 
   it('creates a pending order if the collective is active and the payment method type is manual', async () => {
     const host = await models.Collective.create({
-      slug: 'host',
+      slug: 'host-collective',
```
Betree: We cannot use `host` as a collective slug anymore in tests as it is a reserved keyword

PR #1782, opened 2019-03-11
Archive Collective

flickz:
This PR contains API code archive/deactivate collective, as explained in issue [#1317](https://github.com/opencollective/opencollective/issues/1317).
API Checklist:

- [x] Add a archivedAt column to the collective table
- [x]  Add a GraphQL mutation to archive a collective.
- [x] Add a property to the GraphQL collective object to know if a collective had been archived
- [x] Check that the authenticated user is an admin
- [x] Delete all tiers in the collective
- [x] Cancel all subscriptions to the collective
- [x]  Update `deactivatedAt=now()` & `isActive=false`


server/graphql/v1/mutations/collectives.js
```diff
@@ -631,3 +632,42 @@ export async function claimCollective(_, args, req) {
...
+
+  if (!req.remoteUser.isAdmin(collective.id)) {
+    throw new errors.Unauthorized({
+      message: 'You need to be logged in as a core contributor.',
+    });
+  }
+
+  let tiers = await collective.getTiers();
+  tiers = tiers.map(async tier => {
+    return await tier.destroy();
```
znarf: This is confusing:

- don't transform the `tiers` variable into an array of delete promises
- the use of await doesn't make sense to me, it's inside an async function with a single instruction

server/graphql/v1/mutations/collectives.js
```diff
@@ -631,3 +632,42 @@ export async function claimCollective(_, args, req) {
...
+  });
+
+  // Cancel all active subscription
+  let orders = await collective.getIncomingOrders({
+    where: { status: status.ACTIVE, [Op.and]: { status: status.PENDING } },
+    include: [{ model: models.Subscription }, { model: models.Collective, as: 'collective' }],
+  });
+
+  orders = orders.map(async order => {
```
znarf: This is confusing:

- don't transform the orders variable into an array of promises, let's keep it an array of orders. if necessary, create another variable.

server/graphql/v1/mutations/collectives.js
```diff
@@ -631,3 +632,42 @@ export async function claimCollective(_, args, req) {
...
+    return await Promise.all([order.update({ status: status.CANCELLED }), order.Subscription.deactivate()]);
+  });
+
+  await Promise.all([...orders, ...tiers]);
```
znarf: What is the concurrency strategy? Absolutely everything in parallel? Have you considered doing it sequentially? If we do it in parallel, maybe it's better having limits.

server/graphql/v1/mutations/collectives.js
```diff
@@ -631,3 +632,42 @@ export async function claimCollective(_, args, req) {
...
+  const collective = await models.Collective.findByPk(args.id);
+  if (!collective) {
+    throw new errors.NotFound({
+      message: `Collective with id ${args.id} not found`,
+    });
+  }
+
+  if (!req.remoteUser.isAdmin(collective.id)) {
+    throw new errors.Unauthorized({
+      message: 'You need to be logged in as a core contributor.',
```
znarf: The message is misleading.

PR #1804, opened 2019-03-19
Fix PayPal balance calculation

Betree:
See https://github.com/opencollective/opencollective/issues/1596 for issue root cause.

### DIFF
--- a/server/paymentProviders/paypal/adaptive.js
+++ b/server/paymentProviders/paypal/adaptive.js
@@ -97,7 +97,10 @@ export default {
             paymentMethod.currency,
           )} across ${totalTransactions} transactions between ${firstTransactionAt} and ${lastTransactionAt}`,
         );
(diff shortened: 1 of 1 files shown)


server/paymentProviders/paypal/adaptive.js
```diff
@@ -97,7 +97,9 @@ export default {
             paymentMethod.currency,
           )} across ${totalTransactions} transactions between ${firstTransactionAt} and ${lastTransactionAt}`,
         );
-        return 200000 + totalSpent; // total spent has a negative value (in cents)
+        // total spent has a negative value (in cents)
+        const initialBalance = paymentMethod.data.balance || 200000;
```
znarf: What's `200000`?

Betree: It's the max amount for PayPal pre-approval. I've just moved it to a `paypalPreApprovalLimit` variable to make things clearer.

PR #1806, opened 2019-03-19
chore(dbdump): update default opencollective_dvl.pgsql with latest schema

xdamman:
It looks like someone forgot to run `./scripts/db_migrate.sh` after changing the schema of the db.


znarf:
@xdamman While I have nothing against updating the dump regularly, I don't see why it's breaking. If we follow the steps in the Readme:
 - for a fresh restore, the dump should be automatically migrated.
 - for an existing database, this is not helping anyway

I'm currently reviewing all that aspects in my PR, this is why I'm interested in the topic. https://github.com/opencollective/opencollective-api/pull/1800

xdamman:
How is the dump automatically migrated on a fresh restore?

It was breaking because the db already existed locally and wasn’t migrated. Then I used the restore script using the dbdump which wasn’t migrated. So I guess I didn’t use the right command to restore my local opencollective_dvl db?

znarf:
@xdamman current, I would expect a `dropdb opencollective_dvl` followed by a `npm postinstall` but maybe it's because I know it's taking care of the migration.

In my PR, I'm introducing `npm run db:restore`, maybe this should take care of the migration automatically.

PR #1805, opened 2019-03-19
Archive Collective: unarchive

flickz:
This PR continues the archive collective process by adding `unarchiveCollective` resolver to collective mutation.

Ref ([#1826](https://github.com/opencollective/opencollective/issues/1826))

PR #1812, opened 2019-03-20
Archive Collective: Remove "delete tiers" from archiveCollective

flickz:
Currently we are deleting tiers in [`archiveCollective`](https://github.com/opencollective/opencollective-api/blob/master/server/graphql/v1/mutations/collectives.js#L661) mutation resolver, this PR removes that as explained in ([#1824](https://github.com/opencollective/opencollective/issues/1824))

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -656,16 +656,9 @@ export async function archiveCollective(_, args, req) {
   }
 
   return collective
-    .getTiers()
-    .then(tiers => {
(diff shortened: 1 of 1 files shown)


znarf:
@flickz this is containing commits to from https://github.com/opencollective/opencollective-api/pull/1805

Would be great if you could isolate the commits so that PRs can be merged independently.

flickz:
@znarf Oh let me fix that.

flickz:
Fixed

PR #1814, opened 2019-03-20
Archive Collective: delay subscription cancellation

flickz:
This PR cancels subscription when executing daily charge subscription for an archived collective and notifies the user through email using a new `archived.collective.hbs` email template.

Ref ([#1825](https://github.com/opencollective/opencollective/issues/1825))


piamancini:
Suggested changes: 

> Archived Collective: Subscription cancelled to APEX

Archived Collective: Your subscription to APEX has been cancelled. 

> 
> Body:
> Dear Xavier Damman,
> APEX has been archived, we won't be able to process your order and your subscription has now been cancelled.
> If you have questions, you can always email us at support@opencollective.com.
> Warmly,
> – Team Open Collective

Dear Xavier Damman, 

The APEX collective has been archived by its core contributors and it is no longer accepting contributions. We are unable to process your order and your subscription has been cancelled. 

If you have any questions, you can always email us at support@opencollective.com.

Warmly 

-- The Open Collective Team

server/lib/subscriptions.js
```diff
@@ -253,6 +265,24 @@ export async function cancelSubscriptionAndNotifyUser(order) {
   return sendFailedEmail(order, true);
 }
 
+/** Send `archived.collective` email */
+export async function notifyUserForArchivedCollective(order) {
+  const user = order.createdByUser;
```
Betree: Unfortunately we don't have any database constraint to ensure that `createdByUser` is not null. Though we only have one order like that in the database, we need to be defensive here to ensure we don't crash subscriptions if we end up with one.

znarf: Agree, let's rely on information provided by the `order.fromCollective` (or similar, not sure about the syntax).

flickz: `order.createdByUser`  is used by other methods within that file too, basically to access `generateLoginLink` method and user `email` but I don't think `order.fromCollective` has the method nor the `email`. But one way to we can do it is to query `user` with `order.fromCollective.id`, what do you think?

znarf: You're right, `order.createdByUser` is the current practice and there is no point in refactoring that here. We should however keep that in mind, and later refactor to notify **accounts: (collectives/organizations/individuals)** and not **users**.

PR #1819, opened 2019-03-23
Delete account

flickz:
This PR:
 - Renames the current [`deleteCollective`](https://github.com/opencollective/opencollective-api/blob/master/server/graphql/v1/mutations/collectives.js#L514-L535) to `deleteCollectiveEvent` because it is essentially used to delete collective event in the frontend. 

- Adds a new `deleteCollective` to collective mutation resolver that:
             - Checks the user is logged in, collective is available, user is collective admin.
             - Checks the collective is deletable by making sure the collective has no transaction, order and PAID expense.
              - Sequentially soft deletes collective `members`, `tiers`, `expenses`, `paymentMethods`, `connectedAccounts` and `collective` itself. Each with concurrency limit of `3`. 

-  Adds `deleteUserCollective` to collective mutation resolver that:
           -  Checks the user is logged in
            - Checks the user collective is deletable by making sure the user has no transaction, order and PAID expense.
             - Sequentially soft deletes user `membership`, `expenses`, `paymentMethods`,  `connectedAccounts`, `userCollective` and `user`. Each with concurrency limit of `3`. 

- Adds `isDeletable` to collective property.

Ref ([#1793](https://github.com/opencollective/opencollective/issues/1793))

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -511,7 +511,7 @@ export async function approveCollective(remoteUser, CollectiveId) {
   return collective.update({ isActive: true });
 }
 
-export function deleteCollective(_, args, req) {
+export function deleteEvent(_, args, req) {
   if (!req.remoteUser) {
     throw new errors.Unauthorized({
       message: 'You need to be logged in to delete a collective',
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -17,7 +17,7 @@ import { types } from '../../../constants/collectives';
 const debugClaim = debug('claim');
 const debugGithub = debug('github');
 const debugArchive = debug('archive');
-const debugDeleteCollective = debug('delete');
+const debugDelete = debug('delete');
 
 export async function createCollective(_, args, req) {
   if (!req.remoteUser) {
@@ -279,7 +279,7 @@ export async function createCollectiveFromGithub(_, args, req) {
   collectiveData.currency = 'USD';
   collectiveData.CreatedByUserId = user.id;
   collectiveData.LastEditedByUserId = user.id;
-  collectiveData.tiers = [
+  collectiveData.teirs = [
     {
       name: 'backer',
       title: 'Backers',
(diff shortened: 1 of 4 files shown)


znarf:
@flickz Nice. Please check the failing test.

test/graphql.mutation.test.js
```diff
@@ -1001,7 +1001,7 @@ describe('Mutation Tests', () => {
             mutation createOrder($order: OrderInputType!) {
               createOrder(order: $order) {
                 id,
-                createdByUser {
```
znarf: @flickz check that

znarf:
Because this is a breaking change with Frontend, we need to make sure we sync the respective deployments.

znarf:
Good for me. As a follow up, I would extract the core actions out of GraphQL so that hey can be shared with Forest, CLI tools, etc ...

znarf:
@flickz After a second look, why not group together `deleteCollective`and `deleteUserCollective`? Feels to me they are similar enough and `deleteUserCollective` just need an extra step.

server/graphql/v1/mutations/collectives.js
```diff
@@ -278,7 +279,7 @@ export async function createCollectiveFromGithub(_, args, req) {
   collectiveData.currency = 'USD';
   collectiveData.CreatedByUserId = user.id;
   collectiveData.LastEditedByUserId = user.id;
-  collectiveData.teirs = [
```
znarf: Actually, what's that? Was that broken since we launched the new onboarding?

server/graphql/v1/mutations/collectives.js
```diff
@@ -710,3 +711,222 @@ export async function unarchiveCollective(_, args, req) {
...
+        members,
+        member => {
+          return member.destroy();
+        },
+        { concurrency: 3 },
+      );
+    })
+    .then(() => debugDelete('deleteCollectiveMembers'))
+    .then(async () => {
+      const expenses = await models.Expense.findAll({
```
Betree: Doing `models.Expenses.destroy({ where: { CollectiveId: collective.id } })` would have been much more efficient and concise.

PR #1837, opened 2019-03-28
Fix broken tiers

flickz:


znarf:
Let's double/triple check that.

This could be a case where trying to fix it will actually introduce a functional regression.

znarf:
So, let's remove that and edit the code at https://github.com/opencollective/opencollective-api/blob/master/server/models/Collective.js#L72 to make sure it works as expected (title, description, defaults).

We'll first need to check with @alannallama and @piamancini what should exactly be the default tiers from now.

alannallama:
@znarf sorry I just saw this. Are you still waiting on info from us for this?

znarf:
@alannallama The question is: are you currently happy with the default tiers for the Open Source Collectives?

PR #1838, opened 2019-03-29
Purge cloudflare cache for collective pages on certain actions

Betree:
Resolve https://github.com/opencollective/opencollective/issues/1870

### DIFF
--- a/server/lib/cloudflare.js
+++ b/server/lib/cloudflare.js
@@ -1,0 +1,44 @@
+import config from 'config';
+import cloudflare from 'cloudflare';
+
+import logger from './logger';
+
+// Load config
+const cfConfig = config.cloudflare || {};
+
+// Check config
+const isLiveServer = ['staging', 'production'].includes(config.env);
+const hasConfig = Boolean(cfConfig.email && cfConfig.key && cfConfig.zone);
+
+if (isLiveServer && !hasConfig) {
+  logger.warn('Your Cloudflare config is imcomplete, you must provide an email, a key and a zone');
+} else if (!isLiveServer && hasConfig) {
+  logger.info('A Cloudflare config was provided on a Live/Test environment. Some methods will be stubbed.');
+}
+
+const CloudflareLib = cloudflare({ email: cfConfig.email, key: cfConfig.key });
+
+// Export some helpers
+
+/**
+ * Purge the given page from cloudflare's cache. In dev environments this function
+ * will only log to the console and will skip the call to cloudflare.
+ *
+ * @param {string|array} pagePaths - a path or an array of paths `/eslint`, `['/', '/about']`
+ * @returns {Promise}
+ */
+export const purgeCacheForPage = pagePaths => {
+  const prepareURL = pagePath => {
+    const prependSlash = pagePath[0] === '/' ? '' : '/';
+    return `${config.host.website}${prependSlash}${pagePath}`;
+  };
+
+  const urlsToPurge = Array.isArray(pagePaths) ? pagePaths.map(prepareURL) : [prepareURL(pagePaths)];
+  logger.info(`Asking cloudflare to purge the cache for ${urlsToPurge}`);
+
+  if (!isLiveServer || !hasConfig) {
+    return Promise.resolve({ success: true, errors: [], messages: [], result: {} });
+  }
+
+  return CloudflareLib.zones.purgeCache(cfConfig.zone, { files: urlsToPurge }).catch(logger.error);
+};
--- a/config/custom-environment-variables.json
+++ b/config/custom-environment-variables.json
@@ -54,6 +54,11 @@
       "bucket": "AWS_S3_BUCKET"
     }
   },
+  "cloudflare": {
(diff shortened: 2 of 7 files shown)


server/gateways/cloudflare.js
```diff
@@ -0,0 +1,55 @@
...
+import cloudflare from 'cloudflare';
+import { get } from 'lodash';
+
+import logger from '../lib/logger';
+
+// Load config
+const cfConfig = {
+  email: get(config, 'clouflare.email'),
+  key: get(config, 'clouflare.key'),
+};
```
znarf: Looks unnecessary compared to `config.cloudflare`.

server/gateways/cloudflare.js
```diff
@@ -0,0 +1,55 @@
...
+
+export const zone = get(config, 'clouflare.zone');
```
znarf: Why should that be exported?

server/gateways/cloudflare.js
```diff
@@ -0,0 +1,55 @@
...
+
+// Check config
+const isLiveServer = ['staging', 'production'].includes(config.env);
+const hasConfig = Boolean(cfConfig.email && cfConfig.key && zone);
+
+if (isLiveServer && !hasConfig) {
+  logger.warn('Your Cloudflare config is imcomplete, you must provide an email, a key and a zone');
```
znarf: Let's avoid "you" for such messages?

server/gateways/cloudflare.js
```diff
@@ -0,0 +1,55 @@
+import config from 'config';
```
znarf: Why not put this library in `lib` instead of `gateways`?

znarf:
Looks good, let's try on staging before merging.

server/gateways/cloudflare.js
Betree: Almost all queries from `cloudflare` library expect a `zone` to be passed in params

server/gateways/cloudflare.js
Betree: This object is to get passed to `cloudflare(...)` (the lib) and it only expect an `email` and a `key`. If we pass the full object it will also include the `zone` parameter.

However I understand that this can be confusing, I'll refactor

server/gateways/cloudflare.js
Betree: But I'll remove the cloudflare export to only export the methods that will be used

server/gateways/cloudflare.js
Betree: `gateways` seemed to me like the place where we wrap third-party libraries. What's the purpose of this folder if that's not for that?

znarf: Then I would drop `gateways` and have everything in `lib`, this is where most wrappers are already.

Betree:
Tests on staging went well. Also added the configuration to prod.

:warning: Cloudflare doesn't allow to generate multiple API KEYs, so the key is shared between staging and production.

server/lib/cloudflare.js
```diff
@@ -0,0 +1,44 @@
...
+
+// Load config
+const cfConfig = config.cloudflare || {};
+
+// Check config
+const isLiveServer = ['staging', 'production'].includes(config.env);
+const hasConfig = Boolean(cfConfig.email && cfConfig.key && cfConfig.zone);
+
+if (isLiveServer && !hasConfig) {
+  logger.warn('Your Cloudflare config is imcomplete, you must provide an email, a key and a zone');
```
znarf: Same comment than the other day about "your", don't think we should formulate errors/warnings like that.

server/lib/cloudflare.js
```diff
@@ -0,0 +1,44 @@
...
+ * Purge the given page from cloudflare's cache. In dev environments this function
+ * will only log to the console and will skip the call to cloudflare.
+ *
+ * @param {string|array} pagePaths - a path or an array of paths `/eslint`, `['/', '/about']`
+ * @returns {Promise}
+ */
+export const purgeCacheForPage = pagePaths => {
+  const prepareURL = pagePath => {
+    const prependSlash = pagePath[0] === '/' ? '' : '/';
+    return `${config.host.website}${prependSlash}${pagePath}`;
```
znarf: Have you tested?

My experience of Cloudflare is that we need to empty `https://frontend.opencollective.com/babel` not `https://opencollective.com/babel`. I personally have the habit of purging both, not sure it's justified.

Betree: I tested a bit naively, just ensured the collective page was properly updated.

I just did more advanced testing with `curl -svo /dev/null` to get the headers and it seems like I was lucky during my tests; the cache is not purged properly.

However I don't have more luck by doing:

```es6
api.zones.purgeCache('...', {files: ['https://staging.opencollective.com/captainfact_io', 'https://staging-frontend.opencollective.com/captainfact_io']})
```

znarf: It's `https://frontend-staging.opencollective.com`.

PR #1846, opened 2019-04-01
Replaced Knox with Aws-Sdk

vickywane:
Since the Knox library was no longer maintained,  the official aws-sdk was decided to replace it

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,6 +129,7 @@
     "graphiql": "^0.13.0",
     "husky": "^1.3.1",
     "jsdoc": "^3.5.5",
+    "aws-sdk": "^2.432.0",
     "jsdoc-simple-theme": "^1.2.6",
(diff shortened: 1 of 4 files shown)


znarf:
Thank you!

I'm sorry, the knox file was moved by @Betree recently https://github.com/opencollective/opencollective-api/pull/1839, so you will need to update your branch to avoid conflicts. In case it's difficult maybe restart from a fresh version of master.

Also, can you publish the changes on a named branch instead of master? (you can name it `knox-aws-sdk` for example). It's always a good practice and will allow us to trigger CircleCI on the PR.

Finally, when issuing a PR, always double check the changes,  presently there is a `server/gateways/fix 1.txt`, you should remove that.

vickywane:
Alright.  
That you for the explanation.  
Feel free to close this PR as I would create another after the adjustments

PR #1783, opened 2019-03-13
Update API to json2csv version 4

eloyekunle:
v4 Changes that concern us:

- The actual JSON parson is now done via a `parse` method (i.e. From `json2csv(...)` to `json2csv.parse(...)`.
- The arguments accepted by the `parse` method is slightly different from that accepted by the previous `json2csv` function (From `json2csv({data, fields})` to `json2csv.parse(data, {fields})`.
- The parse is no longer done asynchronously, but the `parse` method generates and returns the CSV synchronously.

Fixes opencollective/opencollective#1777
Closes #1770

PR #1849, opened 2019-04-02
Consolidated yearly invoice #1783

pietgeursen:
# [Consolidated yearly invoice](https://github.com/opencollective/opencollective/issues/1783)

Until now all invoices were generated for one month intervals. This feature needs invoices generated for one year intervals, but this PR goes a step further and allows querying over an arbitrary date range.

This PR:
- adds a new graphql endpoint `InvoiceByDateRange`.
- defines a new graphql scalar type `IsoDateString` which is used for validating date strings.
- defines a new graphql input type `InvoiceInputType`, the type passed as an argument to the `InvoiceByDateRange` query and has shape:

```js
{
      dateFrom: { type: IsoDateString },
      dateTo: { type: IsoDateString },
      collectiveSlug: { type: GraphQLString },
      fromCollectiveSlug: { type: GraphQLString },
}
```

### DIFF
--- a/server/graphql/v1/types.js
+++ b/server/graphql/v1/types.js
@@ -1,17 +1,21 @@
 import {
   GraphQLBoolean,
   GraphQLEnumType,
+  GraphQLError,
   GraphQLFloat,
   GraphQLInt,
   GraphQLInputObjectType,
   GraphQLList,
   GraphQLObjectType,
   GraphQLString,
+  GraphQLScalarType,
 } from 'graphql';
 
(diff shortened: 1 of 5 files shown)


server/graphql/v1/types.js
```diff
@@ -337,6 +341,27 @@ export const LocationType = new GraphQLObjectType({
   }),
 });
 
+export const IsoDateString = new GraphQLScalarType({
```
Betree: This is going to be useful in the future. Nice!

PR #1854, opened 2019-04-03
Feat / Webhook Notifications

eloyekunle:
- [x] Add new `NotificationType` GraphQL object type, based on the `Notification` model.
- [x] Return `notifications` with a Collective. Makes it possible to query and edit it on the Edit Collective page.
- [x] Add `webhook` notification type.
- [x] Implement `editWebhooks` mutation.
- [x] Implement `subscribe` and `unsubscribe` endpoints for Zapier to subscribe to webhooks.

Ref: opencollective/opencollective#1839

### DIFF
--- a/migrations/20190403113644-notification-composite-index.js
+++ b/migrations/20190403113644-notification-composite-index.js
@@ -5,13 +5,9 @@ module.exports = {
     await queryInterface.addIndex('Notifications', ['channel', 'type', 'webhookUrl', 'CollectiveId'], {
       indicesType: 'UNIQUE',
     });
-    await queryInterface.addColumn('Notifications', 'deletedAt', {
-      type: DataTypes.DATE,
-    });
   },
 
   down: async function(queryInterface, Sequelize) {
     await queryInterface.removeIndex('Notifications', ['channel', 'type', 'webhookUrl', 'CollectiveId']);
-    await queryInterface.removeColumn('Notifications', 'deletedAt');
   },
 };
--- a/migrations/20190403113644-notification-composite-index.js
+++ b/migrations/20190403113644-notification-composite-index.js
@@ -1,13 +1,19 @@
 'use strict';
 
 module.exports = {
-  up: async function(queryInterface, DataTypes) {
-    await queryInterface.addIndex('Notifications', ['channel', 'type', 'webhookUrl', 'CollectiveId'], {
-      indicesType: 'UNIQUE',
-    });
+  up: function(queryInterface, DataTypes) {
+    return queryInterface
+      .addIndex('Notifications', ['channel', 'type', 'webhookUrl', 'CollectiveId'], {
+        indicesType: 'UNIQUE',
+      })
+      .then(() => {
+        queryInterface.removeIndex('Notifications', ['type', 'CollectiveId', 'UserId']);
+      });
   },
 
-  down: async function(queryInterface, Sequelize) {
-    await queryInterface.removeIndex('Notifications', ['channel', 'type', 'webhookUrl', 'CollectiveId']);
+  down: function(queryInterface, Sequelize) {
+    queryInterface.removeIndex('Notifications', ['channel', 'type', 'webhookUrl', 'CollectiveId']).then(() => {
+      queryInterface.addIndex('Notifications', ['type', 'CollectiveId', 'UserId'], { indicesType: 'UNIQUE' });
+    });
   },
 };
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -607,6 +607,7 @@ export const CollectiveInterfaceType = new GraphQLInterfaceType({
           offset: { type: GraphQLInt },
           channel: { type: GraphQLString },
           type: { type: GraphQLString },
+          active: { type: GraphQLBoolean },
         },
       },
       maxQuantity: { type: GraphQLInt },
@@ -1138,10 +1139,23 @@ const CollectiveFields = () => {
         offset: { type: GraphQLInt },
         channel: { type: GraphQLString },
         type: { type: GraphQLString },
+        active: { type: GraphQLBoolean },
       },
       resolve(collective, args) {
-        return models.Notifications.findAll({
-          where: { CollectiveId: collective.id, args: { channel, type } },
+        let where = { CollectiveId: collective.id };
+
+        if (args.channel) {
+          where.channel = args.channel;
+        }
+        if (args.type) {
+          where.type = args.type;
+        }
+        if (args.active) {
+          where.active = args.active;
+        }
+
+        return models.Notification.findAll({
+          where: where,
           limit: args.limit,
           offset: args.offset,
         });
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -1142,7 +1142,7 @@ const CollectiveFields = () => {
         active: { type: GraphQLBoolean },
       },
       resolve(collective, args) {
-        let where = { CollectiveId: collective.id };
+        const where = { CollectiveId: collective.id };
 
         if (args.channel) {
           where.channel = args.channel;
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -26,6 +26,7 @@ import {
   OrderStatusType,
   PaginatedPaymentMethodsType,
   ImageFormatType,
+  NotificationType,
 } from './types';
 
 import { OrderDirectionType, TransactionInterfaceType } from './TransactionInterface';
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -598,6 +598,16 @@ export const CollectiveInterfaceType = new GraphQLInterfaceType({
           offset: { type: GraphQLInt },
         },
       },
+      notifications: {
+        type: new GraphQLList(NotificationType),
+        description: 'List of all notifications for this collective',
+        args: {
+          limit: { type: GraphQLInt },
+          offset: { type: GraphQLInt },
+          channel: { type: GraphQLString },
+          type: { type: GraphQLString },
+        },
+      },
       maxQuantity: { type: GraphQLInt },
       tiers: {
         type: new GraphQLList(TierType),
@@ -1118,6 +1128,22 @@ const CollectiveFields = () => {
           limit: args.limit,
           offset: args.offset,
         }).then(memberships => memberships.memberCollective);
+      },
+    },
+    notifications: {
+      type: new GraphQLList(NotificationType),
+      args: {
+        limit: { type: GraphQLInt },
+        offset: { type: GraphQLInt },
+        channel: { type: GraphQLString },
+        type: { type: GraphQLString },
+      },
+      resolve(collective, args) {
+        return models.Notifications.findAll({
+          where: { CollectiveId: collective.id, args: { channel, type } },
+          limit: args.limit,
+          offset: args.offset,
+        });
       },
     },
     maxQuantity: {
--- a/server/graphql/v1/inputTypes.js
+++ b/server/graphql/v1/inputTypes.js
@@ -161,7 +161,7 @@ export const CollectiveInputType = new GraphQLInputObjectType({
     settings: { type: GraphQLJSON },
     data: { type: GraphQLJSON },
     members: { type: new GraphQLList(MemberInputType) },
-    notification: { type: new GraphQLList(NotificationInputType) },
+    notifications: { type: new GraphQLList(NotificationInputType) },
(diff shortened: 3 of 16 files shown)


server/models/Collective.js
```diff
@@ -1555,6 +1556,33 @@ export default function(Sequelize, DataTypes) {
       .then(() => this.getTiers());
   };
 
+  // edit the notifications of this collective (create/update/remove)
```
xdamman: This is a very bad pattern to edit notifications of a collective. I know that you used what we (I) did for editing members, but it's bad. It was a shortcut and it's bug prone (we should also refactor `editMembers` and `editTiers` at some point).
I'd highly recommend to create new mutations to keep things clear.

Also as a paradigm, I don't think it's right to have `editNotifications` linked to `editCollective`. You shouldn't have to be an admin of a collective to register a webhook for events happening to that collective (though it's fine to assume that for a MVP if it makes things simpler). The only thing is that if you are not an admin of the collective, you cannot receive private information about a transaction (typically the email of a user or payment method or attachment to an expense).

server/graphql/v1/inputTypes.js
```diff
@@ -145,6 +159,7 @@ export const CollectiveInputType = new GraphQLInputObjectType({
     settings: { type: GraphQLJSON },
     data: { type: GraphQLJSON },
     members: { type: new GraphQLList(MemberInputType) },
+    notification: { type: new GraphQLList(NotificationInputType) },
```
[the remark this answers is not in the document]
eloyekunle: Fixed

server/models/Collective.js
eloyekunle: @xdamman I have created a new mutation for editing notifications.

server/graphql/v1/mutations.js
```diff
@@ -717,6 +720,44 @@ const mutations = {
...
+      notifications: { type: new GraphQLList(NotificationInputType) },
+    },
+    resolve(_, args, req) {
+      return editNotifications(args, req.remoteUser);
+    },
+  },
+  createNotification: {
+    type: NotificationType,
+    args: {
+      notification: NotificationInputType,
```
Betree: This doesn't work, we need to be explicit here:

```suggestion
      notification: { type: NotificationInputType },
```

server/graphql/v1/mutations.js
```diff
@@ -717,6 +720,44 @@ const mutations = {
...
+    },
+    resolve(_, args, req) {
+      return createNotification(args, req.remoteUser);
+    },
+  },
+  createWebhook: {
+    type: NotificationType,
+    args: {
+      collectiveSlug: { type: new GraphQLNonNull(GraphQLString) },
+      notification: NotificationInputType,
```
Betree: This doesn't work, we need to be explicit here:

```suggestion
      notification: { type: NotificationInputType },
```

server/graphql/v1/mutations/notifications.js
```diff
@@ -0,0 +1,88 @@
...
+import activities from '../../../constants/activities';
+import { channels } from '../../../constants';
+import { Forbidden } from '../../errors';
+
+const NotificationPermissionError = new Forbidden({
+  message: "This notification does not exist or you don't have the permission to edit it.",
+});
+
+export async function editNotifications(args, remoteUser) {
+  if (!remoteUser) {
```
Betree: We also need to check that `remoteUser` is admin of collective `args.id`. See https://github.com/opencollective/opencollective-api/blob/32f473bdbd17056bf8dbef7e6e6c392c32fa8747/server/graphql/v1/mutations/collectives.js#L691 for an example

server/graphql/v1/mutations/notifications.js
```diff
@@ -0,0 +1,88 @@
...
+        return models.Notification.create(notification);
+      });
+    });
+}
+
+/**
+ * Creates a new notification
+ */
+export async function createNotification(args, remoteUser) {
+  if (!remoteUser) {
```
Betree: We also need to check that `remoteUser` is admin of collective `CollectiveId`. See https://github.com/opencollective/opencollective-api/blob/32f473bdbd17056bf8dbef7e6e6c392c32fa8747/server/graphql/v1/mutations/collectives.js#L691 for an example

server/graphql/v1/mutations/notifications.js
```diff
@@ -0,0 +1,88 @@
...
+  const collective = models.Collective.find({ where: { slug: args.collectiveSlug } });
+
+  args.notification.channel = channels.WEBHOOK;
+  args.notification.CollectiveId = collective.id;
+
+  return createNotification(args, remoteUser);
+}
+
+export async function deleteNotification(args, remoteUser) {
+  if (!remoteUser) {
```
Betree: We also need to check that `remoteUser` is admin of collective `CollectiveId`. See https://github.com/opencollective/opencollective-api/blob/32f473bdbd17056bf8dbef7e6e6c392c32fa8747/server/graphql/v1/mutations/collectives.js#L691 for an example

server/graphql/v1/types.js
```diff
@@ -869,6 +869,72 @@ export const CommentType = new GraphQLObjectType({
...
 });
 
+export const NotificationType = new GraphQLObjectType({
+  name: 'NotificationType',
+  description: 'This represents a Notification',
+  fields: () => {
+    return {
+      id: {
+        type: GraphQLInt,
+        resolve(notification) {
```
Betree: Not a requested change, just an info: when the field map the schema it is not necessary to add a `resolve` function. GraphQL resolves to `model => model[fieldName]` by default.

migrations/20190403113644-notification-composite-index.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+module.exports = {
+  up: async function(queryInterface, DataTypes) {
+    await queryInterface.addIndex('Notifications', ['channel', 'type', 'webhookUrl', 'CollectiveId'], {
```
Betree: We don't have enough database constraints in our current schema. This index is much appreciated!

server/graphql/v1/types.js
eloyekunle: Oh, just knowing this. Thanks.

server/graphql/v1/mutations/notifications.js
eloyekunle: I think we want non-admins to be able to register event webhooks as well?

server/graphql/v1/mutations/notifications.js
eloyekunle: Since non-admins will be able to register webhooks, they should also be allowed to delete webhooks they register.

server/graphql/v1/mutations/notifications.js
Betree: I'll let @znarf confirm that as I don't know how the frontend implementation is going to look like. If webhooks are meant to be created from the admin interface ("edit collective") then we should be protective per default.

server/graphql/v1/mutations/notifications.js
Betree: Yes, the check below (`remoteUser.id === notification.UserId || remoteUser.isAdmin(notification.CollectiveId)`) looks good to me

server/graphql/v1/mutations/notifications.js
Betree: As this lives in the `Edit collective` menu that is only accessible to Collective admins, I can confirm that we should check for `req.remoteUser.isAdmin(CollectiveId)`

eloyekunle: This `createNotification` mutation is supposed to be called by Zapier, and we want non-admins to integrate via Zapier as well.

The `Edit collective` menu calls the `editNotifications` mutation which is defined above.

A difference between the two mutations is that `createNotification` persists the user ID as well, while `editNotifications` does not because the latter is supposed to be collective-wide, while the former can be integrated by anybody.

Betree: My bad, it's all good then. Thanks for sharing the details.

server/graphql/v1/mutations/notifications.js
```diff
@@ -0,0 +1,94 @@
...
+
+export async function editNotifications(args, remoteUser) {
+  if (!(remoteUser && remoteUser.isAdmin(args.id))) {
+    throw NotificationPermissionError;
+  }
+
+  if (!args.notifications) return Promise.resolve();
+  let newNotifications = [];
+
+  return models.Notification.findAll({ where: { CollectiveId: args.id, UserId: null } })
```
Betree: As we're loading all notifications here, this is going to delete all existing notifications (for `email`, `twitter`, `gitter` and `slack`) that already exists for the collective.

Betree: A simple way to fix this would be to rename this function `editWebhooks` and to make it deal only with webhooks.

eloyekunle: Addressed.

server/graphql/v1/mutations.js
```diff
@@ -717,6 +720,44 @@ const mutations = {
       return paymentMethodsMutation.removePaymentMethod(args.id, req.remoteUser);
     },
   },
+  editNotifications: {
+    type: new GraphQLList(NotificationType),
+    args: {
+      id: { type: new GraphQLNonNull(GraphQLInt) },
```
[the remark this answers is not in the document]
eloyekunle: Addressed.

server/graphql/v1/mutations/notifications.js
```diff
@@ -0,0 +1,94 @@
...
+    .then(oldNotifications => {
+      const diff = oldNotifications
+        .filter(
+          x1 =>
```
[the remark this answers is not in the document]
eloyekunle: Addresses.

server/graphql/v1/mutations/notifications.js
```diff
@@ -0,0 +1,94 @@
...
+
+      return models.Notification.destroy({ where: { id: { [Op.in]: diff } } });
+    })
+    .then(() => {
+      return Promise.map(newNotifications, notification => {
+        if (!(values(activities).includes(notification.type) && values(channels).includes(notification.channel)))
+          return;
+
+        notification.CollectiveId = args.id;
+        return models.Notification.create(notification);
```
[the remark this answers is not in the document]
eloyekunle: I've implemented webhooks limitation.

server/graphql/v1/CollectiveInterface.js
```diff
@@ -1120,6 +1132,35 @@ const CollectiveFields = () => {
...
+      type: new GraphQLList(NotificationType),
+      args: {
+        limit: { type: GraphQLInt },
+        offset: { type: GraphQLInt },
+        channel: { type: GraphQLString },
+        type: { type: GraphQLString },
+        active: { type: GraphQLBoolean },
+      },
+      resolve(collective, args) {
+        let where = { CollectiveId: collective.id };
```
Betree: This is not re-assigned, we can go with `const`

server/graphql/v1/mutations/notifications.js
```diff
@@ -0,0 +1,110 @@
...
+        );
+      }),
+  );
+
+  models.Notification.destroy({ where: { id: { [Op.in]: toDelete } } });
+
+  return Promise.map(toCreate, notification => {
+    if (!(values(activities).includes(notification.type) && notification.channel === channels.WEBHOOK)) return;
+
+    notification.CollectiveId = args.id;
```
Betree:
```suggestion
    notification.CollectiveId = args.collectiveId;
```

server/graphql/v1/CollectiveInterface.js
eloyekunle: Addressed.

server/graphql/v1/mutations/notifications.js
Betree: @eloyekunle As discussed on Slack, this is where we should ensure that user is admin of the collective, otherwise anyone could create a webhook to a malicious endpoint and make private data leak.

eloyekunle: @Betree I have updated the mutations to check admin status.

server/graphql/v1/mutations/notifications.js
```diff
@@ -0,0 +1,110 @@
...
+    notification.CollectiveId = args.collectiveId;
+    return models.Notification.create(notification);
+  });
+}
+
+/**
+ * Creates a Webhook subscription for a collective given a collective slug.
+ */
+export async function createWebhook(args, remoteUser) {
+  if (!(remoteUser && remoteUser.isAdmin(args.collectiveId))) {
```
Betree: I think `createWebhook` is taking a `collectiveSlug` (not a `collectiveId`), does that work?

eloyekunle: It doesn't. I've fixed it.

znarf:
Thank you all. It's time to merge !

eloyekunle:
Nice!

PR #1855, opened 2019-04-03
Ability to query the collective of an host

znarf:
To enable that, add the collective id to the host settings, ie: `{"hostCollective":{"id":83}}`.

Related:
https://github.com/opencollective/opencollective/issues/1884
https://github.com/opencollective/opencollective-frontend/pull/1622

### DIFF
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -525,6 +525,7 @@ export const CollectiveInterfaceType = new GraphQLInterfaceType({
       isArchived: { type: GraphQLBoolean },
       isDeletable: { type: GraphQLBoolean },
       host: { type: CollectiveInterfaceType },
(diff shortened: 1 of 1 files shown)


server/graphql/v1/CollectiveInterface.js
```diff
@@ -525,6 +525,7 @@ export const CollectiveInterfaceType = new GraphQLInterfaceType({
       isArchived: { type: GraphQLBoolean },
       isDeletable: { type: GraphQLBoolean },
       host: { type: CollectiveInterfaceType },
+      collective: { type: CollectiveInterfaceType },
```
Betree: Having `collective.collective` may be confusing. Can we find a more accurate name for this like `hostCollective`, `hostFrontCollective`?

znarf: I agree, `hostCollective` perhaps.

PR #1859, opened 2019-04-03
Knox replacement

vickywane:
Aws-sdk replacement  for Knox


vickywane:
Sorry I had to close the former PR.  
I lost my old copy of the project and had to start afresh which lead to some conflicts in the two branches. 

By using the linters I have been able to fix the problem and also running yarn install solves the first issue

znarf:
@vickywane in this project, you should not use `yarn` but `npm install`, and you should commit the modifications to `package-lock.json`.

vickywane:
Sorry about that. 
My personal preference has been Yarn over npm.  
Once I'll push once I make the changes.  

Is yarn responsible for the lint error?

znarf:
@vickywane yes, without the updated `package-lock.json`, we can't install the packages and this is breaking all the tests.

vickywane:
I have updated the required files. 
Can we run the tests again?

znarf:
Unfortunately, the tests are not passing. Have you tried `npx mocha test/images.routes.test.js` ?

vickywane:
@francois I just used the command and went through the file.  
2 failed tests that I don't know how to solve 😔

znarf:
@vickywane you need to make the test work then, if they don't, the feature is likely to be broken.

vickywane:
I refactored the way I used the aws-sdk in my new commit.. The logs shown from the tests do not indicate anything.  I however tested my algorithm in a sandbox and I worked.

znarf:
@vickywane ok, let's try to find more about this error.

znarf:
If you edit the test, you can inspect the error:

```
      .expect(function(res) {
        console.log(res);
      })
      .expect(200)
```

server/lib/awsClient.js
```diff
@@ -0,0 +1,17 @@
+import config from 'config';
+import aws from 'aws-sdk';
+
+// S3 bucket
+let awsClient;
+
+if (config.aws.s3.key) {
+  awsClient = aws.config.update({
```
znarf: Looks like `aws.config.update` is not returning anything, leading to `awsClient` being undefined.

server/controllers/images.js
```diff
@@ -1,11 +1,13 @@
 import path from 'path';
 import uuidv1 from 'uuid/v1';
+import config from 'config';
 
 import errors from '../lib/errors';
-import knox from '../lib/knox';
+import S3 from '../lib/awsClient';
```
znarf: Inspect `S3`, it should not be `undefined`

server/controllers/images.js
```diff
@@ -35,30 +37,33 @@ export default function uploadImage(req, res, next) {
     );
   }
 
-  if (!knox) {
+  if (!S3) {
```
znarf: This is currently failing here, because `S3` is undefined.

server/lib/awsClient.js
```diff
@@ -0,0 +1,17 @@
...
+    accessKeyId: config.aws.s3.key,
+    secretAcessKey: config.aws.s3.secret,
+    region: 'us-west-1',
+  });
+}
+
+const S3 = new awsClient.S3();
```
znarf: This is breaking here because `awsClient` is undefined.

vickywane:
Thank you for the head-start.  It would keep working on it

vickywane:
@znarf.  
I turns connection to s3 is initialized using ```js 
new aws.S3({})
```

This change however is not reflected in the docs.  I got it through trial by error.  

I however cannot access the config file hence when I use arbitrary values for the keys everything seems fine.  

I think am enjoying this!

znarf:
Sorry, I can't start reviewing the code itself until the tests are passing. 

- You should address my comments from 3 days ago about `awsClient` being undefined. This is absolutely broken.
- Why did you remove the `multer` package? This is breaking everything.

Looks you're continuously submitting code that is broken and not tested. Before asking for review, the automated tests should pass and you should also test the functionality by yourself.

vickywane:
I didnt make any new commit.  

I only notice that awsClient was being undefined because of a change in the documentation .

The tests errors have changed since then

vickywane:
I installed the multer package in a bid to substitute it with the aws - sdk but that didn't happen so I uninstalled it.  

I totally forgot it being utilised elsewhere

vickywane:
Closing this til all tests pass locally

PR #1860, opened 2019-04-04
fix(emails): fall back to ui-avatars.com when no avatar

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/1906

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -221,7 +221,7 @@ export default function(Sequelize, DataTypes) {
             const image = `https://logo.clearbit.com/${getDomain(this.website)}`;
             return image;
           }
-          return `https://ui-avatars.com/api/?rounded=true&name=${this.name}`;
+          return `https://ui-avatars.com/api/?rounded=true&name=${encodeURIComponent(this.name)}`;
         },
       },
 
@@ -1131,7 +1131,8 @@ export default function(Sequelize, DataTypes) {
               backerCollective.isActive = order.Subscription.isActive;
             }
             backerCollective.image =
-              backerCollective.image || `https://ui-avatars.com/api/?rounded=true&name=${backerCollective.name}`;
+              backerCollective.image ||
+              `https://ui-avatars.com/api/?rounded=true&name=${encodeURIComponent(backerCollective.name)}`;
             debug('adding to tier', TierId, 'backer: ', backerCollective.dataValues.slug);
             tiersById[TierId].dataValues.users.push(backerCollective.dataValues);
           });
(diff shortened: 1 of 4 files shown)


server/models/Collective.js
```diff
@@ -898,11 +899,12 @@ export default function(Sequelize, DataTypes) {
    * @param {*} startDate beginning of the time period
    * @param {*} endDate end of the time period
    */
-  Collective.prototype.getNewOrders = async function(startDate = 0, endDate = new Date()) {
```
znarf: Looks it's not related to the PR.

server/models/Collective.js
```diff
@@ -221,6 +221,7 @@ export default function(Sequelize, DataTypes) {
             const image = `https://logo.clearbit.com/${getDomain(this.website)}`;
             return image;
           }
+          return `https://ui-avatars.com/api/?rounded=true&name=${encodeURIComponent(this.name)}`;
```
znarf: Shouldn't we only use ui-avatars for 'USER'?

znarf: What's the impact of that change in the frontend?

server/models/Collective.js
```diff
@@ -898,11 +899,12 @@ export default function(Sequelize, DataTypes) {
    * @param {*} startDate beginning of the time period
    * @param {*} endDate end of the time period
    */
-  Collective.prototype.getNewOrders = async function(startDate = 0, endDate = new Date()) {
+  Collective.prototype.getNewOrders = async function(startDate = 0, endDate = new Date(), options = {}) {
     const orders = await models.Order.findAll({
       where: {
         CollectiveId: this.id,
         createdAt: { [Op.gte]: startDate, [Op.lt]: endDate },
+        ...options.where,
```
znarf: Looks not related to the PR.

server/models/Collective.js
xdamman: Good idea to only do it for `USER`.

It means that we will have to check for `ui-avatars.com` instead of `undefined` if we want to show a different default avatar on the frontend. But in the meantime, it will fix this: https://github.com/opencollective/opencollective/issues/1903

server/models/Collective.js
xdamman: ah right

server/models/Collective.js
xdamman: removing this for cleanness.

znarf:
I think we should fix that differently.

- The API should not be aware of `ui-avatars`, it's the responsibility of the `images` service.
- In the sake of simplification, we should also drop the SVG initials in frontend and only rely on `ui-avatars` (through the `images` service).

The responsibility of the API should then only be to expose the URL of the image service: `https://images.opencollective.com/{slug}/avatar.png`

znarf:
I introduced some helpers in https://github.com/opencollective/opencollective-api/pull/1909

`collective.getImageUrl();`

znarf:
Closing the PR. We will fix the issue differently.

PR #1861, opened 2019-04-04
fix(report): only include paid orders in monthly collective report

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/1885


cron/monthly/collective-report.js
```diff
@@ -167,7 +167,7 @@ const processCollective = collective => {
     collective.getExpenses(null, startDate, endDate),
     collective.getRelatedCollectives(3, 0, 'c."createdAt"', 'DESC'),
     collective.getBackersStats(startDate, endDate),
-    collective.getNewOrders(startDate, endDate),
+    collective.getNewOrders(startDate, endDate, { status: 'PAID' }),
```
znarf: What about `status: 'ACTIVE'`?

xdamman: Good point. Fixing.

PR #1862, opened 2019-04-05
Use "open source" tag instead of "opensource"

Betree:
See https://github.com/opencollective/opencollective/issues/1886

### DIFF
--- a/server/lib/utils.js
+++ b/server/lib/utils.js
@@ -412,17 +412,19 @@ export function exportToPDF(template, data, options) {
 
 /**
  * Default host id, set this for new collectives created through our flow
+ *
+ * @param {"opensource" | null} category of the collective
  */
-export const defaultHostCollective = tag => {
+export const defaultHostCollective = category => {
   if (process.env.NODE_ENV === 'production' || process.env.NODE_ENV === 'staging') {
-    if (tag === 'opensource') {
+    if (category === 'opensource') {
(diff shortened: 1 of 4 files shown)


server/lib/utils.js
```diff
@@ -412,17 +412,19 @@ export function exportToPDF(template, data, options) {
 
 /**
  * Default host id, set this for new collectives created through our flow
+ *
+ * @param {"opensource" | null} category of the collective
  */
-export const defaultHostCollective = tag => {
+export const defaultHostCollective = category => {
   if (process.env.NODE_ENV === 'production' || process.env.NODE_ENV === 'staging') {
-    if (tag === 'opensource') {
```
znarf: not "open source"?

server/lib/utils.js
```diff
@@ -412,17 +412,19 @@ export function exportToPDF(template, data, options) {
...
+    if (category === 'opensource') {
       return { id: 772, CollectiveId: 11004, ParentCollectiveId: 83 }; // Open Source Host Collective
     } else {
       return {}; // Don't automatically assign a host anymore
     }
   }
   if (process.env.NODE_ENV === 'development' || process.env.E2E_TEST) {
-    if (tag === 'opensource') {
```
znarf: not "open source"?

server/lib/utils.js
Betree: I was hesitating on renaming these. We call this function from multiple places and we never really pass any tags to it, only static strings. So I preferred renaming the parameter from `tag` to `category` to remove any confusion

PR #1918, opened 2019-04-23
Fix / Upgrade Sequelize

eloyekunle:
Upgrades Sequelize to 5.8.2, and Sequelize-CLI to 5.4.0.

Closes #1592 
Closes #1788 

Ref: opencollective/opencollective#1873


server/models/Transaction.js
```diff
@@ -397,6 +397,11 @@ export default (Sequelize, DataTypes) => {
     transaction.TransactionGroup = uuidv4();
     transaction.hostCurrencyFxRate = transaction.hostCurrencyFxRate || 1;
 
+    if (!isUndefined(transaction.amountInHostCurrency)) {
```
znarf: @eloyekunle do you know what was the breaking change in sequelize?

@Betree can you review this?

test/graphql.updateOrder.test.js
```diff
@@ -8,6 +8,7 @@ import models from '../server/models';
 const ordersData = utils.data('orders');
 
 const order = {
+  id: null,
```
znarf: @eloyekunle do you know what was the breaking change in sequelize?

package.json
```diff
@@ -98,9 +98,9 @@
     "request-as-curl": "0.1.0",
     "request-promise": "4.2.3",
     "sanitize-html": "1.20.1",
-    "sequelize": "4.43.0",
-    "sequelize-cli": "4.0.0",
-    "sequelize-temporal": "1.0.6",
+    "sequelize": "^5.7.6",
```
znarf: We use fixed versions. Should be `5.7.6` or whatever is the most recent.

znarf:
It's really not super important, but half of this PR is coming from commits I authored in https://github.com/opencollective/opencollective-api/tree/greenkeeper/sequelize-5.1.0.

For history purpose and review, it would have been clearer to keep my commits and not squash them.

eloyekunle:
@znarf I initially kept the commits, until today when I needed to rebase and I was fixing issues on each of 17 commits, so I decided to squash so I could rebase once.

package.json
eloyekunle: Done

server/models/Transaction.js
eloyekunle: It seems Sequelize was used to coercing Double values to fit into Integer columns until v5.

The `amountInHostCurrency` column is an integer column, but we were passing in floats in some cases so it was returning errors.

test/graphql.updateOrder.test.js
eloyekunle: In this case, we had a test saying return an error if there is no id. 

Sequelize v4 worked fine when `id` was undefined, but v5 gives errors so I needed to make it `null` to pass.

eloyekunle:
@znarf I've fixed the history.

Betree:
This https://github.com/opencollective/opencollective-api/blob/b802b0eed236e30f1a04eef7a989f2dc5c3c07c3/server/models/Collective.js#L260

is going to break because of:

> Validators
>
> Custom validators defined per attribute (as opposed to the custom validators defined in the model's options) now run when the attribute's value is null and allowNull is true (while previously they didn't run and the validation succeeded immediately). To avoid problems when upgrading, please check all your custom validators defined per attribute, where allowNull is true, and make sure all these validators behave correctly when the value is null. See #9143.

eloyekunle:
@Betree Good catch! I've added a check for this case.

CircleCI isn't building because this is from a forked branch.

znarf:
Thank you @eloyekunle! Merging.

A $300 bounty was attached to this issue, see: https://github.com/opencollective/opencollective/issues/1873.

To claim the bounty, you can:

  a) submit an invoice to https://opencollective.com/engineering
  b) or ask for an Open Collective gift card of the same value (to donate to other collectives)

eloyekunle:
@znarf  Great, I've [submitted an expense](https://opencollective.com/engineering/expenses/7898).

PR #1919, opened 2019-04-23
Introduce minimum amount in Tier management

flickz:
This PR introduces `minimumAmount` to tier data replacing the need to use `Min(amount, presets[0])`. 

Ref ([#1938](https://github.com/opencollective/opencollective/issues/1938))

### DIFF
--- a/migrations/20190418120035-add_minmumAmount_column_to_tiers.js
+++ b/migrations/20190418120035-add_minmumAmount_column_to_tiers.js
@@ -26,10 +26,12 @@ module.exports = {
           let minimumAmount = null;
           if (!presets && amount) {
             minimumAmount = amount;
-          } else if (presets && !amount) {
+          } else if (presets && presets.length > 0 && !amount) {
             minimumAmount = Math.min(...presets);
-          } else if (presets && amount) {
+          } else if (presets && presets.length > 0 && amount) {
             minimumAmount = Math.min(amount, ...presets);
+          } else if (presets.length === 0 && amount) {
+            minimumAmount = amount;
           }
 
           return queryInterface.sequelize.query(
--- a/migrations/20190418120035-add_minmumAmount_column_to_tiers.js
+++ b/migrations/20190418120035-add_minmumAmount_column_to_tiers.js
@@ -1,0 +1,64 @@
+'use strict';
+import { map } from 'bluebird';
+const colName = 'minimumAmount';
+
+/**
+ * Add a `minimumAmount` column to tiers to remove ambiguity.
+ */
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    const colParams = {
+      type: Sequelize.INTEGER,
+    };
+
+    return queryInterface
+      .addColumn('Tiers', colName, colParams)
+      .then(() => {
+        return queryInterface.sequelize.query(`
+        SELECT * FROM "Tiers"
+      `);
+      })
+      .then(results => {
+        const tiers = results[0];
+        return map(tiers, tier => {
+          const presets = tier.presets;
+          const amount = tier.amount;
+          let minimumAmount = null;
+          if (!presets && amount) {
+            minimumAmount = amount;
+          } else if (presets && !amount) {
+            minimumAmount = Math.min(presets[0], presets[1]);
+          } else if (presets && amount) {
+            minimumAmount = Math.min(amount, presets[0]);
+          }
+
+          return queryInterface.sequelize.query(
+            `
+          UPDATE "Tiers"
+            SET "minimumAmount" = :minimumAmount
+          WHERE "id" = :tierId
+        `,
+            {
+              replacements: {
+                minimumAmount,
+                tierId: tier.id,
+              },
+            },
+          );
+        });
+      })
+      .then(() => {
+        console.log('>>> Done!');
+      })
+      .catch(async err => {
+        // Remove the table if any error occur in the process
+        // to prevent having inconsistent data
+        await queryInterface.removeColumn('Tiers', colName);
+        throw err;
+      });
+  },
+
+  down: async queryInterface => {
+    await queryInterface.removeColumn('Tiers', colName);
+  },
+};
--- a/migrations/20190418120035-add_minmumAmount_column_to_tiers.js
+++ b/migrations/20190418120035-add_minmumAmount_column_to_tiers.js
@@ -27,9 +27,9 @@ module.exports = {
           if (!presets && amount) {
             minimumAmount = amount;
           } else if (presets && !amount) {
-            minimumAmount = Math.min(presets[0], presets[1]);
+            minimumAmount = Math.min(...presets);
           } else if (presets && amount) {
-            minimumAmount = Math.min(amount, presets[0]);
+            minimumAmount = Math.min(amount, ...presets);
           }
 
           return queryInterface.sequelize.query(
--- a/migrations/20190418120035-add_minmumAmount_column_to_tiers.js
+++ b/migrations/20190418120035-add_minmumAmount_column_to_tiers.js
@@ -24,13 +24,11 @@ module.exports = {
           const presets = tier.presets;
           const amount = tier.amount;
           let minimumAmount = null;
-          if (!presets && amount) {
-            minimumAmount = amount;
+          if (presets && presets.length > 0 && amount) {
+            minimumAmount = Math.min(amount, ...presets);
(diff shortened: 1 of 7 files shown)


migrations/20190418120035-add_minmumAmount_column_to_tiers.js
```diff
@@ -0,0 +1,64 @@
...
+      .then(results => {
+        const tiers = results[0];
+        return map(tiers, tier => {
+          const presets = tier.presets;
+          const amount = tier.amount;
+          let minimumAmount = null;
+          if (!presets && amount) {
+            minimumAmount = amount;
+          } else if (presets && !amount) {
+            minimumAmount = Math.min(presets[0], presets[1]);
```
znarf: Why `presets[0], presets[1]` and not all presets?

znarf:
@flickz Can you also edit `amountStr` in `Tier.js`?

znarf:
@flickz let's make it clear that we're using the smallest preset instead of assuming that `presets[0]` is the smallest. Unless there is a guarantee that `presets[0]` is the smallest of course.

server/graphql/v1/mutations/orders.js
```diff
@@ -393,7 +393,7 @@ export async function createOrder(order, loaders, remoteUser, reqIp) {
 
     // If using a tier, amount can never be less than the minimum amount
     if (tier && tier.presets) {
-      const minValue = min(isNil(tier.amount) ? tier.presets : [...tier.presets, tier.amount]);
+      const minValue = tier.minimumAmount || min(isNil(tier.amount) ? tier.presets : [...tier.presets, tier.amount]);
```
znarf: Can we stick to just use `tier.minimumAmount` here and assume it's set?

flickz: Yes we can

server/models/Tier.js
```diff
@@ -69,6 +69,10 @@ export default function(Sequelize, DataTypes) {
         type: DataTypes.ARRAY(DataTypes.INTEGER),
       },
 
+      minimumAmount: {
+        type: DataTypes.INTEGER,
```
Betree: I recently made changes to enforce more restrictive validations on models, like in https://github.com/opencollective/opencollective-api/pull/1776/files#diff-768438be827d00a869e92ca592010413R201

Here we know that min amount should never be less than `0`.

migrations/20190418120035-add_minmumAmount_column_to_tiers.js
```diff
@@ -0,0 +1,64 @@
...
+import { map } from 'bluebird';
+const colName = 'minimumAmount';
+
+/**
+ * Add a `minimumAmount` column to tiers to remove ambiguity.
+ */
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    const colParams = {
+      type: Sequelize.INTEGER,
```
Betree: `null` being a valid value for this column, is there any benefit for us to make a difference between `minAmount=0` an `minAmount=null`?

server/models/Tier.js
```diff
@@ -165,13 +169,7 @@ export default function(Sequelize, DataTypes) {
...
 
         amountStr() {
-          let str;
-          if (this.presets) {
-            const minimumAmount = Math.min(this.presets[0], this.amount);
-            str = `${formatCurrency(minimumAmount, this.currency)}+`;
-          } else {
-            str = formatCurrency(this.amount, this.currency);
-          }
+          let str = `${formatCurrency(this.minimumAmount, this.currency)}+`;
```
znarf: What happens if it's `null`?

@Betree Generally, should we assume it's `100` cents in the current currency.

Betree: I'm not sure, some collectives offer a free tier to get people into specific group. But we could put `100` as default in the frontend

znarf:
No change to `TierInputType`? Is that tested?

znarf:
We need also to support `minimumAmount` everywhere we create Tiers automatically such as https://github.com/opencollective/opencollective-api/blob/master/server/models/Collective.js#L72

server/models/Tier.js
flickz: @Betree & @znarf Should I add this?

znarf:
@flickz To be able to merge the API while being compatible with the old Frontend, we need a last small change.

In the mutation, if `minimumAmount` is not sent, we should set it to `minimumAmount = Math.min(amount, ...presets);`. Can you do that?

flickz:
Yes, I'll do that now

znarf:
@flickz Sorry, I was actually asking about the mutation to editTiers.

flickz:
Oh alright, I'll change that.

migrations/20190418120035-add_minmumAmount_column_to_tiers.js
Betree: To be clearer, is `allowNull=true` (the default) useful for us, or should we set `allowNull=false`?

server/graphql/v1/mutations/tiers.js
```diff
@@ -6,6 +7,13 @@ export function editTiers(_, args, req) {
     throw new errors.Unauthorized('You need to be logged in to edit tiers');
   }
 
+  const tiers = map(args.tiers, tier => {
+    if (tier.presets && !tier.minimumAmount) {
```
znarf: There should be a `minimumAmount` even if tier.presets is empty (= `amount`).

server/graphql/v1/mutations/tiers.js
```diff
@@ -6,6 +7,15 @@ export function editTiers(_, args, req) {
...
```
znarf: I find there are some ambiguity in the current code, the following one would be clearer:

```
if (!tier.minimumAmount) {
  if (tier.presets && tier.presets.length > 0 && tier.amount) {
     tier.minimumAmount = Math.min(tier.amount, ...tier.presets);
  } else if (tier.presets && tier.presets.length > 0) {
    tier.minimumAmount = Math.min(...tier.presets);
  } else if (tier.amount) {
    tier.minimumAmount = tier.amount;
  }
}
 ```

What do you think?

Comment (znarf) [reply server/graphql/v1/mutations/tiers.js]: Also make sure the logic exactly match the one in the migration for consistency.

File: server/graphql/v1/mutations/tiers.js [hunk 8]
```diff
@@ -6,6 +7,19 @@ export function editTiers(_, args, req) {
     throw new errors.Unauthorized('You need to be logged in to edit tiers');
   }
 
+  const tiers = map(args.tiers, tier => {
+    if (!tier.minimumAmount) {
```
znarf: Why would `tier.minimumAmount` be falsey?

I see 2 cases:

- frontend is not updated yet and the behavior is right
- frontend is updated, is sending minimumAmount but it's empty because the admin didn't define a minimumAmount. In that case, I don't think we should re-process it.

flickz: We should throw an error instead?

znarf: Why would we throw an error? Is it mandatory?

flickz: No it is not but we also can't have it as `null` since `amount` is mandatory.

znarf: Once the feature is available on frontend, we should not mess with user submitted values.

If the user decide the minimumAmount is null, it should stay null.

flickz: Alright, I can open another PR to remove that after they've both been deployed.

znarf: The API code should be compatible with the old and new versions of frontend.

server/models/Tier.js
znarf: Good feedback, implementing that myself. cc @flickz

znarf: By the way @Betree, why is `min` in `amount` not in a `validate` object?

```
      amount: {
        type: DataTypes.INTEGER, // In cents
        min: 0,
      },
```

server/models/Tier.js
znarf: We'll not implement that in this PR.

server/models/Tier.js
Betree: @znarf Seems like a mistake from https://github.com/opencollective/opencollective-api/commit/616100eedd229548224810f599539846d81ce9e3. I missed it while adding validations, I'll fix it myself

PR #1931, opened 2019-04-29
Fix host collectives GraphQL error

flickz:
This PR removes `dataValues` destructure causing the error and followed existing pattern used by other methods.
Ref([#1926](https://github.com/opencollective/opencollective/issues/1926))

### DIFF
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -355,8 +355,7 @@ const getCollectivesWithBalance = async (where = {}, options) => {
     sequelize.query(`${sql(allFields)} LIMIT ${limit} OFFSET ${offset}`, params),
   ]);
 
-  const total = totalResult ? get(totalResult, 'dataValues.total') : 0;
-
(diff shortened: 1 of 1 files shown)


server/lib/queries.js
```diff
@@ -350,18 +350,13 @@ const getCollectivesWithBalance = async (where = {}, options) => {
...
-      },
-    ],
-    collectives,
-  ] = await Promise.all([
+  const [[totalResult], collectives] = await Promise.all([
     sequelize.query(`${sql('COUNT(c.*) OVER() as "total"')} LIMIT 1`, params),
     sequelize.query(`${sql(allFields)} LIMIT ${limit} OFFSET ${offset}`, params),
   ]);
 
+  const total = totalResult ? get(totalResult, 'dataValues.total') : 0;
```
Betree: `get` knows how to deal with `undefined`, so this can also be written without ternary check: 

```es6
const total = get(totalResult, 'dataValues.total', 0)
```

PR #1958, opened 2019-05-07
Fix missing account transactions in GraphQL API v2

znarf:


znarf:
Merging without having tests passing because the breaking tests are not related to the change. This  will have to be addressed later anyway.

PR #1968, opened 2019-05-10
Applications management improvements

Betree:
Related to https://github.com/opencollective/opencollective-frontend/pull/1749

- Remove `updateApplication` query (was unused)
- Add a limit on the number of applications per user (default=15)

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -44,6 +44,7 @@
       "update": 10,
       "confirmToken": 15
     },
+    "maxNumberOfAppsPerUser": 15,
     "maxWebhooksPerUserPerCollective": 50
   },
   "email": {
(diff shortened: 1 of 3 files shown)


znarf:
This was from before, but there is some confusion between using `createdByUser` and `CollectiveId`. To be future proof, I think we should rely more on `CollectiveId` than `createdByUser`, it's a cleaner relationship.

For:
- checking `maxNumberOfAppsPerUser`
- when retrieving in `CollectiveInterface` -> `UserCollectiveType` -> `applications`
- maybe in the the access control too , what is today `req.remoteUser.id !== app.CreatedByUserId`

znarf:
Also, while you're cleaning, there is a whole REST API to manage applications that could be deleted if we're not using it. Check `routes.js` and `controllers/applications.js`.

Betree:
I've removed the REST endpoints, do you know what they were used for?

Regarding https://github.com/opencollective/opencollective-api/pull/1968#issuecomment-491336286 I feel like this is larger change that could have side effects, I'd be more comfortable to do it in another PR. Is it ok if I extract it to a separate issue?

PR #1976, opened 2019-05-13
Generate emails with service

znarf:


znarf:
We will look at that again later.

PR #1979, opened 2019-05-14
make sure the hostFeePercent matches the hostFeePercent of the host

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/2000

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1309,7 +1309,7 @@ export default function(Sequelize, DataTypes) {
   };
 
   Collective.prototype.updateHostFee = async function(hostFeePercent, remoteUser) {
-    if (!hostFeePercent || !remoteUser || hostFeePercent === this.hostFeePercent) {
+    if (typeof hostFeePercent === undefined || !remoteUser || hostFeePercent === this.hostFeePercent) {
       return;
     }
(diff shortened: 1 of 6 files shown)


znarf:
@xdamman have you thought about running a migration to update the past entries?

znarf:
@xdamman Have you seen my comment? What's needed to complete this and merge?

xdamman:
I've seen your comment but I don't have the time to write such migration.
It's not straightforward. I've computed what each host needs to reimburse on the issue. I think it's good enough. 
It's more important to make sure we don't have that issue again in the future so let's merge and deploy this.

znarf:
So, let's say we merge that. How do we ensure that `hostFeePercent` is the right one in the database?

xdamman:
As documented in the issue, I already ran the query to update those `hostFeePercent` for two main hosts that were impacted, I can do it for the remaining ones once deployed.

znarf:
Running that automatically as a migration is the way to do it. No need to sync as a team, the query can be reviewed in the PR and it's executed exactly at the right time.

znarf:
@xdamman Slow feedback loop here :-/ Merging to avoid issues for future collectives. Please make sure the database is updated after deploy. Still think it would have been cleaner with a simple migration.

PR #1988, opened 2019-05-15
Add ability to create private updates

eloyekunle:
Related: opencollective/opencollective-frontend#1796

Ref: opencollective/opencollective#1923

### DIFF
--- a/migrations/20190515180630-updates-is-private.js
+++ b/migrations/20190515180630-updates-is-private.js
@@ -1,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.addColumn('Updates', 'isPrivate', {
+      type: Sequelize.BOOLEAN,
+      defaultValue: false,
+    });
+  },
+
+  down: (queryInterface, Sequelize) => {
+    return queryInterface.removeColumn('Updates', 'isPrivate');
+  },
+};
--- a/migrations/20190515180630-updates-is-private.js
+++ b/migrations/20190515180630-updates-is-private.js
@@ -2,13 +2,22 @@
 
 module.exports = {
   up: (queryInterface, Sequelize) => {
-    return queryInterface.addColumn('Updates', 'isPrivate', {
-      type: Sequelize.BOOLEAN,
-      defaultValue: false,
-    });
+    return queryInterface
+      .addColumn('Updates', 'isPrivate', {
+        type: Sequelize.BOOLEAN,
+        defaultValue: false,
+      })
+      .then(() => {
+        queryInterface.addColumn('UpdateHistories', 'isPrivate', {
+          type: Sequelize.BOOLEAN,
+          defaultValue: false,
+        });
+      });
   },
 
   down: (queryInterface, Sequelize) => {
-    return queryInterface.removeColumn('Updates', 'isPrivate');
+    return queryInterface.removeColumn('Updates', 'isPrivate').then(() => {
+      queryInterface.removeColumn('UpdateHistories', 'isPrivate');
+    });
   },
 };
--- a/migrations/20190515180630-updates-is-private.js
+++ b/migrations/20190515180630-updates-is-private.js
@@ -8,7 +8,7 @@ module.exports = {
         defaultValue: false,
       })
       .then(() => {
-        queryInterface.addColumn('UpdateHistories', 'isPrivate', {
+        return queryInterface.addColumn('UpdateHistories', 'isPrivate', {
           type: Sequelize.BOOLEAN,
           defaultValue: false,
         });
--- a/migrations/20190515180630-updates-is-private.js
+++ b/migrations/20190515180630-updates-is-private.js
@@ -11,6 +11,7 @@ module.exports = {
         return queryInterface.addColumn('UpdateHistories', 'isPrivate', {
           type: Sequelize.BOOLEAN,
           defaultValue: false,
+          allowNull: false,
         });
       });
   },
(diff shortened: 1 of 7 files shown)


znarf:
@eloyekunle Why have you chosen the "stripContent" approach instead of just restricting access to the update? Is there a requirement to see the title?

eloyekunle:
@znarf From the original specifications and discussions in the issue, there's a requirement to show the title while hiding the actual content.

znarf:
@eloyekunle Great, got it. 

What about implementing the permission to read the content in the GraphQL getter instead of manipulating the object with `stripContent`?

eloyekunle:
@znarf I'll look into that.

eloyekunle:
@znarf I've removed the `stripContent` approach in favor of modifying GraphQL getters.

server/models/Update.js
```diff
@@ -121,6 +121,16 @@ export default function(Sequelize, DataTypes) {
 
       image: DataTypes.STRING,
 
+      isPrivate: {
+        type: DataTypes.BOOLEAN,
+        defaultValue: false,
+      },
+
+      contentStripped: {
+        type: new DataTypes.VIRTUAL(DataTypes.BOOLEAN),
```
znarf: What is `DataTypes.VIRTUAL`? I can't find documentation on it.

server/models/Update.js
```diff
@@ -121,6 +121,16 @@ export default function(Sequelize, DataTypes) {
 
       image: DataTypes.STRING,
 
+      isPrivate: {
+        type: DataTypes.BOOLEAN,
+        defaultValue: false,
+      },
+
+      contentStripped: {
```
znarf: Maybe prefix with `is`. I'm wondering about a better name but I'm not decided yet, ie: `isContentAvailable`. Let's see what @Betree thinks about that.

server/models/Update.js
eloyekunle: http://docs.sequelizejs.com/class/lib/data-types.js~VIRTUAL.html

It's used when the property is not backed by a database column.

server/models/Update.js
eloyekunle: Yes, `isContentAvailable` is a better name.

server/models/Update.js
znarf: What's the benefit of exposing that in the model? (compared to having it only on GraphQL side).

eloyekunle: No benefit, I added it when I was manipulating with `stripContent`, so it can be removed now.

server/models/Update.js
eloyekunle: I've renamed it to `isContentAvailable`.

server/models/Update.js
eloyekunle: I've removed it.

znarf:
@Betree ok for you?

migrations/20190515180630-updates-is-private.js
```diff
@@ -0,0 +1,23 @@
...
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface
+      .addColumn('Updates', 'isPrivate', {
+        type: Sequelize.BOOLEAN,
+        defaultValue: false,
+      })
+      .then(() => {
+        queryInterface.addColumn('UpdateHistories', 'isPrivate', {
```
Betree:
```suggestion
        return queryInterface.addColumn('UpdateHistories', 'isPrivate', {
```

So sequelize can properly await for the migration to be done

migrations/20190515180630-updates-is-private.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface
+      .addColumn('Updates', 'isPrivate', {
+        type: Sequelize.BOOLEAN,
+        defaultValue: false,
```
Betree: If we're not interested in `null` then let's add a `allowNull: false` on the column

server/graphql/v1/mutations/updates.js
```diff
@@ -27,6 +27,7 @@ export async function createUpdate(_, args, req) {
     markdown,
     html: strip_tags(args.update.html),
     CollectiveId,
+    isPrivate: !!args.update.isPrivate,
```
Betree: Is this `!!` necessary?

server/graphql/v1/types.js
```diff
@@ -658,6 +658,18 @@ export const UpdateType = new GraphQLObjectType({
           return update.image;
         },
       },
+      isPrivate: {
+        type: GraphQLBoolean,
+        resolve(update) {
+          return update.isPrivate;
+        },
+      },
+      isContentAvailable: {
```
Betree: `isContentAvailable` doesn't seems accurate to me if this is about permission. Maybe `isAllowedToSee`, or `userCanSeeUpdate` would be better. Whether you want to keep the actual or prefer to change, there should be a `description` on the field to be explicit about what it does.

Betree:
Let's merge this ASAP so we can properly test https://github.com/opencollective/opencollective-frontend/pull/1796

migrations/20190515180630-updates-is-private.js
eloyekunle: Added.

server/graphql/v1/mutations/updates.js
eloyekunle: No, it's not. I've removed it.

server/graphql/v1/types.js
eloyekunle: I've renamed 'isContentAvailable' to 'userCanSeeUpdate', and added a description.

PR #2002, opened 2019-05-17
Remove edit permission for collective creator

znarf:
Fix: https://github.com/opencollective/opencollective/issues/1992


server/graphql/v1/mutations/collectives.js
```diff
@@ -410,10 +410,7 @@ export function editCollective(_, args, req) {
           req.remoteUser.hasRole(['ADMIN', 'HOST', 'BACKER'], parentCollective.id)
         );
       } else {
-        return (
-          req.remoteUser.id === collective.CreatedByUserId ||
-          req.remoteUser.hasRole(['ADMIN', 'HOST'], newCollectiveData.id)
-        );
+        return req.remoteUser.isAdmin(newCollectiveData.id);
```
znarf: This is the fix.

PR #2005, opened 2019-05-18
CI: update setup

znarf:
- use `opencollective` instead of `ubuntu` user

### DIFF
--- a/config/circleci.json
+++ b/config/circleci.json
@@ -1,6 +1,6 @@
 {
   "database": {
-    "url": "postgres://ubuntu@localhost:5432/opencollective_dvl",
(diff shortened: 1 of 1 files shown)


eloyekunle:
@znarf Seems this also needs to be done in the frontend.

znarf:
@eloyekunle it was done and merged at the same time: https://github.com/opencollective/opencollective-frontend/pull/1808

eloyekunle:
I'm getting a CI error here: https://circleci.com/gh/opencollective/opencollective-frontend/21574

znarf:
@eloyekunle It's because the matching API branch was not updated. We should deleted it instead.

eloyekunle:
Oh okay, will do that. Thanks.

PR #2009, opened 2019-05-21
Tier API validation

flickz:
In the effort to validate `tiers`, this PR:

- Add `amountType` (as `ENUM` of `('FLEXIBLE', 'FIXED')`) to `Tier` column and migration script that sets the value of `amountType` to `FIXED` when `tier.presets` is `NULL` and `FLEXIBLE` if otherwise.

- Adds validation to tiers, for each tiers, ensure the "default amount" (amount) is one of the presets when the amount type is "flexible amount" and there is no presets smaller than the "minimum amount" (minimum_amount)

Ref [#1905](https://github.com/opencollective/opencollective/issues/1905)

### DIFF
--- a/migrations/20190509094659-add-amountType-column-to-tiers.js
+++ b/migrations/20190509094659-add-amountType-column-to-tiers.js
@@ -1,0 +1,58 @@
+'use strict';
+import { map } from 'bluebird';
+
+const colName = 'amountType';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    const colParams = {
+      type: Sequelize.ENUM('FLEXIBLE', 'FIXED'),
+    };
+
+    return queryInterface
+      .addColumn('Tiers', colName, colParams)
+      .then(() => {
+        return queryInterface.sequelize.query(`
+          SELECT * FROM "Tiers"
+        `);
+      })
+      .then(results => {
+        const tiers = results[0];
+        return map(tiers, tier => {
+          const presets = tier.presets;
+          let amountType;
+          if (presets) {
+            amountType = 'FLEXIBLE';
+          } else {
+            amountType = 'FIXED';
+          }
+          return queryInterface.sequelize.query(
+            `
+              UPDATE "Tiers"
+                SET "amountType" = :amountType
+              WHERE "id" = :tierId
+            `,
+            {
+              replacements: {
+                amountType,
+                tierId: tier.id,
+              },
+            },
+          );
+        });
+      })
+      .then(() => {
+        console.log('>>> Done!');
+      })
+      .catch(async err => {
+        // Remove the table if any error occur in the process
+        // to prevent having inconsistent data
+        await queryInterface.removeColumn('Tiers', colName);
+        throw err;
+      });
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.removeColumn('Tiers', colName);
+  },
+};
--- a/migrations/20190509094659-add-amountType-column-to-tiers.js
+++ b/migrations/20190509094659-add-amountType-column-to-tiers.js
@@ -13,33 +13,17 @@ module.exports = {
       .addColumn('Tiers', colName, colParams)
       .then(() => {
         return queryInterface.sequelize.query(`
-          SELECT * FROM "Tiers"
+          UPDATE "Tiers"
+            SET "amountType" = 'FIXED'
+          WHERE "presets" IS NULL
         `);
       })
-      .then(results => {
-        const tiers = results[0];
-        return map(tiers, tier => {
-          const presets = tier.presets;
-          let amountType;
-          if (presets) {
-            amountType = 'FLEXIBLE';
-          } else {
-            amountType = 'FIXED';
-          }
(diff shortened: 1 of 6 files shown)


server/graphql/v1/mutations/collectives.js
```diff
@@ -380,6 +380,23 @@ export function editCollective(_, args, req) {
...
+  forEach(tiers, tier => {
+    const presets = tier.presets || [];
+    const amount = tier.amount;
+    const amountType = tier.amountType;
+    const minPreset = Math.min(...presets);
+    if (amountType === 'FLEXIBLE' && presets.indexOf(amount) === -1) {
+      throw new Error(`In ${tier.name}'s tier "Default amount" must be one of suggested values amounts`);
+    }
+
+    if (amountType === 'FLEXIBLE' && minPreset < tier.minimumAmount) {
```
znarf: What if `tier.minimumAmount` is not set?

server/graphql/v1/mutations/collectives.js
```diff
@@ -380,6 +380,37 @@ export function editCollective(_, args, req) {
 
   let collective, parentCollective;
 
+  const tiers = args.collective.tiers;
+  forEach(tiers, tier => {
+    const presets = tier.presets || [];
+    const amount = tier.amount;
+    const name = tier.name;
+    const amountType = tier.amountType;
```
znarf: What if `amountType` is not sent by the client?

Should we make it mandatory to deploy frontend first or can we have a simple transition strategy (better)?

flickz: The backend should be deployed first, since we can easily determine `amountType`  from the tier data without being sent by the frontend. 
Basically I just need to add:
```js
tier.amountType = tier.presets? 'FLEXIBLE' : 'FIXED'
```

znarf: Ok, let's do that.

server/graphql/v1/mutations/collectives.js
```diff
@@ -380,6 +380,42 @@ export function editCollective(_, args, req) {
...
+  forEach(tiers, tier => {
+    const presets = tier.presets;
+    const amount = tier.amount;
+    const name = tier.name;
+
+    if (!tier.amountType) {
+      tier.amountType = presets ? 'FLEXIBLE' : 'FIXED';
+    }
+
+    const amountType = tier.amountType;
```
znarf: Just a suggestion

```suggestion
    // if amountType is not passed we guess it for retro-compatibility
    const amountType = tier.amountType || (presets ? 'FLEXIBLE' : 'FIXED');
```

migrations/20190509094659-add-amountType-column-to-tiers.js
```diff
@@ -0,0 +1,58 @@
...
+    };
+
+    return queryInterface
+      .addColumn('Tiers', colName, colParams)
+      .then(() => {
+        return queryInterface.sequelize.query(`
+          SELECT * FROM "Tiers"
+        `);
+      })
+      .then(results => {
```
znarf: Like in previous similar PR, it's better to skip the model and do a single SQL query directly.

server/graphql/v1/mutations/collectives.js
```diff
@@ -380,6 +380,42 @@ export function editCollective(_, args, req) {
...
+
+    if (!name || name.trim().length === 0) {
+      throw new Error('Name field is required for all tiers');
+    }
+
+    if (tier.type !== 'TICKET' && amountType === 'FIXED' && !amount) {
+      throw new Error(`In ${name}'s tier, "Amount" is required`);
+    }
+
+    if (amountType === 'FLEXIBLE' && !tier.minimumAmount) {
```
znarf: `minimumAmount` is not required. If it was to address my previous comment, I was just asking to check that "minimum amount cannot be less than minimum suggested amounts" only if a value is provided.

znarf:
Before merging, can we have a frontend PR that only cares about reporting backend errors?

As discussed, we will take care of full frontend validation later.

flickz:
This is already part of the `amountTypePR` https://github.com/opencollective/opencollective-frontend/pull/1823/commits/9e564fa7606169e42b3df39581dc8e4ffc109e24

Screenshot from 2019-05-21 13-28-10

znarf:
Ok. Fantastic.

server/graphql/v1/mutations/collectives.js
```diff
@@ -380,6 +380,38 @@ export function editCollective(_, args, req) {
...
+    }
+
+    const amountType = tier.amountType;
+    const minPreset = presets ? Math.min(...presets) : null;
+
+    if (!name || name.trim().length === 0) {
+      throw new Error('Name field is required for all tiers');
+    }
+
+    if (tier.type !== 'TICKET' && amountType === 'FIXED' && !amount) {
```
Betree: This validations should happen at the model level, this will not cover tiers edited with `editTier` nor `editTiers` mutations.

znarf: Good point.

PR #2019, opened 2019-05-23
Update GraphQL to v14

eloyekunle:
Related: https://github.com/opencollective/opencollective-frontend/pull/1846
Ref: opencollective/opencollective#1658

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6323,25 +6323,25 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": false,
+          "resolved": "",
           "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
           "optional": true
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "resolved": false,
+          "resolved": "",
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
           "optional": true
         },
         "aproba": {
           "version": "1.2.0",
-          "resolved": false,
+          "resolved": "",
           "integrity": "sha512-Y9J6ZjXtoYh8RnXVCMOU/ttDmk1aBjunq9vO0ta5x85WDQiQfUF9sIPBITdbiiIVcBo03Hi3jMxigBtsddlXRw==",
           "optional": true
         },
         "are-we-there-yet": {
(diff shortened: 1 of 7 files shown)


eloyekunle:
@znarf 

Here's why the Dates will always be coerced to their Unix time representation: https://github.com/graphql/graphql-js/blob/d4140766e0d2add0f82b22fe6e6354de4afc0f3c/src/type/scalars.js#L125.

I'll be using a custom GraphQL scalar type to represent the datetimes.

znarf:
@eloyekunle Looks like a great solution! Do you understand why the `test-bdd` test are not passing?

eloyekunle:
It's related to expenses. I'm looking into it now.

eloyekunle:
@znarf It's fixed now.

znarf:
@eloyekunle seems like a warning of a breaking change in the API.

What is change in the GraphQL package that is triggering that? Can that be fixed in the API instead of updating the test?

znarf:
I pushed a matching branch on frontend to test if the update is breaking anything in e2e tests.

https://circleci.com/gh/opencollective/opencollective-frontend/tree/feat%2Fupdate-graphql

eloyekunle:
The breaking change was made here: graphql/graphql-js#1382.

It means cases where the frontend sends `Int` values as `String` need to be changed.

znarf:
@eloyekunle Can the breaking change fixed in the API instead of updating the test?

eloyekunle:
@znarf 

Trying to fix it means creating a new scalar type for `Integers` that will also accept `Strings` and coerce them to `Integers`.

I've recognized two instances affected in the frontend and opened a PR for it: https://github.com/opencollective/opencollective-frontend/pull/1846.

znarf:
@eloyekunle Thank you for your answer. Hope our coverage is great! 🤞

eloyekunle:
The main things to look out for are cases where the Frontend is currently sending `Integers` as `Strings`. In such cases, the `String` should be parsed to `Integer` first, before sending to the API.

I'm sure there'll only be a few (if any) such cases left.

znarf:
@eloyekunle What's the behavior if we don't? It's triggering a GraphQL error in the new version?

eloyekunle:
Yes, it triggers an error.
Something like:

```
GraphQL v1 error: Variable "$id" got invalid value "10890"; Expected type Int; Int cannot represent non-integer value: "10890"
```

eloyekunle:
Ping @znarf, What's left to merge this?

znarf:
@eloyekunle It's good. I usually prefer to merge and deploy changes like that early in the week.

PR #2026, opened 2019-05-24
WIP: Automatically make a private update public after a certain period

brymut:
Add a field to support automatic change of a private update to public on a specific
date.


brymut:
CC: opencollective/opencollective#2058

brymut:
Hi @Betree, I've managed to get something going for the API side of the spec for opencollective/opencollective#2058 could I please get some feedback on this when you can?

cron/daily/make-private-updates-public.js
```diff
@@ -0,0 +1,45 @@
...
+
+const today = new Date(new Date().setUTCHours(0, 0, 0, 0)).toISOString();
+logger.info(`Updates for ${today}`);
+
+// get all the private updates with the makePublicOn <= today
+// and make them public
+
+let updates = 0;
+
+models.Update.findAll({
```
Betree: Updating row per row makes the code pretty verbose and it's not performant: if we have 100 updates to do, we'll trigger 100 SQL queries.

A more concise way to write this would be to do something like:

```es6
models.Update.update(
  {
    isPrivate: false,
  }, 
  {
    where: {
      isPrivate: true,
      makePublicOn: { [Op.lte]: today },
    }
  }
).then(([affectedCount]) => {
  logger.info(`Number of private updates made public: ${affectedCount}`)
});
```

See https://docs.sequelizejs.com/manual/instances.html#working-in-bulk--creating--updating-and-destroying-multiple-rows-at-once-

brymut: Thanks for this, I missed the bulk operations part of the docs, should've looked that up earlier, I'll make this change asap.

brymut:
Hi @Betree, may have another quick look at this?

brymut:
Rebased this branch onto the latest master.

test/utils.js
```diff
@@ -183,6 +185,21 @@ export const separator = length => {
   console.log(`\n${separator}\n`);
 };
 
+export const makePublic = () => {
```
Betree: Let's have a more explicit name for this one, like `makeUpdatePublic`.

brymut: will change that ASAP

test/utils.js
```diff
@@ -183,6 +185,21 @@ export const separator = length => {
...
+    execOptions,
+    (error, stdout, stderr) => {
+      if (error) {
+        logger.error(stderr);
+        throw error;
+      }
+      logger.verbose(stdout);
+    },
+  );
+};
```
znarf: I would prefer if we would just do something like `models.Update.makePrivateUpdatesPublic()`

cron/daily/make-private-updates-public.js
```diff
@@ -0,0 +1,29 @@
...
+// get the ISOtime for the day @ midnight to compare
+// against the update.makePublic on time field
+
+const today = new Date(new Date().setUTCHours(0, 0, 0, 0)).toISOString();
+logger.info(`Updates for ${today}`);
+
+// get all the private updates with the makePublicOn <= today
+// and make them public
+
+models.Update.update(
```
znarf: I would extract that to the model to be easily available elsewhere, ie: `models.Update.makePrivateUpdatesPublic`

brymut:
Hi @znarf, I'm sorry I didn't notice this mix up earlier but the review I had requested was for https://github.com/opencollective/opencollective-api/pull/2128#event-2427756443

cron/daily/make-private-updates-public.js
brymut: extracted to a new Update method. (.makePublic())

test/utils.js
znarf: @brymut Can you address this comment?

brymut: Hi, I'm sorry. I don't know how I missed this.

Betree:
Done in https://github.com/opencollective/opencollective-api/pull/2563

PR #2034, opened 2019-05-27
Add sign in with publicKey

friedger:
This PR
- adds field `publicKey` to UserInputType
- adds `redirect` to sign in requests if the user input comes with a `publicKey` value. The `redirect` value is the encrypted redirect link.

This is a proof of concept towards 
https://github.com/opencollective/opencollective/issues/1749

Corresponding UI looks like this:
https://github.com/opencollective/opencollective-frontend/pull/1833

Probably it is not necessary to include blockstack.js as only function `encryptContent` is needed. 

After sign in the user name contains the blockstack id of the user:
Screenshot from 2019-05-27 11-07-28

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "babel-plugin-add-module-exports": "1.0.2",
     "babel-plugin-lodash": "3.3.4",
     "bcrypt": "3.0.6",
+    "blockstack": "^19.1.0",
     "bluebird": "3.5.5",
     "body-parser": "1.19.0",
     "child-process-promise": "2.2.1",
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -21,11 +21,35 @@ export const exists = async (req, res) => {
   if (!isValidEmail(email)) {
     return res.send({ exists: false });
   } else {
-    const user = await models.User.findOne({
-      attributes: ['id'],
-      where: { email },
+    const publicKey = req.query.publicKey;
+    if (publicKey) {
+      const user = await models.User.findOne({
+        attributes: ['id', 'email'],
+        where: { publicKey },
+      });
+      const exists = user && user.email === email;
+      return res.send({ exists: Boolean(exists) });
+    } else {
+      const user = await models.User.findOne({
+        attributes: ['id'],
+        where: { email },
+      });
+      return res.send({ exists: Boolean(user) });
+    }
+  }
+};
+
+const findOne = user => {
+  if (user.publicKey) {
+    return models.User.findOne({ where: { publicKey: user.publicKey } }).then(u => {
+      if (u.email === user.email.toLowerCase()) {
+        return u;
+      } else {
+        return null;
+      }
     });
-    return res.send({ exists: Boolean(user) });
+  } else {
+    return models.User.findOne({ where: { email: user.email.toLowerCase() } });
   }
 };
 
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -27,6 +27,7 @@
     "babel-plugin-add-module-exports": "1.0.2",
     "babel-plugin-lodash": "3.3.4",
     "bcrypt": "3.0.6",
+    "blockstack": "^19.1.0",
```
znarf: You said integrating this `blockstack` library might not be mandatory. What would be the alternative? If it's simple, then let's do that.

server/controllers/users.js
```diff
@@ -34,19 +59,25 @@ export const exists = async (req, res) => {
...
-      return emailLib.send('user.new.token', u.email, { loginLink }, { sendEvenIfNotProduction: true });
+      emailLib.send('user.new.token', u.email, { loginLink }, { sendEvenIfNotProduction: true });
+      if (user.publicKey) {
+        loginLink = blockstackLib.encryptLink(user.publicKey, loginLink);
+      }
     })
     .then(() => {
       const response = { success: true };
       // For e2e testing, we enable testuser+(admin|member)@opencollective.com to automatically receive the login link
-      if (process.env.NODE_ENV !== 'production' && user.email.match(/.*test.*@opencollective.com$/)) {
```
znarf: If there is no e2e tests, then let's not do that.

server/controllers/users.js
```diff
@@ -34,19 +59,25 @@ export const exists = async (req, res) => {
...
+      if (
+        (process.env.NODE_ENV !== 'production' && user.email.match(/.*test.*@opencollective.com$/)) ||
+        user.publicKey
```
znarf: If there is no e2e tests, then let's not modify that.

server/controllers/users.js
```diff
@@ -34,19 +59,25 @@ export const exists = async (req, res) => {
 export const signin = (req, res, next) => {
```
znarf: At this stage, can you have a specific `signin` endpoint for blockstack? And use a dedicated `exists` method?

To minimize footprint in `server/controllers/users.js`, supporting functions may be loaded from `lib/blockstack.js`

server/graphql/v1/mutations.js
```diff
@@ -204,12 +206,25 @@ const mutations = {
           throw new Error('User already exists for given email');
         }
 
+        if (
```
znarf: I don't think it's useful to do any modification in these GraphQL queries. The `signin` endpoint should be enough.

friedger: This is for creating a new profile with blockstack.

znarf: Ok, sorry, I was a bit wrong. `createUser` was indeed used for signups.

But, please handle that in a separate endpoint (ie: blockstackSignin / blockstackSignup), or a separated GraphQL mutation (ie: createUserWithBlockstack).

package.json
friedger: I had a look but it is a bit more complicated.

server/graphql/v1/mutations.js
friedger: I have re-added it and moved everything into separate methods.

friedger:
I have added the updateUserPublicKey mutation to better handle existing users

friedger:
Can I do more to get this one merged?

znarf:
@friedger nothing on the API side, this is more on the frontend side that we had open questions. We need to have another look and test again all the different cases.

znarf:
This was pretty good work and I'm sorry we let this PR go stale.

Unfortunately, after discussing it with the team last week, there was no consensus to move it forward and I'll have to close it.

This could be revisited once supporting multiple identity providers become a priority.

PR #2040, opened 2019-05-28
Enable host collective admin to refund transaction 

flickz:
Following [#1864](https://github.com/opencollective/opencollective-frontend/pull/1864), this PR allows collective host admin to refund transaction.

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -766,16 +766,22 @@ export async function refundTransaction(_, args, req) {
   const transaction = await models.Transaction.findByPk(args.id, {
     include: [models.Order, models.PaymentMethod],
   });
+
   if (!transaction) {
     throw new errors.NotFound({ message: 'Transaction not found' });
   }
 
+  const collective = await models.Collective.findByPk(transaction.CollectiveId);
+  const isHost = await collective.isHost();
+  const HostCollectiveId = isHost ? collective.id : await collective.getHostCollectiveId();
+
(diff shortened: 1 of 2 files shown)


server/graphql/v1/mutations/orders.js
```diff
@@ -766,16 +766,22 @@ export async function refundTransaction(_, args, req) {
...
     include: [models.Order, models.PaymentMethod],
   });
+
   if (!transaction) {
     throw new errors.NotFound({ message: 'Transaction not found' });
   }
 
+  const collective = await models.Collective.findByPk(transaction.CollectiveId);
+  const isHost = await collective.isHost();
+  const HostCollectiveId = isHost ? collective.id : await collective.getHostCollectiveId();
```
znarf: Is that implemented like something elsewhere? In which case would the `transaction.CollectiveId` be a host?

flickz: Yeah in the case of adding fund manually, noticed that from inspecting this issue  https://github.com/opencollective/opencollective/issues/2042 locally. 

Also I think manual fund transactions should not have refund button shown, because I got the error below when I attempt to refund transaction created from "Add fund" form, was planning to open an issue regarding that.

```
 [GraphQL error]: Message: paymentMethod.refundTransaction is not a function,
Location: [object Object], Path: refundTransaction
```

znarf: There is an issue for that I think (payment methods lacking refundTransaction).

PR #2044, opened 2019-05-29
Add `addFundsToCollective` mutation resolver 

flickz:
This PR attempts to fix [#2042](https://github.com/opencollective/opencollective/issues/2042) by creating a separate mutation resolver for "Add funds" form instead of the congested `createOrder` currently used. 

The resolver ensure the following that  LoggedIn user is a root or host collective admin and the operation is performed regardless of `order.user.email` existence.

### DIFF
--- a/server/graphql/v1/mutations.js
+++ b/server/graphql/v1/mutations.js
@@ -17,6 +17,7 @@ import {
   updateSubscription,
   refundTransaction,
   addFundsToOrg,
+  addFundsToCollective,
   completePledge,
   markOrderAsPaid,
   updateOrderInfo,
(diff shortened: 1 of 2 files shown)


znarf:
Is there anything that can be removed from `createOrder`now?

flickz:
My few attempts have at least one thing breaking, it's a very fragile function, might take a little while.

flickz:
@znarf  I think it's safe to remove [this](https://github.com/opencollective/opencollective-api/blob/master/server/graphql/v1/mutations/orders.js#L249-L263
) but I'd appreciate if you can also double check so I don't miss anything.

znarf:
@flickz Let's do it. And fix https://github.com/opencollective/opencollective/issues/1616

PR #2059, opened 2019-06-03
Remove COLLECTIVE_APPROVE email notification for Github

flickz:
This PR completes [#2068](https://github.com/opencollective/opencollective/issues/2068) fix.

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -316,7 +316,7 @@ export async function createCollectiveFromGithub(_, args, req) {
   const host = await models.Collective.findByPk(defaultHostCollective('opensource').CollectiveId);
   const promises = [
     collective.addUserWithRole(user, roles.ADMIN),
-    collective.addHost(host, user),
+    collective.addHost(host, user, { activity: activities.COLLECTIVE_CREATED_GITHUB }),
     collective.update({ isActive: true }),
   ];
 
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -316,7 +316,7 @@ export async function createCollectiveFromGithub(_, args, req) {
   const host = await models.Collective.findByPk(defaultHostCollective('opensource').CollectiveId);
   const promises = [
     collective.addUserWithRole(user, roles.ADMIN),
-    collective.addHost(host, user, { skipApplyActivity: true }),
+    collective.addHost(host, user, { skipCollectiveApplyActivity: true }),
     collective.update({ isActive: true }),
   ];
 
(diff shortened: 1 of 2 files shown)


server/models/Collective.js
```diff
@@ -1454,13 +1455,15 @@ export default function(Sequelize, DataTypes) {
...
           },
         };
-        promises.push(
-          models.Activity.create({
-            CollectiveId: this.id,
-            type: activities.COLLECTIVE_APPLY,
-            data,
-          }),
-        );
+        if (options.activity && options.activity !== activities.COLLECTIVE_CREATED_GITHUB) {
```
znarf: This is wrong. This will opt out everything that doesn't have an activity.

server/graphql/v1/mutations/collectives.js
```diff
@@ -316,7 +316,7 @@ export async function createCollectiveFromGithub(_, args, req) {
   const host = await models.Collective.findByPk(defaultHostCollective('opensource').CollectiveId);
   const promises = [
     collective.addUserWithRole(user, roles.ADMIN),
-    collective.addHost(host, user),
+    collective.addHost(host, user, { activity: activities.COLLECTIVE_CREATED_GITHUB }),
```
znarf: Be clearer about the intent of the option, just pass something like `skipApplyActivity`.

server/models/Collective.js
flickz: Oh didn't think it through, thanks.

server/graphql/v1/mutations/collectives.js
```diff
@@ -316,7 +316,7 @@ export async function createCollectiveFromGithub(_, args, req) {
   const host = await models.Collective.findByPk(defaultHostCollective('opensource').CollectiveId);
   const promises = [
     collective.addUserWithRole(user, roles.ADMIN),
-    collective.addHost(host, user),
+    collective.addHost(host, user, { skipApplyActivity: true }),
```
znarf:
```suggestion
    collective.addHost(host, user, { skipCollectiveApplyActivity: true }),
```

server/models/Collective.js
```diff
@@ -1454,13 +1455,16 @@ export default function(Sequelize, DataTypes) {
...
         };
-        promises.push(
-          models.Activity.create({
-            CollectiveId: this.id,
-            type: activities.COLLECTIVE_APPLY,
-            data,
-          }),
-        );
+
+        if (!options || !options.skipApplyActivity) {
```
znarf:
```suggestion
        if (!options || !options.skipCollectiveApplyActivity) {
```

znarf:
@flickz already integrated the proposed changes

PR #2063, opened 2019-06-04
Allow editing of "core contributor since" dates #2076

leonardohof:
Resolve opencollective/opencollective#2076


leonardohof:
https://github.com/opencollective/opencollective/issues/2076

PR #2074, opened 2019-06-05
Feat/add legal docs table

pietgeursen:
- [x] Adds a new table and model for `LegalDocuments`
- [x] Adds a new table and model for `RequiredLegalDocument`
- [x] Adds a cron job to send HelloWorks document requests to users who need to fill out a document.
- [x] Exposes some extra info in graphql on `Expense` so that hosts can avoid paying out users who need to fill in their forms. There's a new field `userTaxFormRequiredBeforePayment`.

A user collective can have many `LegalDocuments` (they can fill in many of them over the years and of different document types)

`LegalDocuments` has a composite unique constraint so that there will only be one record for a given combination of `year`, `CollectiveId`(the collective filling in the form) and `documentType`. 

A host collective can have many `RequiredLegalDocuments`.  
eg. Open Source Collective will have a required document of type `US_TAX_FORM`. Note that:
- it only has one at the moment but the data model supports adding more in the future.
- some other host might require the same document type and the user will not have to fill in another form.

Part of the work for [fixing how open source collective collects tax forms](https://github.com/opencollective/opencollective/issues/1774)


package.json
```diff
@@ -58,6 +58,7 @@
     "handlebars": "4.1.2",
     "hashids": "1.2.2",
     "he": "1.2.0",
+    "helloworks-sdk": "pietgeursen/helloworks-nodejs-sdk#patch-1",
```
pietgeursen: Waiting for this [pr](https://github.com/hellosign/helloworks-nodejs-sdk/pull/5)

server/lib/taxForms.js
```diff
@@ -0,0 +1,88 @@
...
+    year,
+    UserId,
+  });
+  const hasUserCompletedDocument = await LegalDocument.hasUserCompletedDocument({
+    documentType: US_TAX_FORM,
+    year,
+    user,
+  });
+
+  return requiredDocuments.length > 0 && isOverThreshold && !hasUserCompletedDocument;
```
Betree: Maybe we can avoid additional queries by checking things just after the queries?

```es6
const requiredDocuments = await host.getRequiredLegalDocuments({ ... })
if (!requiredDocuments.length) {
  return false;
}
```

server/lib/taxForms.js
```diff
@@ -0,0 +1,88 @@
...
+}
+
+export function SendHelloWorksTaxForm({ client, callbackUrl, workflowId, year }) {
+  return async function sendHelloWorksUsTaxForm(user) {
+    const participants = {
+      participant_swVuvW: {
+        type: 'email',
+        value: user.email,
+        fullName: `${user.firstName} ${user.lastName}`, // Is this an internationalisation problem?
```
Betree: I think it's ok

```suggestion
        fullName: `${user.firstName} ${user.lastName}`,
```

server/lib/taxForms.js
```diff
@@ -0,0 +1,88 @@
...
+          participants,
+          metadata: {
+            email: user.email,
+            year,
+          },
+        })
+        .then(() =>
+          LegalDocument.findOrCreate({ where: { documentType: US_TAX_FORM, year, CollectiveId: userCollective.id } }),
+        )
+        // .then(() => LegalDocument.findByTypeYearUser({ documentType: US_TAX_FORM, year, user }))
```
Betree:
```suggestion
```

server/lib/taxForms.js
```diff
@@ -0,0 +1,88 @@
+import models from '../models';
+import { uniqBy } from 'lodash';
+const { RequiredLegalDocument, LegalDocument, Collective, User } = models;
+const {
+  documentType: { US_TAX_FORM },
+} = RequiredLegalDocument;
+
+export async function findUsersThatNeedToBeSentTaxForm({ invoiceTotalThreshold, year }) {
```
Betree: An elegant way of checking that could be to have a `LEFT OUTER JOIN` between `RequiredLegalDocuments` and `LegalDocuments`, with a JOIN condition on year/user and `requestStatus != RECEIVED`

server/models/Collective.js
```diff
@@ -26,6 +26,9 @@ import roles from '../constants/roles';
 import activities from '../constants/activities';
 import { HOST_FEE_PERCENT } from '../constants/transactions';
 import { types } from '../constants/collectives';
+import status from '../constants/expense_status';
+
+const { PENDING, APPROVED, PAID } = status;
```
Betree: It can be confusing to have this in the global scope for `Collective` as a collective can be pending for approval or approved by the host and the code will not reflect that we're talking about expenses when using `PENDING` below.

```es6
import ExpenseStatus from '../constants/expense_status';

if (... === ExpenseStatus.PENDING) {
  ... 
}
```

Would make more sense to me

server/controllers/helloworks.js
```diff
@@ -0,0 +1,105 @@
...
+      saveFileToTempStorage({ filename: key, buffer });
+      return Promise.resolve({ Location: key });
+    }
+
+    // TODO: encrypt the form here.
+
+    return new Promise((resolve, reject) => {
+      s3.upload({ Body: buffer, bucket, key }, (err, data) => {
+        if (err) {
+          console.log('error uploading file to s3: ', err);
```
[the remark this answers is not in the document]
pietgeursen: Done.

server/controllers/helloworks.js
```diff
@@ -0,0 +1,105 @@
...
+    const key = createTaxFormFilename({ id, year, documentType });
+
+    if (!s3) {
+      // s3 may not be set in a dev env
+      console.error('s3 is not set, saving file to temp folder. This should only be done in development');
+      saveFileToTempStorage({ filename: key, buffer });
+      return Promise.resolve({ Location: key });
+    }
+
+    // TODO: encrypt the form here.
```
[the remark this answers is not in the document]
pietgeursen: Removed.

server/controllers/helloworks.js
```diff
@@ -0,0 +1,105 @@
...
+async function callback(req, res) {
+  console.log(req.body);
+  const {
+    body: { status, workflow_id: workflowId, data, id, metadata },
+  } = req;
+
+  if (status && status === 'completed' && workflowId == HELLO_WORKS_WORKFLOW_ID) {
+    const { email, year } = metadata;
+    const documentId = Object.keys(data)[0];
+    console.log(`workflowId: ${workflowId}, documentId: ${documentId}`);
```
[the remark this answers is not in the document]
pietgeursen: Removed.

server/lib/taxForms.js
pietgeursen: Removed

server/models/Collective.js
pietgeursen: Oh, good point, good to know. Fixed.

server/graphql/v1/types.js
```diff
@@ -576,6 +577,19 @@ export const ExpenseType = new GraphQLObjectType({
...
+        type: GraphQLBoolean,
+        async resolve(expense) {
+          const incurredYear = moment(expense.incurredAt).year();
+
+          return isUserTaxFormRequiredBeforePayment({
+            year: incurredYear,
+            invoiceTotalThreshold: 600e2,
+            HostCollectiveId: expense.CollectiveId,
+            UserId: expense.UserId,
+          }); // TODO define the threshold once
```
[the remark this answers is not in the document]
pietgeursen: Removed.

server/controllers/helloworks.js
```diff
@@ -0,0 +1,105 @@
...
+        } else {
+          resolve(data);
+        }
+      });
+    });
+  };
+}
+
+function saveFileToTempStorage({ buffer, filename }) {
+  fs.writeFile(`/tmp/${filename}`, buffer, console.log);
```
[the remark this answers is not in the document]
pietgeursen: Yep.

server/lib/taxForms.js
pietgeursen: Yeah, cool. Done.

server/lib/taxForms.js
pietgeursen: I've spend some time looking at this and I can't see how this would work. We need to get the user total expenses as well.

Also, this query is only used by the cron job so it's not in a hot code path, so can we just leave it as is?

Betree: Totally, I missed the fact that this was only used by the CRON and I'm really fine with it.

GitHub should let you precise on individual review comments if they are just thought or actual blockers :)

server/graphql/v1/types.js
```diff
@@ -576,6 +577,19 @@ export const ExpenseType = new GraphQLObjectType({
           });
         },
       },
+      userTaxFormRequiredBeforePayment: {
+        type: GraphQLBoolean,
+        async resolve(expense) {
+          const incurredYear = moment(expense.incurredAt).year();
+
+          return isUserTaxFormRequiredBeforePayment({
```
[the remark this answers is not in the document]
pietgeursen: > Not a blocker for me if this is just for the host dashboard or for "show expense" page (https://opencollective.com/______/expenses/7988)

Yes, this would only be used in those cases. It doesn't make sense to tell the public about tax form status on expenses.

My intuition is that the front end should only query for this information when it really needs it. 

Plus I just looked at the loader file for the first time and I'm not sure exactly what to do in there.

So, are you happy to leave this as is and we just make sure the frontend only queries this when needed?

package.json
pietgeursen: Actually, let's not block this PR on that one. I don't know when they're going to get to it. My change was tiny and only used by the cron job code anyway.

server/graphql/v1/types.js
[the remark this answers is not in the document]
Betree: The purpose of `loaders` is to batch the queries. Instead of querying the DB for each entry (in this case, each expense) the goal is to make one big query where you fetch all the data and to do the reconciliation in JS.

But yes, if this is only going to be fetched in the host dashboard then I'm ok with this.

package.json
znarf: Can you switch to the regular one?

znarf:
Looks great! Good to merge from my perspective.

config/default.json
```diff
@@ -108,5 +108,18 @@
   "recaptcha": {
     "siteKey": "6LcyeXoUAAAAAFtdHDZfsxncFUkD9NqydqbIFcCK",
     "secretKey": "6LcyeXoUAAAAACqNaR6EZJ_LG0V0N35uNkab3ibj"
+  },
+  "helloworks": {
+    "key": "RO7aWNLJx6CzhRix",
+    "secret": "FanG4FjeV3aGHjpS2A1YckyWuhvxfKHoZN5Q5bHO",
+    "workflowId": "qdUbX5nw8sMZzykz",
```
znarf: Are these ok to share with everyone for dev and staging?

package.json
pietgeursen: Yep, done.

config/default.json
pietgeursen: Yes, they're test keys.

PR #2076, opened 2019-06-05
Enable isPledged in GraphQL API v1

znarf:


znarf:
@Betree yes, this was not working.

` Cannot query field "isPledged" on type "CollectiveInterface". Did you mean to use an inline fragment on "Collective", "Event", "User", or "Organization"?`

Betree:
Ok I missed that. Thanks for fixing it!

PR #2084, opened 2019-06-06
feat: filter orders by sent / received

znarf:


Betree:
Not easy to name this param :thinking: 

A param taking an enum like `SENT|RECEIVED|BOTH` is appealing but hard to name. Here are few suggestions but none of them sounds perfect:
- origin
- direction

Otherwise the two booleans are not chocking to me, we should just have a `description` to make their purpose clear.

znarf:
I already don't like much the name `order` (as in our model table) because in our API, we can order results and _ordering orders_ is confusing.

It's a bit the same for direction, because _order direction_ is confusing. You may grab orders filtered by direction sorted by x in ascending direction.

To avoid _order_ confusion, we may rename `orderBy` by `sortBy`. Not easy.

znarf:
@Betree I found a great name: `filter` :-P

PR #2091, opened 2019-06-07
don't send a notification when expense approved to host if collective hasn't been approved

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/1967

### DIFF
--- a/server/models/Expense.js
+++ b/server/models/Expense.js
@@ -194,7 +194,7 @@ export default function(Sequelize, DataTypes) {
       CollectiveId: this.collective.id,
       data: {
         host: get(host, 'minimal'),
-        collective: this.collective.minimal,
+        collective: { ...this.collective.minimal, isActive: this.collective.isActive },
         user: user.minimal,
         fromCollective: userCollective.minimal,
         expense: this.info,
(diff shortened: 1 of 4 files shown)


server/models/Expense.js
```diff
@@ -194,7 +194,7 @@ export default function(Sequelize, DataTypes) {
       CollectiveId: this.collective.id,
       data: {
         host: get(host, 'minimal'),
-        collective: this.collective.minimal,
+        collective: { ...this.collective.minimal, isActive: this.collective.isActive },
```
znarf: can we add `isActive` to minimal instead?

xdamman: Do we really want to "pollute" the `minimal` used elsewhere with this information that is only needed here?

znarf: I don't know the history behind it. Looks weird to extend it with just one property.

Now that I'm looking at it, I'm wondering why the following are `minimal`:

```
twitterHandle,
githubHandle,
publicUr,
mission,
isSupercollective,
```

xdamman: This entire thing is bogus. It was initially meant for storing a minimum set of data for the activity table. But we should revisit that at some point.

PR #2093, opened 2019-06-07
fix(expenses): only return expenses from collectives that have been approved by the host

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/2107

### DIFF
--- a/server/graphql/v1/queries.js
+++ b/server/graphql/v1/queries.js
@@ -666,14 +666,15 @@ const queries = {
           throw new Error('Collective not found');
         }
         const getCollectiveIds = () => {
-          // if is host, we get all the expenses across all the hosted collectives
+          // if is host, we get all the expenses across all the hosted collectives that are active
           if (args.includeHostedCollectives) {
-            return models.Member.findAll({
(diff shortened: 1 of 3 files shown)


server/graphql/v1/queries.js
```diff
@@ -666,14 +666,15 @@ const queries = {
...
+            return models.Collective.findAll({
+              attributes: ['id'],
               where: {
-                MemberCollectiveId: collective.id,
-                role: 'HOST',
+                HostCollectiveId: collective.id,
+                isActive: true,
               },
-            }).map(member => member.CollectiveId);
+            }).map(c => c.id);
```
znarf:
```suggestion
            }).map(collective => collective.id);
```

It's better with an explicit variable name.

znarf:
@xdamman you didn't answer to my feedback?

server/graphql/v1/queries.js
xdamman: I actually had written "collective" in full at first and then renamed it to avoid confusion with a variable in the scope above. And since it's a very short lived variable, I thought it was ok to use just `c`.

PR #2118, opened 2019-06-13
Remove findOrCreate for collective pledge in createOrder

flickz:
When a new collective is created, a hook is expected to run to set the default [opencollective service paymentMethod](https://github.com/opencollective/opencollective-api/blob/master/server/models/Collective.js#L566-L583), the hook triggers a foreign key constraint error when a collective is created using `findOrCreate` method but doesn't using standalone `create` method. Basically, creation with `create` method behaves differently to `findOrCreate`. 

While the main cause of this hasn't been figured out but this PR fix the [issue](https://github.com/opencollective/opencollective/issues/2098)

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -189,10 +189,10 @@ export async function createOrder(order, loaders, remoteUser, reqIp) {
         defaults: order.collective,
       }))[0];
     } else if (order.collective.githubHandle) {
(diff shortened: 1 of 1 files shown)


znarf:
code source for findOrCreate: https://github.com/sequelize/sequelize/blob/6c76c240467656c585f6edaf8c5292f4d9749b8b/lib/model.js#L2274

Betree:
Good fix, may be a bug with sequelize itself

PR #2121, opened 2019-06-14
Add character validation to collective slug

flickz:
This PR adds character validation to `Collective` model in `slug` field, the validation ensures no characters other than alphanumeric and hyphens are contained in collective slug.

Screenshot from 2019-06-14 15-47-17

Ref [#2077](https://github.com/opencollective/opencollective/issues/2077)

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -149,6 +149,11 @@ export default function(Sequelize, DataTypes) {
             args: [collectiveSlugBlacklist],
             msg: 'The slug given for this collective is a reserved keyword',
           },
+          isValid(value) {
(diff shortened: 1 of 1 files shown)


znarf:
We need to check that the error is properly managed/reported when creating an individual account or a collective.

znarf:
@flickz have you checked that the error is properly reported in those situations?

flickz:
Sorry for the delay, during creation of collective or individual account, collective or individual slugs are generated automatically, the generated slug can only have hyphens. Any attempt to edit it with invalid character, the validation errors pops up.

PR #2127, opened 2019-06-17
Create contributors concept

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2102

---

- [x] Base `Contributor` definition
- [x] Collective resolvers
  - [x] Add an index on Transactions > `UsingVirtualCardFromCollectiveId`
- [x] Tier resolvers

About new index on `UsingVirtualCardFromCollectiveId`:
This divides the complexity of the query for collective members by 2 (using pgAdmin for estimation). As we use it in many places (eg. to calculate collective's balances) we can expect performance improvement on many places.

**Bonus**
- [ ] Benchmark fetching webpack members VS contributors


Betree:
@znarf Ready for review

PR #2128, opened 2019-06-18
refactor(dataloader-sequelizer): Upgrade dataloader-sequelizer package

brymut:
Upgrade dataloader-sequelizer package to v2.0.0, breaking change requiring
switch to use of Context API.
CC: opencollective/opencollective#2094

### DIFF
--- a/server/graphql/loaders.js
+++ b/server/graphql/loaders.js
@@ -2,14 +2,10 @@ import models, { sequelize, Op } from '../models';
 import { getListOfAccessibleMembers } from '../lib/auth';
 import { TransactionTypes } from '../constants/transactions';
 import DataLoader from 'dataloader';
-import dataloaderSequelize from 'dataloader-sequelize';
+import { createContext } from 'dataloader-sequelize';
+
 import { get, groupBy } from 'lodash';
 import debugLib from 'debug';
-
-dataloaderSequelize(models.Order);
-dataloaderSequelize(models.Transaction);
-dataloaderSequelize(models.Collective);
-dataloaderSequelize(models.Expense);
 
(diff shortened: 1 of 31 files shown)


brymut:
Hi @znarf, I'm sorry this took longer than planned. Got stuck on some family commitments and my day job wasn't doing me any favours. I should have communicated on here better.

But here it is, rebased and working off of an origin branch. Could you please have a look whenever you can.

server/graphql/loaders.js
```diff
@@ -2,15 +2,11 @@ import models, { sequelize, Op } from '../models';
...
-import dataloaderSequelize from 'dataloader-sequelize';
+import { createContext } from 'dataloader-sequelize';
+
 import { get, groupBy } from 'lodash';
 import debugLib from 'debug';
 
-dataloaderSequelize(models.Order);
-dataloaderSequelize(models.Transaction);
-dataloaderSequelize(models.Collective);
-dataloaderSequelize(models.Expense);
```
znarf: My understanding is that this was enabling automatic dataloading within Sequelize. For example, all calls to `models.Collective.findByPk` would be cached for the request thanks to that.

If that's right, it means that all call to `models.Collective.findByPk` in the GraphQL API needs to be replaced by `req.loaders.Collective.byId.load`.

brymut: Hi, thanks for pointing this out, had to revisit how dataloaders work. My previous understanding was that they were used further down during the request when data had already been called & cached from sequelize, not automatically. I've made some changes for the Collective loaders.

znarf:
@brymut This was a great start! Hope you will find the time to complete it.

PR #2129, opened 2019-06-18
perf: update email unsubscription url

EslamHiko:
Update email unsubscription URL to point to the frontend. created in this PR : https://github.com/opencollective/opencollective-frontend/pull/1959


znarf:
We need to wait on https://github.com/opencollective/opencollective-frontend/pull/1959 to be merged and deployed.

PR #2138, opened 2019-06-18
Update status of orders with totalAmount=0

znarf:
Fix: https://github.com/opencollective/opencollective/issues/1963


johanricher:
Should this update be already propagated to opencollective.com? I don't seem to see any change.

znarf:
@johanricher not yet, this needs to be deployed.

Note for the future, we should not close issues until they are deployed.

johanricher:
That was my intuition, thanks for the confirmation! I'll wait. :)

PR #2141, opened 2019-06-18
Custom fields for Tiers

flickz:
This PR follows up [#2128](https://github.com/opencollective/opencollective/issues/2128) API specification.

### DIFF
--- a/server/graphql/v1/inputTypes.js
+++ b/server/graphql/v1/inputTypes.js
@@ -8,6 +8,7 @@ import {
   GraphQLString,
   GraphQLScalarType,
   GraphQLError,
+  GraphQLEnumType,
 } from 'graphql';
 
 import GraphQLJSON from 'graphql-type-json';
@@ -67,13 +68,25 @@ export const PaymentMethodDataVirtualCardInputType = new GraphQLInputObjectType(
   }),
 });
 
+const CustomFieldType = new GraphQLEnumType({
+  name: 'CustomFieldType',
+  description: 'Type of custom field',
+  values: {
+    number: {},
+    text: {},
+    email: {},
+    date: {},
+    radio: {},
+    url: {},
+  },
+});
+
 export const CustomFieldsInputType = new GraphQLInputObjectType({
   name: 'CustomFieldsInputType',
   description: 'Input for custom fields for order',
   fields: () => ({
-    type: { type: GraphQLString },
+    type: { type: CustomFieldType },
     name: { type: GraphQLString },
-    value: { type: GraphQLString },
     label: { type: GraphQLString },
     required: { type: GraphQLBoolean },
   }),
@@ -286,7 +299,7 @@ export const TierInputType = new GraphQLInputObjectType({
       description: 'amount that you are trying to raise with this tier',
     },
     password: { type: GraphQLString },
-    settings: { type: GraphQLJSON },
+    customFields: { type: new GraphQLList(CustomFieldsInputType) },
     startsAt: {
       type: GraphQLString,
       description: 'Start of the campaign',
@@ -328,7 +341,7 @@ export const OrderInputType = new GraphQLInputObjectType({
     fromCollective: { type: CollectiveAttributesInputType },
     collective: { type: new GraphQLNonNull(CollectiveAttributesInputType) },
     tier: { type: TierInputType },
-    customFields: { type: new GraphQLList(CustomFieldsInputType) },
+    customData: { type: GraphQLJSON },
     recaptchaToken: { type: GraphQLString },
     // For taxes
     taxAmount: {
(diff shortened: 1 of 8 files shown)


znarf:
No change to `OrderInputType`?

flickz:
Oh I forgot, have added it.

server/graphql/v1/inputTypes.js
```diff
@@ -315,6 +316,7 @@ export const OrderInputType = new GraphQLInputObjectType({
     fromCollective: { type: CollectiveAttributesInputType },
     collective: { type: new GraphQLNonNull(CollectiveAttributesInputType) },
     tier: { type: TierInputType },
+    customFields: { type: GraphQLJSON },
```
znarf: Can we create a `type`for it?

flickz: I'm on it.

server/graphql/v1/inputTypes.js
```diff
@@ -67,6 +67,18 @@ export const PaymentMethodDataVirtualCardInputType = new GraphQLInputObjectType(
   }),
 });
 
+export const CustomFieldsInputType = new GraphQLInputObjectType({
+  name: 'CustomFieldsInputType',
+  description: 'Input for custom fields for order',
+  fields: () => ({
+    type: { type: GraphQLString },
```
znarf: This should be an enum, so we know the allowed values.

server/graphql/v1/inputTypes.js
```diff
@@ -274,6 +286,7 @@ export const TierInputType = new GraphQLInputObjectType({
       description: 'amount that you are trying to raise with this tier',
     },
     password: { type: GraphQLString },
+    settings: { type: GraphQLJSON },
```
znarf: If we support it in the Input, we should ask for `customFields`, not settings. We don't want to have to take care of implementing a merging strategy when updating `settings`.

server/graphql/v1/inputTypes.js
```diff
@@ -315,6 +328,7 @@ export const OrderInputType = new GraphQLInputObjectType({
     fromCollective: { type: CollectiveAttributesInputType },
     collective: { type: new GraphQLNonNull(CollectiveAttributesInputType) },
     tier: { type: TierInputType },
+    customFields: { type: new GraphQLList(CustomFieldsInputType) },
```
znarf: This is simpler than that. This can just be an object with `key: value` pairs.

And we can name it `customValues` or `customData` to not mistake it with the `customFields` from the `Tier`.

server/graphql/v1/types.js
```diff
@@ -1502,9 +1508,9 @@ export const OrderType = new GraphQLObjectType({
       },
       data: {
         type: GraphQLJSON,
-        description: 'Additional information on order: tax',
+        description: 'Additional information on order: tax and custom fields',
         resolve(order) {
-          return pick(order.data, ['tax']) || null;
```
znarf: Why remove that, wasn't it a good practice of picking only what we want to return?

server/graphql/v1/inputTypes.js
znarf: And this is where we can support `CustomFieldsInputType`.

znarf:
Please add some tests involving a Tier with `customFields` and an Order with `customValues`.

server/graphql/v1/inputTypes.js
flickz: Enum of?

znarf: All possible values for `type`.

server/graphql/v1/inputTypes.js
flickz: Not sure I understand this completely, following the API spec:

> Add a new settings (JSON) column to Tiers that will contain additional data & Expose settings through GraphQL

What is changing now?

server/graphql/v1/inputTypes.js
flickz: > This is simpler than that. This can just be an object with key: value pairs.

Is it not suppose to be an array of custom fields even though we have only one field currently, just against future expansion?

I.e 
```js
[
 {
  "name": "jsonUrl",
  "label": "URL of the JSON dependency file",
  "type": "url",
  "required": true,
  }
]
````

File: server/graphql/v1/types.js (hunk 5 above)
Comment (flickz) [reply server/graphql/v1/types.js]: Oh that was my bad. Will fix it.

File: server/graphql/v1/inputTypes.js (hunk 2 above)
Comment (flickz) [reply server/graphql/v1/inputTypes.js]: okay.

File: server/graphql/v1/inputTypes.js (hunk 4 above)
Comment (znarf) [reply server/graphql/v1/inputTypes.js]: In this case, in the Order table, we don't care about `label`, `type` and `required`, we just want keys and values.

```
{"jsonUrl": "http://backyourstack.com/heroku/dependencies.json"}
```

server/graphql/v1/inputTypes.js
znarf: You can discard what I wrote, the way it is implemented it will add complexity for not much gain.

Just let's make sure we don't lose the settings when we update the `Tier` from the interface. (it could be easy to introduce a bug that is resetting the settings in that case, because we don't surface settings in in the inputs).

flickz:
Added the test now.

PR #2144, opened 2019-06-19
Incognito feature

xdamman:
Required for https://github.com/opencollective/opencollective-frontend/pull/1984


server/graphql/v1/mutations/orders.js
```diff
@@ -286,7 +286,10 @@ export async function createOrder(order, loaders, remoteUser, reqIp) {
         possibleRoles.push(roles.MEMBER);
       }
 
-      if (!remoteUser.hasRole(possibleRoles, order.fromCollective.id)) {
+      if (
+        !remoteUser.hasRole(possibleRoles, order.fromCollective.id) &&
+        fromCollective.CreatedByUserId !== remoteUser.id
```
Betree: If I create a collective then later get removed from core contributors, I'll still be able to make orders on behalf of the collective I've created. It's a permission issue.

server/graphql/v1/mutations/collectives.js
```diff
@@ -111,15 +111,23 @@ export async function createCollective(_, args, req) {
...
-    collective.editPaymentMethods(args.collective.paymentMethods, {
-      CreatedByUserId: req.remoteUser.id,
-    }),
   ];
 
+  if (collectiveData.tiers) {
+    promises.push(collective.editTiers(collectiveData.tiers));
+  }
+
+  if (collectiveData.paymentMethods) {
```
Betree: If I want to delete all the payment methods for the collective, this will return false and it won't work.

server/graphql/v1/mutations/collectives.js
```diff
@@ -111,15 +111,23 @@ export async function createCollective(_, args, req) {
...
-    collective.editTiers(collectiveData.tiers),
     collective.addUserWithRole(req.remoteUser, roles.ADMIN, {
       CreatedByUserId: req.remoteUser.id,
     }),
-    collective.editPaymentMethods(args.collective.paymentMethods, {
-      CreatedByUserId: req.remoteUser.id,
-    }),
   ];
 
+  if (collectiveData.tiers) {
```
Betree: If I want to delete all tiers for the collective, this will return false and it won't work.

server/graphql/v1/mutations/orders.js
znarf: While I guess you're not supposed to be removed as an admin of this collective, why can't we just rely on roles here? It's cleaner and easier to maintain. As seen in: https://github.com/opencollective/opencollective-api/pull/2002/files

server/graphql/v1/mutations/collectives.js
```diff
@@ -111,15 +111,23 @@ export async function createCollective(_, args, req) {
...
+    promises.push(collective.editTiers(collectiveData.tiers));
+  }
+
+  if (collectiveData.paymentMethods) {
+    promises.push(
+      collective.editPaymentMethods(args.collective.paymentMethods, {
```
znarf:
```suggestion
      collective.editPaymentMethods(collectiveData.paymentMethods, {
```

clearer

xdamman:
While would you want to do that? What’s the difference between anonymous, pseudonymous, a user and an organization? They are basically all the same from the platform stand point. In all cases you never expose the personal identifiable information (email address), and you let the user decide what information to show publicly. 
This feature is really only about enabling you to have multiple identities / profiles so that you have full control over them.

server/graphql/v1/mutations/orders.js
xdamman: @Betree that’s a great point. While it will never happen in the case of a collective of type USER, this is a very valid argument for organizations and collectives. No reason while you should always be considered as admin. 

I’ll see if can add an ADMIN role in the Members table. I agree with @znarf that it would be the cleanest. If not I’ll restrict this to collectives of type USER.

Betree: > The Internet generation needs organizations that reflect who we are; where anybody can contribute to a shared mission; **where leaders can easily change**

:slightly_smiling_face:

znarf:
@xdamman When creating anonymous accounts, the intent is clearly special and we should be able to recognize that in the database. Like you're recognizing an Organization account with `type=ORGANIZATION`. This will help in many places, like:
- computing statistics
- displaying the proper 'anonymous' avatar
- display them in their own section in the user menu

xdamman:
@znarf what about simply `collective.name === null`?

Betree:
@xdamman `collective.name` has a `not null` constraint, I wouldn't sacrifice this additional security to introduce a new feature. That also creates new questions, like what happens if I set an empty name in my profile page, does it become an anonymous profile (then we need to be explicit to the users)? If I set a name on my anonymous profile, what happens?

`type === ANONYMOUS` seems simpler and more explicit to me, we don't have to document it: anyone seeing that will understand that this is an anonymous account. `name === null` would be a secret knowledge that we'll need to explain to new contributors.

znarf:
I anticipated that it could be a bigger change to introduce a new `type=ANONYMOUS`, this is why I also proposed `type="USER" isAnonymous=TRUE` which I think is reasonable.

server/graphql/v1/TransactionInterface.js
```diff
@@ -179,10 +179,16 @@ const TransactionFields = () => {
...
+      async resolve(transaction, args, req) {
+        // We don't return the user if the transaction has been created by someone who wanted to remain anonymous
+        // This is very suboptimal. We should probably record the CreatedByCollectiveId (or better CreatedByProfileId) instead of the User.
         if (transaction && transaction.getCreatedByUser) {
+          const fromCollective = await transaction.getFromCollective();
+          if (fromCollective.isAnonymous && (!req.remoteUser || !req.remoteUser.isAdmin(transaction.CollectiveId)))
+            return {};
+          const collective = await transaction.getCollective();
+          if (collective.isAnonymous && (!req.remoteUser || !req.remoteUser.isAdmin(transaction.FromCollectiveId)))
+            return {};
```
znarf: Have you considered dropping `createdByUser` altogether? I'm pretty sure we don't need it / can avoid it.

server/graphql/v1/types.js
```diff
@@ -313,7 +316,10 @@ export const MemberType = new GraphQLObjectType({
       member: {
         type: CollectiveInterfaceType,
         resolve(member, args, req) {
-          return member.memberCollective || req.loaders.collective.findById.load(member.MemberCollectiveId);
+          const memberCollective =
+            member.memberCollective || req.loaders.collective.findById.load(member.MemberCollectiveId);
+          memberCollective.inTheContextOfCollectiveId = member.CollectiveId;
```
znarf: This doesn't "smell" good.

server/graphql/v1/TransactionInterface.js
xdamman: Maybe, it just feels like out of scope for this PR. I wanted to keep everything else "as is".

znarf: If this can avoid that kind of suboptimal code, this is worth it.

server/graphql/v1/types.js
xdamman: Not exactly sure what to do with this feedback ¯\\\_(ツ)_/¯ 

The goal here is to keep track of the context of the collective in which this user is.
As an admin of the collective, I should only be able to access the email address of an anonymous backer in the context of that collective. Otherwise the permission system would give me access if the same anonymous user made another donation to another collective that I am an admin of.

znarf: The pattern is not good. Modifying an object on the fly like that will be source of bugs.

For example, what if what you get out of the resolver is not a copy but a shared object and you end up sending the modified version in unexpected places?

The removal of `createdByUser` in transactions could be a simpler solution.

xdamman: It's used in the REST api. 
https://github.com/OpenCollective/opencollective-api/blob/3c1cbfd4a72b3d19987cae33559a135fc5fa3cef/server/routes.js#L219
https://github.com/OpenCollective/opencollective-api/blob/3c1cbfd4a72b3d19987cae33559a135fc5fa3cef/server/graphql/v1/restapi.js#L23-L26

I don't know if anyone is using it. So what do we do?

> For example, what if what you get out of the resolver is not a copy but a shared object and you end up sending the modified version in unexpected places?

I don't understand. This is not going to interfere with anything else. And if it's missing it falls back to the safe scenario which is not sending the personal data.

znarf: If it's only used in the REST API, it's already a very limited usage (I only have 2 users in mind), and very likely no one is relying on the specific field. Not a blocker for me.

xdamman: Isn't webpack relying on getting the email addresses of their backers for their voting platform https://webpack.js.org/vote/ ?

znarf: No trace in the logs, I only see Triplebyte using it.

xdamman: Well, one too much. I could see why they would want to get the email addresses of the people using the gift card.

Then maybe we could simplify by not returning the `Transaction.createdByUser` when it's anonymous even if you are an admin?

But it's weird that only in that case we don't return it. It feels like a bug.

znarf: Can you ask them if they rely on that feature?

znarf: I'm sure they don't rely on it because this functionality was broken until February and they never complained. https://github.com/opencollective/opencollective-api/commit/8b36cc52ba15b5b081b6263c355ebe90048401dc

xdamman:
> I think we should not try to protect that because it's not possible to do it properly. There is too many places this would need to be enforced in the API.

It doesn't hurt to not show it through the GraphQL API which is the primary way we access data.

I'll look for other places. If you think of some, let me know.

server/graphql/v1/types.js
Betree: As such the `inTheContextOfCollectiveId` really looks like a hack. It's not documented, not enforced in the model and you have to go through the code to understand what it does. That will confuse any external contributors, or us if we look at it in 2 months.

@znarf's solution looks good to me.

But if we really want to go with the current pattern, we should at least make it explicit so we don't create new technical debt. We could do this by adding the field to the model (`DataTypes.VIRTUAL`) with a proper description.

server/models/User.js
```diff
@@ -477,60 +484,52 @@ export default (Sequelize, DataTypes) => {
...
+    };
+    user.collective = await models.Collective.create(userCollectiveData, sequelizeParams);
+
+    // It's difficult to predict when the image will be updated by findImageForUser
+    // So we skip that in test environment to make it more predictable
+    if (config.env !== 'test' && config.env !== 'circleci') {
+      user.collective.findImageForUser(user);
+    }
+    user.CollectiveId = user.collective.id;
+    await user.save(sequelizeParams);
```
Betree: I'd rather return the result of `user.save(sequelizeParams)` directly, as formatters apply to the data when saving the model before and after `save` can be different.

znarf:
> It doesn't hurt to not show it through the GraphQL API which is the primary way we access data. I'll look for other places. If you think of some, let me know.

I'm insisting that we should not try to be smart and try limiting what we can query on Incognito accounts for now. This is setting the expectations too high, and this is flawed by design because this doesn't match our model.

Think of it as Bitcoin address, you can see all transactions belonging to one and it's fine. If we want full Incognito, to respect our model, we should create an new account for each transaction. This was actually considered and might be done as a follow up.

server/graphql/v1/CollectiveInterface.js
```diff
@@ -744,7 +745,12 @@ const CollectiveFields = () => {
     },
     createdByUser: {
       type: UserType,
-      resolve(collective) {
+      resolve(collective, args, req) {
+        if (
+          collective.isAnonymous &&
+          (!req.remoteUser || !req.remoteUser.isAdmin(collective.inTheContextOfCollectiveId))
+        )
+          return {};
```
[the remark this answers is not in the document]
xdamman: Otherwise it breaks down the graph when you try to return `user.firstName` for example.
Otherwise you need to always test that `user` passed to the resolver is not null.

server/models/User.js
xdamman: it doesn't matter since we are waiting for the result of `user.save`.
But sure. Doesn't hurt.

server/graphql/v1/types.js
xdamman: Would it make more sense to have that variable in the context of the resolver (the 3rd argument `req`)?

We have to return the `createdByUser` when the host is logged in.

server/models/User.js
Betree: Does `user.save` mutates the object with the return?

server/graphql/v1/types.js
Betree: > Would it make more sense to have that variable in the context of the resolver (the 3rd argument req)?

Yes

xdamman:
It's not about trying to be smart or about setting false expectations. We don't show the slug of an incognito profile on the frontend because we don't want to link to it, so there is really no reason to enable that on the API side.

I have very limited time to work on this. Originally I was just supposed to do a proof of concept.
Now it's really becoming a drag. It costs me a lot of context switching and I never see the end of it. It's very frustrating.

It's been a month already that this pull request has been opened. We need to either kill it if you feel strongly that it has to be implemented differently or accept that it's good enough as a first iteration. You can always iterate on it if you feel like some parts need to change / can be improved. But I'm not a full time developer so asking little changes here and there are very costly for me and take forever.

@alanna and other users are also waiting for this to be delivered: https://github.com/opencollective/opencollective/issues/1841#issuecomment-506135817

Let's do a post mortem after this to avoid such frustrations in the future (I'm sure you have yours as well).

server/graphql/v1/types.js
znarf: > We have to return the createdByUser when the host is logged in.

I'm not following. What is the relation with the current discussion?

Also, my feedback to move forward simple and efficient, deprecate `createdByUser` in transactions and send a fake object in it **all the time**.

znarf:
I'm sorry for your frustration and I'm sure we'll talk about that in a retrospective.

My point on "setting false expectations" is actually to make the feature simpler and the delivery simpler, not the opposite. By trying to prevent the Incognito transactions from being browsed, you need more checks and more code, and it's super easy to find ways to bypass it. So I just suggest to not do it, that's all.

server/models/User.js
xdamman: I'd expect so. But I'm not sure. Doesn't matter here.

server/graphql/v1/types.js
xdamman: This is in the context of getting the `createdByUser` based on the `memberCollectiveId` which references the incognito profile.

But from GraphQL's perspective, it doesn't matter much what the path is.

I tried but we can't keep track of the context in which a memberCollective is in the context of the resolver because this could be overwritten if we browse members from multiple collectives.

So look, I know it's a bit hacky but it works and it has tests.
I can't come up with anything better for now with the constrain that we have to return the personal info of the incognito profile if logged in as an admin of the host.

I'll add a `DataTypes.VIRTUAL` as suggested by @Betree .

I'm trying my best to help you guys so that you don't have to build this particular feature but I'm not full time on the code and I'm not as talented as you are. So I can help, but getting to your level of code quality would require a different level of commitment and focus on the code that I just can't offer right now. Thank you for your understanding.

server/models/User.js
Betree: This was a genuine question, I don't know if `save` mutates the object and sequelize's documentation isn't helping much.

server/models/Collective.js
```diff
@@ -746,7 +765,7 @@ export default function(Sequelize, DataTypes) {
     }
 
     if (this.type === 'USER') {
-      if (user && user.email && this.name && this.name !== 'anonymous') {
+      if (user && user.email && this.name && this.name !== 'incognito') {
```
znarf: We wan't to keep that for previous "anonymous" accounts.

```suggestion
      if (user && user.email && this.name && this.name !== 'anonymous' && this.name !== 'incognito') {
```

server/lib/payments.js
```diff
@@ -451,7 +451,7 @@ const sendOrderProcessingEmail = async order => {
     user: user.info,
     collective: collective.info,
     host: host.info,
-    fromCollective: fromCollective.minimal,
```
znarf: I don't see the point of this change.

xdamman: need to pass the `isIncognito` to the email template.

PR #2146, opened 2019-06-20
Encrypt tax forms before uploading to s3

pietgeursen:
Encrypts tax forms on the server before uploading to s3.


package.json
```diff
@@ -100,6 +100,8 @@
     "sequelize-historical": "5.1.0",
     "showdown": "1.9.0",
     "stripe": "7.4.0",
+    "tweetnacl": "^1.0.1",
```
Betree: We use fixed versions (`1.0.1`)

package.json
```diff
@@ -100,6 +100,8 @@
...
+    "tweetnacl-util": "^0.15.0",
```
Betree: We use fixed versions (`0.15.0`)

test/lib.encryption.test.js
```diff
@@ -0,0 +1,16 @@
...
+  it('it encrypts and decrypts ok', () => {
+    const message = 'OpenCollective Rules';
+    const buff = Buffer.from(message);
+    const key = generateKey();
+
+    const encrypted = encrypt(buff, key);
+
+    const result = decrypt(encrypted, key);
+
+    expect(result).to.eq(message);
```
Betree: I would just add this to ensure the message is indeed encrypted (otherwise early returns in encrypt+decrypt could bypass the test):

```es6
expect(encrypted).to.not.eq(message);
```

package.json
pietgeursen: done.

test/lib.encryption.test.js
pietgeursen: done.

package.json
pietgeursen: done

PR #2152, opened 2019-06-22
fix(homepage): better selected of sponsors to feature

xdamman:
Improving the query to select the sponsors to feature on the homepage.

I'm only taking into account sponsors that have donated to more than one collective in the past 3 months.

Total amount donated is not limited to last 3 months.

Related to https://github.com/opencollective/opencollective/issues/1529


Betree:
Sounds good! Also this doesn't take into account the money donated through gift cards, it would be great to add that too.

PR #2153, opened 2019-06-23
WIP: Daily cron to dispatch donation to dependencies collectives

flickz:
Following up [#2129](https://github.com/opencollective/opencollective/issues/2129).
PS: This is experimental as most of the data used are dummy.

My understanding of this feature is to create a custom Tier as type `PREPAID` on "host collective", like every other tiers, using it follows the same contribution process except later on a `customField` will be shown on the contribution `details`  step to get the json url of dependencies that would benefits. The order goes through the process (`createOrder` mutation resolver process), payment is executed and the "host collective" gets the money initially. 
Now comes the daily cron which this PR is trying to address, sequentially following:

- The `orders` are fetched with the following conditions, `order.status === ACTIVE`,  `order.Tier.type === PREPAID` && `order.Subscription.nextChargeDate === Date.now()`.  (The subscription has been added yet to the PR).

- Using `bluebird` `map` method, the orders are iterated through and each `order` goes through the following operation: 
                    - First thing is to get the dependencies that would be benefiting as expected to be provided in `order.data.customFields.jsonUrl`, using the url in a fetch function, the dependencies data is expected to be returned in below structure.
```js
const sampleDependecies = [
      { weight: 50, opencollective: { id: 43, name: 'Apex', slug: 'apex' } },
      { weight: 100, opencollective: { id: 10887, name: 'Nodejs Tech', slug: 'nodejs-tech' } },
    ];
```
- Next is iterating through the dependencies using the `map` method (using this method allows us to handle concurrency properly), and for each dependency, we compute their share of the original donation using their `weight` property. To compute their share, the sum of all dependencies weight is used to  express each dependency `weight` as percentage of their shares from the `amount`. An order (call it sub order :slightly_smiling_face: ) with `paymentMethod` is created in order is to dispatch donation to the current dependency collective. All information needed to create the order will be fetched, like dependency collective information, the order data looks like below:
```js
const orderData = {
        createdByUserId: order.createdByUserId,
        FromCollectiveId: order.FromCollectiveId,
        CollectiveId: collective.id,
        quantity: order.quantity,
        processedAt: new Date(),
        totalAmount,
        currency: order.currency,
        status: status.PENDING,
        paymentMethod: {
          primary: false,
          initalBalance: shareableAmount - totalAmount,
          type: 'prepaid',
          CollectiveId: order.FromCollectiveId,
          service: 'opencollective',
          customerId: order.FromCollectiveId.slug,
          currency: order.currency,
          data: { HostCollectiveId: fromHostId },
        },
      };
```
The next part will be setting `paymentMethod`, currently the [`libPayments.executeOrder` ](https://github.com/opencollective/opencollective-api/blob/master/server/lib/payments.js#L310) method needed to process the order, requires `models.User` instance in it's argument which isn't available in this context and it is where I'm stuck currently in this PR. I also don't know if the function can process order with this type of `paymentMethod` unless it is further expanded. At this stage, if the payment goes through for all dependencies the original `order` subscription `nextChargeDate` should be updated.

The explanation above only gives overview of what is happening in this PR, the cron would further need some constraints to be added in order to handle cases of where a charge doesn't go through or dependencies collective doesn't exist or handling concurrency, error management. 

This is the state of the code & my understanding of what I'm working, let me know what I'm not getting or missing.

### DIFF
--- a/cron/daily/dispatch_prepaid_subscription.js
+++ b/cron/daily/dispatch_prepaid_subscription.js
@@ -2,14 +2,14 @@
 import fetch from 'isomorphic-fetch';
 import uuidV4 from 'uuid/v4';
 import debugLib from 'debug';
-import { map } from 'bluebird';
+import moment from 'moment';
+import { map, filter } from 'bluebird';
 import { Op } from 'sequelize';
 
 import '../../server/env';
 
 import models from '../../server/models';
 import * as libPayments from '../../server/lib/payments';
-import { getNextChargeAndPeriodStartDates } from '../../server/lib/subscriptions';
 import status from '../../server/constants/order_status';
 
 const debug = debugLib('dispatch_prepaid_subscription');
@@ -36,26 +36,36 @@ async function run() {
           isActive: true,
           deletedAt: null,
           deactivatedAt: null,
-          activatedAt: { [Op.lte]: new Date() },
-          nextChargeDate: { [Op.lte]: new Date() },
         },
       },
     ],
   });
 
-  return map(allOrders, async order => {
+  return filter(allOrders, order => {
+    return order.Subscription.data && needsDispatching(order.Subscription.data.nextDispatchDate);
+  }).map(async order => {
     // Amount shareable amongst dependencies
     const shareableAmount = order.totalAmount;
     const jsonUrl = order.data.customData.jsonUrl;
     const depRecommendation = await fetchDependcies(jsonUrl);
     const sumOfWeights = depRecommendation.reduce((sum, dependency) => dependency.weigh + sum, 0);
+    const HostCollectiveId = await order.collective.getHostCollectiveId();
+
+    const paymentMethod = await models.PaymentMethod.create({
+      initialBalance: shareableAmount,
+      currency: order.currency,
+      CollectiveId: order.FromCollectiveId,
+      customerId: order.fromCollective.slug,
+      service: 'opencollective',
+      type: 'prepaid',
+      uuid: uuidV4(),
+      data: { HostCollectiveId },
+    });
 
     return map(depRecommendation, async dependency => {
       // Check if the collective is avaliable
       const collective = await models.Collective.findByPk(dependency.opencollective.id);
       const totalAmount = computeAmount(shareableAmount, sumOfWeights, dependency.weigh);
-      const HostCollectiveId = await order.collective.getHostCollectiveId();
-      // const pm = await order.collective.getPaymentMethod({ service: 'opencollective', type: 'prepaid' }, false);
 
       const orderData = {
         CreatedByUserId: order.CreatedByUserId,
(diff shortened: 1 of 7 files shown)


cron/daily/dispatch_prepaid_subscription.js
```diff
@@ -0,0 +1,142 @@
...
+  });
+
+  return filter(allOrders, order => {
+    return order.Subscription.data && needsDispatching(order.Subscription.data.nextDispatchDate);
+  }).map(async order => {
+    // Amount shareable amongst dependencies
+    const shareableAmount = order.totalAmount;
+    const jsonUrl = order.data.customData.jsonUrl;
+    const depRecommendation = await fetchDependcies(jsonUrl);
+    const sumOfWeights = depRecommendation.reduce((sum, dependency) => dependency.weigh + sum, 0);
```
znarf: let's fix the `weigh` typo everywhere please

cron/daily/dispatch_prepaid_subscription.js
```diff
@@ -0,0 +1,142 @@
...
+  });
+}
+
+const computeAmount = (totalAmount, sumOfWeights, dependencyWeight) => {
+  // Express each weight as percentage
+  const percentage = (dependencyWeight / sumOfWeights) * 100;
+  return Math.round((percentage / 100) * totalAmount);
+};
+
+const fetchDependcies = jsonUrl => {
```
znarf: typo fetchDependcies -> fetchDependencies

server/lib/subscriptions.js
```diff
@@ -328,3 +331,98 @@ export async function sendThankYouEmail(order, transaction) {
     },
   );
 }
+
+export function getNextDispatchingDate(interval, currentDispatchDate) {
```
znarf: Put everything in a new file instead of adding it to `subscriptions`.

server/graphql/v1/mutations/orders.js
```diff
@@ -960,3 +965,41 @@ export async function addFundsToCollective(order, remoteUser) {
 
   return models.Order.findByPk(orderCreated.id);
 }
+
+export async function createOrderForDispatch(order, loaders, remoteUser, reqIp) {
```
znarf: - the mutation should be named `dispatchOrder`
- it should take as argument one **already processed** order
- and should return a list of the generated Orders, with the proper GraphQL type

server/graphql/v1/types.js
```diff
@@ -1688,8 +1689,20 @@ export const OrderType = new GraphQLObjectType({
       data: {
         type: GraphQLJSON,
         description: 'Additional information on order: tax and custom fields',
-        resolve(order) {
-          return pick(order.data, ['tax', 'customData']) || null;
+        async resolve(order) {
+          const data = pick(order.data, ['tax', 'customData', 'lastDispatchedOrdersId']) || null;
```
znarf: If moving with the other suggestions, this should not be needed.

server/lib/dispatcher.js
```diff
@@ -0,0 +1,112 @@
...
+    // Check if the collective is avaliable
+    const collective = await models.Collective.findByPk(dependency.opencollective.id);
+    const totalAmount = computeAmount(shareableAmount, sumOfWeights, dependency.weight);
+
+    const orderData = {
+      CreatedByUserId: order.CreatedByUserId,
+      FromCollectiveId: order.FromCollectiveId,
+      CollectiveId: collective.id,
+      quantity: order.quantity,
+      description: order.description,
```
znarf: The description is not right, it should be something custom.

ie: `Monthly donation to {collective.name} through Back Your Stack`

PR #2173, opened 2019-06-27
Fix deleting application

znarf:


Betree:
See https://github.com/opencollective/opencollective/issues/2166

PR #2174, opened 2019-06-27
fix(applications): Set Int for ID type for delete mutation

Betree:
For https://github.com/opencollective/opencollective/issues/2166


Betree:
Closing in favor of https://github.com/opencollective/opencollective-api/pull/2173

PR #2193, opened 2019-07-02
Update countryISO from UK to GB

flickz:
Fix [#2184](https://github.com/opencollective/opencollective/issues/2184)


znarf:
This was introduced in https://github.com/opencollective/opencollective-api/blob/master/migrations/20190321222546-populate-collectives-countries.js cc @Betree

PR #2208, opened 2019-07-04
chore(Contributor): Add type field

Betree:
This will add the collective type (`COLLECTIVE`, `USER`...etc) to contributors.

Ping @znarf


znarf:
I harcoded it as `USER`where we were discussing it. Maybe you want to change that.

Betree:
@znarf Do you remember where it was?

znarf:
@Betree ContributorsGrid in frontend

PR #2268, opened 2019-07-16
chore(editCollective): Better check for permissions

Betree:
This PR also disable the ability to edit collective `type` and `ParentCollectiveId`.

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -316,9 +316,8 @@ export function editCollective(_, args, req) {
   }
 
   const newCollectiveData = {
-    ...omit(args.collective, ['location']),
+    ...omit(args.collective, ['location', 'type', 'ParentCollectiveId']),
     LastEditedByUserId: req.remoteUser.id,
-    type: args.collective.type || 'COLLECTIVE',
   };
 
   // Set location values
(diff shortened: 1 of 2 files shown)


server/graphql/v1/mutations/collectives.js
```diff
@@ -316,9 +316,8 @@ export function editCollective(_, args, req) {
   }
 
   const newCollectiveData = {
-    ...omit(args.collective, ['location']),
```
znarf: For the future, maybe we better whitelist here.

server/graphql/v1/mutations/collectives.js
```diff
@@ -341,52 +340,41 @@ export function editCollective(_, args, req) {
...
     .then(() => {
-      // If trying to edit parent collective, ensure user is an admin of it
-      const newParentCollectiveId = args.collective.ParentCollectiveId;
-      const isChangingParent = newParentCollectiveId && newParentCollectiveId !== collective.ParentCollectiveId;
-      if (isChangingParent && !req.remoteUser.isAdmin(newParentCollectiveId)) {
-        throw new errors.Unauthorized({
-          message: `You must be logged in as the creator of this Event or as an admin of the ${parentCollective.slug} collective to edit this Event Collective`,
+      if (collective.ParentCollectiveId) {
+        return req.loaders.collective.findById.load(collective.ParentCollectiveId).then(pc => {
+          if (!pc) {
```
znarf: Not sure I would throw an exception here.

If the ParentCollectiveId is invalid, then we can't edit the Collective anymore.

It's not supposed to happen, but I don't think there is a point checking that here.

Betree: Not supposed to happen indeed but I'd rather have this check in place, otherwise the behaviour when `ParentCollectiveId` is incorrect will be harder to predict: `isAdmin` will return false in https://github.com/opencollective/opencollective-api/blob/94379ff62f64b5f36ccbdef45ff8fc8045855508/server/graphql/v1/mutations/collectives.js#L369 but you have to know (or go down the code) to understand that `isAdmin(null) === false`

znarf: Ok. This needs a strong refactor anyway (the pattern to load `collective` and `parentCollective` is ugly) so that will certainly be tackled at this time.

PR #2269, opened 2019-07-16
Refactor createMember / removeMember

znarf:
When auditing mutations I found that these functions could  be way clearer with a small refactor.

Here we are.

- use await / async
- rename variables
- clearer ACLs

### DIFF
--- a/server/graphql/v1/mutations/members.js
+++ b/server/graphql/v1/mutations/members.js
@@ -4,93 +4,80 @@ import models from '../../../models';
 import errors from '../../../lib/errors';
 import roles from '../../../constants/roles';
 
-export function createMember(_, args, req) {
-  let collective;
-
-  const checkPermission = () => {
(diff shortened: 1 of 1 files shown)


server/graphql/v1/mutations/members.js
```diff
@@ -4,93 +4,80 @@ import models from '../../../models';
...
+  if (args.member.id) {
+    user = await models.User.findOne({ where: { CollectiveId: args.member.id } });
+  }
+  if (!user) {
+    user = await models.User.findOrCreateByEmail(args.member.email, args.member);
+  }
 
-  const checkPermission = () => {
+  return models.Member.create({
+    // NOTE: doesn't look like a good idea to set CreatedByUserId != req.remoteUser.id
```
Betree: Agree, I was worried about this too. It seemed fine in the code I inspected though.

PR #2276, opened 2019-07-17
Introduce hash in imageUrl based on image property

znarf:
if used, will help bypass cache on update


znarf:
@Betree agree for the database storage

PR #2307, opened 2019-07-23
Feat/add expense type

pietgeursen:
Adds a migration for an expense type so that we can record whether an expense is a "Receipt" or an "Invoice".

Expenses of type "Receipt" should not be counted towards tax form totals and this PR tests that too.

This will need to merge at the same time as [this PR](https://github.com/opencollective/opencollective-frontend/pull/2232)

### DIFF
--- a/migrations/20190723034059-add-type-to-expense-table.js
+++ b/migrations/20190723034059-add-type-to-expense-table.js
@@ -1,0 +1,12 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.addColumn('Expenses', 'type', {
+      type: Sequelize.STRING(),
+    });
+  },
+  down: (queryInterface, Sequelize) => {
+    return queryInterface.removeColumn('Expenses', 'type');
+  },
+};
--- a/migrations/20190723034059-add-type-to-expense-table.js
+++ b/migrations/20190723034059-add-type-to-expense-table.js
@@ -5,6 +5,7 @@ module.exports = {
     return Promise.all([
       queryInterface.addColumn('Expenses', 'type', {
         type: Sequelize.ENUM('RECEIPT', 'INVOICE', 'UNCLASSIFIED'),
+        defaultValue: 'UNCLASSIFIED',
       }),
       queryInterface.sequelize.query(`
         UPDATE "Expenses"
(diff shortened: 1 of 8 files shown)


server/models/Expense.js
```diff
@@ -83,7 +84,7 @@ export default function(Sequelize, DataTypes) {
       status: {
         type: DataTypes.STRING,
         defaultValue: status.PENDING,
-        allowNull: false,
+        allowNull: true,
```
Betree: This won't work because there is also a `NOT NULL` constraint in the DB for this field, DB and model should be in sync. However I don't understand why we need this?

migrations/20190723034059-add-type-to-expense-table.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.addColumn('Expenses', 'type', {
+      type: Sequelize.STRING(),
```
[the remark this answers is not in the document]
pietgeursen: Yeah I did this originally and it would be my preferred option, but I couldn't work out how to make the down migration work because it doesn't remove the enum.

I'll have another look at it. I might just have to write some raw sql.

server/models/Expense.js
pietgeursen: Woops, that's just a total mistake and should not be there. I must have edited the wrong property. Good catch!

migrations/20190723034059-add-type-to-expense-table.js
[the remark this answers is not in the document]
Betree: Enums are good if that's not too complicated. Apparently this can be done with (haven't tried):

```es6
return queryInterface.sequelize.query(
  'DROP TYPE "EnumName";'
);
```

Crazy that sequelize doesn't provide a helper for this :fearful:

migrations/20190723034059-add-type-to-expense-table.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return Promise.all([
+      queryInterface.addColumn('Expenses', 'type', {
```
Betree: Any column added to the `Expenses` table must be added to `ExpenseHistories` as well. Same goes for the followings:

> 'Collectives', 'Comments', 'Expenses', 'Subscriptions', 'Updates'

You were not supposed to know it, it's not documented. I'm adding a test right now to ensure that we keep the models in sync: https://github.com/opencollective/opencollective-api/pull/2323

In the mean time you can check https://github.com/opencollective/opencollective-api/blob/53ea1c189c83ae506e53155359ddb1c22e8c6079/migrations/20190327172521-add-is-pledged-to-collectives.js#L14 for an example of how we do that.

PR #2363, opened 2019-08-03
Changed button link from view 'Latest Expenses' to 'View Expenses' as…

Gbahdeyboh:
… requested in issue #2203


package-lock.json
```diff
@@ -1,15274 +0,0 @@
-{
```
Betree: Should not be deleted

templates/emails/collective.expense.approved.hbs
```diff
@@ -30,7 +30,7 @@ Subject: Your expense to {{collective.name}} for {{{currency expense.amount curr
   </a>
 
   <br /><br />
-  <a href="{{actions.viewLatestExpenses}}" class="btn"><div>View Latest Expenses</div></a>
+  <a href="{{actions.viewLatestExpenses}}" class="btn"><div>View Expense</div></a>
```
Betree: Changing the label only makes sense if we change the link, the goal here is to redirect to the expense URL

znarf:
Closing because of inactivity. Feel free to create a new PR when committing new changes.

PR #2370, opened 2019-08-05
Add ApprovedAt to Collectives

flickz:
This PR attempt to fix [#1878](https://github.com/opencollective/opencollective/issues/1878) && [#2100](https://github.com/opencollective/opencollective/issues/2100).

- Adds `approvedAt` to Collective table
- Add migration file to set the value of `collective.approvedAt` to colllective host member `since` date.
- Update the value of `approvedAt` when collective is approved, change host or archived.
- Add `isApproved` property to `Collective` query.
- Remove collective host in `archiveCollective` mutation.


znarf:
We should avoid complex behavior in migrations (20190805114605-set-approved-collective-approvedAt-column-value.js). We should ideally stick to pure SQL.

- In `20190805102047-add-approvedAt-column-to-collective.js`, can you add `approvedAt` to a default value like today for approved collectives?
- Can you move `20190805114605-set-approved-collective-approvedAt-column-value.js` in  `scripts` and adjust the syntax so that it's not a mutation anymore? We will trigger it manually.

flickz:
Okay, let me adjust that.

server/graphql/v1/mutations/collectives.js
```diff
@@ -634,7 +634,19 @@ export async function archiveCollective(_, args, req) {
...
+    where: {
+      CollectiveId: collective.id,
+      MemberCollectiveId: collective.HostCollectiveId,
+      role: roles.HOST,
+    },
+  });
+
+  if (membership) {
+    membership.destroy();
+  }
```
znarf: Thank you for adding that. I think we need to also add `HostCollectiveId: null`in `collective.update`.

flickz: Yes we do, I just added it.

migrations/20190805102047-add-approvedAt-column-to-collective.js
```diff
@@ -0,0 +1,23 @@
...
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    const colParms = {
+      type: Sequelize.DATE,
+    };
+    await queryInterface.addColumn('Collectives', colName, colParms);
+    await queryInterface.addColumn('CollectiveHistories', colName, colParms);
+    return queryInterface.sequelize.query(`
+      UPDATE "Collectives" c
+        SET "approvedAt" = '2019-08-06'
```
znarf: Sorry, also missed that until now. I think we should limit the update to `type=COLLECTIVE`. Also `HostCollectiveId IS NOT NULL`to be on the safe side.

flickz: Okay.

PR #2386, opened 2019-08-07
Fix: Prevent collective from having zero admin.

flickz:
This PR fix [#2289](https://github.com/opencollective/opencollective/issues/2289), by basically adding a condition that prevents a collective admin member from removing self from a collective.

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1600,11 +1600,7 @@ export default function(Sequelize, DataTypes) {
     })
       .then(oldMembers => {
         // remove the members that are not present anymore
-        const diff = differenceBy(
-          oldMembers.map(t => ({ id: t.id, MemberCollectiveId: t.MemberCollectiveId })),
-          members.map(t => ({ id: t.id, MemberCollectiveId: t.MemberCollectiveId })),
-          'id',
-        );
+        const diff = differenceBy(oldMembers, members, 'id');
         if (diff.length === 0) {
           return null;
         } else {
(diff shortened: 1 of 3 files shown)


server/models/Collective.js
```diff
@@ -1587,12 +1600,22 @@ export default function(Sequelize, DataTypes) {
     })
       .then(oldMembers => {
         // remove the members that are not present anymore
-        const diff = difference(oldMembers.map(t => t.id), members.map(t => t.id));
+        const diff = differenceBy(
+          oldMembers.map(t => ({ id: t.id, MemberCollectiveId: t.MemberCollectiveId })),
```
Betree: It `.map(t => ({ id: t.id, MemberCollectiveId: t.MemberCollectiveId }))` necessary? As we're extracting the `id` later with `const diffMemberIds = diff.map(m => m.id);` maybe this could work by doing simply `differenceBy(oldMembers, members, 'id')`

znarf:
Let's back that with test.

server/models/Collective.js
```diff
@@ -1600,11 +1600,7 @@ export default function(Sequelize, DataTypes) {
     })
       .then(oldMembers => {
         // remove the members that are not present anymore
-        const diff = differenceBy(
-          oldMembers.map(t => ({ id: t.id, MemberCollectiveId: t.MemberCollectiveId })),
-          members.map(t => ({ id: t.id, MemberCollectiveId: t.MemberCollectiveId })),
-          'id',
-        );
+        const diff = differenceBy(oldMembers, members, 'id');
```
znarf: Great.

PR #2390, opened 2019-08-08
PayPal: Ensure fees and total amount are truncated

Betree:
Resolve https://opencollective.freshdesk.com/a/tickets/1105
The reason why I couldn't reproduce on staging was because the PayPal fees depends on the environment. However by simulating 56 cents fees for PayPal I was able to reproduce:

> 0.56 * 100
56.00000000000001

This fix ensure we truncate the cent amounts for both the total amount and the fees.

### DIFF
--- a/server/paymentProviders/paypal/payment.js
+++ b/server/paymentProviders/paypal/payment.js
@@ -100,15 +100,15 @@ export async function executePayment(order) {
 
 /** Create transaction in our database to reflect a PayPal charge */
 export async function createTransaction(order, paymentInfo) {
-  /* The `* 100` in the next lines convert from PayPal format in
-     dollars to Open Collective format in cents */
   const transaction = paymentInfo.transactions[0];
-  const amountFromPayPal = parseFloat(transaction.amount.total) * 100;
-  const paypalFee = parseFloat(get(transaction, 'related_resources.0.sale.transaction_fee.value', '0.0')) * 100;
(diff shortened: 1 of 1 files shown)


server/paymentProviders/paypal/payment.js
```diff
@@ -100,15 +100,15 @@ export async function executePayment(order) {
...
 export async function createTransaction(order, paymentInfo) {
-  /* The `* 100` in the next lines convert from PayPal format in
-     dollars to Open Collective format in cents */
   const transaction = paymentInfo.transactions[0];
-  const amountFromPayPal = parseFloat(transaction.amount.total) * 100;
-  const paypalFee = parseFloat(get(transaction, 'related_resources.0.sale.transaction_fee.value', '0.0')) * 100;
+  const amountFromPayPal = parseFloat(transaction.amount.total);
+  const paypalFee = parseFloat(get(transaction, 'related_resources.0.sale.transaction_fee.value', '0.0'));
+  const amountFromPayPalInCents = Math.trunc(amountFromPayPal * 100);
+  const paypalFeeInCents = Math.trunc(paypalFee * 100);
```
znarf: @Betree Just curious, why `Math.trunc` vs `Math.round`?

Betree: The examples I used were rounded above the value. I just made some more research and found out that this is not always the case:

> 0.29 * 100 
28.999999999999996

I'll make a PR with an update, thanks for pointing that out!

PR #2391, opened 2019-08-08
PaymentMethods: Remove with `destroy` method

Betree:
Change the behaviour introduced in https://github.com/opencollective/opencollective-api/pull/1823 to remove payment methods rather than setting their `CollectiveId` to `null` (we probably want to keep this info). I've checked to ensure payment methods are properly soft-deleted (`deletedAt = NOW()`).

Also changed this for `removePaymentMethod` (the endpoint called from /edit) to use `deletedAt` rather than `archivedAt`.

@znarf do you know what's supposed to be the difference between `archivedAt` and `deletedAt`? Is it just legacy or does it have a meaning?


znarf:
Let's look at that once Stripe CSA is merged.

Betree:
@znarf is this still required, should I solve the conflicts?

PR #2407, opened 2019-08-12
BackYourStack dispatcher

znarf:
Continuing from https://github.com/opencollective/opencollective-api/pull/2153

### DIFF
--- a/server/lib/backyourstack/dispatcher.js
+++ b/server/lib/backyourstack/dispatcher.js
@@ -41,17 +41,28 @@ export async function dispatchFunds(order) {
   });
   const shareableAmount = transaction.netAmountInCollectiveCurrency;
   const jsonUrl = order.data.customData.jsonUrl;
-  let depRecommendation;
+  const collectives = [];
+  let depRecommendations;
 
   try {
-    depRecommendation = await fetchDependencies(jsonUrl);
+    depRecommendations = await fetchDependencies(jsonUrl);
   } catch (err) {
     debug('Error fetching dependcies', err);
     console.error(err);
     throw new Error('Unable to fetch dependencies, please ensure the url is correct');
   }
 
-  const sumOfWeights = depRecommendation.reduce((sum, dependency) => dependency.weight + sum, 0);
+  for (const depRecommended of depRecommendations) {
+    const collective = await models.Collective.findOne({ where: { slug: depRecommended.opencollective.slug } });
+    if (!collective) {
+      console.log(`Unable to fetch collective with slug ${depRecommended.opencollective.slug}`);
+      return;
+    }
+    collective.totalAmount = computeAmount(shareableAmount, sumOfWeights, depRecommended.weight);
+    collectives.push(collective);
+  }
+
+  const sumOfWeights = collectives.reduce((sum, dependency) => dependency.weight + sum, 0);
   let HostCollectiveId;
   if (!order.collective) {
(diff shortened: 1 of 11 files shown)


server/lib/backyourstack/dispatcher.js
```diff
@@ -0,0 +1,120 @@
...
+    type: 'prepaid',
+    uuid: uuidV4(),
+    data: { HostCollectiveId },
+  });
+
+  return map(
+    depRecommendation,
+    async dependency => {
+      // Check if the collective is avaliable
+      const collective = await models.Collective.findOne({ where: { slug: dependency.opencollective.slug } });
```
znarf: @flickz this should be refactored so we filter existing collectives before. The goal is to compute `sumOfWeights` only after the collectives are filtered.

flickz: I've updated this now

server/lib/backyourstack/dispatcher.js
```diff
@@ -0,0 +1,123 @@
...
+    debug('Error fetching dependcies', err);
+    console.error(err);
+    throw new Error('Unable to fetch dependencies, please ensure the url is correct');
+  }
+
+  for (const depRecommended of depRecommendations) {
+    const collective = await models.Collective.findOne({ where: { slug: depRecommended.opencollective.slug } });
+    if (!collective) {
+      console.log(`Unable to fetch collective with slug ${depRecommended.opencollective.slug}`);
+      return;
```
znarf: Have you rely tried that?

server/lib/backyourstack/dispatcher.js
```diff
@@ -0,0 +1,123 @@
...
+    }
+    collective.totalAmount = computeAmount(shareableAmount, sumOfWeights, depRecommended.weight);
```
znarf: Is `sumOfWeights` defined at this point?

server/lib/backyourstack/dispatcher.js
```diff
@@ -0,0 +1,123 @@
...
+        CreatedByUserId: order.CreatedByUserId,
+        FromCollectiveId: order.FromCollectiveId,
+        CollectiveId: collective.id,
+        quantity: order.quantity,
+        description: `Monthly donation to ${collective.name} through BackYourStack`,
+        totalAmount: collective.totalAmount,
+        currency: order.currency,
+        status: status.PENDING,
+      };
+      delete collective.totalAmount;
```
znarf: What's the purpose of that?

server/paymentProviders/opencollective/prepaid.js
```diff
@@ -80,9 +80,17 @@ async function processOrder(order) {
     throw new Error("This payment method doesn't have enough funds to complete this order");
   }
 
+  let hostFeeInHostCurrency, platformFeeInHostCurrency;
+
+  if (options.skipPlatformFee && options.skipHostFee) {
```
znarf: That doesn't make sense. You should handle these properties individually.

server/lib/backyourstack/dispatcher.js
znarf: This needs to be addressed.

server/paymentProviders/opencollective/prepaid.js
flickz: Oh thought they were together. I'll change it

PR #2428, opened 2019-08-19
Stripe CSA

znarf:
Backend side of https://github.com/opencollective/opencollective/issues/2340


server/graphql/v1/mutations/orders.js
```diff
@@ -528,39 +527,75 @@ export async function createOrder(order, loaders, remoteUser, reqIp) {
...
+    ],
+  });
+  if (!existingOrder) {
+    throw new errors.NotFound({ message: 'Order not found' });
+  }
+
+  if (!remoteUser.isAdmin(existingOrder.FromCollectiveId)) {
+    throw new errors.Unauthorized({ message: "You don't have permission to confirm this order" });
+  }
+
```
Betree: Maybe we should also check that the order is not already confirmed to prevent any double spending?

server/graphql/v1/types.js
```diff
@@ -1653,6 +1653,24 @@ export const OrderType = new GraphQLObjectType({
...
+      stripeAccount: {
+        type: GraphQLString,
+        resolve(order) {
+          return order.stripeAccount;
+        },
+      },
+      stripeResponse: {
+        type: GraphQLJSON,
+        resolve(order) {
+          return order.stripeResponse;
```
Betree: Is there any risk of leaking private info if we return the Stripe response?

server/lib/subscriptions.js
```diff
@@ -88,6 +88,13 @@ export async function processOrderWithSubscription(options, order) {
         transaction = await paymentsLib.processOrder(order);
         orderProcessedStatus = 'success';
       } catch (error) {
+        if (error.stripeResponse) {
+          if (error.stripeResponse.paymentIntent) {
```
Betree: Ideally `subscriptions` should be agnostic about what Stripe is, it's the responsibility of `server/paymentProviders/stripe/*` to deal with Stripe errors

server/graphql/v1/mutations/orders.js
```diff
@@ -528,39 +527,80 @@ export async function createOrder(order, loaders, remoteUser, reqIp) {
...
+  });
+
+  if (!existingOrder) {
+    throw new errors.NotFound({ message: 'Order not found' });
+  }
+  if (!remoteUser.isAdmin(existingOrder.FromCollectiveId)) {
+    throw new errors.Unauthorized({ message: "You don't have permission to confirm this order" });
+  }
+  if (existingOrder.status !== 'ERROR' && existingOrder.status !== 'PENDING') {
+    throw new errors.NotFound({ message: 'Order can only be confirmed if ERROR or PENDING.' });
```
Betree: `NotFound` doesn't feels to me like the right error type. Would be great to have a `ActionAlreadyDone` (or similar) error type for this case, I'd like to be able to catch it in the Frontend to show it as an info (`This order has alrady been confirmed`) rather than an error.

znarf: Good that you spotted that. I absolutely didn't want to do that, copy/paste accident!

migrations/20190822070158-add-saved-column-to-payment-methods.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('PaymentMethods', 'saved', {
+      type: Sequelize.BOOLEAN,
+      defaultValue: false,
```
Betree: If default is `false`, is there not a risk of not seeing all the other payment methods when they're created? I'm thinking about `virtualcard` and `collective` types in particular

server/graphql/v1/mutations/orders.js
```diff
@@ -118,9 +118,12 @@ async function checkRecaptcha(order, remoteUser, reqIp) {
 }
 
 export async function createOrder(order, loaders, remoteUser, reqIp) {
+  // console.log(order);
```
Betree:
```suggestion
```

server/graphql/v1/mutations/orders.js
```diff
@@ -510,36 +504,98 @@ export async function createOrder(order, loaders, remoteUser, reqIp) {
...
+      { model: models.Collective, as: 'fromCollective' },
+      { model: models.PaymentMethod, as: 'paymentMethod' },
+      { model: models.Subscription, as: 'Subscription' },
+    ],
+  });
+
+  if (!order) {
+    throw new errors.NotFound({ message: 'Order not found' });
+  }
+  if (!remoteUser.isAdmin(order.FromCollectiveId)) {
```
Betree: Tricky case: if order is made with a gift card, to one that will need to confirm it is the gift card emitter - not the person ordering. In such case we'll need to check `PaymentMethod.CollectiveId`

server/graphql/v1/mutations/orders.js
```diff
@@ -510,36 +504,98 @@ export async function createOrder(order, loaders, remoteUser, reqIp) {
...
+      await order.save();
     }
 
     return order;
   } catch (error) {
-    debug('createOrder mutation error: ', error);
-    if (orderCreated && !orderCreated.processedAt) {
-      // TODO: Order should be updated with data JSON field to store the error to review later
-      orderCreated.update({ status: status.ERROR });
+    // console.log(error);
```
Betree:
```suggestion
```

server/graphql/v1/mutations/orders.js
```diff
@@ -688,34 +744,51 @@ export async function updateSubscription(remoteUser, args) {
...
+        await order.Subscription.update({
+          nextChargeDate: updatedDates.nextChargeDate,
+          chargeRetryCount,
+        });
+      }
 
-    order = await order.update({ PaymentMethodId: newPm.id });
+      order = await order.update({ PaymentMethodId: newPm.id });
+    } catch (error) {
+      console.log(error);
```
Betree:
```suggestion
```

migrations/20190822070158-add-saved-column-to-payment-methods.js
znarf: What would you recommend?

znarf: Technically, I decided to only use this "saved" property for credit cards. For now.

PR #2433, opened 2019-08-20
Prevent admin from changing role without any other admin in collective.

flickz:
This PR follows up https://github.com/opencollective/opencollective/issues/2289#issuecomment-522887742

Screenshot 2019-08-20 at 8 18 14 PM

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1621,17 +1621,12 @@ export default function(Sequelize, DataTypes) {
           return models.Member.update({ deletedAt: new Date() }, { where: { id: { [Op.in]: diffMemberIds } } });
         }
       })
-      .then(async () => {
-        const oldAdminMembers = await this.getMembers({
-          where: { role: { [Op.in]: [roles.ADMIN] } },
-        });
-        const adminMembers = members.filter(m => m.role === roles.ADMIN);
-        const diff = differenceBy(oldAdminMembers, adminMembers, 'role');
-
-        if (oldAdminMembers.length === 1 && diff.length !== 0) {
-          throw new Error('A collective requires having at least one admin member.');
+      .then(() => {
+        if (members.filter(m => m.role === roles.ADMIN).length === 0) {
+          throw new Error('There must always be at least one collective admin');
         }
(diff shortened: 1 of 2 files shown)


znarf:
Looks good for me. @Betree can you also check?

server/models/Collective.js
```diff
@@ -1621,8 +1621,17 @@ export default function(Sequelize, DataTypes) {
           return models.Member.update({ deletedAt: new Date() }, { where: { id: { [Op.in]: diffMemberIds } } });
         }
       })
-      .then(() => {
-        return Promise.map(members, member => {
+      .then(async () => {
+        const oldAdminMembers = await this.getMembers({
+          where: { role: { [Op.in]: [roles.ADMIN] } },
```
Betree: Just a suggestion

```suggestion
          where: { role: roles.ADMIN },
```

server/models/Collective.js
```diff
@@ -1621,8 +1621,17 @@ export default function(Sequelize, DataTypes) {
...
+        });
+        const adminMembers = members.filter(m => m.role === roles.ADMIN);
+        const diff = differenceBy(oldAdminMembers, adminMembers, 'role');
```
Betree: Shouldn't we make the difference by `id` rather than `role`? All `role`s in `oldAdminMembers` and `adminMembers` are `ADMIN` right?

server/models/Collective.js
```diff
@@ -1621,8 +1621,17 @@ export default function(Sequelize, DataTypes) {
...
+
+        if (oldAdminMembers.length === 1 && diff.length !== 0) {
```
Betree: I don't really understand this, will it work if there are two admins in my collective and I change the role to `MEMBER` for both of them?

Betree:
@flickz Actually we don't even need to check the remove. `members` is supposed to contains the final list of members, so we can check every possible case simply with a simple test at the top:

```es6
if (members.filter(m => m.role === roles.ADMIN).length === 0) {
  throw new Error('There must always be at least one collective admin');
}
```

flickz:
@Betree Perfect! Thank you.

PR #2439, opened 2019-08-21
Changes helloworks sdk to be Piet's fork

pietgeursen:
Something is broken with version 1.1.0 of the official SDK.

I've created an [issue here](https://github.com/hellosign/helloworks-nodejs-sdk/issues/9)

I've tested that [my fork](https://github.com/pietgeursen/helloworks-nodejs-sdk) works on staging.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2890,7 +2890,7 @@
       "dependencies": {
         "callsites": {
           "version": "2.0.0",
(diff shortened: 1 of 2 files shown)


znarf:
@pietgeursen let's be more strict with the commit messages. it's not great to have such in the project history.

Here that could be:

 ```
fix(hello-works): temporary switch back to previous fork of helloworks-sdk

see upstream issue: https://github.com/hellosign/helloworks-nodejs-sdk/issues/9
```

PR #2444, opened 2019-08-22
Kill referrals

znarf:
I decided to not remove the data for now, in case we still want to access it in the near future.


znarf:
@znarf I will work on the frontend part too.

PR #2468, opened 2019-08-28
Correctly fetch incognito profiles for admins

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2371 and https://opencollective.freshdesk.com/a/tickets/1260

This will ensure that we still get the user details even when using incognito mode. Previously features like Event > Export CSV were missing personal information for incognito users.

### DIFF
--- a/server/graphql/loaders.js
+++ b/server/graphql/loaders.js
@@ -5,6 +5,7 @@ import DataLoader from 'dataloader';
 import dataloaderSequelize from 'dataloader-sequelize';
 import { get, groupBy } from 'lodash';
 import debugLib from 'debug';
+import queries from '../lib/queries';
 
 dataloaderSequelize(models.Order);
 dataloaderSequelize(models.Transaction);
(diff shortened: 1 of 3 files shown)


xdamman:
Thanks for taking care of this. It wasn’t that obvious.

Just making sure: the rule should be that only admins of collective or host can access the email addresses of incognito users. There is a catch though compared to how we used to check if one could access someone’s private infos: before we would just check that you were an admin of one of the collectives where the user is a member of. But in this case, this is not enough. Because I could have donated to collective A as xdamman, to collective B as incognito. If you are admin of Collective A, you should not be able to get my email when exporting the list of members of Collective B (whereas it didn’t matter before).

Betree:
> Thanks for taking care of this. It wasn’t that obvious.
> 
> Just making sure: the rule should be that only admins of collective or host can access the email addresses of incognito users. There is a catch though compared to how we used to check if one could access someone’s private infos: before we would just check that you were an admin of one of the collectives where the user is a member of. But in this case, this is not enough. Because I could have donated to collective A as xdamman, to collective B as incognito. If you are admin of Collective A, you should not be able to get my email when exporting the list of members of Collective B (whereas it didn’t matter before).

That's a very good point. It increases a lot the complexity for this fix and I need to deliver something very quickly to help WWC that has an event coming in less than two days and need the emails of their attendees. Here are the steps I've chosen to follow:
- Send attendee's details (email + name) to WWC including incognito users
- Close this pull request
- Disable ordering tickets as incognito in the contribution flow

We'll figure out a better way of doing thing in a second time if we want to allow this behaviour.

PR #2493, opened 2019-08-30
User: Enforce email address

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2039
Probably caused by https://github.com/opencollective/opencollective/issues/1179

`null` emails were allowed in https://github.com/opencollective/opencollective-api/pull/344 but the code where it was required isn't there anymore.

All tests updated by this PR are just added an `email: randEmail()` except https://github.com/opencollective/opencollective-api/pull/2493/files#diff-c91eac63018a8e6fd82d82d8873413a2 where the logic was clearly changed to ensure we don't accept null emails.

### DIFF
--- a/migrations/20190830121917-enforce-users-emails.js
+++ b/migrations/20190830121917-enforce-users-emails.js
@@ -1,0 +1,58 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    // Add a default email to all users with active memberships
+    await queryInterface.sequelize.query(`
+      UPDATE  "Users" AS u
+      SET     "email" = FORMAT('missing-email-%s@opencollective.com', u.id)
+      FROM    "Members" AS m
+      WHERE   m."MemberCollectiveId" = u."CollectiveId"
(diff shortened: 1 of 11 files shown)


Betree:
@znarf Tests have been updated , this PR is now ready for review

migrations/20190830121917-enforce-users-emails.js
```diff
@@ -0,0 +1,36 @@
...
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    // All users without email currently in the DB have no transactions nor expenses, thus
+    // it should be safe to delete them all directly.
+    await queryInterface.sequelize.query(`
+      UPDATE  "Users"
+      SET     "deletedAt" = NOW()
+      WHERE   "email" IS NULL
+      AND     "deletedAt" IS NULL
+    `);
```
znarf: As discussed yesterday, we should also process the matching `Collective type=user`

Betree:
I'll merge this one in the next 24h if there's no blocker

PR #2503, opened 2019-09-02
Authentication enhancements

Betree:
- Enforce `algorithms` in `jwt.verify`. The lib we use is already protected, this adds an additional security to ensure no-one can mess with the algorithm used to decode the tokens.

- Add a `kid` field in JWT headers. This doesn't change anything now, but it will help with migrations strategies if we later want to update the keys or the algorithm used.

- Remove an unused function: `authenticateInternalUserByJwt`

---

New token format (decoded):

image


znarf:
Could that be a good time to review the content of the payload also?

Betree:
@znarf I've removed `email` and `id` from the payload as I could not find any usage for them (user `id` is already provided by `sub`)

PR #2531, opened 2019-09-08
Update README.md

bkmgit:
Add a bit more information on deployment of database. Such information should probably be in the main readme, not in a secondary document.


znarf:
The proposed wording is too specific, it's not just about "Ensure that users and a database can be created".

We decided to have the database setup guide separated because it's too much information for the main readme.

znarf:
Closing because the improvement is not clear enough.

PR #2532, opened 2019-09-08
Update postgres.md

bkmgit:
Add setup information for Fedora 30

### DIFF
--- a/docs/postgres.md
+++ b/docs/postgres.md
@@ -18,6 +18,40 @@ Get the app from [Postgres.app](http://postgresapp.com/). Install it.
 
 Then to enable the CLI tools, follow the steps from: https://postgresapp.com/documentation/cli-tools.html
 
+## On Fedora 30
+
+Install required packages from repositories and initialize database
+```
+sudo dnf install postgresql postgresql-server postgresql-server-devel postgis postgis-devel
+sudo postgresql-setup --initdb
+```
+
+Export configuration variables
+```
+export PGHOST=localhost
+export PGUSER=opencollective
+```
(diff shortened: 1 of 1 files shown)


docs/postgres.md
```diff
@@ -18,6 +18,40 @@ Get the app from [Postgres.app](http://postgresapp.com/). Install it.
...
+Install required packages from repositories and initialize database
+```
+sudo dnf install postgresql postgresql-server postgresql-server-devel postgis postgis-devel
+sudo postgresql-setup --initdb
+```
+
+Export configuration variables
+```
+export PGHOST=localhost
+export PGUSER=opencollective
```
znarf: What's the purpose of these, at which step are they required?

bkmgit: There are only instructions for installation and setup using Homebrew and Docker. This gives information on installation and setup on Fedora Linux without use of Docker.

znarf: I commented on line 32 specifically. I'm questioning about the `export`s.

bkmgit: These are needed in the development environment for the application to be able to use the Postgres database once Postgres has been installed and setup. The Docker instructions have these as well, but not sure if one wants to break up the installation instructions as I am not sure what happens on Mac using homebrew.

znarf: The application is not using that itself. Only `npm run db:restore` might be needing it in some circumstances.

What happens if you don't set them? I just want to avoid unnecessary steps if possible.

znarf:
This PR went stale.

PR #2548, opened 2019-09-11
Monthly report email: Empty "You have contributed" #2298

leonardohof:
Resolve opencollective/opencollective#2298

### DIFF
--- a/templates/emails/user.monthlyreport.hbs
+++ b/templates/emails/user.monthlyreport.hbs
@@ -231,7 +231,7 @@ Subject: {{{capitalize month}}} Report
               (<a href="{{../config.host.website}}/{{../fromCollective.slug}}/subscriptions">increase your contribution</a>)
               (<a href="https://twitter.com/intent/tweet?text={{{encodeURIComponent nextGoal.tweet}}}">share their goal</a>)
               {{else}}
-              You have contributed {{{currency order.totalAmount currency=order.currency}}}.
+              You have contributed {{{order.totalAmount}}}.
               (<a href="{{../config.host.website}}/{{slug}}/donate">increase your contribution</a>)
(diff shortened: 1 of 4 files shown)


leonardohof:
image

templates/emails/user.monthlyreport.hbs
```diff
@@ -231,7 +231,7 @@ Subject: {{{capitalize month}}} Report
               (<a href="{{../config.host.website}}/{{../fromCollective.slug}}/subscriptions">increase your contribution</a>)
               (<a href="https://twitter.com/intent/tweet?text={{{encodeURIComponent nextGoal.tweet}}}">share their goal</a>)
               {{else}}
-              You have contributed {{{currency order.totalAmount currency=order.currency}}}.
+              You have contributed {{{order.totalAmount}}}.
```
Betree: I don't fully understand why we need to change this because the `currency` function seems to work in other places. From my understanding, the problem is that the empty string that we set in https://github.com/opencollective/opencollective-api/blob/a156bbe47f877b52e46c3f7d7b8f4916446fb6f7/cron/monthly/user-report.js#L402 is not replaced by the real amount.

leonardohof: @Betree Cause I changed to sum all the orders and convert to string at the end on the 'user-report.js'. So, this conversion is not necessary anymore.

Betree: Alright, makes sense!

leonardohof:
> The code looks good to me and seems to work when tested locally.
> 
> Let's have another review from @znarf before merging.

Do you have any updates? @znarf?

znarf:
@leonardohof sorry, I missed it. All good for me.

PR #2567, opened 2019-09-17
Add expenseId to TransactionType to add link to individual expense ge from all transactions view

EslamHiko:
This pull request adds expenseId to Transaction type.

this pr to close this issue : https://github.com/opencollective/opencollective/issues/1590

### DIFF
--- a/server/graphql/v1/TransactionInterface.js
+++ b/server/graphql/v1/TransactionInterface.js
@@ -13,7 +13,7 @@ import {
 
 import { CollectiveInterfaceType, UserCollectiveType } from './CollectiveInterface';
 
-import { SubscriptionType, OrderType, PaymentMethodType, UserType, DateString } from './types';
+import { SubscriptionType, OrderType, PaymentMethodType, UserType, DateString, ExpenseType } from './types';
 
 export const TransactionInterfaceType = new GraphQLInterfaceType({
   name: 'Transaction',
@@ -297,13 +297,13 @@ export const TransactionExpenseType = new GraphQLObjectType({
             : null;
         },
       },
-      expenseId: {
-        type: GraphQLInt,
+      expense: {
+        type: ExpenseType,
         resolve(transaction, args, req) {
           // If it's a sequelize model transaction, it means it has the method getExpenseFromViewer
           // otherwise we return null
-          return transaction.getExpenseForViewer
-            ? transaction.getExpenseForViewer(req.remoteUser).then(expense => expense && expense.id)
+          return transaction.getExpense
+            ? transaction.getExpenseForViewer(req.remoteUser).then(expense => expense)
             : null;
         },
       },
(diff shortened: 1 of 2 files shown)


znarf:
Seems to me it's not following the [specs](https://github.com/opencollective/opencollective/issues/1590#issuecomment-517407840) from @Betree 

- I think it's a good idea to introduce `transaction.expense` as @Betree  suggested. Exposing `transaction.expense.id` is more elegant and future proof than `transaction.expenseId`.
- I'm not sure for the Dataloader, `getExpenseForViewer` does an important job at filtering the information. This is required and I'm not sure if that is appropriate to have this behavior in a Dataloader.

EslamHiko:
@znarf I updated my pr we now have all the expense object filtered through getExpenseForViewer. I suggest wrapping category,attachment,privateMessage in the new expense field, what do you think?

server/graphql/v1/TransactionInterface.js
```diff
@@ -297,13 +297,13 @@ export const TransactionExpenseType = new GraphQLObjectType({
...
-      expenseId: {
-        type: GraphQLInt,
+      expense: {
+        type: ExpenseType,
         resolve(transaction, args, req) {
           // If it's a sequelize model transaction, it means it has the method getExpenseFromViewer
           // otherwise we return null
-          return transaction.getExpenseForViewer
-            ? transaction.getExpenseForViewer(req.remoteUser).then(expense => expense && expense.id)
+          return transaction.getExpense
```
znarf: Let's keep `getExpenseForViewer` here, no?

znarf:
>  I suggest wrapping category, attachment, privateMessage in the new expense field

In theory yes, but in practice we need to keep compatibility, so it it should stay like that for now.

We can start trying to get rid of it in the frontend, but this should be another PR.

server/graphql/v1/TransactionInterface.js
EslamHiko: I think `getExpenseForViewer` is better for the next pr that will wrap the other attributes, or we change it for `getExpense` for now?

znarf: Let's use `getExpenseForViewer`, I think it's what we want now, and for the future.

EslamHiko: @znarf done.

znarf: Sorry, I just checked with @Betree and he confirmed he would like to use Dataloader as it was originally written in the spec. Instead of `getExpense` /  `getExpenseForViewer`.

1) introduce an expense dataloader `expense.findById` similar to https://github.com/opencollective/opencollective-api/blob/master/server/graphql/loaders.js#L56
2) use it, ie: `return req.loaders.expense.findById.load(transaction.ExpenseId);`

EslamHiko: @znarf Done.

EslamHiko:
@znarf Working on it.

PR #2578, opened 2019-09-19
deprecate getExpenseForViewer

EslamHiko:
This Pr is a follow up to this PR: https://github.com/opencollective/opencollective-api/pull/2567

Also, I updated the front end in this PR: https://github.com/opencollective/opencollective-frontend/pull/2575

### DIFF
--- a/server/graphql/loaders.js
+++ b/server/graphql/loaders.js
@@ -55,7 +55,15 @@ export const loaders = req => {
   return {
     expense: {
       findById: new DataLoader(ids =>
-        models.Expense.findAll({ where: { id: { [Op.in]: ids } } }).then(results => sortResults(ids, results, 'id')),
+        models.Expense.findAll({ where: { id: { [Op.in]: ids } } }).then(results => {
+          const viewer = req.remoteUser;
+          return sortResults(ids, results, 'id').map(result => {
+            if (!result) return null;
+            if (viewer && viewer.isAdmin(result.CollectiveId)) return result.info;
+            if (viewer && viewer.id === result.UserId) return result.info;
+            return result.public;
+          });
+        }),
       ),
     },
     collective: {
--- a/server/graphql/loaders.js
+++ b/server/graphql/loaders.js
@@ -56,13 +56,7 @@ export const loaders = req => {
     expense: {
       findById: new DataLoader(ids =>
         models.Expense.findAll({ where: { id: { [Op.in]: ids } } }).then(results => {
-          const viewer = req.remoteUser;
-          return sortResults(ids, results, 'id').map(result => {
-            if (!result) return null;
-            if (viewer && viewer.isAdmin(result.CollectiveId)) return result.info;
-            if (viewer && viewer.id === result.UserId) return result.info;
-            return result.public;
-          });
+          return sortResults(ids, results, 'id');
         }),
       ),
     },
--- a/server/graphql/v1/TransactionInterface.js
+++ b/server/graphql/v1/TransactionInterface.js
@@ -279,7 +279,9 @@ export const TransactionExpenseType = new GraphQLObjectType({
       expense: {
         type: ExpenseType,
         resolve(transaction, args, req) {
-          return req.loaders.expense.findById.load(transaction.ExpenseId);
+          if (transaction.ExpenseId) {
+            return req.loaders.expense.findById.load(transaction.ExpenseId);
+          }
         },
       },
     };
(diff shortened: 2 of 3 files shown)


Betree:
@znarf was it specified that we wanted to remove `getExpenseForViewer`?

znarf:
@Betree it's now effectively useless and should be removed. However, I have problems with other aspects of this PR and have yet to provide clear feedback: modification to dataLoader, straight removal of properties instead of deprecation.

EslamHiko:
@znarf @Betree Friendly ping.

server/graphql/loaders.js
```diff
@@ -55,7 +55,15 @@ export const loaders = req => {
...
-        models.Expense.findAll({ where: { id: { [Op.in]: ids } } }).then(results => sortResults(ids, results, 'id')),
+        models.Expense.findAll({ where: { id: { [Op.in]: ids } } }).then(results => {
+          const viewer = req.remoteUser;
+          return sortResults(ids, results, 'id').map(result => {
+            if (!result) return null;
+            if (viewer && viewer.isAdmin(result.CollectiveId)) return result.info;
+            if (viewer && viewer.id === result.UserId) return result.info;
+            return result.public;
+          });
+        }),
```
znarf: That should not be modified. There are individual checks on the private properties that make this code not needed. See https://github.com/opencollective/opencollective-api/blob/master/server/graphql/v1/types.js#L662 etc ...

server/graphql/v1/TransactionInterface.js
```diff
@@ -277,40 +276,12 @@ export const TransactionExpenseType = new GraphQLObjectType({
           return transaction.description || expense;
         },
       },
-      privateMessage: {
```
znarf: At this stage, these properties should be deprecated but not deleted. Just add a `deprecationReason` like in other places in the schema.

server/graphql/v1/TransactionInterface.js
```diff
@@ -277,40 +276,12 @@ export const TransactionExpenseType = new GraphQLObjectType({
...
-        type: GraphQLString,
-        resolve(transaction, args, req) {
-          // If it's a sequelize model transaction, it means it has the method getExpenseFromViewer
-          // otherwise we return null
-          return transaction.getExpenseForViewer
-            ? transaction.getExpenseForViewer(req.remoteUser).then(expense => expense && expense.privateMessage)
```
znarf: If not deleted, replace `getExpenseForViewer`by the new dataloader.

server/graphql/v1/TransactionInterface.js
```diff
@@ -277,40 +276,12 @@ export const TransactionExpenseType = new GraphQLObjectType({
...
-            : null;
-        },
-      },
-      category: {
-        type: GraphQLString,
-        resolve(transaction, args, req) {
-          // If it's a sequelize model transaction, it means it has the method getExpenseFromViewer
-          // otherwise we return null
-          return transaction.getExpenseForViewer
-            ? transaction.getExpenseForViewer(req.remoteUser).then(expense => expense && expense.category)
```
znarf: If not deleted, replace `getExpenseForViewer`by the new dataloader.

server/graphql/v1/TransactionInterface.js
```diff
@@ -277,40 +276,12 @@ export const TransactionExpenseType = new GraphQLObjectType({
...
-          return req.loaders.expense.findById.load(transaction.ExpenseId);
-        },
-      },
-      attachment: {
-        type: GraphQLString,
-        resolve(transaction, args, req) {
-          // If it's a sequelize model transaction, it means it has the method getExpenseFromViewer
-          // otherwise we return null
-          return transaction.getExpenseForViewer
-            ? transaction.getExpenseForViewer(req.remoteUser).then(expense => expense && expense.attachment)
```
znarf: If not deleted, replace `getExpenseForViewer`by the new dataloader.

EslamHiko:
@znarf Done.

server/graphql/v1/TransactionInterface.js
```diff
@@ -49,7 +49,6 @@ export const TransactionInterfaceType = new GraphQLInterfaceType({
       collective: { type: CollectiveInterfaceType },
       type: { type: GraphQLString },
       description: { type: GraphQLString },
-      privateMessage: { type: GraphQLString },
```
znarf: Why remove `privateMessage` here?

EslamHiko: @znarf I did it because `privateMessage` is loaded either from `Order` or `Expense` it's not loaded from the `Transaction` itself, should I return it back?

znarf: It's all right.

PR #2586, opened 2019-09-23
Custom email templates for Texas Camp

rocketeerbkw:
I've added two email templates that link to our attendee questionnaire in the ticket email.

### DIFF
--- a/server/lib/email.js
+++ b/server/lib/email.js
@@ -243,6 +243,14 @@ const generateEmailFromTemplate = (template, recipient, data = {}, options = {})
   if (template === 'ticket.confirmed') {
     // if (slug === 'sustainoss') template += '.sustainoss';
     if (slug === 'fearlesscitiesbrussels') template += '.fearlesscitiesbrussels';
+
+    if (slug === 'drupalatx') {
+      const eventSlug = get(data, 'event.slug');
+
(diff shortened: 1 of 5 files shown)


server/lib/email.js
```diff
@@ -243,6 +243,14 @@ const generateEmailFromTemplate = (template, recipient, data = {}, options = {})
   if (template === 'ticket.confirmed') {
     // if (slug === 'sustainoss') template += '.sustainoss';
     if (slug === 'fearlesscitiesbrussels') template += '.fearlesscitiesbrussels';
+
+    if (slug === 'drupalatx') {
+      const eventSlug = get(options, 'event.slug');
```
znarf: I haven't tested but seems to me that `event` will never be found in`options` but in `data`.


See:
https://github.com/opencollective/opencollective-api/blob/master/server/lib/notifications.js#L186
https://github.com/opencollective/opencollective-api/blob/master/server/lib/notifications.js#L147

rocketeerbkw: Thanks, I thought I tested locally, but must've used the wrong method. It seems like `data` is the right one. I updated the branch.

znarf:
Merging. It will not be immediately deployed. I will comment on this issue when it will be.

PR #2595, opened 2019-09-23
Cache heavy queries for the new collective page

znarf:
When benchmarking the babel page, I figured this could shave the new collective page from multiple seconds.


znarf:
@Betree `limit` is part of the `cacheKey` here, what precisely do you have in mind when you say we should tweak it?

znarf:
Let's have chat about other details (on demand cache activation/de-activation, ttl, invalidation) and I will make it good to go.

PR #2621, opened 2019-09-29
wip

EvanBurchard:


znarf:
@EvanBurchard what was done in https://github.com/opencollective/opencollective-api/pull/2128 was a good start. Actually, looking at it I just realized that this PR was looking almost complete, never noticed the last 2 commits until now.

PR #2632, opened 2019-09-30
search(collectives) dont index archived collectives, remove recently …

EugeneHlushko:
…archived from index, manual run index to cleanup historical data

- dont index archived collectives
- remove recently archived collectives from index
- manual run index to cleanup historical data
- Closes https://github.com/opencollective/opencollective/issues/2397

### DIFF
--- a/scripts/collective-archived-search-index-cleanup.js
+++ b/scripts/collective-archived-search-index-cleanup.js
@@ -38,7 +38,10 @@ const done = error => {
 const cleanIndex = async () => {
   const collectives = await models.Collective.findAll({
     where: {
-      isArchived: true,
+      isActive: false,
+      deactivatedAt: {
+        [Op.not]: null,
+      },
       type: {
         [Op.or]: [collectiveTypes.COLLECTIVE, collectiveTypes.ORGANIZATION],
       },
--- a/scripts/collective-archived-search-index-cleanup.js
+++ b/scripts/collective-archived-search-index-cleanup.js
@@ -1,14 +1,14 @@
 #!/usr/bin/env node
-import '../../server/env';
+import '../server/env';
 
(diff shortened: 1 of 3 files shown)


scripts/collective-archived-search-index-cleanup.js
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env node
```
EugeneHlushko: Need to run this once to clean up historical data

EugeneHlushko:
Right @znarf , i couldnt confirm this earlier due to posgres install process slowdowns and based it on quick code search. Now it up and running and looks like `isArchived` is only available in graphql v1 as a combination of:

```
    isArchived: {
      description: 'Returns whether this collective is archived',
      type: GraphQLBoolean,
      resolve(collective) {
        return Boolean(collective.deactivatedAt && !collective.isActive);
      },
    },
```

So i've fixed queries and tested them

EugeneHlushko:
Hi @znarf 
I've cleaned up the "manual run" script in the last commit, although i am not sure on the SQL approach. Can you point me to an example of SQL usage in the codebase where i can have a look at to change the script, also you are talking about using SQL for the one-time script in `scripts/collective-archived-search-index-cleanup.js` right? Would love to get feedback in the code in the future vs raw comments, because sometimes its hard to relate the message to specific addition/change in the new codebase.

znarf:
Well it was just about this `yesterday` variable, I wondered what you wanted to do with it.

Now that it's gone, I understand it was useless, so feel free to ignore my comment about SQL, nothing to change.

EugeneHlushko:
@znarf got it! Sorry for late response here i have tidy-ed up the code, can you re-review or point me at what else i can add here?

PR #2637, opened 2019-10-01
Set new nextChargeDate for subscription made after middle of the month.

flickz:
This PR attempts to use https://github.com/opencollective/opencollective/issues/1215#issuecomment-519858300 to fix [#1215](https://github.com/opencollective/opencollective/issues/1215).

### DIFF
--- a/server/lib/subscriptions.js
+++ b/server/lib/subscriptions.js
@@ -200,7 +200,7 @@ export function getNextChargeAndPeriodStartDates(status, order) {
       nextChargeDate.add(1, 'years');
     }
 
-    if (status === 'new' && order.Subscription.interval === intervals.MONTH && dayOfTheMonth >= 15) {
+    if (status === 'new' && order.Subscription.interval === intervals.MONTH && dayOfTheMonth > 15) {
       nextChargeDate.add(1, 'months').startOf('month');
     } else if (status === 'new') {
       nextChargeDate.startOf('month');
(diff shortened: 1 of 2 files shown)


server/lib/subscriptions.js
```diff
@@ -190,16 +191,21 @@ export async function handleRetryStatus(order, transaction) {
...
-    if (order.Subscription.interval === 'month') {
+    if (order.Subscription.interval === intervals.MONTH) {
       nextChargeDate.add(1, 'months');
-    } else if (order.Subscription.interval === 'year') {
+    } else if (order.Subscription.interval === intervals.YEAR) {
       nextChargeDate.add(1, 'years');
     }
-    if (status === 'new') {
+
+    if (status === 'new' && order.Subscription.interval === intervals.MONTH && dayOfTheMonth >= 15) {
```
znarf:
```suggestion
    if (status === 'new' && order.Subscription.interval === intervals.MONTH && dayOfTheMonth > 15) {
```

Would do after the 15.

server/lib/subscriptions.js
```diff
@@ -190,16 +191,21 @@ export async function handleRetryStatus(order, transaction) {
...
-    if (order.Subscription.interval === 'month') {
+    if (order.Subscription.interval === intervals.MONTH) {
       nextChargeDate.add(1, 'months');
-    } else if (order.Subscription.interval === 'year') {
+    } else if (order.Subscription.interval === intervals.YEAR) {
       nextChargeDate.add(1, 'years');
     }
-    if (status === 'new') {
+
+    if (status === 'new' && order.Subscription.interval === intervals.MONTH && dayOfTheMonth > 15) {
```
znarf: This is not really nice to read.

Why not do:

```
if (status === 'new' && order.Subscription.interval === intervals.MONTH && nextChargeDate.date() > 15) {
  nextChargeDate.add(1, 'months')
}
if (status === 'new') {
  nextChargeDate.startOf('month');
}
```

I would also add some code comments to explain why we're doing it.

PR #2638, opened 2019-10-01
feat: added post-merge-install hook

maddhruv:
https://npmjs.com/package/post-merge-install

The tool helps us a lot at ClearTax

It will run `npm install` when there are some changes in `package.json` or `package-lock.json`

Same as - https://github.com/opencollective/opencollective-frontend/pull/2654


znarf:
Cool, let's have a try.

PR #2651, opened 2019-10-02
Allow YouTube and Vimeo iframes in HTML sanitizer

maddhruv:
https://github.com/opencollective/opencollective/issues/2043

Added `youtube` as a trusted source, no clue if `vimeo` or `asciinema` could also be good options in use


Betree:
@maddhruv Did you try using that in a rich description to see if it works?

maddhruv:
I have looked into it, the sanitizer works fine with the configuration in this PR. But the problem is that the `html` value/property for an update contains string formatted in a different way than that of the `summary` which we render using the `HTMLContent` component (Note: study specific to collective updates)

For Instance
summary - `<a href="https://github.com/maddhruv">Github</a>`
while the
html - `<p>&lt;a href=\"https://github.com/maddhruv\"&gt;Github&lt;/a&gt;</p><p>&lt;iframe width=\"560\" height=\"315\" src=\"https://www.youtube.com/embed/mjPPCgz_J4Y\" frameborder=\"0\" allow=\"accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture\" allowfullscreen&gt;&lt;/iframe&gt;</p>`

where the editor treats this HTML as it is but the `a` and `iframe` coming inside are just string (not HTML DOMs)

Betree:
Let's include a screenshot, from what I know YouTube's embed iframe has a few more attriubtes that seems quite important for the video to work properly (see example bellow). Also let's authorize `www.youtube-nocookie.com` (official YouTube endpoint with privacy enhanced mode).

```html
<iframe width="560" height="315" src="https://www.youtube.com/embed/1E7tz0zRlpE" frameborder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
```

maddhruv:
Screenshot 2019-10-04 at 13 14 18
Caption: Status summary as in `HTMLContent`

Screenshot 2019-10-04 at 13 15 11
Caption: Status content (`html`) in div with `dangerouslySetInnerHTML`

> has a few more attriubtes

I only see `width` and `height` as what user might declare for the `view`, and moreover, there is no sense of allowing the various services - `accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture` neither does the frameborder and others

> Also let's authorize www.youtube-nocookie.com

done 👍

maddhruv:
with this fiddle - https://jsfiddle.net/k6gqnmoj/ you can check that just by passing the `src` the iframe/embed works fine except the dimensions of it

Betree:
> I only see width and height as what user might declare for the view, and moreover, there is no sense of allowing the various services - accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture neither does the frameborder and others

- `frameborder` is for not displaying a border around the iframe, I guess it helps with cross-browser compatibility and I don't see any risk in allowing it
- `allowfullscreen` seems pretty important to me, it's what allows for the iframe to go full screen. Which you want for a video.

Regarding `allow`:
- `autoplay` for automatically start the video on page load. I think we should not prevent users from doing that if they want to.
- `encrypted-media` may be required for some DRM-protected videos
- `gyroscope` for rotating video when screen layout changes

So there **is** sense in allowing these various attributes. What are the reasons why we shouldn't do so?

Betree:
> with this fiddle - https://jsfiddle.net/k6gqnmoj/ you can check that just by passing the `src` the iframe/embed works fine except the dimensions of it

It doesn't, I cant go fullscreen without the `allowfullscreen` attribute

image

maddhruv:
> It doesn't, I cant go fullscreen without the allowfullscreen attribute

That I don't know if we want to allow - allow videos to get full screen

Betree:
> > It doesn't, I cant go fullscreen without the allowfullscreen attribute
> 
> That I don't know if we want to allow - allow videos to get full screen

Why wouldn't we?

maddhruv:
> Why wouldn't we?

I don't have an answer for this, but rather a question in perspective of providing limited experiences with 3rd parties (here iframes), I have seen at many websites where they don't allow full screens or views conflicting their designs

Betree:
As long as we control who the third parties are (only YouTube for now) I don't see any downsides or dangers to allow that. But I see downsides in preventing collectives from doing so, because we're limiting their experience for no strong reason.

maddhruv:
Okay great
Have added the same

Betree:
What about the other mentioned in https://github.com/opencollective/opencollective-api/pull/2651#issuecomment-538296369, is there any reason for not allowing them?

maddhruv:
Sorry missed the comment

server/lib/utils.js
```diff
@@ -65,7 +65,21 @@ export function strip_tags(str, allowedTags) {
...
+        'width',
+        'height',
+        {
+          name: 'allow',
+          multiple: true,
+          values: ['autoplay', 'encrypted-media', 'gyroscope'],
+        },
+      ],
     },
+    allowedIframeHostnames: ['www.youtube.com', 'www.youtube-nocookie.com'],
```
Betree: Adding Vimeo while we're at it

```suggestion
    allowedIframeHostnames: ['www.youtube.com', 'www.youtube-nocookie.com', 'player.vimeo.com'],
```

PR #2652, opened 2019-10-02
Enhancement: Host admin can un-pay an expense

flickz:
This PR follows up https://github.com/opencollective/opencollective/issues/2324 to give admin host ability to mark paid expense with other payout method as unpaid. 

- [x] Complete the `markExpenseAsUnpaid` mutation.
- [x] Add unit test

### DIFF
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -382,7 +382,7 @@ export async function markExpenseAsUnpaid(remoteUser, ExpenseId) {
   }
 
   if (expense.payoutMethod !== 'other') {
-    throw new errors.Unauthorized('Only expense with other payout method can be marked unpaid');
+    throw new errors.Unauthorized('Only expenses with "other" payout method can be marked as unpaid');
   }
 
   const transaction = await models.Transaction.findOne({
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -370,7 +370,7 @@ export async function markExpenseAsUnpaid(remoteUser, ExpenseId) {
   });
 
   if (!expense) {
-    throw new errors.NotFound('No expense transaction found');
+    throw new errors.NotFound('No expense found');
   }
 
   if (!canMarkExpenseUnpaid(remoteUser, expense)) {
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -362,7 +362,7 @@ export async function payExpense(remoteUser, expenseId, fees = {}) {
 
 export async function markExpenseAsUnpaid(remoteUser, ExpenseId) {
   if (!remoteUser) {
-    throw new errors.Unauthorized('You need to be logged in to pay an expense');
+    throw new errors.Unauthorized('You need to be logged in to unpay an expense');
   }
 
   const expense = await models.Expense.findByPk(ExpenseId, {
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -374,7 +374,7 @@ export async function markExpenseAsUnpaid(remoteUser, ExpenseId) {
   }
 
   if (!canMarkExpenseUnpaid(remoteUser, expense)) {
-    throw new errors.Unauthorized("You don't have permission to mark this expense unpaid");
+    throw new errors.Unauthorized("You don't have permission to mark this expense as unpaid");
   }
 
   if (expense.status !== statuses.PAID) {
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -378,7 +378,7 @@ export async function markExpenseAsUnpaid(remoteUser, ExpenseId) {
   }
 
   if (expense.status !== statuses.PAID) {
-    throw new errors.Unauthorized('Expense has not been paid');
+    throw new errors.Unauthorized('Expense has not been paid yet');
   }
 
   if (expense.payoutMethod !== 'other') {
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -5,6 +5,7 @@ import statuses from '../../../constants/expense_status';
 import activities from '../../../constants/activities';
 import models from '../../../models';
 import paymentProviders from '../../../paymentProviders';
+import * as libPayments from '../../../lib/payments';
 import { formatCurrency } from '../../../lib/utils';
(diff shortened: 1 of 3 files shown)


server/graphql/v1/mutations/expenses.js
```diff
@@ -348,3 +359,38 @@ export async function payExpense(remoteUser, expenseId, fees = {}) {
...
+
+  if (expense.payoutMethod !== 'other') {
+    throw new errors.Unauthorized('Only expense with other payout method can be marked unpaid');
+  }
+
+  const transaction = await models.Transaction.findOne({
+    where: { ExpenseId },
+    include: [{ model: models.Expense }],
+  });
+  const refundedTransaction = await libPayments.createRefundTransaction(transaction, 0, null, expense.User);
```
Betree: I'm not sure about the `0`. When hosts mark expenses as paid, they can set a `paymentProcessorFeeInHostCurrency`. If the initial payment processor fee is not 0, then this will take the amount out  of the host budget. But that's tricky.

@alanna When un-paying expenses with `Other` payout method, should we assume that:
1. Payment processor fees are lost in the payout, and we should take them out of the host's budget
2. OR that fees are reimbursed by whatever the payout method and it's fine to mark them as such

My assumption will be that 2. is the valid one, because the reason for marking expenses as unpaid is usually that the payment has failed and no fees is generally taken in those cases. 

But maybe host should be able to set this when marking as unpaid (ex: Click on the button -> Checkbox: Has the payout provider refunded the payment processor fees? -> Yes/No).

server/graphql/v1/mutations/expenses.js
```diff
@@ -28,6 +29,16 @@ function canUpdateExpenseStatus(remoteUser, expense) {
   return false;
 }
 
+/**
+ * Only admin of expense.collective.host can mark expenses unpaid
+ */
+function canMarkExpenseUnpaid(remoteUser, expense) {
+  if (remoteUser.hasRole([roles.HOST, roles.ADMIN], expense.collective.HostCollectiveId)) {
```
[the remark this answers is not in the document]
flickz: Has both HOST and ADIM role.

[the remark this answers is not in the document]
Betree: But why would a user have the role `HOST` on the host?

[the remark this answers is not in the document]
Betree: Host has the `HOST` role for collectives it hosts, but a host itself has no host

alanna:
I think a checkbox is the best option ("Also refund payment processor fees?") because both scenarios are possible.

server/graphql/v1/mutations/expenses.js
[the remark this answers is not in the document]
znarf: Agree with @Betree, doesn't make sense.

I had a small doubt and looked at 
https://github.com/opencollective/opencollective-api/blob/master/server/models/User.js#L309

But when admin of the host, the user is granted ADMIN role not the HOST role.

If it appears elsewhere, let's remove it. No cargo cult code :-)

server/graphql/v1/mutations/expenses.js
znarf: Let's wait on @alanna's feedback.

server/graphql/v1/mutations/expenses.js
```diff
@@ -348,3 +359,38 @@ export async function payExpense(remoteUser, expenseId, fees = {}) {
...
+  if (!canMarkExpenseUnpaid(remoteUser, expense)) {
+    throw new errors.Unauthorized("You don't have permission to mark this expense unpaid");
+  }
+
+  if (expense.status !== statuses.PAID) {
+    throw new errors.Unauthorized('Expense has not been paid');
+  }
+
+  if (expense.payoutMethod !== 'other') {
+    throw new errors.Unauthorized('Only expense with other payout method can be marked unpaid');
```
znarf:
```suggestion
    throw new errors.Unauthorized('Only expenses with "other" payout method can be marked as unpaid');
```

server/graphql/v1/mutations/expenses.js
```diff
@@ -348,3 +359,38 @@ export async function payExpense(remoteUser, expenseId, fees = {}) {
 
   return markExpenseAsPaid(expense);
 }
+
+export async function markExpenseAsUnpaid(remoteUser, ExpenseId) {
+  if (!remoteUser) {
+    throw new errors.Unauthorized('You need to be logged in to pay an expense');
```
znarf:
```suggestion
    throw new errors.Unauthorized('You need to be logged in to unpay an expense');
```

server/graphql/v1/mutations/expenses.js
```diff
@@ -348,3 +359,38 @@ export async function payExpense(remoteUser, expenseId, fees = {}) {
...
+  }
+
+  const expense = await models.Expense.findByPk(ExpenseId, {
+    include: [{ model: models.Collective, as: 'collective' }, { model: models.User, as: 'User' }],
+  });
+
+  if (!expense) {
+    throw new errors.NotFound('No expense transaction found');
```
znarf:
```suggestion
    throw new errors.NotFound('No expense found');
```

server/graphql/v1/mutations/expenses.js
```diff
@@ -348,3 +359,38 @@ export async function payExpense(remoteUser, expenseId, fees = {}) {
...
+  }
+
+  if (!canMarkExpenseUnpaid(remoteUser, expense)) {
+    throw new errors.Unauthorized("You don't have permission to mark this expense unpaid");
```
znarf:
```suggestion
    throw new errors.Unauthorized("You don't have permission to mark this expense as unpaid");
```

server/graphql/v1/mutations/expenses.js
```diff
@@ -348,3 +359,38 @@ export async function payExpense(remoteUser, expenseId, fees = {}) {
...
+  }
+
+  if (expense.status !== statuses.PAID) {
+    throw new errors.Unauthorized('Expense has not been paid');
```
znarf:
```suggestion
    throw new errors.Unauthorized('Expense has not been paid yet');
```

server/graphql/v1/mutations/expenses.js
flickz: @Betree What do you think?

Screenshot 2019-10-06 at 9 19 57 AM

Betree: That'd be perfect. I would have it checked by default (based on my assumption above).

Let's add the param for it to the query and based on that, pass either `0` or `transaction.paymentProcessorFeeInHostCurrency` based on that.

PR #2656, opened 2019-10-03
feat(refund): allow collective admins to issue refunds

xdamman:


server/graphql/v1/mutations/orders.js
```diff
@@ -845,7 +845,7 @@ export async function refundTransaction(_, args, req) {
   //   b. Host Collective receiving the donation
   //   c. Site Admin
```
znarf: @xdamman can you perhaps change these comments which are now outdated?

xdamman: Ah sorry, only seeing your comment after merging. Shall I do another PR or just push another commit for that small change in master?

xdamman: New PR: https://github.com/opencollective/opencollective-api/pull/2662

Betree:
Would have been great to make a security review of the refund process before merging that.

PR #2657, opened 2019-10-03
test(graphql.collective.test.js): fix Fragile API Test

NotJohs:
fix fragile test resulting from unstable array element position

Fix [#2386](https://github.com/opencollective/opencollective/issues/2386)

### DIFF
--- a/test/graphql.collective.test.js
+++ b/test/graphql.collective.test.js
@@ -802,9 +802,10 @@ describe('graphql.collective.test.js', () => {
       expect(res.errors).to.not.exist;
       const members = res.data.editCollective.members;
       expect(members.length).to.equal(4);
-      expect(members[2].role).to.equal('MEMBER');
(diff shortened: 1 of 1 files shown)


test/graphql.collective.test.js
```diff
@@ -802,9 +802,10 @@ describe('graphql.collective.test.js', () => {
       expect(res.errors).to.not.exist;
       const members = res.data.editCollective.members;
       expect(members.length).to.equal(4);
-      expect(members[2].role).to.equal('MEMBER');
-      expect(members[2].member.name).to.equal('member1');
-      expect(members[2].member.email).to.equal('member1@hail.com');
+      const member1 = members.filter(m => m.member.name === 'member1');
```
znarf: Just a matter of style, instead of using `filter` I would use `find` that is returning the first found element. Then instead of using `member1[0]` you can just use `member1`.

PR #2660, opened 2019-10-03
show paypal email only when expense.payoutMethod is set to 'paypal'

EslamHiko:
This pr solves: https://github.com/opencollective/opencollective/issues/1912


znarf:

PR #2661, opened 2019-10-03
sort expenses based on createdAt

EslamHiko:
This Pr is to solve: https://github.com/opencollective/opencollective/issues/2351

Before : 

before

After : 

after


znarf:

PR #2662, opened 2019-10-03
chore(refund): fix comments

xdamman:


server/graphql/v1/mutations/orders.js
```diff
@@ -842,8 +842,9 @@ export async function refundTransaction(_, args, req) {
 
   // 1. Verify user permission. User must be either
   //   a. User that created transaction (within 24h) -- Not implemented yet
```
znarf: Let's remove A because it's been here for quite some time and we're not implementing it.

server/graphql/v1/mutations/orders.js
```diff
@@ -842,8 +842,9 @@ export async function refundTransaction(_, args, req) {
...
-  //   b. Host Collective receiving the donation
-  //   c. Site Admin
+  //   b. Admin of the collective that received the donation
+  //   b. Admin of the Host Collective that received the donation
```
znarf: Let's make sure there is not 2 times B

PR #2665, opened 2019-10-03
flush cloudflare cache of collective's page after refunding

EslamHiko:
this Pr solves: https://github.com/opencollective/opencollective/issues/2354


znarf:

PR #2667, opened 2019-10-03
prevent guests from creating orders

EslamHiko:
This Pr is to solve: https://github.com/opencollective/opencollective/issues/1616


znarf:
- Can you rebase and push the result?

znarf:
Triggered CI

EslamHiko:
@znarf I don't think the failure is related to my PR, it fails in this test : https://github.com/opencollective/opencollective-api/blob/master/test/graphql.collective.test.js#L753

also, it's failing in https://github.com/opencollective/opencollective-api/pull/2671 for the same reason.

znarf:
@EslamHiko could you rebase to see if tests are now passing?

PR #2671, opened 2019-10-03
Add 'Reply-To' Header and replyTo as an option to notifyAdminsOfCollective

EslamHiko:
This PR is to solve: https://github.com/opencollective/opencollective/issues/2332

It needs testing.


Betree:
Can't we use `hello@${collective.slug}.opencollective.com` so that all collective admins will be notified?

EslamHiko:
lint fails because of 
```
require-atomic-updates
  290:7  error    Unexpected lexical declaration in case block                                                                                                    no-case-declarations
  291:7  error    Unexpected lexical declaration in case block 
```
any ideas to fix the lint ?

Betree:
@EslamHiko You can see the reason for this as well as details on how to fix it on https://eslint.org/docs/rules/no-case-declarations

znarf:
Triggering CI

znarf:
> Can't we use hello@${collective.slug}.opencollective.com so that all collective admins will be notified?

@Betree sticking to what was asked in the issue for now, but why not? @alanna what do you think?

znarf:
@EslamHiko could you rebase to see if tests are now passing?

server/lib/notifications.js
```diff
@@ -286,15 +286,21 @@ async function notifyByEmail(activity) {
       notifyAdminsOfCollective(activity.data.collective.id, activity);
       break;
 
-    case activityType.COLLECTIVE_APPLY:
+    case activityType.COLLECTIVE_APPLY: {
+      const user = await models.User.findByPk(activity.data.user.id);
```
Betree: The place where the activity is created doesn't seems to include the user id in the payload. But it does include the user email, so we can probably skip the DB query: https://github.com/opencollective/opencollective-api/blob/fde4e0b20e7d53dab447f2fa7f166c180ced57ea/server/models/Collective.js#L1510-L1548

EslamHiko: Done 👍

server/lib/notifications.js
```diff
@@ -286,15 +286,17 @@ async function notifyByEmail(activity) {
       notifyAdminsOfCollective(activity.data.collective.id, activity);
       break;
 
-    case activityType.COLLECTIVE_APPLY:
+    case activityType.COLLECTIVE_APPLY: {
+
```
znarf:
```suggestion
```

prettier is not happy with this empty line

PR #2672, opened 2019-10-04
[feat]: disconnect account api

maddhruv:
Solves **1** and **2** of **Disconnecting** at https://github.com/opencollective/opencollective/issues/1845

which are

- [x] - It's currently impossible to disconnect a Twitter account
- [x] - It's currently impossible to disconnect a GitHub account

#### Includes
- `Disconnect` api endpoint with `DELETE` method
- `delete` method on ConnectedAccount model - which delete the given/attached account

Related with - https://github.com/opencollective/opencollective-frontend/pull/2738

### DIFF
--- a/server/controllers/connectedAccounts.js
+++ b/server/controllers/connectedAccounts.js
@@ -6,6 +6,7 @@ import models, { Op } from '../models';
 import errors from '../lib/errors';
 import paymentProviders from '../paymentProviders';
 import * as github from '../lib/github';
+import { mustBeLoggedInTo } from '../lib/auth';
 
 const { ConnectedAccount, User } = models;
 
@@ -136,18 +137,30 @@ export const disconnect = async (req, res) => {
   const { remoteUser } = req;
 
   try {
+    mustBeLoggedInTo(remoteUser, 'disconnect this connected account');
+
+    if (!remoteUser.isAdmin(CollectiveId)) {
+      throw new errors.Unauthorized({
+        message: 'You are either logged out or not authorized to disconnect this account',
+      });
+    }
+
     const account = await ConnectedAccount.findOne({
       where: { service, CollectiveId },
     });
 
-    await account.delete(remoteUser);
+    await account.delete();
 
     res.send({
       deleted: true,
       service,
     });
   } catch (err) {
-    res.send(err);
+    res.send({
+      error: {
+        message: err.message,
+      },
+    });
   }
 };
 
--- a/server/controllers/connectedAccounts.js
+++ b/server/controllers/connectedAccounts.js
@@ -149,7 +149,9 @@ export const disconnect = async (req, res) => {
       where: { service, CollectiveId },
     });
 
-    await account.delete();
+    if (account) {
+      await account.destroy();
+    }
 
     res.send({
       deleted: true,
--- a/server/controllers/connectedAccounts.js
+++ b/server/controllers/connectedAccounts.js
@@ -131,6 +131,26 @@ export const createOrUpdate = (req, res, next, accessToken, data, emails) => {
   }
 };
 
+export const disconnect = async (req, res) => {
+  const { collectiveId: CollectiveId, service } = req.params;
+  const { remoteUser } = req;
+
+  try {
+    const Account = await ConnectedAccount.findOne({
+      where: { service, CollectiveId },
+    });
+
+    await Account.delete(remoteUser);
+
+    res.send({
+      deleted: true,
+      service,
+    });
+  } catch (err) {
+    res.send(err);
+  }
+};
+
 export const verify = (req, res, next) => {
   const payload = req.jwtPayload;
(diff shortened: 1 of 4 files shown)


maddhruv:
> we now try to do everything in GraphQL

I totally agree with this, but seeing the already existing connect accounts auth api as REST, I thought there might be some reason using REST for those (may be somewhere else those are being implemented as REST) so thought of creating a similar REST Endpoint

server/models/ConnectedAccount.js
```diff
@@ -72,5 +75,15 @@ export default (Sequelize, DataTypes) => {
...
+  ConnectedAccount.prototype.delete = async function(remoteUser) {
+    mustBeLoggedInTo(remoteUser, 'disconnect this connected account');
+    if (remoteUser.id !== this.CreatedByUserId || !remoteUser.isAdmin(this.CollectiveId)) {
+      throw new errors.Unauthorized({
+        message: 'You are either logged out or not authorized to disconnect this account',
+      });
+    }
+    return this.destroy();
+  };
+
```
Betree: It seems to me that checking authentication/permissions and generating GraphQL errors is not the role of the model (whose role should be to solely map the data model to the database). Though I don't know this part very well, this code could be problematic for example in the Forest admin interface where we may call `destroy` as an admin action without providing a `remoteUser`.

I'll add that it makes the API for it pretty unclear, when reading at this line above it's hard to tell why we provide the remoteUser to the function at first glance.

```es6
await Account.delete(remoteUser);
```

Note that this pattern has been followed in other parts of the code in the past so it's not a big issue but as a design principle I'd be in favor of moving the permission check in the controllers/graphql endpoints for any new development (in this case `server/controllers/connectedAccounts.js`)

server/controllers/connectedAccounts.js
```diff
@@ -131,6 +131,26 @@ export const createOrUpdate = (req, res, next, accessToken, data, emails) => {
   }
 };
 
+export const disconnect = async (req, res) => {
+  const { collectiveId: CollectiveId, service } = req.params;
+  const { remoteUser } = req;
+
+  try {
+    const Account = await ConnectedAccount.findOne({
```
Betree: Let's default to camel case for variables

```suggestion
    const account = await ConnectedAccount.findOne({
```

server/controllers/connectedAccounts.js
```diff
@@ -131,6 +131,26 @@ export const createOrUpdate = (req, res, next, accessToken, data, emails) => {
...
+      where: { service, CollectiveId },
+    });
+
+    await Account.delete(remoteUser);
```
Betree: Let's default to camel case for variables

```suggestion
    await account.delete(remoteUser);
```

maddhruv: resolved with - https://github.com/opencollective/opencollective-api/pull/2672/commits/ee8a906d7dc974721f519c49107488d3d31990d5

server/controllers/connectedAccounts.js
maddhruv: resolved with https://github.com/opencollective/opencollective-api/pull/2672/commits/ee8a906d7dc974721f519c49107488d3d31990d5

server/models/ConnectedAccount.js
maddhruv: Please review again, I have refactored the authorization part.

server/controllers/connectedAccounts.js
```diff
@@ -131,6 +132,38 @@ export const createOrUpdate = (req, res, next, accessToken, data, emails) => {
...
+      throw new errors.Unauthorized({
+        message: 'You are either logged out or not authorized to disconnect this account',
+      });
+    }
+
+    const account = await ConnectedAccount.findOne({
+      where: { service, CollectiveId },
+    });
+
+    await account.delete();
```
Betree: Let's be safe

```suggestion
    if (account) {
        await account.destroy();
    }
```

server/models/ConnectedAccount.js
```diff
@@ -72,5 +72,9 @@ export default (Sequelize, DataTypes) => {
     });
   };
 
+  ConnectedAccount.prototype.delete = async function() {
+    return this.destroy();
+  };
+
```
Betree: No need to add more code, we can use the `destroy` function directly

server/controllers/connectedAccounts.js
maddhruv: resolved with https://github.com/opencollective/opencollective-api/pull/2672/commits/6e3ad30b03e212430eb30a015eb716703ec39d96

server/models/ConnectedAccount.js
maddhruv: resolved with https://github.com/opencollective/opencollective-api/pull/2672/commits/6e3ad30b03e212430eb30a015eb716703ec39d96

PR #2683, opened 2019-10-05
Fix/delete currency fxrate

NotJohs:
re [#1662](https://github.com/opencollective/opencollective/issues/1662)


server/lib/activities.js
```diff
@@ -227,8 +227,11 @@ export default {
         switch (activity.data.transaction.type) {
           case TransactionTypes.CREDIT:
             if (userTwitter) {
+              // let fxRate = ''
+              // currencylib.convertToCurrency(1, 'USD', currency)
+              //   .then(amount => fxRate += amount);
```
Betree: Let's not keep commented code

server/lib/currency.js
```diff
@@ -18,6 +18,21 @@ function getDate(date = 'latest') {
...
+    CAD: value => `${value} CAD`,
+    EUR: value => `€${value}`,
+    GBP: value => `£${value}`,
+    INR: value => `₹${value}`,
+    MXN: value => `${value} MXN`,
+    SEK: value => `kr ${value}`,
+    USD: value => `$${value}`,
+    UYU: value => `$U ${value}`,
+  };
+  return currencyFormats[currency];
```
Betree: Let's take this as an opportunity to improve what we have. Currently if we use this function with an unknown currency it will crash. We can fix it easily with the following:

```suggestion
  return currencyFormats[currency] || value => `${value} ${currency}`;
```

server/lib/queries.js
```diff
@@ -20,17 +20,10 @@ const generateFXConversionSQL = aggregate => {
...
-
-  let sql = 'CASE ';
-  sql += fxConversion
-    .map(currency => `WHEN ${currencyColumn} = '${currency[0]}' THEN ${amountColumn} / ${currency[1]}`)
-    .join('\n');
-  sql += 'ELSE 0 END';
-
+  let sql = '';
+  convertToCurrency(1, 'USD', currencyColumn).then(
+    amount => (sql += currencyColumn ? `${amountColumn} / ${amount}` : '0'),
```
Betree: Because this call is asynchronous, the function will always return an empty string.

server/lib/activities.js
NotJohs: Understood

server/lib/currency.js
NotJohs: Will implement

server/lib/queries.js
NotJohs: Understood. Couldn't properly implement this.  Will Fix

NotJohs:
@Betree Reviewed.

server/lib/queries.js
```diff
@@ -114,13 +104,13 @@ const getTotalAnnualBudgetForHost = HostCollectiveId => {
...
 
-const getTotalAnnualBudget = () => {
+const getTotalAnnualBudget = async () => {
   return sequelize
     .query(
       `
   SELECT
     (SELECT
-      COALESCE(SUM(${generateFXConversionSQL()} * 12),0)
+      COALESCE(SUM(${await generateFXConversionSQL()} * 12),0)
```
znarf: Can you move the await on top, put the value in a variable and use the variable here instead?

NotJohs: Understood.

server/lib/activities.js
```diff
@@ -262,7 +262,7 @@ export default {
       case activities.SUBSCRIPTION_CONFIRMED:
         if (userTwitter) {
           tweet = encodeURIComponent(
-            `@${userTwitter} thanks for your ${currencies[currency].format(recurringAmount)} donation to ${
+            `@${userTwitter} thanks for your ${formatCurrency(currency, recurringAmount)} donation to ${
```
Betree: Usage is much cleaner like that, thanks!

server/lib/currency.js
```diff
@@ -18,6 +18,19 @@ function getDate(date = 'latest') {
...
+export function formatCurrency(currency, value) {
+  const _currency = currency.toUpperCase();
+  const currencyFormats = {
+    EUR: `€${value}`,
+    GBP: `£${value}`,
+    INR: `₹${value}`,
+    SEK: `kr ${value}`,
+    USD: `$${value}`,
+    UYU: `$U ${value}`,
+  };
```
Betree: Generating all the possible strings to return only one seems overkill, it was good to have functions there and to only call the one that you're going to use.

server/lib/queries.js
```diff
@@ -20,18 +20,8 @@ const generateFXConversionSQL = aggregate => {
...
-  }
-
-  let sql = 'CASE ';
-  sql += fxConversion
-    .map(currency => `WHEN ${currencyColumn} = '${currency[0]}' THEN ${amountColumn} / ${currency[1]}`)
-    .join('\n');
-  sql += 'ELSE 0 END';
-
-  return sql;
+  const amount = await convertToCurrency(1, 'USD', currencyColumn);
```
Betree: I don't understand how this is going to work, from my understanding it is going to call `getFxRate` like this:

> getFxRate(1, 'USD', 't.currency')

Which I guess will return an error?

Am I missing something?

NotJohs: Screenshot 2019-11-02 at 6 16 04 AM

the amount is passed to the callback

znarf: That piece of code `generateFXConversionSQL` is absolutely wrong. The concern from @Betree was 100% valid.

NotJohs: Reviewing this.

NotJohs: @znarf I understand where the bugs are in the code.

PR #2685, opened 2019-10-06
Change how pledge historical information is stored

Betree:
With https://github.com/opencollective/opencollective-frontend/pull/2760 we now rely on `isPledged` to know if the collective is **currently** pledged. With the current system, there's a risk that we show the pledge page if a collective that has been pledged in the past is archived (because we check `isPledged && !isActive`).

This PR:
- Sets the `isPledged` flag to false when claiming the collective
- Sets a new flag instead: `data.hasBeenPledged` so that we keep this information
- Limits the number of allowed fields when creating a collective for a pledge (because the creator of the collective is not an admin in this case, we want to be more restrictive)

### DIFF
--- a/cron/daily/60-clean-orders.js
+++ b/cron/daily/60-clean-orders.js
@@ -18,10 +18,18 @@ Promise.all([
   AND "Collectives"."id" = "Orders"."CollectiveId"
   AND "Collectives"."isPledged" = FALSE
   AND "Collectives"."HostCollectiveId" IS NOT NULL
-  AND "Orders"."createdAt" <  (NOW() - interval '2 month')`,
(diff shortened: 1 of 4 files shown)


Betree:
As reported by @znarf, we need to make some changes to ensure that we don't delete all the pledges in `cron/daily/clean-orders.js` when the collective is claimed.

Betree:
Added https://github.com/opencollective/opencollective-api/pull/2685/commits/e033ba46c6c36bf1178f819cc68845dfdfa3f086 to make sure we don't delete pledges in the new system with `clean-orders`.

Betree:
@znarf this is now ready for review

PR #2703, opened 2019-10-08
feat(one-use-login-link): Changes login links to single use

sbinlondon:
Addresses opencollective/opencollective#2192

single-use-login

This creates single-use login links by adding a field to the first JWT token that is processed when a user signs in.

Hopefully I've updated all the relevant database and GraphQL bits, but let me know if I've missed anything.

Feedback and questions welcome 👍

### DIFF
--- a/.forestadmin-schema.json
+++ b/.forestadmin-schema.json
@@ -5288,7 +5288,7 @@
           "validations": []
         },
         {
-          "field": "lastLogin",
+          "field": "lastLoginAt",
           "type": "Date",
           "defaultValue": null,
           "enums": null,
--- a/.forestadmin-schema.json
+++ b/.forestadmin-schema.json
@@ -5288,6 +5288,21 @@
           "validations": []
         },
         {
+          "field": "lastLogin",
+          "type": "Date",
+          "defaultValue": null,
+          "enums": null,
+          "integration": null,
+          "isFilterable": true,
+          "isReadOnly": false,
+          "isRequired": false,
+          "isSortable": true,
+          "isVirtual": false,
+          "reference": null,
+          "inverseOf": null,
+          "validations": []
+        },
+        {
           "field": "transactions",
           "type": ["Number"],
           "defaultValue": null,
--- a/migrations/05-user-model.js
+++ b/migrations/05-user-model.js
@@ -60,9 +60,6 @@ module.exports = {
         defaultValue: DataTypes.NOW,
       },
       seenAt: DataTypes.DATE,
-      lastLogin: {
-        type: DataTypes.DATE,
-      },
       deletedAt: {
         type: DataTypes.DATE,
       },
--- a/migrations/05-user-model.js
+++ b/migrations/05-user-model.js
@@ -60,6 +60,9 @@ module.exports = {
         defaultValue: DataTypes.NOW,
       },
       seenAt: DataTypes.DATE,
+      lastLogin: {
+        type: DataTypes.DATE,
+      },
       deletedAt: {
         type: DataTypes.DATE,
       },
--- a/migrations/20191007173134-add-lastlogin-column-to-users.js
+++ b/migrations/20191007173134-add-lastlogin-column-to-users.js
@@ -2,12 +2,12 @@
 
 module.exports = {
   up: function(queryInterface, DataTypes) {
-    return queryInterface.addColumn('Users', 'lastLogin', {
+    return queryInterface.addColumn('Users', 'lastLoginAt', {
       type: DataTypes.DATE,
     });
   },
 
   down: function(queryInterface) {
-    return queryInterface.removeColumn('Users', 'lastLogin');
+    return queryInterface.removeColumn('Users', 'lastLoginAt');
   },
 };
--- a/migrations/20191007173134-add-lastlogin-column-to-users.js
+++ b/migrations/20191007173134-add-lastlogin-column-to-users.js
@@ -1,0 +1,13 @@
+'use strict';
+
+module.exports = {
+  up: function(queryInterface, DataTypes) {
+    return queryInterface.addColumn('Users', 'lastLogin', {
+      type: DataTypes.DATE,
+    });
+  },
+
+  down: function(queryInterface) {
+    return queryInterface.removeColumn('Users', 'lastLogin');
+  },
+};
--- a/server/middleware/security/authentication.js
+++ b/server/middleware/security/authentication.js
@@ -84,9 +84,24 @@ export const checkJwtExpiry = (req, res, next) => {
 export const _authenticateUserByJwt = (req, res, next) => {
   if (!req.jwtPayload) return next();
   const userid = Number(req.jwtPayload.sub);
+  const lastLogin = req.jwtPayload.last;
   User.findByPk(userid)
     .then(user => {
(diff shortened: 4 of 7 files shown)


sbinlondon:
Can't seem to request a review (since I don't have write access) so tagging @znarf 😄

migrations/20191007173134-add-lastlogin-column-to-users.js
```diff
@@ -0,0 +1,13 @@
...
+  up: function(queryInterface, DataTypes) {
+    return queryInterface.addColumn('Users', 'lastLogin', {
+      type: DataTypes.DATE,
+    });
+  },
+
+  down: function(queryInterface) {
+    return queryInterface.removeColumn('Users', 'lastLogin');
+  },
+};
```
znarf: Have you considered naming the column `lastLoginAt`?

That would match the convention we have for Date columns.

.forestadmin-schema.json
```diff
@@ -5287,6 +5287,21 @@
...
+          "integration": null,
+          "isFilterable": true,
+          "isReadOnly": false,
+          "isRequired": false,
+          "isSortable": true,
+          "isVirtual": false,
+          "reference": null,
+          "inverseOf": null,
+          "validations": []
+        },
```
znarf: I'm curious how you got that modified? Was it done automatically?

This file is a bit new for us, and we're still trying to figure the best practices when modifying it.

server/middleware/security/authentication.js
```diff
@@ -84,9 +84,24 @@ export const checkJwtExpiry = (req, res, next) => {
 export const _authenticateUserByJwt = (req, res, next) => {
   if (!req.jwtPayload) return next();
   const userid = Number(req.jwtPayload.sub);
+  const lastLogin = req.jwtPayload.last;
```
znarf: You set `lastLogin` here but are still using `req.jwtPayload.last` later. Maybe you should try to always use `lastLogin` once it's set?

server/middleware/security/authentication.js
```diff
@@ -84,9 +84,24 @@ export const checkJwtExpiry = (req, res, next) => {
...
+        } else {
+          const jwtDate = new Date(lastLogin);
+          const dbLogin = user.lastLogin;
+          if (jwtDate.getTime() !== dbLogin.getTime()) {
+            throw errors.Unauthorized('Invalid login link - multiple uses');
+          }
+          const now = new Date();
+          user.update({ lastLogin: now });
+        }
+      }
```
znarf: Some code comments would help understand this section. I also think it there is some potential to make it easier to read.

migrations/05-user-model.js
```diff
@@ -60,6 +60,9 @@ module.exports = {
         defaultValue: DataTypes.NOW,
       },
       seenAt: DataTypes.DATE,
+      lastLogin: {
+        type: DataTypes.DATE,
+      },
```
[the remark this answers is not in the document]
sbinlondon: Ah, good point - sorry about that! I'll be sure to remove it.

migrations/20191007173134-add-lastlogin-column-to-users.js
sbinlondon: That makes more sense. I'll be sure to update that.

.forestadmin-schema.json
sbinlondon: I did it myself. I was trying to figure out how to make sure my new `lastLogin` column was included everywhere it ought to be and thought that this needed to be modified as well.

server/middleware/security/authentication.js
sbinlondon: Agreed! I thought for some reason that perhaps we wouldn't want to be obvious about the naming of the last login timestamp in a JWT, since they can be decoded, but that's probably just me overthinking 😅

.forestadmin-schema.json
znarf: No Forest API keys were required for that?

server/middleware/security/authentication.js
znarf: For example, one explanation before the `if` and `else` conditions could be interesting.

sbinlondon:
I've updated the PR based on your comments @znarf. I hope the comments in the code makes it better to read; I've also renamed some of the variables to hopefully make things more clear.

server/middleware/security/authentication.js
znarf: It's now already super clear but I'm wondering if this could be simplified.

I see that in both cases we do the following:

```
const now = new Date();
user.update({ lastLoginAt: now });
```

Maybe we could have that only once by moving it out of the conditions. I can imagine this should then allow some further simplification.

sbinlondon: I see! Really good point. I'll update that now 😊

.forestadmin-schema.json
sbinlondon: No - i just edited the file. It didn't fuss or anything 🤔

znarf: Ok, I get it!

znarf: I introduced yesterday the first version of a tool to update that file. https://github.com/opencollective/opencollective-api/pull/2694 It's not documented yet and this is why I was interested.

server/middleware/security/authentication.js
sbinlondon: Updated the code. Let me know when you want to chat through it or if you think it needs some more changes!

znarf:
Looks good for me, asking a second review from @Betree

server/middleware/security/authentication.js
```diff
@@ -84,9 +84,33 @@ export const checkJwtExpiry = (req, res, next) => {
...
+         * conditions are met. For normal logins, we check that the lastLoginAt
+         * in the JWT matches the lastLoginAt in the db. If so, we allow the user
+         * to log in, and update the lastLoginAt.
+         */
+        if (req.jwtPayload.lastLoginAt !== 'firstLogin' || user.lastLoginAt) {
+          const lastLoginInJwt = new Date(lastLoginAt);
+          const lastLoginInDb = user.lastLoginAt;
+          if (lastLoginInJwt.getTime() !== lastLoginInDb.getTime()) {
+            throw errors.Unauthorized('Invalid login link - multiple uses');
+          }
```
Betree: Can't `user.lastLoginAt` be `null`? (it would crash in the condition below)

sbinlondon: You're right, when the db initialises, then `lastLoginAt` will be null. But it shouldn't crash in the `if` statement because of the code that decides whether you go into the `if` statement or not (line 104).

If it's the user's first login, i.e. they just signed up, then line 104 takes them around the `if` statement and straight into the bit where we overwrite the `lastLoginAt` in the db with the first login time. (This is because both of the conditions are falsy.)

So in all other cases, `user.lastloginAt` should actually be a timestamp, meaning we can: 1) prevent re-use of the first login token if `req.jwtPayload.lastLoginAt !== 'firstLogin'` evaluates to true; and 2) takes us into the `if` statement if it isn't the user's first login, allowing us to check that the `lastLoginAt` in the JWT matches that in the db.

Betree: Alright, thanks for the details

PR #2705, opened 2019-10-08
Block applications if host is not open to them

Betree:
Resolve https://opencollective.freshdesk.com/a/tickets/2139
Related to https://github.com/opencollective/opencollective/issues/2165

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -204,7 +204,7 @@ export async function createCollectiveFromGithub(_, args, req) {
     const host = await models.Collective.findByPk(defaultHostCollective('opensource').CollectiveId);
     const promises = [
       collective.addUserWithRole(user, roles.ADMIN),
-      collective.addHost(host, user),
+      collective.addHost(host, user, { shouldAutomaticallyApprove: true }),
       collective.update({ isActive: true, approvedAt: new Date() }),
     ];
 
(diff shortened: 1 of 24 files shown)


server/models/Collective.js
```diff
@@ -1433,25 +1441,27 @@ export default function(Sequelize, DataTypes) {
       CollectiveId: this.id,
     };
 
-    let isActive = false;
-    let approvedAt = null;
-    if (creatorUser.isAdmin) {
```
Betree: I checked the code and found no valid reason to verify that this method exist, we always pass a valid user. It was introduced in https://github.com/opencollective/opencollective-api/commit/91d1f1974d605449a8383f4ae66af276e1038b40 to fix the tests, because we don't pass a `user` in the tests.

I'm gonna fix the tests, because adding this kind of checks is not a good practice and can be dangerous.

server/graphql/v1/mutations/collectives.js
```diff
@@ -590,10 +590,7 @@ export async function claimCollective(_, args, req) {
   // create default tiers
   const host = await models.Collective.findByPk(defaultHostCollective('opensource').CollectiveId);
 
-  collective = await collective.addHost(host, {
-    ...req.remoteUser.minimal,
-    isAdmin: () => true,
```
Betree: I also removed this pattern to instead introduce a `shouldAutomaticallyApprove` boolean in `options`

server/lib/queries.js
```diff
@@ -34,7 +34,7 @@ const generateFXConversionSQL = aggregate => {
   return sql;
 };
 
-const getPublicHostsByTotalCollectives = args => {
+const getPublicHostsByTotalCollectives = async args => {
```
Betree: I changed this query because it was not mapping results to models, thus the `collective.canApply` call was failing in graphql enpoint. It will alto fix the `total` number returned, which was invalid.

PR #2723, opened 2019-10-10
Fix: non-descriptive error on add funds form

flickz:
This PR fix https://github.com/opencollective/opencollective/issues/2488

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -975,7 +975,7 @@ export async function addFundsToCollective(order, remoteUser) {
     user = remoteUser;
   }
 
-  if (order.fromCollective.id) {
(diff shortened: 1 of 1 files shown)


flickz:
`error: GraphQL v1 error: We couldn't generate a unique slug for this collective`

I was about to confirm, what are the data required  to addFundToCollective, so we can enforce that if not provided before it's attempt to create Organization with empty data.

Betree:
`Cannot read property id of undefined` is bad, but `We couldn't generate a unique slug for this collective` is not very informative neither about what  the user should do next to solve it.

In opencollective/opencollective#2488, the use case asked by Alanna (entering just a name and an email) seems valid to me. We should support that.

flickz:
Closing this in favour of https://github.com/opencollective/opencollective-api/pull/2724

PR #2725, opened 2019-10-10
chore: remove npx for maildev

maddhruv:
### Description
As `maildev` is not installed as a local dependency for this node app, `npx` will always try to fetch it and then run it and that is what is happening right now - it takes time loading the _binary` for `maildev` then run it. This way - adding it as a `devDependency` it will not always try to fetch it, rather pick it from the local _node_modules_


Betree:
I believe this was done for a reason, maildev is pretty heavy and we don't want it to be installed by default. I'll let @znarf confirm.

znarf:
This was effectively the reason.

znarf:
I checked and found the reason, it was because of outdated dependencies generating warnings. 
https://github.com/opencollective/opencollective-api/commit/a2454da05c22d360a80f35b2db2c662a4aafdb21

If it's now fixed, it seems so, I'm good with the change.
https://github.com/djfarrelly/MailDev/commit/76f14df32e3ba19901c5a4bb58c1d1b9d4beb9a3#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

PR #2742, opened 2019-10-14
BYS: Filter recommended dependencies based on the ones selected by the user

flickz:
Following up https://github.com/opencollective/backyourstack/pull/244

The user selected collective ids are saved along with `jsonUrl` in `order.data` as `selectedCollectives`. For first and future dispatch, only dependencies with collectives id matching one of `selectedCollectives` will be dispatched.


znarf:
We will not do that.

PR #2743, opened 2019-10-15
Feat/edit public message from grid

jzabala:
Backend for https://github.com/opencollective/opencollective/issues/2433

Closes https://github.com/opencollective/opencollective/issues/2433

### DIFF
--- a/server/graphql/v1/mutations/members.js
+++ b/server/graphql/v1/mutations/members.js
@@ -2,6 +2,7 @@ import { pick } from 'lodash';
 
 import models from '../../../models';
 import errors from '../../../lib/errors';
+import cache from '../../../lib/cache';
 import roles from '../../../constants/roles';
 
 export async function createMember(_, args, req) {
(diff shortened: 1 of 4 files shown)


server/graphql/v1/mutations/members.js
```diff
@@ -93,3 +95,36 @@ export async function editMembership(_, args, req) {
 
   return member.update(pick(args, ['publicMessage']));
 }
+
+/** A mutation to edit the public message of all matching members. */
+export async function editPublicMessage(_, { FromCollectiveId, CollectiveId, message }, req) {
+  if (!req.remoteUser || !req.remoteUser.isAdmin(FromCollectiveId)) {
+    throw new errors.Unauthorized("You don't have the permission to edit member public message");
+  }
+  const updatedMembers = await models.Member.update(
```
Betree: From variable names, it's unclear what this function's return is. One would have to check `update`'s documentation to know that it returns an array with the count in first position.

What about the following:

```suggestion
  const [nbUpdated, updatedMembers] = await models.Member.update(
```

jzabala: Great comment @Betree. I totally agree with your comment. I'll make the change. Thanks for the feedback.

jzabala:
Awesome. @znarf @Betree  thank you very much for taking the time to review the PR. I think one of you needs to merge it since I am not authorized to do so 🙂.

Screen Shot 2019-10-24 at 8 47 13 AM

PR #2745, opened 2019-10-15
Fix one-time sign in links

Betree:
Follow up on https://github.com/opencollective/opencollective-api/pull/2703 (that was reverted in https://github.com/opencollective/opencollective-api/pull/2744).

This didn't introduce problems on production nor staging because it wasn't deployed, but we've seen some related failures in Frontend CI.

When creating login tokens, `lastLoginAt` was set to `firstLogin` in `createCollective` mutation, but not in the other endpoints. The function was crashing on this line in two situations: https://github.com/sbinlondon/opencollective-api/blob/420e9e826778ce96f5bd529ce1c7afec5fb633d2/server/middleware/security/authentication.js#L104

1. For all existing users, because the had their `lastLoginAt` set to `null` but they were calling the `signin` endpoint (not the `createCollective` endpoint) so the JWT was not set to `firstLogin`

2. (not tested, but I suppose) When you sign up, don't use the "first login" email, then ask for another token by signing in with your email.

I've added a condition to ignore `lastLoginAt` if it's not set for the user and replaced the concept of "first login" by the fact that a null value for this field will work by default.

--

While investigating I also found that we had a `seenAt` column in the `Users` table serving the same purpose. I've removed it (because the new one has a better name) but migrated all the data so we keep the existing information.

I also removed the legacy `User.auth` method and `canEditCurrentCollective` that didn't seemed to be used anywhere.

---

@sbinlondon feel free to review (I've added you to https://github.com/orgs/opencollective/teams/contributors) or add commits.

### DIFF
--- a/.forestadmin-schema.json
+++ b/.forestadmin-schema.json
@@ -5288,7 +5288,7 @@
           "validations": []
         },
         {
-          "field": "lastLogin",
+          "field": "lastLoginAt",
           "type": "Date",
           "defaultValue": null,
           "enums": null,
--- a/.forestadmin-schema.json
+++ b/.forestadmin-schema.json
@@ -5288,6 +5288,21 @@
           "validations": []
         },
         {
+          "field": "lastLogin",
+          "type": "Date",
+          "defaultValue": null,
+          "enums": null,
+          "integration": null,
+          "isFilterable": true,
+          "isReadOnly": false,
+          "isRequired": false,
+          "isSortable": true,
+          "isVirtual": false,
+          "reference": null,
+          "inverseOf": null,
+          "validations": []
+        },
+        {
           "field": "transactions",
           "type": ["Number"],
           "defaultValue": null,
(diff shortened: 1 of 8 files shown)


sbinlondon:
Apologies for the issues! I'm sorry it caused this much trouble. 😫

I'll have a look a bit later once I am home from work if that's ok. Thank you for adding me to the contributors list!

Betree:
> Apologies for the issues! I'm sorry it caused this much trouble.

Don't worry, the impact was very limited and no user actually faced this issue.

sbinlondon:
Thanks for fixing this @Betree. Do you think that it needs further work now or is it ready to go?

I'm hoping that next time I'll know the codebase a bit better and won't run into so many issues implementing a feature 🙈

Betree:
I think it's ready, waiting for a review from @znarf to confirm.

znarf:
Works as expected for me, merging.

PR #2759, opened 2019-10-17
Add mutation for contact collective form

flickz:
Resolve https://github.com/opencollective/opencollective/issues/2532

API for https://github.com/opencollective/opencollective-frontend/pull/2848

Resolver: `sendMessageToCollectives`
Parameters: `id `,  `[subject]` `message`

Returns: `{ success: true }`


server/graphql/v1/mutations/collectives.js
```diff
@@ -911,3 +922,51 @@ export async function deleteUserCollective(_, args, req) {
...
+  const existingCount = (await cache.get(countCacheKey)) || 0;
+  if (existingCount > maxEmailMessagePerHour) {
+    throw new errors.RateLimitExceeded();
+  }
+  const fromCollective = await models.Collective.findByPk(req.remoteUser.CollectiveId);
+  const data = {
+    fromCollective,
+    collective,
+    subject: args.subject || null,
+    message: sanitize(args.message.trim()),
```
Betree: As per https://www.npmjs.com/package/sanitize-html#what-if-i-dont-want-to-allow-any-tags, I think we have to pass the following params to strip all HTML:

```es6
{
  allowedTags: [],
  allowedAttributes: {}
}

**Comment by flickz** [comment]
> canContact on the GraphQL collective type. Maybe we should add it to the Collective.prototype to centralize it?

Yes we should, just like we did it `canApply()`

File: server/graphql/v1/mutations/collectives.js [hunk 2]
```diff
@@ -911,3 +915,54 @@ export async function deleteUserCollective(_, args, req) {
...
+    fromCollective,
+    collective,
+    subject: args.subject || null,
+    message: sanitize(args.message.trim(), {
+      allowedTags: [],
+      allowedAttributes: {},
+    }),
+    user,
+  };
+  const recipient = `hello@${collective.slug}.opencollective.com`;
```
Betree: I missed this during my first review, the goal being to remove our reliance on `hello@___` email for this feature, we should avoid using it. If that doesn't already exist, we should create a function to easily target collective admins when we send emails.

Betree:
PR ready:
- Updated the behavior to use `Activity` (so we can notify all admins and they can un-subscribe)
- Updated email design

localhost_1080_email_OgGMxcd1_html (1)

PR #2767, opened 2019-10-21
Fix deleted events show up in Contributions

bernawil:
When deleting a Collective, all Members from its hosted events should be also deleted so it doesn't show up in a User's memberOf query.

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -734,7 +734,7 @@ export async function deleteCollective(_, args, req) {
 
   const eventCollectiveIds = await models.Collective.findAll({
     attributes: ['id'],
-    where: { ParentCollectiveId: collective.id },
+    where: { ParentCollectiveId: collective.id, type: types.EVENT },
   }).then(collectives => collectives.map(({ id }) => id));
 
   return models.Member.findAll({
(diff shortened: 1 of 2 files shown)


server/graphql/v1/mutations/collectives.js
```diff
@@ -732,8 +732,13 @@ export async function deleteCollective(_, args, req) {
     throw new Error('Can not delete collective with paid expenses.');
   }
 
+  const eventCollectiveIds = await models.Collective.findAll({
+    attributes: ['id'],
+    where: { ParentCollectiveId: collective.id },
```
Betree: `ParentCollectiveId` can have other uses, we need to double check with the type

```suggestion
    where: { ParentCollectiveId: collective.id, type: types.EVENT },
```

jeffin143:
@Betree @maddhruv Pr solves this issues https://github.com/opencollective/opencollective/issues/2516 ?

znarf:
Ok, we will not do that in the end.

- Events should not be automatically deleted if they have a balance. Events should be only manually deleted. Also, similar to Collectives, deletion should be impossible if they had transactions. Only "archiving" should be possible.
- Collectives should not be deleted if they have non-deleted events. Only "archiving" should then be possible.

This work will be tackled in other issues:
- Deleting Events https://github.com/opencollective/opencollective/issues/2496
- Preventing Deletion of Collective with Events https://github.com/opencollective/opencollective/issues/2577

PR #2776, opened 2019-10-22
Limit hello@* emails to collectives, events and hosts

Betree:
This will prevent catch-all emails from being used to contact organizations and users.

See https://github.com/opencollective/opencollective-api/blob/1e3e888f765c732d9dd3254ef715f3dec3b8a745/server/models/Collective.js#L864 for `canContact` definition.

### DIFF
--- a/server/controllers/services/email.js
+++ b/server/controllers/services/email.js
@@ -183,16 +183,26 @@ export const webhook = (req, res, next) => {
   // If an email is sent to [info|hello|members|admins|organizers]@:collectiveSlug.opencollective.com,
   // we simply forward it to admins who subscribed to that mailinglist (no approval process)
   if (mailinglist === 'admins') {
-    return sendEmailToList(recipient, {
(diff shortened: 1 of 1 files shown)


server/controllers/services/email.js
```diff
@@ -183,16 +183,26 @@ export const webhook = (req, res, next) => {
...
-      subject: email.subject,
-      body,
-      from: email.from,
-    })
-      .then(() => res.send('ok'))
-      .catch(e => {
-        debugWebhook('Error: ', e);
-        next(e);
-      });
+    return models.Collective.findOne({ where: { slug: collectiveSlug } }).then(collective => {
```
znarf: Rewrite with `async/await` perhaps?

Betree: No touching to be able to merge now but I'll be more careful about removing promises to switch to async/await when possible in the future

PR #2786, opened 2019-10-23
Update to node 12

znarf:


maddhruv:
Hey @znarf can we bump it to 
```
Node 12.13.0
npm 6.12.0
```
As `12.13.0` is now an LTS version

znarf:
@maddhruv yes we can, thank you for spotting that. I didn't check properly what was the latest

package.json
```diff
@@ -5,8 +5,8 @@
   "author": "@philmod @arnaudbenard @xdamman @asood123",
   "main": "server/index.js",
   "engines": {
-    "node": "11.8.0",
-    "npm": "6.5.0"
+    "node": "12.13.0",
+    "npm": "6.9.0"
```
maddhruv: npm `6.12.0` is the latest which comes with node `12.13.0`

maddhruv:
in `package-lock.json` some of the package resolvers are resolved to `https://` which is good but some are resolved to `http://` 😕 which I don't understand
@znarf can you do a fresh install

znarf:
@maddhruv no, it's npm buggy behavior that is like that for many years, we're living with that

maddhruv:
> npm buggy behavior

😕 I don't get it, what sort of buggy behavior

znarf:
@maddhruv https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/39

maddhruv:
https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/53
Do we have to do this?
I don't think `http` vs `https` for npm registry will make more difference but still?

znarf:
Removing the modifications to `package-lock.json` because they were out of scope in the first place.

PR #2790, opened 2019-10-23
Skip activity when adding gift card issuer as member

znarf:
- add a context parameter in `addUserWithRole` to be able to pass
  - `order`
  - `skipActivity` flag
- remove `addBackerToCollective` because it's an unnecessary abstraction that makes the code more complicated to modify and navigate
- extract complex behavior in helpers methods to make `addUserWithRole` itself simpler
  - `createCollectiveMemberCreatedActivity`
  - `sendNewMemberEmail`

Fix: https://github.com/opencollective/opencollective/issues/2551

Also fix an issue where the order in this notification email was "fantasy". Can't find the related ticket, issue or discussion.

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -306,13 +306,6 @@ export const executeOrder = async (user, order, options) => {
     order.paymentMethod.save();
   }
 
-  // Credit card charges are synchronous. If the transaction is
-  // created here it means that the payment went through so it's
-  // safe to create subscription after this.
-
-  // The order will be updated to ACTIVE
-  order.interval && transaction && (await createSubscription(order));
-
   // Register user as collective backer
   await order.collective.findOrAddUserWithRole(
     { id: user.id, CollectiveId: order.FromCollectiveId },
@@ -332,6 +325,13 @@ export const executeOrder = async (user, order, options) => {
       { order, skipActivity: true },
     );
   }
+
+  // Credit card charges are synchronous. If the transaction is
+  // created here it means that the payment went through so it's
+  // safe to create subscription after this.
+
+  // The order will be updated to ACTIVE
+  order.interval && transaction && (await createSubscription(order));
 };
 
 const validatePayment = payment => {
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -232,13 +232,6 @@ export const sendEmailNotifications = (order, transaction) => {
   }
 };
 
-export const addBackerToCollective = async (user, collective, TierId) => {
-  return await collective.findOrAddUserWithRole(user, roles.BACKER, {
-    CreatedByUserId: user.id,
-    TierId,
-  });
-};
-
 export const createSubscription = async order => {
   const subscription = await models.Subscription.create({
     amount: order.totalAmount,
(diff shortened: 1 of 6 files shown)


server/models/Collective.js
```diff
@@ -1265,134 +1265,133 @@ export default function(Sequelize, DataTypes) {
...
+      case roles.MEMBER:
+      case roles.ADMIN:
+        await this.sendNewMemberEmail(user, role, member, sequelizeParams);
+        break;
+    }
+
+    return member;
+  };
+
+  Collective.prototype.createCollectiveMemberCreatedActivity = async function(member, context, sequelizeParams) {
```
Betree: I find that name quite verbose, because we're in 'Collective' I feel like `createMemberCreatedActivity` would be enough

server/models/Collective.js
```diff
@@ -1265,134 +1265,133 @@ export default function(Sequelize, DataTypes) {
...
+
+    const memberUser = await models.User.findByPk(
+      user.id,
+      { include: [{ model: models.Collective, as: 'collective' }] },
+      sequelizeParams,
+    );
+
+    // We don't notify if the new member is the logged in user
+    if (get(remoteUser, 'collective.id') === get(memberUser, 'collective.id')) {
+      return;
```
Betree: I don't understand this change

server/lib/payments.js
```diff
@@ -313,32 +306,32 @@ export const executeOrder = async (user, order, options) => {
     order.paymentMethod.save();
   }
 
+  // Credit card charges are synchronous. If the transaction is
+  // created here it means that the payment went through so it's
+  // safe to create subscription after this.
+
+  // The order will be updated to ACTIVE
+  order.interval && transaction && (await createSubscription(order));
+
```
Betree: If for any reason the membership creation fails, we will still create the subscription. Shouldn't we move that after `Register user as collective backer` so that is the membership is not created we don't add the `subscription`?

znarf: It was like that before. Why would the membership creation fail?

I thought the opposite, I wanted to make sure to create the subscription first and then take care of things like membership creation.

It's a bit out of scope and happy to change it back.

server/models/Collective.js
znarf: I agree but I wanted to make it consistent. Will change it I think.

server/models/Collective.js
znarf: We're in `sendNewMemberEmail` not in `addUserWithRole` so it doesn't have to return anything.

server/lib/payments.js
Betree: > It was like that before.

From what I see it was at the end of the function before this PR no?

znarf: Sorry, my answer was not clear at all. I definitely modified that and what you suggest is to go back to the previous behavior.

But, I'm not sure why we should create the subscription after the membership and not the opposite? It sounds more logic to create the subscription first, no?

Betree: Thanks for clarifying! I would have put the membership creation before because the eventuality of having a member without subscription sounds better to me than the eventuality of having a subscription without a member. But it's not a strong argument, that's a preference. I was curious about the rational behind this change, I'm fine with it.

PR #2795, opened 2019-10-24
BackYourStack: Improve first dispatch performance

flickz:
This PR follows up https://github.com/opencollective/backyourstack/pull/248

Checks for `firstDispatchRecommendations` in custom data, use it instead of fetching for recommendations on backyourstack again. At the end of dispatch, remove `firstDispatchRecommendations` from order customData.

### DIFF
--- a/server/graphql/v1/mutations/backyourstack.js
+++ b/server/graphql/v1/mutations/backyourstack.js
@@ -1,4 +1,6 @@
 import moment from 'moment';
+import { omit } from 'lodash';
+
 import models from '../../../models';
 import { dispatchFunds, getNextDispatchingDate, needsDispatching } from '../../../lib/backyourstack/dispatcher';
 
@@ -39,8 +41,10 @@ export async function dispatchOrder(orderId) {
     subscription.data = {
       nextDispatchDate: getNextDispatchingDate(subscription.interval, currentDispatchDate),
     };
-
+    // Remove firstDispatchRecommendations in custom data after the first dispatch is over
+    order.data = omit(order.data, ['customData.firstDispatchRecommendations']);
     await subscription.save();
+    await order.save();
   }
 
   return dispatchedOrders;
--- a/server/lib/backyourstack/dispatcher.js
+++ b/server/lib/backyourstack/dispatcher.js
@@ -80,13 +80,17 @@ export async function dispatchFunds(order) {
   const shareableAmount = transaction.netAmountInCollectiveCurrency;
 
   const collectives = [];
-
   let depRecommendations;
   try {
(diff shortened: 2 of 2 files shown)


znarf:
I'm unsure about the proposed solution, I think we should focus on making `backing.json` fast in the first place.

flickz:
From my testing and observations today using Microsoft organization with highest number of repositories, has about 69 recommendations available for dispatch, it takes between 20-23s to complete dispatch with this proposed solution while it takes between 29-31s without.

Opencollective API has a default timeout of 25s, the request from backyourstack times out and returns with error while the dispatching process continues to run behind. 

Even this proposed solution will fail at some point, thinking about it again. 

> I think we should focus on making backing.json fast in the first place

```
Microsoft example
Fetching from `backing.json` time - 2674.518ms
Total dispatch time - 31048ms

Subtracting the time it took fetch recommendation from the total dispatch time, 
the dispatch will still fail because time exceeds 25s default timeout
```

I think`backing.json` is actually fast with the cache, it's only slow when it goes to github to refetch.

Overall we need a future proof solution that works regardless of the amount repositories and recommended collectives. We can't optimize our way out to achieve this. Right now the chance of  dispatch failure is directly proportional to the number of collectives available for dispatch at once. 

My future proof solution is to consider batching dispatch, for recommendations > 25. This gives us confidence with the dispatching process irrespective of the numbers of recommendations available. 
The batching implementation is unclear in my head currently, hence why I'm not providing technical details.

Also It is important to note that this problem will not occur for recurring dispatch. Only first dispatch has this timeout issue.

znarf:
I don't understand where we are currently spending the 6 to 10 seconds you're optimizing here. What exactly is slow and we are avoiding with this solution?

flickz:
Actually what I was trying to do here is this,  we decided not  save recommended collectives because we wanted to be able to reanalyze repositories, so dependencies that just joined opencollective can be included, we do this through `backing.json` endpoint,  which is a good idea for recurring dispatch but for the first dispatch we have the recommended dependencies collective already, why go back to `backing.json` to analyze and fetch them again?  This is the idea behind this proposed solution, we skip the process for first dispatch. This improves the time it takes to dispatch.

This part of the I've been overlooking acatually takes  70% of the time
https://github.com/opencollective/opencollective-api/blob/master/server/lib/backyourstack/dispatcher.js#L121-L149

This in particular `paymentsLib.executeOrder(order.createdByUser, orderCreated)`

znarf:
Closing for now as I don't think this the right answer to the problem.

I've written about it here: https://github.com/opencollective/backyourstack/issues/247#issuecomment-546838595

PR #2798, opened 2019-10-25
Remove post-merge-install

znarf:
It doesn't play well with our deploy script which is performing a merge on a branch.


maddhruv:
Hey @znarf can you explain a little more about this issue. We would be happy to work on it and improve `post-merge-install`

znarf:
No issue with the tool, I just don’t want a “npm install” in the middle of the deployment process.

PR #2799, opened 2019-10-25
Delete "api v1" payment method endpoint

znarf:
This should be supported first in the REST instance or deprecated if not used.

- [x] Before deploying, the Cloudflare Worker should be updated to send that route to the REST service.


znarf:
Blocked by coverage.

znarf:
It should now be served from the REST service.

PR #2808, opened 2019-10-28
Users: Ability for us to block contact feature

Betree:
This PR introduces a way for us to easily block someone misusing the contact form.

It introduces a `data` field in `Users`, where we can set a `disableContact` field that will:
- Hide the button from the UI (because `canContact` will be false)
- Throw an error if user tries to call the `sendMessage` endpoint

@znarf let me know what you think of this solution and if having a `data` field for `Users` is welcome or not.

### DIFF
--- a/migrations/20191028101813-add-data-to-users.js
+++ b/migrations/20191028101813-add-data-to-users.js
@@ -1,0 +1,11 @@
+'use strict';
+
+module.exports = {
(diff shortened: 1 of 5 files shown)


Betree:
`data` could be renamed to something more specific like `allowedFeatures` if we want to re-use it for similar intents in the future.

PR #2392, opened 2019-08-08
Replaced rounded monetary amounts in emails with non-rounded ones

Gbahdeyboh:

PR #2809, opened 2019-10-28
authentication: Don't fail if token lastLoginAt is invalid

Betree:
This will mitigate the current error in frontend E2E tests while we investigate on the root cause


Betree:
Also fixed a test broken in https://github.com/opencollective/opencollective-api/pull/2392 (we missed it because PR was on a fork).

See https://circleci.com/gh/opencollective/opencollective-api/21223?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #2820, opened 2019-10-29
BackYourStack: Update dispatch process to Async

flickz:
Following up https://github.com/opencollective/backyourstack/issues/247#issuecomment-546838595

This PR updates the dispatch process from previous synchronous method, the collective admin will be notified about dispatch confirmation through an email as shown below:

Screenshot 2019-10-29 at 7 16 00 PM

Screenshot 2019-10-29 at 7 16 18 PM

### DIFF
--- a/server/graphql/v1/mutations.js
+++ b/server/graphql/v1/mutations.js
@@ -901,7 +901,12 @@ const mutations = {
     },
   },
   backyourstackDispatchOrder: {
-    type: new GraphQLList(OrderType),
+    type: new GraphQLObjectType({
+      name: 'BackYourStackResult',
+      fields: {
+        dispatching: { type: GraphQLBoolean },
+      },
+    }),
     args: {
       id: {
         type: new GraphQLNonNull(GraphQLInt),
--- a/server/graphql/v1/mutations.js
+++ b/server/graphql/v1/mutations.js
@@ -902,7 +902,7 @@ const mutations = {
   },
   backyourstackDispatchOrder: {
     type: new GraphQLObjectType({
-      name: 'BackYourStackResult',
+      name: 'BackYourStackDispatchState',
       fields: {
         dispatching: { type: GraphQLBoolean },
       },
--- a/server/graphql/v1/mutations/backyourstack.js
+++ b/server/graphql/v1/mutations/backyourstack.js
@@ -1,6 +1,6 @@
 import moment from 'moment';
 import models from '../../../models';
-import { dispatchFunds, getNextDispatchingDate, needsDispatching } from '../../../lib/backyourstack/dispatcher';
+import { backgroundDispatch, needsDispatching } from '../../../lib/backyourstack/dispatcher';
 
 import status from '../../../constants/order_status';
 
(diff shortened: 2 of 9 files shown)


server/lib/backyourstack/dispatcher.js
```diff
@@ -148,3 +149,35 @@ export async function dispatchFunds(order) {
     { concurrency: 3 },
   );
 }
+
+export function backgroundDispatch(order, subscription) {
+  dispatchFunds(order)
```
znarf: async/await please

server/graphql/v1/mutations/backyourstack.js
```diff
@@ -25,23 +25,7 @@ export async function dispatchOrder(orderId) {
...
-    subscription.data = {
-      nextDispatchDate: getNextDispatchingDate(subscription.interval, currentDispatchDate),
-    };
-
-    await subscription.save();
-  }
-
-  return dispatchedOrders;
+  // Funds are dispatched asynchronously
+  backgroundDispatch(order, subscription);
```
znarf: Don't think `backgroundDispatch` is a great name, it should simply `dispatch`. The function itself is not "background", it's just a dispatcher.

server/lib/notifications.js
```diff
@@ -315,5 +315,14 @@ async function notifyByEmail(activity) {
         template: 'collective.created.opensource',
       });
       break;
+
+    case activityType.BACKYOURSTACK_DISPATCH_CONFIRMED:
+      activity.data.orders.map(async order => {
```
znarf: Does that work? I'm not aware of Array.map being async.

I understand this can technically work, but the syntax should make clear the behavior is understood and not confuse future developers.

server/graphql/v1/mutations.js
```diff
@@ -878,7 +878,12 @@ const mutations = {
     },
   },
   backyourstackDispatchOrder: {
-    type: new GraphQLList(OrderType),
+    type: new GraphQLObjectType({
+      name: 'BackYourStackResult',
```
znarf: BackYourStackDispatchResult ?

server/graphql/v1/mutations/backyourstack.js
```diff
@@ -25,23 +25,7 @@ export async function dispatchOrder(orderId) {
...
-    const currentDispatchDate = (subscription.data && subscription.data.nextDispatchDate) || new Date();
-    subscription.data = {
-      nextDispatchDate: getNextDispatchingDate(subscription.interval, currentDispatchDate),
-    };
-
-    await subscription.save();
-  }
-
-  return dispatchedOrders;
+  // Funds are dispatched asynchronously
```
znarf: Maybe be more explicit, "Funds are dispatched asynchronously, we're not waiting here on purpose"

server/graphql/v1/mutations/backyourstack.js
```diff
@@ -25,23 +25,7 @@ export async function dispatchOrder(orderId) {
...
+  backgroundDispatch(order, subscription);
+  return { dispatching: true };
```
znarf: Looks a bit like an useless object in the end. Why not. I don't have something strongly better for now.

znarf:
The email looks great. It should later be duplicated and sent each month on dispatch.

server/lib/backyourstack/dispatcher.js
```diff
@@ -148,3 +149,31 @@ export async function dispatchFunds(order) {
...
+      UserId: order.CreatedByUserId,
+      CollectiveId: collective.id,
+      data: {
+        orders: dispatchedOrders,
+        collective: collective.info,
+      },
+    });
+  } catch (err) {
+    console.log('>>>> Background dispatch failed');
+    console.error(err);
```
znarf: What do you plan to do if the dispatch is failing and we end up here in production?

flickz: I was about to ask for suggestion on Slack, I'm thinking we can implement a way to retry dispatch just like we have for subscriptions, I also think in a situation where parts of the fund has been dispatched and failure occurred halfway, we don't want to retry that process instead have human investigation, having a real time alert system like "Sentry"  might be helpful. Another challenge will be how to determine what failure to retry.

znarf: Can you propose a plan, short-term, mid-term, long-term? So I can give input on it?

PR #2827, opened 2019-10-30
Integrate maildev in the developer environment

znarf:
No need to run a separate tab anymore.


znarf:
@Betree right now, it's on demand with `MAILDEV=1 npm run dev`. Do you think we should make it the default?

PR #2842, opened 2019-11-04
Implement retries in stripe payment processing

znarf:
- Switch to a two steps Stripe payment process `create` -> `confirm`
- Let Stripe library retry these steps if they fails

fix https://github.com/opencollective/opencollective/issues/2584

~~According to E2E tests, this is not ready.~~

> ~~You cannot confirm this PaymentIntent because it's missing a payment method. To confirm the PaymentIntent with cus_GXUt1sPXWkmqEy, specify a payment method attached to this customer along with the customer ID.~~ FIXED

~~Update:  according to E2E tests, 3Dsecure is not working as expected anymore.~~

Seems ready!

### DIFF
--- a/server/paymentProviders/stripe/creditcard.js
+++ b/server/paymentProviders/stripe/creditcard.js
@@ -113,6 +113,7 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
       customer: hostStripeCustomer.id,
       description: order.description,
       confirm: false,
+      confirmation_method: 'manual',
       metadata: {
         from: `${config.host.website}/${order.fromCollective.slug}`,
         to: `${config.host.website}/${order.collective.slug}`,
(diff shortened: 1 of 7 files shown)


server/paymentProviders/stripe/creditcard.js
```diff
@@ -93,37 +93,37 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
   // Make sure data is available (breaking in some old tests)
   order.data = order.data || {};
 
-  let paymentIntent;
-  if (!order.data || !order.data.paymentIntent) {
-    const payload = {
+  let paymentIntent = order.data && order.data.paymentIntent;
```
Betree: Looking at the line above (`order.data = order.data || {};`) I think we can do directly 

```suggestion
  let paymentIntent = order.data.paymentIntent;
```

PR #2864, opened 2019-11-07
fix -2577 : Prevent collectives to be deleted when they have events attached.

jeffin143:
Handling : https://github.com/opencollective/opencollective/issues/2577

### DIFF
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -1015,7 +1015,7 @@ const CollectiveFields = () => {
         const expenseCount = await models.Expense.count({
           where: { CollectiveId: collective.id, status: 'PAID' },
         });
-        const eventCount = await models.Collectives.count({
+        const eventCount = await models.Collective.count({
           where: { ParentCollectiveId: collective.id, type: types.EVENT },
         });
 
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -1015,8 +1015,11 @@ const CollectiveFields = () => {
         const expenseCount = await models.Expense.count({
           where: { CollectiveId: collective.id, status: 'PAID' },
         });
+        const eventCount = await models.Collectives.count({
+          where: { ParentCollectiveId: collective.id, type: types.EVENT },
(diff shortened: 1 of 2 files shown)


server/graphql/v1/CollectiveInterface.js
```diff
@@ -1015,8 +1015,10 @@ const CollectiveFields = () => {
         const expenseCount = await models.Expense.count({
           where: { CollectiveId: collective.id, status: 'PAID' },
         });
-
```
znarf: Keep the line break?

server/graphql/v1/CollectiveInterface.js
```diff
@@ -1015,8 +1015,11 @@ const CollectiveFields = () => {
         const expenseCount = await models.Expense.count({
           where: { CollectiveId: collective.id, status: 'PAID' },
         });
+        const eventCount = await models.Collectives.count({
```
znarf: This was not tested. `models.Collectives` is undefined and this is broken.

server/graphql/v1/mutations/collectives.js
```diff
@@ -729,6 +729,14 @@ export async function deleteCollective(_, args, req) {
     throw new Error('Can not delete collective with paid expenses.');
   }
 
+  const eventCount = await models.Collectives.count({
+    where: { ParentCollectiveId: collective.id, type: types.EVENT },
```
znarf: This is similarly broken.

jeffin143: > This is similarly broken.

How to check whether it was fixed or not , how should I re run the build ??

znarf: You should just test the feature manually. `models.Collectives` is undefined so this should not go far.

jeffin143: > You should just test the feature manually. `models.Collectives` is undefined so this should not go far.

sorry about that, I missed it first time, since that function was not called so often and hence I thought may be it's ok, but `isdeletable` was called for almost everything and hence it was throwing error, I would be cautious next time.

Rectified the issue, made changes at both the places

znarf:
Rebased

PR #2868, opened 2019-11-08
fix 'Unhandled rejection SequelizeDatabaseError' error messages

EslamHiko:
This PR fixes : https://github.com/opencollective/opencollective/issues/2607
References : https://sequelize.org/master/class/lib/model.js~Model.html (Schema Section) , https://stackoverflow.com/questions/39580973/node-js-defining-postgres-schemas-in-sequelize/39581304#39581304

Before : 
before

After:
after

### DIFF
--- a/server/models/Activity.js
+++ b/server/models/Activity.js
@@ -24,6 +24,8 @@ export default function(Sequelize, DataTypes) {
     },
   );
 
+  Activity.schema('public');
+
   return Activity;
 }
 
--- a/server/models/Application.js
+++ b/server/models/Application.js
@@ -99,5 +99,7 @@ export default function(Sequelize, DataTypes) {
     return Application.build(props).save();
   };
 
+  Application.schema('public');
+
   return Application;
 }
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1701,6 +1701,8 @@ export default function(Sequelize, DataTypes) {
       });
   };
 
+  Collective.schema('public');
+
   // edit the tiers of this collective (create/update/remove)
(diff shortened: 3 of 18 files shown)


znarf:
Great, thank you!

- Why do we need to do that for these specific tables? Not all?
- When is Sequelize precisely creating the tables in this case?

EslamHiko:
- Why do we need to do that for these specific tables? Not all? : I think we can do this for all the other tables, I didn't want to modify unnecessary tables.
I did this for these specific tables because after I did it only for `Transaction` model `Collectives` and `Transactions` Errors disappeared but 'Notifications', `Members` and `PaymentMethods` Errors appeared. The strange thing, I tried solving `Members` error by modifying `Member` model but the error didn't disappear until I did this for `User` Model.

- When is Sequelize precisely creating the tables in this case? : Sequelize creates tables when the model trigger `create` function ref: https://stackoverflow.com/a/52267740.

znarf:
> I think we can do this for all the other tables

I think we should do it as as matter of consistency. Is it possible that there is a single command to do that?

> When is Sequelize precisely creating the tables in this case?

I found that we're reseting the schema quite often in the tests: https://github.com/opencollective/opencollective-api/blob/master/test/utils.js#L47 This is when we're creating and re-creating the tables I think.

EslamHiko:
- I think we should do it as as matter of consistency. Is it possible that there is a single command to do that? : I tried to add `schema: 'public'` here https://github.com/opencollective/opencollective-api/blob/master/server/models/index.js#L56 but this error showed up `Couldn't recreate the schema`.

PR #2869, opened 2019-11-09
Fix/load more expense comments

jzabala:
This PR adds the _expense_ and _expese->comments_ to the api v2. This is going to be used to solve the issue loading more comments for a expense reported on https://github.com/opencollective/opencollective/issues/2590

It also solves an issue I found when requesting the comments of a expense in the api v1.

### DIFF
--- a/server/graphql/v2/input/CommentCreate.js
+++ b/server/graphql/v2/input/CommentCreate.js
@@ -1,0 +1,17 @@
+import { GraphQLString, GraphQLInt, GraphQLInputObjectType, GraphQLNonNull } from 'graphql';
+
+/**
+ * Input type to use as the type for the comment input in createComment mutation.
+ */
+const CommentCreate = new GraphQLInputObjectType({
+  name: 'CommentCreate',
+  fields: () => ({
+    markdown: { type: GraphQLString },
+    html: { type: GraphQLString },
+    CollectiveId: { type: new GraphQLNonNull(GraphQLInt) },
+    ExpenseId: { type: GraphQLInt },
+    UpdateId: { type: GraphQLInt },
+  }),
+});
+
+export { CommentCreate };
--- a/server/graphql/v2/input/CommentCreate.js
+++ b/server/graphql/v2/input/CommentCreate.js
@@ -1,4 +1,4 @@
-import { GraphQLString, GraphQLInt, GraphQLInputObjectType, GraphQLNonNull } from 'graphql';
+import { GraphQLString, GraphQLInt, GraphQLInputObjectType } from 'graphql';
 
 /**
  * Input type to use as the type for the comment input in createComment mutation.
@@ -8,7 +8,6 @@ const CommentCreate = new GraphQLInputObjectType({
   fields: () => ({
     markdown: { type: GraphQLString },
     html: { type: GraphQLString },
-    CollectiveId: { type: new GraphQLNonNull(GraphQLInt) },
     ExpenseId: { type: GraphQLInt },
     UpdateId: { type: GraphQLInt },
   }),
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -29,6 +29,120 @@ import { NotFound } from '../../errors';
 
 import models, { Op } from '../../../models';
 
+const accountFieldsDefinition = () => ({
+  // _internal_id: {
+  //   type: GraphQLInt,
+  //   description: 'The internal database identifier (should not be public)',
+  // },
+  id: {
+    type: GraphQLString,
+    description: 'The public id identifying the account (ie: 5v08jk63-w4g9nbpz-j7qmyder-p7ozax5g)',
+  },
+  slug: {
+    type: GraphQLString,
+    description: 'The slug identifying the account (ie: babel)',
+  },
+  type: {
+    type: AccountType,
+    description: 'The type of the account (BOT/COLLECTIVE/EVENT/ORGANIZATION/INDIVIDUAL)',
+  },
+  name: {
+    type: GraphQLString,
+  },
+  description: {
+    type: GraphQLString,
+  },
+  website: {
+    type: GraphQLString,
+  },
+  twitterHandle: {
+    type: GraphQLString,
+  },
+  githubHandle: {
+    type: GraphQLString,
+  },
+  currency: {
+    type: GraphQLString,
+  },
+  imageUrl: {
+    type: GraphQLString,
+    args: {
+      height: { type: GraphQLInt },
+      format: {
+        type: ImageFormat,
+      },
+    },
+  },
+  createdAt: {
+    type: GraphQLDateTime,
+    description: 'The time of creation',
+  },
+  updatedAt: {
+    type: GraphQLDateTime,
+    description: 'The time of last update',
+  },
+  // stats: {
+  //   type: AccountStats,
+  // },
+  members: {
+    type: MemberCollection,
+    args: {
+      limit: { type: GraphQLInt, defaultValue: 100 },
+      offset: { type: GraphQLInt, defaultValue: 0 },
+      role: { type: new GraphQLList(MemberRole) },
+      accountType: {
+        type: new GraphQLList(AccountType),
+        description: 'Type of accounts (BOT/COLLECTIVE/EVENT/ORGANIZATION/INDIVIDUAL)',
+      },
+    },
+  },
+  memberOf: {
+    type: MemberOfCollection,
+    args: {
+      limit: { type: GraphQLInt, defaultValue: 100 },
+      offset: { type: GraphQLInt, defaultValue: 0 },
+      role: { type: new GraphQLList(MemberRole) },
+      accountType: {
+        type: new GraphQLList(AccountType),
+        description: 'Type of accounts (BOT/COLLECTIVE/EVENT/ORGANIZATION/INDIVIDUAL)',
+      },
+    },
+  },
+  transactions: {
+    type: TransactionCollection,
+    args: {
+      limit: { type: GraphQLInt, defaultValue: 100 },
+      offset: { type: GraphQLInt, defaultValue: 0 },
+      type: {
+        type: TransactionType,
+        description: 'Type of transaction (DEBIT/CREDIT)',
+      },
+      orderBy: {
+        type: ChronologicalOrder,
+      },
+    },
+  },
+  orders: {
+    type: OrderCollection,
+    args: {
+      limit: { type: GraphQLInt, defaultValue: 100 },
+      offset: { type: GraphQLInt, defaultValue: 0 },
+      filter: { type: AccountOrdersFilter },
+      status: { type: new GraphQLList(OrderStatus) },
+      tierSlug: { type: GraphQLString },
+      orderBy: {
+        type: ChronologicalOrder,
+      },
+    },
+  },
+});
+
+export const Account = new GraphQLInterfaceType({
+  name: 'Account',
+  description: 'Account interface shared by all kind of accounts (Bot, Collective, Event, User, Organization)',
+  fields: accountFieldsDefinition,
+});
+
 const accountTransactions = {
   type: TransactionCollection,
   args: {
@@ -113,58 +227,17 @@ const accountOrders = {
 };
 
 export const AccountFields = {
-  // _internal_id: {
-  //   type: GraphQLInt,
-  //   resolve(transaction) {
-  //     return transaction.id;
-  //   },
-  // },
+  ...accountFieldsDefinition(),
   id: {
     type: GraphQLString,
     resolve(collective) {
       return idEncode(collective.id, 'account');
     },
   },
-  slug: {
-    type: GraphQLString,
-    resolve(collective) {
-      return collective.slug;
-    },
-  },
   type: {
     type: AccountType,
     resolve(collective) {
       return invert(AccountTypeToModelMapping)[collective.type];
-    },
-  },
-  name: {
-    type: GraphQLString,
-    resolve(collective) {
-      return collective.name;
-    },
-  },
-  description: {
-    type: GraphQLString,
-    resolve(collective) {
-      return collective.description;
-    },
-  },
-  website: {
-    type: GraphQLString,
-    resolve(collective) {
-      return collective.website;
-    },
-  },
-  twitterHandle: {
-    type: GraphQLString,
-    resolve(collective) {
-      return collective.twitterHandle;
-    },
-  },
-  githubHandle: {
-    type: GraphQLString,
-    resolve(collective) {
-      return collective.githubHandle;
     },
   },
   imageUrl: {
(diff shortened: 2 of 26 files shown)


server/graphql/v2/interface/Account.js
```diff
@@ -29,6 +29,120 @@ import { NotFound } from '../../errors';
 
 import models, { Op } from '../../../models';
 
+const accountFieldsDefinition = () => ({
```
znarf: What is the purpose of this change?

server/graphql/loaders.js
```diff
@@ -258,7 +258,7 @@ export const loaders = req => {
           where: { ExpenseId: { [Op.in]: ExpenseIds } },
           group: ['ExpenseId'],
         })
-          .then(results => sortResults(ExpenseIds, results, 'ExpenseId'))
+          .then(results => sortResults(ExpenseIds, results, 'ExpenseId', { count: 0 }))
```
znarf: Can you maybe land that patch as a separated PR to fast track it?

server/graphql/v2/object/Collective.js
```diff
@@ -10,6 +10,53 @@ export const Collective = new GraphQLObjectType({
...
+        return req.loaders.collective.findById.load(parentCollective.HostCollectiveId);
+      } else {
+        return null;
+      }
+    });
+  }
+  return null;
+}
+
+const Stats = new GraphQLObjectType({
```
znarf: I don't like this object. I guess it is to mock the API v1? Why not having balance as property of the Collective itself?

server/graphql/v2/object/Comment.ts
```diff
@@ -0,0 +1,40 @@
...
+import { Account } from '../interface/Account';
+
+const Comment = new GraphQLObjectType({
+  name: 'Comment',
+  description: 'This represents an Comment',
+  fields: () => {
+    return {
+      id: {
+        type: GraphQLInt,
+      },
```
znarf: As a design principle of the API v2, we decided to not expose the internal identifiers anymore. We're generally using obfuscated ids as replacement, which is not perfect but it's like that.

server/graphql/v2/object/Expense.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { CommentCollection } from '../collection/CommentCollection';
+import { CollectionArgs } from '../interface/Collection';
+
+const Expense = new GraphQLObjectType({
+  name: 'Expense',
+  description: 'This represents an Expense',
+  fields: () => {
+    return {
+      id: { type: GraphQLInt },
```
znarf: Same remark for public ids.

server/graphql/v2/query/ExpenseQuery.ts
```diff
@@ -0,0 +1,19 @@
+import { GraphQLNonNull, GraphQLInt } from 'graphql';
+
+import { Expense } from '../object/Expense';
+import models from '../../../models';
+
+const ExpenseQuery = {
+  type: Expense,
+  args: {
+    id: {
```
znarf: Compared to what I said elsewhere about publics ids, I guess we effectively still need to be able to query by id there. Would be great if we didn't have to.

server/routes.js
```diff
@@ -133,6 +133,9 @@ export default app => {
   app.param('paranoidtransactionid', params.paranoidtransactionid);
   app.param('expenseid', params.expenseid);
 
+  const isProduction = process.env.NODE_ENV === 'production';
+  const isStaging = process.env.NODE_ENV === 'staging';
+  const isDevelopment = !isProduction && !isStaging;
```
znarf: Maybe a breaking line here.

server/routes.js
```diff
@@ -133,6 +133,9 @@ export default app => {
   app.param('paranoidtransactionid', params.paranoidtransactionid);
   app.param('expenseid', params.expenseid);
 
+  const isProduction = process.env.NODE_ENV === 'production';
+  const isStaging = process.env.NODE_ENV === 'staging';
```
znarf: While refactoring that. We now favor `config.env` over `process.env.NODE_ENV`.

server/routes.js
znarf: What about `test` and `circlecli` environments. Maybe just check for `=== 'development'` there.

server/graphql/v2/scalar/DateString.ts
```diff
@@ -0,0 +1,10 @@
+import { GraphQLScalarType } from 'graphql';
+
+const DateString = new GraphQLScalarType({
+  name: 'DateString',
+  serialize: value => {
+    return value.toString();
+  },
+});
+
+export { DateString };
```
znarf: Same remark for .ts extension

server/graphql/loaders.js
jzabala: Sure. I'll cherry pick this commit and remove ti from this PR.

server/graphql/v2/interface/Account.js
jzabala: I noticed a ton a repetition on fields that didn't need a special resolver. For example:
```
  id: {
    type: GraphQLString,
    description: 'The public id identifying the account (ie: 5v08jk63-w4g9nbpz-j7qmyder-p7ozax5g)',
  },
  slug: {
    type: GraphQLString,
    description: 'The slug identifying the account (ie: babel)',
  },
  type: {
    type: AccountType,
    description: 'The type of the account (BOT/COLLECTIVE/EVENT/ORGANIZATION/INDIVIDUAL)',
  },
  name: {
    type: GraphQLString,
  },
  description: {
    type: GraphQLString,
  },
  website: {
    type: GraphQLString,
  },
  twitterHandle: {
    type: GraphQLString,
  },
  githubHandle: {
    type: GraphQLString,
  },
  currency: {
    type: GraphQLString,
  },
  imageUrl: {
    type: GraphQLString,
    args: {
      height: { type: GraphQLInt },
      format: {
        type: ImageFormat,
      },
    },
  },
  createdAt: {
    type: GraphQLDateTime,
    description: 'The time of creation',
  },
  updatedAt: {
    type: GraphQLDateTime,
    description: 'The time of last update',
  },
```
Width this change we can avoid repeating them in both places in the same file. I noticed this in the `Transaction.js` interface too but, I preferred to  wait for you opinion before refactoring more files.

server/graphql/v2/object/Collective.js
jzabala: Yeah, I just mocked API V1 there. Let me check how that `stats` is used in other places on V1. If I don't find any problems I'll add the balance to the Collective itself.

server/graphql/v2/object/Comment.ts
jzabala: Oh I see. That is what this code is for: `return idEncode(transaction.id, 'transaction');`. Ok, I'll make the change there.

server/graphql/v2/object/Expense.ts
jzabala: I'll make the change.

server/graphql/v2/query/ExpenseQuery.ts
jzabala: Well we have a `idDecode` function. I'll send the coded `id` and decode it before making the db query.

server/routes.js
jzabala: yeah, I thought the same but that would break that current functionality for API V1 and maybe the workflow of another dev includes to check the playground during test. I don't know jajaja. But yeah, I can just check for `development`

server/graphql/v2/scalar/DateString.ts
jzabala: I didn't see any comments related to `.ts` files.

server/graphql/v2/interface/Account.js
znarf: But the resolver is missing now. Are you saying it's optional in these simple cases?

server/graphql/v2/scalar/DateString.ts
znarf: I'm not sure where it went!

It was saying I would prefer to not use .ts extension if we're not using typescript within the file.

server/graphql/v2/interface/Account.js
jzabala: Yeah, if we are returning from the db an object like :`{ id: 1, description: "..." }` and the field names are `id` and `description` the engine can resolve them with the values from the object. It is when we have custom logic that we need the resolver function.

znarf: It's great to know, never realized that!

server/graphql/v2/scalar/DateString.ts
jzabala: But won't this cases problems with the ts engine. For example I suppose that the ts engine is not analyzing and compiling js files. So, we are probably losing type safety when we import js files into ts files and vice versa. I am not expert in ts that is just a humble guess.

znarf: Out of consistency, let's keep it as .js for now and we can bring this topic in a separated issue. So we can quietly look at the situation and take a clear informed decision.

server/graphql/loaders.js
jzabala: Done: https://github.com/opencollective/opencollective-api/pull/2877

server/graphql/v2/object/Collective.js
jzabala: I checked and the stats type from api v1 is a little bit bigger that the one I create here (I only copied what I needed) but yeah, it makes sense for a collective to have its balance a part of its fields. I'll move it and delete the type. As the migration progress we can decide where to put the rest of the stats fields.

server/graphql/v2/query/ExpenseQuery.ts
jzabala: After thinking about this one I realized that for the moment we have to use the unencoded version of the `id`. In the frontend we'll be using the expense unencoded id coming from the api v1. I still think it is safe to return the encoded version in the type. Apollo will just register two expenses instead of one.

server/graphql/v2/object/Collective.js
jzabala: Done!

server/graphql/v2/object/Comment.ts
jzabala: Done. I create a new function called `idEncodeResolver`.  It receives a type and returns a resolver that does the encoding.

server/graphql/v2/object/Expense.ts
jzabala: Done!

server/graphql/v2/query/ExpenseQuery.ts
jzabala: I left a TODO comment there. So that we remember to decode the id after the v1 to v2 migration.

server/routes.js
jzabala: Done.

server/routes.js
jzabala: Done!

jzabala: Done!

server/graphql/v2/scalar/DateString.ts
jzabala: Done. I found [this article](https://mariusschulz.com/blog/type-checking-javascript-files-with-checkjs-in-typescript) about the subject.

jzabala:
@znarf I finished all the changes. 

I rebase this one from https://github.com/opencollective/opencollective-api/pull/2877 in hopes that one gets first to `master`.

I also added a couple more test cases.

server/graphql/v2/object/Collective.js
```diff
@@ -10,6 +10,34 @@ export const Collective = new GraphQLObjectType({
...
+      host: {
+        description: 'Get the host collective that is receiving the money on behalf of this collective',
+        type: Account,
+        resolve: resolveHost,
+      },
     };
   },
 });
+
+function resolveHost(collective, _, req) {
```
znarf: I see that this logic is copy/pasted from somewhere else.

I would extract that in some shared utility between v1/v2, similar to loaders https://github.com/opencollective/opencollective-api/blob/master/server/graphql/loaders.js which are also share between v1 and v2.

server/graphql/v2/object/Comment.js
```diff
@@ -0,0 +1,46 @@
...
+      id: {
+        type: GraphQLInt,
+        /** TODO: Uncomment after editComment mutation is migrated to v2.
+         * The string encoded id is not compatible with the edit comment mutation
+         * from v1.
+         */
+        // resolve: idEncodeResolver('comment'),
+      },
+      createdAt: {
+        type: DateString,
```
znarf: In API v2, dates are `GraphQLDateTime`.

server/graphql/v2/query/ExpenseQuery.js
```diff
@@ -0,0 +1,23 @@
...
+import models from '../../../models';
+
+const ExpenseQuery = {
+  type: Expense,
+  args: {
+    id: {
+      /** TODO: When migration is done decode the id field in the resolver before querying the database.
+       * At the moment of creating this field the migration from v1 to v2 has just started and the id
+       * that is going to be used to query the expense is the unencoded one from the api v1.
+       */
```
znarf: Then, the query should be able to resolve `id` in both forms from this day.

Either we support 2 different `args` for that, `id` / `internal_id`.

Or maybe smarter, we regexp match the `id` to check if it's in the new form. If it is, then we decode it. If it's not, we can check that the id is as integer. If it's not, we throw an error.

server/graphql/v2/object/Comment.js
```diff
@@ -0,0 +1,46 @@
...
+  fields: () => {
+    return {
+      id: {
+        type: GraphQLInt,
+        /** TODO: Uncomment after editComment mutation is migrated to v2.
+         * The string encoded id is not compatible with the edit comment mutation
+         * from v1.
+         */
+        // resolve: idEncodeResolver('comment'),
+      },
```
znarf: `id` should be in the new form from this day. If `internal_id` is needed, we can expose it but already mark it as deprecated.

znarf:
Deploying that branch on staging so we can already test the frontend part.

server/graphql/v2/query/ExpenseQuery.js
jzabala: Checking at the database I can see all ids are of type integer. I think checking if the string is a number or not would be enough here.

server/graphql/v2/object/Collective.js
jzabala: Done. There is a `common` folder shared by v1 and v2.

server/graphql/v2/object/Comment.js
jzabala: Done.

server/graphql/v2/object/Comment.js
jzabala: I didn't like the idea of creating a new field deprecated by default. So, I implemented the whole comment crud in v2.

test/utils.js
```diff
@@ -268,3 +268,25 @@ export function expectNoErrorsFromResult(res) {
...
   expect(res.errors).to.not.exist;
 }
+
+/**
+ * This callback type is called `propertyCallback` and is called for every property of an object.
+ *
+ * @callback propertyCallback
+ * @param {string} key - Object key.
+ * @param {*} value - Object value at the given keys.
+ */
```
Betree: I don't understand what this comment stands for

jzabala: This is how jsdoc recommends commenting callback functions https://jsdoc.app/tags-param.html#callback-functions. But since the comment intention was not clear I am inclined to remove it.

jzabala: @Betree I tried to improve the callback definition comment a little bit. Let me know what you think about it know 🙂

jzabala:
By the way, while working on the implementation on v2 I saw some curios implementations. For example the `FromCollectiveId` on create comment is taken from the logged user https://github.com/opencollective/opencollective-api/blob/0e4f433fbf900b6f4331e6688124a5b80b06b3cd/server/graphql/v1/mutations/comments.js#L27 that I think is right but it is possible to change it in the edit https://github.com/opencollective/opencollective-api/blob/0e4f433fbf900b6f4331e6688124a5b80b06b3cd/server/models/Comment.js#L178.

I don't think this should be a changeable property. I made it changeable in v2 since I wanted to confirm it first.

----------------------------

When you create/update a comment you can send the `ExpenseId` and `CollectiveId`. I suppose that collective id is the one the expense belongs to. Someone with knowledge of our api could messed around with the data creating comments for not related `ExpenseId` and `CollectiveId`.
https://github.com/opencollective/opencollective-api/blob/0e4f433fbf900b6f4331e6688124a5b80b06b3cd/server/graphql/v1/inputTypes.js#L374
https://github.com/opencollective/opencollective-api/blob/0e4f433fbf900b6f4331e6688124a5b80b06b3cd/server/graphql/v1/inputTypes.js#L388

znarf:
> I don't think this should be a changeable property. 

I don't see any reason why we would change `FromCollectiveId`, we should not allow that.

> Someone with knowledge of our api could messed around with the data creating comments for not related ExpenseId and CollectiveId.

Good catch. We should make sure there is no mismatch when creating comments. I don't feel CollectiveId is useful as input. If required, it should be deduced from the Expense.

jzabala:
> > I don't think this should be a changeable property.
> 
> I don't see any reason why we would change `FromCollectiveId`, we should not allow that.
> 
> > Someone with knowledge of our api could messed around with the data creating comments for not related ExpenseId and CollectiveId.
> 
> Good catch. We should make sure there is no mismatch when creating comments. I don't feel CollectiveId is useful as input. If required, it should be deduced from the Expense.

@znarf done. I deprecated the args in v1 and remove them from v2. You can check that latest commit on this PR: https://github.com/opencollective/opencollective-api/pull/2869/commits/be6d60fee2fe234424e3e00ef3196719451d4487

The frontend is up to date with these changes.

server/graphql/v2/input/CommentCreate.js
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Input type to use as the type for the comment input in createComment mutation.
+ */
+const CommentCreate = new GraphQLInputObjectType({
+  name: 'CommentCreate',
+  fields: () => ({
+    markdown: { type: GraphQLString },
+    html: { type: GraphQLString },
+    ExpenseId: { type: GraphQLInt },
+    UpdateId: { type: GraphQLInt },
```
znarf: I kind of missed that but one of the design goals of the GraphQL API v2 is to stop exposing directly internal integers. We should think about an alternative.

jzabala: 🤔 this looks like the one that got away 🙂.

Let me create a PR to change it.

Also I would like to do a little bit of refactoring to use some of the functions I introduced in the PR.

PR #2877, opened 2019-11-11
Fix/requesting expense comments when empty

jzabala:
This PR solves a bug found while working on issue [#2590](https://github.com/opencollective/opencollective/issues/2590).

When requesting expense comments from `Expense->comments` field, if there are not comments the api v1 breaks.

### DIFF
--- a/server/graphql/loaders.js
+++ b/server/graphql/loaders.js
@@ -258,7 +258,7 @@ export const loaders = req => {
           where: { ExpenseId: { [Op.in]: ExpenseIds } },
           group: ['ExpenseId'],
         })
-          .then(results => sortResults(ExpenseIds, results, 'ExpenseId'))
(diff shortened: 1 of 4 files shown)


znarf:
@jzabala Seems there is one extra commit?

jzabala:
> @jzabala Seems there is one extra commit?

Yeah, I created this commit after moving the function `expectNoErrorsFromResult` to the utils file and creating the function `graphqlQueryV2`. I use the function `expectNoErrorsFromResult` in the new test I create to verify this bug so I just cherry picked both. I can rebase the PR and remove the function `graphqlQueryV2` from the first commit but since those changes will get to master soon (I am almost done with https://github.com/opencollective/opencollective-api/pull/2869), I just left them there.

PR #2883, opened 2019-11-12
Fix graphql.expenses.test deadlock

EslamHiko:
This is a PR to fix : https://github.com/opencollective/opencollective/issues/2623

What was exactly happening that the `#markExpenseAsUnpaid`'s beforeEach `utils.resetTestDB` was still working after it has done. so when`#unapproveExpense`'s  beforeEach `utils.resetTestDB` was called `#markExpenseAsUnpaid`'s beforeEach `utils.resetTestDB` wasn't done yet and that caused the deadlock.

Before : 

before44

After : 

after44

After With Running All Tests : 

after442

### DIFF
--- a/test/graphql.expenses.test.js
+++ b/test/graphql.expenses.test.js
@@ -90,9 +90,13 @@ const addFunds = async (user, hostCollective, collective, amount) => {
 };
 
 describe('GraphQL Expenses API', () => {
-  beforeEach(utils.resetTestDB);
+  beforeEach(async () => {
+    await utils.resetTestDB();
+  });
 
-  beforeEach(utils.resetCaches);
+  beforeEach(async () => {
+    await utils.resetCaches();
+  });
 
   describe('#allExpenses', () => {
     it('fails if collective not found', async () => {


znarf:
Triggering CI. And yes, as a recurring contributor, you're welcome and encouraged to push your branches to our repository.

EslamHiko:
@Betree I think it was starting 2 promises at the same time starting a Session of Postgresql while the other session hasn't finished yet.

I will do it next PR.

@znarf I don't know why errors still exist in CircleCI tests, I'm using ubuntu.

znarf:
It's actually difficult to guarantee that the fix is working because the issue is not 100% reproducible.

@EslamHiko it seems you are pretty confident the fix is the right one, how where you able to test in on your side?

EslamHiko:
@znarf I'm running `npm run test` on my local machine and it disappeared after my modification I don't know how could CircleCI be different from my machine in running tests :confused: .

znarf:
@EslamHiko Sounds good! It never happened on my own machine so I'm curious.

znarf:
Thank you! Merging so we can try it.

PR #2894, opened 2019-11-15
Feat/add docker files

jzabala:
To try this branch you need to install [Docker and Docker Compose](https://docs.docker.com/v17.09/engine/installation/).

# Commands

Before running any of the predefined commands for the first time or every time you want to reinstall the node modules and run the db migrations, remove the `node_modules` folder (`rm -rf node_modules`).

#### `docker-compose up`

Setup the api and runs `npm start`. 

#### `OPENCOLLECTIVE_DOCKER_INIT_CMD=dev docker-compose up` 

Setup the api in a way that restart the server when files change.

#### `OPENCOLLECTIVE_DOCKER_INIT_CMD=e2e docker-compose up`

Setup the api in e2e mode.

#### `OPENCOLLECTIVE_DOCKER_INIT_CMD=test docker-compose up`

Setup the api and runs the unit/integration tests. After the tests are done running you need to stop the app manually with `ctrl+c`.

The db is restored and migrated every time you run this command.

#### `OPENCOLLECTIVE_DOCKER_INIT_CMD="<`npm run ...` command>" docker-compose up`

Start the db and run the custom `npm run ...` command.

Only the predefined commands will install node modules and run migrations on your behalf. Before using a custom command make sure the environment is in the state you want.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,33 @@
+version: '3.7'
+
+services:
+  opencollective_api:
+    build:
+      context: . # Build the api image from the Dockerfile
+    command: ${OPENCOLLECTIVE_DOCKER_INIT_CMD:-start} # Use this env var to send pre-defined and custom commands
+    ports:
+      - '3060:3060' # App port
+      - '1080:1080' # Mail gui port
+      - '1025:1025'
+      - '9229:9229' # Debug port
+    volumes:
+      - .:/usr/src/app # Mount the app in the container
+    depends_on:
+      - opencollective_db # Add dependency on the db service
+    environment:
+      - PGHOST=opencollective_db # The db host is the name of the service
+      - PGUSER=postgres
+      - PG_URL=postgres://opencollective@opencollective_db:5432/opencollective_dvl
+      - PG_MAINTENANCE_URL=postgres://opencollective_db:5432/postgres
+
+  opencollective_db:
+    image: mdillon/postgis:9.6
+    volumes:
+      - opencollective-db-data:/var/lib/postgresql/data # Persist data between mounts
+    ports:
+      - '5432:5432' # Export PG port to the host allowing to access the db with a pg client
+    environment:
+      - POSTGRES_USER=postgres # Default user
+
+volumes:
+  opencollective-db-data:
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -8,7 +8,7 @@ installNodeModules() {
   npm install
 }
 
-waitForPostgresToByReady() {
+waitForPostgresToBeReady() {
   # Code by https://starkandwayne.com/blog/how-to-know-when-your-postgres-service-is-ready/
   until pg_isready
   do
@@ -19,7 +19,7 @@ waitForPostgresToByReady() {
 
 seedDB() {
   echo "Seeding database"
-  waitForPostgresToByReady
+  waitForPostgresToBeReady
   npm run postinstall
 }
 
@@ -34,7 +34,7 @@ doesDatabaseExists() {
 
 setupTestingDB() {
   echo "Setting up testing db"
-  waitForPostgresToByReady
+  waitForPostgresToBeReady
   if ! doesDatabaseExists "opencollective_test"; then
     npm run db:setup
     createdb opencollective_test || true
@@ -68,7 +68,7 @@ case $1 in
     echo "Starting in e2e mode"
     setupEnvironmentIfNodeModulesDoesNotExists
     if ! $setupScriptRan; then
-      waitForPostgresToByReady
+      waitForPostgresToBeReady
       npm run db:restore && npm run db:migrate
     fi
     npm run build
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -7,6 +7,7 @@ installNodeModules() {
   echo "Installing node modules"
   npm install
 }
+
 waitForPostgresToByReady() {
   # Code by https://starkandwayne.com/blog/how-to-know-when-your-postgres-service-is-ready/
   until pg_isready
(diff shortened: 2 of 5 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    environment:
+      - PGHOST=opencollective_db # The db host is the name of the service
+      - PGUSER=postgres
+      - PG_URL=postgres://opencollective@opencollective_db:5432/opencollective_dvl
+      - PG_MAINTENANCE_URL=postgres://opencollective_db:5432/postgres
+
+  opencollective_db:
+    image: mdillon/postgis:9.6
+    volumes:
+      - opencollective-db-data:/var/lib/postgresql/data # Persist data between mounts
```
Betree: Is that going to work on Windows?

jzabala: It should. In this case a named volume is created with the name _opencollective-db-data_ associated to the location _/var/lib/postgresql/data_ in the container. You can check this name with the command `docker volume ls`.

Here is an overview on the [different kinds of volumes](https://success.docker.com/article/different-types-of-volumes). We could also remove this section and create an anonymous volume but as the doc says it can be difficult to refer to the same volume over time.

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
+version: '3.7'
+
+services:
+  opencollective_api:
+    build:
+      context: . # Build the api image from the Dockerfile
+    command: ${OPENCOLLECTIVE_DOCKER_INIT_CMD} # Use this env var to send pre-defined and custom commands
```
Betree: Currently running `docker-compose up` generates a warning:

> WARNING: The OPENCOLLECTIVE_DOCKER_INIT_CMD variable is not set. Defaulting to a blank string.

Maybe we should add a default value here to avoid that?

znarf:
I just had a test and it seems the Docker image is reusing the node_modules folder. It seems to be a problem because some modules are not working if they were compiled on the host (here Mac OS) and we are trying to run them in the container.

jzabala:
> I just had a test and it seems the Docker image is reusing the node_modules folder. It seems to be a problem because some modules are not working if they were compiled on the host (here Mac OS) and we are trying to run them in the container.

Yeah, this is totally the case. That is why I advice at the beginning that it is needed to remove the `node_modules` before running any of the command.

**Why to install the node_modules in the container?** Because it makes this solution OS independent. Everything is done in Docker from the same image, reducing OS related errors to the minimum.

**Why do the container expose the node_modules to the host?** I do that because of Text Editors (in my case VSCode). VS analyzes the `node_modules` and in my case it uses prettier, lint and other tools versions from the ones installed in this project. I also have a feature that autocompletes `node_modules` when I am importing them.

**Should we just remove the node_modules from the container everything we run a command?** I thought about that but to install the `node_modules` takes time and removing and installing them everything a command is run would be a nightmare. A better option would be to add a `-f` or `-r` param to the command like: `start -r`, and to rebuild the whole thing from there, but I wanted to gather comments before dedicating more time to this.

docker-compose.yml
jzabala: Sure. Working on it.

jzabala: Great catch. Working on it.

jzabala: Done!

scripts/setup.sh
```diff
@@ -0,0 +1,86 @@
...
+set -e
+
+setupScriptRan=false
+
+installNodeModules() {
+  echo "Installing node modules"
+  npm install
+}
+
+waitForPostgresToByReady() {
```
AliMirlou:
```suggestion
waitForPostgresToBeReady() {
```

scripts/setup.sh
```diff
@@ -0,0 +1,86 @@
...
+  until pg_isready
+  do
+    echo "Waiting for postgres"
+    sleep 2;
+  done
+}
+
+seedDB() {
+  echo "Seeding database"
+  waitForPostgresToByReady
```
AliMirlou:
```suggestion
  waitForPostgresToBeReady
```

scripts/setup.sh
```diff
@@ -0,0 +1,86 @@
...
+  if psql -lqt | cut -d \| -f 1 | grep -qw $1; then
+    return 0
+  else
+    return 1
+  fi
+}
+
+setupTestingDB() {
+  echo "Setting up testing db"
+  waitForPostgresToByReady
```
AliMirlou:
```suggestion
  waitForPostgresToBeReady
```

scripts/setup.sh
```diff
@@ -0,0 +1,86 @@
...
+  'dev')
+    echo "Starting in dev mode"
+    setupEnvironmentIfNodeModulesDoesNotExists
+    exec npm run dev
+    ;;
+  'e2e')
+    echo "Starting in e2e mode"
+    setupEnvironmentIfNodeModulesDoesNotExists
+    if ! $setupScriptRan; then
+      waitForPostgresToByReady
```
AliMirlou:
```suggestion
      waitForPostgresToBeReady
```

scripts/setup.sh
Betree: Good catch!

znarf:
@kewitz Could you have a look at that and share your opinion when you have some time?

znarf:
@kewitz Could you take care of that? Edit what you think is worth being changed? Then merge? 

But also please make sure not not fall in a rabbit hole ...

kewitz:
The complex thing about implement docker is that it kinda needs to be a silver bullet, otherwise it can cause some trouble to the community.
@jzabala anything to add to the discussion? Are you interested in moving this forward?

znarf:
This was pretty good work and we should keep the branch for reference in the future.

Closing the issue for now as there is no consensus within the team on how to move it forward.

PR #2898, opened 2019-11-18
Email reminder for pending orders

leonardohof:
Resolve opencollective/opencollective#1508

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -225,6 +225,8 @@ export const sendEmailNotifications = (order, transaction) => {
     sendOrderProcessingEmail(order);
     if (isProvider('opencollective.giftcard', order.paymentMethod)) {
       sendSupportEmailForManualIntervention(order); // async
+    } else {
+      sendManualPendingOrderEmail(order);
     }
   } else {
     order.transaction = transaction;
@@ -422,6 +424,43 @@ const sendOrderProcessingEmail = async order => {
     });
   }
   return emailLib.send('order.processing', user.email, data, {
+    from: `${collective.name} <hello@${collective.slug}.opencollective.com>`,
+  });
+};
+
+const sendManualPendingOrderEmail = async order => {
+  const { collective, fromCollective } = order;
+  const user = order.createdByUser;
+  const host = await collective.getHostCollective();
+  const data = {
+    order: order.info,
+    user: user.info,
+    collective: collective.info,
+    host: host.info,
+    fromCollective: fromCollective.activity,
+    pendingOrderLink: `${config.host.website}/${collective.slug}/orders/${order.id}`,
+  };
+
+  return emailLib.send('order.new.pendingFinancialContribution', user.email, data, {
+    from: `${collective.name} <hello@${collective.slug}.opencollective.com>`,
+  });
+};
+
+export const sendReminderPendingOrderEmail = async order => {
+  const { collective, fromCollective } = order;
+  const user = order.createdByUser;
+  const host = await collective.getHostCollective();
+  const data = {
+    order: order.info,
+    user: user.info,
+    collective: collective.info,
+    host: host.info,
+    fromCollective: fromCollective.activity,
+    markAsPaidLink: `${config.host.website}/orders/${order.id}/mark-as-paid`,
+    viewDetailsLink: `${config.host.website}/${collective.slug}/orders/${order.id}`,
+  };
+
+  return emailLib.send('order.reminder.pendingFinancialContribution', user.email, data, {
     from: `${collective.name} <hello@${collective.slug}.opencollective.com>`,
   });
 };
--- a/cron/daily/80-pending-orders-reminder.js
+++ b/cron/daily/80-pending-orders-reminder.js
@@ -1,0 +1,41 @@
+#!/usr/bin/env node
+import '../../server/env';
+
+import models, { Op } from '../../server/models';
+import status from '../../server/constants/order_status';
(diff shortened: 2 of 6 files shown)


leonardohof:
@znarf Could you review, please?

cron/daily/100-pending-orders-reminder.js
```diff
@@ -0,0 +1,39 @@
...
+      { model: models.User, as: 'createdByUser' },
+      { model: models.Collective, as: 'collective' },
+    ],
+  });
+
+  return orders;
+};
+
+const run = async () => {
+  //
```
Betree:
```suggestion
```

cron/daily/100-pending-orders-reminder.js
```diff
@@ -0,0 +1,39 @@
...
+  const reminderDate = process.env.START_DATE ? new Date(process.env.START_DATE) : new Date();
+  reminderDate.setDate(reminderDate.getDate() - REMINDER_DAYS);
+
+  const orders = await fetchPendingOrders(reminderDate);
+  orders.forEach(async order => {
+    await libPayments.sendReminderPedingOrderEmail(order);
```
Betree:
```suggestion
    await libPayments.sendReminderPendingOrderEmail(order);
```

server/lib/payments.js
```diff
@@ -225,6 +225,8 @@ export const sendEmailNotifications = (order, transaction) => {
     sendOrderProcessingEmail(order);
     if (isProvider('opencollective.giftcard', order.paymentMethod)) {
       sendSupportEmailForManualIntervention(order); // async
+    } else {
+      sendManualPedingOrderEmail(order);
```
Betree:
```suggestion
      sendManualPendingOrderEmail(order);
```

cron/daily/100-pending-orders-reminder.js
```diff
@@ -0,0 +1,39 @@
...
+
+  return orders;
+};
+
+const run = async () => {
+  //
+  const reminderDate = process.env.START_DATE ? new Date(process.env.START_DATE) : new Date();
+  reminderDate.setDate(reminderDate.getDate() - REMINDER_DAYS);
+
+  const orders = await fetchPendingOrders(reminderDate);
```
Betree: My understanding is that this will send a reminder email for each pending order every day during 4 days, while what we want is to send **only one** email after 4 days. Am I missing something?

server/lib/payments.js
```diff
@@ -425,3 +427,40 @@ const sendOrderProcessingEmail = async order => {
...
+    fromCollective: fromCollective.activity,
+    pendingOrderLink: `${config.host.website}/${collective.slug}/orders/${order.id}`,
+  };
+
+  return emailLib.send('order.new.pendingFinancialContribution', user.email, data, {
+    from: `${collective.name} <hello@${collective.slug}.opencollective.com>`,
+  });
+};
+
+export const sendReminderPedingOrderEmail = async order => {
```
Betree:
```suggestion
export const sendReminderPendingOrderEmail = async order => {
```

server/lib/payments.js
```diff
@@ -425,3 +427,40 @@ const sendOrderProcessingEmail = async order => {
     from: `${collective.name} <hello@${collective.slug}.opencollective.com>`,
   });
 };
+
+const sendManualPedingOrderEmail = async order => {
```
Betree:
```suggestion
const sendManualPendingOrderEmail = async order => {
```

cron/daily/100-pending-orders-reminder.js
leonardohof: No, It's a misunderstanding from me. I fixed it.

cron/daily/100-pending-orders-reminder.js
```diff
@@ -0,0 +1,41 @@
...
+  orders.forEach(async order => {
+    await libPayments.sendReminderPendingOrderEmail(order);
+  });
+};
+
+run();
```
znarf: Please rename it as `80` instead of `100`. We're relying on that to run the jobs sequentially. A bit old school but ...

leonardohof: Changed.

server/lib/payments.js
```diff
@@ -225,6 +225,8 @@ export const sendEmailNotifications = (order, transaction) => {
     sendOrderProcessingEmail(order);
     if (isProvider('opencollective.giftcard', order.paymentMethod)) {
       sendSupportEmailForManualIntervention(order); // async
+    } else {
```
znarf: @Betree are we sure pledges are not falling in that case?

Betree: I'm not, good catch. I'm going to test it manually

Betree: I can confirm that it's safe, because we check `isActive` in `createOrder` (https://github.com/opencollective/opencollective-api/blob/e5d82f6a75838bb6cbcb7c3d74561e690544fbd7/server/graphql/v1/mutations/orders.js#L625) and never call `executeOrder` for pledged collectives

znarf:
I think some of the `PENDING` orders will not be manual payments. Maybe at some point we should rename them 'MANUAL' to clarify the situation. Looking at that: https://github.com/opencollective/opencollective/issues/1994#issuecomment-503435051 It can still be:

- Non confirmed orders after Stripe SCA
- API crashes

I don't think this should prevent the current issue to be merged, but just a note for the future.

PR #2901, opened 2019-11-19
GraphQL v1: accept slugs as query parameters in allCollectives

NotJohs:
… and return collective data.
Re [#2565](https://github.com/opencollective/opencollective/issues/2565)

### DIFF
--- a/server/graphql/v1/queries.js
+++ b/server/graphql/v1/queries.js
@@ -795,7 +795,7 @@ const queries = {
   allCollectives: {
     type: CollectiveSearchResultsType,
     args: {
-      renderCollectivesFromList: {
+      slugs: {
         type: new GraphQLList(GraphQLString),
         description: 'Fetch collectives with a list of collective slug',
       },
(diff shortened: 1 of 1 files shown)


server/graphql/v1/queries.js
```diff
@@ -795,6 +795,10 @@ const queries = {
   allCollectives: {
     type: CollectiveSearchResultsType,
     args: {
+      renderCollectivesFromList: {
```
znarf: should be named `slugs`

server/graphql/v1/queries.js
```diff
@@ -859,6 +863,12 @@ const queries = {
         include: [],
       };
 
+      if (args.renderCollectivesFromList) {
+        query.where.id = await Promise.all(
+          args.renderCollectivesFromList.map(async id => (typeof id === 'string' ? await fetchCollectiveId(id) : id)),
+        );
```
znarf: Should just be a where `IN` condition with the slugs. As simple as it can be.

NotJohs: Yes. I just made that change.

server/graphql/v1/queries.js
NotJohs: Understood.

PR #2902, opened 2019-11-19
WIP: Fix "Relation {Table} does not exist" errors in CI #2607

leonardohof:
Resolve opencollective/opencollective#2607


znarf:
@leonardohof can you check the diff, it seems too many things were committed. Otherwise, looks good!

PR #2909, opened 2019-11-19
Don't create membership for PENDING orders

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2643
Related to https://opencollective.freshdesk.com/a/tickets/2921

Only register as a backer member when a transaction is created. This doesn't break event's free tickets thanks to https://github.com/opencollective/opencollective-api/blob/e5d82f6a75838bb6cbcb7c3d74561e690544fbd7/server/graphql/v1/mutations/orders.js#L500

Properly tested: membership is created when the order is confirmed.


Betree:
@znarf fixed

PR #2911, opened 2019-11-19
Deleting events #2496

leonardohof:
Resolve opencollective/opencollective#2496

### DIFF
--- a/server/graphql/v1/inputTypes.js
+++ b/server/graphql/v1/inputTypes.js
@@ -187,6 +187,7 @@ export const CollectiveInputType = new GraphQLInputObjectType({
     firstName: { type: GraphQLString },
     lastName: { type: GraphQLString },
     isIncognito: { type: GraphQLBoolean },
+    isDeletable: { type: GraphQLBoolean },
   }),
 });
 


server/graphql/v1/inputTypes.js
```diff
@@ -187,6 +187,7 @@ export const CollectiveInputType = new GraphQLInputObjectType({
     firstName: { type: GraphQLString },
     lastName: { type: GraphQLString },
     isIncognito: { type: GraphQLBoolean },
+    isDeletable: { type: GraphQLBoolean },
```
znarf: I see that you're following the example of `isIncognito`, which was not a great model in the first place, I'm sad we accepted that change when it was introduced.

I think we should better filter these properties before submitting to GraphQL.

leonardohof: I'm not sure If I understood what you're saying. Could you give me more details? Is the problem that the field being "calculated" on the GraphQL?
image

znarf: Yes, it's a dynamic "calculated" field, there is no reason to add it in the Input because we will never want to update it. I think you're adding it here to prevent an error, but I think we should not send it in the first place.

leonardohof: Yes, I added here because the e2e tests for the front were broking asking for this field on the schema.

znarf:
Closing the PR because at this point this modification should not be needed and is not desirable.

I think we should be able to solve https://github.com/opencollective/opencollective-frontend/pull/2996 without modifying the API.

PR #2951, opened 2019-11-25
chore(bys): hide prepaid method created for backyourstack

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2647

### DIFF
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -1541,6 +1541,10 @@ const CollectiveFields = () => {
         // Filter only "saved" stripe Payment Methods
         // In the future we should only return the "saved" whatever the service
         paymentMethods = paymentMethods.filter(pm => pm.service !== 'stripe' || pm.saved);
+
+        // Filter out prepaid method created for backyourstack.
+        paymentMethods = paymentMethods.filter(pm => pm.data && pm.data.created_for !== 'backyourstack');
+
         if (args.service) {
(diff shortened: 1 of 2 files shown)


server/lib/backyourstack/dispatcher.js
```diff
@@ -63,7 +63,10 @@ async function createPaymentMethod(originalCreditTransaction) {
     service: 'opencollective',
     type: 'prepaid',
     uuid: uuidV4(),
-    data: { HostCollectiveId: originalCreditTransaction.HostCollectiveId },
+    data: {
+      HostCollectiveId: originalCreditTransaction.HostCollectiveId,
+      created_for: 'backyourstack',
```
znarf: We generally don't use Snake case (underscore).

znarf: Prefer some generic property such as `hidden: true`.

server/graphql/v1/CollectiveInterface.js
```diff
@@ -1529,6 +1529,10 @@ const CollectiveFields = () => {
         // Filter only "saved" stripe Payment Methods
         // In the future we should only return the "saved" whatever the service
         paymentMethods = paymentMethods.filter(pm => pm.service !== 'stripe' || pm.saved);
+
+        // Filter out prepaid method created for backyourstack.
+        paymentMethods = paymentMethods.filter(pm => pm.data && pm.data.created_for !== 'backyourstack');
```
znarf: What happens if `pm.data` is null?

PR #2961, opened 2019-11-26
Fix deadlocks while running test

gmarcos87:
This pull request resolves:
- [x] removes "...DatabaseError: deadlock detected.." https://github.com/opencollective/opencollective/issues/2623

### DIFF
--- a/test/graphql.mutation.test.js
+++ b/test/graphql.mutation.test.js
@@ -49,7 +49,11 @@ describe('Mutation Tests', () => {
 
   after(() => sandbox.restore());
 
-  beforeEach('reset db', () => utils.resetTestDB());
+  beforeEach('reset db', async () => {
+    await new Promise(res => setTimeout(res, 500));
+    await utils.resetTestDB();
+    return;
+  });
 
   beforeEach('create user1', () => models.User.createUserWithCollective(utils.data('user1')).tap(u => (user1 = u)));
   beforeEach('create host user 1', () =>
--- a/test/graphql.mutation.test.js
+++ b/test/graphql.mutation.test.js
@@ -52,7 +52,6 @@ describe('Mutation Tests', () => {
   beforeEach('reset db', async () => {
     await new Promise(res => setTimeout(res, 500));
     await utils.resetTestDB();
-    return;
   });
 
   beforeEach('create user1', () => models.User.createUserWithCollective(utils.data('user1')).tap(u => (user1 = u)));
--- a/test/graphql.paymentMethods.test.js
+++ b/test/graphql.paymentMethods.test.js
@@ -18,7 +18,6 @@ describe('graphql.paymentMethods.test.js', () => {
   beforeEach(async () => {
     await new Promise(res => setTimeout(res, 500));
     await utils.resetTestDB();
-    return;
   });
 
   beforeEach(() =>
--- a/test/graphql.paymentMethods.test.js
+++ b/test/graphql.paymentMethods.test.js
@@ -15,7 +15,11 @@ import { randEmail } from './stores';
 let host, admin, user, collective, paypalPaymentMethod;
 
 describe('graphql.paymentMethods.test.js', () => {
-  beforeEach(() => utils.resetTestDB());
+  beforeEach(async () => {
+    await new Promise(res => setTimeout(res, 500));
+    await utils.resetTestDB();
+    return;
+  });
 
   beforeEach(() =>
     models.User.createUserWithCollective({
--- a/config/test.json
+++ b/config/test.json
@@ -1,6 +1,10 @@
 {
   "database": {
-    "url": "postgres://opencollective@127.0.0.1:5432/opencollective_test"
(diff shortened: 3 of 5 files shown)


.circleci/config.yml
```diff
@@ -71,6 +71,7 @@ jobs:
       - run: mkdir -p $CIRCLE_ARTIFACTS $CIRCLE_TEST_REPORTS
       - run: npm ci
       - run: npm run db:setup
+      - run: npm run db:migrate
```
znarf: What is the impact of the migration? It was skipped on purpose, the model is supposedly already up to date, isn't it?

test/graphql.paymentMethods.test.js
```diff
@@ -15,7 +15,11 @@ import { randEmail } from './stores';
 let host, admin, user, collective, paypalPaymentMethod;
 
 describe('graphql.paymentMethods.test.js', () => {
-  beforeEach(() => utils.resetTestDB());
+  beforeEach(async () => {
+    await utils.resetTestDB();
+    await utils.resetTestModelDb('CollectiveHistory');
```
znarf: Why do we need to to that? What does that solve?

.circleci/config.yml
gmarcos87: Without migration sequelize.sync does not find the tables in postgres and displays the error before creating each one. That's why running the migrations avoids the error log in the tests.

znarf: Thank you, I'll dig deeper into it.

test/graphql.paymentMethods.test.js
gmarcos87: After many tests I came to the conclusion that it can be changed. Its original goal was to ensure the creation of the table "CollectiveHistory" which failed repeatedly.
I think the problem is due to an error in sequelize: sequelize.sync not completed when resolving (https://github.com/sequelize/sequelize/issues/8918) or by asynchronous events from which no resolution is expected (I saw some in the hooks) and that events/functions use the database that is being recreated by sequelize.sync at the same time.

If I add time before and after running sequelize.sync all problems disappear. The amount of time varies depending on the execution environment (memory, processor, Internet connection, etc.).

During the debugging process the only thing I noticed that improved the success rate was to enable the __logQueryParameters__ and __logging__ options. 😯 I suppose it slows down some processes and thereby giving others time to conclude gracefully.

In the case of the file ```graphql.paymentMethods.test``` (where the problem with "CollectiveHistory" occurs) I had to add time before running sequelize.sync, which gives me to understand that some postgres or sequelize process is announced as finished when it actually continues to block writing, or that some hook continues to work asynchronously while table deletion occurs.
_________________
Rather than worrying about the appearance of logs, what I would like to understand is the impact of these lags on production. I don't think there is any creation/droppingse of tables in production, so this is not a threat, but it can be remembered as something to revisit.

.circleci/config.yml
znarf: So, after investigation, the outcome is great, we do want to have the schema created before running the tests.

However, I don't think it's good idea and create the schema with `npm run db:migrate`, it's slow and ineffective. I think it's better to use `sequelize.sync` and add it in the setup script.

I explored this solution in: https://github.com/opencollective/opencollective-api/pull/2973

gmarcos87:
Great, I can do a rebase from ##2973, include that commit and remove the ones that aren't needed.

test/graphql.mutation.test.js
```diff
@@ -49,7 +49,11 @@ describe('Mutation Tests', () => {
 
   after(() => sandbox.restore());
 
-  beforeEach('reset db', () => utils.resetTestDB());
+  beforeEach('reset db', async () => {
+    await new Promise(res => setTimeout(res, 500));
```
znarf: Could we replace that by a try/catch/wait/retry? I'm afaid to slow down all tests by 500ms.

znarf:
Looks like the rebase was not properly done, my changes are appearing in the PR.

gmarcos87:
> Looks like the rebase was not properly done, my changes are appearing in the PR.

I had rebased to #2973. I removed the commit and updated it with master.

test/graphql.mutation.test.js
gmarcos87: try/catch do not fulfill the same function. The problem is that there are actions that continue to need the db even after the test case is over. This 500 millisecond wait gives them time to finish before the db is deleted and recreated. It doesn't happen all the time just in a couple of cases (it's only in graphql.mutation.test.js), the total time increases only in two or three seconds.

znarf: Ok, hadn't noticed it was only it's only in graphql.mutation.test.js.

I was thinking about.

```
try {
  await utils.resetTestDB()
} catch (e) {
  await new Promise(res => setTimeout(res, 500));
  await utils.resetTestDB()
}
```

gmarcos87: utils.resetTestDB() doesn't fail, what fails are the processes that were left in the background and over which we can't catch the errors.

PR #2968, opened 2019-11-27
Including IP address in sign in emails

CanerSezgin:
Issue: https://github.com/opencollective/opencollective/issues/2654

Screenshot from 2019-11-28 02-34-41

### DIFF
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -39,10 +39,9 @@ export const signin = (req, res, next) => {
     .then(u => u || models.User.createUserWithCollective(user))
     .then(u => {
       loginLink = u.generateLoginLink(redirect || '/', websiteUrl);
-      clientIP = req.headers['x-forwarded-for'] || req.connection.remoteAddress;
+      clientIP = req.ip;
       if (config.env === 'development') {
         logger.info(`Login Link: ${loginLink}`);
-        logger.info(`Client IP: ${clientIP}`);
       }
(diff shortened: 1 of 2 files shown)


server/controllers/users.js
```diff
@@ -34,14 +34,17 @@ export const exists = async (req, res) => {
 export const signin = (req, res, next) => {
   const { user, redirect, websiteUrl } = req.body;
   let loginLink;
+  let clientIP;
   return models.User.findOne({ where: { email: user.email.toLowerCase() } })
     .then(u => u || models.User.createUserWithCollective(user))
     .then(u => {
       loginLink = u.generateLoginLink(redirect || '/', websiteUrl);
+      clientIP = req.headers['x-forwarded-for'] || req.connection.remoteAddress;
```
znarf: Should be `req.ip` https://expressjs.com/en/api.html#req.ip

We expect it to be accurate, because we have it properly configured in express. See: https://github.com/opencollective/opencollective-api/blob/master/server/lib/express.js#L26

server/controllers/users.js
```diff
@@ -34,14 +34,17 @@ export const exists = async (req, res) => {
...
       if (config.env === 'development') {
         logger.info(`Login Link: ${loginLink}`);
+        logger.info(`Client IP: ${clientIP}`);
```
znarf: That should not be necessary. We log the `loginLink` not for troubleshooting but to use it to login! This is why it's relevant.

PR #2969, opened 2019-11-28
Tier: Deal with empty slugs

Betree:
Resolve https://sentry.io/organizations/open-collective/issues/1346473289/?project=1736806&query=is%3Aunresolved+assigned%3Ame&statsPeriod=14d

### DIFF
--- a/server/models/Tier.js
+++ b/server/models/Tier.js
@@ -7,7 +7,7 @@ import { Op } from 'sequelize';
 
 import CustomDataTypes from './DataTypes';
 import { maxInteger } from '../constants/math';
-import { capitalize, pluralize, days, formatCurrency, strip_tags } from '../lib/utils';
+import { capitalize, days, formatCurrency, strip_tags } from '../lib/utils';
 import { isSupportedVideoProvider, supportedVideoProviders } from '../lib/validators';
 
 const debug = debugLib('tier');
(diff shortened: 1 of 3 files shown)


server/models/Tier.js
```diff
@@ -248,7 +256,7 @@ export default function(Sequelize, DataTypes) {
         },
 
         title() {
-          return capitalize(pluralize(this.name));
+          return capitalize(this.name);
```
Betree: `capitalize` without passing a `count` has no effect https://github.com/opencollective/opencollective-api/blob/e5d82f6a75838bb6cbcb7c3d74561e690544fbd7/server/lib/utils.js#L404

server/models/Tier.js
```diff
@@ -48,10 +51,15 @@ export default function(Sequelize, DataTypes) {
...
         allowNull: false,
         set(name) {
+          this.setDataValue('name', name);
+
           if (!this.getDataValue('slug')) {
-            this.slug = name;
+            // Try to generate the slug from the name. If it fails, for example if tier
+            // name is '😵️' we gracefully fallback on tier type
+            const slugFromName = slugify(name);
+            const slug = slugFromName || slugify(this.type || 'TIER');
```
znarf: What happens if slug has a duplicate?

Betree: There's no unique constraint on tier's slugs (we use the id in the URL to differentiate) so it should be fine

PR #2970, opened 2019-11-28
Tier: Optimizations for availableQuantity

Betree:
- Use GraphQL loader.
- Never query DB if no `maxQuantity` is set. Will avoid 99% of the DB queries as this field is not set for most of them.

### DIFF
--- a/server/models/Tier.js
+++ b/server/models/Tier.js
@@ -319,6 +319,10 @@ export default function(Sequelize, DataTypes) {
 
   // TODO: Check for maxQuantityPerUser
   Tier.prototype.availableQuantity = function() {
+    if (!this.maxQuantity) {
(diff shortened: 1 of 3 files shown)


server/models/Tier.js
```diff
@@ -319,6 +319,10 @@ export default function(Sequelize, DataTypes) {
 
   // TODO: Check for maxQuantityPerUser
   Tier.prototype.availableQuantity = function() {
```
znarf: Can we somehow consider dropping this method?

Betree: It's used by `checkAvailableQuantity` below. We can eventually merge the two functions, but I think being able to do `tier.availableQuantity()` just to know how many are left is also nice.

PR #2987, opened 2019-12-03
Backyourstack: Fix broken test

flickz:
Resolve https://github.com/opencollective/opencollective/issues/2648

### DIFF
--- a/server/lib/backyourstack/dispatcher.js
+++ b/server/lib/backyourstack/dispatcher.js
@@ -176,6 +176,9 @@ export async function dispatch(order, subscription) {
       },
     });
   } catch (err) {
+    if (process.env.NODE_ENV !== 'production') {
+      throw err;
+    }
     console.log('>>>> Background dispatch failed');
     console.error(err);
   }
--- a/server/lib/backyourstack/dispatcher.js
+++ b/server/lib/backyourstack/dispatcher.js
@@ -176,9 +176,6 @@ export async function dispatch(order, subscription) {
       },
     });
   } catch (err) {
-    if (process.env.NODE_ENV !== 'production') {
-      throw err;
-    }
     console.log('>>>> Background dispatch failed');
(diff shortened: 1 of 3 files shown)


znarf:
The issue was asking to also cover a failure. Make sure it's properly handled. Can you check that?

flickz:
I'm still on it.

znarf:
I don't get the last change, I expected to see a new test covering that a failure is properly handled.

flickz:
I thought the goal was to be able to detect error that occurred during the process of dispatching, instead of having errors ignored while the tests continue to pass. 

This change will cause the test to fail whenever an error occurs.

flickz:
> I don't get the last change, I expected to see a new test covering that a failure is properly handled.

I understand now, I'll add additional test

test/graphql.createOrder.backyourstack.test.js
```diff
@@ -187,10 +194,21 @@ describe('createOrder', () => {
...
+  });
+
+  it('returns  error when it is not time for dispatch', async () => {
     // When the order is created
     const res = await utils.graphqlQuery(backyourstackDispatchOrder, { id: orderCreated.id }, xdamman);
-    // There should be no errors
-    res.errors && console.error(res.errors);
-    expect(res.errors).to.not.exist;
+    // There should be an error
+    expect(res.errors).to.exist;
```
znarf: Can you test the error message?

server/lib/backyourstack/dispatcher.js
```diff
@@ -176,6 +176,9 @@ export async function dispatch(order, subscription) {
       },
     });
   } catch (err) {
+    if (process.env.NODE_ENV !== 'production') {
+      throw err;
+    }
```
znarf: We should not have a specific behavior like that for production. Remove it or make it the default.

PR #2981, opened 2019-12-02
Upgrade dataloader-sequelize to v2

gmarcos87:
resolve https://github.com/opencollective/opencollective/issues/2094

PR #2989, opened 2019-12-03
Refactor loaders

Betree:
A proposal to refactor the giant `loaders.js` file:

- Have helpers in `server/graphql/loaders/helpers.ts`
- Create loaders in files like `server/graphql/loaders/comments.ts`
- Tie everything up in `server/graphql/loaders/index.js`

Moved `server/graphql/loaders/comments.ts` for demonstration purpose.

image

Also introduced a `sortResultsSimple` that does the same job as `sortResults` but without any magic implemented.


znarf:
@Betree uhm, this will heavily conflict with https://github.com/opencollective/opencollective-api/pull/2981

Betree:
I missed that in https://github.com/opencollective/opencollective-api/pull/2981, let's wait for it to be merged, the changes in this PR can easily be re-done if needed.

gmarcos87:
Hello, I think it's a great idea to split ```loaders.js```, it was a headache to work on such a big file and now it's hard to review all the changes.
If you think it's necessary I can adapt the pull-request again after you split the file.

Whatever you want to happen first is fine with me.

Betree:
@gmarcos87 Agree that the length of this file makes it hard to work with :wink: 

But let's prioritize https://github.com/opencollective/opencollective-api/pull/2981 over this one, it's a far more important change. I did this revamp cause I needed to add a loader and wanted to do it clean, but there's no rush.

Betree:
Rebased the PR upon changes from https://github.com/opencollective/opencollective-api/pull/2981.

It's very similar to the initial change, main difference is that loaders should now export functions instead of objects so that we can pass down the `request` and the `cache` (see `generateCommentsLoader`).

gmarcos87:
Hi @betree, So far everything look fine, you're going to create a file for each dataloader right? 
I recommend running the E2E test before the merge, in the case of #2981 detected a bug that api's tests do not catch.

PS: Is there any preference on the use of typescript over javascirpt? So I take it into account for future contributions.

Betree:
> So far everything look fine, you're going to create a file for each dataloader right?

@gmarcos87 I was thinking of grouping them by business unit (one file for collectives, one for expenses, one for contributors...etc).

> I recommend running the E2E test before the merge, in the case of #2981 detected a bug that api's tests do not catch.

@gmarcos87 Should be fine, I tested the endpoints manually.

> Is there any preference on the use of typescript over javascirpt? So I take it into account for future contributions.

As per https://github.com/opencollective/opencollective/blob/1d2b8016f64e79390724a82d4707943da46bbd8f/rfcs/006-typescript.md#L53-L55:

> We don't have a plan to migrate the existing codebase for now, nor to enforce it for all new
developments. However, we should encourage its usage for issues like https://github.com/opencollective/opencollective/issues/2274
where type checking and interfaces have a lot of benefits to offer.

For dataloaders I think it's interesting because we can easily specify the expected input and return types, see for example https://github.com/opencollective/opencollective-api/blob/bc396bf6753f9cbab493e1da3d075a91c5195ddb/server/graphql/loaders/conversation.ts#L4-L7

---

@znarf I'd like to remove `createDataLoaderWithOptions`. It was introduced in https://github.com/opencollective/opencollective-api/pull/1005 to cache the created `DataLoader` but because dataloaders are instantiated only once per request I don't understand what's its purpose. Do you have more insights?

znarf:
@Betree I'm not sure, feel free to remove it and ship that in an isolated deploy so we can monitor the impact in production performance

Betree:
@znarf Alright, I'm merging this PR as-is and will follow up with `createDataLoaderWithOptions` separately

PR #3007, opened 2019-12-06
chore(payExpense): ignore adaptive fees when payExpense is forceManual

flickz:
Following up https://github.com/opencollective/opencollective/issues/1976#issuecomment-562347697

### DIFF
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -312,7 +312,7 @@ export async function payExpense(remoteUser, args) {
 
   const fxrate = await getFxRate(expense.collective.currency, host.currency);
   const feesInHostCurrency = {};
-  if (paymentProviders[expense.payoutMethod] && !(expense.payoutMethod === 'paypal' && args.forceManual)) {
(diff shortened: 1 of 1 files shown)


server/graphql/v1/mutations/expenses.js
```diff
@@ -312,7 +312,7 @@ export async function payExpense(remoteUser, args) {
 
   const fxrate = await getFxRate(expense.collective.currency, host.currency);
   const feesInHostCurrency = {};
-  if (paymentProviders[expense.payoutMethod]) {
+  if (paymentProviders[expense.payoutMethod] && !(expense.payoutMethod === 'paypal' && args.forceManual)) {
```
znarf: This condition is not clear, it seems that the following code is highly related to Paypal. What about:

```if (paymentProviders[expense.payoutMethod] === 'paypal' && !args.forceManual)) {```

@Betree what do you think?

Comment (Betree) [reply server/graphql/v1/mutations/expenses.js]: Seems indeed that the code in the condition is for PayPal only.

`paymentProviders` holds the library, not the payment method name I think so the proper condition is probably something like:

```es6
if (expense.payoutMethod === 'paypal' && !args.forceManual)
```

PR #3008, opened 2019-12-06
Support for expiring credit cards notification

sbinlondon:
Relates opencollective/opencollective#2684

Also lots of cleaning up to do (logs, commits, etc) so please ignore that 😅 

Thank you @znarf for helping out with the mutation and Sequelize code. I was able to take that and run with it. I'll be sure to tell you all about the fun bug that had me stopped in my tracks for a while in the stand-up today... 😬 

As discussed this: 

* takes the id of the old payment method and the Stripe info from the new card
* finds the old payments method and its attached subs
* updates them with the new info
* and then deletes the old payment method so it no longer shows on the user's account

All feedback and questions and suggestions welcome! Still learning a lot about mutations/Sequelize.

### DIFF
--- a/cron/monthly/expiring-cards.js
+++ b/cron/monthly/expiring-cards.js
@@ -1,0 +1,87 @@
+#!/usr/bin/env node
+import '../../server/env';
+
+import logger from '../../server/lib/logger';
+import models from '../../server/models';
+import * as libPayments from '../../server/lib/payments';
+
+// Only run on the first of the month
+const today = new Date();
+const date = today.getDate();
+const month = today.getMonth() + 1;
+const year = today.getFullYear();
+
+if (process.env.NODE_ENV === 'production' && date !== 1 && !process.env.OFFCYCLE) {
+  console.log('NODE_ENV is production and today is not the first of month, script aborted!');
+  process.exit();
+}
+
+// link payment method id in Orders to payment method id in the payment method we're updating
+
+const fetchExpiringCreditCards = async (month, year) => {
+  const expiringCards = await models.PaymentMethod.findAll({
+    where: {
+      type: 'creditcard',
+      data: {
+        expMonth: month,
+        expYear: year,
+      },
+    },
+    include: [
+      {
+        model: models.Order,
+        where: { status: 'ACTIVE' },
+      },
+    ],
+  });
+
+  return expiringCards;
+};
+
+const run = async () => {
+  const cards = await fetchExpiringCreditCards(month, year);
+
+  for (const card of cards) {
+    try {
+      const { id, CollectiveId, name } = card;
+      const brand = card.data.brand || 'credit card';
+
+      const collective = await models.Collective.findByPk(CollectiveId);
+      const adminUsers = await collective.getAdminUsers();
+
+      for (const adminUser of adminUsers) {
+        const { slug } = collective;
+        const collectiveName = collective.name;
+        const { email } = adminUser;
+        const userId = adminUser.id;
+
+        const data = {
+          id,
+          brand,
+          name,
+          userId,
+          CollectiveId,
+          collectiveName,
+          slug,
+          email,
+        };
+
+        logger.info(
+          `Payment method ${data.id} for collective '${data.slug}' is expiring, sending an email to ${data.email}`,
+        );
+        if (!process.env.DRY_RUN) {
+          await libPayments.sendExpiringCreditCardUpdateEmail(data);
+        }
+      }
+    } catch (e) {
+      console.log(e);
+    }
+  }
+
+  logger.info('Done sending credit card update emails.');
+  process.exit();
+};
+
+if (require.main === module) {
+  run();
+}
--- a/server/graphql/v1/mutations/paymentMethods.js
+++ b/server/graphql/v1/mutations/paymentMethods.js
@@ -195,3 +195,36 @@ export async function updatePaymentMethod(args, remoteUser) {
 
   return models.PaymentMethod.update(pick(args, allowedFields), { where: { id: paymentMethod.id } });
 }
+
+/** Update payment method with given args */
+export async function replaceCreditCard(args, remoteUser) {
+  const oldPaymentMethod = await models.PaymentMethod.findByPk(args.id);
+  if (!oldPaymentMethod || !remoteUser || !remoteUser.isAdmin(oldPaymentMethod.CollectiveId)) {
+    throw PaymentMethodPermissionError;
+  }
+
+  const createArgs = {
+    ...pick(args, ['CollectiveId', 'name', 'token', 'data']),
+    service: 'stripe',
+    type: 'creditcard',
+  };
+
+  const newPaymentMethod = await createPaymentMethod(createArgs, remoteUser);
+
+  // Update orders (using Sequelize)
+  // first arg in new thing, second arg is old thing it's replacing
+  await models.Order.update(
+    { PaymentMethodId: newPaymentMethod.id },
+    {
+      where: {
+        PaymentMethodId: oldPaymentMethod.id,
+        status: 'ACTIVE',
+      },
+    },
+  );
+
+  // Delete or hide the old Payment Method (using Sequelize) - destroy instead of delete
+  await oldPaymentMethod.destroy();
+
+  return newPaymentMethod;
+}
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -469,3 +469,14 @@ export const sendReminderPendingOrderEmail = async order => {
     });
   }
 };
+
+export const sendExpiringCreditCardUpdateEmail = async data => {
+  data = {
+    ...data,
+    updateDetailsLink: `${config.host.website}/${data.slug}/paymentmethod/${data.id}/update`,
+  };
+
+  return emailLib.send('payment.creditcard.expiring', data.email, data, {
+    from: `${data.slug} <hello@${data.slug}.opencollective.com>`,
+  });
+};
--- a/templates/emails/payment.creditcard.expiring.hbs
+++ b/templates/emails/payment.creditcard.expiring.hbs
@@ -1,0 +1,46 @@
+Subject: Your {{brand}} ending in {{name}} is about to expire
+
+<style>
+  code {
+    display: inline-block;
+    border: 1px solid #ccc;
+    background: #ddd;
+    padding: 1px 3px;
+    margin: 0px 1px;
+  }
+</style>
+
+{{> header}}
+
+{{#if collectiveName}}
+<p>Dear {{collectiveName}},</p>
+{{else}}
+<p>Hello,</p>
+{{/if}}
+
+<p>
+  Your {{brand}} ending in {{name}} is about to expire. To continue your subscriptions without interruption, you should
+  update this card before the end of the month.
+</p>
+
+<p>Add a new credit card <a href="{{updateDetailsLink}}">here</a> to keep your subscriptions going.</p>
+
+<p>If you have any questions, please contact <a href="https://opencollective.com/support">Open Collective support</a>.
+</p>
+
+<p>Warmly,</p>
+
+<p>
+  – Team Open Collective
+</p>
+
+{{> footer}}
+
+<p
+  style="text-align:center;font-family:'Helvetica Neue',Helvetica,Arial,sans-serif;box-sizing:border-box;font-size:12px;color:#999;">
+  OpenCollective, Inc.<br />
+  EIN: 47-3896707<br />
+  340 S LEMON AVE #3717<br />
+  Walnut CA 91789<br />
+  USA
+</p>
--- a/server/graphql/v1/mutations.js
+++ b/server/graphql/v1/mutations.js
@@ -727,6 +727,20 @@ const mutations = {
       );
     },
   },
+  replaceCreditCard: {
+    type: PaymentMethodType,
+    description: 'Replace a payment method',
+    args: {
+      id: { type: new GraphQLNonNull(GraphQLInt) },
+      CollectiveId: { type: new GraphQLNonNull(GraphQLInt) },
+      name: { type: new GraphQLNonNull(GraphQLString) },
+      token: { type: new GraphQLNonNull(GraphQLString) },
+      data: { type: new GraphQLNonNull(StripeCreditCardDataInputType) },
+    },
+    resolve: async (_, args, req) => {
+      return paymentMethodsMutation.replaceCreditCard(args, req.remoteUser);
+    },
+  },
   createVirtualCards: {
     type: new GraphQLList(PaymentMethodType),
     args: {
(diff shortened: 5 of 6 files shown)


server/graphql/v1/mutations/paymentMethods.js
```diff
@@ -195,3 +201,37 @@ export async function updatePaymentMethod(args, remoteUser) {
...
+/** Update payment method with given args */
+export async function replaceCreditCard(args, remoteUser) {
+  const oldPaymentMethod = await models.PaymentMethod.findByPk(args.id);
+  if (!oldPaymentMethod || !remoteUser || !remoteUser.isAdmin(oldPaymentMethod.CollectiveId)) {
+    throw PaymentMethodPermissionError;
+  }
+
+  console.log('old payment method', oldPaymentMethod);
+
+  const { id, ...noOldIdArgs } = args;
```
znarf: Use lodash `omit` ?

sbinlondon: Ah nice, I did not know about that 👍

sbinlondon:
Cleaned up an rebased onto current master

cron/monthly/expiring-cards.js
```diff
@@ -0,0 +1,88 @@
...
+import models from '../../server/models';
+import * as libPayments from '../../server/lib/payments';
+
+// Only run on the first of the month
+const today = new Date();
+const date = today.getDate();
+const month = today.getMonth();
+const year = today.getFullYear();
+
+if (process.env.NODE_ENV === 'production' && date !== 1) {
```
znarf: Would be great to add the `OFFCYCLE`, similar to other monthly jobs.

server/graphql/v1/mutations/paymentMethods.js
```diff
@@ -195,3 +195,28 @@ export async function updatePaymentMethod(args, remoteUser) {
...
 }
+
+/** Update payment method with given args */
+export async function replaceCreditCard(args, remoteUser) {
+  const oldPaymentMethod = await models.PaymentMethod.findByPk(args.id);
+  if (!oldPaymentMethod || !remoteUser || !remoteUser.isAdmin(oldPaymentMethod.CollectiveId)) {
+    throw PaymentMethodPermissionError;
+  }
+
+  args = omit(args, ['id']);
```
znarf: It's not a problem today, but more as a code style, I would avoid rewriting `args` to a new value.

server/graphql/v1/mutations/paymentMethods.js
```diff
@@ -195,3 +195,28 @@ export async function updatePaymentMethod(args, remoteUser) {
...
+  // first arg in new thing, second arg is old thing it's replacing
+  await models.Order.update(
+    { PaymentMethodId: newPaymentMethod.id },
+    { where: { PaymentMethodId: oldPaymentMethod.id } },
+  );
+
+  // Delete or hide the old Payment Method (using Sequelize) - destroy instead of delete
+  await oldPaymentMethod.destroy();
+
+  // TODO: What should we return? The new Credit Card perhaps
```
znarf: We can now remove this comment.

server/lib/email.js
```diff
@@ -336,6 +336,7 @@ const generateEmailFromTemplateAndSend = async (template, recipient, data, optio
   }
 
   const notificationIsActive = await isNotificationActive(template, data);
+
```
znarf: It the file doesn't change, it should not appear as changed.

server/models/PaymentMethod.js
```diff
@@ -187,6 +187,7 @@ export default function(Sequelize, DataTypes) {
 
       hooks: {
         beforeCreate: instance => {
+          console.log('instance', instance);
```
znarf: Remove console.log and the 2 others.

templates/emails/payment.creditcard.expiring.hbs
```diff
@@ -0,0 +1,45 @@
...
+{{> toplogo}}
+
+{{#if firstName}}
+<p>Dear {{firstName}},</p>
+{{else}}
+<p>Hi!</p>
+{{/if}}
+
+<p>
+  A credit card used for your recurring subscriptions will expire at the end of this month.
```
znarf: As a reminder, final copy needs to be provided.

znarf:
@sbinlondon what's the difference between this PR and this one https://github.com/opencollective/opencollective-api/pull/3010? Can we close the other?

templates/emails/payment.creditcard.expiring.hbs
sbinlondon: Provided by design or would you like me to come up with something?

sbinlondon:
Thanks for the comments, will update tomorrow 😊

cron/monthly/expiring-cards.js
sbinlondon: Notes: OFFCYCLE allows us to fire the script whenever you want, not just the first of the month

server/graphql/v1/mutations/paymentMethods.js
sbinlondon: Note: see proposed solution in Slack

cron/monthly/expiring-cards.js
```diff
@@ -0,0 +1,88 @@
...
+const fetchExpiringCreditCards = async (month, year) => {
+  const expiringCards = await models.PaymentMethod.findAll({
+    where: {
+      type: 'creditcard',
+      data: {
+        expMonth: month,
+        expYear: year,
+      },
+    },
+  });
```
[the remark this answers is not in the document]
znarf: Technically, Orders with status=ACTIVE

templates/emails/payment.creditcard.expiring.hbs
sbinlondon: Note: will come up with something together today.

server/graphql/v1/mutations/paymentMethods.js
```diff
@@ -195,3 +195,28 @@ export async function updatePaymentMethod(args, remoteUser) {
...
+
+  args = omit(args, ['id']);
+
+  const newPaymentMethod = await createPaymentMethod({ ...args, service: 'stripe', type: 'creditcard' }, remoteUser);
+
+  // Update orders (using Sequelize)
+  // first arg in new thing, second arg is old thing it's replacing
+  await models.Order.update(
+    { PaymentMethodId: newPaymentMethod.id },
+    { where: { PaymentMethodId: oldPaymentMethod.id } },
```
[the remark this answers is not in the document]
sbinlondon: Added

cron/monthly/expiring-cards.js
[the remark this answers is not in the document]
sbinlondon: Added

cron/monthly/expiring-cards.js
```diff
@@ -0,0 +1,102 @@
...
+    } catch (e) {
+      console.log(e);
+    }
+  }
+
+  logger.info('Done sending credit card update emails.');
+  process.exit();
+};
+
+run();
```
kewitz: I suggest checking if this is not being accidentally imported with:
```js
if (require.main === module) {
  run();
}
```

server/lib/payments.js
```diff
@@ -468,3 +468,14 @@ export const sendReminderPendingOrderEmail = async order => {
...
 };
+
+export const sendExpiringCreditCardUpdateEmail = async data => {
+  data = {
+    ...data,
+    updateDetailsLink: `${config.host.website}/${data.slug}/paymentmethod/${data.id}/update`,
+  };
+
+  return emailLib.send('payment.creditcard.expiring', data.email, data, {
+    from: `${data.slug} <hello@${data.slug}.opencollective.com>`,
```
kewitz: I think we're deprecating the _hello@collective_ email, we should check this with @Betree.

znarf: In the end, I think we will need to write an utility function to compute this email address. In some cases, we will send `hello`, and in others it will be `no-reply`. We'll need to update everywhere it's used.

kewitz: #out-of-scope.

Betree: Yes we'll make the changes with https://github.com/opencollective/opencollective/issues/2847, it's the standard today so I'm ok with using it here.

cron/monthly/expiring-cards.js
```diff
@@ -0,0 +1,102 @@
...
+  for (const card of cards) {
+    try {
+      const { id, CollectiveId, name } = card;
+      const brand = card.data.brand || 'credit card';
+
+      const collective = await fetchCollectiveData(CollectiveId);
+      const user = await fetchUserData(collective.CreatedByUserId);
+
+      const { slug } = collective;
+      const { firstName, email } = user;
```
znarf: The only `name` we can rely on is `collective.name`

cron/monthly/expiring-cards.js
```diff
@@ -0,0 +1,102 @@
...
+const run = async () => {
+  const cards = await fetchExpiringCreditCards(month, year);
+
+  for (const card of cards) {
+    try {
+      const { id, CollectiveId, name } = card;
+      const brand = card.data.brand || 'credit card';
+
+      const collective = await fetchCollectiveData(CollectiveId);
+      const user = await fetchUserData(collective.CreatedByUserId);
```
Betree: Not a requested change but an alternative way to write this is to use `findByPk` (find by primary key)

```suggestion
      const user = await models.User.findByPk(collective.CreatedByUserId);
```

cron/monthly/expiring-cards.js
```diff
@@ -0,0 +1,102 @@
...
+        brand,
+        name,
+        userId,
+        CollectiveId,
+        slug,
+        firstName,
+        email,
+      };
+
+      console.log(
```
Betree: Maybe use the `logger` instead?

server/graphql/v1/mutations/paymentMethods.js
```diff
@@ -195,3 +195,32 @@ export async function updatePaymentMethod(args, remoteUser) {
...
 }
+
+/** Update payment method with given args */
+export async function replaceCreditCard(args, remoteUser) {
+  const oldPaymentMethod = await models.PaymentMethod.findByPk(args.id);
+  if (!oldPaymentMethod || !remoteUser || !remoteUser.isAdmin(oldPaymentMethod.CollectiveId)) {
+    throw PaymentMethodPermissionError;
+  }
+
+  const createArgs = { ...omit(args, ['id']), service: 'stripe', type: 'creditcard' };
```
Betree: As a general practice I think it's better to whitelist (`pick`) the fields you want rather than blacklist (`omit`). The reason is that any parameter added to the graphql endpoint will otherwise end up here without the implementer knowing it.

server/graphql/v1/mutations/paymentMethods.js
```diff
@@ -195,3 +195,32 @@ export async function updatePaymentMethod(args, remoteUser) {
...
+
+  const newPaymentMethod = await createPaymentMethod(createArgs, remoteUser);
```
Betree: We're not checking that `remoteUser` is an admin of `createArgs.CollectiveId`, so in the current state users can use this endpoint to create payment methods for collectives they don't own.

cron/monthly/expiring-cards.js
znarf: Agree. Simpler.

server/graphql/v1/mutations/paymentMethods.js
znarf: @Betree can you double check? `createPaymentMethod` should be taking care of that

cron/monthly/expiring-cards.js
```diff
@@ -0,0 +1,102 @@
...
+        `Payment method ${data.id} for collective '${data.slug}' is expiring, sending an email to ${data.email}`,
+      );
+      if (!process.env.DRY_RUN) {
+        await libPayments.sendExpiringCreditCardUpdateEmail(data);
```
znarf: As discussed last week, we should not rely on the CreatedByUserId. We should contact admins of the Collective the Payment Method is attached to.

server/graphql/v1/mutations/paymentMethods.js
Betree: Ah yes didn't see that. All good then!

server/lib/payments.js
sbinlondon: Marking resolved since this is being addressed in another PR 👍

server/graphql/v1/mutations/paymentMethods.js
sbinlondon: Updated to use pick instead of omit

cron/monthly/expiring-cards.js
```diff
@@ -0,0 +1,103 @@
...
+
+const run = async () => {
+  const cards = await fetchExpiringCreditCards(month, year);
+
+  for (const card of cards) {
+    try {
+      const { id, CollectiveId, name } = card;
+      const brand = card.data.brand || 'credit card';
+
+      const collective = await fetchCollectiveData(CollectiveId);
```
znarf: Maybe replace by `models.Collective.findByPk` here too.

cron/monthly/expiring-cards.js
```diff
@@ -0,0 +1,103 @@
...
+      const adminUsers = await collective.getAdminUsers();
+
+      for (const adminUser of adminUsers) {
+        try {
```
znarf: What kind of exceptions are you thinking about catching?

server/graphql/v1/mutations/paymentMethods.js
```diff
@@ -195,3 +195,36 @@ export async function updatePaymentMethod(args, remoteUser) {
...
+
+/** Update payment method with given args */
+export async function replaceCreditCard(args, remoteUser) {
+  const oldPaymentMethod = await models.PaymentMethod.findByPk(args.id);
+  if (!oldPaymentMethod || !remoteUser || !remoteUser.isAdmin(oldPaymentMethod.CollectiveId)) {
+    throw PaymentMethodPermissionError;
+  }
+
+  const createArgs = {
+    ...pick(args, ['CollectiveId', 'name', 'token', 'data']),
```
znarf: Great!

cron/monthly/expiring-cards.js
```diff
@@ -0,0 +1,89 @@
...
+  for (const card of cards) {
+    try {
+      const { id, CollectiveId, name } = card;
+      const brand = card.data.brand || 'credit card';
+
+      const collective = await models.Collective.findByPk(CollectiveId);
+      const adminUsers = await collective.getAdminUsers();
+
+      for (const adminUser of adminUsers) {
+        const user = await models.User.findByPk(adminUser.id);
```
znarf: Sorry, late comment! This is useless,`adminUser` and `user` are the same thing.

PR #3009, opened 2019-12-06
Auth: remove dead code

Betree:


znarf:
I like the sound of dead code being deleted.

PR #3026, opened 2019-12-10
Notify organization admins when funds are added

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2705

Screenshot 2019-12-10 at 12 44 19 PM


Betree:
We should probably mention the host that added the funds, because added funds are linked to a specific host an can only be used within this host.

PR #3028, opened 2019-12-10
Run e2e tests on API CI

znarf:
This will run E2E tests on all API pushes.

### DIFF
--- a/scripts/ci_checkout_frontend.sh
+++ b/scripts/ci_checkout_frontend.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+if [[ ! -z "${CIRCLE_BRANCH}" ]]; then
+  CI_BRANCH="${CIRCLE_BRANCH}"
+fi
+
+if [[ ! -z "${GITHUB_REF}" ]]; then
+  CI_BRANCH="${GITHUB_REF/refs\/heads\//}"
+fi
+
+CI_BRANCH=${CI_BRANCH:=master}
+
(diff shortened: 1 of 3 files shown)


scripts/ci_checkout_frontend.sh
```diff
@@ -0,0 +1,52 @@
...
+
+if [[ ! -z "${GITHUB_REF}" ]]; then
+  CI_BRANCH="${GITHUB_REF/refs\/heads\//}"
+fi
+
+CI_BRANCH=${CI_BRANCH:=master}
+
+TARBALL_URL="https://codeload.github.com/opencollective/opencollective-frontend/tar.gz/"
+echo "> Check ${TARBALL_URL}${CI_BRANCH}"
+if curl -s --head --request GET "${TARBALL_URL}${CI_BRANCH}" | grep "200" > /dev/null
```
Betree: Nice that we keep matching branches :+1:

znarf: I tried to implement that with a cleaner syntax using GitHub actions checkout. https://github.com/actions/checkout 

Was almost working but was ultimately failing because the first attempt was sending a CI failure. Will share a PR and open an issue on GitHub side to ask for that feature to be somehow supported.

znarf: Related links:
- https://github.com/opencollective/opencollective-api/pull/3031
- https://github.com/actions/checkout/issues/105

PR #3039, opened 2019-12-12
chore(collective): return getTotalAmountSpent in collective currency

flickz:
Resolves https://github.com/opencollective/opencollective/issues/2678 and https://github.com/opencollective/opencollective/issues/1408

### DIFF
--- a/test/server/graphql/v1/collective.test.js
+++ b/test/server/graphql/v1/collective.test.js
@@ -559,11 +559,15 @@ describe('graphql.collective.test.js', () => {
     beforeEach(async () => {
       // Given some users
       const currency = 'EUR';
+      const currencyInUSD = 'USD';
+
       const { piamancini } = await store.newUser('piamancini', { currency });
       const { xdamman } = await store.newUser('xdamman', { currency });
       const { user } = await store.newUser('alexandrasaveljeva', { currency });
       const { user0 } = await store.newUser('user0', { currency });
       const { user1 } = await store.newUser('user1', { currency });
+      const { olu } = await store.newUser('olu', { currencyInUSD });
+
       // And given a collective with a host that has stripe enabled
       const { hostAdmin, hostCollective, collective } = await store.newCollectiveWithHost(
         'brusselstogether collective',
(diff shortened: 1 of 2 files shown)


znarf:
Can we cover that with a test?

znarf:
Can we have a test that is testing the feature in its complexity? I would like to see a multi currency case.

test/server/graphql/v1/collective.test.js
```diff
@@ -735,6 +768,45 @@ describe('graphql.collective.test.js', () => {
...
+      for (const transaction of transactions) {
+        let amount = -parseInt(transaction.netAmountInCollectiveCurrency, 10);
+        if (transaction.currency !== collective.currency) {
+          const fxRate = await getFxRate(transaction.currency, collective.currency);
+          amount = fxRate * amount;
+        }
+        totalAmountSpent = amount + totalAmountSpent;
+      }
+
+      expect(collective.stats.totalAmountSpent).to.equal(totalAmountSpent);
```
znarf: Why not hardcode the totalAmountSpent?

flickz: Because a slight change in the value of exchange rate will cause the test to break.

znarf: Then I would say you need to mock the getFxRate.

In the current version, you're copying the code in the test, that's not the right approach,

test/server/graphql/v1/collective.test.js
```diff
@@ -735,6 +767,34 @@ describe('graphql.collective.test.js', () => {
...
+
+      const result = await utils.graphqlQuery(query, {
+        slug: 'olu',
+        type: 'DEBIT',
+      });
+
+      result.errors && console.error(result.errors);
+      expect(result.errors).to.not.exist;
+      const collective = result.data.Collective;
+      const totalAmountSpent = 1000 * 1.1 + 1000;
```
Betree: @flickz didn't you say [here](https://github.com/opencollective/opencollective-api/pull/3039/files#r357670268) that hard-coding wasn't the right approach because a slight change in the value of exchange rate will cause the test to break?

flickz: True but I realized the rate was hardcoded for test in https://github.com/opencollective/opencollective-api/blob/master/server/lib/currency.js#L76

PR #3043, opened 2019-12-12
Test parallelization

znarf:
Separated graphql tests from others, to split execution in half.

That was based on data sampled from a previous run:
Capture d’écran 2019-12-12 à 20 47 39

@Betree Should we proceed with that change?

### DIFF
--- a/config/test.json
+++ b/config/test.json
@@ -16,7 +16,7 @@
     }
   },
   "mailgun": {
-    "user": "test-mailgun-user",
-    "password": "test-mailgun-password"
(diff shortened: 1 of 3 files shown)


config/test.json
```diff
@@ -16,7 +16,7 @@
     }
   },
   "mailgun": {
-    "user": "test-mailgun-user",
-    "password": "test-mailgun-password"
+    "user": "",
+    "password": ""
```
Betree: Good catch for the empty strings

PR #3050, opened 2019-12-13
Remove clearbit

znarf:
There was no ticket for that but I saw an opportunity to simplify our code base and remove this dependency.

- it's not clear why we would need clearbit today, if we need the name of a user, we ask them
- for privacy reasons, we should not enrich a user profile without their consent
- seems it was implemented with the intent to get `website`, `twitterHandle` and `image`, but it's now disabled


server/lib/userlib.js
```diff
@@ -45,24 +13,4 @@ export default {
     }
     return null;
   },
-
-  updateUserInfoFromClearbit(user) {
-    // don't try to fetch user details if user hasn't provided a first name (i.e. if they wanted to remain incognito)
-    if (!user.email || !user.firstName) {
```
znarf: If no `firstName` we're doing nothing.

server/lib/userlib.js
```diff
@@ -45,24 +13,4 @@ export default {
...
-      return Promise.resolve();
-    }
-    return this.getUserData(user.email).then(userData => {
-      if (userData) {
-        user.firstName = user.firstName || userData.name.givenName;
-        user.lastName = user.lastName || userData.name.familyName;
```
znarf: We're only setting `lastName` if `firstName` was previously sent and `lastName` was empty. Doesn't make sense!

PR #3060, opened 2019-12-16
Members: Ensure one member is created for each tier

Betree:
Fix https://github.com/opencollective/opencollective/issues/2723
Require https://github.com/opencollective/opencollective-frontend/pull/3187 for tests update

### DIFF
--- a/server/paymentProviders/paypal/payment.js
+++ b/server/paymentProviders/paypal/payment.js
@@ -2,7 +2,6 @@ import config from 'config';
 import { get } from 'lodash';
 
 import models from '../../models';
-import roles from '../../constants/roles';
 
 import * as constants from '../../constants/transactions';
 import * as libpayments from '../../lib/payments';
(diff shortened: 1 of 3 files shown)


server/paymentProviders/paypal/payment.js
```diff
@@ -135,21 +135,12 @@ export async function createTransaction(order, paymentInfo) {
...
-  return order.collective.findOrAddUserWithRole(donorInfo, roles.BACKER, tierInfo, { order });
-}
-
 /** Process order in paypal and create transactions in our db */
 export async function processOrder(order) {
   const paymentInfo = await executePayment(order);
   logger.info('PayPal Payment');
   logger.info(paymentInfo);
   const transaction = await createTransaction(order, paymentInfo);
-  await addUserToCollective(order);
```
Betree: Removed this because the membership is created at the upper level, in `executeOrder` https://github.com/opencollective/opencollective-api/blob/47e075b43dd2997e944ee7f0da3f088ef4e62a0b/server/lib/payments.js#L302

PR #3096, opened 2019-12-23
Member invitation

Betree:
API changes for https://github.com/opencollective/opencollective-frontend/pull/3232


server/graphql/v1/CollectiveInterface.js
```diff
@@ -1077,6 +1077,7 @@ const CollectiveFields = () => {
         const query = {
           limit: args.limit,
           offset: args.offset,
+          order: [['id', 'ASC']],
```
Betree: This is to keep the order consistent when editing members

test/server/graphql/v1/createOrder.incognito.test.js
```diff
@@ -111,12 +111,18 @@ describe('server/graphql/v1/createOrder.incognito', () => {
...
+      const adminMember = collectiveData.members.find(m => m.member.id === adminUser.CollectiveId);
+      const backerMember = collectiveData.members.find(m => m.member.id === backerUser.CollectiveId);
+      const incognitoMember = collectiveData.members.find(m => m.member.id === incognitoCollective.id);
+      const hostMember = collectiveData.members.find(m => m.member.id === hostCollective.id);
+      expect(adminMember.member.createdByUser.firstName).to.equal('admin');
+      expect(backerMember.member.createdByUser.firstName).to.equal('backer');
+      expect(incognitoMember.member.slug).to.not.be.null;
+      expect(incognitoMember.member.createdByUser.firstName).to.be.null;
+      expect(incognitoMember.member.createdByUser.email).to.be.null;
+      expect(hostMember.member.createdByUser.firstName).to.equal('host');
```
Betree: Had to update these tests because members return order is now fixed. Used something more robust.

PR #3102, opened 2019-12-24
Add rate limiting library

Betree:


znarf:
Is it to replace `expressLimiter`?

Betree:
@znarf no it's just a simple wrapper around cache easily check and increment limits for actions (ie. number of gift cards created). I just comitted the two files using it, it should be clearer now.

PR #3119, opened 2020-01-02
Notify slack when we receive suspicious collective data (SPAM)

CanerSezgin:
Issue: https://github.com/opencollective/opencollective/issues/2750

@Betree 
Before going further, couple of more questions. 

1) I am planning to call **getSuspiciousKeywords** before creating collective. You can find example for `server/graphql/v1/mutations/collectives.js `-> **createCollective** function. Is this the correct position ? 

2) Should I call same function for these 3 (only) **createCollective**, **createCollectiveFromGithub**, **editCollective** If this is the case, I am going to create a function for this approach as well. (I guess again better place is spam.ts)
_PS: with longDescription if it is applicable_
```
const suspiciosKeywordCheckList = ['name', 'website', 'description'];
    suspiciosKeywordCheckList.forEach(prop => {
      const suspiciousKeywords = getSuspiciousKeywords(collectiveData[prop]);
      if (suspiciousKeywords.length) {
        console.log('Suspicious Keyword !!!', suspiciousKeywords, prop);
      }
});
```

3) According to the answer of the second question, we may think to create a class rather than some functions in spam.ts (I suppose, more spam business logic will be added later)

4) About slack notification, I don't have any access to **spam-reports** channel I guess. This is why I couldn't try but basically I suppose that I need to call only one function at the same place instead of console log in the example code block. If it is the case could you please send me function for sending notification. (or basically I can leave it to you guys)

### DIFF
--- a/server/lib/spam.ts
+++ b/server/lib/spam.ts
@@ -1,49 +1,131 @@
+import { clamp } from 'lodash';
+import config from 'config';
 import slackLib from '../lib/slack';
-import activities from '../constants/activities';
 
-interface Collective {
-  id: string;
-}
+// Watched content. Please keep everything in there lowercase.
+const ANALYZED_FIELDS: string[] = ['name', 'website', 'description', 'longDescription'];
+const SPAM_KEYWORDS: string[] = ['keto', 'porn', 'pills', 'male enhancement'];
+const BLACKLISTED_DOMAINS = [
+  'supplementslove.com',
+  'supplementspeak.com',
+  'fitcareketo.com',
+  'supplementgear.com',
+  'totalketopills.com',
+  'diets2try.com',
+  'dragonsdenketo.com',
+  'ketovatrudiet.info',
+  'completefoods.co',
+  'creativehealthcart.com',
+  'onnitsupplements.com',
+  'topcbdoilhub.com',
+  'nutrifitweb.com',
+  'healthygossips.com',
+  'healthtalkrev.com',
+  'allsupplementshop.com',
+  'offer4cart.com',
+  'hulkpills.com',
+  'supplementgo.com',
+  'fitnessmegamart.com',
+  'healthyaustralia.com.au',
+  'healthyslimdiet.com',
+  'totaldiet4you.com',
+  'trypurenutrition.com',
+  'reviewsbox.org',
+];
 
-const collectiveCheckList: string[] = ['name', 'website', 'description', 'longDescription'];
-const blackList: string[] = ['keto', 'porn', 'pills'];
+/** Return type when running a spam analysis */
+type SpamAnalysisResult = {
+  /** A score between 0 and 1. 0=NotSpam, 1=IsSpamForSure */
+  score: number;
+  /** Detected spam keywords */
+  keywords: string[];
+  /** Detected blacklisted domains */
+  domains: string[];
+};
 
 /**
- * Returns SuspiciousKeywords
- * @param {String} content
+ * Returns suspicious keywords found in content
  */
 const getSuspiciousKeywords = (content: string): string[] => {
-  const suspiciousWords = [];
-  blackList.forEach(keyword => {
-    if (content.includes(keyword)) {
-      suspiciousWords.push(keyword);
-    }
-  });
-  return suspiciousWords;
-}
-
-export const collectiveSpamCheck = (collective: Collective): null => {
-  const result = {
-    collective,
-    warnings: {},
-  };
-  collectiveCheckList.forEach(prop => {
-    const suspiciousKeywords = getSuspiciousKeywords(collective[prop] || '');
-    if (suspiciousKeywords.length) {
-      result.warnings[prop] = suspiciousKeywords;
-    }
-  });
-
-  // Send Notification on Slack
-  // TODO: Add WEBHOOK
-  if (Object.keys(result).length) {
-      slackLib.postActivityOnPublicChannel(
-        {
-          type: activities.COLLECTIVE_BADWORD_DETECTED,
-          data: result,
-        },
-        'WEBHOOK',
-      );
+  if (!content) {
+    return [];
   }
-  return null;
-}
+
+  return SPAM_KEYWORDS.reduce((result, keyword) => {
+    if (content.toLowerCase().includes(keyword)) {
+      result.push(keyword);
+    }
+    return result;
+  }, []);
+};
+/**
+ *
+ * Returns blacklisted domains found in content
+ */
+const getBlacklistedDomains = (content: string): string[] => {
+  if (!content) {
+    return [];
+  }
+
+  return BLACKLISTED_DOMAINS.reduce((result, domain) => {
+    if (content.toLowerCase().includes(domain)) {
+      result.push(domain);
+    }
+    return result;
+  }, []);
+};
+
+/**
+ * Checks the values for this collective to try to determinate if it's a spammy profile.
+ */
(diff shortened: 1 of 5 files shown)


Betree:
> 1. I am planning to call **getSuspiciousKeywords** before creating collective. You can find example for `server/graphql/v1/mutations/collectives.js `-> **createCollective** function. Is this the correct position ?
> 2) Should I call same function for these 3 (only) **createCollective**, **createCollectiveFromGithub**, **editCollective** If this is the case, I am going to create a function for this approach as well. (I guess again better place is spam.ts)
_PS: with longDescription if it is applicable_

To make sure that we cover all the paths, I'd say that the best place to implement would be in the `afterCreate`: https://github.com/opencollective/opencollective-api/blob/edb538ea1161d7033600dacfd02979673a3fa350/server/models/Collective.js#L637

> 3. According to the answer of the second question, we may think to create a class rather than some functions in spam.ts (I suppose, more spam business logic will be added later)

This is up to you, this repo usually tends to use mostly functional code but feel free to use a class if you think that it is the best pattern to implement this.

> 4. About slack notification, I don't have any access to **spam-reports** channel I guess. This is why I couldn't try but basically I suppose that I need to call only one function at the same place instead of console log in the example code block. If it is the case could you please send me function for sending notification. (or basically I can leave it to you guys)

You can eventually try with this (public) channel: https://opencollective.slack.com/archives/C1T3E0T5F

Betree:
You'll need to rebase on master to get the tests working (see https://github.com/opencollective/opencollective-api/pull/3123). Sorry for the inconvenience!

CanerSezgin:
> > 1. Should I call same function for these 3 (only) **createCollective**, **createCollectiveFromGithub**, **editCollective** If this is the case, I am going to create a function for this approach as well. (I guess again better place is spam.ts)
> >    _PS: with longDescription if it is applicable_
> 
> To make sure that we cover all the paths, I'd say that the best place to implement would be in the `afterCreate`:
> 
> https://github.com/opencollective/opencollective-api/blob/edb538ea1161d7033600dacfd02979673a3fa350/server/models/Collective.js#L637
> 

I am just calling spamController as you suggest in **afterCreate**, I put it inside if check but it might be redundant. I don't know what kind of instance we have here but I saw there is a similar logic above this is why created like this

```
if (instance.type === 'COLLECTIVE' || instance.type === 'EVENT') {
            await models.PaymentMethod.create({
                         ...Current Codes...
            });
}

if (instance.type === 'COLLECTIVE') {
      const spamCheck = spamController.collectiveCheck(instance);
}
```

With this approach, when I created collective from frontend, it works fine for  **createCollective**, couldn't try for **createCollectiveFromGithub** but it will be ok most prob as you suggest. 
However when I edit (directly from frontend), I didn't see **afterCreate** has been called. I guess we need to call spamController somewhere else to use with edit. 

Waiting your comments on this @Betree 
Meanwhile, couldn't try to send these messages on slack, will try to implement this. 

For now, I am returning this kind of array as a result, it shows which part contains bannedWords
```
{ 
  collectiveId: 10894,
  warnings: { 
     name: [ 'keto', 'pills' ], 
     description: [ 'keto' ] 
  } 
}
```

CanerSezgin:
Another question, what do you suggest for this error. I am using it in another file but TS wants me to use in the same file that class is defined.
`[error]server/lib/spam.ts(43,13): error TS6133: 'collectiveCheck' is declared but its value is never read.`

Betree:
> However when I edit (directly from frontend), I didn't see afterCreate has been called. I guess we need to call spamController somewhere else to use with edit.

`afterCreate` is only called after creation. You can also try `afterUpdate` or `afterSave` hooks: https://sequelize.org/master/manual/hooks.html

> Another question, what do you suggest for this error. I am using it in another file but TS wants me to use in the same file that class is defined.

It should no happen if you export the class/function

CanerSezgin:
Hello @Betree

When I merged **master** into **branch** I could get rid of the other changes (when I run rebase, it hadn't worked as you remember) It is kinda ready now and I think you can review all. 

I left a TODO inside code, I just need you to try on slack (I couldn't send request on dev mode) and also you need to set up correct webhook for it. 

Example messages that I create is below. I have changed little bit of the flow. Such as if there is a bad word more than one property (name, description, website etc) I am returning a message it says this property includes these bad words. (I use **\n** inside string but not sure how slack is handling this)

>    Field "name" for <qketoed lsfpills |http://localhost:3000/qketoed-lsfpills> was set to a suspicious value/s: [ keto, pills ]
>     Field "website" for <qketoed lsfpills |http://localhost:3000/qketoed-lsfpills> was set to a suspicious value/s: [ keto, pills ]
>     Field "description" for <qketoed lsfpills |http://localhost:3000/qketoed-lsfpills> was set to a suspicious value/s: [ pills ]

Betree:
Actually there are changes that you can make before I move forward, adding a review now

server/models/Collective.js
```diff
@@ -653,8 +654,17 @@ export default function(Sequelize, DataTypes) {
             });
           }
 
+          if (instance.type === 'COLLECTIVE') {
```
Betree: We should actually run that on all types of collectives (events, users, orgs...etc)

server/models/Collective.js
```diff
@@ -653,8 +654,17 @@ export default function(Sequelize, DataTypes) {
...
+            spamController.collectiveCheck(instance);
+          }
+
           return null;
         },
+        afterUpdate: async instance => {
+          if (instance.type === 'COLLECTIVE') {
```
Betree: Same: we should actually run that on all types of collectives (events, users, orgs...etc)

server/lib/spam.ts
```diff
@@ -0,0 +1,70 @@
...
+      }
+    });
+    return suspiciousWords;
+  }
+
+  /**
+   *
+   * @param collective {Collective}
+   */
+  private collectiveCheck(collective: Collective): { collective: Collective; warnings: object } {
```
Betree: Good idea to return a map for `warnings`, it's a good way to bundle all of them.

Why returning the collective? You pass it as a parameter so you already have it when you call this function.

server/lib/spam.ts
```diff
@@ -0,0 +1,70 @@
...
+    // TODO: Add WEBHOOK
+    if (Object.keys(result).length) {
+        slackLib.postActivityOnPublicChannel(
+          {
+            type: activities.COLLECTIVE_BADWORD_DETECTED,
+            data: result,
+          },
+          'WEBHOOK',
+        );
+    }
```
Betree: If using an activity then you should not use `slackLib` directly, you should just create the activity

server/lib/spam.ts
CanerSezgin: it's my bad, the structure is a bit different than this. Then, I moved the whole logic inside spam.js and call slack function inside. Now don't need to return anything. Resolved.

server/lib/spam.ts
CanerSezgin: Couldn't get how to do this. You can send me an example or basically take it from here, again it is slack related (you need to add webhook etc in anyway). Not sure which functions/files should I use exactly.

CanerSezgin:
@Betree except from this one https://github.com/opencollective/opencollective-api/pull/3119#discussion_r364689056 all requests have been sorted out. Thanks for reviewing.

Betree:
I'll take it from there, thanks @CanerSezgin :+1:

PR #3076, opened 2019-12-19
Host Plans

znarf:
Implement: https://github.com/opencollective/opencollective/issues/2662

PR #3132, opened 2020-01-06
WIP: Wrapping up Host plans

kewitz:
Contributes to: https://github.com/opencollective/opencollective-api/pull/3076

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -990,9 +990,8 @@ export async function addFundsToCollective(order, remoteUser) {
   }
 
   // Check limits
-  const hostPlan = host.getPlan();
-  const hostTotalAddedFunds = await host.getTotalAddedFunds();
-  if (hostPlan.addFundsLimit && hostPlan.addFundsLimit <= hostTotalAddedFunds) {
+  const hostPlan = await host.getPlan();
+  if (hostPlan.addedFundsLimit && hostPlan.addedFundsLimit <= hostPlan.addedFunds) {
     throw new errors.PlanLimit({
       message:
         'The limit of "Added Funds" for the host has been reached. Please contact support@opencollective.com if you think this is an error.',
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -641,14 +641,13 @@ export async function completePledge(remoteUser, order) {
  * Cancel user's subscription. We don't prevent this event is user is limited (canUseFeature -> false)
  * because we don't want to prevent users from cancelling their subscriptions.
  */
-export function cancelSubscription(remoteUser, orderId) {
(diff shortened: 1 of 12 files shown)


server/models/Collective.js
```diff
@@ -868,6 +868,23 @@ export default function(Sequelize, DataTypes) {
...
 
+  /**
+   * If the collective is a host, it needs to remove existing hosted collectives before
+   * deactivating it as a host.
+   */
+  Collective.prototype.deactivateAsHost = async function() {
+    const hostedCollectivesCount = await this.getHostedCollectivesCount();
+    if (hostedCollectivesCount >= 1) {
+      throw new Error(
+        `You cannot deactivate your collective while still hosting ${hostedCollectivesCount} other collectives. Please contact support.`,
```
znarf: The copy is unclear, it's not deactivating the collective.

Please contact support: support@opencollective.com.

server/graphql/v1/mutations/orders.js
```diff
@@ -641,65 +642,61 @@ export async function completePledge(remoteUser, order) {
...
+    UserId: order.CreatedByUserId,
+    data: {
+      subscription: order.Subscription,
+      collective: order.collective.minimal,
+      user: remoteUser.minimal,
+      fromCollective: order.fromCollective.minimal,
+    },
+  });
+
+  if (order.Tier && order.collective.slug === PLANS_COLLECTIVE_SLUG) {
```
znarf: Maybe add a code comment explaining why we're doing that.

server/graphql/v1/mutations/orders.js
```diff
@@ -641,65 +642,61 @@ export async function completePledge(remoteUser, order) {
...
+    const plan = findKey(plans, { slug: order.Tier.slug });
+    if (plan && Boolean(order.fromCollective.plan)) {
+      await order.fromCollective.update({ plan: null });
```
znarf: We have to think about people upgrading from a plan to another, the best way will be to subscribe to the new plan first, then cancel the old one.

Maybe we should not do that synchronously in the end. Daily cron job and alert us?

server/graphql/v1/mutations/orders.js
kewitz: Deprecate this in favor of a worker that will properly process downgrades once a month.

server/lib/plans.ts
```diff
@@ -0,0 +1,21 @@
+import { findKey, get } from 'lodash';
+
+import plans, { PLANS_COLLECTIVE_SLUG } from '../constants/plans';
+
+const isSubscribeOrUpgrade = (newPlan: string, oldPlan?: string | null): boolean => {
+  return !oldPlan ? true : get(plans, `${newPlan}.tier`) > get(plans, `${oldPlan}.tier`);
```
znarf: Maybe `level` would be better so we don't make mistake about what `tier` is.

PR #3149, opened 2020-01-08
Fix wrong receipt links in emails

Betree:
Fix https://github.com/opencollective/opencollective/issues/2783 by linking to the collective transactions page rather than the invoices service.

Also deleted `templates/emails/ticket.confirmed.sustainoss.hbs ` because its integration was removed in https://github.com/opencollective/opencollective-api/pull/1876.


Betree:
Moved to https://github.com/opencollective/opencollective-api/pull/3162.
Not sure why, but CI wouldn't pass on this PR - even with an empty commit.

PR #3151, opened 2020-01-08
WIP: Downgrade/Cancellation cron worker

kewitz:
Deals with simple cases of pending cancellations and downgrades.

### DIFF
--- a/cron/monthly/verify-plans.js
+++ b/cron/monthly/verify-plans.js
@@ -1,10 +1,6 @@
 #!/usr/bin/env node
 import '../../server/env';
 
-/**
- * Makes sure all collectives.plan are up-to-date, downgrading and cancelling plans if needed.
- */
-
 // Only run on the first of the month
 const today = new Date();
 if (process.env.NODE_ENV === 'production' && today.getDate() !== 1) {
@@ -12,28 +8,22 @@ if (process.env.NODE_ENV === 'production' && today.getDate() !== 1) {
   process.exit();
 }
 
-process.env.PORT = 3066;
-
 import { Op } from 'sequelize';
-import { compact, findKey, get, groupBy, map, values, pipe } from 'lodash/fp';
+import { findKey, groupBy } from 'lodash';
 import debugLib from 'debug';
 
 import models from '../../server/models';
-import { promiseSeq } from '../../server/lib/utils';
 import plans, { PLANS_COLLECTIVE_SLUG } from '../../server/constants/plans';
 import orderStatus from '../../server/constants/order_status';
 import emailLib from '../../server/lib/email';
 
 const debug = debugLib('verify-plans');
 
-const d = new Date();
-d.setMonth(d.getMonth() - 1);
-
-const START_DATE = new Date(d.getFullYear(), d.getMonth(), 1);
-const END_DATE = new Date(d.getFullYear(), d.getMonth() + 1, 1);
 const REPORT_EMAIL = 'ops@opencollective.com';
 const BATCH_SIZE = 10;
-const EXISTING_PLANS_SLUGS = pipe(values, map(get('slug')), compact)(plans);
+const EXISTING_PLANS_SLUGS = Object.values(plans)
+  .map(p => p.slug)
+  .filter(Boolean);
 
 const LEVELS = {
   LEGACY: 'LEGACY',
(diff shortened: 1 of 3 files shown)


test/cron/monthly/verify-plans.test.js
```diff
@@ -0,0 +1,76 @@
...
+import { fakeCollective, fakeTier, multiple } from '../../test-helpers/fake-data';
+import * as utils from '../../utils';
+import models from '../../../server/models';
+import orderStatus from '../../../server/constants/order_status';
+import plans, { PLANS_COLLECTIVE_SLUG } from '../../../server/constants/plans';
+import { run as verifyPlans } from '../../../cron/monthly/verify-plans';
+
+const arrayToJSON = map(a => a.toJSON());
+
+const asserCollectionIsUnchanged = async col => {
```
znarf: ```asserCollectionIsUnchanged```

cron/monthly/verify-plans.js
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  // Time we spent running the whole script
+  const now = new Date();
+  const end = now - START_DATE;
+  if (body.length === 0) {
+    const text = `No subscriptions pending charges found\n\nTotal time taken: ${end}ms`;
+    const subject = `Ø Monthly Plan Verification Report - ${now.toLocaleDateString()}`;
+    emailLib.sendMessage(REPORT_EMAIL, subject, '', { text });
+  } else if (body.lenght > 0) {
```
znarf: `body.length`

cron/monthly/verify-plans.js
```diff
@@ -0,0 +1,148 @@
...
+ */
+
+// Only run on the first of the month
+const today = new Date();
+if (process.env.NODE_ENV === 'production' && today.getDate() !== 1) {
+  console.log('NODE_ENV is production and today is not the first of month, script aborted!');
+  process.exit();
+}
+
+process.env.PORT = 3066;
```
znarf: What's that?

cron/monthly/verify-plans.js
```diff
@@ -0,0 +1,148 @@
...
+
+import { Op } from 'sequelize';
+import { compact, findKey, get, groupBy, map, values, pipe } from 'lodash/fp';
```
znarf: Why not just use `lodash` ?

cron/monthly/verify-plans.js
```diff
@@ -0,0 +1,148 @@
...
+  EXCEPTION: 'EXCEPTION',
+};
+
+export async function run(options = {}) {
+  const collectives = await models.Collective.findAll({ where: { plan: { [Op.ne]: null } } });
+  debug(`There is/are ${collectives.length} subscribed to our plans...`);
+
+  const info = [];
+
+  await promiseSeq(
```
znarf: I don't think the charge-subscription script is a good example to follow in best practices.

I would simplify that and just loop over collectives without concurrency.

```
for (const collective of collectives) {
 ... 
}
```

cron/monthly/verify-plans.js
```diff
@@ -0,0 +1,148 @@
...
+import orderStatus from '../../server/constants/order_status';
+import emailLib from '../../server/lib/email';
+
+const debug = debugLib('verify-plans');
+
+const d = new Date();
+d.setMonth(d.getMonth() - 1);
+
+const START_DATE = new Date(d.getFullYear(), d.getMonth(), 1);
+const END_DATE = new Date(d.getFullYear(), d.getMonth() + 1, 1);
```
znarf: `START_DATE` and `END_DATE` is not really used, are they useful?

`START_DATE` is used for time measurement later but I would just drop that, this script is gonna be fast.

cron/monthly/verify-plans.js
```diff
@@ -0,0 +1,148 @@
...
+      debug(`Processing collective #${c.id}...`);
+      // Custom or legacy plans, we're ignoring this because this was manually set.
+      if (!plans[c.plan].slug) {
+        return info.push({
+          level: LEVELS.LEGACY,
+          message: `${c.slug} is using legacy plan, ignoring.`,
+        });
+      }
+
+      const [lastOrder] = await models.Order.findAll({
```
znarf: `findOne`

cron/monthly/verify-plans.js
```diff
@@ -0,0 +1,148 @@
...
+  if (EXCEPTION) {
+    subjectIcon = '🚨';
+    body = [`🚨 Exceptions:\n`, ...EXCEPTION.map(info => `${info.message}\n`), `\n`, ...body];
+  }
+
+  // Time we spent running the whole script
+  const now = new Date();
+  const end = now - START_DATE;
+  if (body.length === 0) {
+    const text = `No subscriptions pending charges found\n\nTotal time taken: ${end}ms`;
```
znarf: Review text.

cron/monthly/verify-plans.js
```diff
@@ -0,0 +1,148 @@
...
+const debug = debugLib('verify-plans');
+
+const d = new Date();
+d.setMonth(d.getMonth() - 1);
+
+const START_DATE = new Date(d.getFullYear(), d.getMonth(), 1);
+const END_DATE = new Date(d.getFullYear(), d.getMonth() + 1, 1);
+const REPORT_EMAIL = 'ops@opencollective.com';
+const BATCH_SIZE = 10;
+const EXISTING_PLANS_SLUGS = pipe(values, map(get('slug')), compact)(plans);
```
znarf: Difficult to follow if not familiar with fp, this can be simpler such as, `Object.values(plans).map(p => p.slug).filter(Boolean)`

cron/monthly/verify-plans.js
```diff
@@ -0,0 +1,148 @@
...
+      }
+
+      const [lastOrder] = await models.Order.findAll({
+        include: [
+          { model: models.Collective, as: 'collective', where: { slug: PLANS_COLLECTIVE_SLUG } },
+          { model: models.Collective, as: 'fromCollective', where: { id: c.id } },
+          { model: models.Subscription, as: 'Subscription' },
+          { model: models.Tier, as: 'Tier', where: { slug: { [Op.in]: EXISTING_PLANS_SLUGS } } },
+        ],
+        limit: 1,
```
znarf: limit will not be useful if using findOne

cron/monthly/verify-plans.js
```diff
@@ -0,0 +1,148 @@
...
+        order: [['updatedAt', 'DESC']],
+      });
+
+      const lastOrderPlan = findKey({ slug: lastOrder.Tier.slug }, plans);
```
znarf: Is it because it's the fp version of findKey? Doesn't match the documentation. https://lodash.com/docs/4.17.15#findKey

```findKey(users, { 'age': 1, 'active': true });```

File: cron/monthly/verify-plans.js (hunk 3 above)
Comment (kewitz) [reply cron/monthly/verify-plans.js]: Not sure, it is included in every monthly cron worker, I thought it was a dependency related hack.

File: cron/monthly/verify-plans.js (hunk 11 above)
Comment (kewitz) [reply cron/monthly/verify-plans.js]: Yes, in the functional programming version you first create functions and then invoke with the data.
I'll replace this with pure JS like suggested above.

File: cron/monthly/verify-plans.js (hunk 4 above)
Comment (kewitz) [reply cron/monthly/verify-plans.js]: I used this for composing the functions to eval `EXISTING_PLANS_SLUGS`.
Returning this to normal `lodash` after replacing that with plain JS.

File: cron/monthly/verify-plans.js [hunk 12]
```diff
@@ -0,0 +1,142 @@
...
+        message,
+      });
+    }
+  }
+
+  return info;
+}
+
+const script = async () => {
+  const info = await run({ batch: BATCH_SIZE });
```
znarf: BATCH_SIZE not needed anymore

PR #3153, opened 2020-01-08
Refactor subcollectives: Use Members rather than ParentCollectiveId

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2633

### DIFF
--- a/server/models/Member.js
+++ b/server/models/Member.js
@@ -59,19 +59,8 @@ export default function(Sequelize, DataTypes) {
         defaultValue: 'member',
         validate: {
           isIn: {
-            args: [
-              [
-                roles.HOST,
-                roles.ADMIN,
-                roles.MEMBER,
-                roles.BACKER,
-                roles.CONTRIBUTOR,
-                roles.ATTENDEE,
-                roles.FOLLOWER,
-                roles.FUNDRAISER,
-              ],
-            ],
-            msg: 'Must be host, admin, member, backer, contributor, attendee, fundraiser or follower',
+            args: [Object.values(roles)],
+            msg: `Must be one of ${Object.values(roles)}`,
           },
         },
       },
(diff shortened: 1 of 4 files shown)


server/models/Member.js
```diff
@@ -59,19 +59,8 @@ export default function(Sequelize, DataTypes) {
...
-                roles.BACKER,
-                roles.CONTRIBUTOR,
-                roles.ATTENDEE,
-                roles.FOLLOWER,
-                roles.FUNDRAISER,
-              ],
-            ],
-            msg: 'Must be host, admin, member, backer, contributor, attendee, fundraiser or follower',
+            args: [Object.values(roles)],
+            msg: `Must be one of ${Object.values(roles)}`,
```
znarf: Nice!

Betree:
Currently the terminology to show them in the frontend is `Member collective`:

image

This comment is interesting to bring in this discussion to better understand the need:

- https://github.com/opencollective/opencollective/issues/2633#issuecomment-555911039

> We basically have two needs for sub-collectives:
> 
>   1. different currencies
>   2. specific causes (like collecting money for speaker training, collecting money for conference scholarships). this way we want to offer donors more control over how their donations may be used.
> 
> I would add that in our case we really do want one collective to be considered the parent and others as the sub-collectives (especially for the cause specific collectives).

To that we need to add that this feature also replaces the concept of "Super Collective" that used tags to group projects that are part of an organization, that we refer to as `These x collectives are part of our organization` in the frontend.

So whatever we choose, the term needs to acknowledge that the relation is directional (it's not just connected, one of the collective is "above" the other).

Based on that, I can think of these other alternatives to sub-collective:
- Incorporated collective
- Included collective
- Integrated collective
- Embedded collective
- Member collective (brings confusion with the `MEMBER` role)

znarf:
Ok, let's move forward with sub-collective.

PR #3169, opened 2020-01-10
CollectiveInterface: Use parentCollective for event's isApproved

Betree:
That will make it easier to know if it's possible to contribute to an event in the frontend.

This will also set `isApproved` to `false` if no host has been set.


znarf:
@Betree Thank you. The `patch` config needs to be tweaked too it seems.

Betree:
@znarf 

> Rebased.
> 
> @Betree Should we we move isApproved() to the model?

I hesitated, went lazy because I don't want to load the parent in the model (to make sure we keep using the GraphQL loaders). I made the change now, using a synchronous version that don't accept being called with collectives.

PR #3174, opened 2020-01-10
Mocha Config moved from mocha.opts to mocharc.json

CanerSezgin:
Issue: https://github.com/opencollective/opencollective/issues/2787

I have put config into `test/mocharc.json` and updated package.json test script with `--config test/.mocharc.json"` 

I think this is the best and cleanest solution but this also can be added into directly package.json 
or these extensions (**.js .jsonc .yml .json**) can be used.

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,0 +1,8 @@
+{
+  "timeout": 20000,
+  "extension": ["js", "ts"],
+  "require": ["./test/setup-babel.js"],
+  "file": ["./test/setup.js", "./server/env.js"],
+  "recursive": true,
+  "exit": true
+}
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,7 +1,0 @@
---timeout 20000
---extension js,ts
---require ./test/setup-babel.js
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -187,7 +187,7 @@
     "script": "babel-node --extensions .js,.ts $1",
     "sequelize": "babel-node --extensions .js,.ts -- ./node_modules/.bin/sequelize --config config/sequelize-cli.js",
     "start": "node ./dist/index.js",
-    "test": "TZ=UTC nyc --reporter=lcov mocha",
+    "test": "TZ=UTC nyc --reporter=lcov mocha --config test/.mocharc.json",
     "test:clean": "rm -rf .nyc_output coverage",
     "test:coverage": "codecov",
     "test:graphql": "TZ=UTC mocha test/server/graphql",
```
znarf: Looks like the config is missing in all the other commands.

In the end, this might be verbose and unpractical to always require to pass it on the CLI.

Let's settle on the root file approach (changed my mind compared to the original specification).

PR #3186, opened 2020-01-13
CI: Matching branch functionality can only work on pushes

znarf:
On pull_request events, it's trying to match with PR branches and it's not what we want.

ie: downloading https://codeload.github.com/opencollective/opencollective-frontend/tar.gz/refs/pull/3177/merge

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,14 +1,6 @@
 name: E2E
 
-on:
-  push:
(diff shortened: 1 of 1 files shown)


znarf:
Fix CI failures on:
- https://github.com/opencollective/opencollective-api/pull/3177
- https://github.com/opencollective/opencollective-api/pull/3163

PR #3190, opened 2020-01-13
Multiple expense's attachments

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2803

This PR ended up being way larger than I initially thought it would be.

It fully preserves the compatibility with the existing behaviors:
- submitting an expense with the `attachment` field will just store one
- the old `expense.attachment` field returns the first attachment

Yet the new fields should provide full control for the frontend implementation.

Only thing left to do: apply these changes to API v2 (that we will consume for the expense flow). That will be done in a separate PR.

### DIFF
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -1,9 +1,9 @@
-import { get, omit } from 'lodash';
+import { get, omit, pick, flatten } from 'lodash';
 import errors from '../../../lib/errors';
 import roles from '../../../constants/roles';
 import statuses from '../../../constants/expense_status';
 import activities from '../../../constants/activities';
-import models from '../../../models';
+import models, { sequelize } from '../../../models';
 import paymentProviders from '../../../paymentProviders';
 import * as libPayments from '../../../lib/payments';
 import { formatCurrency } from '../../../lib/utils';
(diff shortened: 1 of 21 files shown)


server/graphql/v1/mutations/expenses.js
```diff
@@ -95,6 +95,26 @@ export async function updateExpenseStatus(remoteUser, expenseId, status) {
...
+  }
+
+  // Check amounts
+  const sumAttachments = attachments.reduce((total, attachment) => total + attachment.amount, 0);
+  if (sumAttachments !== expenseData.amount) {
+    throw new ValidationFailed({
+      message: `The sum of all attachments must be equal to the total expense's amount. Expense's total is ${expenseData.amount}, but the total of attachments was ${sumAttachments}.`,
+    });
+  }
+};
```
kewitz: Interesting.
This is basically a runtime integration test.
In my head, having the sum being calculated in the backend seemed simpler but if we take into consideration the frontend and third-party use of the API, I think this saves us some code/complexity.

Betree: I would actually prefer to compute the amount from attachments dynamically as you suggested  and not store any amount in `Expense` but we have too much code relying on this column (queries for statistics for example) in too many places, the conversion seemed tedious and risky.

kewitz: Perfect.

PR #3202, opened 2020-01-14
Unallocated funds query

Betree:


Betree:
@znarf I'm not sure, having it in `server/sql` would be fine to me. But these queries are meant to be usable directly (not only in the code), so I'm not chocked to have it on the root.

I recently realized some of the benefits of writing queries like that rather than directly in the code:
- IDEs understands that its SQL (you have linting, auto-format...etc)
- Queries can easily be tested in isolation (see `opencollective-api/test/sql`)
- It forces you to write queries with standard parameters (`$myVar`) rather than sequelize (`:myVar`) which is good cause a lot of DB tools understand them, so you can just copy/paste the query to use it manually

I don't have a strong opinion on this folder, but if it grows we should definitely think about a clean way to organize it.

znarf:
@Betree could that be part of another repository?

Betree:
That would not be practical. For example this unallocated funds query is meant to be imported from the code to send a special email (https://github.com/opencollective/opencollective/issues/2811). Having to modify the query and then update the subrepo would be time consuming, I don't think the small benefit of having a clear separation would be worth it.

My [comment above](https://github.com/opencollective/opencollective-api/pull/3202#issuecomment-577072493) was about how I think we should use more of these queries in the code, if that's the way we go then we should definitely keep them in the same repo.

Betree:
Also I'm happy to have access to all the test helpers and models directly to write tests like https://github.com/opencollective/opencollective-api/blob/8efbe1213217d331cf537b8c6c7e4f5521dd0e87/test/sql/ban-collectives.sql.test.js. That'd be complex to achieve in a separate repo.

znarf:
@Betree Ok! I was not getting it would be use somewhere in the codebase.

PR #3203, opened 2020-01-14
Feat: Make one-time PayPal donations host agnostic

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/2489

Before testing, make sure to add valid credentials to your host settings:
```ts
{
  paymentMethods: {
    paypal: {
      clientId: string,
      clientSecret: string
    }
  }
}
```
You can fetch that from our existing staging environment variables or create new ones [here](https://developer.paypal.com/developer/applications/create).

Before merging make sure:
- [x] Existing Collective was updated in our staging DB.
- [x] Existing Collective was updated in our production DB.

### DIFF
--- a/server/paymentProviders/paypal/payment.js
+++ b/server/paymentProviders/paypal/payment.js
@@ -21,8 +21,7 @@ export function paypalUrl(path) {
 }
 
 /** Exchange clientid and secretid by an auth token with PayPal API */
-export async function retrieveOAuthToken() {
-  const { clientId, clientSecret } = config.paypal.payment;
+export async function retrieveOAuthToken({ clientId, clientSecret }) {
   const url = paypalUrl('oauth2/token');
   const body = 'grant_type=client_credentials';
   /* The OAuth token entrypoint uses Basic HTTP Auth */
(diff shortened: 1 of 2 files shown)


kewitz:
@znarf since this is a limited, custom implementation for OpenCollective Foundation, I say we ship it this way to solve the problem and rethink the whole refactor after digging into PayPal's marketplace documentation.

server/paymentProviders/paypal/payment.js
```diff
@@ -36,9 +35,13 @@ export async function retrieveOAuthToken() {
 }
 
 /** Assemble POST requests for communicating with PayPal API */
-export async function paypalRequest(urlPath, body) {
+export async function paypalRequest(urlPath, body, hostCollective) {
+  const clientSecrets = get(hostCollective.get('settings', { raw: true }), 'paymentMethods.paypal');
```
Betree: Storing secrets in `settings` is not safe, this field is public

kewitz:
Agreed, I'll move to ConnectedAccounts for security reasons.
Moreover, what is currently preventing raw SQL queries from getting ConnectedAccounts?

Betree:
@kewitz `ConnectedAccounts` were supposedly designed with the constraint of having private data in them. So there shouldn't be any query returning that data without filtering, or the person who did it introduced a vulnerability in the code.

`settings` was designed as a public field and used as such, so as of today it's ok to query it by other means and return the data. I don't even know if this is done somewhere, but if it does it's not a security issue because we consider this field public.

Something cool we could do for this kind of things in the future is to use the `data` field instead. For most tables this field is private. Unfortunately the `Collective` model doesn't follow this pattern.
- `settings` -> field to store public data that doesn't justify creating a table column
- `settings` -> field to store private data that doesn't justify creating a table column

kewitz:
Rebased.

PR #3231, opened 2020-01-20
Multiple payout methods per collective

Betree:
Resolve https://github.com/opencollective/opencollective/issues/2822

## Breaking changes

:warning: `findByEmailOrPaypalEmail` has been removed. As a consequence we're not longer trying to match accounts on the PayPal email when creating one through redeeming gift cards or adding funds (we only match on the `email` field). We could re-introduce it, but I was anyway concerned about the security aspect of this because we have no way to know if the person really owns the `paypalEmail` that they enter in the input field.

## Legacy `payoutMethod`

The legacy `payoutMethod` has been moved to `legacyPayoutMethod`. The reason for keeping it is that we still have some checks for the `donation` expense payout type that we'll need to get rid of to remove the column and I didn't want to introduce the deprecated concept of `Donation` to the new `PayoutMethods` table. I'll open a separate issue about migrating the old entries to remove this column.

## Payout method types

I've introduced two payout method types:
- **PAYPAL**: data must be formatted like `{ email: validEmail }`
- **OTHER**: data must be formatted like `{ content: '...' }`

This is meant to be extended with new types, for example:
- **TRANSFERWISE_BANK_ACCOUNT**: data will be whatever the API needs us to store

## Graphql queries changes

No breaking changes, the legacy API endpoints will continue working.

New endpoints:
- `collective.payoutMethods`: authenticated, fetch the payout methods for this collective
- `expense.PayoutMethod`: the payout method used for this expense
- `expense.canSeePrivateInfo`: whether logged in user can see private info

## Graphql mutation changes

No breaking changes, the legacy API endpoints will continue working.

`createExpense` and `editExpense` now accept a `PayoutMethod` parameter to configure the payout method for the expense.


znarf:
For Bank Transfers, I don't think it's suitable to store the value in a string. It would be better to have a structured object, knowing that the structure of a "bank account" is different across countries.

server/lib/handlebars.js
```diff
@@ -3,7 +3,7 @@ import moment from 'moment-timezone';
 import { resizeImage, capitalize, formatCurrencyObject, pluralize } from './utils';
 
 // from https://stackoverflow.com/questions/8853396/logical-operator-in-a-handlebars-js-if-conditional
-handlebars.registerHelper('ifCond', (v1, operator, v2, options) => {
+handlebars.registerHelper('ifCond', function(v1, operator, v2, options) {
```
Betree: Add to do this because arrow functions have their own context (`this` is reset) thus is was impossible to access template's parameters in a `ifCond`.

server/graphql/v1/types.js
```diff
@@ -34,6 +34,7 @@ import { isUserTaxFormRequiredBeforePayment } from '../../lib/tax-forms';
 import { getCollectiveAvatarUrl } from '../../lib/collectivelib';
 import * as commonComment from '../common/comment';
 import { canViewExpensePrivateInfo, getExpenseAttachments } from '../common/expenses';
+import { PayoutMethodType as PayoutMethodTypes } from '../../models/PayoutMethod';
```
kewitz: Should we rename the Enum?

Betree: Yes, I just saw that we don't have a perfect consistency on that but mist of the time it's plural, so let's have plural.

PR #3238, opened 2020-01-21
Improve mocha setup

znarf:
- ignore non-test files
- split again GraphQL tests from others in CI (faster results)
- remove `seed-users-with-expenses.js`
- fix a recently introduced warning while running tests
- fix a common failure on Mac OS
- update root describe of tests with a consistent string

Fix https://github.com/opencollective/opencollective/issues/2831 (ignore non-test files and delete `seed-users-with-expenses.js`)


package.json
```diff
@@ -185,22 +185,21 @@
...
+    "test": "TZ=UTC nyc mocha",
     "test:clean": "rm -rf .nyc_output coverage",
     "test:coverage": "codecov",
-    "test:graphql": "TZ=UTC mocha test/server/graphql",
-    "test:lib": "TZ=UTC mocha test/server/lib",
-    "test:models": "TZ=UTC mocha test/server/models",
-    "test:payment-providers": "TZ=UTC mocha test/server/paymentProviders",
-    "test:routes": "TZ=UTC mocha test/server/routes",
-    "test:scripts": "TZ=UTC mocha test/scripts",
-    "test:watch": "TZ=UTC nyc mocha --watch",
```
Betree: I missed that this was removed, I use this command very often to do TDD. @znarf are you ok if I add it back?

Betree: Actually it's fine, I always specify a file when doing that so adding `--watch` at the end is not more complex

znarf: Sorry, let's bring it back!

PR #3239, opened 2020-01-21
Remove script to reset DB

Betree:
When I tried to run `npm run db:reset` it failed with the error below. By looking at the script, my impression is that it's deprecated and not maintained because it's not supposed to seed data itself, that's the role of `db:setup`. 

I solved my case by dropping the DB manually and running `npm run db:setup`. In the end I prefer this approach, it makes sure that you delete the database you intended to.

We can eventually create another script that does ` DROP DB ...` then db:setup if there's a need for it but I personally don't see it.

```
error: uncaughtException: exec(...).tap is not a function
TypeError: exec(...).tap is not a function
    at Object.<anonymous> (/home/user/Workspace/Projects/opencollective/opencollective-api/scripts/reset.js:19:4)
    at Module._compile (internal/modules/cjs/loader.js:774:30)
    at Module._compile (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/pirates/lib/index.js:99:24)
    at Module._extensions..js (internal/modules/cjs/loader.js:785:10)
    at Object.newLoader [as .js] (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/pirates/lib/index.js:104:7)
    at Module.load (internal/modules/cjs/loader.js:641:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:837:10)
    at Object.<anonymous> (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/@babel/node/lib/_babel-node.js:180:21)
    at Module._compile (internal/modules/cjs/loader.js:774:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:785:10)
    at Module.load (internal/modules/cjs/loader.js:641:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:837:10)

```


Betree:
Depcheck suggested I should also remove `child-process-promise` :+1:

PR #3244, opened 2020-01-22
Make hello@ email address opt-out

Betree:


znarf:
The problem with this change is that collectives currently relying on it will all be disabled until they opt-in.

Betree:
Yes, I just opened this for it to be ready but we still need to talk about the alternatives.

I personally don't know if anyone's using these addresses.

PR #3245, opened 2020-01-22
Don't check Host Plan to compute canApply

znarf:
Fix: https://github.com/opencollective/opencollective/issues/2837


server/models/Collective.js
```diff
@@ -905,13 +905,7 @@ export default function(Sequelize, DataTypes) {
    * It does **not** check that the collective is indeed a host.
    */
   Collective.prototype.canApply = async function() {
```
kewitz: You can also remove `async`.

znarf: We used to have an eslint rule for that, not sure what happened to it.

PR #3256, opened 2020-01-23
Internal Transferwise Payout MVP

kewitz:
Related to https://github.com/opencollective/opencollective/issues/2836
Fix https://github.com/opencollective/opencollective/issues/2890

## Testing instructions

1. Run `npm run db:migrate`
1. Create a PayoutMethod for your current testing user with
    ```js
    data: {"accountHolderName":"Leo Kewitz","currency":"EUR","type":"iban","legalType":"PRIVATE","details":{"IBAN":"DE89370400440532013000"}}
    ```
1. Create a ConnectedAccount for Open Source Collective Host with
    ```js
    service: 'transferwise'
    token: '33b5e94d-9815-4ebc-b970-3612b6aec332'
    data: {"type":"business","id":220192}
    ```
1. Submit an expense to any collective hosted by Open Source Collective.
    - Select "Bank Transfer" for payout method.
1. As an admin of that collective, approve the expense.
1. As an admin of Open Source Collective, click on the `Pay with Transferwise` button.
    1. The Expense should display _PROCESSING_ status
    1. The transaction should have been created.
    1. The fee should display in the details of the transaction.
1. Run `npm run script cron/hourly/check-pending-transferwise-transactions.js`
    1. It should log the processing expense and mark it as paid if the transfer is completed.


test/server/paymentProviders/transferwise/index.test.ts
```diff
@@ -0,0 +1,138 @@
...
+          type: 'iban',
+          legalType: 'PRIVATE',
+          details: {
+            IBAN: 'DE89370400440532013000',
+          },
+        },
+      },
+    });
+    expense = await fakeExpense({
+      payoutMethod: 'transferwise',
```
Betree: This field is not going to be there with the payout method refactor. I'd recommend waiting for https://github.com/opencollective/opencollective-api/pull/3231 to be merged before going too far, otherwise you're gonna face multiple conflicts and design issues.

server/models/Expense.js
```diff
@@ -244,6 +246,10 @@ export default function(Sequelize, DataTypes) {
     return this.getUser().then(user => user.paypalEmail || user.email);
   };
 
+  Expense.prototype.getFromCollective = function() {
+    return models.Collective.findByPk(this.FromCollectiveId);
```
Betree: In https://github.com/opencollective/opencollective-api/pull/3231/files#diff-8f509dcd331d60429c7be6e52df71925R205 I've modeled the relation properly by adding:

```es6
  m.Expense.belongsTo(m.Collective, {
    foreignKey: 'FromCollectiveId',
    as: 'fromCollective',
  });
```

You can add that here if you want, with it you don't need to create the method manually sequelize will do it for you.

server/paymentProviders/index.ts
```diff
@@ -0,0 +1,11 @@
+import opencollective from './opencollective';
+import paypal from './paypal';
+import stripe from './stripe';
+import transferwise from './transferwise';
+
+export default {
```
Betree: Thank you :pray: This magic import loop was not very nice.

Betree:
We need to merge https://github.com/opencollective/opencollective-api/pull/3231 (or rebase on it) before reviewing, it's too difficult to see the changes in the current state. I'll try to move on with this today.

kewitz:
Rebased.

config/custom-environment-variables.json
```diff
@@ -144,5 +144,8 @@
   },
   "githubFlow": {
     "minNbStars": "GITHUB_FLOW_MIN_NB_STARS"
+  },
+  "transferwise": {
+    "api": "TRANSFERWISE_API"
```
znarf: "api" doesn't mean much, should be "apiUrl"

server/graphql/v1/mutations/expenses.js
```diff
@@ -455,6 +471,12 @@ export async function payExpense(remoteUser, args) {
   if (expense.status === statuses.PAID) {
     throw new errors.Unauthorized('Expense has already been paid');
   }
+  if (expense.status === statuses.PROCESSING) {
+    throw new errors.Unauthorized('Expense has already been paid but is still being processed');
+  }
+  if (expense.status === statuses.ERROR) {
+    throw new errors.Unauthorized('Expense has already been paid and it failed, please create a new one');
```
Betree: We should be able to retry the payment if it fails, but we can figure that out later in a separate issue

server/graphql/v1/mutations/expenses.js
```diff
@@ -478,11 +499,22 @@ export async function payExpense(remoteUser, args) {
...
+  if (payoutMethodType === PayoutMethodTypes.BANK_ACCOUNT) {
+    const [connectedAccount] = await host.getConnectedAccounts({
+      where: { service: 'transferwise', deletedAt: null },
+    });
+    if (!connectedAccount) {
+      throw new Error('Host is not connected to Transferwise');
+    }
+    const quote = await paymentProviders.transferwise.getTemporaryQuote(connectedAccount, payoutMethod, expense);
+    // Notice this is the FX rate between Host and Collective, that's why we use `fxrate`.
+    fees.paymentProcessorFeeInCollectiveCurrency = Math.round((quote.fee / fxrate) * 100);
```
Betree: Not a requested change: I've introduced a `floatAmountToCents` helper in https://github.com/opencollective/opencollective-api/blob/10b10bd8f12071bf66b6da184ebbab2311b5dc48/server/lib/math.js to do that. Someone not aware of the floating numbers problem may wonder why we `Math.round` and I think this little helper can help to clarify that.

Anyway, kudos for thinking about that :+1:

server/lib/currency.js
```diff
@@ -37,21 +37,18 @@ if (!get(config, 'fixer.accessKey') && !['staging', 'production'].includes(confi
 export function getFxRate(fromCurrency, toCurrency, date = 'latest') {
   debug('>>> getFxRate for ', date, fromCurrency, toCurrency);
 
-  if (!get(config, 'fixer.accessKey')) {
+  if (fromCurrency === toCurrency) {
+    return Promise.resolve(1);
+  } else if (!fromCurrency || !toCurrency) {
+    return Promise.resolve(1);
+  } else if (!get(config, 'fixer.accessKey')) {
```
Betree: :+1:

server/lib/transferwise.ts
```diff
@@ -0,0 +1,124 @@
...
+        source: sourceCurrency,
+        target: targetCurrency,
+        rateType: 'FIXED',
+        type: 'BALANCE_PAYOUT',
+        targetAmount,
+        sourceAmount,
+      },
+      { headers: { Authorization: `Bearer ${token}` } },
+    )
+    .then(get('data'));
```
Betree: Maybe add a catch to log the error and return something that we can display to the user (I can imagine that errors may contain private infos)

server/lib/transferwise.ts
```diff
@@ -0,0 +1,124 @@
...
+  token: string,
+  { profileId: profile, currency, type, accountHolderName, legalType, details }: CreateRecipientAccount,
+): Promise<RecipientAccount> =>
+  axios
+    .post(
+      `${config.transferwise.api}/v1/accounts`,
+      { profile, currency, type, accountHolderName, legalType, details },
+      { headers: { Authorization: `Bearer ${token}` } },
+    )
+    .then(({ data: { details, ...recipient } }: { data: RecipientAccount }) => ({
```
Betree: Same here, a catch would be nice

server/paymentProviders/transferwise/index.ts
```diff
@@ -0,0 +1,75 @@
...
+import { Quote } from '../../types/transferwise';
+
+async function populateProfileId(connectedAccount): Promise<void> {
+  if (!connectedAccount.data.profile) {
+    const profiles = await transferwise.getProfiles(connectedAccount.token);
+    const profile =
+      profiles.find(p => p.type === connectedAccount.type) || profiles.find(p => p.type === 'business') || profiles[0];
+    if (profile) {
+      connectedAccount.set({ data: { ...connectedAccount.data, ...profile } });
+      await connectedAccount.save();
```
Betree: Not a requested change: this can also be written with a single `update` line 

```es6
await connectedAccount.update({ data: { ...connectedAccount.data, ...profile } });
```

cron/hourly/check-pending-transferwise-transactions.js
```diff
@@ -0,0 +1,67 @@
...
+  const transfer = await transferwiseLib.getTransfer(connectedAccount.token, transaction.data.transfer.id);
+  if (transfer.status === 'processing') {
+    console.warn(`Transfer is still being processed, nothing to do but wait.`);
+  } else if (transfer.status === 'outgoing_payment_sent') {
+    console.log(`Transfer sent, marking expense as paid.`);
+    await expense.setPaid(expense.lastEditedById);
+    await expense.createActivity(activities.COLLECTIVE_EXPENSE_PAID);
+  } else if (transfer.status === 'funds_refunded') {
+    console.warn(`Transfer ${transfer.status}, setting status to Error and deleting existing transactions.`);
+    await models.Transaction.destroy({ where: { ExpenseId: expense.id } });
```
znarf: Normally, the `Transaction` table is immutable.

Might be ok for a MVP but to be considered for the future.

cron/hourly/check-pending-transferwise-transactions.js
```diff
@@ -0,0 +1,67 @@
...
+  if (!transaction) {
+    throw new Error(`Could not find any transactions associated with expense.`);
+  }
+  const transfer = await transferwiseLib.getTransfer(connectedAccount.token, transaction.data.transfer.id);
+  if (transfer.status === 'processing') {
+    console.warn(`Transfer is still being processed, nothing to do but wait.`);
+  } else if (transfer.status === 'outgoing_payment_sent') {
+    console.log(`Transfer sent, marking expense as paid.`);
+    await expense.setPaid(expense.lastEditedById);
+    await expense.createActivity(activities.COLLECTIVE_EXPENSE_PAID);
```
znarf: I think only at this point the Transaction should be created.

server/graphql/v1/mutations/expenses.js
```diff
@@ -455,6 +471,12 @@ export async function payExpense(remoteUser, args) {
   if (expense.status === statuses.PAID) {
     throw new errors.Unauthorized('Expense has already been paid');
   }
+  if (expense.status === statuses.PROCESSING) {
+    throw new errors.Unauthorized('Expense has already been paid but is still being processed');
```
znarf: If it's processing, it's not paid.

server/lib/currency.js
```diff
@@ -37,21 +37,18 @@ if (!get(config, 'fixer.accessKey') && !['staging', 'production'].includes(confi
 export function getFxRate(fromCurrency, toCurrency, date = 'latest') {
   debug('>>> getFxRate for ', date, fromCurrency, toCurrency);
 
-  if (!get(config, 'fixer.accessKey')) {
+  if (fromCurrency === toCurrency) {
+    return Promise.resolve(1);
+  } else if (!fromCurrency || !toCurrency) {
```
znarf: Shouldn't we throw exception for that?

server/lib/transferwise.ts
```diff
@@ -0,0 +1,124 @@
+import axios from 'axios';
+import config from 'config';
+import { get, omitBy, isNull } from 'lodash/fp';
```
znarf: I'm still not 100% bringing these `fp` patterns in the codebase.

server/lib/transferwise.ts
znarf: I was surprised by the semantic of `get('data')`.

cron/hourly/check-pending-transferwise-transactions.js
kewitz: I agree. The problem is that we don't consider expenses when calculating the available balance of a collective, which means that while the transaction isn't created, the expense amount will be still available for the collective to use.
Since this payment process is async and can take up to a few days, that's a big issue.
Do you have any suggestions on how we could solve that?

test/test-helpers/fake-data.js
```diff
@@ -276,3 +276,49 @@ export const fakeTier = async (tierData = {}) => {
...
+    ...connectedAccountData,
+  });
+};
+
+/**
+ * Creates a fake transaction. All params are optionals.
+ */
+export const fakeTransaction = async (transactionData = {}) => {
+  const amount = transactionData.amount || randAmount(10, 100) * 100;
+  let CreatedByUserId = get(transactionData, 'CreatedByUserId') || get(transactionData, 'createdByUser.id');
```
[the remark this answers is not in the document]
kewitz: Agreed.

server/graphql/v1/mutations/expenses.js
kewitz: Agreed, I think that that could be achieved by just removing this. I'll review and fix it now.

server/lib/transferwise.ts
kewitz: Thanks for catching that, I also don't think it is a good idea to mix normal lodash and fp.

server/lib/currency.js
kewitz: Not sure, I'm worried that there's some edge-case here that I'm not aware.
I think we should keep the same logic.

server/paymentProviders/transferwise/index.ts
kewitz: Not really, I'm counting on the fact that this method will mutate `connectedAccount`.
This means `populateProfileId()` is not a pure function, but if that's a problem I can rewrite the logic around this.

server/lib/currency.js
znarf: I missed it was just moved from somewhere else. It's fine then.

server/paymentProviders/transferwise/index.ts
Betree: `update` does mutate the object, as per the docs:

> This is the same as calling set and then calling save but it only saves the exact values passed to it, making it more atomic and safer.

PR #3305, opened 2020-02-04
Add GraphQL API v2 ids to graphql v1 order/transaction

znarf:
This will be used for forward compatibility with GraphQL API v2.

Required by: https://github.com/opencollective/opencollective-frontend/pull/3466

### DIFF
--- a/test/server/lib/utils.test.js
+++ b/test/server/lib/utils.test.js
@@ -21,9 +21,7 @@ describe('server/lib/utils', () => {
     expect(res.card.expYear).to.equal(obj.card.expYear);
   });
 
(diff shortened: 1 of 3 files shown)


test/server/lib/utils.test.js
```diff
@@ -21,9 +21,7 @@ describe('server/lib/utils', () => {
     expect(res.card.expYear).to.equal(obj.card.expYear);
   });
 
-  it('exports PDF', function(done) {
-    this.timeout(10000);
-
+  it('exports PDF', done => {
```
znarf: For some reason, test were currently timeouting (likely unrelated) so I had to changed that.

PR #3318, opened 2020-02-07
update copy day 14 if no expenses

piamancini:
can we make this 
Day 14, if no expense 
Thanks!


znarf:
@piamancini isn't that already day14? note that it's also sent at day35

piamancini:
Just day 14 - we should remove the 35th

PR #3322, opened 2020-02-07
update copy on-boarding Day 7

piamancini:
updated copy


znarf:
@piamancini should this one be moved to day3 ?

PR #3343, opened 2020-02-10
Prepare API V2 for new create expense flow

Betree:
API changes for https://github.com/opencollective/opencollective-frontend/pull/3477
Includes a cherrypick of https://github.com/opencollective/opencollective-api/pull/3377 for the `AccountInput` type.

### DIFF
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -16,6 +16,7 @@ import { canUseFeature } from '../../../lib/user-permissions';
 import FEATURE from '../../../constants/feature';
 import { FeatureNotAllowedForUser, ValidationFailed } from '../../errors';
 import { PayoutMethodTypes } from '../../../models/PayoutMethod';
+import { types as collectiveTypes } from '../../../constants/collectives';
 
 const debug = debugLib('expenses');
 
@@ -126,6 +127,10 @@ const checkExpenseAttachments = (expenseData, attachments) => {
     throw new ValidationFailed({
       message: `The sum of all attachments must be equal to the total expense's amount. Expense's total is ${expenseData.amount}, but the total of attachments was ${sumAttachments}.`,
     });
+  } else if (!sumAttachments) {
+    throw new ValidationFailed({
+      message: `The sum of all attachments must be above 0`,
+    });
   }
 
   // If expense is a receipt (not an invoice) then files must be attached
@@ -203,7 +208,7 @@ export async function createExpense(remoteUser, expenseData) {
   if (expenseData.attachment && expenseData.attachments) {
     throw new ValidationFailed({ message: 'Fields "attachment" and "attachments" are exclusive, please use only one' });
   } else if (expenseData.attachment) {
-    // Convert legacy attachment param to new format
+    // @deprecated Convert legacy attachment param to new format
     attachmentsData = [{ amount: expenseData.amount, url: expenseData.attachment }];
   }
 
@@ -212,6 +217,8 @@ export async function createExpense(remoteUser, expenseData) {
   const collective = await models.Collective.findByPk(expenseData.collective.id);
   if (!collective) {
     throw new errors.ValidationFailed('Collective not found');
+  } else if (![collectiveTypes.COLLECTIVE, collectiveTypes.EVENT].includes(collective.type)) {
+    throw new errors.ValidationFailed('Expenses can only be submitted to collectives and events');
   }
 
   // For now we only add expenses from user's collectives
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -100,6 +100,17 @@ export async function updateExpenseStatus(remoteUser, expenseId, status) {
   return res;
 }
 
+/** Compute the total amount of expense from attachments */
+const getTotalAmountFromAttachments = attachments => {
+  if (!attachments) {
+    return 0;
+  } else {
+    return attachments.reduce((total, attachment) => {
+      return total + attachment.amount;
+    }, 0);
+  }
+};
+
 /** Check expense's attachments values, throw if something's wrong */
 const checkExpenseAttachments = (expenseData, attachments) => {
   // Check the number of attachments
@@ -110,7 +121,7 @@ const checkExpenseAttachments = (expenseData, attachments) => {
   }
 
   // Check amounts
-  const sumAttachments = attachments.reduce((total, attachment) => total + attachment.amount, 0);
+  const sumAttachments = getTotalAmountFromAttachments(attachments);
   if (sumAttachments !== expenseData.amount) {
     throw new ValidationFailed({
       message: `The sum of all attachments must be equal to the total expense's amount. Expense's total is ${expenseData.amount}, but the total of attachments was ${sumAttachments}.`,
@@ -223,6 +234,7 @@ export async function createExpense(remoteUser, expenseData) {
         incurredAt: expenseData.incurredAt || new Date(),
         PayoutMethodId: payoutMethod && payoutMethod.id,
         legacyPayoutMethod: models.Expense.getLegacyPayoutMethodTypeFromPayoutMethod(payoutMethod),
+        amount: expenseData.amount || getTotalAmountFromAttachments(attachmentsData),
       },
       { transaction: t },
     );
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -144,7 +144,7 @@ const checkExpenseAttachments = (expenseData, attachments) => {
   }
 };
 
-const EXPENSE_EDITABLE_FIELDS = ['amount', 'description', 'category', 'type', 'privateMessage'];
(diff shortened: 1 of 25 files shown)


kewitz:
@Betree I rebased this and solved the conflict.

server/graphql/v2/object/Expense.js
```diff
@@ -42,6 +45,18 @@ const Expense = new GraphQLObjectType({
           };
         },
       },
+      payee: {
+        type: Account,
+        description: 'The account being paid by this expense',
+        async resolve(expense, _, req) {
+          // Set the permissions for account's fields
+          const canSeePrivateInfo = (await canViewExpensePrivateInfo(expense, req)).userLocation;
```
kewitz: Shouldn't this be `canSeeUserLocation`? Or maybe drop `.userLocation`?

server/graphql/v2/object/Expense.js
```diff
@@ -42,6 +45,18 @@ const Expense = new GraphQLObjectType({
...
+          setContextPermission(req, PERMISSION_TYPE.SEE_ACCOUNT_LOCATION, expense.FromCollectiveId, canSeePrivateInfo);
```
kewitz: Not sure how useful is this, couldn't we just use `canViewExpensePrivateInfo` directly?

server/graphql/v1/mutations/expenses.js
```diff
@@ -212,6 +217,8 @@ export async function createExpense(remoteUser, expenseData) {
   const collective = await models.Collective.findByPk(expenseData.collective.id);
   if (!collective) {
     throw new errors.ValidationFailed('Collective not found');
+  } else if (![collectiveTypes.COLLECTIVE, collectiveTypes.EVENT].includes(collective.type)) {
+    throw new errors.ValidationFailed('Expenses can only be submitted to collectives and events');
```
kewitz: Hmm, why are we adding this limitation?
This seems to be an interesting feature.

server/graphql/v2/input/ExpenseCreate.ts
```diff
@@ -0,0 +1,35 @@
+import { GraphQLString, GraphQLInputObjectType, GraphQLList, GraphQLNonNull } from 'graphql';
+import { ExpenseType } from '../enum/ExpenseType';
+import { ExpenseAttachmentCreate } from './ExpenseAttachmentCreate';
+import { PayoutMethodInput } from './PayoutMethodInput';
+import { AccountInput } from './AccountInput';
+
+/**
+ * Input type to use as the type for the expense input in createExpense mutation.
+ */
+const ExpenseCreate = new GraphQLInputObjectType({
```
kewitz: Shouldn't this be named `ExpenseInput`?

Betree: The idea is to have different input types for create and edit mutations, in a similar fashion to what was done with CommentCreate and CommentEdit. For example there's no reason to accept an expense ID in the create path, and their may be some params that we don't want to allow in the edit path, splitting in two types makes that possible.

kewitz: Cool, that makes sense!

server/graphql/v1/mutations/expenses.js
Betree: Expenses have always had this limitation, only before this PR it was only enforced in the frontend. The goal of this line is to make it explicit to prevent inconsistencies in the data if someone misuse the endpoint.

server/graphql/v2/object/Expense.js
Betree: Some permissions depend on the context. For example, you're normally not allowed to watch the address of a user. But if you're a host admin validating an expense that the user has submitted, then you're allowed to see it.

Problem is: we have no clue in the User's location resolver about this context. We could eventually go up in the AST from there to get it, but it would get pretty messy quickly (and I'd argue that the user location resolver should not have any logic related to expenses).

So how to represent this kind of permissions check that depends on the parent?

For the incognito feature the decision was to store it in the collective as a virtual field (`inTheContextOfUserId`). We didn't block this decision because the feature needed to be released fast, yet it is problematic: there's no reason for the collective model to know about Graphql's context and anyone anywhere can change this value with whatever they want.

With this small helper, what I tried to do is to store this information in a standardize way in the GraphQl context. I believe storing it there is right, because it's what this object is about: you normally can't access this data, but in the context of **this query** you're allowed to.

Do you see better ways to achieve that?

kewitz: Cool, I get it now why this is happening.
One alternative is to set the user role in the context and write the permission logic in the resolver.

What I mean here is that the context would hold and enum value related to its user role _e.g._ `ROLE.HOST_ADMIN` or `ROLE.PARENT_COLLECTIVE_ADMIN` or `ROLE.OWNER`.
In this scenario, the resolver would take that generic context information and be explicit about its access logic:
```ts
resolve(individual, _, req) {
  if (
    req.role === ROLE.HOST_ADMIN ||
    req.role === ROLE.OWNER
  ) {
    return individual.location;
  }
}
```
The idea would be to work on a more abstracted level than going straight to granular permissions.
WDYT?

Betree: I could work with that approach, but the permissions here are too vague: 
- Host admin of which host? Owner of what? 
- It doesn't account for users with multiple roles (ie. owner of the expense **and** host admin)

Let's take a more complex query:
 
```
Host
└── Collective
  └── MemberCollectives
      └── Expenses
          ├── PayoutMethod => can only see if owner of PayoutMethod or host admin
          │   └── data
          └── User
              └── email => can only see if collective admin, owner or host admin
```

With your proposal, for example, how would you make sure that the `ROLE.HOST_ADMIN` was set for the proper collective? Or said another way, how would you know if the role was set at the `Collective` or at the `MemberCollective` level?

We could address this question by adding something like:

```ts
// Data
req.roles = {
  Expense: {
    42: ['ADMIN', 'OWNER']
  },
  Collective: {
    56: ['ADMIN']
  }
}

// Lib
type Entity = 'Expense' | 'Colllective' | ...
type Role = 'ADMIN' | 'OWNER'
setContextRole(req, entity: Entity, entityId, role: Role) { ... }
hasContextRole(req, entity, entityId) {...}
```

But then we go back to the initial problem: the user's resolver at the bottom of the tree doesn't know which expense it was retrieved from.

Betree: I could work with that approach, but the permissions here are too vague: 
- Host admin of which host? Owner of what? 
- It doesn't account for users with multiple roles (ie. owner of the expense **and** host admin)

Let's take a more complex query:
 
```
Host
└── Collective
  └── MemberCollectives
      └── Expenses
          ├── PayoutMethod => can only see if owner of PayoutMethod or host admin
          │   └── data
          └── User
              └── email => can only see if collective admin, owner or host admin
```

With your proposal, for example, how would you make sure that the `ROLE.HOST_ADMIN` was set for the proper collective? Or said another way, how would you know if the role was set at the `Collective` or at the `MemberCollective` level?

We could address this question by adding something like:

```ts
// Data
req.roles = {
  Expense: {
    42: ['ADMIN', 'OWNER']
  },
  Collective: {
    56: ['ADMIN']
  }
}

// Lib
type Entity = 'Expense' | 'Colllective' | ...
type Role = 'ADMIN' | 'OWNER'
setContextRole(req, entity: Entity, entityId, role: Role) { ... }
hasContextRole(req, entity, entityId) {...}
```

But then we go back to the initial problem: the user's resolver at the bottom of the tree doesn't know which expense it was retrieved from.

Betree: Another example to demonstrate that it is actually dangerous:

```
Collective
└── RelatedCollective
    └── private
```

If `RelatedCollective` resolver forgets to set `req.roles = COLLECTIVE_ADMIN`, then you'll get access to `private` without being an admin of RelatedCollective.

Also, this needs to be verified, but if resolvers can run in parallel (which I assume they do when they're asynchronous) then we can't trust that `req.roles` because we won't be able to know which part of the code filled it.

server/graphql/v2/object/Expense.js
Betree: Simpler functions would be better indeed, I initially did that to return the full object in https://github.com/opencollective/opencollective-api/blob/94800abaf8fc5fb6c8d80c8e31079a1b2884feda/server/graphql/v1/types.js#L813 but let's not make the rest of the code more complex for just one endpoint.

Betree: Simpler functions would be better indeed, I initially did that to return the full object in https://github.com/opencollective/opencollective-api/blob/94800abaf8fc5fb6c8d80c8e31079a1b2884feda/server/graphql/v1/types.js#L813 but let's not make the rest of the code more complex for just one endpoint.

server/graphql/v2/object/Expense.js
Betree: I'm merging to we can move forward, feel free to open an issue on this topic if you want to explore further alternatives

kewitz: Sure! This is a wicked problem that requires hard thinking.
I'm fine with the current implementation, perhaps it is just a bit complex because the problem is also complex.

PR #3355, opened 2020-02-12
PaymentMethod: filter fields in getOrCreate

Betree:


server/models/PaymentMethod.js
```diff
@@ -488,7 +488,7 @@ export default function(Sequelize, DataTypes) {
       /*
       if (paymentMethod.token) {
         const paymentMethodWithToken = await models.PaymentMethod.findOne({
-          where: { token: paymentMethod.token },
+          where: { token: paymentMethod.token, CollectiveId: paymentMethod.CollectiveId },
```
Betree: If we ever uncomment this it will make it safer by ensuring we're not using someone else's token

PR #3359, opened 2020-02-13
Delete homepage related code

znarf:
Now that we have a new homepage, we can safely delete related code.﻿


znarf:
Change in coverage is interesting but the code is still used in `user-report` so I will ignore it.

PR #3369, opened 2020-02-17
Add TransferWise required account info resolvers

kewitz:
Requesting review to validate the implementation, this PR is still missing tests.

Requests look like this:
```gql
query {
  collective(slug: "apex") {
    host {
      transferwise {
        availableCurrencies
        requiredFields(currency: EUR) {
          type
          title
          fields {
            name
            group {
              key
              name
              type
              required
              example
              valuesAllowed {
                key
                name
              }
            }
          }
        }
      }
    }
  }
}
```


server/graphql/v2/query/RequiredBankInformationQuery.ts
```diff
@@ -0,0 +1,23 @@
+import { GraphQLInt, GraphQLNonNull } from 'graphql';
+
+import RequiredBankInformation from '../object/RequiredBankInformation';
+import models from '../../../models';
+
+const RequiredBankInformationQuery = {
+  type: RequiredBankInformation,
+  args: {
+    collectiveId: {
```
znarf: We avoid exposing auto-increment ids in GraphQL API v2, alternatives: slug, hash-ids.

server/paymentProviders/transferwise/index.ts
```diff
@@ -67,7 +68,57 @@ async function payExpense(connectedAccount, payoutMethod, expense): Promise<any>
...
+    targetCurrency: currency,
+    targetAmount: 100,
+  });
+  const requiredFields = await transferwise.getAccountRequirements(connectedAccount.token, quote.id);
+  cache.set(cacheKey, requiredFields, 24 * 60 * 60 /* a whole day and we could probably increase */);
+  return requiredFields;
+}
+
+async function getAvailableCurrencies(collective): Promise<any> {
+  const cacheKey = `req_avail_currs_${collective.id}`;
```
znarf: We should avoid cryptic `avail_currs`, better: `available_currencies` Maybe prefix key with `transferwise_` .

server/paymentProviders/transferwise/index.ts
```diff
@@ -67,7 +68,57 @@ async function payExpense(connectedAccount, payoutMethod, expense): Promise<any>
   return { quote, recipient, transfer, fund };
 }
 
+async function getRequiredBankInformation(collective, currency: string): Promise<any> {
+  const cacheKey = `req_bank_info_${collective.id}_to_${currency}`;
```
znarf: maybe prefix with `transferwise` instead of `req`

kewitz:
Rebased.

server/graphql/v2/query/RequiredBankInformationQuery.ts
```diff
@@ -0,0 +1,24 @@
...
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'The public id identifying the host collective',
+    },
+  },
+  async resolve(_, args, req): Promise<any> {
+    if (!req.remoteUser) {
+      return null;
+    }
+    const slug = args.slug.toLowerCase();
+    req.collective = await models.Collective.findBySlug(slug);
```
Betree: We should avoid using `context` like this whenever possible because context is global (it is shared with by all resolvers). 

1. We're going to have problems if a different endpoint starts using the same `collective` field for a totally different purpose.
2. If you're not aware of both files, it won't be clear why we're touching the context (where does this `collective` comes from?).

An easy way to solve this is to have a `CollectiveId` field inside of `RequiredBankInformation` and return `{ CollectiveId }` instead of an empty object. Better: I'd suggest cherrypicking what @znarf did in https://github.com/opencollective/opencollective-api/blob/972431161bbc79d8531445e414a7333925e5321f/server/graphql/v2/input/AccountInput.js so that you take an `account` in args rather than a `slug` or an `id`.

server/lib/transferwise.ts
```diff
@@ -177,3 +177,29 @@ export const getTransfer = async (token: string, transferId: number): Promise<an
...
+export const getAccountRequirements = async (token: string, quoteId: number): Promise<any> => {
+  try {
+    const response = await axios.get(`${config.transferwise.api}/v1/quotes/${quoteId}/account-requirements`, {
+      headers: { Authorization: `Bearer ${token}` },
+    });
+    return getData(response);
+  } catch (e) {
+    const message = `Unable to get account requirements data: ${getAxiosError(e)}`;
+    logger.error(message, { quoteId });
+    throw new Error(message);
```
Betree: Can error include sensitive information? If that's the case, we should log the full error but return something like this:
> An unknown error happened with Transferwise. Please contact support@opencollective.com.

server/paymentProviders/transferwise/index.ts
```diff
@@ -67,7 +68,57 @@ async function payExpense(connectedAccount, payoutMethod, expense): Promise<any>
   return { quote, recipient, transfer, fund };
 }
 
+async function getRequiredBankInformation(collective, currency: string): Promise<any> {
+  const cacheKey = `transferwise_required_bank_info_${collective.id}_to_${currency}`;
```
Betree: Good idea to cache this :+1: 

Maybe we could cache at the host level, because most of the time collectives have the same currency than the host and for OSC which has 1700+ collectives this can make a difference.
```suggestion
  const cacheKey = `transferwise_required_bank_info_${hostCollective.id}_${collective.currency}_to_${currency}`;
```

server/paymentProviders/transferwise/index.ts
```diff
@@ -67,7 +68,57 @@ async function payExpense(connectedAccount, payoutMethod, expense): Promise<any>
...
+    targetCurrency: currency,
+    targetAmount: 100,
+  });
+  const requiredFields = await transferwise.getAccountRequirements(connectedAccount.token, quote.id);
+  cache.set(cacheKey, requiredFields, 24 * 60 * 60 /* a whole day and we could probably increase */);
+  return requiredFields;
+}
+
+async function getAvailableCurrencies(collective): Promise<any> {
+  const cacheKey = `transferwise_available_currencies_${collective.id}`;
```
Betree: Same here, great that you used a cache, let's have it at the host level if that's possible.

server/graphql/v2/query/RequiredBankInformationQuery.ts
kewitz: Perfect, thanks!

server/paymentProviders/transferwise/index.ts
kewitz: Nice catch! I refactored the way we move things around and I'm now caching on the host level! Thank you!

server/lib/transferwise.ts
kewitz: Agreed.

kewitz: I'm not completely sure, but since this error is generated by TW there's no way to guarantee. I agree with the solution you proposed.

server/paymentProviders/transferwise/index.ts
```diff
@@ -1,8 +1,14 @@
 import uuidv4 from 'uuid/v4';
 
 import * as transferwise from '../../lib/transferwise';
+import cache from '../../lib/cache';
+import models from '../../models';
 import { Quote } from '../../types/transferwise';
 
+export const blackListedCurrencies = [
+  'BRL', // Businesses customers are not supported yet.
```
Betree: Not sure I understand this one

server/graphql/v2/object/RequiredBankInformation.ts
```diff
@@ -0,0 +1,31 @@
+import { GraphQLNonNull, GraphQLString, GraphQLObjectType } from 'graphql';
+import GraphQLJSON from 'graphql-type-json';
+
+import transferwise from '../../../paymentProviders/transferwise';
+
+const RequiredBankInformation = new GraphQLObjectType({
+  name: 'RequiredBankInformation',
+  description: 'A list of required bank information inputs',
+  fields: {
+    fields: {
```
znarf: `fields: { fields: {` Is that correct?

server/graphql/v2/query/RequiredBankInformationQuery.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const RequiredBankInformationQuery = {
+  type: RequiredBankInformation,
+  args: {
+    slug: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'The public id identifying the host collective',
+    },
+  },
+  async resolve(_, args, req): Promise<any> {
```
znarf: I don't see the benefit of Typescript in this file.

Should we create .ts files even when we're not really using it?

server/graphql/v2/query/RequiredBankInformationQuery.ts
```diff
@@ -0,0 +1,25 @@
+import { GraphQLNonNull, GraphQLString } from 'graphql';
+
+import RequiredBankInformation from '../object/RequiredBankInformation';
+import models from '../../../models';
+
+const RequiredBankInformationQuery = {
```
znarf: I would not do that as a top-level query. If this is a property of the host, it should be at property on the `Account` level, perhaps on a dedicated `Host` object.

server/graphql/v2/object/RequiredBankInformation.ts
```diff
@@ -0,0 +1,31 @@
...
+          description: 'The 3 letter code identifying the currency you want to receive (ie: USD, EUR, BRL, GBP)',
+        },
+      },
+      type: GraphQLJSON,
+      async resolve({ host }, args) {
+        return await transferwise.getRequiredBankInformation(host, args.currency);
+      },
+    },
+    currencies: {
+      type: GraphQLJSON,
```
znarf: Could we express that as something more precise than just `GraphQLJSON`.

How is the data looking like?

server/graphql/v2/object/RequiredBankInformation.ts
```diff
@@ -0,0 +1,31 @@
...
+import transferwise from '../../../paymentProviders/transferwise';
+
+const RequiredBankInformation = new GraphQLObjectType({
+  name: 'RequiredBankInformation',
+  description: 'A list of required bank information inputs',
+  fields: {
+    fields: {
+      args: {
+        currency: {
+          type: new GraphQLNonNull(GraphQLString),
```
znarf: Can we express that as a `Currency`? it's already existing as `enum`.

server/graphql/v2/object/RequiredBankInformation.ts
```diff
@@ -0,0 +1,31 @@
...
+          description: 'The 3 letter code identifying the currency you want to receive (ie: USD, EUR, BRL, GBP)',
+        },
+      },
+      type: GraphQLJSON,
```
znarf: Could we express that as something more precise than just GraphQLJSON.

How is the data looking like?

server/graphql/v2/object/RequiredBankInformation.ts
znarf: Ok, got it now. Still look funky when you read it.

znarf:
I would push for the following:

```
query {
  Host(slug: "opensource") {
    requiredBankInformation {
      currencies
  }
}
```

I think we can also find a better API design for `requiredBankInformation`. Some ideas:

```
query {
  Host(slug: "opensource") {
    transferwise {
      currencies
  }
}

query {
  Host(slug: "opensource") {
    transferwise {
      supportedCurrencies
  }
}


query {
  Host(slug: "opensource") {
    transferwise {
      requiredBankInformation
  }
}
```

I know the idea is that we could reuse the same "required bank information" for SEPA payouts. But I think it's clearer right now if we express the relationship with transferwise explicitely. Also match better the code architecture that is behind.

server/paymentProviders/transferwise/index.ts
kewitz: This is a TransferWise limitation, it only allows bank transfers between private parts.
That's why I'm currently blacklisting that.

kewitz: The right way to do is to check if the host has a private or a business account and allow if it's private. But I don't think we should worry about that right now.

server/graphql/v2/query/RequiredBankInformationQuery.ts
kewitz: I think it is a good idea because you get static checking from other imported TS.
I don't see a lot of value in requiring us to be explicit about the returning value though.
It should provide more value when we migrate all models to TS.

server/graphql/v2/object/RequiredBankInformation.ts
kewitz: We don't have a way to guarantee the subset of currencies this host can transfer to will be the same that we have in the enum.

kewitz:
@znarf Yes, I pretty much agree. I think dealing with this as if it was some generic data is a bit of wishful thinking.
I'll refactor to relocate these fields underneath host and namespace as Transferwise.

server/graphql/v2/object/RequiredBankInformation.ts
kewitz: Renamed to required fields.

server/graphql/v2/query/RequiredBankInformationQuery.ts
kewitz: Moved to Account so we can update the existing query to
```gql
host {
  transferwise {
    availableCurrencies
    requiredFields(currency: "EUR")
  }
}
```

server/graphql/v2/object/RequiredBankInformation.ts
znarf: Ok. We could possibly add all standardized currency codes in this GraphQL type:
https://www.iban.com/currency-codes

server/graphql/v2/object/RequiredBankInformation.ts
kewitz: I'll structure that.

server/graphql/v2/query/RequiredBankInformationQuery.ts
Betree: I agree with that. Even if you're not actively typing everything, that doesn't mean you're not using typescript. For example:

```es6
const x = 42;
const y = 'test';
const z = x / y; // typescript will not be happy
```

On the other hand, I don't see any downside about using ts files

PR #3375, opened 2020-02-18
Host Plans follow up

znarf:
- Everyone is allowed to add `manualPayments`
- Add `bankTransfers` and `bankTransfersLimit`

Implement partially https://github.com/opencollective/opencollective/issues/2912


znarf:
Merging because I want that to go live. @kewitz let me know if you have feedback

PR #3377, opened 2020-02-18
GraphQL API v2 createCollective mutation

znarf:


server/graphql/v2/input/AccountInput.js
```diff
@@ -0,0 +1,35 @@
...
+  if (input.id) {
+    const id = idDecode(input.id, 'account');
+    collective = await models.Collective.findByPk(id);
+  } else if (input.legacyId) {
+    collective = await models.Collective.findByPk(input.legacyId);
+  } else if (input.slug) {
+    const slug = input.slug.toLowerCase();
+    collective = await models.Collective.findBySlug(slug);
+  } else {
+    return new Error('Please provide a slug or an id');
```
Betree: Shouldn't it be throw rather than return?

```suggestion
    throw new Error('Please provide a slug or an id');
```

server/graphql/v2/input/AccountInput.js
```diff
@@ -0,0 +1,35 @@
...
+const AccountInput = new GraphQLInputObjectType({
+  name: 'AccountInput',
+  fields: () => ({
+    id: { type: GraphQLString },
+    slug: { type: GraphQLString },
+    legacyId: { type: GraphQLInt },
+  }),
+});
+
+const fetchAccountWithInput = async input => {
```
Betree: Ideally we should accept the `throwIfMissing` here. Sending this arg from frontend is not very nice, it was introduced because fetching a collective that slug doesn't exist in a `collective` query should return `null`, not an error but we didn't want to break the existing. 

In mutations however it's ok to throw if the provided collective doesn't exist.

So for new endpoints we'll normally won't have to expose the `throwIfMissing` param publicly.

server/graphql/v2/input/AccountInput.js
```diff
@@ -0,0 +1,45 @@
...
+      description: 'The slug identifying the account (ie: babel for https://opencollective.com/babel)',
+    },
+  }),
+});
+
+const fetchAccountWithInput = async (input, options) => {
+  let collective;
+  if (input.id) {
+    const id = idDecode(input.id, 'account');
+    collective = await models.Collective.findByPk(id);
```
Betree: This fails with the following message:
> Argument passed to findByPk is invalid: 83

It's because `idDecode` returns a string, it goes away if I do `parseInt(idDecode(input.id, 'account'))`. Seems to me that `idDecode` should return an integer no?

server/graphql/v2/input/AccountInput.js
```diff
@@ -0,0 +1,45 @@
...
+    const slug = input.slug.toLowerCase();
+    collective = await models.Collective.findBySlug(slug);
+  } else {
+    throw new Error('Please provide an id or a slug');
+  }
+  if (!collective && options && options.throwIfMissing) {
+    throw new NotFound({ message: 'Account Not Found' });
+  }
+  return collective;
+};
```
Betree: Can we use the GraphQL loaders in there to make it safe to use this in sub-fields?

server/graphql/v2/input/AccountInput.js
```diff
@@ -0,0 +1,45 @@
...
+  }),
+});
+
+const fetchAccountWithInput = async (input, options) => {
+  let collective;
+  if (input.id) {
+    const id = idDecode(input.id, 'account');
+    collective = await models.Collective.findByPk(id);
+  } else if (input.legacyId) {
+    collective = await models.Collective.findByPk(input.legacyId);
```
Betree: Would be good to use GraphQL loaders in there to make it safe with subfields. Example:

```gql
account(id: 'x') {
  expenses(fromAccount: { id: 'y' }) {
    ...
  }
}
```

server/graphql/v2/input/AccountInput.js
znarf: I just spent one hour on that! `idDecode` actually returns an **array** and it's the problem.

Fixing that in this commit: https://github.com/opencollective/opencollective-api/pull/3377/commits/65e0496397a4b83c6730d52d541673da5314582a

server/graphql/v2/mutation/CreateCollectiveMutations.js
```diff
@@ -0,0 +1,130 @@
...
+  const { remoteUser, loaders } = req;
+
+  if (!remoteUser) {
+    throw new errors.Unauthorized({
+      message: 'You need to be logged in to create a collective',
+    });
+  }
+
+  const collectiveData = {
+    ...pick(args.collective, ['name', 'slug', 'description', 'tags']),
```
sbinlondon: So we will only need to input name, slug, desc, and tags from the form for the Collective to be created? (Plus the info from the remote user, etc...)

test/server/graphql/v2/mutation/CreateCollectiveMutations.test.js
```diff
@@ -0,0 +1,44 @@
...
+const createCollectiveQuery = `
+    mutation createCollective($collective: CollectiveCreate!) {
+      createCollective(collective: $collective) {
+        name
+        slug
+        tags
+      }
+    }`;
+
+const newCollectiveData = {
```
sbinlondon: The test is helpful to show an example of the kind of data I will need to input. Thanks! 🙌

server/graphql/v2/mutation/CreateCollectiveMutations.js
```diff
@@ -0,0 +1,130 @@
...
+      collective: {
+        type: new GraphQLNonNull(CollectiveCreate),
+      },
+      host: {
+        type: AccountInput,
+      },
+      automateApprovalWithGithub: {
+        type: GraphQLBoolean,
+        defaultValue: false,
+      },
```
znarf: @sbinlondon As top parameter, there is also `host` (ie: `{"slug": "opensource"}`) and `automateApprovalWithGithub` (boolean).

server/graphql/v2/input/CollectiveCreate.js
```diff
@@ -0,0 +1,14 @@
+import { GraphQLString, GraphQLNonNull, GraphQLList, GraphQLInputObjectType } from 'graphql';
+
+const CollectiveCreate = new GraphQLInputObjectType({
+  name: 'CollectiveCreate',
+  fields: () => ({
+    name: { type: new GraphQLNonNull(GraphQLString) },
+    slug: { type: new GraphQLNonNull(GraphQLString) },
+    description: { type: new GraphQLNonNull(GraphQLString) },
+    tags: { type: new GraphQLList(GraphQLString) },
+    githubHandle: { type: GraphQLString },
```
znarf: @sbinlondon As collective params, there is also `githubHandle`.

sbinlondon: But this is only with open source projects, right?

server/graphql/v2/mutation/CreateCollectiveMutations.js
sbinlondon: And when would we automate approval with Github?

znarf: Always when we pick the GitHub path.

server/graphql/v2/input/CollectiveCreate.js
znarf: Yes

server/graphql/v2/input/AccountInput.js
znarf: @Betree Actually went one step back, this is now the fixing commit: https://github.com/opencollective/opencollective-api/pull/3377/commits/ba8d36138d37ff4106f658c21bb71e1f210e2afb `getDecodedId` is not removed.

server/graphql/v1/mutations/collectives.js
```diff
@@ -243,46 +243,20 @@ export async function createCollectiveFromGithub(_, args, req) {
...
-      });
-    }
-    const allRepos = await github.getAllOrganizationPublicRepos(githubHandle).catch(() => null);
-    const repoWith100stars = allRepos.find(repo => repo.stargazers_count >= config.githubFlow.minNbStars);
-    if (!repoWith100stars) {
-      throw new errors.ValidationFailed({
-        message: `The organization need at least one repository with ${config.githubFlow.minNbStars} GitHub stars to be pledged.`,
-      });
+  try {
+    await github.handleOpenSourceAutomatedApproval(githubHandle, req.remoteUser);
```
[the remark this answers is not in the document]
znarf: Addressed in https://github.com/opencollective/opencollective-api/pull/3421

PR #3378, opened 2020-02-18
Enforce Bank Transfer limit when creating bank transfer orders

kewitz:
Related to opencollective/opencollective#2912.


znarf:
Let's keep in mind that we'll need to properly handle currency at some point.

PR #3391, opened 2020-02-20
Prevent spammy collectives from being created

Betree:
The only case where this can happen if it's the score is `1`, which implies that we detected a spam domain (ie `supplementslove.com`). If we detect such a domain:

1. We prevent the collective creation, with an error message like this:

> Our systems detected suspicious activity from your account and prevented the creation of this collective. If this is a mistake, please contact support@opencollective.com

2. We put the user that tried to create the collective into limited mode (see https://docs.opencollective.com/help/internal/moderation#limit-user-account)

3. We notify slack #abuse channel

## Why I think it's a good solution

Something I've noticed from the SPAM job is that spammers usually post a lot of ads for the same website (sometimes more than 30+ profiles for the same one). Doing this is a cheap and efficient way to block recurring spammers by preventing the collective creation. Banning collectives being an expensive task, everything we can do to prevent them from being created will save us time.

We could think that they can counter-react to this measure by offuscating the URL (for example with URL shorteners) but I think that goes against their incentive: having their URL linked from another website is good for SEO, that's what they're probably after and they'll loose that if they try to offuscate the URL.


server/lib/spam.ts
```diff
@@ -2,10 +2,35 @@ import { clamp } from 'lodash';
 import config from 'config';
 import slackLib from '../lib/slack';
 
-// Watched content. Please keep everything in there lowercase.
+/** Return type when running a spam analysis */
+export type SpamAnalysisRport = {
```
znarf: Typo

server/models/Collective.js
```diff
@@ -650,6 +656,31 @@ export default function(Sequelize, DataTypes) {
...
+            if (user) {
+              await user.limitAcount(spamReport);
+            }
+
+            // Notify Slack
+            notifyTeamAboutPreventedCollectiveCreate(spamReport);
+
+            // Prevent collective creation
+            throw new Error(
+              'Our systems detected suspicious activity from your account and prevented the creation of this collective. If this is a mistake, please contact support@opencollective.com,',
```
znarf: More generic error message perhaps?

PR #3392, opened 2020-02-21
feat: add collective name to update emails

bolariin:
Resolves https://github.com/opencollective/opencollective/issues/2919


znarf:
Thank you for the contribution! Could you share a screenshot so people can see how it looks?

bolariin:
Here are the screenshots

* Without the collective logo
Screenshot without logo

* With the collective logo
New update email with name included

PR #3398, opened 2020-02-21
Store IP and User Agent

znarf:
- Store ip and userAgent of creationRequest
- Store ip and userAgent of lastLoginRequest

Implement: https://github.com/opencollective/opencollective/issues/2930

### DIFF
--- a/server/models/User.js
+++ b/server/models/User.js
@@ -460,7 +460,7 @@ export default (Sequelize, DataTypes) => {
 
     const sequelizeParams = transaction ? { transaction } : undefined;
     debug('createUserWithCollective', userData);
-    const cleanUserData = pick(userData, ['email', 'firstName', 'lastName', 'newsletterOptIn']);
+    const cleanUserData = pick(userData, ['email', 'firstName', 'lastName', 'newsletterOptIn', 'data']);
     const user = await User.create(cleanUserData, sequelizeParams);
     let name = userData.firstName;
     if (name && userData.lastName) {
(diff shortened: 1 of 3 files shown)


znarf:
Implement https://github.com/opencollective/opencollective/issues/2930

server/models/User.js
```diff
@@ -446,7 +446,7 @@ export default (Sequelize, DataTypes) => {
 
     const sequelizeParams = transaction ? { transaction } : undefined;
     debug('createUserWithCollective', userData);
-    const cleanUserData = pick(userData, ['email', 'firstName', 'lastName', 'newsletterOptIn']);
+    const cleanUserData = pick(userData, ['email', 'firstName', 'lastName', 'newsletterOptIn', 'data']);
```
Betree: I'd prefer if we keep the `data` out of users reach, we use this field to store info that they should not be able to alter

znarf: @Betree what is the actionable?

Betree: To pass it in a way users can't submit it themselves.

With the current code if you call `/api/singin` with the following payload, it will be accepted:
```js
{ 
  ...userData, 
  lastSignInRequest: { ip: 'fake-ip' }, 
  feature: { allTheFeaturesThatIWant: true } 
}
```

A potential way to solve this can be to change the signature of `createUserWithCollective` to:
```
User.createUserWithCollective = async (userData, {transaction, creationRequest})
```

Or your could also update the user after the creation in the relevant paths.

PR #3407, opened 2020-02-24
payExpense: Protect manual payout when no payout method is set

Betree:
Fix https://opencollective.slack.com/archives/GDTJY7S48/p1582237171005300

Before this PR trying to pay an expense manually would fail if the payout method type was paypal and no paypal email was set for user.


Betree:
Fixed the data manually rather than introducing this hack

PR #3408, opened 2020-02-24
TransferWise Payout Form: Fixes and API updates

kewitz:
Based on https://github.com/opencollective/opencollective-api/pull/3343
Relates to https://github.com/opencollective/opencollective/issues/2858

Make sure the Host you're using is connected to TransferWise
```js
connectedAccount: {
  service: 'transferwise',
  token: '33b5e94d-9815-4ebc-b970-3612b6aec332',
  data: {"type":"business","id":220192},
}
```

### DIFF
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -171,10 +171,9 @@ const accountFieldsDefinition = () => ({
     type: TransferWise,
     async resolve(collective) {
       const connectedAccount = await models.ConnectedAccount.findOne({
-        where: { service: 'transferwise', CollectiveId: collective.id, deletedAt: null },
+        where: { service: 'transferwise', CollectiveId: collective.id },
       });
       if (connectedAccount) {
-        collective.connectedAccount = connectedAccount;
         return collective;
       } else {
         return null;
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -169,8 +169,16 @@ const accountFieldsDefinition = () => ({
   },
   transferwise: {
     type: TransferWise,
-    resolve(collective) {
-      return collective;
+    async resolve(collective) {
+      const connectedAccount = await models.ConnectedAccount.findOne({
+        where: { service: 'transferwise', CollectiveId: collective.id, deletedAt: null },
+      });
+      if (connectedAccount) {
+        collective.connectedAccount = connectedAccount;
+        return collective;
+      } else {
+        return null;
+      }
     },
   },
   payoutMethods: {
--- a/server/lib/transferwise.ts
+++ b/server/lib/transferwise.ts
@@ -45,9 +45,7 @@ export const createQuote = async (
     return getData(response);
   } catch (e) {
     logger.error(`Unable to create quote: ${getAxiosError(e)}`, data);
-    const publicMessage =
-      get(e, 'response.data.errors[0].message') || `Sorry, we can't make transfers to ${targetCurrency}.`;
-    throw new Error(publicMessage);
+    throw new Error(`Sorry, we can't make transfers to ${targetCurrency}.`);
   }
 };
 
--- a/server/lib/transferwise.ts
+++ b/server/lib/transferwise.ts
@@ -1,6 +1,6 @@
 import axios, { AxiosError } from 'axios';
 import config from 'config';
-import { get, omitBy, isNull } from 'lodash';
+import { omitBy, isNull } from 'lodash';
 
 import logger from './logger';
 import { Quote, RecipientAccount } from '../types/transferwise';
--- a/server/lib/transferwise.ts
+++ b/server/lib/transferwise.ts
@@ -1,6 +1,6 @@
 import axios, { AxiosError } from 'axios';
 import config from 'config';
-import { omitBy, isNull } from 'lodash';
+import { get, omitBy, isNull } from 'lodash';
 
 import logger from './logger';
 import { Quote, RecipientAccount } from '../types/transferwise';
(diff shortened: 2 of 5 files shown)


server/graphql/v2/input/AccountInput.js
```diff
@@ -0,0 +1,45 @@
...
+  } else {
+    throw new Error('Please provide an id or a slug');
+  }
+  if (!collective && options && options.throwIfMissing) {
+    throw new NotFound({ message: 'Account Not Found' });
+  }
+  return collective;
+};
+
+export { AccountInput, fetchAccountWithInput };
```
znarf: This is now on master. Rebase.

Betree:
I'll rebase and merge https://github.com/opencollective/opencollective-api/pull/3343 first to simplify the review

kewitz:
Rebased.

server/graphql/v2/interface/Account.js
```diff
@@ -169,8 +169,16 @@ const accountFieldsDefinition = () => ({
   },
   transferwise: {
     type: TransferWise,
-    resolve(collective) {
-      return collective;
+    async resolve(collective) {
+      const connectedAccount = await models.ConnectedAccount.findOne({
+        where: { service: 'transferwise', CollectiveId: collective.id, deletedAt: null },
```
Betree: I don't think that you have to pass `deletedAt: null`, if `paranoid: false` is not added this is normally added by default.

server/graphql/v2/interface/Account.js
```diff
@@ -169,8 +169,16 @@ const accountFieldsDefinition = () => ({
...
+      });
+      if (connectedAccount) {
+        collective.connectedAccount = connectedAccount;
```
Betree: This is a bit hacky, we should avoid adding fields on models that are not defined. What will happen if a different part of the code starts using the same field for storing the Twitter account because it needs it? Without the context, how will the person reading the code for `transferwise/index.ts` understand where the field is defined?
Also it will probably trigger an error if we move the file to typescript. 

What was wrong with loading the proper `connectedAccount` in `transferwise/index.ts` like you did before?

server/lib/transferwise.ts
```diff
@@ -44,9 +44,10 @@ export const createQuote = async (
     });
     return getData(response);
   } catch (e) {
-    const message = `Unable to create quote: ${getAxiosError(e)}`;
-    logger.error(message, data);
-    throw new Error(message);
+    logger.error(`Unable to create quote: ${getAxiosError(e)}`, data);
+    const publicMessage =
+      get(e, 'response.data.errors[0].message') || `Sorry, we can't make transfers to ${targetCurrency}.`;
```
Betree: I think we discussed this before but I don't remember the conclusion, are we 100% sure that the error message from transferwise can't include any sensitive information?

server/paymentProviders/transferwise/index.ts
```diff
@@ -73,56 +79,60 @@ async function payExpense(connectedAccount, payoutMethod, expense): Promise<any>
...
-    .map(c => c.currencyCode);
-  cache.set(cacheKey, currencies, 24 * 60 * 60 /* a whole day and we could probably increase */);
-  return currencies;
+  const currencyInfo = find(await getAvailableCurrencies(host), { code: currency });
+
+  const quote = await transferwise.createQuote(connectedAccount.token, {
+    profileId: connectedAccount.data.id,
+    sourceCurrency: host.currency,
+    targetCurrency: currency,
+    sourceAmount: currencyInfo.minInvoiceAmount * 20,
```
Betree: Won't this will crash if we call `getRequiredBankInformation` with a currency that is not supported?

server/paymentProviders/transferwise/index.ts
```diff
@@ -73,56 +79,60 @@ async function payExpense(connectedAccount, payoutMethod, expense): Promise<any>
...
   const fromCache = await cache.get(cacheKey);
   if (fromCache) {
     return fromCache;
   }
 
-  const connectedAccount = await models.ConnectedAccount.findOne({
-    where: { service: 'transferwise', CollectiveId: host.id, deletedAt: null },
-  });
+  const connectedAccount =
+    host.connectedAccount?.service === 'transferwise'
```
znarf: What is this syntax?

server/paymentProviders/transferwise/index.ts
kewitz: It will throw, but the currencies available are also being generated by TransferWise.
I already blacklisted a few currencies that could present specific restrictions, but if we're aiming to be 100% safe I think we could whitelist just a few currencies we're sure it works.

server/paymentProviders/transferwise/index.ts
kewitz: This is the optional chaining syntax.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

kewitz: This is the optional chaining syntax.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

server/graphql/v2/interface/Account.js
kewitz: Agreed, this implementation is hacky, I'll just use Sequelize's include to preload connected accounts and use `find()` in `transferwise/index.ts`.
The idea here is validating the collective is actually connected to TransferWise before calling related methods so we can cover the case where the collective is not using TW.
It is also an optimization so we can load it once and use it later.

server/lib/transferwise.ts
kewitz: Since errors are not being documented by TransferWise, we're not 100% sure.
By what was experienced during development, although the error contains metadata related to the request, the messages are usually generic.
I'll make sure to throw generic errors on our side, at least for the methods that are being used to generate the form.

server/paymentProviders/transferwise/index.ts
Betree: It's always better to gracefully fail, we could just add above:

```ts
const currencyInfo = find(await getAvailableCurrencies(host), { code: currency });
if (!currencyInfo) {
  throw new Error('This currency is not supported');
}
```

server/paymentProviders/transferwise/index.ts
Betree: @znarf apparently our Babel version already support it too, so we can also use it in API & Frontend JS files

PR #3422, opened 2020-02-26
chore: refresh contributor cache when new member added to Collective

bolariin:
Resolves https://github.com/opencollective/opencollective/issues/2494

### DIFF
--- a/server/graphql/v1/mutations/members.js
+++ b/server/graphql/v1/mutations/members.js
@@ -1,6 +1,5 @@
 import models from '../../../models';
 import errors from '../../../lib/errors';
-import { invalidateContributorsCache } from '../../../lib/contributors';
 
 /** A mutation to edit the public message of all matching members. */
 export async function editPublicMessage(_, { FromCollectiveId, CollectiveId, message }, req) {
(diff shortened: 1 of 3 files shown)


bolariin:
> Isn't there an `afterUpdate` hook or similar?

there is an `afterUpdate` hook I can replace it with. I will make that change

server/graphql/v1/mutations/members.js
```diff
@@ -27,6 +26,5 @@ export async function editPublicMessage(_, { FromCollectiveId, CollectiveId, mes
    * After updating the public message it is necessary to update the cache
    * used in the collective page.
    */
```
Betree: We should remove this comment because we're not updating the cache anymore here

server/models/Member.js
```diff
@@ -110,6 +111,18 @@ export default function(Sequelize, DataTypes) {
...
       },
+      hooks: {
+        afterCreate: instance => {
+          if (instance.role !== roles.FOLLOWER) {
+            invalidateContributorsCache(instance.CollectiveId);
+          }
+          return null;
+        },
+        afterUpdate: instance => {
+          invalidateContributorsCache(instance.CollectiveId);
```
Betree: Can we add the same `if (instance.role !== roles.FOLLOWER)` check here?

bolariin: I thought about a case, not sure it is valid. I was wondering in the case that role of the member is updated to `FOLLOWER`, wouldn't we want to update the contributors to account for the change?

Betree: We don't use the role `FOLLOWER` yet (it will be implemented with https://github.com/opencollective/opencollective/issues/1445) but I don't think that we'll have such cases, if an existing member wants to follow a collective then we'll probably create a new `Member` entry rather than updating the existing one.

PR #3427, opened 2020-02-27
fix: add await on expense createActivity to resolve deadlock in tests

bolariin:
Resolves https://github.com/opencollective/opencollective/issues/2906


znarf:
Thank you so much @bolariin. It's merged.

PR #3434, opened 2020-02-27
refactor(applications): check CollectiveId instead of CreatedByUser

bolariin:
resolves https://github.com/opencollective/opencollective/issues/2029

### DIFF
--- a/migrations/20200308233148-enforce-applications-collective-id.js
+++ b/migrations/20200308233148-enforce-applications-collective-id.js
@@ -1,0 +1,24 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(`
+      UPDATE ONLY "Applications" a
+      SET "CollectiveId" = u."CollectiveId"
+      FROM "Users" u
+      WHERE u.id = a."CreatedByUserId"
+      AND a."CollectiveId" IS NULL;
+    `);
+
+    await queryInterface.sequelize.query(`
+      ALTER TABLE "Applications" ALTER COLUMN "CollectiveId" SET NOT NULL;
+    `);
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    // We can't completely revert
+    await queryInterface.sequelize.query(`
+      ALTER TABLE "Applications" ALTER COLUMN "CollectiveId" INT NULL;
+    `);
+  },
+};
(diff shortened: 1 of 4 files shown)


bolariin:
> We have two applications in the database with `CollectiveId` = NULL (sounds like something that we should not allow in the first place). We should do something about them before merging, otherwise we're going to break them.

Would it possible to update those rows?

> ), tho I think should probably also hide the whole list in the `applications` resolver if you're not allowed to - I don't see any benefit in allowing other users to list your applications (id, type, name, description).

I'm not sure I completely understand when you say hide the whole list in the `applications` resolver

Betree:
> Would it possible to update those rows?

Yes, the best to do that is to create a migration with `npm run db:migration:create -- --name enforce-applications-collective-id`.

You can use the following query to update the applications:

```sql
UPDATE ONLY "Applications" a
SET "CollectiveId" = u."CollectiveId"
FROM "Users" u 
WHERE u.id = a."CreatedByUserId" 
AND a."CollectiveId" IS NULL
```

Then, in the same migration. we should also add a `NOT NULL` constraint to the `Applications.CollectiveId` column to make sure it doesn't happens in the future. Let me know if you need more info on that.

> I'm not sure I completely understand when you say hide the whole list in the applications resolver

In my opinion, there's no reason to allow users to list someone else's applications. So in the application resolver (https://github.com/opencollective/opencollective-api/blob/0f710e52086298cd7b1255a8eb34a8935978b5d2/server/graphql/v1/CollectiveInterface.js#L1798) I'd rather add:

```es6
applications: {
  type: new GraphQLList(ApplicationType),
  resolve(userCollective, _, req) {
    if (req.remoteUser && req.remoteUser.isAdmin(userCollective.id)) {
      return models.Application.findAll({
        where: { CollectiveId: userCollective.id },
      });
    }
  },
},
```

znarf:
@Betree the two applications mentioned are different from the others, we should double check their use case internally.

Betree:
@znarf You mean by contacting the users?

In any case I don't think that it should prevent us from fixing them, adding their `CollectiveId` should not break anything.

znarf:
@Betree I mean: let's discuss these specific cases elsewhere.

Betree:
@znarf Yes no problem with that, but I still think that we should create this migration no? It doesn't cost much and adding a constraint to the DB will make sure that this doesn't happen again in the future. It's also a requirement for this PR to solve it before merging because otherwise we will break these two applications.

znarf:
Thank you @bolariin. Merging!

PR #3440, opened 2020-02-28
Tweak GitHub scope based on context

znarf:
Looks like `public_repo` is not needed and it's the controversial permission (for good reason).


znarf:
Fix: https://github.com/opencollective/opencollective/issues/1034

PR #3441, opened 2020-02-28
Reset minimumAmount and presets for FIXED tiers

znarf:
Fix: https://github.com/opencollective/opencollective/issues/2939

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1861,6 +1861,10 @@ export default function(Sequelize, DataTypes) {
       })
       .then(() => {
         return Promise.map(tiers, tier => {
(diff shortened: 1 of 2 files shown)


server/models/Collective.js
```diff
@@ -1862,6 +1862,10 @@ export default function(Sequelize, DataTypes) {
       .then(() => {
         return Promise.map(tiers, tier => {
           if (tier.id) {
+            if (tier.amountType === 'FIXED') {
```
Betree: Why doing it only if the tier has an id?

znarf: Good call, doesn't hurt to also do that for creations.

PR #3442, opened 2020-02-28
Disable expiring cards cron until we set a proper schedule

znarf:
we decided to not sent the 1st of the month, which is tomorrow


sbinlondon:
Looks good to me - will open an issue and PR fixing the dates and adding a reminder

znarf:
Merging, this needs to go live today.

PR #3444, opened 2020-02-28
GQLV2 endpoints for the new expense page

Betree:
## New pattern: fetch permissions from frontend

This PR introduces a pattern that we haven't used before: providing a map of the permissions to the client (see `ExpensePermissions`) as the single source of truth for checking what a user can do with an object. I believe this is something we should tend towards for future developments for the following reasons:
- Permissions is something that will never change from one frontend to another
- Ability to check more complex scenarios without fetching additional data from the frontend
- Only one code to check permissions everywhere = easier to test, less chances of doing it wrong
- Never hide a security issue: if button is hidden, API shouldn't accept the action nether

## Changes in context-permissions

I've updated the documentation for `server/graphql/common/context-permissions.ts` to better explain the what & why.

`setContextPermission` has been renamed to `allowContextPermission` and the value parameter has been removed because setting this field to `false` usually doesn't make sense:

- `getContextPermission` already returns a falsy value by default
- If another parent up the tree set this to true, it makes no sense to remove this permission down the tree

### DIFF
--- a/server/constants/expense_type.js
+++ b/server/constants/expense_type.js
@@ -6,4 +6,5 @@
 export default {
   INVOICE: 'INVOICE',
   RECEIPT: 'RECEIPT',
+  UNCLASSIFIED: 'UNCLASSIFIED',
 };
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -369,7 +369,9 @@ export async function deleteExpense(remoteUser, expenseId) {
   }
 
   if (!canDeleteExpense(remoteUser, expense)) {
-    throw new errors.Unauthorized("You don't have permission to delete this expense");
+    throw new errors.Unauthorized(
+      "You don't have permission to delete this expense or it needs to be rejected before being deleted",
+    );
   }
 
   const res = await expense.destroy();
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -17,21 +17,9 @@ import FEATURE from '../../../constants/feature';
 import { FeatureNotAllowedForUser, ValidationFailed } from '../../errors';
 import { PayoutMethodTypes } from '../../../models/PayoutMethod';
 import { types as collectiveTypes } from '../../../constants/collectives';
+import { canUpdateExpenseStatus, canEditExpense, canDeleteExpense } from '../../common/expenses';
 
 const debug = debugLib('expenses');
-
-/**
- * Only admin of expense.collective or of expense.collective.host can approve/reject expenses
- */
-function canUpdateExpenseStatus(remoteUser, expense) {
-  if (remoteUser.hasRole([roles.ADMIN], expense.CollectiveId)) {
-    return true;
-  }
-  if (remoteUser.hasRole([roles.ADMIN], expense.collective.HostCollectiveId)) {
-    return true;
-  }
-  return false;
-}
 
 /**
  * Only admin of expense.collective.host can mark expenses unpaid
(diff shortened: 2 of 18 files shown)


server/graphql/v2/input/ExpenseReferenceInput.ts
```diff
@@ -0,0 +1,32 @@
+import { GraphQLInt, GraphQLString, GraphQLInputObjectType } from 'graphql';
+import { idDecode, IDENTIFIER_TYPES } from '../identifiers';
+
+const ExpenseReferenceInput = new GraphQLInputObjectType({
```
Betree: I've tried to follow the pattern defined in https://github.com/opencollective/opencollective/issues/2935 for this one. Let me know if `Reference` is not the final name.

server/constants/expense_type.js
```diff
@@ -6,4 +6,5 @@
 export default {
   INVOICE: 'INVOICE',
   RECEIPT: 'RECEIPT',
+  UNCLASSIFIED: 'UNCLASSIFIED',
```
Betree: This wasn't in the enum, yet we have entries for it in the DB.

server/graphql/v1/mutations/expenses.js
```diff
@@ -395,14 +368,10 @@ export async function deleteExpense(remoteUser, expenseId) {
     throw new errors.NotFound('Expense not found');
   }
 
-  if (!canEditExpense(remoteUser, expense)) {
+  if (!canDeleteExpense(remoteUser, expense)) {
     throw new errors.Unauthorized("You don't have permission to delete this expense");
```
kewitz:
```suggestion
    throw new errors.Unauthorized("You don't have permission to delete this expense or this expense needs to be rejected before being deleted");
```

server/graphql/v2/input/CommentCreate.js
```diff
@@ -1,15 +1,22 @@
...
  * Input type to use as the type for the comment input in createComment mutation.
  */
 const CommentCreate = new GraphQLInputObjectType({
   name: 'CommentCreate',
   fields: () => ({
-    markdown: { type: GraphQLString },
+    markdown: { type: GraphQLString, deprecationReason: '2020-02-26: Please use html' },
     html: { type: GraphQLString },
-    ExpenseId: { type: GraphQLInt },
-    UpdateId: { type: GraphQLInt },
```
kewitz: Are we dropping `UpdateId`?

server/graphql/v2/object/PayoutMethod.ts
```diff
@@ -10,18 +11,43 @@ const PayoutMethod = new GraphQLObjectType({
...
+        }
+      },
     },
     data: {
       type: GraphQLJSON,
+      description: 'The actual data for this payout method. Content depends on the type.',
+      resolve: (payoutMethod, _, req): object => {
+        if (
+          req.remoteUser?.isAdmin(payoutMethod.CollectiveId) ||
+          getContextPermission(req, PERMISSION_TYPE.SEE_PAYOUT_METHOD_DATA, payoutMethod.id)
```
kewitz: Isn't the `isAdmin` redundant in this case?

server/graphql/v2/input/CommentCreate.js
Betree: Right now it's not possible to add comments on updates, this was added in the expectation that we should be able to do it in the future. We'll revisit once comments on updates will be possible.

server/graphql/v2/object/PayoutMethod.ts
Betree: The context permission is about whitelisting the field for certain contexts (in this case, an expense where you would be an admin of the host). But no matter what the context is, you should always have access to your own payout method's data.

For example when you list the payout methods for a collective you own with the following query, the permission should not depend on the context - you're always allowed to see your own payout methods no matter what the context is.

```gql
loggedInUser {
  payoutMethods {
    data // Should have access, because user is the owner
  }
}
```

PR #3449, opened 2020-02-28
Fix: Stubs, Sandboxes and ESLint

kewitz:
I would also like to drop the warning for functions without explicit returning types because TS can automatically validate that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -172,7 +172,7 @@
     "forestadmin-schema:update": "npm run script scripts/forestadmin-schema-update && prettier .forestadmin-schema.json --write",
     "git:clean": "./scripts/git_clean.sh",
     "hint": "npm run lint:quiet",
-    "lint": "eslint .",
+    "lint": "eslint '**/*.{js,ts,jsx,tsx}'",
     "lint:check": "npm run lint -- --quiet",
     "lint:fix": "npm run lint -- --fix",
     "maildev": "maildev",
--- a/package.json
+++ b/package.json
@@ -172,7 +172,7 @@
     "forestadmin-schema:update": "npm run script scripts/forestadmin-schema-update && prettier .forestadmin-schema.json --write",
     "git:clean": "./scripts/git_clean.sh",
     "hint": "npm run lint:quiet",
-    "lint": "eslint '**/*.{js,ts,jsx,tsx}'",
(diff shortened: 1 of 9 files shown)


Betree:
@kewitz thanks for taking care of that!

I personally like the fact that it enforces to set the return type for functions. But I'm not sure what we're referring to here, can you provide a short example as well as arguments on why we should drop it?

package.json
```diff
@@ -172,7 +172,7 @@
     "forestadmin-schema:update": "npm run script scripts/forestadmin-schema-update && prettier .forestadmin-schema.json --write",
     "git:clean": "./scripts/git_clean.sh",
     "hint": "npm run lint:quiet",
-    "lint": "eslint .",
+    "lint": "eslint '**/*.{js,ts,jsx,tsx}'",
```
znarf: What about `"lint": "eslint . --ext=js,ts"` instead?

kewitz:
@Betree mainly because TypeScript is really good at deducing the return type by itself:
image
Especially when you use a lot of native operators and functions, which is pretty much our case.
The exception is when you use a library that is not typed, in that case, TS wouldn't be able to deduce the output type.

package.json
kewitz: Yep, simpler.

Betree:
@kewitz I'm ok to have it just as a warning, but I don't like the idea of removing the check entirely because explicit returns type are a good practice. They make sure that the calling code can expect a stable API from your function.

Let's take an example:
```ts
// Implements this **in a typescript file**
const myFunc = arg => {
  return Promise.resolve(42);
}

// Implements some code **in a JS file**
myFunc().then(console.log);

// Update first example
const myFunc = arg => {
  return arg ? Promise.resolve(42) : 55;
}
```

Now if you use explicit types Typescript will prevent you from doing the latest, because you've specified that `myFunc` should always return a promise. If you don't use explicit types, the type check will pass and the failure will only happen at runtime. 

Also if we ever want to move to strict typescript flags in the future (which I hope we'll be able to do one day) implicit returns types are not allowed so we'll have to go back and type everything. I'd rather keep a good habit from the start.

kewitz:
@Betree you're right, in this mixed JS/TS context it makes sense.

PR #3450, opened 2020-02-29
fix: send donation confirmation on donation to an event

bolariin:
resolves https://github.com/opencollective/opencollective/issues/2891


znarf:
Thank you so much @bolariin. Merging!

PR #3458, opened 2020-03-02
Update expiring cards dates and add reminder

sbinlondon:
References opencollective/opencollective#2940

Just a small change:

* Change date to 7th/21st for emails
* Adds `reminder` case to code and to email template

Tested locally and once you update your credit card after the first email, it means you won't get the reminder email during the second run, so no extra code is needed.

Let me know if there are any changes needed and if not, I can move it back to the `cron/monthly` folder.

### DIFF
--- a/cron/disabled/expiring-cards.js
+++ b/cron/disabled/expiring-cards.js
@@ -40,8 +40,7 @@ const fetchExpiringCreditCards = async (month, year) => {
 
 const run = async () => {
   const cards = await fetchExpiringCreditCards(month, year);
-  let reminder;
-  date === 21 ? (reminder = true) : (reminder = false);
+  const reminder = date === 21 ? true : false;
 
   for (const card of cards) {
     try {
--- a/cron/disabled/expiring-cards.js
+++ b/cron/disabled/expiring-cards.js
@@ -5,14 +5,14 @@ import logger from '../../server/lib/logger';
 import models from '../../server/models';
 import * as libPayments from '../../server/lib/payments';
 
-// Only run on the first of the month
+// Run on the 7th and 21st of the month
 const today = new Date();
 const date = today.getDate();
 const month = today.getMonth() + 1;
 const year = today.getFullYear();
 
-if (process.env.NODE_ENV === 'production' && date !== 1 && !process.env.OFFCYCLE) {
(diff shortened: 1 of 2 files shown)


cron/disabled/expiring-cards.js
```diff
@@ -40,6 +40,8 @@ const fetchExpiringCreditCards = async (month, year) => {
 
 const run = async () => {
   const cards = await fetchExpiringCreditCards(month, year);
+  let reminder;
+  date === 21 ? (reminder = true) : (reminder = false);
```
znarf: `const reminder = date === 21 ? true : false;`

templates/emails/payment.creditcard.expiring.hbs
```diff
@@ -18,10 +22,19 @@ Subject: Your {{brand}} ending in {{name}} is about to expire
 <p>Hello,</p>
 {{/if}}
 
+
+{{#if reminder}}
```
znarf: Maybe split in 2 paragraphs so we don't have to duplicate too much the copy?

cron/disabled/expiring-cards.js
```diff
@@ -5,14 +5,14 @@ import logger from '../../server/lib/logger';
...
-// Only run on the first of the month
+// Run on the 7th and 21st of the month
 const today = new Date();
 const date = today.getDate();
 const month = today.getMonth() + 1;
 const year = today.getFullYear();
 
-if (process.env.NODE_ENV === 'production' && date !== 1 && !process.env.OFFCYCLE) {
-  console.log('NODE_ENV is production and today is not the first of month, script aborted!');
+if (process.env.NODE_ENV === 'production' && (date !== 7 || date !== 21) && !process.env.OFFCYCLE) {
```
Betree: Seems to me that this condition will never be falsy, for example with date=21:

```
// true                                   && (true       || false      ) && true
if (process.env.NODE_ENV === 'production' && (date !== 7 || date !== 21) && !process.env.OFFCYCLE) {
```

sbinlondon: Apologies - the || here should be an &&, correct?

znarf:
Let's merge like that and send the next batches manually.

We'll then activate the job if everything goes well.

PR #3460, opened 2020-03-02
TransferWise: Add Webhook Endpoint

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/2942


data/transferwise.webhook.live.pub
```diff
@@ -0,0 +1,9 @@
+-----BEGIN PUBLIC KEY-----
+MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAvO8vXV+JksBzZAY6GhSO
+XdoTCfhXaaiZ+qAbtaDBiu2AGkGVpmEygFmWP4Li9m5+Ni85BhVvZOodM9epgW3F
+bA5Q1SexvAF1PPjX4JpMstak/QhAgl1qMSqEevL8cmUeTgcMuVWCJmlge9h7B1CS
+D4rtlimGZozG39rUBDg6Qt2K+P4wBfLblL0k4C4YUdLnpGYEDIth+i8XsRpFlogx
+CAFyH9+knYsDbR43UJ9shtc42Ybd40Afihj8KnYKXzchyQ42aC8aZ/h5hyZ28yVy
+Oj3Vos0VdBIs/gAyJ/4yyQFCXYte64I7ssrlbGRaco4nKF3HmaNhxwyKyJafz19e
+HwIDAQAB
+-----END PUBLIC KEY-----
```
znarf: Can we think about another place for this file? This data folder doesn't sound ideal (and it's actually on my kill list).

server/lib/express.js
```diff
@@ -37,7 +37,17 @@ export default function(app) {
   }
 
   // Body parser.
-  app.use(bodyParser.json({ limit: '50mb' }));
+  app.use(
+    bodyParser.json({
+      limit: '50mb',
+      verify(req, res, buf) {
+        const url = req.originalUrl;
```
znarf: Can you add a comment block explaining that?

data/transferwise.webhook.live.pub
kewitz: 1. `lib/transferwise.ts` into `lib/transferwise/index.ts` and put the keys inside the same folder.
2. Have these keys as strings in `lib/transferwise.ts`.
3. Something else.

kewitz:
Rebased.

PR #3472, opened 2020-03-04
fix(postgres): fix implicit disabling of certificate verification

bolariin:
resolves https://github.com/opencollective/opencollective/issues/2897

### DIFF
--- a/config/production.json
+++ b/config/production.json
@@ -8,7 +8,9 @@
         "acquire": 60000
       },
       "dialectOptions": {
-        "ssl": true
+        "ssl": {
+          "rejectUnauthorized": false
+        }
       },
       "benchmark": true
     }


config/production.json
```diff
@@ -8,7 +8,9 @@
         "acquire": 60000
       },
       "dialectOptions": {
-        "ssl": true
+        "ssl": {
+          "rejectUnauthorized": true
```
kewitz: If we set it to `true` we'll need to supply the certificate, otherwise, the connection won't work.
If we want to explicitly disable, this should be set to false.

znarf: Good catch @kewitz ! I looked at it many times and missed it.

kewitz: Yes, this argument is passed to `pg` which passes to Node's `tls` library, you can check it here:
https://nodejs.org/api/tls.html#tls_tls_connect_options_callback
Notice that our Postgres host is `amazonaws.com` if you're going after that certificate.

PR #3473, opened 2020-03-04
Simpler deployment process

znarf:
As discussed, the current deployment process is complex for no real upside. It's difficult to follow the succeeding merges on `staging`, `production` and `master` branches. Then, respective pushes to `github` and `heroku`.

As no one ever complained about deployment on the frontend side, this PR is bringing the simpler process to API.


sbinlondon:
Looks good! No complaints from me about a simpler deploy process 😊

PR #3489, opened 2020-03-06
Add auth-simple-oauth-app for unauthenticated requests

znarf:
Unauthenticated requests were broken in last update. This is an attempt at fixing it.

See: https://github.com/octokit/rest.js/issues/1636


server/lib/github.js
```diff
@@ -13,7 +13,6 @@ const compactRepo = repo => {
     'description', // (1)
     'owner', // (1) (4)
     'stargazers_count', // (1) (2) (4)
-    'topics', // (1)
```
znarf: Removing topics as it's not needed anymore.

server/lib/github.js
```diff
@@ -31,12 +30,10 @@ const compactRepo = repo => {
 export function getOctokit(accessToken) {
   const octokitParams = {};
 
-  octokitParams.previews = ['mercy-preview'];
```
znarf: Topics being removed, we can remove this preview header.

znarf:
Merging to give it a try.

PR #3494, opened 2020-03-09
Migrate JSON columns to JSONB

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/2974

### DIFF
--- a/migrations/20200309140500-JSONB.js
+++ b/migrations/20200309140500-JSONB.js
@@ -1,0 +1,161 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    /** Activity */
+    await queryInterface.changeColumn('Activities', 'data', {
+      type: Sequelize.JSONB,
+    });
+
+    /** Collective */
+    await queryInterface.changeColumn('Collectives', 'data', {
+      type: Sequelize.JSONB,
+    });
+    await queryInterface.changeColumn('CollectiveHistories', 'data', {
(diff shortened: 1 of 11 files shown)


znarf:
@kewitz how much time do you estimate the migration to need in production?

kewitz:
@znarf that's the question. Any straightforward way to simulate/estimate this?

kewitz:
I think this should take a couple of minutes, tops.
From https://stackoverflow.com/a/50715139:
> I don't know how this compares to the accepted answer performance-wise, but I tested this on a table with 120 000 records, each record having four json columns and it took me about a minute to migrate that table. Of course, I guess it depends on how complex the json structure is.

kewitz:
On my old Macbook Pro i5, dockered postgres.
```
leo in ~/Code/opencollective/api on ⌥ fix/move-to-jsonb > PG_DATABASE=opencollective_prod_snapshot npm run db:migrate

> opencollective-api@2.0.567 db:migrate /Users/leo/Code/opencollective/api
> ./scripts/sequelize.sh db:migrate

babel-node --extensions .js,.ts ./node_modules/.bin/sequelize --models-path server/models/ --config config/sequelize-cli.js db:migrate

Sequelize CLI [Node: 12.14.0, CLI: 5.5.1, ORM: 5.21.5]

Loaded configuration file "config/sequelize-cli.js".
== 20200309140500-JSONB: migrating =======
== 20200309140500-JSONB: migrated (238.733s)
```

PR #3500, opened 2020-03-10
Add email field in the email template sent to contributors

JoyLubega:
Resolve: opencollective/opencollective/issues/2297

Screen shots
Screen Shot 2020-03-12 at 12 16 03 AM

### DIFF
--- a/server/lib/notifications.js
+++ b/server/lib/notifications.js
@@ -145,7 +145,6 @@ async function notifyUserId(UserId, activity, options = {}) {
 export async function notifyAdminsOfCollective(CollectiveId, activity, options = {}) {
   debug('notify admins of CollectiveId', CollectiveId);
   const collective = await models.Collective.findByPk(CollectiveId);
-
   if (!collective) {
     throw new Error(
       `notifyAdminsOfCollective> can't notify ${activity.type}: no collective found with id ${CollectiveId}`,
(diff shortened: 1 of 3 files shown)


JoyLubega:
hello @znarf  I request for some feedback on this PR.

server/lib/notifications.js
```diff
@@ -214,6 +215,10 @@ async function notifyByEmail(activity) {
 
     case activityType.COLLECTIVE_MEMBER_CREATED:
       twitter.tweetActivity(activity);
+      if (activity.data.member.CreatedByUserId) {
+        const userAdmin = await models.User.findByPk(activity.data.member.CreatedByUserId);
```
znarf, line 219: `CreatedByUserId` should not be used because there is no guarantee that this what you need here.

You should look at the `memberCollective` and query the User table with its `CollectiveId`

znarf: There is also no guarantee that the member is an user/individual and that userAdmin will exist, you should take that into account.

templates/emails/collective.member.created.hbs
```diff
@@ -11,7 +11,7 @@ Subject: {{member.memberCollective.name}} joined {{collective.name}} as {{{toLow
 {{/if}}
 
 <p>Just a quick note to let you know that {{#if member.memberCollective.isIncognito}}someone (incognito){{else}}<a
-    href="{{config.host.website}}/{{member.memberCollective.slug}}">{{member.memberCollective.name}}</a>{{/if}} is now a
+    href="{{config.host.website}}/{{member.memberCollective.slug}}">{{member.memberCollective.name}} </a> with email: {{member.memberCollective.email}} {{/if}} is now a
```
znarf, line 14: There is no guarantee that email will be there, there should be a condition.

JoyLubega:
I think I am going to query the email field straight from the model

Betree:
The issue has been resolved in https://github.com/opencollective/opencollective-api/pull/4001. 
Thank you for the effort, the adopted solution was actually pretty close from what we have here!

PR #3508, opened 2020-03-11
Enforce Unique Token/Service ConnectedAccount Combo

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/2957

### DIFF
--- a/server/models/ConnectedAccount.js
+++ b/server/models/ConnectedAccount.js
@@ -43,6 +43,13 @@ export default (Sequelize, DataTypes) => {
     {
       paranoid: true,
 
+      indexes: [
+        {
+          unique: true,
+          fields: ['service', 'token'],
+        },
+      ],
+
       getterMethods: {
         info() {
           return {
--- a/migrations/20200311112600-remove-duplicated-connected-accounts.js
+++ b/migrations/20200311112600-remove-duplicated-connected-accounts.js
@@ -1,0 +1,22 @@
+'use strict';
+
+module.exports = {
+  up: async queryInterface => {
+    await queryInterface.sequelize.query(`
+      DELETE * FROM "ConnectedAccounts" WHERE id NOT IN (
+        SELECT DISTINCT ON (token, service, username) id
+        FROM "ConnectedAccounts"
+        WHERE "deletedAt" IS NULL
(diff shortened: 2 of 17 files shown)


server/models/ConnectedAccount.js
```diff
@@ -43,6 +43,13 @@ export default (Sequelize, DataTypes) => {
     {
       paranoid: true,
 
+      indexes: [
```
Betree: Is this enough to create the index in DB, no need to create a migration?

kewitz: Nice catch @Betree! It is not needed if you use `db.sync()`, since we don't, I'll add it to the migration.

kewitz:
Rebased.

kewitz:
@znarf @Betree: what should we do with this migration? Are we OK with restricting this through the DB or should I manually validate this through the GraphQL API for TransferWise only?

Betree:
I am. DB constraints are always better, they make sure that the data stays consistent no matter what we do in the code.

Are there any cases or existing data that doesn't play nice with it?

znarf:
I'm just unsure about the deletion policy. For example, we need to protect Twitter connectedAccounts that are actively used and configured for automated tweets.

kewitz:
I agree with @znarf, this is a legit use for duplicated tokens. I'll implement this in the mutator level for now.

znarf:
@kewitz we still want that, need to go extra mile and figure a sensible deletion policy.

kewitz:
Cool, I'll generate a report and properly analyze all cases this week.

znarf:
@kewitz I'm gonna work on that.

kewitz:
Thanks @znarf!

znarf:
I'm almost done for Twitter, main strategy is:
- Ignore deleted ConnectedAccounts and Collectives
- Delete duplicates that are not Collective or User=isHost or Organization=isHost (there is no point connecting Twitter out of these cases)
- Delete duplicates that don't have Tweets configured or are explicitly turned off
- Deletion of the least used/useful (no transaction, no tweet going out)

With this strategy, I'm done to 1 duplicate.

znarf:
Stale, closing.

PR #3509, opened 2020-03-11
feat(collective): update currency handler

joshikeno:
- add checks for currency change and invoke handler function
    - return error if conditions to update currency not meet
Re [#3607](https://github.com/opencollective/opencollective-frontend/pull/3607)
Part of [#2730](https://github.com/opencollective/opencollective/issues/2730)

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1532,6 +1532,16 @@ export default function(Sequelize, DataTypes) {
     return this;
   };
 
+  Collective.prototype.updateCurrency = async function(currency, remoteUser) {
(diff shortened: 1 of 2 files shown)


server/models/Collective.js
```diff
@@ -1532,6 +1532,16 @@ export default function(Sequelize, DataTypes) {
     return this;
   };
 
+  Collective.prototype.updateCurrency = async function(currency, remoteUser) {
+    if (typeof currency === undefined || !remoteUser || currency === this.currency || !remoteUser.isAdmin(this.id)) {
+      return;
+    }
+    if ([types.COLLECTIVE, types.EVENT].includes(this.type) || (await this.isHost())) {
```
znarf: Test if User or Organization instead of excluding Collective and Event

PR #3524, opened 2020-03-16
Remove wild console.log

kewitz:
Found a huge array of transactions being logged on PaperTrail.
I think this should solve it.

image

You can log in to PaperTrail through Heroku dashboard and visit  https://my.papertrailapp.com/systems/opencollective-prod-api/events?focus=1176217716166164496&selected=1176217716166164496.


kewitz:
Rebased.

PR #3525, opened 2020-03-16
Optimize TransferWise webhook transaction query

kewitz:
Actually tested with production DB clone.
I feel like I overcomplicated the first time trying to make it more consistent with the webhook query, but in the end, these are totally different logics.


kewitz:
Rebased.

kewitz:
@Betree yes, it is taking around 1~2 seconds, which is acceptable.
Turn out I had two problems and the prior migration was fixing just one.

PR #3526, opened 2020-03-16
Fix: rewrite unreliable snapshot test

kewitz:
The difference is being caused by transactions swapping position in the returning array, which by the code we use to generate these transactions is acceptable.
I don't think this reflects some actual major problem and I recommend deprecating snapshot tests.


kewitz:
Resolves https://github.com/opencollective/opencollective-api/pull/3526.

PR #3537, opened 2020-03-18
feat(event): make event hosted, approved and active when unarchiving

flickz:
Following up https://github.com/opencollective/opencollective-frontend/pull/3630#issuecomment-599660467

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -662,10 +662,11 @@ export async function unarchiveCollective(_, args, req) {
   }
 
   if (collective.type === types.EVENT) {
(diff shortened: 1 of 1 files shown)


server/graphql/v1/mutations/collectives.js
```diff
@@ -661,6 +661,15 @@ export async function unarchiveCollective(_, args, req) {
     });
   }
 
+  if (collective.type === types.EVENT) {
+    return collective.update({
+      deactivatedAt: null,
+      isActive: true,
+      HostCollectiveId: collective.ParentCollectiveId,
```
znarf: That's not correct. HostCollectiveId should be the `HostCollectiveId` of the parent collective.

PR #3545, opened 2020-03-19
GQLV2: Add mutation to edit collective settings

Betree:
This PR adds a mutation to edit collective settings. I've chosen to do that by creating an `editAccountSetting` that only updates one key in the settings rather than passing the full object like we do today in GQLV1:

```es6
editAccountSetting({ account, key: 'tos', value: 'https://opencollective.com/tos' })
```

The reasons for this choice are:
1. **It prevent bugs from stripping the full settings object**
For example if we do `setSettings({ ...settings, new: 'value' })` in frontend but we forgot to fetch `settings`, it could strip all other keys.
2. **It prevent from saving outdated settings**
With the v1 `editSettings` mutation, if you load two settings page, save the first one then the second your first save will be stripped out because the second page has no way to be notified that the data changed.
3. **It makes it easy to introduce whitelisting**, because we can start applying it on new entries today without affecting the existing settings. After merging I'll create a proposal issue to cleanup the keys an normalize a bit how we use `settings`. For now I've whitelisted all the existing keys using the following query:
```sql
SELECT DISTINCT jsonb_object_keys(settings) FROM "Collectives"
```


server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -0,0 +1,37 @@
+import { GraphQLNonNull } from 'graphql';
+import GraphQLJSON from 'graphql-type-json';
+import { AccountReferenceInput, fetchAccountWithReference } from '../input/AccountReferenceInput';
+import { Account } from '../interface/Account';
+import { Unauthorized, Forbidden } from '../../errors';
+
+const accountMutations = {
+  editAccountSettings: {
+    type: new GraphQLNonNull(Account),
```
znarf: Is that best practice vs returning the settings?

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -0,0 +1,37 @@
...
+        throw new Unauthorized();
+      }
+
+      const account = await fetchAccountWithReference(args.account, { loaders: req.loaders, throwIfMissing: true });
+
+      if (!req.remoteUser.isAdmin(account.id)) {
+        throw new Forbidden();
+      }
+
+      return account.update({ settings: args.settings });
```
znarf: Should we consider any blacklist/whitelist strategy?

Should we consider merging by default? To not erase existing settings?

Betree: A work has already been started to implement whitelisting at the collective level: https://github.com/opencollective/opencollective-api/blob/a0400e1f4b18cf259a0f22c9f9449c463f5f242c/server/models/Collective.js#L317. We need to improve that because it's not a hard whitelist right now, but it's a different issue: we must ensure we don't break any existing settings. A problem induced by whitelisting is that any change will then require an API update, it's handy to be able to update the settings without touching the API.

Regarding the merge vs erase, while I agree it's safer, the major issue is that it will prevent keys from being removed.

server/graphql/v2/mutation/AccountMutations.ts
Betree: If we just return the `settings` object we'll need to update Apollo's cache manually. By returning the full account Apollo will be able to update it automatically.

I have no strong opinion on which one is cleaner, but I like the idea that any returned entity should have a unique identifier. In the case of `settings`, its identifier is the account id.

server/graphql/v2/mutation/AccountMutations.ts
Betree: The new `editAccountSetting` whitelists the `key` being submitted.

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -0,0 +1,50 @@
...
+          dbTransaction: transaction,
+          lock: true,
+          throwIfMissing: true,
+        });
+
+        if (!req.remoteUser.isAdmin(account.id)) {
+          throw new Forbidden();
+        }
+
+        const settings = { ...account.settings, [args.key]: args.value };
```
kewitz: We should use `lodash.set` to support nested properties.

Betree: Interesting idea, I didn't think about that. It will make whitelisting and usage a bit less straightforward because we won't be able to use the `AccountSettingsKey` enum anymore for the `key` field but updating nested values will be safer.

@opencollective/core-developers any opinion on this?

Betree: I made the update, the safety and usage are really better when we allow nested keys.

```es6
editAccountSetting({ account, key: 'collectivePage.background.zoom', value: 0.5 }) 
```

Thanks @kewitz for the suggestion :+1:

PR #3546, opened 2020-03-20
Collective Sponsorship

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/2985.

The idea is adding the flat `sponsored = true` to collective.data for when we want to waive fees and be able to track orders that were sponsored.

**TODO**
- [x] Waive fee from sponsored collectives
- [x] Add resolvers for sponsorship values.

### DIFF
--- a/migrations/20200323163000-collective-platform-fee.js
+++ b/migrations/20200323163000-collective-platform-fee.js
@@ -1,0 +1,21 @@
+'use strict';
+
+const OC_FEE_PERCENT = 5;
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('Collectives', 'platformFeePercent', {
+      type: Sequelize.INTEGER,
+      defaultValue: OC_FEE_PERCENT,
+    });
+    await queryInterface.addColumn('CollectiveHistories', 'platformFeePercent', {
+      type: Sequelize.INTEGER,
+      defaultValue: OC_FEE_PERCENT,
+    });
+  },
+
+  down: async queryInterface => {
+    await queryInterface.removeColumn('Collectives', 'platformFeePercent');
+    await queryInterface.removeColumn('CollectiveHistories', 'platformFeePercent');
+  },
+};
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -51,7 +51,7 @@ import { capitalize, flattenArray, getDomain, formatCurrency, cleanTags, md5, st
 
 import roles, { MemberRoleLabels } from '../constants/roles';
 import activities from '../constants/activities';
-import { HOST_FEE_PERCENT } from '../constants/transactions';
+import { HOST_FEE_PERCENT, OC_FEE_PERCENT } from '../constants/transactions';
 import { types } from '../constants/collectives';
 import expenseStatus from '../constants/expense_status';
 import expenseTypes from '../constants/expense_type';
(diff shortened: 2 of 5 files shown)


xdamman:
This is great! It will really save a lot of (error prone) admin work as more people create #COVID19 related collectives.

znarf:
@kewitz Thank you for this PR! 

However, I would like to move forward with a more generic approach for now. Let's add a `platformFeePercent` column to the Collective table (default to `5`) and use its value in the payment processing, similar to what we're doing with `hostFeePercent`. And let's skip the GraphQL resolvers while we're working on specs for the frontend part.

kewitz:
@znarf this should be consistent with current implementations.

migrations/20200323163000-collective-platform-fee.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const OC_FEE_PERCENT = 5;
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('Collectives', 'platformFeePercent', {
+      type: Sequelize.INTEGER,
+      defaultValue: OC_FEE_PERCENT,
+    });
```
znarf: Small detail, is it possible to insert it next to `hostFeePercent`?

znarf: I checked and it's possible in MySQL but not in Postgres.

znarf:
Rebased. Edited.

migrations/20200323163000-collective-platform-fee.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const OC_FEE_PERCENT = 5;
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('Collectives', 'platformFeePercent', {
+      type: Sequelize.INTEGER,
+      defaultValue: OC_FEE_PERCENT,
```
Betree: If `null` is not an acceptable value for this column, can we add `allowNull`:

```suggestion
      defaultValue: OC_FEE_PERCENT,
      allowNull: false,
```

znarf: Maybe platformFeePercent doesn't make sense for Organization or User, so it should be nullable and only set for Collectives and Events. We should then make sure to set it when creating new Collectives and Events.

Also, it's weird to have hostFeePercent and platformFeePercent to be of different types (float vs integer). What should we settle on?

xdamman: Looks like integer is better here since it's already a percentage value. Hard to imagine less than 1% precision. My $0.02 ;-)

migrations/20200323163000-collective-platform-fee.js
kewitz: No, that's not possible in Postgres. Adding column _after_ is a MySQL feature.

migrations/20200323163000-collective-platform-fee.js
kewitz: @znarf that's the only inconsistency between `platformFeePercent` and `hostFeePercent`, but reviewing the code I can guarantee nothing will go bad.
Never use float when you can use integer.

znarf: Like no one ever will want to cut platform fees in half and say 2.5% ? ;-)

znarf:
@kewitz We need to follow up and consolidate, seems some comments were not taken in consideration. Out of clarity, `platformFeePercent` should only be set for Collectives and Events. We should then make sure to set it when creating new Collectives and Events.

kewitz:
@znarf I'll follow up now, I thought the only thing holding it back was the `after` column, my bad.

kewitz:
@znarf isn't that an overkill? `hostFeePercent` is also being set for every collective.

znarf:
Ideally, we should also review `hostFeePercent` usage, I recently figured it was inconsistent.

PR #3549, opened 2020-03-20
Use bundled version of body parser

znarf:
One less package!


server/lib/express.js
```diff
@@ -50,7 +50,7 @@ export default function(app) {
       },
     }),
   );
-  app.use(bodyParser.urlencoded({ limit: '50mb', extended: true }));
+  app.use(express.urlencoded({ limit: '50mb' }));
```
znarf: `extended: true` is the default

PR #3570, opened 2020-03-24
Following fixes to Sponsorship

kewitz:
Following #3546

### DIFF
--- a/migrations/20200323163000-collective-platform-fee.js
+++ b/migrations/20200323163000-collective-platform-fee.js
@@ -5,11 +5,11 @@ const OC_FEE_PERCENT = 5;
 module.exports = {
   up: async (queryInterface, Sequelize) => {
     await queryInterface.addColumn('Collectives', 'platformFeePercent', {
-      type: Sequelize.INTEGER,
+      type: Sequelize.FLOAT,
       defaultValue: OC_FEE_PERCENT,
     });
     await queryInterface.addColumn('CollectiveHistories', 'platformFeePercent', {
-      type: Sequelize.INTEGER,
+      type: Sequelize.FLOAT,
       defaultValue: OC_FEE_PERCENT,
     });
   },
--- a/migrations/20200323163000-collective-platform-fee.js
+++ b/migrations/20200323163000-collective-platform-fee.js
@@ -1,16 +1,12 @@
 'use strict';
 
-const OC_FEE_PERCENT = 5;
-
 module.exports = {
   up: async (queryInterface, Sequelize) => {
     await queryInterface.addColumn('Collectives', 'platformFeePercent', {
       type: Sequelize.FLOAT,
-      defaultValue: OC_FEE_PERCENT,
     });
     await queryInterface.addColumn('CollectiveHistories', 'platformFeePercent', {
       type: Sequelize.FLOAT,
-      defaultValue: OC_FEE_PERCENT,
     });
   },
 
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -70,6 +70,10 @@ export async function createCollective(_, args, req) {
     // The currency of the new created collective if not specified should be the one of its direct parent or the host (in this order)
     collectiveData.currency = collectiveData.currency || parentCollective.currency;
     collectiveData.HostCollectiveId = parentCollective.HostCollectiveId;
+
+    if (collectiveData.type === types.EVENT) {
+      collectiveData.platformFeePercent = parentCollective.platformFeePercent;
+    }
   }
 
   if (collectiveData.HostCollectiveId) {
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -237,7 +237,6 @@ export default function (Sequelize, DataTypes) {
 
       platformFeePercent: {
         type: DataTypes.INTEGER,
-        defaultValue: OC_FEE_PERCENT,
         validate: {
           min: 0,
           max: 100,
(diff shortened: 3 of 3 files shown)


server/models/Collective.js
```diff
@@ -680,6 +680,15 @@ export default function (Sequelize, DataTypes) {
             // Prevent collective creation
             throw new Error('Collective creation failed');
           }
+
+          // Set default platformFeePercent
+          if (!instance.platformFeePercent) {
+            if (instance.type === 'COLLECTIVE' || instance.type === 'EVENT') {
```
znarf: Are we sure that `instance.type === 'COLLECTIVE'` is set at this point and that we're not relying on the default value of the column?

Should we do the opposite and exclude ORGANIZATION and USER` ?

Should EVENT inherit the value of their parent collective?

migrations/20200323163000-collective-platform-fee.js
```diff
@@ -5,11 +5,11 @@ const OC_FEE_PERCENT = 5;
 module.exports = {
   up: async (queryInterface, Sequelize) => {
     await queryInterface.addColumn('Collectives', 'platformFeePercent', {
-      type: Sequelize.INTEGER,
+      type: Sequelize.FLOAT,
       defaultValue: OC_FEE_PERCENT,
```
znarf: Should we remove the defaultValue?

server/models/Collective.js
kewitz: Good question, this seems to be a bit inconsistent.
I think excluding ORG and USER should be safer.

server/graphql/v1/mutations/collectives.js
```diff
@@ -70,6 +70,10 @@ export async function createCollective(_, args, req) {
     // The currency of the new created collective if not specified should be the one of its direct parent or the host (in this order)
     collectiveData.currency = collectiveData.currency || parentCollective.currency;
     collectiveData.HostCollectiveId = parentCollective.HostCollectiveId;
+
+    if (collectiveData.type === types.EVENT) {
+      collectiveData.platformFeePercent = collectiveData.platformFeePercent || parentCollective.platformFeePercent;
```
znarf: Why would `collectiveData.platformFeePercent` be set?

znarf: Should we also process `hostFeePercent` similarly here?

server/graphql/v1/mutations/collectives.js
```diff
@@ -90,6 +94,7 @@ export async function createCollective(_, args, req) {
 
     collectiveData.currency = collectiveData.currency || hostCollective.currency;
     collectiveData.hostFeePercent = hostCollective.hostFeePercent;
+    collectiveData.platformFeePercent = collectiveData.platformFeePercent || hostCollective.platformFeePercent;
```
znarf: We're now using GraphQL v2 for Collective (type=COLLECTIVE) creation: https://github.com/opencollective/opencollective-api/blob/master/server/graphql/v2/mutation/CreateCollectiveMutations.js

server/graphql/v1/mutations/collectives.js
kewitz: Only if we drop default value.

server/graphql/v1/mutations/collectives.js
```diff
@@ -90,6 +96,9 @@ export async function createCollective(_, args, req) {
 
     collectiveData.currency = collectiveData.currency || hostCollective.currency;
     collectiveData.hostFeePercent = hostCollective.hostFeePercent;
+    collectiveData.platformFeePercent = isNullOrUndefined(collectiveData.platformFeePercent)
+      ? hostCollective.platformFeePercent
```
znarf: `hostCollective.platformFeePercent` is not supposed to be set or used, I think we should default on the `OC_FEE_PERCENT` constant (so actually not do anything and remove this line).

kewitz: OK.

server/graphql/v1/mutations/collectives.js
```diff
@@ -25,6 +25,8 @@ const DEFAULT_COLLECTIVE_SETTINGS = {
   features: { conversations: true },
 };
 
+const isNullOrUndefined = x => x === null || x === undefined;
```
znarf: 'isNullOrUndefined' is assigned a value but never used

server/graphql/v2/mutation/CreateCollectiveMutations.js
```diff
@@ -77,6 +77,10 @@ async function createCollective(_, args, req) {
     }
   }
 
+  if (host) {
+    collectiveData.platformFeePercent = host.platformFeePercent;
+  }
+
```
znarf: We don't want that.

server/models/Collective.js
```diff
@@ -680,6 +679,15 @@ export default function (Sequelize, DataTypes) {
             // Prevent collective creation
             throw new Error('Collective creation failed');
           }
+
+          // Set default platformFeePercent
+          if (!instance.platformFeePercent) {
```
Betree: My understanding is that this will reset `platformFeePercent` if we try to create a collective with `platformFeePercent === 0`. Not a big deal because I suppose we will add this flag after the collective creation, but using `isNil` could have been better:

```es6
if (!isNil(instance.platformFeePercent)) {
```

kewitz: Yes, that's actually better.
I've implemented it here: https://github.com/opencollective/opencollective-api/pull/3583/files#diff-768438be827d00a869e92ca592010413R684.
Although I wrote it explicitly in plain JS.

PR #3598, opened 2020-03-27
EditCollective: Prevent editing tiers if not event or collective

Betree:
Fix https://github.com/opencollective/opencollective/issues/3014

I'm not sure where this was introduced. The real solution to avoid that in the future would be https://github.com/opencollective/opencollective/issues/1451.


znarf:
@Betree I think this might have been introduced in https://github.com/opencollective/opencollective-frontend/pull/3630

Betree:
@znarf good catch, https://github.com/opencollective/opencollective-frontend/pull/3630/files#diff-a24a56b7f34652e27e424c18037c2e33R78 probably made us return an array with an empty object VS an empty array as it was before.

PR #3599, opened 2020-03-27
Run prettier:write

Betree:
Our `prettier:write` was not up to date on master.

I suppose this was introduced in https://github.com/opencollective/opencollective-api/pull/3576 (see https://github.com/prettier/prettier/blob/master/CHANGELOG.md).


Betree:
Seems that my local version was not up to date

PR #3611, opened 2020-03-30
Fix all occurences of Eslint camelcase rule

znarf:
And strictly enforce it from now.

### DIFF
--- a/server/controllers/connectedAccounts.js
+++ b/server/controllers/connectedAccounts.js
@@ -11,7 +11,7 @@ import { mustBeLoggedInTo } from '../lib/auth';
 const { ConnectedAccount, User } = models;
 
 export const createOrUpdate = (req, res, next, accessToken, data, emails) => {
-  const { utm_source, redirect } = req.query;
+  const { utm_source: utmSource, redirect } = req.query;
   const { service } = req.params;
   const attrs = { service };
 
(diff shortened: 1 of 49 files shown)


server/controllers/connectedAccounts.js
```diff
@@ -11,7 +11,7 @@ import { mustBeLoggedInTo } from '../lib/auth';
 const { ConnectedAccount, User } = models;
 
 export const createOrUpdate = (req, res, next, accessToken, data, emails) => {
-  const { utm_source, redirect } = req.query;
+  const { utm_source, redirect } = req.query; // eslint-disable-line camelcase
```
Betree: For this one an alias could be cleaner than ignoring the warning

```suggestion
  const { utm_source: utmSource, redirect } = req.query;
```

server/controllers/connectedAccounts.js
```diff
@@ -75,7 +75,7 @@ export const createOrUpdate = (req, res, next, accessToken, data, emails) => {
           const token = user.generateConnectedAccountVerifiedToken(caId, data.profile.username);
           const newLocation = redirect
             ? `${redirect}?token=${token}`
-            : `${config.host.website}/github/apply/${token}?utm_source=${utm_source}`;
+            : `${config.host.website}/github/apply/${token}?utm_source=${utm_source}`; // eslint-disable-line camelcase
```
Betree: If we rename the variable

```suggestion
            : `${config.host.website}/github/apply/${token}?utm_source=${utmSource}`;
```

PR #3613, opened 2020-03-30
Security: safeguard payExpense from quick approve and pay

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/2958


kewitz:
@Betree that's an interesting suggestion.
In that case, we would need to notify admins of the host, otherwise, they might never notice a bad expense being paid.

kewitz:
Require better specs.

PR #3621, opened 2020-03-31
Eslint simple-import-sort

znarf:


znarf:
This is ready to merge, we can later integrate the rule in our eslint-config-opencollective.

PR #3623, opened 2020-03-31
Consider currency when calculating plan usage

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3029
I tried to move the computational complexity to Postgres and process in runtime only what we need.

### DIFF
--- a/server/constants/paymentMethod.ts
+++ b/server/constants/paymentMethod.ts
@@ -1,0 +1,2 @@
+export const services = ['paypal', 'stripe', 'opencollective', 'prepaid'];
+export const types = ['creditcard', 'prepaid', 'payment', 'collective', 'adaptive', 'virtualcard'];
--- a/server/constants/paymentMethod.ts
+++ b/server/constants/paymentMethod.ts
@@ -1,2 +1,0 @@
-export const services = ['paypal', 'stripe', 'opencollective', 'prepaid'];
-export const types = ['creditcard', 'prepaid', 'payment', 'collective', 'adaptive', 'virtualcard'];
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -18,6 +18,7 @@ import {
   sumBy,
   keys,
   omit,
+  sum,
   defaults,
   includes,
   isNull,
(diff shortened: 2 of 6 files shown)


server/models/PaymentMethod.js
```diff
@@ -14,16 +14,13 @@ import * as libpayments from '../lib/payments';
 import { isTestToken } from '../lib/stripe';
 
 import { maxInteger } from '../constants/math';
+import { services, types } from '../constants/paymentMethod';
```
znarf: Maybe use a `*` import so you can later write `paymentMethod.services` and `paymentMethod.types` which is clearer.

server/constants/paymentMethod.ts
```diff
@@ -0,0 +1,2 @@
+export const services = ['paypal', 'stripe', 'opencollective', 'prepaid'];
```
znarf: Maybe name the file with plural? -> `paymentMethod.ts`

server/models/Collective.js
```diff
@@ -2469,34 +2470,65 @@ export default function (Sequelize, DataTypes) {
...
+          currency: { [Op.ne]: 'USD' },
+          hostCurrency: { [Op.ne]: 'USD' },
         },
-      ],
-      raw: true,
-    });
+        include,
+        raw: true,
+      }),
+    ]);
```
znarf: I understand what's attempted, but I think it would be clearer to get the sum for all currencies in one SQL query, then proceed with the rates. Especially that the complex code has to be repeated multiple times.

We don't even need a special case for USD, because getFxRate knows it just need to skip conversion if USD -> USD.

server/constants/paymentMethod.ts
```diff
@@ -0,0 +1,2 @@
...
+export const types = ['creditcard', 'prepaid', 'payment', 'collective', 'adaptive', 'virtualcard'];
```
Betree, line 2: **[:sparkles: **Enhancement**]** I'd prefer if we can set explicit names here, like `paymentMethodServices`. Because these variables are non-mutable (*real* constants) I'd event argue they should be capitalized: `PAYMENT_METHOD_SERVICES` / `PAYMENT_METHOD_TYPES`.

For example for the collectives constants we also export a `type` variable in `server/constants/collectives.ts` and I find it very unpractical for two reasons:
1. If you're deep in a file, it may be unclear what this `type` variable is and where it's coming from
2. It prevents the variable names from being used in the whole file, and these are very common variable names

Sure you can solve that by aliasing your imports, but I'd rather have explicit names by default to simplify their usage.

---

<details>
<br/>

:no_entry_sign: **Blocker**

:sparkles: **Enhancement**

</details>

znarf: For the same problem, I proposed https://github.com/opencollective/opencollective-api/pull/3623#discussion_r401182819 but your proposal is better.

kewitz: Fixed in 0ddfca9.

server/models/Collective.js
kewitz: Cool, I simplified it, but it brings a bit of complexity to JS runtime, check it out: 0c34197.

PR #3627, opened 2020-04-01
GQLV2: Add mutation to edit expenses

Betree:


server/graphql/v1/inputTypes.js
```diff
@@ -485,7 +485,6 @@ export const ExpenseInputType = new GraphQLInputObjectType({
         deprecationReason: '2020-21-01: Please use PayoutMethod',
       },
       PayoutMethod: {
-        id: { type: GraphQLInt },
```
Betree, line 488: @kewitz this was added in https://github.com/opencollective/opencollective-api/commit/a4693da584c86dda36d10bfdd95cc4f915cdb957 but I assume it has no effect because this is not a GraphQLObject, the fields are defined in `PayoutMethodInputType` not here. Can you confirm if it was just a typo?

server/graphql/v2/mutation/ExpenseMutations.ts
```diff
@@ -38,6 +39,42 @@ const expenseMutations = {
...
+          type: expense.payoutMethod.type,
+        },
+        attachments: expense.attachments?.map(attachment => ({
+          id: attachment.id && idDecode(attachment.id, IDENTIFIER_TYPES.EXPENSE_ATTACHMENT),
+          url: attachment.url,
+          amount: attachment.amount,
+          incurredAt: attachment.incurredAt,
+          description: attachment.description,
+        })),
+        fromCollective: null, // TODO payee
```
Betree, line 74: Will be tackled with https://github.com/opencollective/opencollective/issues/2266

server/graphql/v1/inputTypes.js
kewitz: @betree, this was not a typo.
I temporarily had to add support for PayoutMethod in the current Expense flow, so we can select our TransferWise payout method when submitting expenses.

Betree: But the input fields were already defined in `PayoutMethodInputType`, I don't understand why we needed to add an `id` here and how it influences the schema.

kewitz: Ah, now I get it. Yes, my mistake!

PR #3629, opened 2020-04-01
Pledges: Bypass Github validations on CI

Betree:
https://github.com/opencollective/opencollective-frontend/pull/3803 progress is blocked because Github validations fails for verifying the pledged repo/org. This is because we don't have a Github token setup, so the requests can't be triggered.

Alternatives solutions:
- Setup a dummy Github account and set the token in CI
- Completely bypass the Github verification on CI rather than relying on the name starting with `valid`. Simpler, but we won't be able to test the errors cases.

### DIFF
--- a/server/lib/github.js
+++ b/server/lib/github.js
@@ -140,17 +140,26 @@ export async function getOrgMemberships(accessToken) {
 }
 
 export async function checkGithubExists(githubHandle, accessToken) {
-  if (githubHandle.includes('/')) {
+  const isRepo = githubHandle.includes('/');
+  const invalidOrgMsg = 'We could not verify the GitHub organization exists';
+  const invalidRepoMsg = 'We could not verify the GitHub repository exists';
+
+  // On CI, only throw if the repo name doesn't start with "valid"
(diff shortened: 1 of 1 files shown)


znarf:
I'm happy to introduce a GitHub configuration in CI if that is required.

Betree:
@znarf introducing a Github token in CI may be cleaner than what I did in this PR (having specifics `if`s for the CI is not great). We need to make sure that it's a dummy token tho because we won't be able to protect it.

Let me know which option you prefer.

znarf:
I wanted to create a new oAuth App and just use it for that. It will technically be a secret in GitHub, but yes it can be extracted with dedicated code and PR.

Betree:
@znarf That works for me

Betree:
@znarf can I merge this one to unblock the contributor (he's working on a bounty and is almost done), we can eventually follow up with the enhancement later?

znarf:
@Betree ok

Betree:
Closing in favor of https://github.com/opencollective/opencollective-api/pull/3680

PR #3631, opened 2020-04-01
Enforce Pay with TransferWise plan limit

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3028
Based on #3623

### DIFF
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -9,7 +9,7 @@ import paymentProviders from '../../../paymentProviders';
 import * as libPayments from '../../../lib/payments';
 import { formatCurrency } from '../../../lib/utils';
 import { floatAmountToCents } from '../../../lib/math';
-import { handleTransferwisePaymentLimit } from '../../../lib/plans';
+import { handleTransferwisePayoutsLimit } from '../../../lib/plans';
 import { createFromPaidExpense as createTransactionFromPaidExpense } from '../../../lib/transactions';
 import { getFxRate } from '../../../lib/currency';
 import debugLib from 'debug';
@@ -518,7 +518,7 @@ async function payExpenseWithTransferwise(host, payoutMethod, expense, fees, rem
     throw new Error('Host is not connected to Transferwise');
   }
 
-  await handleTransferwisePaymentLimit(host, expense, { throwException: true });
+  await handleTransferwisePayoutsLimit(host);
 
   const data = await paymentProviders.transferwise.payExpense(connectedAccount, payoutMethod, expense);
   const transactions = await createTransactions(host, expense, fees, data);
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -9,6 +9,7 @@ import paymentProviders from '../../../paymentProviders';
 import * as libPayments from '../../../lib/payments';
 import { formatCurrency } from '../../../lib/utils';
 import { floatAmountToCents } from '../../../lib/math';
+import { handleTransferWisePaymentLimit } from '../../../lib/plans';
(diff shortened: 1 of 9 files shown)


server/lib/plans.ts
```diff
@@ -95,6 +95,19 @@ export async function handleHostPlanBankTransfersLimit(host, { throwException =
   }
 }
 
+export async function handleTransferWisePaymentLimit(host, expense, { throwException = false }): Promise<void> {
+  const hostPlan = await host.getPlan();
+  if (
+    throwException &&
+    hostPlan.payedWithTransferWiseLimit !== null &&
+    hostPlan.payedWithTransferWise + expense.amount >= hostPlan.payedWithTransferWiseLimit
```
znarf, line 103: We stopped doing that. You're now allowed to go over the limit, but you can't anymore once you're over limit. So you can simply remove the `+ expense.amount`

kewitz: Nice, I'll fix it and rebase, thanks!

server/lib/plans.ts
```diff
@@ -95,6 +95,19 @@ export async function handleHostPlanBankTransfersLimit(host, { throwException =
   }
 }
 
+export async function handleTransferWisePaymentLimit(host, expense, { throwException = false }): Promise<void> {
```
znarf, line 98:
```suggestion
export async function handleTransferwisePayoutLimit(host, expense, { throwException = false }): Promise<void> {
```

server/graphql/v1/mutations/expenses.js
```diff
@@ -445,11 +446,12 @@ async function payExpenseWithTransferwise(host, payoutMethod, expense, fees) {
   const [connectedAccount] = await host.getConnectedAccounts({
     where: { service: 'transferwise', deletedAt: null },
   });
-
   if (!connectedAccount) {
     throw new Error('Host is not connected to Transferwise');
   }
 
+  await handleTransferWisePaymentLimit(host, expense, { throwException: true });
```
znarf, line 453: How should we handle the limitation on the frontend side?

kewitz: We can disable the button and display a tooltip like we're doing with the other plan limitations.
The backend error message shouldn't be too invasive tho.

server/lib/plans.ts
kewitz: Thanks, I missed that one.

kewitz:
Thanks @znarf! The frontend is here: https://github.com/opencollective/opencollective-frontend/pull/3879

znarf:
Sorry, a bit of annoying last minute feedback, I think it should be `transferwisePayouts` plural.

And `handleTransferwisePaymentLimit` should be `handleTransferwisePayoutsLimit`. Maybe even `handleHostTransferwisePayoutsLimit` or `handleHostPlanTransferwisePayoutsLimit` to match other similar methods.

server/lib/plans.ts
```diff
@@ -95,6 +95,19 @@ export async function handleHostPlanBankTransfersLimit(host, { throwException =
   }
 }
 
+export async function handleTransferwisePaymentLimit(host, expense, { throwException = false }): Promise<void> {
```
Betree, line 98: **[:ice_cream: Details & personal preferences]** It's a strange API, because if `throwException` is false (default) then the function will never do anything. Should probably either return a boolean in such cases or remove this param.

znarf: Not a justification but an explanation, The API is copied from `handleHostPlanAddedFundsLimit` which has ability to throwException AND notifyAdmins.

kewitz: Exact, `handleHostPlanAddedFunsLimit` was used as a boilerplate. I thought the consistency was better here but I agree it is misleading.
I'll refactor both functions to remove unused parameters.

kewitz:
Rebased, changes starting at 8f1868b.

PR #3635, opened 2020-04-02
custom email templates for the-social-change-agency

xdamman:
It would be good to have an interface for hosts to be able to customise such email templates.

In the meantime, here are custom ones for https://opencollective.com/the-social-change-agency

It's quite urgent as they are onboarding a lot of new collectives now for the COVID-19 pandemic.

Thank you!


leonardohof:
> Why not use the same mechanism as 'thankyou' emails in `generateEmailFromTemplate` ?

@znarf ... are you saying, to put the logic of getting additional data (userEmail) there?

xdamman:
@znarf There is no properly documented way to do this. So I  just looked at the other example in the code of how to customize the collective.created notification (there is such example for meetups). Clearly this needs refactoring but this is beyond the goal of this PR.

znarf:
@xdamman This didn't answer my question. Customizing based on slug is usually done in. https://github.com/opencollective/opencollective-api/blob/master/server/lib/email.js#L250 I was thinking it's better to consolidate such behavior in a single place.

znarf:
While looking at it, I found that we also need to do it like that for the Open Source Collective. Working on such PR now.

znarf:
https://github.com/opencollective/opencollective-api/pull/3651

PR #3639, opened 2020-04-03
Inherit platformFeePercent from Host

kewitz:


server/graphql/v1/mutations/collectives.js
```diff
@@ -94,6 +94,7 @@ export async function createCollective(_, args, req) {
 
     collectiveData.currency = collectiveData.currency || hostCollective.currency;
     collectiveData.hostFeePercent = hostCollective.hostFeePercent;
+    collectiveData.platformFeePercent = hostCollective.platformFeePercent;
```
znarf, line 97: Only if it's set, non null.

znarf: Maybe you should handle that within `addHost`?

server/graphql/v2/mutation/CreateCollectiveMutations.js
```diff
@@ -85,6 +85,7 @@ async function createCollective(_, args, req) {
   // Add the host if any
   if (host) {
     await collective.addHost(host, remoteUser, { shouldAutomaticallyApprove });
+    await collective.update({ platformFeePercent: host.platformFeePercent });
```
znarf, line 88: If done within addHost, not necessary.

server/graphql/v1/mutations/collectives.js
kewitz: Nice catch, we should add the not null conditional.
I don't think we should move this to `addHost` because I think we should keep this behaviour only for newly created collectives. If we move this to `addHost` we might end up changing platformFee for collectives that are changing host. WDYT?

znarf: @kewitz I think that's actually a desirable effect. Many Collectives will actually be created without host and it will be added later.

server/models/Collective.js
```diff
@@ -1613,6 +1613,7 @@ export default function (Sequelize, DataTypes) {
     const updatedValues = {
       HostCollectiveId: hostCollective.id,
       hostFeePercent: hostCollective.hostFeePercent,
+      platformFeePercent: hostCollective.platformFeePercent !== null ? hostCollective.platformFeePercent : undefined,
```
znarf, line 1616: What is the expected behavior here? If `hostCollective.platformFeePercent` is `null` ?

We should make sure that it's reset to `OC_FEE_PERCENT`.

kewitz: I want to fall-back to the `beforeCreate` hook in the collective model.
That hook will make sure to set the right OC_FEE_PERCENT.

znarf: At this stage, it's not a creation, it's an update.

kewitz: I'm using `undefined` because I don't want to convey the idea that this will be necessarily set to `null`.

kewitz: Nice, so undefined will make sure we're not updating its value, right?

kewitz:
@znarf I'm not sure if we're miscommunication, but I think this code covers all the cases.

kewitz:
@znarf true, but that's also the case if you create any COVID collective.
Either way, this will require a manual migration to get these collectives back to the default platform fee.

PR #3640, opened 2020-04-03
Expense Tags

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/2823

### DIFF
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -356,25 +356,23 @@ export async function editExpense(remoteUser, expenseData) {
       PayoutMethodId !== expense.PayoutMethodId,
     );
 
-    if (cleanExpenseData) {
-      const existingTags = expense.tags || [];
-      let tags = cleanExpenseData.tags;
-      if (cleanExpenseData.category) {
-        tags = [cleanExpenseData.category, ...existingTags];
-      }
-      return expense.update(
-        {
-          ...cleanExpenseData,
-          lastEditedById: remoteUser.id,
-          incurredAt: expenseData.incurredAt || new Date(),
-          status: shouldUpdateStatus ? 'PENDING' : expense.status,
-          PayoutMethodId: PayoutMethodId,
-          legacyPayoutMethod: models.Expense.getLegacyPayoutMethodTypeFromPayoutMethod(payoutMethod),
-          tags,
-        },
-        { transaction: t },
-      );
+    const existingTags = expense.tags || [];
+    let tags = cleanExpenseData.tags;
+    if (cleanExpenseData.category) {
+      tags = [cleanExpenseData.category, ...existingTags];
     }
+    return expense.update(
+      {
+        ...cleanExpenseData,
+        lastEditedById: remoteUser.id,
+        incurredAt: expenseData.incurredAt || new Date(),
+        status: shouldUpdateStatus ? 'PENDING' : expense.status,
+        PayoutMethodId: PayoutMethodId,
+        legacyPayoutMethod: models.Expense.getLegacyPayoutMethodTypeFromPayoutMethod(payoutMethod),
+        tags,
+      },
+      { transaction: t },
+    );
   });
 
   await updatedExpense.createActivity(activities.COLLECTIVE_EXPENSE_UPDATED, remoteUser);
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -224,7 +224,7 @@ export async function createExpense(remoteUser, expenseData) {
       {
         ...pick(expenseData, EXPENSE_EDITABLE_FIELDS),
         currency: collective.currency,
-        tags: expenseData.category ? [expenseData.category] : null,
+        tags: expenseData.category ? [expenseData.category] : expenseData.tags,
         status: statuses.PENDING,
         CollectiveId: collective.id,
         FromCollectiveId: fromCollective.id,
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -135,7 +135,7 @@ const checkExpenseAttachments = (expenseData, attachments) => {
   }
 };
 
-const EXPENSE_EDITABLE_FIELDS = ['amount', 'description', 'category', 'type', 'privateMessage', 'invoiceInfo'];
+const EXPENSE_EDITABLE_FIELDS = ['amount', 'description', 'category', 'type', 'tags', 'privateMessage', 'invoiceInfo'];
 
 const getPaypalPaymentMethodFromExpenseData = async (expenseData, remoteUser, fromCollective, dbTransaction) => {
   if (expenseData.PayoutMethod) {
@@ -224,6 +224,7 @@ export async function createExpense(remoteUser, expenseData) {
       {
         ...pick(expenseData, EXPENSE_EDITABLE_FIELDS),
         currency: collective.currency,
+        tags: expenseData.category ? [expenseData.category] : null,
         status: statuses.PENDING,
         CollectiveId: collective.id,
         FromCollectiveId: fromCollective.id,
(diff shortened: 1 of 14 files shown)


server/graphql/v1/mutations/expenses.js
```diff
@@ -340,17 +341,25 @@ export async function editExpense(remoteUser, expenseData) {
...
-        ...cleanExpenseData,
-        lastEditedById: remoteUser.id,
-        incurredAt: expenseData.incurredAt || new Date(),
-        status: shouldUpdateStatus ? 'PENDING' : expense.status,
-        PayoutMethodId: PayoutMethodId,
-        legacyPayoutMethod: models.Expense.getLegacyPayoutMethodTypeFromPayoutMethod(payoutMethod),
-      },
-      { transaction: t },
-    );
+    if (cleanExpenseData) {
```
znarf, line 344: Why is this condition needed?

server/graphql/v2/object/TransferWise.ts
```diff
@@ -52,7 +51,7 @@ export const TransferWise = new GraphQLObjectType({
     requiredFields: {
       args: {
         currency: {
-          type: new GraphQLNonNull(TransferWiseCurrency),
+          type: new GraphQLNonNull(GraphQLString),
```
znarf, line 54: What's the relationship with the PR?

server/models/Expense.js
```diff
@@ -141,6 +142,32 @@ export default function (Sequelize, DataTypes) {
...
+              })
+              .filter(tag => {
+                return tag && tag.length > 0;
+              });
+          }
+
+          if (!tags || tags.length === 0) {
+            this.setDataValue('tags', null);
+          } else if (tags) {
+            this.setDataValue('tags', Array.from(new Set(tags)));
```
znarf, line 166: Is that a de-duplication? In this case, why is it not done before? And why not use lodash?

server/models/Expense.js
```diff
@@ -141,6 +142,32 @@ export default function (Sequelize, DataTypes) {
       deletedAt: {
         type: DataTypes.DATE,
       },
+
+      tags: {
+        type: DataTypes.ARRAY(DataTypes.STRING),
+        set(tags) {
+          if (tags) {
```
znarf, line 149: Can we add a comment explaining why we're doing that?

znarf: I know see that the code was duplicated from somewhere else, can that be extracted in the `tags` lib?

server/lib/tags.ts
```diff
@@ -0,0 +1,17 @@
+export const validateTags = (tags: string[]): void => {
+  if (tags) {
+    // Limit to max 30 tags
+    if (tags.length > 30) {
+      throw new Error(`Conversations cannot have more than 30 tags. Please remove ${30 - tags.length} tag(s).`);
```
znarf, line 5: If it's used for something else that Conversations, the message is wrong.

server/graphql/v1/mutations/expenses.js
```diff
@@ -209,6 +209,7 @@ export async function createExpense(remoteUser, expenseData) {
       {
         ...pick(expenseData, EXPENSE_EDITABLE_FIELDS),
         currency: collective.currency,
+        tags: expenseData.category ? [expenseData.category] : null,
```
Betree, line 212: **[:sparkles: **Enhancement**]** `ExpenseCreateInput` already has a `tags` field. I'd go with:

```suggestion
        tags: expenseData.category ? [expenseData.category] : expenseData.tags,
```

server/graphql/v1/queries.js
```diff
@@ -675,7 +675,7 @@ const queries = {
         query.where.status = args.status;
       }
       if (args.category) {
-        query.where.category = { [Op.iLike]: args.category };
+        query.where[Op.and] = [sequelize.literal(`tags[1] ILIKE ${sequelize.escape(args.category)}`)];
```
Betree, line 678: **[:sparkles: **Enhancement**]** Would be better to search among all tags (not sure about the syntax below)

```suggestion
        query.where.tags = { [Op.contains]: [args.category] };
```

server/graphql/v1/types.js
```diff
@@ -790,7 +790,13 @@ export const ExpenseType = new GraphQLObjectType({
       category: {
         type: GraphQLString,
```
Betree, line 791: **[:sparkles: **Enhancement**]** We can deprecate the field

```suggestion
        type: GraphQLString,
        deprecationReason: '2020-04-01 - Please use tags',
```

---

<details>
<br/>

:no_entry_sign: **Blocker**

:sparkles: **Enhancement**

</details>

server/models/Expense.js
```diff
@@ -141,6 +142,32 @@ export default function (Sequelize, DataTypes) {
...
+            tags = tags
+              .map(tag => {
+                if (tag) {
+                  const upperCase = tag.toUpperCase();
+                  const cleanTag = upperCase.trim().replace(/\s+/g, ' ');
+                  return stripHTML(cleanTag);
```
Betree, line 155: **[:sparkles: **Enhancement**]** As mentioned on Slack, this `stripHTML` is not required because we don't use tags in a `dangerouslySetInnerHtml`. I went a bit too far by adding it to conversations tags :slightly_smiling_face:

server/models/Expense.js
Betree: Doing that in the setter seems like the good place to me, that way you make sure that tags are always de-duplicated, no matter where & how you're creating/updating them. Agree that lodash could simplify the code.

Betree: Doing that in the setter seems like the good place to me, that way you make sure that tags are always de-duplicated, no matter where & how you're creating/updating them. Agree that lodash could simplify the code.

znarf: By earlier, I just wanted to say before `(!tags || tags.length === 0)` and not directly in the `setDataValue`.

Betree: Good point!

server/graphql/v2/object/TransferWise.ts
kewitz: The expense flow was broken due to an update in TransferWise, I deprecated the type because I couldn't render the Expense edit page.

server/graphql/v1/mutations/expenses.js
kewitz: Not sure how this ended up here, removing it...

server/graphql/v1/queries.js
kewitz: Good idea, the operator is `var ILIKE ANY(tags)`.

server/lib/tags.ts
kewitz: 🎯

kewitz:
Rebased and fixed.

server/graphql/v2/object/Expense.js
```diff
@@ -168,6 +168,12 @@ const Expense = new GraphQLObjectType({
           return req.loaders.Expense.activities.load(expense.id);
         },
       },
+      tags: {
+        type: new GraphQLList(GraphQLString),
+        resolve(expense) {
+          return expense.tags;
+        },
```
Betree, line 175: **[:ice_cream: Details & personal preferences]** It's a detail, but could simplify things on frontend

```suggestion
        type: new GraphQLNonNull(new GraphQLList(GraphQLString)),
        resolve(expense) {
          return expense.tags || [];
        },
```

server/lib/tags.ts
```diff
@@ -0,0 +1,42 @@
...
+          return tag && tag.length > 0;
+        }),
+    );
+  }
+
+  if (!tags || tags.length === 0) {
+    this.setDataValue('tags', null);
+  } else if (tags) {
+    this.setDataValue('tags', tags);
+  }
```
Betree, line 41: This doesn't look right to me, `setTags` has its own context so `this` here probably not refers to the models.

I'd prefer if we can rename the function `cleanTags`, return the value and do the `setDataValue` from the model

kewitz: The reason why `this` works in this context is because `setTags` is an actual `function`. If it was an arrow expression, `this` would be undefined.

Do you think this is too technical and we should just extract the functional expressions?

Betree: > The reason why this works in this context is because setTags is an actual function. If it was an arrow expression, this would be undefined.

Indeed, that explains why it works :+1: 

> Do you think this is too technical and we should just extract the functional expressions?

I do, it's seems more useful to me to have a generic helper to clean tags that doesn't know anything about models and can be called from any code. It's also more practical to write unit tests if the function has no side effect.

znarf:
@kewitz According to https://github.com/opencollective/opencollective-api/pull/3698 There were some missed leftovers. Some were mentionned by @Betree and not taken care of. Can you double check?

kewitz:
@znarf, just noticed the list of files, I'll review that ASAP.

PR #3649, opened 2020-04-06
GQLV2/Expenses: Implement activities

Betree:
API requirements for showing the expense activities on the new expense page.

image

## Example

**Query**
```gql
{
  expense(expense: {legacyId: 2382}) {
    id
    description
    activities {
      id
      type
      createdAt
      individual {
        id
        slug
      }
    }
  }
}
```

**Result**
```json
{
  "data": {
    "expense": {
      "id": "qdrmg0z3-wq7py50o-zd589e4a-xlov6nk3",
      "description": "New expense 79049 edited",
      "activities": [
        {
          "id": "z06danmy-p4vz3b6g-6zpj9wox-e7r8qglk",
          "type": "COLLECTIVE_EXPENSE_CREATED",
          "createdAt": "2020-02-21T12:41:43.370Z",
          "individual": {
            "id": "9lxn69kv-0gwmo5pe-nxbdr3y8-ae7pqz4r",
            "slug": "testuseradmin"
          }
        },
        {
          "id": "9lxn69kv-0gwmo5pz-mlvbdr3y-8ae7pqz4",
          "type": "COLLECTIVE_EXPENSE_APPROVED",
          "createdAt": "2020-02-21T12:41:47.493Z",
          "individual": {
            "id": "9lxn69kv-0gwmo5pe-nxbdr3y8-ae7pqz4r",
            "slug": "testuseradmin"
          }
        },
        {
          "id": "qdrmg0z3-wq7py507-qzwb89e4-axlov6nk",
          "type": "COLLECTIVE_EXPENSE_UPDATED",
          "createdAt": "2020-02-21T12:41:50.476Z",
          "individual": {
            "id": "9lxn69kv-0gwmo5pe-nxbdr3y8-ae7pqz4r",
            "slug": "testuseradmin"
          }
        },
        {
          "id": "nyvwlnxe-389pojvw-zl756daq-407rmkgz",
          "type": "COLLECTIVE_EXPENSE_APPROVED",
          "createdAt": "2020-02-21T12:41:51.142Z",
          "individual": {
            "id": "9lxn69kv-0gwmo5pe-nxbdr3y8-ae7pqz4r",
            "slug": "testuseradmin"
          }
        }
      ]
    }
  }
}
```

### DIFF
--- a/server/graphql/loaders/collective.ts
+++ b/server/graphql/loaders/collective.ts
@@ -1,0 +1,29 @@
+import models, { sequelize } from '../../models';
+import { sortResultsSimple } from './helpers';
+import DataLoader from 'dataloader';
+
+export default {
+  /**
+   * Returns the collective (account) for this user ID, including incognito profiles
+   */
+  byUserId: (): DataLoader<number, object> => {
+    return new DataLoader(async userIds => {
+      const collectives = await sequelize.query(
+        ` SELECT      c.*, u.id AS __user_id__
+          FROM        "Collectives" c
+          INNER JOIN  "Users" u ON u."CollectiveId" = c.id
+          WHERE       u.id in (:userIds)
+          AND         c."deletedAt" IS NULL
+          GROUP BY    u."id", c.id`,
+        {
+          type: sequelize.QueryTypes.SELECT,
+          model: models.Collective,
+          mapToModel: true,
+          replacements: { userIds },
+        },
+      );
+
+      return sortResultsSimple(userIds, collectives, result => result.dataValues['__user_id__']);
+    });
+  },
+};
--- a/server/graphql/loaders/expenses.ts
+++ b/server/graphql/loaders/expenses.ts
@@ -1,4 +1,5 @@
 import DataLoader from 'dataloader';
+import ACTIVITY from '../../constants/activities';
 import models, { Op } from '../../models';
 import { ExpenseAttachment } from '../../models/ExpenseAttachment';
 import { sortResultsArray } from './helpers';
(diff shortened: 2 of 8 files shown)


server/graphql/v2/interface/Activity.ts
```diff
@@ -0,0 +1,44 @@
...
+    account: {
+      type: Account,
+      description: 'The account concerned by this activity, if any',
+      resolve: async (activity, _, req): Promise<object> => {
+        if (activity.CollectiveId) {
+          return req.loaders.Collective.byId.load(activity.CollectiveId);
+        }
+      },
+    },
+    individual: {
```
Betree, line 34: Maybe we should rename this to `fromIndividual` for more clarity

kewitz: I think we don't need "from" because there's no "to" in an activity.
Perhaps we should keep naming it Account for consistency?

kewitz: I think we don't need "from" because there's no "to" in an activity.
Perhaps we should keep naming it Account for consistency?

Betree: There's already an `account` field. Also this one will always be an individual, it cannot be a collective, an org or an event.

But to you mean `individualAccount`?

kewitz: For the case where an individual is taking action in the name of a collective?
Interesting, I think it is a good idea to bind the data-type. Meaning that if both Individual and Account are returning the same data type, they should have similar names, specially because we don't use TS in the frontend.
I think `individualAccount` could be a nice solution.

Betree: > For the case where an individual is taking action in the name of a collective?

Even then, this endpoint will return the user that triggered the action (we rely on the `UserId` field). The goal in such case is to know which user triggered the action in the name of the collective.


> Meaning that if both Individual and Account are returning the same data type

It's not exactly the case, the endpoint returns an `Individual` object,  which inherits of the `Account` interface but has its own type.

kewitz: Nice, thanks for explaining it to me. Perhaps Individual is enough then, sorry for overcomplicating.

Betree: No over-complication :slightly_smiling_face: Asking for more clarity is always good, no matter if it results in a code change or not!

PR #3651, opened 2020-04-06
Refactor customized emails

znarf:


xdamman:
This assumes that there is an opensource version of every template no? Would it fall back to the general template if such custom email for a particular template doesn’t exist?

znarf:
@xdamman In this case, it's just for `collective.created`.

server/lib/email.js
```diff
@@ -226,6 +226,7 @@ const isWhitelistedDomain = email => {
  */
 const generateEmailFromTemplate = (template, recipient, data = {}, options = {}) => {
   const slug = get(options, 'collective.slug') || get(data, 'collective.slug') || 'undefined';
+  const hostSlug = get(data, 'collective.slug');
```
Betree, line 229: Shouldn't that be

```suggestion
  const hostSlug = get(data, 'collective.host.slug');
```

?

znarf: @Betree Oops. Definitely !

PR #3653, opened 2020-04-06
Expense activities enhancements

Betree:
Require https://github.com/opencollective/opencollective-api/pull/3649

Before this PR, all activities created for expenses had their `UserId` set to the expense's submitter user ID. This was wrong because this field is supposed to tell you who **triggered** the action (ie. pay, approve, etc.). This PR intends to fix that by providing a user for all expense activities, or no user at all when it's a "system" action (ie. expense paid after processing).

Also added activities for expenses marked as unpaid and unapproved.

## About the migration

All the activities related to the update of an expense are corrupted and the only way to properly fix it by using `ExpenseHistories` would require a huge amount of work.

This PR sets the `UserId` to `null` for all these corrupted entries, aka it's better store nothing than to store false information. It will also make the fix possible in the future if we want to invest in that because these entries will be clearly identifiable.

In the interface, they will just appear as `Expense approved` without the user who triggered the action being displayed.

image


test/server/graphql/v1/expenses.test.js
```diff
@@ -1272,7 +1272,7 @@ describe('server/graphql/v1/expenses', () => {
           hostFeeInCollectiveCurrency +
           platformFeeInCollectiveCurrency;
         expect(balance).to.equal(initialBalance - expensePlusFees);
-        await utils.waitForCondition(() => emailSendMessageSpy.callCount === 4);
+        await utils.waitForCondition(() => emailSendMessageSpy.callCount === 2);
```
Betree, line 1275: Because the `createActivity` was previously in the `afterUpdate` hook, we were sending two more emails here for "expense approved". This was not correct because the expense was already approved before the update (see https://github.com/opencollective/opencollective-api/pull/3653/files#diff-35fd0807cc45439c1187333c6f4c30d6R1239) so there's no reason to send these two emails.

PR #3661, opened 2020-04-07
Filter PayPal adaptive paymentMethods

znarf:


znarf:
@Betree You're right and this was properly catched by the test suite. I will switch to a frontend approach.

PR #3662, opened 2020-04-07
Expenses attached files

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3022
Requirement for https://github.com/opencollective/opencollective-frontend/pull/3903

### DIFF
--- a/migrations/20200407101229-create-expense-attached-files.js
+++ b/migrations/20200407101229-create-expense-attached-files.js
@@ -1,0 +1,47 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, DataTypes) => {
+    await queryInterface.createTable('ExpenseAttachedFiles', {
+      id: {
+        type: DataTypes.INTEGER,
+        primaryKey: true,
+        autoIncrement: true,
+      },
+      ExpenseId: {
+        type: DataTypes.INTEGER,
+        references: { key: 'id', model: 'Expenses' },
+        onDelete: 'CASCADE',
+        onUpdate: 'CASCADE',
+        allowNull: false,
+      },
+      CreatedByUserId: {
+        type: DataTypes.INTEGER,
+        references: { key: 'id', model: 'Users' },
+        onDelete: 'SET NULL',
+        onUpdate: 'CASCADE',
+        allowNull: true,
+      },
+      url: {
+        type: DataTypes.STRING,
+        allowNull: false,
+      },
+      createdAt: {
+        type: DataTypes.DATE,
+        defaultValue: DataTypes.NOW,
+        allowNull: false,
+      },
+      updatedAt: {
+        type: DataTypes.DATE,
+        defaultValue: DataTypes.NOW,
+        allowNull: false,
+      },
+    });
+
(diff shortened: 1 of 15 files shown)


server/lib/images.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Returns true if image is a valid OC image uploaded in our S3 bucket.
+ * Allow any image on non-production environments
+ */
+export const isValidOCImage = (url: string): boolean => {
+  if (process.env.NODE_ENV !== 'production') {
+    return true;
+  } else {
+    return url.startsWith(`${process.env.AWS_S3_BUCKET}.s3-us-west-1.amazonaws.com`);
```
Betree, line 9: @znarf do you think that's a safe way to check that uploaded image URLs are valid?

migrations/20200407101229-create-expense-attached-files.js
```diff
@@ -0,0 +1,47 @@
...
+        allowNull: false,
+      },
+      updatedAt: {
+        type: DataTypes.DATE,
+        defaultValue: DataTypes.NOW,
+        allowNull: false,
+      },
+    });
+
+    await queryInterface.addIndex('ExpenseAttachedFiles', ['ExpenseId']);
```
kewitz, line 41: 💯

server/graphql/v2/input/ExpenseAttachedFileInput.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+export const ExpenseAttachedFileInput = new GraphQLInputObjectType({
+  name: 'ExpenseAttachedFileInput',
+  fields: {
+    id: {
+      type: GraphQLString,
+      description: 'ID of the file',
+    },
+    url: {
+      type: new GraphQLNonNull(GraphQLString),
```
znarf, line 14: Maybe should we have considered a URL scalar for that.

server/lib/images.ts
znarf: It doesn't start with http ?

znarf: We should not use `process.env` but use `config` to read configuration.

server/graphql/v2/input/ExpenseAttachedFileInput.ts
Betree: We can introduce that, I'll do it as a separate PR tho because `ExpenseItems` have an `url` field that needs to be updated too.

server/lib/images.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Returns true if image is a valid OC image uploaded in our S3 bucket.
+ * Allow any image on non-production environments
+ */
+export const isValidOCImage = (url: string): boolean => {
```
znarf, line 5: While we're at it, not a fan of names containing `OC`, remove or use `OpenCollective`. Suggested names: `isValidUploadedImage`. Or, maybe it should be more generic and be `isValidUploadedAttachment`

Betree: Going with `isValidUploadedImage`

PR #3663, opened 2020-04-07
Graphql v2: hosts query

znarf:
This PR contains a refactor of all our Collections, no breaking change.

Then, it's introducing a new query to fetch hosts on the platform.

Finally adding some statistics to account and host.


sbinlondon:
Looks good to me. As I understand it, fewer args/variables to pass? Thanks for doing this so quickly!

znarf:
@sbinlondon in this version, I decided to rely on defaults for most variables. We can re-introduce that later if needed:

```
      onlyOpenHosts: true,
      minNbCollectivesHosted: 0,
      orderBy: 'collectives',
      orderDirection: 'DESC',
```

PR #3668, opened 2020-04-08
CI: don't use azure repositories as they are often unavailable

znarf:


znarf:
Closing as stability improved and they said they took actions for better reliability.

Will revisit if it's a problem in the future.

PR #3671, opened 2020-04-08
Rename ExpenseAttachments to ExpenseItems

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3039


server/graphql/v1/types.js
```diff
@@ -829,18 +817,6 @@ export const ExpenseType = new GraphQLObjectType({
           }
         },
       },
-      canSeePrivateInfo: {
```
Betree, line 832: This one wasn't used for API V1, so it's easier to delete with together with `ExpenseViewPermissions`

PR #3680, opened 2020-04-09
Configure GitHub client in CI

znarf:


znarf:
@Betree This should also be done on frontend side. Doing that now.

PR #3683, opened 2020-04-09
Migrate edit URL in template

znarf:
Small fix.

Pinging you @Betree because I'm gonna remove support for these URLs in frontend, seems like we waited enough.


Betree:
Agree, no blocker for me to remove in frontend. Especially because the fallback (ending up in general settings) is not that bad.

PR #3684, opened 2020-04-09
Deprecate apollo-errors

kewitz:
Typing should prevent us from making the same mistake again and it will also suggest what properties else you could be using.


znarf:
> the same mistake again

@kewitz What are you talking about?

Betree:
@znarf I think Leo is referring to the misuse of the errors where we pass the message directly rather than in an object (ie. `new Unauthorized({ message: '' })`).

I agree with the change, I've misused this constructor a bunch of time. It's strange that we have to pass the `message` for an error through an object rather than a direct parameter.

znarf:
Should we change the API and remove the `message` everywhere then?

kewitz:
My suggestion, in that case, would be to deprecate the outdated `apollo-errors` and extend the native ApolloError.

znarf:
@kewitz I'm open to it.

kewitz:
Squashed the first commit, since it was related to the previous refactor.

znarf:
Please rebase and merge!

PR #3693, opened 2020-04-10
chore: update handlebars to version 4.7.1

bolariin:
Resolves https://github.com/opencollective/opencollective/issues/2802

I made use of `info` in `getterMethods` of the model as it was already used within the codebase. From what I could see the errors came from passing the sequelize instances directly as data.

### DIFF
--- a/cron/daily/10-backyourstack-dispatch-subscriptions.js
+++ b/cron/daily/10-backyourstack-dispatch-subscriptions.js
@@ -71,7 +71,7 @@ async function run() {
             CollectiveId: order.fromCollective.id,
             data: {
               orders: dispatchedOrders,
-              collective: order.fromCollective,
+              collective: order.fromCollective.info,
               recurringDispatch: true,
             },
           });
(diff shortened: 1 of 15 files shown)


znarf:
@Betree could you have a look at this?

server/models/User.js
```diff
@@ -198,6 +198,7 @@ export default (Sequelize, DataTypes) => {
             emailWaitingForValidation: this.emailWaitingForValidation,
             createdAt: this.createdAt,
             updatedAt: this.updatedAt,
+            emailConfirmationToken: this.emailConfirmationToken,
```
Betree, line 201: This is a private information that should never be shared, it should not appear in `info`.

Because there's only one email relying on that, I would suggest passing it manually where it's needed.

Betree:
@bolariin changes are good. I still need to test the emails, I'll try to do it today.

Betree:
Sorry for the late update, busy week! Testing that now.

PR #3694, opened 2020-04-12
feat: introduce URL scalar in GQLv2

bolariin:
resolves https://github.com/opencollective/opencollective/issues/3043

### DIFF
--- a/server/graphql/v2/scalar/URL.ts
+++ b/server/graphql/v2/scalar/URL.ts
@@ -1,0 +1,20 @@
+import { URL } from 'url';
+
+import { GraphQLScalarType } from 'graphql';
+
+import { ValidationFailed } from '../../errors';
+
+export default new GraphQLScalarType({
+  name: 'URL',
+  description: "A field whose value conforms to the standard URL format as specified in RFC3986: https://www.ietf.org/rfc/rfc3986.txt.",
+  parseValue(value: string): string {
+    try {
+      return new URL(value).toString()
+    } catch {
+      throw new ValidationFailed(`Not a valid URL: ${value}`);
+    }
+  },
+  serialize(value: string): string {
+    return value;
+  },
+});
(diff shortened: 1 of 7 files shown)


server/graphql/v2/scalar/URL.ts
```diff
@@ -0,0 +1,17 @@
+import { GraphQLScalarType } from 'graphql';
+import { ValidationFailed } from '../../errors';
+
+export default new GraphQLScalarType({
+  name: 'URL',
+  description: "A field whose value conforms to the standard URL format as specified in RFC3986: https://www.ietf.org/rfc/rfc3986.txt.",
+  parseValue(value: string): string {
+    try {
+      return new URL(value).toString()
```
znarf, line 9: Why is `URL` not imported? I'm curious to know how that works.

`import { URL } from 'url';`

znarf:
What is the release strategy? I guess it's a breaking change and frontend needs to be updated at the same time as API?

server/graphql/v2/scalar/URL.ts
bolariin: I noticed that in that `URL` was not imported in code sample that was [linked](https://github.com/Urigo/graphql-scalars/blob/master/src/resolvers/URL.ts) in the issue description. When i tested it out and it worked I just assumed I didn't need to add the import statement.

I will change this to include the import statement

bolariin:
> What is the release strategy? I guess it's a breaking change and frontend needs to be updated at the same time as API?

I'm not too sure but I would assume it would be better to update the frontend at the same time.

Betree:
> I'm not too sure but I would assume it would be better to update the frontend at the same time.

Agree. I don't think  that it's a breaking change for frontend so to speak because if we deploy only the API, I think the frontend will still work (the input is still a string - only the validations are different).

**But** it will break the CI for frontend, because the schema will be outdated and ESLint will not be happy.

TLDR ; we should merge frontend & API at the same time but no need to deploy at the same time.

server/graphql/v2/scalar/URL.ts
Betree: `URL` is in the global scope (https://nodejs.org/api/url.html). Agree that it's better to explicitly import, an ESLint rule for that could be nice.

znarf: @Betree Ok. `URL` is effectively part of node globals. https://nodejs.org/api/globals.html

Perfect, both forms are acceptable then.

PR #3698, opened 2020-04-13
fix(export:csv): using e.tags instead of e.category

xdamman:
Export of open collective on drive.opencollective.com was not up to date.
It has to be run manually by someone who has access to the database (read).

This is the command to run:

```PG_DATABASE=opencollective_prod_snapshot npm run export:csv```

It was broken because the schema for the `Expenses` table has changed.
This is fixing it.

Related to https://github.com/opencollective/opencollective/issues/3004 (but financial data still needs to be updated)

### DIFF
--- a/scripts/monthly_data_exports.js
+++ b/scripts/monthly_data_exports.js
@@ -71,7 +71,7 @@ const queries = [
     query: `
     SELECT
     t."createdAt", c.slug as "collective slug", t.type as "transaction type", t.amount::float / 100,
-    t.currency, fc.slug as "from slug", fc.type as "from type", t.description, e.category as "expense category",
+    t.currency, fc.slug as "from slug", fc.type as "from type", t.description, e.tags as "expense tags",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by znarf** [comment]
Thank you! Tests are not passing. Having a try with a rebase.

File: scripts/monthly_data_exports.js [hunk 1]
```diff
@@ -71,7 +71,7 @@ const queries = [
     query: `
     SELECT
     t."createdAt", c.slug as "collective slug", t.type as "transaction type", t.amount::float / 100,
-    t.currency, fc.slug as "from slug", fc.type as "from type", t.description, e.category as "expense category",
+    t.currency, fc.slug as "from slug", fc.type as "from type", t.description, e.tags as "expense tags",
```
kewitz, line 74: Detail: the category was migrated as the first tag of the expense.
If you want to fetch that you can use `e.tags[1] as category`, otherwise the output will have the array of all tags displayed like `"[""OTHER""]"`.

PR #3701, opened 2020-04-14
Customize expense payee

Betree:
Ability to use organization profiles as payees for expenses if they're not hosts.

Also added a filter on `isMemberOf` to filter out hosts on frontend payee select.

### DIFF
--- a/server/graphql/v2/mutation/ExpenseMutations.ts
+++ b/server/graphql/v2/mutation/ExpenseMutations.ts
@@ -46,7 +46,7 @@ const expenseMutations = {
         amount: items.reduce((total, item) => total + item.amount, 0),
         PayoutMethod: payoutMethod,
         collective: await fetchAccountWithReference(args.account, req),
-        fromCollective: args.expense.payee,
+        fromCollective: await fetchAccountWithReference(args.expense.payee, { throwIfMissing: true }),
       });
     },
   },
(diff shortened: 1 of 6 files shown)


server/graphql/v2/mutation/ExpenseMutations.ts
```diff
@@ -46,7 +46,7 @@ const expenseMutations = {
         amount: items.reduce((total, item) => total + item.amount, 0),
         PayoutMethod: payoutMethod,
         collective: await fetchAccountWithReference(args.account, req),
-        fromCollective: args.expense.payee,
+        fromCollective: await fetchAccountWithReference(args.expense.payee, { throwIfMissing: true }),
```
kewitz, line 49: We don't need to add any checks here because we're using legacy methods. ✅

server/graphql/v2/mutation/ExpenseMutations.ts
```diff
@@ -89,7 +89,7 @@ const expenseMutations = {
           id: attachedFile.id && idDecode(attachedFile.id, IDENTIFIER_TYPES.EXPENSE_ITEM),
           url: attachedFile.url,
         })),
-        fromCollective: null, // TODO payee
+        fromCollective: await fetchAccountWithReference(expense.payee, { throwIfMissing: true }),
```
kewitz, line 92: Same works here.

kewitz:
Everything looks clean, just make sure tests are passing.

PR #3702, opened 2020-04-14
Fix missing expense.category updates

kewitz:
Related to #3640.


cron/monthly/user-report.js
```diff
@@ -457,16 +457,16 @@ const computeStats = async (collectives, currency = 'USD') => {
       stats.expenses += expenses.length;
       await Promise.map(expenses, async expense => {
         const amountInBackerCurrency = await convertToCurrency(expense.amount, expense.currency, currency);
-        categories[expense.category] = categories[expense.category] || {
```
Betree, line 460: I don't know if it has an incidence, but acknowledging that we now store expenses without categories in `categories[undefined]` instead of `categories[null]`

PR #3713, opened 2020-04-15
Refactor: make express initialization async

znarf:
Trying to unblock https://github.com/opencollective/opencollective-api/pull/3708

The Forest Middleware is moving to an async signature and that is forcing us to reconsider the way we're initializing `express`.

I'm not 100% happy, let me know if you have ideas to make that better.

### DIFF
--- a/scripts/forestadmin-schema-update.js
+++ b/scripts/forestadmin-schema-update.js
@@ -7,4 +7,8 @@ if (!process.env.FOREST_ENV_SECRET || !process.env.FOREST_AUTH_SECRET) {
   process.exit();
 }
 
-forest.init();
+async function main() {
+  await forest.init();
+}
+
+main();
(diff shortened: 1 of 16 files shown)


scripts/forestadmin-schema-update.js
```diff
@@ -7,4 +7,8 @@ if (!process.env.FOREST_ENV_SECRET || !process.env.FOREST_AUTH_SECRET) {
   process.exit();
 }
 
-forest.init();
+async function main() {
+  await forest.init();
+}
```
Betree, line 12: [Top-level await](https://github.com/tc39/proposal-top-level-await) is at stage 3. Soon!

znarf: Before writing that, I looked if babel could support it. They support the syntax, but not the feature.

PR #3714, opened 2020-04-15
Make sure to use GraphQL errors

znarf:
I was checking this file and I noticed that we were still relying on `lib/errors'.

Is that expected @kewitz ?


znarf:
Ok, extended the PR to cover all errors.

Even when it was loading the right error library, I made the usage consistent (named import).

PR #3725, opened 2020-04-17
Fix ValidationError when slug clashes with existing deleted one.

kewitz:
This is enforced either way through index.

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -200,6 +200,7 @@ export async function createCollectiveFromGithub(_, args, req) {
   if (process.env.NODE_ENV !== 'production' && user.email.match(/.*test.*@opencollective.com$/)) {
     const existingCollective = models.Collective.findOne({
       where: { slug: collectiveData.slug.toLowerCase() },
+      paranoid: false,
     });
     if (existingCollective) {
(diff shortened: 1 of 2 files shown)


Betree:
Nice, it would be great to better display sql errors but this fix does the job

znarf:
@kewitz Could you audit the codebase and find similar occurrences?

znarf:
Also note that when we delete accounts using the normal path, we're already making sure to free the slug, see `deleteCollective` in GraphQL v1.

kewitz:
@znarf, nice catch! So this is really an edge-case... should we just hard-delete test-collective?

znarf:
We should make sure to free all the slug of soft deleted collectives.

kewitz:
Actually, we shouldn't solve it like this.
No point in include deleted rows in the query since it will be an ever-increasing complexity to process.
This is a super-edge case, I'll make sure we're freeing up slugs and write a migration to free-up soft-deleted sluges.

kewitz:
@Betree since our policy is to not hard delete collectives, I don't like to have an ever-increasing query tied to the onboard flow.
Also, in the case of us failing to validate the existing collectives, the unique validator will throw and we'll be able to properly display the Sequelize validation error in the frontend.

PR #3727, opened 2020-04-17
Add Information About Installing Postgres Client Tools on Linux

SudharakaP:
This adds information to postgres.md about installing Postgres client tools on Ubuntu. We currently only have information about Postgres client tool installation on Mac. :smile:

### DIFF
--- a/docs/postgres.md
+++ b/docs/postgres.md
@@ -35,12 +35,20 @@ export PGHOST=localhost
 export PGUSER=postgres
 ```
 
-You'll also need to have Postgres client tools like `psql`, `dropdb`, `createuser` locally available to run our scripts. In macOS you can install those using Homebrew with:
+You'll also need to have Postgres client tools like `psql`, `dropdb`, `createuser` locally available to run our scripts.
+
+#### MacOS
+
+In macOS you can install those using Homebrew with:
 
 ```
 brew install libpq
 echo 'export PATH="/usr/local/opt/libpq/bin:$PATH"' >> ~/.bash_profile
 ```
+
+#### Linux
+
+##### Debian Based Distributions
 
 For Ubuntu 16.04 and above you can execute the following to install Postgres client tools:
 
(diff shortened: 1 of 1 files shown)


SudharakaP:
@znarf : Thanks for the feedback. I've changed it. Let me know if anything needs further change. :smile:

SudharakaP:
> Please check again, you added a Mac section while one was already existing.

Sorry, I am not quite understanding what you have in mind. :smile: 

But this [MacOS and Linux section](https://github.com/SudharakaP/opencollective-api/blob/6c09179179f1deb94e3ccf02d8f6b8e81bbd3335/docs/postgres.md#macos) should be enclosed within the [Docker section](https://github.com/SudharakaP/opencollective-api/blob/6c09179179f1deb94e3ccf02d8f6b8e81bbd3335/docs/postgres.md#with-docker). The [first MacOS section](https://github.com/SudharakaP/opencollective-api/blob/6c09179179f1deb94e3ccf02d8f6b8e81bbd3335/docs/postgres.md#on-macos) is about how to install Postgres on a MacOS. The ones we are talking about are subheadings within Docker; on how to install the Postgres client tools after pulling the docker images. 

Would you like to create a new Docker subheading within the [first MacOS section](https://github.com/SudharakaP/opencollective-api/blob/6c09179179f1deb94e3ccf02d8f6b8e81bbd3335/docs/postgres.md#on-macos) ? :thinking:

znarf:
Sorry, I got it wrong and made it more complicated that it should. You can switch back to the first version and it should be good.

Remove:
- ```#### MacOS```
- ```#### Linux```
- ```##### Debian Based Distributions```

SudharakaP:
@znarf : No problem. Done. :smile: 

On another note I see that you guys run the CI on every single commit; even documentation changes. Would you like to add support for skipping the CI when the commit header has [skip ci] or similar tags? :smile: If that's something useful; I can add that as well.

PR #3586, opened 2020-03-25
Implement Sentry on the API 

cdvx:
Resolves:https://github.com/opencollective/opencollective/issues/2997

chore(api-errors): Implement Sentry on the API
 - Implement sentry for rich error reporting
[resolves 2997]

PR #3741, opened 2020-04-20
Implement Sentry on the API

Betree:
Follow up on https://github.com/opencollective/opencollective-api/pull/3586 after it was reverted in https://github.com/opencollective/opencollective-api/pull/3740. It broke the build when merged, we missed it because we don't run our CI on API by default.


cdvx:
Looking into this

Betree:
Closing for now, will revisit later

PR #3742, opened 2020-04-20
Fix empty webhookUrl

kewitz:
Resolve https://github.com/opencollective/opencollective/issues/3071

The `bulkCreate` method would skip hooks, so I mapped normal create methods.

### DIFF
--- a/server/models/Notification.js
+++ b/server/models/Notification.js
@@ -1,9 +1,10 @@
 import Promise from 'bluebird';
 import debugLib from 'debug';
-import { defaults } from 'lodash';
+import { defaults, isNil } from 'lodash';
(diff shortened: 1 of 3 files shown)


server/models/Notification.js
```diff
@@ -78,6 +79,14 @@ export default function (Sequelize, DataTypes) {
           type: 'unique',
         },
       ],
+      hooks: {
+        beforeCreate(instance) {
+          console.log(instance);
```
Betree, line 84:
```suggestion
```

kewitz:
@Betree, we could, but like the idea of throwing standard errors. If we lower the validation level with DB constraints, we would need a way to digest and pass that error to the frontend.

PR #3743, opened 2020-04-20
Fix unused Collective slugs after deletion

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3075

### DIFF
--- a/server/lib/forest.js
+++ b/server/lib/forest.js
@@ -116,6 +116,8 @@ export default async function (app) {
           });
         });
         // Delete Collective
+        const newSlug = `${collective.slug}-${Date.now()}`;
+        await collective.update({ slug: newSlug });
         await collective.destroy();
       })
       .then(() => {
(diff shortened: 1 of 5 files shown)


server/lib/forest.js
```diff
@@ -116,6 +116,8 @@ export default async function (app) {
           });
         });
         // Delete Collective
+        const newSlug = `${collective.slug}-${Date.now()}`;
+        await collective.update({ slug: newSlug });
```
znarf, line 120: Thank you for doing that!

I forgot we had duplicated logic to delete collectives. Could we follow up and mutualize the code somewhere?

kewitz:
@znarf, DRYed that up in a `beforeDestroy` hook and created a test.

znarf:
@kewitz Thank you. I was also thinking about the deletion of all related entries. We should mutualize that too, but that's another issue/PR.

PR #3749, opened 2020-04-21
Extend expenses permission library

Betree:
Related to https://github.com/opencollective/opencollective/issues/2936

Add new fields related to processing expenses to `ExpensePermissions` in GQLV2:
- `canApprove`
- `canDeleteExpense`
- `canEditExpense`
- `canMarkAsUnpaid`
- `canPayExpense`
- `canReject`
- `canUnapprove`

The frontend so retrieves a single object that tells you what the permissions are, and which buttons you should display:
```es6
{
  canApprove: false,
  canDelete: false,
  canEdit: true,
  canMarkAsUnpaid: false,
  canPay: true,
  canReject: false,
  canSeeInvoiceInfo: true,
  canUnapprove: true
}
```

Also did a bit of refactoring without going too far to try to use the same helpers between GQLV1 & V2. All these new helpers have matching unit tests.

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -1,29 +1,46 @@
 import { expenseStatus, roles } from '../../constants';
+import FEATURE from '../../constants/feature';
+import { canUseFeature } from '../../lib/user-permissions';
 import { ExpenseItem } from '../../models/ExpenseItem';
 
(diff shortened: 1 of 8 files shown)


Betree:
@kewitz thanks!

> I think this is exactly the place where unit-testing excels

I agree, it's also the one where we should always try to reach 100% coverage :slightly_smiling_face:

PR #3757, opened 2020-04-22
Better manage currency updates

znarf:
Fix https://github.com/opencollective/opencollective/issues/3072 but also more edge cases.

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1530,6 +1530,19 @@ export default function (Sequelize, DataTypes) {
     });
   };
 
+  /**
+   * Get Hosted Collectives
+   *
+   * It's expected that child Collectives like EVENTS are returned
+   */
+  Collective.prototype.getHostedCollectives = async function () {
+    const hostedCollectives = await models.Member.findAll({
+      where: { MemberCollectiveId: this.id, role: roles.HOST },
+    });
+    const hostedCollectiveIds = hostedCollectives.map(m => m.CollectiveId);
+    return models.Collective.findAll({ where: { id: { [Op.in]: hostedCollectiveIds } } });
+  };
+
   Collective.prototype.updateHostFee = async function (hostFeePercent, remoteUser) {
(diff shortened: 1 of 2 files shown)


server/models/Collective.js
```diff
@@ -1560,21 +1569,55 @@ export default function (Sequelize, DataTypes) {
...
+      }
+      const collectives = await this.getHostedCollectives();
+      // We use setCurrency so that it will cascade to Tiers
+      await Promise.map(collectives, collective => collective.setCurrency(currency), { concurrency: 3 });
+    }
+
+    // This is currently for COLLECTIVE and EVENTS but we make it generic
+    const tiers = await this.getTiers();
+    if (tiers.length > 0) {
+      await Promise.map(tiers, tier => tier.update({ currency }), { concurrency: 3 });
```
kewitz, line 1617: Should we recalculate the tier value based on FXRate?
I understand why this wouldn't be necessary for some currencies, but there are currencies that this update wouldn't make sense, for example, USD to Yen.

znarf: I thought about this problem but I don't want to go that way and end up with weird values.

We could delete the tiers completely. But, if it's the default tiers, we're not smart enough anyway, it's what we're currently doing (same for all currencies):

```
  if (tiers.length === 0) {
    tiers.push({
      type: 'TIER',
      name: 'backer',
      slug: 'backers',
      amount: 500,
      presets: [500, 1000, 2500, 5000],
      interval: 'month',
      currency: currency,
      minimumAmount: 500,
      amountType: 'FLEXIBLE',
    });
    tiers.push({
      type: 'TIER',
      name: 'sponsor',
      slug: 'sponsors',
      amount: 10000,
      presets: [10000, 25000, 50000],
      interval: 'month',
      currency: currency,
      minimumAmount: 10000,
      amountType: 'FLEXIBLE',
    });
  }
```

If we're in this situation, we're really at the beginning of the collective existence, we don't expect to have processed any transaction.

I don't see a simple solution so we will keep it like that for now.

kewitz: Agreed, I don't think there's an easy solution for this. :shipit:

PR #3760, opened 2020-04-22
GQLV2: Endpoints to process expenses

Betree:
Rather than adding multiple mutations for processing expenses, I chose to add a single `processExpense` endpoint to handle all actions that are meant to change the `status` of an expense.

```ts
processExpense(
  expense: Expense,
  action: APPROVE | UNAPPROVE | REJECT | MARK_AS_UNPAID | PAY,
  paymentParams: ProcessExpensePaymentParams
)
```

The main advantage of doing so is that it simplifies the API usage:

```es6
export const ProcessExpenseButtons = () => {
  const [processExpense] = useMutation(processExpense);
  return (
    <React.Fragment>
      {permissions.canApprove && (
        <StyledButton onClick={() => processExpense({ variables: { action: 'APPROVE' } })}>
          <FormattedMessage id="actions.approve" defaultMessage="Approve" />
        </StyledButton>
      )}
      {permissions.canReject && (
        <StyledButton onClick={() => processExpense({ variables: { action: 'REJECT' } })}>
          <FormattedMessage id="actions.reject" defaultMessage="Reject" />
        </StyledButton>
      )}
    </React.Fragment>
  )
}
```

### DIFF
--- a/server/graphql/v2/object/Host.js
+++ b/server/graphql/v2/object/Host.js
@@ -1,8 +1,10 @@
-import { GraphQLBoolean, GraphQLInt, GraphQLObjectType } from 'graphql';
+import { GraphQLBoolean, GraphQLInt, GraphQLNonNull, GraphQLObjectType } from 'graphql';
 import { get } from 'lodash';
 
 import { Account, AccountFields } from '../interface/Account';
 import URL from '../scalar/URL';
+
+import { HostPlan } from './HostPlan';
 
 export const Host = new GraphQLObjectType({
   name: 'Host',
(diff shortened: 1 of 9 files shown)


server/graphql/v2/object/Host.js
```diff
@@ -36,6 +38,12 @@ export const Host = new GraphQLObjectType({
           return get(collective, 'settings.tos');
         },
       },
+      plan: {
+        type: new GraphQLNonNull(HostPlan),
+        resolve(host) {
+          return host.getPlan();
+        },
+      },
```
Betree, line 46: I had to add this to GQLV2 because the "Pay" button needs to know about the host plan.

PR #3761, opened 2020-04-22
v2: Let users who are admin of a Collective host this Collective

sbinlondon:
I hope I am understanding this correctly.

* The v2 `applyToHost` was checking a single item - `canApply` - to see whether a Collective could apply to a Host.
* for a User who wants to fiscally host a Collective themselves, `isHostAccount` will automatically be `true` but they do not automatically have the `canApply` setting set to true in their settings.
* In the v1 `changeHost` which changes or adds a host to a Collective, it checks whether the potential Host is a Host account via the first method (`isHostAccount`) and then `addHost` within it goes on to check whether a user is the admin of the applying Collective before automatically approving that they can also be the Host

This PR for now adds a line in the v2 `applyToHost` mutation to check in the case of `canApply` being false, whether the host is a `USER`, and then allows `changeHost` and `addHost` to do the work of checking whether they're an admin of the Collective and therefore can host it.

### DIFF
--- a/server/graphql/v2/mutation/CollectiveMutations.ts
+++ b/server/graphql/v2/mutation/CollectiveMutations.ts
@@ -1,7 +1,7 @@
 import { GraphQLNonNull } from 'graphql';
 
 import { types as collectiveTypes } from '../../../constants/collectives';
-import { NotFound,Unauthorized } from '../../errors';
+import { NotFound, Unauthorized } from '../../errors';
 import { AccountReferenceInput, fetchAccountWithReference } from '../input/AccountReferenceInput';
 import { Collective } from '../object/Collective';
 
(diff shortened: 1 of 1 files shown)


sbinlondon:
I am less clear on how to solve the same issue for organizations

server/graphql/v2/mutation/CollectiveMutations.ts
```diff
@@ -47,7 +47,10 @@ const collectiveMutations = {
       }
       const canApply = await host.canApply();
       if (!canApply) {
-        throw new Error('Host is not open to applications');
+        // in the case that the host is the USER, it will be checked in changeHost and addHost whether they are admin of the Collective they are trying to host
+        if (host.type !== 'USER') {
+          throw new Error('Host is not open to applications');
+        }
```
znarf, line 53: I think the proper solution is to delete that whole block.

`addHost` will be smart enough to do the right thing.

znarf:
Also ditch the previous block:

```
      const isHost = await host.isHost();
      if (!isHost) {
        throw new Error('Account is not an host');
      }
```

server/graphql/v2/mutation/CollectiveMutations.ts
znarf: See:

```
    // If we can't automatically approve the collective and it is not open to new applications, reject it
    if (!shouldAutomaticallyApprove && !hostCollective.canApply()) {
      throw new Error('This host is not open to applications');
    }
```

PR #3770, opened 2020-04-24
refactor: improve how summaries are generated for updates

bolariin:
Resolves https://github.com/opencollective/opencollective/issues/2547

This PR adds a new `summary` column to `Updates` table and improved how summaries are generated by using `generateSummaryForHTML` helper which is also used to generate summaries for conversations.

### DIFF
--- a/migrations/20200423203719-add-summary-to-updates.js
+++ b/migrations/20200423203719-add-summary-to-updates.js
@@ -1,0 +1,42 @@
+'use strict';
+
+import Promise from 'bluebird';
+
+import { generateSummaryForHTML } from '../server/lib/sanitize-html';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('Updates', 'summary', { type: Sequelize.STRING });
+    await queryInterface.addColumn('UpdateHistories', 'summary', { type: Sequelize.STRING });
+
+    const updates = await queryInterface.sequelize.query(
+      `
+        SELECT * FROM "Updates"
+        WHERE "summary" IS NULL;
+      `,
+      { type: Sequelize.QueryTypes.SELECT },
+    );
+
+    // Update newly created summary column for existing update entries
+    await Promise.map(updates, updateEntry =>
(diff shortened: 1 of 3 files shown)


bolariin:
I have a question about the `UpdateHIstories` table. It seems like the entries in `UpdateHistories` are also in the `Updates` table and I was wondering what function `UpdateHistories` provides and how is it linked to the `Updates` table?

bolariin:
I was wondering if it's okay to assume that `html` field in `Update` will always have a value? I ask because I noticed a conditional in the `Update` model. 

https://github.com/opencollective/opencollective-api/blob/b35d772366c2c64575947eeff63262a3ebc9fa74/server/models/Update.js#L109-L114

Betree:
@bolariin Sorry for not seeing your comments before, feel free to mention or assign me as a reviewer when working on issues I've authored :slightly_smiling_face: 

> I have a question about the `UpdateHIstories` table. It seems like the entries in `UpdateHistories` are also in the `Updates` table and I was wondering what function `UpdateHistories` provides and how is it linked to the `Updates` table?

This table contains snapshots of the rows in `Updates` for every change (UPDATE or DELETE). This behavior is implemented by [sequelize-temporal](https://github.com/bonaval/sequelize-temporal), that is plugged on the model with `Temporal(Update, Sequelize);`.

> I was wondering if it's okay to assume that `html` field in `Update` will always have a value? I ask because I noticed a conditional in the `Update` model.

I just checked, and there's no entry in the DB where `html=NULL`. We do have 11 updates from 2018 with empty `html` content, but it seems like a bug that has been solved since.

bolariin:
> @bolariin Sorry for not seeing your comments before, feel free to mention or assign me as a reviewer when working on issues I've authored 

Thanks! I will do this from now on

> This table contains snapshots of the rows in `Updates` for every change (UPDATE or DELETE). This behavior is implemented by [sequelize-temporal](https://github.com/bonaval/sequelize-temporal), that is plugged on the model with `Temporal(Update, Sequelize);`.

Thanks for explaining this to me :smile:

PR #3776, opened 2020-04-25
refactor: migrate tags to lowercase

bolariin:
Resolves https://github.com/opencollective/opencollective/issues/3085
Related https://github.com/opencollective/opencollective-frontend/pull/4101

 - [x] Create a migration that transforms all tags for conversations and expenses
 - [x] Update sanitizeTags
 - [x] Update tests

### DIFF
--- a/migrations/20200425013627-change-tags-to-lowercase.js
+++ b/migrations/20200425013627-change-tags-to-lowercase.js
@@ -1,0 +1,39 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(
+      `
+        UPDATE "Conversations" c
+        SET "tags" = lower(tags::text)::text[]
+        WHERE c."tags" IS NOT NULL
+      `,
+    );
+
+    await queryInterface.sequelize.query(
+      `
+        UPDATE "Expenses" e
+        SET "tags" = lower(tags::text)::text[]
+        WHERE e."tags" IS NOT NULL
+      `,
+    );
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(
+      `
+        UPDATE "Conversations" c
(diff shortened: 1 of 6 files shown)


migrations/20200425013627-change-tags-to-lowercase.js
```diff
@@ -0,0 +1,161 @@
...
+      queryInterface.sequelize.query(
+        `
+        UPDATE "Conversations" c
+        SET "tags" = string_to_array(:tags_as_string, ',')
+        WHERE c.id = :id
+      `,
+        {
+          replacements: {
+            id: conversation.id,
+            tags_as_string: convertTagsToLowerCase(conversation),
```
Betree, line 33: A pure SQL solution using [`LOWER`](https://w3resource.com/PostgreSQL/lower-function.php) would be more stable and performant, did you consider using it?

bolariin: No, I didn't consider using `LOWER`. I will update the query.

migrations/20200425013627-change-tags-to-lowercase.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(
+      `
+        UPDATE ONLY "Conversations" c
+        SET "tags" = c_with_lowercase_tags."tags"
+        FROM (SELECT id, lower(tags::text)::text[] as tags from "Conversations") as c_with_lowercase_tags
+        WHERE c.id = c_with_lowercase_tags.id AND c."tags" IS NOT NULL
```
Betree, line 10: Are these join (`FROM`) really necessary? Your approach is good, but I think we can simplify it: 

```suggestion
        UPDATE ONLY "Conversations" c
        SET "tags" = lower(tags::text)::text[] 
        WHERE "tags" IS NOT NULL
```

PR #3789, opened 2020-04-28
Store payee's location in expense

Betree:
To implement the new `address` fields added by design in the expense form to have it in the invoice and that we decided during the call to store inside of expenses.

- Add a `payeeLocation` column on `Expenses`
- Updates all existing expenses to store `fromCollective`'s locations in them
- Add supports for passing a `payeeLocation` in expense mutations (GQLV2 only)
- If no `payeeLocation` is passed, default to `fromCollective.location`
- `expense.payeeLocation` is only visible to expense owner and host admins


server/graphql/v2/input/LocationInput.ts
```diff
@@ -0,0 +1,28 @@
...
+    name: {
+      type: GraphQLString,
+      description: 'A short name for the location (eg. Open Collective Headquarters)',
+    },
+    address: {
+      type: GraphQLString,
+      description: 'Postal address without country (eg. 12 opensource avenue, 7500 Paris)',
+    },
+    country: {
+      type: GraphQLString,
```
Betree, line 16: I'm going to turn this one into an enum or a scalar

PR #3790, opened 2020-04-28
Use structured Bank Information

kewitz:
Resolves opencollective/opencollective/issues/3093
- [x] Add create PayoutMethod mutation
- [x] Render bank account information in the manual email template

### DIFF
--- a/server/graphql/v2/mutation/PayoutMethodMutations.ts
+++ b/server/graphql/v2/mutation/PayoutMethodMutations.ts
@@ -1,0 +1,80 @@
+import { GraphQLNonNull } from 'graphql';
+import { pick } from 'lodash';
+
+import models from '../../../models';
+import { Unauthorized } from '../../errors';
+import { AccountReferenceInput, fetchAccountWithReference } from '../input/AccountReferenceInput';
+import { PayoutMethodInput } from '../input/PayoutMethodInput';
+import PayoutMethod from '../object/PayoutMethod';
+
+const payoutMethodMutations = {
+  createPayoutMethod: {
+    type: PayoutMethod,
+    description: 'Create a new Payout Method to get paid through the platform',
+    args: {
+      payoutMethod: {
+        type: new GraphQLNonNull(PayoutMethodInput),
+        description: 'Payout Method data',
+      },
+      account: {
+        type: new GraphQLNonNull(AccountReferenceInput),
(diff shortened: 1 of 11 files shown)


server/graphql/v2/mutation/PayoutMethodMutations.ts
```diff
@@ -0,0 +1,80 @@
...
+  //       throwIfMissing: true,
+  //     });
+  //     if (!req.remoteUser.isAdmin(connectedAccount.CollectiveId)) {
+  //       throw new Unauthorized("You don't have permission to edit this collective");
+  //     }
+
+  //     await connectedAccount.destroy({ force: true });
+  //     return connectedAccount;
+  //   },
+  // },
```
Betree, line 77:
```suggestion
```

server/graphql/v2/object/PayoutMethod.ts
```diff
@@ -24,7 +24,7 @@ const PayoutMethod = new GraphQLObjectType({
       resolve: (payoutMethod, _, req): string => {
         // Only collective admins can see the name of a payout method
         if (req.remoteUser?.isAdmin(payoutMethod.CollectiveId)) {
-          return payoutMethod.isSaved;
+          return payoutMethod.name;
```
Betree, line 27: :+1:

test/test-helpers/fake-data.js
```diff
@@ -16,6 +16,8 @@ import models from '../../server/models';
 import { PayoutMethodTypes } from '../../server/models/PayoutMethod';
 import { randEmail, randUrl } from '../stores';
 
+export { PAYMENT_METHOD_SERVICES, PAYMENT_METHOD_TYPES, PayoutMethodTypes };
```
Betree, line 19: **[:ice_cream: Details & personal preferences]** Sounds like unnecessary complexity. Why not importing directly from the `/constants` files?

---

<details>
<br/>

:no_entry_sign: **Blocker**

:sparkles: **Enhancement**

</details>

kewitz: The idea is to allow tests to import these constants from the fake-data in order to reduce the number of imports we have in our tests.
Do you think this will cause more confusion?

kewitz:
Rebased.

znarf:
Added a small commit.

PR #3798, opened 2020-04-30
Add Support for Expense Tag Suggestions

SudharakaP:
This adds support for expense tag suggestions for the api layer.

Related to https://github.com/opencollective/opencollective/issues/3088


SudharakaP:
> Works like a charm ok_hand

Wonderful. :smile:

PR #3804, opened 2020-04-30
getHosts: Filter on isHostAccount

Betree:


Betree:
@znarf This test is expecting 11 hosts, but this PRs filters them to 2 https://github.com/opencollective/opencollective-frontend/blob/a75cbd282586c00bd435ca1ed1c339e63d79e8e0/test/cypress/integration/21-collective.create.test.js#L31. It may be related to the problem we have locally where `amtv` has `isHostAccount` to false. Should we investigate that or just fix the E2E test?

Betree:
Closing this one since https://github.com/opencollective/opencollective/issues/3083 was closed. Didn't get feedback to know if this was still needed or not.

PR #3807, opened 2020-04-30
Fix Truncation Issue with the Sanitized Conversation String

SudharakaP:
This fixes the truncation issue found in sanitized conversations.

Resolves https://github.com/opencollective/opencollective/issues/3086

---

*Edit by @Betree: requires https://github.com/opencollective/opencollective-frontend/pull/4404*

### DIFF
--- a/server/lib/sanitize-html.ts
+++ b/server/lib/sanitize-html.ts
@@ -137,8 +137,17 @@ export const generateSummaryForHTML = (content, maxLength = 255) => {
     .replace(/\s+/g, ' ');
 
   // Truncate
-  const truncated = truncate(trimmed, { length: maxLength });
+  const truncated = truncate(trimmed, { length: maxLength, omission: ''});
 
   // Second sanitize pass: an additional precaution in case someones finds a way to play with the trimmed version
-  return sanitizeHTML(truncated, optsSanitizeSummary);
+  const secondSanitized = sanitizeHTML(truncated, optsSanitizeSummary);
+
+  const isTruncated = trimmed.length > maxLength;
+
+  // Check to see if the second sanitization cuts a html tag in the middle
+  if (isTruncated) {
+    return `${secondSanitized.trim()}...`;
+  } else {
+    return secondSanitized;
+  }
 };
--- a/test/server/lib/sanitize-html.test.js
+++ b/test/server/lib/sanitize-html.test.js
@@ -129,19 +129,19 @@ describe('server/lib/sanitize-html', () => {
 
   describe('generateSummaryForHTML', () => {
     it('Sanitizes, trim and truncate', () => {
-      expect(generateSummaryForHTML(fullContent, 4)).to.to.eq('E...');
-      expect(generateSummaryForHTML(fullContent, 32)).to.to.eq('Ergo illi intellegunt quid Ep...');
+      expect(generateSummaryForHTML(fullContent, 4)).to.to.eq('Ergo...');
+      expect(generateSummaryForHTML(fullContent, 32)).to.to.eq('Ergo illi intellegunt quid Epicu...');
       expect(generateSummaryForHTML(fullContent.slice(272), 80)).to.to.eq(
-        'Reges: constructio interrete. <i>Suo genere perveniant ad extremum;</i> Atqui...',
+        'Reges: constructio interrete. <i>Suo genere perveniant ad extremum;</i> Atqui pu...',
       );
       expect(generateSummaryForHTML(fullContent.slice(1000), 1000)).to.to.eq(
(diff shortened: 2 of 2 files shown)


SudharakaP:
@Betree : Done. I had to run a full sanitization to check whether the string contains the `...` at the end or not. Because there can be two cases; 

1) Where the truncated string is equal to, `I'd like to say <stro...` or 

2) Where the truncated string is equal to, `I'd like to say <strong>Hel...`. 

In the first case the final sanitation results in `I'd like to say ` whereas in the second it will be, `I love <strong> Hello ...</strong>`. The only way I can think of to handle both cases is to use a sanitization at then end to get rid of all html tags and then see if it contains the `...`. Let me know if you agree. :smile:

server/lib/sanitize-html.ts
```diff
@@ -140,5 +140,15 @@ export const generateSummaryForHTML = (content, maxLength = 255) => {
   const truncated = truncate(trimmed, { length: maxLength });
 
   // Second sanitize pass: an additional precaution in case someones finds a way to play with the trimmed version
-  return sanitizeHTML(truncated, optsSanitizeSummary);
+  const secondSanitized = sanitizeHTML(truncated, optsSanitizeSummary);
+
+  const isTruncated = trimmed.length > maxLength;
+  const fullySanitized = sanitizeHTML(secondSanitized);
```
Betree, line 146: Sanitizing an already-sanitized string won't have any effect. It's useful after truncating because it will clean invalid markup, but here we're basically doing `sanitizeHTML(sanitizeHTML(truncated, optsSanitizeSummary))`.

Also we must not forget to pass `optsSanitizeSummary` to the function when we use it.

server/lib/sanitize-html.ts
```diff
@@ -140,5 +140,15 @@ export const generateSummaryForHTML = (content, maxLength = 255) => {
...
+
+  // Check to see if the second sanitization cuts a html tag in the middle
+  if (isTruncated && fullySanitized.substring(fullySanitized.length - 3, fullySanitized.length) !== '...') {
```
Betree, line 149: A simpler way to write that

```suggestion
  if (isTruncated && fullySanitized.slice(fullySanitized.length - 3) !== '...') {
```

server/lib/sanitize-html.ts
SudharakaP: The only reason I am using that final sanitize (`const fullySanitized = sanitizeHTML(secondSanitized)`) is to see if it contains the `...` in the if statement below that. Since we don't pass the `optsSanitizeSummary` this will remove all the html tags and thus will give us whether or not it's actually truncated or not. Do you have a better idea to handle this case? :thinking:

server/lib/sanitize-html.ts
SudharakaP: Thanks. Will correct that. :smile:

server/lib/sanitize-html.ts
Betree: I get the idea, I missed that during my first review. The `...` should normally always be outside of tags, we don't want them to be included in whatever styling (bold, strikethrough) that may be occurring here.

So I'd say it should be safe to test `secondSanitized` directly:
```es6
const isTruncated = trimmed.length > maxLength;
const secondSanitized = sanitizeHTML(truncated, optsSanitizeSummary);

// Check to see if the second sanitization cuts a html tag in the middle
if (isTruncated && secondSanitized.slice(secondSanitized.length - 3) !== '...') {
  return `${secondSanitized.trim()}...`;
} else {
   return secondSanitized;
}
```

SudharakaP: > I get the idea, I missed that during my first review. The `...` should normally always be outside of tags, we don't want them to be included in whatever styling (bold, strikethrough) that may be occurring here.
> 
> So I'd say it should be safe to test `secondSanitized` directly:
> 
> ```js
> const isTruncated = trimmed.length > maxLength;
> const secondSanitized = sanitizeHTML(truncated, optsSanitizeSummary);
> 
> // Check to see if the second sanitization cuts a html tag in the middle
> if (isTruncated && secondSanitized.slice(secondSanitized.length - 3) !== '...') {
>   return `${secondSanitized.trim()}...`;
> } else {
>    return secondSanitized;
> }
> ```

@Betree : This will fail the test [`Doesn't cut anchors`](https://github.com/opencollective/opencollective-api/blob/398edc15b23ec27d6ad37c013c8b4c6e37f1a3f8/test/server/lib/sanitize-html.test.js#L142). Because the if condition will pass and we'll add an additional set of `...` at the end. Am I missing something? :thinking: :thinking:

Betree: It's ok to update the test, in this example the `...` should actually be outside of the `strong` because we don't want them to be formatted. 

```es6
expect(generateSummaryForHTML("I'd like to say <strong>Hello World</strong>", 30)).to.to.eq(
  "I'd like to say <strong>Hel</strong>...",
);```

File: server/lib/sanitize-html.ts [hunk 3]
```diff
@@ -137,8 +137,17 @@ export const generateSummaryForHTML = (content, maxLength = 255) => {
...
+  const truncated = truncate(trimmed, { length: maxLength, omission: ''});
 
   // Second sanitize pass: an additional precaution in case someones finds a way to play with the trimmed version
-  return sanitizeHTML(truncated, optsSanitizeSummary);
+  const secondSanitized = sanitizeHTML(truncated, optsSanitizeSummary);
+
+  const isTruncated = trimmed.length > maxLength;
+
+  // Check to see if the second sanitization cuts a html tag in the middle
+  if (isTruncated && secondSanitized.slice(secondSanitized.length - 3) !== '...') {
```
Betree, line 148: Because we now have `omission: ''`, is this `secondSanitized.slice(secondSanitized.length - 3) !== '...'` check still necessary?

SudharakaP: @Betree : Ops, you are correct, that's not needed anymore. I've removed it. Thanks a bunch. :+1:

PR #3832, opened 2020-05-05
Update Foundation email

piamancini:
@znarf is there any way I can see a screenshot of how it looks?


znarf:
@piamancini I can finally come back to you on this.

Capture d’écran 2020-05-15 à 12 25 06

PR #3846, opened 2020-05-07
Fix includeIncognito and introduce roles for memberOf

znarf:
API support for https://github.com/opencollective/opencollective-frontend/pull/4160


znarf:
@Betree Funny! Haven't experienced it but just found it when adding roles.

Betree:
I didn't experience it neither (not sure if it broke something somewhere), just found it after looking at your frontend PR

PR #3850, opened 2020-05-08
Tax Form fix

znarf:


server/graphql/v2/object/Expense.js
```diff
@@ -211,7 +211,7 @@ const Expense = new GraphQLObjectType({
         description: 'Returns the list of legal documents required from the payee before the expense can be payed',
         async resolve(expense) {
           const incurredYear = moment(expense.incurredAt).year();
-          const isW9FormRequired = isUserTaxFormRequiredBeforePayment({
```
Betree, line 214: Good catch :facepalm: I should have added a tests here. Also one that Typescript could have helped with

PR #3857, opened 2020-05-11
Add emoji reactions to Comments

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/3059

This is work in progress. Will push up work as I go along. :smile:


SudharakaP:
@Betree : Trying to finish this up I came across some problems with the API side of things. I'll try to finish this off this week. Could you go through this PR roughly and let me know if there's anything that I am missing which you've mentioned at https://github.com/opencollective/opencollective/issues/3059#issue-600500057 or anything that needs to change. Still trying to understand the full code-base so bear with me. :smile:

migrations/20200507120000-create-comment-reactions.js
```diff
@@ -0,0 +1,51 @@
...
+        type: DataTypes.INTEGER,
+        references: { key: 'id', model: 'Users' },
+        onDelete: 'NO ACTION',
+        onUpdate: 'CASCADE',
+        allowNull: false,
+      },
+      CommentId: {
+        type: DataTypes.INTEGER,
+        references: { key: 'id', model: 'Comments' },
+        onDelete: 'NO ACTION',
```
Betree, line 21: We can delete reactions if the comment gets deleted

```suggestion
        onDelete: 'CASCADE',
```

migrations/20200507120000-create-comment-reactions.js
```diff
@@ -0,0 +1,51 @@
...
+        defaultValue: DataTypes.NOW,
+        allowNull: false,
+      },
+    });
+
+    await queryInterface.addIndex('CommentReactions', ['CommentId']);
+    await queryInterface.addIndex('CommentReactions', ['UserId', 'CommentId', 'emoji'], {
+      indexName: `CommentReactions_UserId_CommentId_emoji`,
+      indicesType: 'UNIQUE',
+    });
```
Betree, line 45: For now we don't need to fetch all comments for a given users, let's have just one focused on comments:

```suggestion
    await queryInterface.addIndex('CommentReactions', ['CommentId', 'FromCollectiveId', 'emoji'], {
      indexName: `CommentReactions_UserId_CommentId_emoji`,
      indicesType: 'UNIQUE',
    });
```

server/constants/emojiTypes.js
```diff
@@ -0,0 +1 @@
+export const EMOJI_TYPES = ['thumbsup', 'thumbsdown', 'smiley', 'tada', 'confused', 'heart', 'rocket', 'eyes'];
```
Betree, line 1: I'd be in favor of using the EMOJIs directly here (ie `EMOJI_TYPES = ['👍️', ...]`)

server/graphql/loaders/comment-reaction.ts
```diff
@@ -0,0 +1,17 @@
+import DataLoader from 'dataloader';
+import models from '../../models';
+import { Op } from 'sequelize';
+
+/**
+ * Loader for all collective's payout methods
+ */
+export default {
+  generateCommentReactionsLoader: (): DataLoader<number, object> =>
```
Betree, line 9: We could add this to `server/graphql/loaders/comments.ts` as `reactionsByCommentId`

server/graphql/loaders/comment-reaction.ts
```diff
@@ -0,0 +1,17 @@
...
+    new DataLoader(async (commentIds: number[]) => {
+      const commentReactions = await models.CommentReaction.findAll({
+        where: { CommentId: { [Op.in]: commentIds } },
+      });
+
+      return commentReactions;
```
Betree, line 15: We need to group the results by comment ID for dataloader to do its job here, check the `sortResults` helper

migrations/20200507120000-create-comment-reactions.js
```diff
@@ -0,0 +1,51 @@
...
+
+module.exports = {
+  up: async (queryInterface, DataTypes) => {
+    await queryInterface.createTable('CommentReactions', {
+      id: {
+        type: DataTypes.INTEGER,
+        primaryKey: true,
+        autoIncrement: true,
+      },
+      UserId: {
```
Betree, line 11: I just realized that I forgot to add a column to reference the `FromCollective` in the specs:

```suggestion
      FromCollectiveId: {
        type: DataTypes.INTEGER,
        references: { key: 'id', model: 'Users' },
        onDelete: 'CASCADE',
        onUpdate: 'CASCADE',
        allowNull: false,
      },
      UserId: {
```

server/graphql/v2/input/CommentReactionCreateInput.js
```diff
@@ -0,0 +1,32 @@
...
+  name: 'CommentReactionCreateInput',
+  fields: () => ({
+    emoji: {
+      type: GraphQLString,
+      description: 'The emoji associated with the reaction',
+    },
+    UserId: {
+      type: GraphQLString,
+      description: 'An unique identifier for the user id associated with this comment reaction',
+    },
```
Betree, line 18: No need for this one, we'll take the user id from `LoggedInUser.id`

```suggestion
```

server/graphql/v2/input/CommentReactionCreateInput.js
```diff
@@ -0,0 +1,32 @@
...
+    CommentId: {
+      type: GraphQLString,
+      description: 'An unique identifier for the comment id associated with this comment reaction',
+    },
```
Betree, line 22: This should be a reference, similar to what we do for `server/graphql/v2/input/ConnectedAccountReferenceInput.ts`

```suggestion
    comment: {
      type: CommentReferenceInput,
      description: 'An unique identifier for the comment id associated with this comment reaction',
    },
```

server/graphql/v2/input/CommentReactionCreateInput.js
```diff
@@ -0,0 +1,32 @@
...
+    expense: {
+      type: ExpenseReferenceInput,
+      description: 'If your comment is linked to an expense, set it here',
+    },
+    ExpenseId: {
+      type: GraphQLInt,
+      deprecationReason: '2019-02-26: Please use the expense field',
+    },
```
Betree, line 30: Not needed
```suggestion
```

server/graphql/v2/mutation/CommentReactionMutations.js
```diff
@@ -0,0 +1,56 @@
...
+import { canUseFeature } from '../../../lib/user-permissions';
+import FEATURE from '../../../constants/feature';
+import { CommentReactionCreateInput } from '../input/CommentReactionCreateInput';
+
+async function createCommentReaction(entity, args, req) {
+  if (!req.remoteUser) {
+    throw new Unauthorized();
+  } else if (!canUseFeature(req.remoteUser, FEATURE.CONVERSATIONS)) {
+    throw new FeatureNotAllowedForUser();
+  }
```
Betree, line 14:
```suggestion
  }
```

server/graphql/v2/mutation/CommentReactionMutations.js
```diff
@@ -0,0 +1,56 @@
...
+
+  return await models.CommentReaction.addCommentReaction(req.remoteUser, comment, emoji);
+}
+
+async function deleteCommentReaction(id, remoteUser) {
+  if (!remoteUser) {
+    throw new Unauthorized();
+  } else if (!canUseFeature(remoteUser, FEATURE.CONVERSATIONS)) {
+    throw new FeatureNotAllowedForUser();
+  }
```
Betree, line 26:
```suggestion
  }
```

server/graphql/v2/mutation/CommentReactionMutations.js
```diff
@@ -0,0 +1,56 @@
...
+  return await models.CommentReaction.addCommentReaction(id);
```
Betree, line 27: We need to make sure that user is allowed to delete the reaction, aka `LoggedInUser.isAdmin(reaction.FromCollectiveId)`

server/models/CommentReaction.js
```diff
@@ -0,0 +1,84 @@
...
+      },
+
+      UserId: {
+        type: DataTypes.INTEGER,
+        references: { key: 'id', model: 'Users' },
+        onDelete: 'NO ACTION',
+        onUpdate: 'CASCADE',
+        allowNull: false,
+      },
+
```
Betree, line 21: Let's add it here too

```suggestion
      FromCollectiveId: {
        type: DataTypes.INTEGER,
        references: { key: 'id', model: 'Users' },
        onDelete: 'CASCADE',
        onUpdate: 'CASCADE',
        allowNull: false,
      },
```

server/graphql/v2/object/CommentReaction.js
```diff
@@ -0,0 +1,38 @@
...
+        description: 'An unique identifier for this comment reaction',
+      },
+      UserId: {
+        type: GraphQLString,
+        description: 'An unique identifier for the user id associated with this comment reaction',
+      },
+      CommentId: {
+        type: GraphQLString,
+        description: 'An unique identifier for the comment id associated with this comment reaction',
+      },
```
Betree, line 20: We won't need that for now

```suggestion
```

server/graphql/v2/object/CommentReaction.js
```diff
@@ -0,0 +1,38 @@
...
+      emoji: {
+        type: GraphQLString,
+        description: 'The emoji associated with this user and comment',
+      },
+
```
Betree, line 25: We need to add `fromAccount` here the return the `fromCollective`

server/graphql/v2/mutation/CommentReactionMutations.js
```diff
@@ -0,0 +1,51 @@
...
+
+  removeCommentReaction: {
+    type: CommentReaction,
+    args: {
+      id: {
+        type: new GraphQLNonNull(GraphQLString),
+      },
+    },
+    resolve(_, { id }, { remoteUser }) {
+      return deleteCommentReaction(id, remoteUser);
```
Betree, line 46: According to the code below, this should be `removeReaction`. We also need to check the permissions before removing the reaction

SudharakaP: Thanks much for the review. I've added the code to check the permissions. Let me know if you see any further issues. :smiley:

server/graphql/v2/mutation/CommentReactionMutations.js
SudharakaP: @Betree : So I've added the condition such that the user who owns the comment reaction to be able to delete it. But the statement `LoggedInUser.isAdmin(reaction.FromCollectiveId)` implies that the admin of the collective should also be able to delete the comment reaction? Is this something we should allow; just wondering since for example GitHub doesn't allow that. Only the creator of the reaction can delete the reaction right? :thinking:

Betree: `reaction.FromCollectiveId` will reference user's collective (`LoggedInUser.collective.id`). So users will only be able to delete their own reactions.

Betree: We tend to rely on the `CollectiveId` to manage permissions. `UserId` are mostly used for logging/history purposes. The idea is that if you post a comment on behalf of your collective/organization, all admins of this profile should be able to edit/delete the comment.

SudharakaP: @Betree : Wonderful. Thanks much for the information. :smile:

SudharakaP:
I don't think we need to squash this anymore as I've interactively re-based this. So I have remove the label; let me know otherwise. :smile:

server/graphql/v2/object/Comment.js
```diff
@@ -1,5 +1,6 @@
-import { GraphQLObjectType, GraphQLString } from 'graphql';
+import { GraphQLList, GraphQLObjectType, GraphQLString } from 'graphql';
 import { GraphQLDateTime } from 'graphql-iso-date';
+import { GraphQLJSONObject } from 'graphql-type-json';
```
kewitz, line 3: For consistency, use `import GraphQLJSON from 'graphql-type-json';`.

SudharakaP: @kewitz : Done. :+1:

Betree:
> Not so sure about how this will scale when we decide to add reactions to other objects, are we duplicating these or moving towards polymorphic associations? If we plan to go towards polymorphic associations, I suggest we start it from the beginning.

The current code supports adding reactions on comments and conversations (because conversations are identified by a root comment). The part that is left for a future iteration is reactions on updates, which does not have specs yet. I don't have a strong experience with polymorphic associations in Postgres, last time I tried it was pretty tedious and I had a bad time setting up the tables and indexes. I just checked how they interface with sequelize, and it's not very straightforward: https://sequelize.org/master/manual/polymorphic-associations.html

The `CommentReactions` table introduced here is really about comment reactions. If we want to support updates too, the solution from the current work is to introduce a new `UpdateReactions` table and refactorize a bit the code to use the same validations library and have common GraphQL interfaces. Having the data stored in a different table is not a bad thing in itself, it's easier to query and index - though obviously, polymorphic associations can help to create something more generic.

All that to say, even if we decide to implement reactions on updates tomorrow, I don't think that we'll have a significant gain by refactoring the code to polymorphic associations. But I'm curious about them, so if you have any good examples or want to use them to build new features I'll be happy to learn more!

kewitz:
@Betree the last time I played with it, I had the feeling it was only worth it when dealing with a minimum amount of models, I don't think I would use it for 2 or 3 models.
Implementing is a bit annoying and I have no idea how it behaves when using raw SQL.
I just wanted to know what was the plan, and if we're not planning to scale reactions to a ton of models I think keeping it simple is a good idea.

SudharakaP:
> The part that is left for a future iteration is reactions on updates, which does not have specs yet.

@Betree : When the specs get cleared up for reactions on updates feel free to let me know. I am willing to do that as well.

PR #3878, opened 2020-05-14
Add Connected Account for e2e and skip Github verification for e2e

sbinlondon:
This adds a Github connected account for the `testuseradmin` for e2e test.

But because it would not be possible to spoof real Github account verification for the tests, it also adds a check to skip the verification is the `NODE_ENV=e2e`.


sbinlondon:
@Betree It seems my workaround to skip Github verification isn't working. I'd assumed we set the NODE_ENV=e2e for CI but it seems to be production.

What are your thoughts on how to get this test to pass?

Related PR on frontend: https://github.com/opencollective/opencollective-frontend/pull/4089

sbinlondon:
Frustratingly the frontend tests pass because we are skipping Github verification if we are in e2e tests or CI.... but here on the API side it causes some tests to fail

sbinlondon:
@kewitz and I decided to try Sinon stubs. For some reason I can't run it locally to test in the e2e environment.

Screenshot 2020-05-14 at 20 12 00

At least the graphql test isn't failing anymore. I feel kind of like I'm banging my head on a brick wall at this point so would love some input and will resume tomorrow.

sbinlondon:
Update: this solution probably won't work. We tried changing 'import' to 'require' but then it crashes when it gets to the imports in files further down the chain, like the Github library we load in the e2e mocks file.

I'm wondering if it's worth all of the hassle for this one test? The backend functionality we are trying to mock (authorizing something with Github during collective creation) is already mocked in some API tests. And on the frontend, we validate that the CCF form works for COVID/community collectives through another test file. So all the separate parts of 'creating a collective with github' are covered. Can we assume that this is good enough?

Betree:
If you used `"start:e2e:server": "node ./dist/e2e.js",`, that's probably the cause for `Can't use import...`, you need to use `babel-node`.

znarf:
Or, use require or compile it.

sbinlondon:
> Or, use require or compile it.

We tried this, the problem is that when it calls more files, it would ask us to use 'require' in those as well. You'd have to change everything to 'require'.

sbinlondon:
> If you used `"start:e2e:server": "node ./dist/e2e.js",`, that's probably the cause for `Can't use import...`, you need to use `babel-node`.

Unfortunately this does not work, once it gets to a .ts file it says it 'can't find it'

Screenshot 2020-05-15 at 11 41 43

Betree:
@sbinlondon Ah yes, I forgot this is API. The other commands in `pakcage.json` use `babel-node --extensions .js,.ts`, using that should work

sbinlondon:
Ok, this at least works now for the API.

Now I need to figure out why the Sinon `resolves()` is passing undefined.

Screenshot 2020-05-15 at 11 56 40

Have also tried `.resolves(true)` (the value we'd expect from these functions) but still getting an undefined error on the frontend.

znarf:
I just checkout the branch and I don't see the point of babel-node given that you remove the following lines. This folder or file is not existing in the branch.

```// Load stubs for E2E tests
  if (process.env.E2E_TEST || process.env.CI) {
    require('../test/mocks/e2e');
  }```

sbinlondon:
Closing. New PR can be seen here:  opencollective/opencollective-api#3890

PR #3880, opened 2020-05-14
Fix expense performances migration

Betree:


migrations/20200514141318-add-collective-index-on-activities.js
```diff
@@ -8,12 +8,15 @@ module.exports = {
...
 
     // Create Activity > ExpenseId, add an index and fill column
-    await queryInterface.addColumn('Activities', 'ExpenseId', {
-      type: Sequelize.INTEGER,
-      references: { key: 'id', model: 'Expenses' },
-      onDelete: 'CASCADE',
-      onUpdate: 'CASCADE',
-    });
+    const activitiesDefinition = await queryInterface.describeTable('Activities');
+    if (!activitiesDefinition.ExpenseId) {
```
Betree, line 12: Added this check because the migration was half-runned on staging

PR #3881, opened 2020-05-14
Add indexes to fix getIncognitoProfile() query perf

kewitz:
This is currently our most expensive query.

### DIFF
--- a/migrations/20200514164200-add-collective-index-for-incognito-query.js
+++ b/migrations/20200514164200-add-collective-index-for-incognito-query.js
@@ -1,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    // Enhance User.prototype.getIncognitoProfile() performance
+    await queryInterface.sequelize.query(`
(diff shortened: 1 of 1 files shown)


znarf:
@kewitz can you integrate feedback and ship that?

kewitz:
@znarf yes, will do first thing after our meeting.

migrations/20200514164200-add-collective-index-for-incognito-query.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    // Enhance User.prototype.getIncognitoProfile() performance
+    await queryInterface.sequelize.query(`
+      CREATE INDEX IF NOT EXISTS collectives_incognito_query ON "public"."Collectives"("CreatedByUserId");
```
znarf, line 7: I would use a more generic name for the index.

kewitz: Cool.
Renamed using the standard field name for the index, I'm keeping the comment as paper-trail.

kewitz:
Well, it looks like it is working.
image

PR #3884, opened 2020-05-15
Restrict unsupported TransferWise currencies

kewitz:
Due to a lack of support to dynamically generated fields in the BankInformationForm, I'm restricting a few more currencies since they break the frontend.

Related to https://sentry.io/share/issue/041da0ad4137499f8e62164cee8ac318/

### DIFF
--- a/server/paymentProviders/transferwise/index.ts
+++ b/server/paymentProviders/transferwise/index.ts
@@ -15,6 +15,14 @@ export const blackListedCurrencies = [
   /** Incomplete requiredFields API or MVP form support */
   'UYU',
   'KRW',
+  'GHS',
+  'HKD',
+  'ILS',
+  'INR',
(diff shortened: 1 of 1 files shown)


kewitz:
@Betree, I need to implement async validation, actually using TransferWise to validate the updated fields.
I've been delaying this to save the effort while waiting for validation. Should I implement it ASAP?

Betree:
Thanks for the details!

> Should I implement it ASAP?

I don't think it's pressing, the error message is clear enough. An issue however could be nice, maybe a bounty if you think it's accessible?

kewitz:
@Betree cool. I have this tracked in the TransferWise project as a card, but I agree it is kinda siloed.
I'll manage this better, I'll convert all those pending cards to proper issues where it is more public.
Thanks!

PR #3909, opened 2020-05-19
Host Budget management

znarf:
Implement: https://github.com/opencollective/opencollective/issues/3161

### DIFF
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -1111,7 +1111,8 @@ const CollectiveFields = () => {
         } else if (collective.type === types.EVENT) {
           const ParentCollectiveId = collective.ParentCollectiveId;
           const parentCollective = ParentCollectiveId && (await req.loaders.Collective.byId.load(ParentCollectiveId));
-          return parentCollective && parentCollective.isApproved();
+          // In the future, we should make it possible to directly read the approvedAt of the event
+          return parentCollective && (parentCollective.isHostAccount || parentCollective.isApproved());
         } else {
           return collective.isApproved();
         }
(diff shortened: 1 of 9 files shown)


server/models/Collective.js
```diff
@@ -1985,9 +1985,7 @@ export default function (Sequelize, DataTypes) {
 
   // edit the tiers of this collective (create/update/remove)
   Collective.prototype.editTiers = function (tiers) {
-    if (this.type !== types.COLLECTIVE && this.type !== types.EVENT) {
-      return [];
-    }
+    // All kind of accounts can have Tiers
```
znarf, line 1988: Check if it was actually required to prevent errors when Users edit their profile.

server/models/Transaction.js
```diff
@@ -528,5 +529,35 @@ export default (Sequelize, DataTypes) => {
...
+    const amount = order.totalAmount;
+    const platformFeePercent = get(order, 'data.platformFeePercent', 0);
+    const platformFee = calcFee(order.totalAmount, platformFeePercent);
+    const payload = {
+      type: 'CREDIT',
+      amount,
+      description: order.description,
+      currency: order.currency,
+      CollectiveId: order.CollectiveId,
+      FromCollectiveId: order.FromCollectiveId,
```
znarf, line 544: Can we null that?

server/graphql/v1/CollectiveInterface.js
```diff
@@ -1111,7 +1111,7 @@ const CollectiveFields = () => {
         } else if (collective.type === types.EVENT) {
           const ParentCollectiveId = collective.ParentCollectiveId;
           const parentCollective = ParentCollectiveId && (await req.loaders.Collective.byId.load(ParentCollectiveId));
-          return parentCollective && parentCollective.isApproved();
+          return parentCollective && (parentCollective.isHostAccount || parentCollective.isApproved());
```
znarf, line 1114: In the future, we should make it possible to directly read the approvedAt of the event

PR #3913, opened 2020-05-19
Connect PayPal for Payouts

kewitz:
Resolves opencollective/opencollective#3140

### DIFF
--- a/server/graphql/v2/mutation/ConnectedAccountMutations.ts
+++ b/server/graphql/v2/mutation/ConnectedAccountMutations.ts
@@ -2,6 +2,7 @@ import { GraphQLNonNull } from 'graphql';
 import { pick } from 'lodash';
 
 import { Service } from '../../../constants/connected_account';
+import * as paypal from '../../../lib/paypal';
 import * as transferwise from '../../../lib/transferwise';
 import models from '../../../models';
 import { Unauthorized, ValidationFailed } from '../../errors';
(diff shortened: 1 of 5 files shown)


server/graphql/v2/mutation/ConnectedAccountMutations.ts
```diff
@@ -37,20 +38,29 @@ const connectedAccountMutations = {
...
 
-      if (args.connectedAccount.service === Service.TRANSFERWISE) {
+      if ([Service.TRANSFERWISE, Service.PAYPAL].includes(args.connectedAccount.service)) {
         if (!args.connectedAccount.token) {
-          throw new ValidationFailed('A token is required for TransferWise accounts');
+          throw new ValidationFailed('A token is required');
         }
         const sameTokenCount = await models.ConnectedAccount.count({
-          where: { service: Service.TRANSFERWISE, token: args.connectedAccount.token },
+          where: { token: args.connectedAccount.token },
```
Betree, line 46: Not fan of this one, having the same token shouldn't conflict if we're not using the same service (even if the risk of having a collision is pretty low). Can we do something like:

```suggestion
          where: { token: args.connectedAccount.token, service: args.connectedAccount.service },
```

Betree, line 46: Not for this PR, but as a follow up we might want to add some rate limit on this endpoint to prevent bruteforce attacks on tokens, ie:

```
> createConnectedAccount(token: "aaaaaa")
Ok
> createConnectedAccount(token: "aaaaab")
Ok
> createConnectedAccount(token: "aaaaac")
This token is already being used ==> Found a token
```

server/lib/paypal.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const getPayPalClient = ({ token, clientId }: ConnectedAccount) => {
+  if (process.env.NODE_ENV === 'production') {
+    return new paypalPayoutsSDK.core.LiveEnvironment(clientId, token);
+  }
+  const environment =
+    process.env.NODE_ENV === 'production'
+      ? new paypalPayoutsSDK.core.LiveEnvironment(clientId, token)
+      : new paypalPayoutsSDK.core.SandboxEnvironment(clientId, token);
```
Betree, line 15: This check seems unnecessary given that we already check `if (process.env.NODE_ENV === 'production')` above

server/lib/paypal.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  return new paypalPayoutsSDK.core.PayPalHttpClient(environment);
+};
+
+export const validateConnectedAccount = async ({ token, clientId }: ConnectedAccount): Promise<void> => {
+  const paypal = getPayPalClient({ token, clientId });
+  console.log(paypal.execute);
+  const t = await paypal.fetchAccessToken();
+  console.log('token', t);
```
Betree, line 24:
```suggestion
  const paypal = getPayPalClient({ token, clientId });
  await paypal.fetchAccessToken();
```

kewitz: Wow, sorry about that.

server/lib/paypal.ts
kewitz: Oh, a wrong partial refactor!
I actually need to return the PayPalHttpClient, this would break in production.

PR #3938, opened 2020-05-26
Hyperwatch POC

znarf:
Expose HTTP structured logs for other internal tools to consume.

- Disabled by default
- Secured by HTTP Basic Auth
- Secret required, no default


server/routes.js
```diff
@@ -197,6 +201,20 @@ export default app => {
...
 
+  // Mount Hyperwatch API and Websocket
+  if (config.hyperwatch && config.hyperwatch.secret && parseToBoolean(config.hyperwatch.enabled) === true) {
+    // We need to setup express-ws here to make Hyperwatch's websocket works
+    expressWs(app);
+    const hyperwatchPath = '/_hyperwatch';
+    const hyperwatchBasicAuth = expressBasicAuth({
+      users: { [config.hyperwatch.username]: config.hyperwatch.secret },
+      challenge: true,
+      realm: config.hyperwatch.realm,
```
Betree, line 212: `config.hyperwatch.realm` seems to be missing in the config

znarf: Added as `null` to the default config.

PR #3942, opened 2020-05-26
Schedule Expense for Payment with PayPal Payouts

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3157

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -2144,8 +2144,9 @@ export default function (Sequelize, DataTypes) {
         CollectiveId: this.id,
         createdAt: { [Op.lt]: until },
       },
+      raw: true,
     });
-    const transactionSum = parseInt(transactions.toJSON().total, 10);
+    const transactionSum = parseInt(transactions.total, 10);
 
     // Subtract expenses scheduled for payment
     const expenses = await models.Expense.findOne({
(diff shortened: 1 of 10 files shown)


server/models/Collective.js
```diff
@@ -2135,17 +2135,31 @@ export default function (Sequelize, DataTypes) {
...
       attributes: [
         [Sequelize.fn('COALESCE', Sequelize.fn('SUM', Sequelize.col('netAmountInCollectiveCurrency')), 0), 'total'],
       ],
       where: {
         CollectiveId: this.id,
         createdAt: { [Op.lt]: until },
       },
-    }).then(result => Promise.resolve(parseInt(result.toJSON().total, 10)));
+    });
+    const transactionSum = parseInt(transactions.toJSON().total, 10);
```
Betree, line 2149: Another way to achieve that is to pass `raw: true` to the query, that way sequelize will return a simple object (rather than a model) and you'll be able to to `transactions.total` directly.

server/models/Collective.js
```diff
@@ -2135,17 +2135,31 @@ export default function (Sequelize, DataTypes) {
...
+
+    // Subtract expenses scheduled for payment
+    const expenses = await models.Expense.findOne({
+      attributes: [[Sequelize.fn('COALESCE', Sequelize.fn('SUM', Sequelize.col('amount')), 0), 'total']],
+      where: {
+        CollectiveId: this.id,
+        status: expenseStatus.SCHEDULED_FOR_PAYMENT,
```
Betree, line 2156: Should we also include `PROCESSING` in there?

If the collective has a $100 balance, we should not allow to have two $75 expenses processing at the same time.

kewitz: Yes, that's coming in the workers PR, already implemented.
In the mid-term, I also want to migrate TransferWise to this implementation, I'm currently creating transactions before finishing processing.

PR #3948, opened 2020-05-27
Cache host plan

znarf:
Fix https://github.com/opencollective/opencollective/issues/3169


kewitz:
Nice! Invalidation looks good, well done!

PR #3974, opened 2020-06-02
PayPal Payout workers

kewitz:
I was able to test this using PayPal's sandbox and it was a bit annoying, feel free to review the code and I'll make sure to set this up in staging once it is merged.

If you really want to test locally:
1. Enable payouts in your host updating settings to have `features.paypalPayouts = true`.
2. Go to your host settings -> sending money and use the Application from https://developer.paypal.com/developer/applications/
    - Use _PayPal OC Inc_ credentials from 1Password
3. Create an expense using PayPal email `paypal-buyer@opencollective.com` to any of the collectives hosted by the host you just connected
4. Approve and Schedule the expense for Payment
5. Manually run the workers
    1. `npm run script cron/hourly/10-pay-paypal-scheduled-expenses-payouts.js` should mark the expense as _processing_
    2. `npm run script cron/hourly/11-check-pending-paypal-expenses.js` should mark the expense as _paid_

### DIFF
--- a/cron/hourly/10-pay-paypal-scheduled-expenses-payouts.js
+++ b/cron/hourly/10-pay-paypal-scheduled-expenses-payouts.js
@@ -7,7 +7,7 @@ import { groupBy, values } from 'lodash';
 import status from '../../server/constants/expense_status';
 import models from '../../server/models';
 import { PayoutMethodTypes } from '../../server/models/PayoutMethod';
-import { payExpensesBatch } from '../../server/paymentProviders/paypal/payouts';
+import * as paypal from '../../server/paymentProviders/paypal/payouts';
 
 export async function run() {
   const expenses = await models.Expense.findAll({
@@ -19,9 +19,9 @@ export async function run() {
       { model: models.PayoutMethod, as: 'PayoutMethod', where: { type: PayoutMethodTypes.PAYPAL } },
     ],
   });
-  const batches = values(groupBy(expenses, 'CollectiveId'));
+  const batches = values(groupBy(expenses, 'collective.HostCollectiveId'));
   for (const batch of batches) {
-    await payExpensesBatch(batch).catch(console.error);
+    await paypal.payExpensesBatch(batch).catch(console.error);
   }
 }
 
--- a/cron/hourly/10-pay-paypal-scheduled-expenses-payouts.js
+++ b/cron/hourly/10-pay-paypal-scheduled-expenses-payouts.js
@@ -5,6 +5,7 @@ import '../../server/env';
 import { groupBy, values } from 'lodash';
 
 import status from '../../server/constants/expense_status';
+import logger from '../../server/lib/logger';
 import models from '../../server/models';
 import { PayoutMethodTypes } from '../../server/models/PayoutMethod';
 import * as paypal from '../../server/paymentProviders/paypal/payouts';
@@ -20,9 +21,12 @@ export async function run() {
     ],
   });
   const batches = values(groupBy(expenses, 'collective.HostCollectiveId'));
+  logger.info(`Processing ${expenses.length} expense(s) scheduled for payment using PayPal Payouts...`);
   for (const batch of batches) {
+    logger.info(`Paying host ${batch[0]?.collective?.HostCollectiveId} batch with ${batch.length} expense(s)...`);
     await paypal.payExpensesBatch(batch).catch(console.error);
   }
+  logger.info('Done!');
 }
 
 if (require.main === module) {
--- a/cron/hourly/10-pay-paypal-scheduled-expenses-payouts.js
+++ b/cron/hourly/10-pay-paypal-scheduled-expenses-payouts.js
@@ -1,0 +1,37 @@
+#!/usr/bin/env node
+
+import '../../server/env';
+
+import { groupBy, values } from 'lodash';
+
+import status from '../../server/constants/expense_status';
+import models from '../../server/models';
+import { PayoutMethodTypes } from '../../server/models/PayoutMethod';
+import { payExpensesBatch } from '../../server/paymentProviders/paypal/payouts';
+
+export async function run() {
+  const expenses = await models.Expense.findAll({
+    where: {
+      status: status.SCHEDULED_FOR_PAYMENT,
+    },
+    include: [
+      { model: models.Collective, as: 'collective' },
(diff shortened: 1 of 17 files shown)


kewitz:
Rebased.

server/models/Collective.js
```diff
@@ -2136,31 +2136,42 @@ export default function (Sequelize, DataTypes) {
...
-      where: {
-        CollectiveId: this.id,
-        createdAt: { [Op.lt]: until },
-      },
-      raw: true,
-    });
-    const transactionSum = parseInt(transactions.total, 10);
+    const result = await this.sequelize.query(
+      `
+        WITH "blockedFunds" AS (
```
Betree, line 2141: Could we have this query in a different file to use the exact same one between the loader and the model?

server/graphql/loaders/index.js
```diff
@@ -54,16 +54,40 @@ export const loaders = req => {
...
+          SELECT
+            e."CollectiveId", COALESCE(sum(e.amount), 0) as sum
+          FROM
+            "Expenses" e
+          WHERE
+            e."CollectiveId" IN (:ids)
+            AND (
+              e.status = 'SCHEDULED_FOR_PAYMENT'
+              OR (
+                e.status = 'PROCESSING' AND e.data ->> 'payout_batch_id' IS NOT NULL
```
Betree, line 70: As per https://github.com/opencollective/opencollective-api/pull/3942#discussion_r433928267, I assume the reason for filtering on `payout_batch_id` is because for Transferwise, processing expenses already have transactions right?

server/lib/paypal.ts
```diff
@@ -1,20 +1,65 @@
-import paypalPayoutsSDK from '@paypal/payouts-sdk';
+import paypal from '@paypal/payouts-sdk';
+
+import { PayoutBatchDetails, PayoutRequestBody, PayoutRequestResult } from '../types/paypal';
+
+const parseError = e => {
+  try {
+    return JSON.parse(e.message).message;
```
Betree, line 7: Looks strange, does PayPal returns an error like `{ message: { message: '___' } }`?

server/paymentProviders/paypal/payouts.ts
```diff
@@ -0,0 +1,125 @@
...
+    throw new Error('All expenses in the batch should belong to the same Collective.');
+  }
+
+  const host = await firstExpense.collective.getHostCollective();
+  if (!host) {
+    throw new Error(`Could not find the host embursing the expense.`);
+  }
+
+  const [connectedAccount] = await host.getConnectedAccounts({
+    where: { service: 'paypal', deletedAt: null },
```
Betree, line 24: Let's be more specific, we also have some entries for `service=paypal type=payment` in the DB used when people pay with PayPal.

```suggestion
    where: { service: 'paypal', type: 'adaptive', deletedAt: null },
```

server/paymentProviders/paypal/payouts.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export const checkBatchStatus = async (batch: any[]): Promise<any[]> => {
+  const [firstExpense] = batch;
+  const host = await firstExpense.collective.getHostCollective();
+  if (!host) {
+    throw new Error(`Could not find the host embursing the expense.`);
+  }
+
+  const [connectedAccount] = await host.getConnectedAccounts({
+    where: { service: 'paypal', deletedAt: null },
```
Betree, line 67: same here

```suggestion
    where: { service: 'paypal', type: 'adaptive', deletedAt: null },
```

server/paymentProviders/paypal/payouts.ts
```diff
@@ -0,0 +1,125 @@
...
+        case 'REVERSED':
+          await expense.setError(expense.lastEditedById);
+          await expense.createActivity(activities.COLLECTIVE_EXPENSE_ERROR);
+          break;
+        // Ignore cases
+        case 'ONHOLD':
+        case 'UNCLAIMED': // Link sent to a non-paypal user, waiting for being claimed.
+        case 'PENDING':
+        default:
+          console.warn(`Expense is still being processed, nothing to do but wait.`);
```
Betree, line 112: Why `warn`ing about this? Is this something we should actively watch?

cron/hourly/10-pay-paypal-scheduled-expenses-payouts.js
```diff
@@ -0,0 +1,37 @@
...
+  const expenses = await models.Expense.findAll({
+    where: {
+      status: status.SCHEDULED_FOR_PAYMENT,
+    },
+    include: [
+      { model: models.Collective, as: 'collective' },
+      { model: models.PayoutMethod, as: 'PayoutMethod', where: { type: PayoutMethodTypes.PAYPAL } },
+    ],
+  });
+  const batches = values(groupBy(expenses, 'CollectiveId'));
```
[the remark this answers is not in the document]
kewitz: Yes, that's the optimum, good catch.

server/models/Collective.js
kewitz: We can, but since `lib/queries` require models we have a cyclical dependency to refactor.
I'll try to do this right in this PR and see if we can DRY this query.

server/graphql/loaders/index.js
kewitz: Yes, that's exactly right.
We have TransferWise expenses that are processing and are already accounted for in a transaction.

server/lib/paypal.ts
kewitz: The problem here is that PayPal returns the body of the error as a JSON object and the library won't parse that.

server/paymentProviders/paypal/payouts.ts
kewitz: @Betree correct me if I'm wrong, but aren't those saved as Payment Methods?

server/paymentProviders/paypal/payouts.ts
kewitz: Not really, I'll reduce the level of importance here.

kewitz:
Rebased.

server/paymentProviders/paypal/payouts.ts
Betree: Yes, my bad! :slightly_smiling_face:

PR #3983, opened 2020-06-03
Implement monthly host fees cron

znarf:
for collectives who opted in for budget.

Part of https://github.com/opencollective/opencollective/issues/3161

Implements: https://github.com/opencollective/opencollective/issues/1723

### DIFF
--- a/cron/monthly/host-fees.js
+++ b/cron/monthly/host-fees.js
@@ -1,0 +1,109 @@
+#!/usr/bin/env node
+import '../../server/env';
+
+import { v4 as uuid } from 'uuid';
+
+import { sumTransactions } from '../../server/lib/hostlib';
+import models, { Op } from '../../server/models';
+
+// Only run on the first of the month
+const today = new Date();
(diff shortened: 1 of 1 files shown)


cron/monthly/host-fees.js
```diff
@@ -0,0 +1,99 @@
...
+
+const dateRange = {
+  createdAt: { [Op.gte]: startDate, [Op.lt]: endDate },
+};
+
+async function main() {
+  const hostsWhere = {
+    isHostAccount: true,
+    isActive: true,
+    HostCollectiveId: { [Op.ne]: null },
```
kewitz, line 37: I'm lacking context here, what is a host that has `HostCollectiveId`?

znarf: We now have this special case of hosts managing their budget. https://github.com/opencollective/opencollective/issues/3161 They work almost like Collectives in that regard. For compatibility reason and to limit the complexity of this feature, they're "hosting themselves". `HostCollectiveId = id`

PR #3984, opened 2020-06-04
enchancement: create contributor member after expense is paid

bolariin:
Resolves https://github.com/opencollective/opencollective/issues/3184

### DIFF
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -285,14 +285,6 @@ export async function createExpense(remoteUser, expenseData) {
     return createdExpense;
   });
 
-  collective.addUserWithRole(remoteUser, roles.CONTRIBUTOR).catch(e => {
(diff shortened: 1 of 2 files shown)


server/graphql/v1/mutations/expenses.js
```diff
@@ -487,6 +479,17 @@ export async function deleteExpense(req, expenseId) {
 async function markExpenseAsPaid(expense, remoteUser) {
   debug('update expense status to PAID', expense.id);
   await expense.setPaid(remoteUser.id);
+
+  // Add user as CONTRIBUTOR
+  const contributor = await models.User.findOne({ where: { CollectiveId: expense.FromCollectiveId } });
```
Betree, line 484: We now support creating expenses as organizations, which means `FromCollectiveId` is sometimes referring to a profile that is not directly linked to a user.

I'd recommend using `expense.UserId` (you can fetch it with `expense.getUser()`).

PR #3993, opened 2020-06-04
PayPal Payouts Webhook

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3218
Based on #3974.

### DIFF
--- a/server/controllers/webhooks.ts
+++ b/server/controllers/webhooks.ts
@@ -1,6 +1,7 @@
-import { NextFunction,Request, Response } from 'express';
+import { NextFunction, Request, Response } from 'express';
 
 import paymentProviders from '../paymentProviders';
+import paypalWebhookHandler from '../paymentProviders/paypal/webhook';
 import transferwiseWebhookHandler from '../paymentProviders/transferwise/webhook';
 
 export async function stripeWebhook(req: Request, res: Response, next: NextFunction): Promise<void> {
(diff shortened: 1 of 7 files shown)


server/controllers/webhooks.ts
```diff
@@ -19,3 +20,12 @@ export async function transferwiseWebhook(
     .then(() => res.sendStatus(200))
     .catch(next);
 }
+
+export async function paypalWebhook(req: Request, res: Response, next: NextFunction): Promise<void> {
+  try {
+    await paypalWebhookHandler(req);
```
Betree, line 26: How do we make sure that this request indeed comes from PayPal, and not from a malicious user playing with the endpoint?

kewitz: That's a very good point.
PayPal has an API endpoint to validate events but that would require us to call PayPal on each event.
My proposal is to back validate the batch_id against the id we have saved in `expense.data`, since this information is secret. (See https://github.com/opencollective/opencollective-api/pull/3993/files#diff-1452b76bdf36ae9a41f0b1ff50047db9R61-R63)
Do you think this is enough or should we properly validate events against PayPal?

Betree: If this `batch_id` is a random ID that you can't predict (ie. `uuid`) and if we're not leaking it elsewhere, then I think it's a good first step that we can merge as a first version.

We can keep the validation of the event with PayPal's API for another PR if it's complex.

kewitz: It is random but I implemented the validation anyway in b467ab4.

PR #4001, opened 2020-06-06
feat: include financial contributor contact email in notification

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/2297

#### Short description of what this resolves:
Now new backer notification will display contact email of a financial contributor if available.

#### Changes proposed in this pull request:

If email is available display it in about section.

# Screenshot

notification

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1473,12 +1473,14 @@ export default function (Sequelize, DataTypes) {
 
     const urlPath = await this.getUrlPath();
     const memberCollective = await models.Collective.findByPk(member.MemberCollectiveId, sequelizeParams);
+    const memberCollectiveUser = await models.User.findByPk(member.CreatedByUserId, sequelizeParams);
 
     const data = {
       collective: { ...this.minimal, urlPath },
       member: {
         ...member.info,
         memberCollective: memberCollective.activity,
+        memberCollectiveUser: memberCollectiveUser.info,
       },
       order: order && {
         ...order.activity,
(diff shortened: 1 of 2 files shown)


snitin315:
@Betree any update on this one?

server/models/Collective.js
```diff
@@ -1472,12 +1472,14 @@ export default function (Sequelize, DataTypes) {
 
     const urlPath = await this.getUrlPath();
     const memberCollective = await models.Collective.findByPk(member.MemberCollectiveId, sequelizeParams);
+    const memberCollectiveUser = await models.User.findByPk(member.CreatedByUserId, sequelizeParams);
```
znarf, line 1475: `CreatedByUserId` fields should not be used for such thing, we should query `User` with the `CollectiveId` instead.

Betree:
As per the latest discussions in https://github.com/opencollective/opencollective/issues/2297, we should only include the email if collective.type is `USER`.

server/models/Collective.js
snitin315: like this `await models.User.findByPk(member.CollectiveId, sequelizeParams)` ?

Betree: `member.CollectiveId` is a collective ID (not a user ID). You can try something like this instead:

```es6
let memberCollectiveUser;
if (memberCollective.type === types.USER && !memberCollective.isIncognito) {
  memberCollectiveUser = await models.User.findOne({ where: { CollectiveId: memberCollective.id } });
}
```

snitin315: Thanks

PR #4003, opened 2020-06-06
refactor: migrate collective long description to html

snitin315:
Refers https://github.com/opencollective/opencollective/issues/3207

Related Frontend PR https://github.com/opencollective/opencollective-frontend/pull/4357

Created a migration that takes all collectives with a markdown description and converts to HTML

### DIFF
--- a/migrations/20200606105847-convert-collective-long-description-to-html.js
+++ b/migrations/20200606105847-convert-collective-long-description-to-html.js
@@ -1,6 +1,7 @@
 'use strict';
 
 import showdown from 'showdown';
+import { buildSanitizerOptions, sanitizeHTML } from '../server/lib/sanitize-html';
 
 const converter = new showdown.Converter();
 
@@ -15,7 +16,19 @@ module.exports = {
       { type: Sequelize.QueryTypes.SELECT },
     );
 
+    const sanitizeOptions = buildSanitizerOptions({
+      titles: true,
+      basicTextFormatting: true,
+      multilineTextFormatting: true,
+      links: true,
+      images: true,
+      videoIframes: true,
+      tables: true,
+    });
+
     for (const collective of collectives) {
+      const htmlContent = converter.makeHtml(collective.longDescription);
+      const sanitizedContent = sanitizeHTML(htmlContent, sanitizeOptions);
       await queryInterface.sequelize.query(
         `
         UPDATE "Collectives" c
@@ -24,7 +37,7 @@ module.exports = {
       `,
         {
           replacements: {
-            longDescription: converter.makeHtml(collective.longDescription),
+            longDescription: sanitizedContent,
             id: collective.id,
           },
         },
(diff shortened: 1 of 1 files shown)


migrations/20200606105847-convert-collective-long-description-to-html.js
```diff
@@ -0,0 +1,60 @@
...
+        WHERE c.id = :id
+      `,
+        {
+          replacements: {
+            longDescription: converter.makeMarkdown(collective.longDescription),
+            id: collective.id,
+          },
+        },
+      ),
+    );
```
Betree, line 58: The purpose of `down` is to rollback the changes if something wrong happens during the migration. What that would mean here is that we would need something to convert from HTML to markdown. We don't have that, and we probably don't want to introduce more complexity just to solve it. So doing this is fine:

```suggestion
    // Can't rollback this one
```

migrations/20200606105847-convert-collective-long-description-to-html.js
```diff
@@ -0,0 +1,60 @@
...
+import Promise from 'bluebird';
+import showdown from 'showdown';
+
+const converter = new showdown.Converter();
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    const collectives = await queryInterface.sequelize.query(
+      `
+        SELECT * FROM "Collectives"
```
Betree, line 12: In production, this query matches 6000+ collectives. Selecting the fields that we want will save some precious RAM:

```suggestion
        SELECT "id", "longDescription" FROM "Collectives"
```

migrations/20200606105847-convert-collective-long-description-to-html.js
```diff
@@ -0,0 +1,60 @@
...
+        WHERE LENGTH("longDescription") > 0 AND "longDescription" NOT LIKE '<%';
+      `,
```
Betree, line 14: No need to include deleted collectives

```suggestion
        WHERE LENGTH("longDescription") > 0 AND "longDescription" NOT LIKE '<%'
        AND "deletedAt" IS NULL;
      `,
```

migrations/20200606105847-convert-collective-long-description-to-html.js
```diff
@@ -0,0 +1,60 @@
...
+        WHERE c.id = :id
+      `,
+        {
+          replacements: {
+            longDescription: converter.makeHtml(collective.longDescription),
+            id: collective.id,
+          },
+        },
+      ),
+    );
```
Betree, line 32: As per the comment above, we have ~6000 collectives concerned by this migration. So this code may start a huge number of parallel DB queries. Could we rather have a `for` loop with an `await` inside (or similar) to make sure we do one query at a time?

Also if you have some insights on the time it takes to run locally, that'd be interesting for us to have.

snitin315:
@Betree Updated :100:, Please take a look.

Betree:
The migration runs ok on staging (~15s). I added a sanitize pass to make sure the result we generate matches what we have with the other editor.

PR #4004, opened 2020-06-07
Order tier/amount updates

sbinlondon:
Adds capability to Order mutations v2 for changing the tier or amount of an order


server/graphql/v2/input/TierReferenceInput.js
```diff
@@ -0,0 +1,19 @@
...
+      description: 'The legacy id assigned to the Tier',
+    },
+    flexible: {
+      type: GraphQLBoolean,
+      description: "Boolen indicated whether a tier's amount is FLEXIBLE (true) or FIXED (false)",
+    },
+    minimumAmount: {
+      type: GraphQLInt,
+      description: 'The minimum amount in cents of the Tier',
+    },
```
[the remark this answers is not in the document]
sbinlondon: Oh, interesting! I thought it was just the new pattern that we were using with v2 to 'normalise' inputs no matter what. That makes more sense.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -177,6 +186,35 @@ const orderMutations = {
...
       }
 
+      // amount
+      if (amount !== order.totalAmount) {
+        if (amount < 100) {
+          throw new Error('Invalid amount.');
+        }
+
+        // If using a tier, amount can never be less than the minimum amount, or else $1.00
+        if (amount < tier.minimumAmount) {
```
znarf, line 196: Have you copied something existing? What if `tier.minimumAmount` is null ?

znarf: If we change tiers, it this the amount of the new tier or the old tier?

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -177,6 +186,35 @@ const orderMutations = {
         order = await order.update({ PaymentMethodId: newPaymentMethod.id });
       }
 
+      // amount
+      if (amount !== order.totalAmount) {
+        if (amount < 100) {
+          throw new Error('Invalid amount.');
+        }
```
znarf, line 193: Edge case to consider: what's happening if "fees on top" are attached to the order?

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -177,6 +186,35 @@ const orderMutations = {
...
+
+      // tier
+      // check if the tier is different from the previous tier
+      if (tier.legacyId !== order.TierId) {
+        // check if the tier is a named tier
+        if (tier.legacyId !== null) {
+          const tierInfo = await models.Tier.findByPk(tier.legacyId);
+
+          if (!tierInfo) {
+            throw new Error(`No tier found with tier id: ${tier.legacyId} for collective ${order.FromCollectiveId}`);
```
znarf, line 211: Doesn't seems right to report `FromCollectiveId` it's `CollectiveId` that we are interested in I think.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -132,6 +133,14 @@ const orderMutations = {
         type: PaymentMethodReferenceInput,
         description: 'Reference to a Payment Method to update the order with',
       },
+      tier: {
+        type: TierReferenceInput,
+        description: 'Reference to a Tier to update the order with',
+      },
+      amount: {
+        type: GraphQLInt,
+        description: 'Reference to an Amount to update the order to',
```
znarf, line 142: It's not a Reference, it's an amount.

I think for better API design, we need to replace `GraphQLInt` with a new `AmountInput` type.

server/graphql/v2/input/TierReferenceInput.js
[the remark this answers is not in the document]
znarf: Yes, `Reference` object are only there to help point to an object that we want to retrieve and work with. Usually, only identifiers are useful in that context.

znarf:
Generally, if a `tier` is passed as argument, try to fetch it early from the database and make sure it's existing and valid. That will clarify the code later on.

server/graphql/v2/mutation/OrderMutations.js
sbinlondon: It's the amount of the tier we are trying to update to.

`tier.minimumAmount` in this instance will either be the actual minimum amount for the tier, if there is one, or else it is set to `100` (or $1.00) since that's the minimum we allow anyways

PR #4021, opened 2020-06-09
Add PaymentMethod to Account interface

sbinlondon:
Adds Payment Method to Account interface, and allows us to query payment methods on v2 Accounts on the frontend. This also means we can get rid of the legacyId in the v2 Payment Method object and use the encoded ids.

### DIFF
--- a/server/graphql/v2/input/PaymentMethodReferenceInput.js
+++ b/server/graphql/v2/input/PaymentMethodReferenceInput.js
@@ -9,7 +9,7 @@ export const PaymentMethodReferenceInput = new GraphQLInputObjectType({
   fields: () => ({
     id: {
       type: GraphQLString,
-      description: 'The encrypted id assigned to the payment method',
+      description: 'The id assigned to the payment method',
     },
   }),
 });
@@ -22,17 +22,10 @@ export const PaymentMethodReferenceInput = new GraphQLInputObjectType({
  *    - dbTransaction: An SQL transaction to run the query. Will skip `loaders`
  *    - lock: If true and `dbTransaction` is set, the row will be locked
  */
-export const fetchPaymentMethodWithReference = async (
-  input,
-  { loaders = null, throwIfMissing = false, dbTransaction = undefined, lock = false } = {},
-) => {
+export const fetchPaymentMethodWithReference = async (input, { throwIfMissing = false } = {}) => {
   // Load payment by ID using GQL loaders if we're not using a transaction & loaders are available
   const loadPaymentById = id => {
-    if (!loaders || dbTransaction) {
-      return models.PaymentMethod.findByPk(id, { transaction: dbTransaction, lock });
-    } else {
-      return loaders.PaymentMethod.byId.load(id);
-    }
+    return models.PaymentMethod.findByPk(id);
   };
 
   let paymentMethod;
--- a/server/graphql/v2/input/PaymentMethodReferenceInput.js
+++ b/server/graphql/v2/input/PaymentMethodReferenceInput.js
@@ -1,11 +1,11 @@
-import { GraphQLInputObjectType, GraphQLInt } from 'graphql';
+import { GraphQLInputObjectType, GraphQLString } from 'graphql';
 
 export const PaymentMethodReferenceInput = new GraphQLInputObjectType({
   name: 'PaymentMethodReferenceInput',
   fields: () => ({
-    legacyId: {
-      type: GraphQLInt,
-      description: 'The legacy id assigned to the payment method',
+    id: {
+      type: GraphQLString,
+      description: 'The encrypted id assigned to the payment method',
     },
   }),
 });
--- a/server/graphql/v2/input/PaymentMethodReferenceInput.js
+++ b/server/graphql/v2/input/PaymentMethodReferenceInput.js
@@ -1,5 +1,9 @@
 import { GraphQLInputObjectType, GraphQLString } from 'graphql';
 
+import models from '../../../models';
+import { NotFound } from '../../errors';
+import { idDecode } from '../identifiers';
+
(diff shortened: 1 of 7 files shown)


server/graphql/v2/object/PaymentMethod.js
```diff
@@ -26,6 +29,66 @@ export const PaymentMethod = new GraphQLObjectType({
...
+          if (paymentMethod.type === 'virtualcard') {
+            if (!req.remoteUser || !req.remoteUser.isAdmin(paymentMethod.CollectiveId)) {
+              return null;
+            }
+            return pick(paymentMethod.data, ['email']);
+          }
+
+          const data = paymentMethod.data;
+          // white list fields to send back; removes fields like CustomerIdForHost
+          const dataSubset = {
```
znarf, line 81: `pick` ?

server/graphql/v2/object/PaymentMethod.js
```diff
@@ -26,6 +29,66 @@ export const PaymentMethod = new GraphQLObjectType({
...
+        resolve(paymentMethod) {
+          return paymentMethod.service;
+        },
+      },
+      type: {
+        type: GraphQLString,
+        resolve(paymentMethod) {
+          return paymentMethod.type;
+        },
+      },
```
znarf, line 43: I would like to change that. We discussed about these `type` and `service` being not great to expose. I will propose something later.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -161,14 +161,16 @@ const orderMutations = {
         throw new Error('Order must be active to be updated');
       }
 
+      const decodedPMId = getDecodedId(args.paymentMethod.id);
```
znarf, line 164: Could be interesting to abstract that, similar to what we do with account `fetchAccountWithReference`. Ie: `fetchPaymentWithReference`.

server/graphql/v2/input/PaymentMethodReferenceInput.js
```diff
@@ -1,11 +1,11 @@
...
 
 export const PaymentMethodReferenceInput = new GraphQLInputObjectType({
   name: 'PaymentMethodReferenceInput',
   fields: () => ({
-    legacyId: {
-      type: GraphQLInt,
-      description: 'The legacy id assigned to the payment method',
+    id: {
+      type: GraphQLString,
+      description: 'The encrypted id assigned to the payment method',
```
znarf, line 8: We should not use the word `encrypted` it's "just" the id.

server/graphql/v2/object/PaymentMethod.js
```diff
@@ -26,6 +29,66 @@ export const PaymentMethod = new GraphQLObjectType({
...
+      balance: {
```
znarf, line 44: Should be an `Amount` type.

server/graphql/v2/object/PaymentMethod.js
```diff
@@ -26,6 +29,66 @@ export const PaymentMethod = new GraphQLObjectType({
...
+        type: GraphQLInt,
+        description: 'Returns the balance in the currency of this paymentMethod',
+        async resolve(paymentMethod, args, req) {
+          const balance = await paymentMethod.getBalanceForUser(req.remoteUser);
+          return balance.amount;
+        },
+      },
+      currency: {
```
znarf, line 52: Should be a `Currency` type

server/graphql/v2/mutation/OrderMutations.js
sbinlondon: Happy to work on that perhaps in a separate PR starting tomorrow?

server/graphql/v2/object/PaymentMethod.js
sbinlondon: Currently the `balance` is an integer, and it's used externally in some payment method libs that rely on it being an int.

Making it an `Amount` type means it would be an object with subfields of `value` and `currency`. Should I go ahead and make this change and then also update the lib?

server/graphql/v2/object/PaymentMethod.js
sbinlondon: There isn't a `Currency` type right now. If we changed the `balance` to an `Amount` type then there would be a `currency` subfield on that.

znarf: We do have an Enum type for currency.

But maybe this field is not actually needed, especially if we're moving the balance to an `Amount` type.

server/graphql/v2/object/PaymentMethod.js
znarf: Could you point me on the lib and function that would need an update?

server/graphql/v2/input/PaymentMethodReferenceInput.js
znarf: We should not use the word `encrypted`.

server/graphql/v2/input/PaymentMethodReferenceInput.js
```diff
@@ -1,11 +1,49 @@
...
+ *    - lock: If true and `dbTransaction` is set, the row will be locked
+ */
+export const fetchPaymentMethodWithReference = async (
+  input,
+  { loaders = null, throwIfMissing = false, dbTransaction = undefined, lock = false } = {},
+) => {
+  // Load payment by ID using GQL loaders if we're not using a transaction & loaders are available
+  const loadPaymentById = id => {
+    if (!loaders || dbTransaction) {
+      return models.PaymentMethod.findByPk(id, { transaction: dbTransaction, lock });
```
znarf, line 32: We can remove dbTransaction and lock if we're not using it for now.

PR #4022, opened 2020-06-09
Use a graphql loader for userTaxFormRequiredBeforePayment & requiredLegalDocuments

flickz:
Resolves https://github.com/opencollective/opencollective/issues/3123

This PR, adds `userTaxFormRequiredBeforePayment` and `requiredLegalDocuments` to expense loader, along with respectives unit tests.

### DIFF
--- a/server/graphql/loaders/expenses.ts
+++ b/server/graphql/loaders/expenses.ts
@@ -96,7 +96,7 @@ export const userTaxFormRequiredBeforePayment = (req): DataLoader<number, boolea
       LEFT JOIN "LegalDocuments" ld ON ld."CollectiveId" = e."FromCollectiveId"
                                         AND ld.year = date_part('year', e."incurredAt")
                                         AND ld."documentType" = 'US_TAX_FORM'
-      WHERE e.status IN ('PENDING', 'APPROVED', 'PAID', 'PROCESSING')
+      WHERE e.status IN ('PENDING', 'APPROVED', 'PAID', 'PROCESSING', 'SCHEDULED_FOR_PAYMENT')
       AND e."UserId" IN (SELECT "UserId" FROM "Expenses" WHERE "Expenses".id IN (:expenseIds))
       AND e.type NOT IN ('RECEIPT')
       AND "incurredAt" BETWEEN e."incurredAt" AND (e."incurredAt" + interval '1 year')
--- a/server/graphql/loaders/expenses.ts
+++ b/server/graphql/loaders/expenses.ts
@@ -17,24 +17,25 @@ const {
 const userTaxFormRequiredBeforePaymentQuery = `
   SELECT 
     e."UserId" "userId", 
-    MAX(e.id) as "expenseId", 
+    MAX(er.id) as "expenseId",
     MAX(ld."requestStatus") as "legalDocRequestStatus",
     MAX(d."documentType") as "requiredDocument",
-    SUM (amount) AS total
+    SUM (e."amount") AS total
   FROM "Expenses" e
+  INNER JOIN "Expenses" er ON er.id IN (:expenseIds)
   INNER JOIN "Collectives" c ON c.id = e."CollectiveId"
   LEFT JOIN "RequiredLegalDocuments" d ON d."HostCollectiveId" = c."HostCollectiveId"
-                                      AND d."documentType" = 'US_TAX_FORM'
+                                    AND d."documentType" = 'US_TAX_FORM'
   LEFT JOIN "LegalDocuments" ld ON ld."CollectiveId" = e."FromCollectiveId"
                                 AND ld.year = date_part('year', e."incurredAt")
                                 AND ld."documentType" = 'US_TAX_FORM'
-  WHERE e.status IN ('PENDING', 'APPROVED', 'PAID', 'PROCESSING', 'SCHEDULED_FOR_PAYMENT')
+  WHERE e."UserId" = er."UserId"
+  AND e.status IN ('PENDING', 'APPROVED', 'PAID', 'PROCESSING', 'SCHEDULED_FOR_PAYMENT')
   AND e.type NOT IN ('RECEIPT')
   AND e."deletedAt" IS NULL
-  AND "incurredAt" BETWEEN date_trunc('year', e."incurredAt") AND (date_trunc('year', e."incurredAt") + interval '1 year')
+  AND e."incurredAt" BETWEEN date_trunc('year', e."incurredAt") AND (date_trunc('year', e."incurredAt") + interval '1 year')
   GROUP BY e."UserId"
-  HAVING MAX(e.id) IN (:expenseIds)
-`
+`;
 
 
 /**
@@ -115,11 +116,9 @@ export const userTaxFormRequiredBeforePayment = (): DataLoader<number, boolean> 
        replacements: { expenseIds }
     });
     const expenseNeedsTaxForm = {}
-
     expenses.forEach((expense) => {
       expenseNeedsTaxForm[expense.expenseId] = expense.requiredDocument && expense.total >= THRESHOLD && expense.legalDocRequestStatus !== RECEIVED;
     })
-
     return expenseIds.map(id => expenseNeedsTaxForm[id] || false);
   });
 }
--- a/server/graphql/loaders/expenses.ts
+++ b/server/graphql/loaders/expenses.ts
@@ -24,16 +24,16 @@ const userTaxFormRequiredBeforePaymentQuery = `
   FROM "Expenses" e
   INNER JOIN "Collectives" c ON c.id = e."CollectiveId"
   LEFT JOIN "RequiredLegalDocuments" d ON d."HostCollectiveId" = c."HostCollectiveId"
-                                        AND d."documentType" = 'US_TAX_FORM'
+                                      AND d."documentType" = 'US_TAX_FORM'
   LEFT JOIN "LegalDocuments" ld ON ld."CollectiveId" = e."FromCollectiveId"
-                                  AND ld.year = date_part('year', e."incurredAt")
-                                  AND ld."documentType" = 'US_TAX_FORM'
+                                AND ld.year = date_part('year', e."incurredAt")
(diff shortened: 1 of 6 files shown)


server/graphql/loaders/expenses.ts
```diff
@@ -72,3 +80,79 @@ export const attachedFiles = (): DataLoader<number, ExpenseAttachedFile[]> => {
...
+      SELECT e.id "expenseId", e."UserId" "userId", ld."requestStatus" "legalDocRequestStatus", d."documentType" "requiredDocument",
+        SUM (amount) AS total
+      FROM "Expenses" e
+      INNER JOIN "Collectives" c ON c.id = e."CollectiveId"
+      LEFT JOIN "RequiredLegalDocuments" d ON d."HostCollectiveId" = c."HostCollectiveId"
+                                              AND d."documentType" = 'US_TAX_FORM'
+      LEFT JOIN "LegalDocuments" ld ON ld."CollectiveId" = e."FromCollectiveId"
+                                        AND ld.year = date_part('year', e."incurredAt")
+                                        AND ld."documentType" = 'US_TAX_FORM'
+      WHERE e.status IN ('PENDING', 'APPROVED', 'PAID', 'PROCESSING')
```
Betree, line 99: We just introduced this new status 

```suggestion
      WHERE e.status IN ('PENDING', 'APPROVED', 'PAID', 'PROCESSING', 'SCHEDULED_FOR_PAYMENT')
```

server/graphql/loaders/expenses.ts
```diff
@@ -72,3 +80,79 @@ export const attachedFiles = (): DataLoader<number, ExpenseAttachedFile[]> => {
...
+      AND e."UserId" IN (SELECT "UserId" FROM "Expenses" WHERE "Expenses".id IN (:expenseIds))
+      AND e.type NOT IN ('RECEIPT')
+      AND "incurredAt" BETWEEN e."incurredAt" AND (e."incurredAt" + interval '1 year')
+      GROUP BY e.id, e."UserId", d."documentType", ld."requestStatus"
```
Betree, line 103: If we group by expense ID (`e.id`) is it really summing the values of all expenses submitted by this user?

server/graphql/loaders/expenses.ts
```diff
@@ -72,3 +80,79 @@ export const attachedFiles = (): DataLoader<number, ExpenseAttachedFile[]> => {
...
+    `, {
+       type: sequelize.QueryTypes.SELECT,
+       model: models.Expense,
+       mapToModel: true,
```
Betree, line 107: Why do we need to map to model? Wouldn't it be simple to have `raw: true` so that we can access values directly without having to go through `dataValues`?

server/graphql/loaders/expenses.ts
```diff
@@ -72,3 +80,79 @@ export const attachedFiles = (): DataLoader<number, ExpenseAttachedFile[]> => {
...
+      // Check if the user has completed document
+      if (expense.legalDocRequestStatus && expense.legalDocRequestStatus === RECEIVED) {
+        data.userTaxFormRequiredBeforePayment = false;
+        return data;
+      }
+      return data;
+    })
+    
+    return sortResultsSimple(expenseIds, expenses, expense => expense.expenseId)
+    .map((expense: []) => expense.userTaxFormRequiredBeforePayment)
```
Betree, line 134: What about doing something like this:

```es6
const expenseNeedsTaxForm = {}
results.forEach((result) => {
  expenseNeedsTaxForm[result.expenseId] = expense.requiredDocument && expense.total >= THRESHOLD && expense.legalDocRequestStatus !== RECEIVED;
})

return expenseIds.map(id => expenseNeedsTaxForm[id] || false)
```

flickz: Brilliant and Elegant! Thanks 😊

server/graphql/loaders/expenses.ts
```diff
@@ -1,12 +1,42 @@
...
+    SUM (amount) AS total
+  FROM "Expenses" e
+  INNER JOIN "Collectives" c ON c.id = e."CollectiveId"
+  LEFT JOIN "RequiredLegalDocuments" d ON d."HostCollectiveId" = c."HostCollectiveId"
+                                        AND d."documentType" = 'US_TAX_FORM'
+  LEFT JOIN "LegalDocuments" ld ON ld."CollectiveId" = e."FromCollectiveId"
+                                  AND ld.year = date_part('year', e."incurredAt")
+                                  AND ld."documentType" = 'US_TAX_FORM'
+  WHERE e.status IN ('PENDING', 'APPROVED', 'PAID', 'PROCESSING', 'SCHEDULED_FOR_PAYMENT')
+  AND e."UserId" IN (SELECT "UserId" FROM "Expenses" WHERE "Expenses".id IN (:expenseIds))
```
Betree, line 32: I don't get the complexity on this line, why not directly this?

```suggestion
  AND e.id IN (:expenseIds)
```

server/graphql/loaders/expenses.ts
```diff
@@ -1,12 +1,42 @@
...
+  AND e.type NOT IN ('RECEIPT')
+  AND "incurredAt" BETWEEN e."incurredAt" AND (e."incurredAt" + interval '1 year')
```
Betree, line 34: If the expense was submitted on `01-08-2018`, this will check all the expenses that were submitted between `01-08-2018` and `01-08-2019`. But we work with fiscal years, what we want is to retrieve the data for expenses that occurred in `2018`.

Betree:
I added more tests in https://github.com/opencollective/opencollective-api/pull/4022/commits/169ef6c75625ee38850b27078419b52b151c51f5. One of them, `requires user tax form before payment when the sum of multiple expenses is above threshold` is not passing, we should support this case.

test/server/graphql/loaders/expense.test.js
```diff
@@ -0,0 +1,131 @@
...
+        const secondExpense = await fakeExpense({
+          amount: 200,
+          CollectiveId: collective.id,
+          FromCollectiveId: user.CollectiveId,
+          UserId: user.id,
+          type: 'INVOICE',
+        });
+        const result1 = await loader.load(firstExpense.id);
+        const result2 = await loader.load(secondExpense.id);
+        expect(result1).to.be.false;
```
Betree, line 57: I don't agree, the result should be `true` whether you're looking at the first expense or the second one. It's the sum that matters.

```suggestion
        expect(result1).to.be.true;
```

server/graphql/loaders/expenses.ts
```diff
@@ -1,12 +1,43 @@
...
+  LEFT JOIN "LegalDocuments" ld ON ld."CollectiveId" = e."FromCollectiveId"
+                                AND ld.year = date_part('year', e."incurredAt")
+                                AND ld."documentType" = 'US_TAX_FORM'
+  WHERE e."UserId" = er."UserId"
+  AND e.status IN ('PENDING', 'APPROVED', 'PAID', 'PROCESSING', 'SCHEDULED_FOR_PAYMENT')
+  AND e.type NOT IN ('RECEIPT')
+  AND e."deletedAt" IS NULL
+  AND e."incurredAt" BETWEEN date_trunc('year', e."incurredAt") AND (date_trunc('year', e."incurredAt") + interval '1 year')
+  GROUP BY e."UserId"
+`;
```
Betree, line 38: We're inverting the join condition and the filter condition here

```suggestion
  SELECT 
    e."UserId" "userId", 
    MAX(er.id) as "expenseId",
    MAX(ld."requestStatus") as "legalDocRequestStatus",
    MAX(d."documentType") as "requiredDocument",
    SUM (e."amount") AS total
  FROM "Expenses" e
  INNER JOIN "Expenses" er ON e."UserId" = er."UserId"
  INNER JOIN "Collectives" c ON c.id = e."CollectiveId"
  LEFT JOIN "RequiredLegalDocuments" d ON d."HostCollectiveId" = c."HostCollectiveId"
                                    AND d."documentType" = 'US_TAX_FORM'
  LEFT JOIN "LegalDocuments" ld ON ld."CollectiveId" = e."FromCollectiveId"
                                AND ld.year = date_part('year', e."incurredAt")
                                AND ld."documentType" = 'US_TAX_FORM'
  WHERE er.id IN (:expenseIds)
  AND e.status IN ('PENDING', 'APPROVED', 'PAID', 'PROCESSING', 'SCHEDULED_FOR_PAYMENT')
  AND e.type NOT IN ('RECEIPT')
  AND e."deletedAt" IS NULL
  AND e."incurredAt" BETWEEN date_trunc('year', e."incurredAt") AND (date_trunc('year', e."incurredAt") + interval '1 year')
  GROUP BY e."UserId"
`;
```

server/graphql/loaders/expenses.ts
```diff
@@ -1,12 +1,43 @@
...
+  SELECT 
+    e."UserId" "userId", 
+    MAX(er.id) as "expenseId",
+    MAX(ld."requestStatus") as "legalDocRequestStatus",
+    MAX(d."documentType") as "requiredDocument",
+    SUM (e."amount") AS total
+  FROM "Expenses" e
+  INNER JOIN "Expenses" er ON er.id IN (:expenseIds)
+  INNER JOIN "Collectives" c ON c.id = e."CollectiveId"
+  LEFT JOIN "RequiredLegalDocuments" d ON d."HostCollectiveId" = c."HostCollectiveId"
```
Betree, line 27: Small optimization: no need to check the expenses for hosts without required legal documents

```suggestion
  INNER JOIN "RequiredLegalDocuments" d ON d."HostCollectiveId" = c."HostCollectiveId"
```

Betree:
@flickz could you rebase on master? Tests fail because they're not up to date with master

znarf:
@Betree @flickz There is a regression in this PR, it seems we're now looking in the floating year (1 year before the current date) instead of looking in the fiscal year (in 2020).

Betree:
This was reported in https://github.com/opencollective/opencollective-api/pull/4022#discussion_r443664123 and an attempt to fix it was made with the `date_trunc` part. The tests for this part may be too permissive, we'll need to improve them as well.

PR #4043, opened 2020-06-11
GraphQL v2 Tier updates

sbinlondon:
* Adds more fields to the Tier object to be used on the frontend
* Adds a TierCollection
* Allows you to query Tiers on a Collective in v2
* Updates the `updateOrder` mutation to use the new v2 stuff passed from the frontend

### DIFF
--- a/server/graphql/v2/input/AmountInput.js
+++ b/server/graphql/v2/input/AmountInput.js
@@ -1,9 +1,10 @@
-import { GraphQLFloat, GraphQLObjectType } from 'graphql';
+import { GraphQLFloat, GraphQLInputObjectType, GraphQLInt } from 'graphql';
 
 import { Currency } from '../enum/Currency';
 
-export const AmountInput = new GraphQLObjectType({
+export const AmountInput = new GraphQLInputObjectType({
   name: 'AmountInput',
+  description: 'Input type for an amount with the value and currency',
   fields: () => ({
     value: {
       type: GraphQLFloat,
@@ -12,6 +13,10 @@ export const AmountInput = new GraphQLObjectType({
     currency: {
       type: Currency,
       description: 'The currency string',
+    },
+    valueInCents: {
+      type: GraphQLInt,
+      description: 'The value in cents',
     },
   }),
 });
--- a/server/graphql/v2/input/TierReferenceInput.js
+++ b/server/graphql/v2/input/TierReferenceInput.js
@@ -1,11 +1,34 @@
-import { GraphQLInputObjectType, GraphQLInt } from 'graphql';
+import { GraphQLInputObjectType, GraphQLString } from 'graphql';
+
+import models from '../../../models';
+import { NotFound } from '../../errors';
+import { idDecode } from '../identifiers';
 
 export const TierReferenceInput = new GraphQLInputObjectType({
   name: 'TierReferenceInput',
   fields: () => ({
-    legacyId: {
-      type: GraphQLInt,
-      description: 'The legacy id assigned to the Tier',
+    id: {
+      type: GraphQLString,
+      description: 'The id assigned to the Tier',
     },
   }),
 });
+
+/**
+ * Retrieves a tier
+ *
+ * @param {string|number} input - id of the tier
+ */
+export const fetchTierWithReference = async input => {
+  let tier;
+  if (input.id) {
+    const id = idDecode(input.id, 'tier');
+    tier = await models.Tier.findByPk(id);
+  } else {
+    throw new Error('Please provide an id');
+  }
+  if (!tier) {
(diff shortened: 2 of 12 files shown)


server/graphql/v2/object/Tier.js
```diff
@@ -62,6 +64,36 @@ export const Tier = new GraphQLObjectType({
...
         },
       },
+      amount: {
+        type: Amount,
+        resolve(tier) {
+          return { value: tier.amount, currency: tier.currency };
+        },
+      },
+      interval: {
+        type: GraphQLString,
```
Betree, line 74: **[:sparkles: **Enhancement**]** Would be nice to create an enum for this one

---

<details>
<br/>

:no_entry_sign: **Blocker**

:sparkles: **Enhancement**

</details>

server/graphql/v2/object/Tier.js
```diff
@@ -62,6 +64,36 @@ export const Tier = new GraphQLObjectType({
...
+        },
+      },
+      presets: {
+        type: new GraphQLList(GraphQLInt),
+        resolve(tier) {
+          return tier.presets;
+        },
+      },
+      amountType: {
+        type: GraphQLString,
```
Betree, line 86: - Should be not nullable
- Would be nice to create an enum for this one

```suggestion
        type: new GraphQLNonNull(GraphQLString),
```

server/graphql/v2/input/AmountInput.js
```diff
@@ -13,5 +14,9 @@ export const AmountInput = new GraphQLObjectType({
       type: Currency,
       description: 'The currency string',
     },
+    valueInCents: {
+      type: GraphQLFloat,
```
znarf, line 18: Should normally be an Int

server/graphql/v2/object/Tier.js
sbinlondon: Oh you're right, I think there might actually be one already 🤔

server/graphql/v2/input/TierReferenceInput.js
```diff
@@ -1,11 +1,38 @@
...
   }),
 });
+
+/**
+ * Retrieves a tier
+ *
+ * @param {string|number} input - id of the tier
+ */
+export const fetchTierWithReference = async input => {
+  const loadTierById = id => {
```
znarf, line 23: Looks like it's not needed, you can simply use `models.Tier.findByPk` later.

sbinlondon: Why would we use `models.Tier.findByPk` in this case but `fetch(Account/PaymentMethod)WithReference` in other places?

znarf: I'm writing about `loadTierById` which looks like an useless abstraction.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -184,41 +183,40 @@ const orderMutations = {
...
               `This tier (#${tierInfo.id}) doesn't belong to the given Collective #${order.CollectiveId}`,
             );
           }
         }
-        // check if the tier is different from the previous tier
-        if (tier.legacyId !== order.TierId) {
-          order = await order.update({ TierId: tier.legacyId });
+
+        // The amount can never be less than $1.00
+        if (args.amount.valueInCents < 100) {
```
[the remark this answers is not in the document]
sbinlondon: There already seems to be a math lib helper called `floatAmountToCents` to I will use that instead.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -184,41 +183,40 @@ const orderMutations = {
...
+          throw new Error('Invalid amount.');
+        }
+
+        // If using a named tier, amount can never be less than the minimum amount
+        if (tierInfo && tierInfo.amountType === 'FLEXIBLE' && args.amount.valueInCents < tierInfo.minimumAmount) {
```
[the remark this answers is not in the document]
sbinlondon: I think this is a good idea and I will write it down for enhancement in a separate PR. Right now I think it's good to get the feature out for testing.

server/graphql/v2/mutation/OrderMutations.js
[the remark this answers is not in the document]
Betree: It's not the same thing, my feedback was not about converting the value to cents but about the `AmountInput` object not being handled correctly. If we only use `amount.value` and ignore `amount.valueInCents` then there's no point in adding this field.

Let's say I'm an OC user willing to implement a feature for this endpoint and I decide to use the `valueInCents` field, my code will not work and I'll be surprised. We've introduced a bug here.

PR #4059, opened 2020-06-15
Expenses List: Search Bar Addition

SudharakaP:
This adds the api part of the search bar.

Fixes https://github.com/opencollective/opencollective/issues/3214

### DIFF
--- a/server/graphql/v2/query/ExpensesQuery.ts
+++ b/server/graphql/v2/query/ExpensesQuery.ts
@@ -59,7 +59,7 @@ const ExpensesQuery = {
     searchTerm: {
       type: GraphQLString,
       description: 'The term to search',
-    }
+    },
   },
   async resolve(_, args, req): Promise<CollectionReturnType> {
     const where = {};
@@ -90,17 +90,17 @@ const ExpensesQuery = {
         { tags: { [Op.overlap]: [args.searchTerm.toLowerCase()] } },
         { '$fromCollective.slug$': { [Op.iLike]: ilikeQuery } },
         { '$fromCollective.name$': { [Op.iLike]: ilikeQuery } },
-        { '$items.description$': { [Op.iLike]: ilikeQuery } },
+        // { '$items.description$': { [Op.iLike]: ilikeQuery } },
       ];
 
-
       include.push(
-        { model: models.Collective, as: 'fromCollective', attributes: [] },
-        { model: models.ExpenseItem, as: 'items', duplicating: false, attributes: [] }
+        { association: 'fromCollective', attributes: [] },
+        // One-to-many relationships with limits are broken in Sequelize. Could be fixed by https://github.com/sequelize/sequelize/issues/4376
+        // { association: 'items', duplicating: false, attributes: [], separate: true },
       );
 
       if (!isNaN(args.searchTerm)) {
-        where[Op.or].push({id: args.searchTerm});
+        where[Op.or].push({ id: args.searchTerm });
       }
     }
 
--- a/server/graphql/v2/query/ExpensesQuery.ts
+++ b/server/graphql/v2/query/ExpensesQuery.ts
@@ -56,6 +56,10 @@ const ExpensesQuery = {
       type: ISODateTime,
       description: 'Only return expenses that were created after this date',
     },
+    searchTerm: {
+      type: GraphQLString,
+      description: 'The term to search',
+    }
   },
   async resolve(_, args, req): Promise<CollectionReturnType> {
     const where = {};
@@ -75,6 +79,29 @@ const ExpensesQuery = {
     if (args.account) {
       const collective = await fetchAccountWithReference(args.account, fetchAccountParams);
       where['CollectiveId'] = collective.id;
+    }
+
+    // Add search filter
+    if (args.searchTerm) {
+      const sanitizedTerm = args.searchTerm.replace(/(_|%|\\)/g, '\\$1');
+      const ilikeQuery = `%${sanitizedTerm}%`;
+      where[Op.or] = [
+        { description: { [Op.iLike]: ilikeQuery } },
+        { tags: { [Op.overlap]: [args.searchTerm.toLowerCase()] } },
+        { '$fromCollective.slug$': { [Op.iLike]: ilikeQuery } },
+        { '$fromCollective.name$': { [Op.iLike]: ilikeQuery } },
+        { '$items.description$': { [Op.iLike]: ilikeQuery } },
+      ];
+
+
+      include.push(
+        { model: models.Collective, as: 'fromCollective', attributes: [] },
+        { model: models.ExpenseItem, as: 'items', duplicating: false, attributes: [] }
+      );
+
+      if (!isNaN(args.searchTerm)) {
+        where[Op.or].push({id: args.searchTerm});
+      }
     }
 
     // Add filters
@@ -111,7 +138,7 @@ const ExpensesQuery = {
 
     const order = [[args.orderBy.field, args.orderBy.direction]];
     const { offset, limit } = args;
-    const result = await models.Expense.findAndCountAll({ where, include, order, offset, limit });
+    const result = await models.Expense.findAndCountAll({ include, where, order, offset, limit });
     return {
       nodes: result.rows,
       totalCount: result.count,


server/graphql/v2/query/ExpensesQuery.ts
```diff
@@ -52,6 +56,26 @@ const ExpensesQuery = {
       where['CollectiveId'] = collective.id;
     }
 
+    // Add search filter
+    let include;
+    if (args.searchTerm) {
+      where[Op.or] = [
+        { description: { [Op.like]: `%${args.searchTerm}%` }},
```
Betree, line 63: Potential SQL injection here, what if my `searchTerm` is something like:

```es6
searchTerm = `_' OR whateverSQLYoudLikeToInject OR ' `
```

SudharakaP: @Betree : Since this was not a [raw query](https://github.com/sequelize/sequelize/issues/6991) my understanding was that it's not subjected to sql injection. Is it not the case? I've tried things like `OR ... OR` and doesn't seem to have a problem. Do you mean we sanitize the string with a custom function before passing this in? :thinking:

Betree: You're right, sequelize seems to sanitize what's passed to `searchTerm`.

However there's still a problem: if you pass a `%` character, it will be interpreted as wildcard and we want to prevent that.

SudharakaP: @Betree : Thanks for the review. I've changed it to use `Op.substring` which I think is better and removes the necessity to use any wildcard. :smile:

Betree: Fantastic!

server/graphql/v2/query/ExpensesQuery.ts
```diff
@@ -56,10 +56,14 @@ const ExpensesQuery = {
...
     },
+    searchTerm: {
+      type: GraphQLString,
+      description: 'The term to search',
+    }
   },
   async resolve(_, args, req): Promise<CollectionReturnType> {
     const where = {};
-    const include = [];
+    let include = [];
```
Betree, line 66: Let's maybe keep that `const` and `push` to make sure we won't overwrite the variable in the future

server/graphql/v2/query/ExpensesQuery.ts
```diff
@@ -77,6 +81,25 @@ const ExpensesQuery = {
...
+      where[Op.or] = [
+        { description: { [Op.substring]: args.searchTerm }},
+        { tags: { [Op.overlap]: [args.searchTerm] }},
+        { '$fromCollective.name$': { [Op.substring]: args.searchTerm }},
+        { '$items.description$': { [Op.substring]: args.searchTerm }}
+      ];
+
+      include = [
+        { model: models.Collective, as: 'fromCollective' },
+        { model: models.ExpenseItem, as: 'items', duplicating: false }
```
Betree, line 95: By adding `attributes: []` we can make sure that the `SELECT` won't include any fields, thus making the query simpler and improving the RAM usage.

```suggestion
        { model: models.Collective, as: 'fromCollective', attributes: [] },
        { model: models.ExpenseItem, as: 'items', duplicating: false, attributes: [] }
```

server/graphql/v2/query/ExpensesQuery.ts
```diff
@@ -77,6 +81,25 @@ const ExpensesQuery = {
...
     }
 
+    // Add search filter
+    if (args.searchTerm) {
+      where[Op.or] = [
+        { description: { [Op.substring]: args.searchTerm }},
+        { tags: { [Op.overlap]: [args.searchTerm] }},
+        { '$fromCollective.name$': { [Op.substring]: args.searchTerm }},
+        { '$items.description$': { [Op.substring]: args.searchTerm }}
+      ];
```
Betree, line 91: `Op.substring` looks very clean, but we need to be case-insensitive (ie. searching `food` should return titles like `Food for the team`). I can suggest:

```suggestion
      const sanitizedTerm = args.searchTerm.replace(/(_|%|\\)/g, '\\$1');
      const ilikeQuery = `${sanitizedTerm}%`;
      where[Op.or] = [
        { description: { [Op.iLike]: ilikeQuery } },
        { tags: { [Op.overlap]: [args.searchTerm.toLowerCase()] } },
        { '$fromCollective.slug$': { [Op.iLike]: ilikeQuery } },
        { '$fromCollective.name$': { [Op.iLike]: ilikeQuery } },
        { '$items.description$': { [Op.iLike]: ilikeQuery } },
      ];
```

server/graphql/v2/query/ExpensesQuery.ts
SudharakaP: Done. :+1:

server/graphql/v2/query/ExpensesQuery.ts
SudharakaP: Done. :+1:

server/graphql/v2/query/ExpensesQuery.ts
SudharakaP: @Betree : I think line 87 should be, `const ilikeQuery = `%${sanitizedTerm}%`;` since otherwise we are missing some search results. Some strings in the middle of sentences will be omitted. This maybe the reason that you see less than 5 results all the time. I've pushed these changes; let me know if you agree. :+1:

SudharakaP:
> Strangely the search term seems to affect the `limit`: when I add a `searchTerm` to the query only five items get returned. To help debug that, it may be useful to add a `logging: console.log` on the query to see the generated SQL.

@Betree : Please see my comment https://github.com/opencollective/opencollective-api/pull/4059#discussion_r442730034. I've pushed everything you suggested; let me know if there's anything else to be done. :smile:

server/graphql/v2/query/ExpensesQuery.ts
Betree: Good catch

Betree:
> > Strangely the search term seems to affect the `limit`: when I add a `searchTerm` to the query only five items get returned. To help debug that, it may be useful to add a `logging: console.log` on the query to see the generated SQL.
> 
> @Betree : Please see my comment [#4059 (comment)](https://github.com/opencollective/opencollective-api/pull/4059#discussion_r442730034). I've pushed everything you suggested; let me know if there's anything else to be done.

I don't think that [#4059 (comment)](https://github.com/opencollective/opencollective-api/pull/4059#discussion_r442730034) was the root cause, because the search properly returned all the entries. The problem seems to come from pagination, I get 4 lists of 5 expenses instead of 2 lists of 10 expenses. I can try again if you're having a hard time reproducing.

SudharakaP:
> > > Strangely the search term seems to affect the `limit`: when I add a `searchTerm` to the query only five items get returned. To help debug that, it may be useful to add a `logging: console.log` on the query to see the generated SQL.
> > 
> > 
> > @Betree : Please see my comment [#4059 (comment)](https://github.com/opencollective/opencollective-api/pull/4059#discussion_r442730034). I've pushed everything you suggested; let me know if there's anything else to be done.
> 
> I don't think that [#4059 (comment)](https://github.com/opencollective/opencollective-api/pull/4059#discussion_r442730034) was the root cause, because the search properly returned all the entries. The problem seems to come from pagination, I get 4 lists of 5 expenses instead of 2 lists of 10 expenses. I can try again if you're having a hard time reproducing.

Hmm... :thinking: Yes, I do seem to have a problem with recreating it. When I search for something with more than 10 results it gives me two pages; first one with 10 results and the second with the rest. 

Peek 2020-06-22 09-42

Am I missing something here? :thinking:

Betree:
It seems like a bug from Sequelize. I tried to log the query to run it manually:

```sql
SELECT "Expense"."id", "Expense"."UserId", "Expense"."FromCollectiveId", "Expense"."payeeLocation", "Expense"."data", "Expense"."CollectiveId", "Expense"."currency", "Expense"."amount", "Expense"."description", "Expense"."legacyPayoutMethod", "Expense"."PayoutMethodId", "Expense"."privateMessage", "Expense"."invoiceInfo", "Expense"."vat", "Expense"."lastEditedById", "Expense"."status", "Expense"."type", "Expense"."incurredAt", "Expense"."createdAt", "Expense"."updatedAt", "Expense"."deletedAt", "Expense"."tags" FROM "Expenses" AS "Expense" LEFT OUTER JOIN "Collectives" AS "fromCollective" ON "Expense"."FromCollectiveId" = "fromCollective"."id" AND ("fromCollective"."deletedAt" IS NULL) LEFT OUTER JOIN "ExpenseItems" AS "items" ON "Expense"."id" = "items"."ExpenseId" AND ("items"."deletedAt" IS NULL) WHERE ("Expense"."deletedAt" IS NULL AND (("Expense"."description" ILIKE '%Food %' OR "Expense"."tags" && ARRAY['food ']::VARCHAR(255)[] OR "fromCollective"."slug" ILIKE '%Food %' OR "fromCollective"."name" ILIKE '%Food %' OR "items"."description" ILIKE '%Food %') AND "Expense"."CollectiveId" = 83)) ORDER BY "Expense"."createdAt" DESC LIMIT 10 OFFSET 0;
```

The query itself returns 10 entries, as expected. But when using `.findAll` or `findAndCountAll`, I only get 4 entries. There are a bunch of issues matching this problem: https://github.com/sequelize/sequelize/issues?q=is%3Aissue+is%3Aopen+limit+include

I'm gonna keep digging.

SudharakaP:
> It seems like a bug from Sequelize. I tried to log the query to run it manually:
> 
> ```sql
> SELECT "Expense"."id", "Expense"."UserId", "Expense"."FromCollectiveId", "Expense"."payeeLocation", "Expense"."data", "Expense"."CollectiveId", "Expense"."currency", "Expense"."amount", "Expense"."description", "Expense"."legacyPayoutMethod", "Expense"."PayoutMethodId", "Expense"."privateMessage", "Expense"."invoiceInfo", "Expense"."vat", "Expense"."lastEditedById", "Expense"."status", "Expense"."type", "Expense"."incurredAt", "Expense"."createdAt", "Expense"."updatedAt", "Expense"."deletedAt", "Expense"."tags" FROM "Expenses" AS "Expense" LEFT OUTER JOIN "Collectives" AS "fromCollective" ON "Expense"."FromCollectiveId" = "fromCollective"."id" AND ("fromCollective"."deletedAt" IS NULL) LEFT OUTER JOIN "ExpenseItems" AS "items" ON "Expense"."id" = "items"."ExpenseId" AND ("items"."deletedAt" IS NULL) WHERE ("Expense"."deletedAt" IS NULL AND (("Expense"."description" ILIKE '%Food %' OR "Expense"."tags" && ARRAY['food ']::VARCHAR(255)[] OR "fromCollective"."slug" ILIKE '%Food %' OR "fromCollective"."name" ILIKE '%Food %' OR "items"."description" ILIKE '%Food %') AND "Expense"."CollectiveId" = 83)) ORDER BY "Expense"."createdAt" DESC LIMIT 10 OFFSET 0;
> ```
> 
> The query itself returns 10 entries, as expected. But when using `.findAll` or `findAndCountAll`, I only get 4 entries. There are a bunch of issues matching this problem: https://github.com/sequelize/sequelize/issues?q=is%3Aissue+is%3Aopen+limit+include
> 
> I'm gonna keep digging.

@Betree : Thanks for the info. Can you let me know where you see this 4 results? Is it through the UI? I wasn't able to recreate such a scenario in my tests. I'll investigate this tomorrow as well. 🙃

Betree:
@SudharakaP Through the UI, or by directly logging the `result` returned by `findAndCountAll`.

It works if I disable the search in `items`, which matches some feedback on the issues that say this bug happens mainly when there's a one-to-many relationship.

Betree:
I've disabled the search in items so we can already merge this an iterate on searching in item's description later.

SudharakaP:
@Betree : No problem. Feel free to assign the future ticket to me if needed.

PR #4063, opened 2020-06-16
Implement TransferWise account validation

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3156

### DIFF
--- a/server/graphql/v2/object/TransferWise.ts
+++ b/server/graphql/v2/object/TransferWise.ts
@@ -1,4 +1,5 @@
 import { GraphQLBoolean, GraphQLInt, GraphQLList, GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
+import GraphQLJSON from 'graphql-type-json';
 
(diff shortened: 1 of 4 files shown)


server/graphql/v2/object/TransferWise.ts
```diff
@@ -17,6 +18,7 @@ const TransferWiseFieldGroup = new GraphQLObjectType({
     name: { type: GraphQLString },
     type: { type: GraphQLString },
     required: { type: GraphQLBoolean },
+    refreshRequirementsOnChange: { type: GraphQLBoolean },
```
Betree, line 21: I don't see this field being used, is it a leftover?

kewitz: It is being used here:
https://github.com/opencollective/opencollective-frontend/pull/4441/files#diff-5790329980938a42b82c77bdd4bf5ef1R109

Betree: Got it! I misread the diff and thought it was part of the `TransferWiseRequiredField` inputs

PR #4078, opened 2020-06-18
WIP: Projects

znarf:
- Introduce `FUND` and `PROJECT` types (GraphQL v1)
- Ability to query `objects` on `collective`, similar to `events` (GraphQL v1)
- Introduce `FUND` and `PROJECT` types (GraphQL v2)
- `createProject` and `createFund` mutations (GraphQL v2)
- refactor `createCollective` mutation (single file + couple of edits to align with new mutations)

Notifications
- Remove onboarding for FUNDS

Permissions
- Introduce `isAdminOfCollective` utility (better check permissions for `EVENT` and `PROJECT`)

Host Budget
- Display "Host" payment methods for Host with budget management features
- Send `balance` instead of infinity in that case

Cleanup:
- delete `eventCollective` mutation


server/graphql/v2/mutation/CreateProjectMutation.js
```diff
@@ -0,0 +1,70 @@
...
+    type: 'PROJECT',
+    ...pick(args.project, ['name', 'slug', 'description']),
+    ...pick(parent, ['currency', 'HostCollectiveId', 'isActive', 'platformFeePercent', 'hostFeePercent']),
+    approvedAt: parent.isActive ? new Date() : null,
+    ParentCollectiveId: parent.id,
+    CreatedByUserId: remoteUser.id,
+    settings: { ...DEFAULT_PROJECT_SETTINGS, ...args.project.settings },
+  };
+
+  const checkSlug = await models.Collective.findOne({ where: { slug: projectData.slug.toLowerCase() } });
```
Betree, line 38: We also need to check the `slug` with `isBlacklistedCollectiveSlug`. Because the page relies on `/${collective}/${project}`, we may also need to add some more entries in this list to avoid any conflict. But I'd argue we should even use a sub-url like `/{collective}/projects/{project}` (ie. `/my-fund/projects/my-project`) to keep it consistent with events and avoid those kind of conflicts in the future.

Related: https://github.com/opencollective/opencollective/issues/1359

znarf: We'll use the "sub-url" pattern like Events. However, I haven't implemented the same "auto unique slug" we have in Events as I would like to reconsider that.

I will definitely add support for `isBlacklistedCollectiveSlug`. Seems, we don't use `isBlacklistedCollectiveSlug` in the new create collective flow and we should add that there too.

PR #4091, opened 2020-06-22
Funds and Projects

znarf:
Tests without matching branch ...

- Introduce `FUND` and `PROJECT` types (GraphQL v1)
- Ability to query `projects` on `collective`, similar to `events` (GraphQL v1)
- Introduce `FUND` and `PROJECT` types (GraphQL v2)
- `createProject` and `createFund` mutations (GraphQL v2)
- refactor `createCollective` mutation (single file + couple of edits to align with new mutations)

Notifications:
- Remove onboarding for `FUND`s

Permissions:
- Introduce `isAdminOfCollective` utility (better check permissions for `EVENT` and `PROJECT`)

Host Budget:
- Display "Host" payment method for Host with budget management features
- Return `balance` instead of `infinity` in that case

Cleanup:
- delete `eventCollective` mutation


server/lib/notifications.js
```diff
@@ -340,7 +340,7 @@ async function notifyByEmail(activity) {
 
     case activityType.COLLECTIVE_APPROVED:
       // Funds MVP
-      if (get(activity, 'data.collective.settings.fund') === true) {
+      if (get(activity, 'data.collective.type') === 'FUND' || get(activity, 'data.collective.settings.fund') === true) {
```
Betree, line 343: Do we still need to check the `data.collective.settings.fund` flag now that we have `collectiveType=FUND`?

znarf: For a short time yes, but that should be ultimately removed (once we update frontend and migrate previous entries).

PR #4099, opened 2020-06-23
Save Fees on Top as Transactions

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3259

### Testing locally
1. Run `dev-20200624-add-transferwise-and-feesOnTop-records` migration.
2. Donate to any given collective not hosted by Open Collective Inc adding a fee on top.
    - Make sure the host of this collective has `settings.feesOnTop: true`
    - Try the same with a collective that has a currency different from 'USD'.


server/graphql/v1/mutations/orders.js
```diff
@@ -450,6 +450,10 @@ export async function createOrder(order, loaders, remoteUser, reqIp) {
     }
     if (order.platformFee) {
       orderData.data.platformFee = order.platformFee;
+      // Let's start separating platformFee from feesOnTop, although we charge the contribution as a fee
+      // the intent is different and should not be mixed with the actual fee we charge from other contributions.
+      // This abstraction should be replicated to the frontend later on.
+      orderData.data.isFeesOnTop = true;
```
znarf, line 456: Can we modify the frontend to make sure it's explicit about "isFeesOnTop" too?

server/lib/payments.js
```diff
@@ -368,7 +378,7 @@ const sendOrderConfirmedEmail = async order => {
         EventCollectiveId: collective.id,
         UserId: user.id,
         recipient: { name: fromCollective.name },
-        order: pick(order, ['totalAmount', 'currency', 'createdAt', 'quantity']),
+        order: order.activity,
```
znarf, line 381: You lost track of `quantity`, is that expected?

znarf: Sorry, nevermind.

server/models/Transaction.js
```diff
@@ -439,51 +440,82 @@ export default (Sequelize, DataTypes) => {
...
+    transaction.CollectiveId = CollectiveId;
+    transaction.PaymentMethodId = transaction.PaymentMethodId || PaymentMethodId;
+    transaction.type = transaction.amount > 0 ? TransactionTypes.CREDIT : TransactionTypes.DEBIT;
+    transaction.hostFeeInHostCurrency = toNegative(transaction.hostFeeInHostCurrency);
+    transaction.platformFeeInHostCurrency = toNegative(transaction.platformFeeInHostCurrency);
+    transaction.taxAmount = toNegative(transaction.taxAmount);
+    transaction.paymentProcessorFeeInHostCurrency = toNegative(transaction.paymentProcessorFeeInHostCurrency);
+    // Separate donation transaction and remove platformFee from main transaction
+    if (transaction.data?.isFeesOnTop) {
+      const platformCurrencyFxRate = await getFxRate(transaction.currency, FEES_ON_TOP_TRANSACTION_PROPERTIES.currency);
```
znarf, line 472: Can you extract that block out in another model (what is specific to isFeesOnTop) to make it clearer?

znarf:
We should not make the same mistake as the Transferwise feature, the default development database should be updated so we can immediately test "Fees On Top" without having to insert anything.

znarf:
Is there a migration strategy for existing "Fees On Top" entries? This is desirable so we don't have to keep the old hacks on the Frontend side forever.

kewitz:
> We should not make the same mistake as the Transferwise feature, the default development database should be updated so we can immediately test "Fees On Top" without having to insert anything.

Agreed.
I tried to dump my DB using our script but the dump was significantly bigger than the current version, even though I only added a couple of records.
I'll make sure to update our dev dump and include everything we need.

kewitz:
> Is there a migration strategy for existing "Fees On Top" entries? This is desirable so we don't have to keep the old hacks on the Frontend side forever.

Doable. I believe we can track existing orders through `data.platformFee`, the only unknown here is related to the exchange rate for different currencies.

znarf:
For the DB, a migration targeting only the dev environment can do the job. Otherwise, you need to make sure to use the same Postgres version as the initial dump was generated with.

server/models/Transaction.js
```diff
@@ -439,51 +440,82 @@ export default (Sequelize, DataTypes) => {
...
+  Transaction.createFromPayload = async ({
     CreatedByUserId,
     FromCollectiveId,
     CollectiveId,
     transaction,
     PaymentMethodId,
   }) => {
     if (!transaction.amount) {
-      return Promise.reject(new Error('transaction.amount cannot be null or zero'));
+      return new Error('transaction.amount cannot be null or zero');
```
Betree, line 451: `return`, not `throw`?

kewitz: Thanks!

kewitz:
Rebased.

PR #4111, opened 2020-06-25
Fix newmember email error for funds

znarf:
make sure we don't send an exception in future failures

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1451,7 +1451,9 @@ export default function (Sequelize, DataTypes) {
 
       case roles.MEMBER:
       case roles.ADMIN:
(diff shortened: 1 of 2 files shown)


server/models/Collective.js
```diff
@@ -1451,7 +1451,9 @@ export default function (Sequelize, DataTypes) {
 
       case roles.MEMBER:
       case roles.ADMIN:
-        await this.sendNewMemberEmail(user, role, member, sequelizeParams);
+        if (this.type !== 'FUND') {
```
Betree, line 1454: Shouldn't that also apply to projects?

znarf: We can do the same, I'll do it. But practically, it's not possible right now to invite people to the project.

PR #4112, opened 2020-06-25
2FA proof of concept

sbinlondon:
EDIT: 

* API part of 2FA is finished
* v2 mutation to add a 2FA token to a user's account and encrypt it
* when logging in we check the decrypted 2FA token against the code the user provides
* tests have been added! 🎉


znarf:
Top!

- we should add it to the `User` table
- it can't be hashed like a password, we need to be able to read the plain version
- later one, we can implement some encryption together with other sensible data (not hash, we need to be able to decode it)

server/graphql/v1/CollectiveInterface.js
```diff
@@ -1849,6 +1852,19 @@ const CollectiveFields = () => {
         return collective;
       },
     },
+    hasTwoFactorAuth: {
+      type: GraphQLBoolean,
+      async resolve(collective) {
+        const user = await models.User.findOne({
+          where: { CollectiveId: collective.id },
+        });
+        if (user.twoFactorAuthToken) {
```
Betree, line 1861: This will crash if we try to fetch this field on something else than a user collective. For example with an org, `models.User.findOne` will return `null`

server/graphql/v2/interface/Account.js
```diff
@@ -224,6 +224,19 @@ const accountFieldsDefinition = () => ({
       return collective;
     },
   },
+  hasTwoFactorAuth: {
+    type: GraphQLBoolean,
+    async resolve(collective) {
+      const user = await models.User.findOne({
+        where: { CollectiveId: collective.id },
+      });
+      if (user.twoFactorAuthToken) {
```
Betree, line 233: Same here, this will crash if we try to fetch this field on something else than a user collective. For example with an org, `models.User.findOne` will return `null`

server/graphql/v2/interface/Account.js
```diff
@@ -224,6 +224,19 @@ const accountFieldsDefinition = () => ({
       return collective;
     },
   },
+  hasTwoFactorAuth: {
```
Betree, line 227: Shouldn't this field be on `Individual` only?

znarf, line 227: I think it should only be on `Individual` objects.

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -48,6 +48,47 @@ const accountMutations = {
...
   },
+  addAccountTwoFactorAuth: {
+    type: new GraphQLNonNull(Account),
+    description: 'Add 2FA to the Account if it does not have it',
+    args: {
+      account: {
+        type: new GraphQLNonNull(AccountReferenceInput),
+        description: 'Account that will have 2FA added to it',
+      },
+      secret: {
```
znarf, line 59: In other places, it's named `*token`, stick to one term everywhere.

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -48,6 +48,47 @@ const accountMutations = {
...
+        type: new GraphQLNonNull(GraphQLString),
+        description: 'The generated secret to save to the Account',
+      },
+    },
+    async resolve(_, args, req): Promise<object> {
+      if (!req.remoteUser) {
+        throw new Unauthorized();
+      }
+
+      return sequelize.transaction(async transaction => {
```
Betree, line 69: I don't see a need for a transaction here, there's no concurrency issue. If we decide to use one, we need to pass it to all the queries (`User.findOne`, `User.update`).

znarf, line 69: We're not using the `transaction`, it's making the code looks more complicated that it should.

Betree:
Also because this issue has sensitive security aspects, I recommend writing tests from the start (it's a good one to do test driven development, if you haven't tried yet `npm run test -- test/my-test.test.js --watch`).

Here we can already test the `addAccountTwoFactorAuth` endpoint to make sure that:
- It returns an error if user is not an admin
- It returns an error if a token is already set
- It saves the new token

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -48,6 +48,41 @@ const accountMutations = {
...
+        throw new Unauthorized();
+      }
+
+      const account = await fetchAccountWithReference(args.account);
+
+      if (!req.remoteUser.isAdminOfCollective(account)) {
+        throw new Forbidden();
+      }
+
+      const user = await models.User.findOne({ where: { id: account.CreatedByUserId } });
```
Betree, line 75: We should not use `CreatedByUserId` for that, for some accounts this can be set to someone else (ie. if you create a profile to someone on their behalf). It's mostly for logging.

Also note that this `findOne` can return `null` if no user is associated with this account (for example, if you pass an organization profile rather than a user profile). 

```suggestion
      const user = await models.User.findOne({ where: { CollectiveId: account.id } });
```

server/middleware/authentication.js
```diff
@@ -371,3 +372,32 @@ export function mustBeLoggedIn(req, res, next) {
...
+/**
+ * Checks if the user has two-factor authentication enabled on their account,
+ * and if they do, we challenge them to auth with it during the login flow
+ */
+export const twoFactorAuthCheck = (req, res, next) => {
+  const { TOTPCode } = req.body;
+  const { updateToken } = res.locals;
+  const email = req.remoteUser.email;
+  let twoFactorAuthToken, verified;
+  return User.findOne({ where: { email: email.toLowerCase() } })
```
Betree, line 385: Why `findOne`? Why not directly use `req.remoteUser`?

server/middleware/authentication.js
```diff
@@ -371,3 +372,32 @@ export function mustBeLoggedIn(req, res, next) {
...
+    .then(u => {
```
Betree, line 386: `async`/`await` could be cleaner than promises here. In any cases it seems like these 3 `.then` could fit in just one.

server/controllers/users.js
```diff
@@ -73,7 +73,22 @@ export const signin = (req, res, next) => {
  *
  * This can be used right after the first login
  */
-export const updateToken = async (req, res) => {
+export const updateToken = async (req, res, next) => {
+  const { checkingTwoFactorAuth, TOTPCode } = req.body;
```
znarf, line 77: The `TOTPCode` variable name doesn't follow our naming practices.

package.json
```diff
@@ -96,6 +96,7 @@
     "sequelize-cli": "5.5.1",
     "sequelize-temporal": "1.0.8",
     "showdown": "1.9.1",
+    "speakeasy": "^2.0.0",
```
znarf, line 99: You mentioned earlier you wanted to avoid this library. Have you changed your thoughts?

server/controllers/users.js
```diff
@@ -73,7 +73,22 @@ export const signin = (req, res, next) => {
...
+
+  // the first time we try, we need to just check if the user has 2FA
+  if (req.remoteUser.twoFactorAuthToken !== null && !checkingTwoFactorAuth) {
```
znarf, line 80: Maybe the variable `checkingTwoFactorAuth` is not useful, you could just check if user passed the `TOTPCode`.

server/middleware/authentication.js
```diff
@@ -371,3 +372,32 @@ export function mustBeLoggedIn(req, res, next) {
     }
   });
 }
+
+/**
+ * Checks if the user has two-factor authentication enabled on their account,
+ * and if they do, we challenge them to auth with it during the login flow
+ */
+export const twoFactorAuthCheck = (req, res, next) => {
```
znarf, line 380: Introducing this separate function from `updateToken` is creating complexities that could be avoided (such as `res.locals` usage).

I would suggest to move everything in `updateToken` and if necessary extract some parts of it in an helper function.

server/graphql/v2/object/Individual.js
```diff
@@ -75,6 +75,19 @@ export const Individual = new GraphQLObjectType({
           }
         },
       },
+      hasTwoFactorAuth: {
```
znarf, line 78: I haven't looked at the updated Frontend yet but I'm wondering where it's needed. Just a note to remember checking that.

package.json
sbinlondon: No, I tried otplib and was not able to make it work with Google Authenticator. I think it's still the most used authenticator library... the technology isn't really changing. So it should be fine.

server/graphql/v2/object/Individual.js
sbinlondon: Yep I have removed it from the larger query and created a small v2 query on Individual for the info needed.

config/ci.json
```diff
@@ -31,5 +31,9 @@
   },
   "fixer": {
     "accessKey": "UNDEFINED_FIXER_ACCESS_KEY"
+  },
+  "twoFactorAuth": {
+    "secretKey": "guineapigs",
+    "cipher": "DES"
```
Betree, line 37: It seems to me that we should just pick the strongest encryption available for us and stick with it. Especially because if you change this, you'll need to invalidate all existing tokens which requires a DB migration.

Is there any value in making the cipher customizable that I'm missing?

sbinlondon: Just a further layer of security I figured, if someone can't guess which one you're using.

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -48,6 +54,47 @@ const accountMutations = {
...
+        throw new NotFound('Account not found.');
+      }
+
+      if (user.twoFactorAuthToken !== null) {
+        throw new Unauthorized('This account already has 2FA enabled.');
+      }
+
+      const encryptedText = crypto[CIPHER].encrypt(args.token, SECRET_KEY).toString();
+
+      await user.update({ twoFactorAuthToken: encryptedText });
```
kewitz, line 93: Don't we need to validate the token in the backend before persisting it?
I don't have any expertise in this so I'm not completely sure, it is just a gut feeling.
If we do, perhaps we should temporarily store cache it in Redis and when it is validated we move to the DB.

Betree: That's a good point, we should do some basic validations on it (length, format...etc). We may also want to make sure that it's unique.

package.json
znarf:
```suggestion
    "speakeasy": "2.0.0",
```

We use fixed dependencies.

package.json
```diff
@@ -132,6 +134,7 @@
     "nodemon": "^2.0.4",
     "npm-run-all": "^4.1.5",
     "nyc": "^15.1.0",
+    "otpauth": "^6.0.4",
```
znarf, line 137: What is the difference between `otpauth` and `speakeasy`?

znarf, line 137:
```suggestion
    "otpauth": "6.0.4",
```

We use fixed dependencies.

znarf:
Let's not forget, the token encryption should later be extracted as a generic functionality that we can use for other sensitive data.

config/ci.json
znarf: As of today, they can guess we'll be using `DES`.

package.json
```diff
@@ -36,6 +36,7 @@
     "connect-redis": "4.0.4",
     "cookie-parser": "1.4.5",
     "cors": "2.8.5",
+    "crypto-js": "^4.0.0",
```
znarf, line 39:
```suggestion
    "crypto-js": "4.0.0",
```

Fixed dependency.

config/ci.json
sbinlondon: We aren't 😃

package.json
sbinlondon: `speakeasy` creates the long secret that you scan in the authenticator app and store in the db. Then the authenticator app creates TOTP codes (the 6-digits) based on the secret.

`otpauth` also creates TOTP codes from a secret.

So they are two sides of the same coin.

server/graphql/v2/mutation/AccountMutations.ts
sbinlondon: This is a good idea - I will add some validation!

sbinlondon: Have added length/format validation. Would be interested in discussing best ways to check uniqueness as a further improvement next PR.

PR #4119, opened 2020-06-28
Allow fiscal hosts send updates

snitin315:
Refers https://github.com/opencollective/opencollective/issues/2314

Related Frontend PR https://github.com/opencollective/opencollective-frontend/pull/4547

### DIFF
--- a/migrations/20200627143142-send-fiscal-host-updates-notification.js
+++ b/migrations/20200627143142-send-fiscal-host-updates-notification.js
@@ -1,0 +1,23 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface
+      .addColumn('Updates', 'notificationAudience', {
+        type: Sequelize.STRING,
+        defaultValue: null,
+      })
+      .then(() => {
+        return queryInterface.addColumn('UpdateHistories', 'notificationAudience', {
+          type: Sequelize.STRING,
+          defaultValue: null,
+        });
+      });
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.removeColumn('Updates', 'notificationAudience').then(() => {
+      queryInterface.removeColumn('UpdateHistories', 'notificationAudience');
+    });
+  },
+};
--- a/server/graphql/v1/mutations.js
+++ b/server/graphql/v1/mutations.js
@@ -83,6 +83,7 @@ import {
   OrderType,
   PaymentMethodType,
   TierType,
+  UpdateAudienceTypeEnum,
   UpdateType,
   UserType,
 } from './types';
@@ -514,6 +515,9 @@ const mutations = {
     args: {
       id: {
         type: new GraphQLNonNull(GraphQLInt),
+      },
+      notificationAudience: {
+        type: new GraphQLNonNull(UpdateAudienceTypeEnum),
       },
     },
     resolve(_, args, req) {
--- a/server/lib/notifications.js
+++ b/server/lib/notifications.js
@@ -94,7 +94,6 @@ function publishToSlackPrivateChannel(activity) {
  */
 async function notifySubscribers(users, activity, options = {}) {
   const { data } = activity;
-
   if (!users || users.length === 0) {
     debug('notifySubscribers: no user to notify for activity', activity.type);
     return;
@@ -166,8 +165,7 @@ export async function notifyAdminsOfCollective(CollectiveId, activity, options =
  */
 export async function notifyHostedCollectiveAdmins(CollectiveId, activity, options = {}) {
   const collective = await models.Collective.findByPk(CollectiveId);
-  const adminUsers = collective.getHostedCollectiveAdmins();
-
+  const adminUsers = await collective.getHostedCollectiveAdmins();
   return notifySubscribers(adminUsers, activity, options);
 }
 /**
--- a/server/lib/notifications.js
+++ b/server/lib/notifications.js
@@ -162,6 +162,15 @@ export async function notifyAdminsOfCollective(CollectiveId, activity, options =
 }
 
 /**
+ * Notify all admins of hosted collectives
+ */
+export async function notifyHostedCollectiveAdmins(CollectiveId, activity, options = {}) {
+  const collective = await models.Collective.findByPk(CollectiveId);
+  const adminUsers = collective.getHostedCollectiveAdmins();
+
+  return notifySubscribers(adminUsers, activity, options);
+}
+/**
  * Notify all the followers of the conversation.
  */
 export async function notififyConversationFollowers(conversation, activity, options = {}) {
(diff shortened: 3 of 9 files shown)


server/graphql/v1/mutations.js
```diff
@@ -515,6 +515,9 @@ const mutations = {
       id: {
         type: new GraphQLNonNull(GraphQLInt),
       },
+      updateNotificationAudience: {
+        type: GraphQLString,
```
Betree, line 519: Can we have an enum for this one? Like:

```es6
export const UpdateAudience = new GraphQLEnumType({
  name: 'UpdateAudience',
  description: 'Defines targets for an update',
  values: {
    ALL: {
      description: 'Will be sent to collective admins and financial contributors',
    },
    COLLECTIVE_ADMINS: {
      description: 'Will be sent to collective admins',
    },
    FINANCIAL_CONTRIBUTORS: {
      description: 'Will be sent to financial contributors',
    },
  },
});

```

server/lib/notifications.js
```diff
@@ -201,14 +201,48 @@ async function notifyByEmail(activity) {
 
     case activityType.COLLECTIVE_UPDATE_PUBLISHED:
       twitter.tweetActivity(activity);
-      activity.data.update = await models.Update.findByPk(activity.data.update.id, {
```
Betree, line 204: Why getting rid of this line, was it not needed?

server/lib/notifications.js
```diff
@@ -201,14 +201,48 @@ async function notifyByEmail(activity) {
...
+      activity.data.collective = await models.Collective.findByPk(activity.data.collective.id);
+      activity.data.hostedCollectives = activity.data.collective.getHostedCollectives();
+
+      if (activity.data.collective.info.isHostAccount) {
+        switch (activity.data.update.updateNotificationAudience) {
+          case 'Hosted Collective Admins':
+            activity.data.hostedCollectives.map(collective => {
+              notifyAdminsOfCollective(collective.id, activity, {
+                from: `${activity.data.collective.name}
+                <no-reply@${activity.data.collective.slug}.opencllective.com>`,
```
Betree, line 214: This `from` param is alway the same, maybe we can extract the full object to an `emailParams` variable?

server/lib/notifications.js
```diff
@@ -201,14 +201,48 @@ async function notifyByEmail(activity) {
...
-        from: `${activity.data.collective.name}
-        <no-reply@${activity.data.collective.slug}.opencollective.com>`,
-      });
+
+      activity.data.collective = await models.Collective.findByPk(activity.data.collective.id);
+      activity.data.hostedCollectives = activity.data.collective.getHostedCollectives();
+
+      if (activity.data.collective.info.isHostAccount) {
+        switch (activity.data.update.updateNotificationAudience) {
+          case 'Hosted Collective Admins':
```
Betree, line 210: Here we should use the values from the enum suggested above, `Hosted Collective Admins` looks like a label, not a key

server/lib/notifications.js
```diff
@@ -201,14 +201,48 @@ async function notifyByEmail(activity) {
...
+            activity.data.hostedCollectives.map(collective => {
```
Betree, line 211: The Open Source Collective is hosting 1200+ collectives. `notifyAdminsOfCollective` triggers at least 2 SQL queries (and usually more) so this loop will trigger more than 2500 SQL queries at the same time, which is really too much stress for the DB, we need to optimize this. Ideally we should have one single SQL query that loads all the hosted collective admins emails at once. A new `notifyHostedCollectiveAdmins` helper would be a good way to achieve that.

migrations/20200627143142-send-fiscal-host-updates-notification.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface
+      .addColumn('Updates', 'updateNotificationAudience', {
```
Betree, line 6: We don't need to repeat the word "update" here, because we're in the "Update" table (like you wouldn't do `Comment.commentTags`, just `Comment.tags`).

```suggestion
      .addColumn('Updates', 'notificationAudience', {
```

server/lib/notifications.js
snitin315: we only need `activity.data.update.collectiveId`, which we already have access to.

server/lib/notifications.js
```diff
@@ -161,6 +161,17 @@ export async function notifyAdminsOfCollective(CollectiveId, activity, options =
...
+/**
+ * Notify all admins of hosted collectives
+ */
+export async function notifyHostedCollectiveAdmins(CollectiveId, activity, options = {}) {
+  const collective = await models.Collective.findByPk(CollectiveId);
+  const hostedCollectives = collective.getHostedCollectives();
+
+  hostedCollectives.map(collective => {
+    notifyAdminsOfCollective(collective.id, activity, options);
+  });
```
Betree, line 173: My previous feedback still applies here.

`notifyAdminsOfCollective` triggers at least 2 SQL queries (and usually more) so this loop will trigger more than 2500 SQL queries at the same time for the Open Source Collective.

Ideally we should have one single SQL query that loads all the hosted collective admins emails at once.

server/graphql/v1/mutations.js
```diff
@@ -515,6 +516,9 @@ const mutations = {
       id: {
         type: new GraphQLNonNull(GraphQLInt),
       },
+      notificationAudience: {
+        type: UpdateAudienceTypeEnum,
```
Betree, line 520: What I understand from the frontend code is that this field will always be set, so let's make it mandatory:

```suggestion
        type: new GraphQLNonNull(UpdateAudienceTypeEnum),
```

snitin315:
@Betree fixed :+1: , thanks for pointing.
Screenshot at 2020-07-09 14-25-00

server/lib/notifications.js
snitin315: ah yes , `info.html` content from here was used in email templates. That's why it went missing.

server/models/Update.js
```diff
@@ -191,6 +196,8 @@ export default function (Sequelize, DataTypes) {
             id: this.id,
             slug: this.slug,
             title: this.title,
+            html: this.html,
+            notificationAudience: this.notificationAudience,
             CollectiveId: this.CollectiveId,
```
snitin315, line 201: added `html` to render in notification template.

templates/emails/collective.update.published.hbs
```diff
@@ -35,7 +35,7 @@ Subject: {{update.title}}
 <br />
 
 <h2>{{update.title}}</h2>
-<div class="byline">by <a href="{{config.host.website}}/{{update.fromCollective.slug}}">{{update.fromCollective.name}}</a></div>
+<div class="byline">by <a href="{{config.host.website}}/{{collective.slug}}">{{collective.name}}</a></div>
```
Betree: `fromCollective` and `collective` are not the same thing. `fromCollective` is the profile who submitted the update (so the user profile). `collective` is the profile where the update was posted.

I've fixed this behavior to include back the user name in my last commit.

server/lib/notifications.js
```diff
@@ -161,6 +161,15 @@ export async function notifyAdminsOfCollective(CollectiveId, activity, options =
...
 }
 
+/**
+ * Notify all admins of hosted collectives
+ */
+export async function notifyHostedCollectiveAdmins(CollectiveId, activity, options = {}) {
+  const collective = await models.Collective.findByPk(CollectiveId);
+  const adminUsers = collective.getHostedCollectiveAdmins();
+
+  return notifySubscribers(adminUsers, activity, options);
```
Betree, line 171: We're lucky that it worked (we're using `bluebird` in this file, which does some magic with promises) but it's an invalid usage, because `notifySubscribers` check this like `users.length`.

It also generates a warning:
```
Unhandled rejection TypeError: options argument must be an object but it is [object String]
    at map (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/map.js:156:35)
    at Promise.map (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/map.js:167:12)
    at /home/user/Workspace/Projects/opencollective/opencollective-api/server/lib/notifications.js:105:11
    at Generator.next (<anonymous>)
    at Generator.tryCatcher (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/util.js:16:23)
    at PromiseSpawn._promiseFulfilled (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/generators.js:97:49)
    at /home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/generators.js:201:15
    at _notifySubscribers (/home/user/Workspace/Projects/opencollective/opencollective-api/server/lib/notifications.js:166:29)
    at notifySubscribers (/home/user/Workspace/Projects/opencollective/opencollective-api/server/lib/notifications.js:130:29)
    at /home/user/Workspace/Projects/opencollective/opencollective-api/server/lib/notifications.js:171:10
    at Generator.next (<anonymous>)
    at Generator.tryCatcher (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/util.js:16:23)
    at PromiseSpawn._promiseFulfilled (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/generators.js:97:49)
    at Promise._settlePromise (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/promise.js:609:26)
    at Promise._settlePromise0 (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/promise.js:649:10)
    at Promise._settlePromises (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/promise.js:729:18)
    at _drainQueueStep (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/async.js:93:12)
    at _drainQueue (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/async.js:86:9)
    at Async._drainQueues (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/async.js:102:5)
    at Immediate.Async.drainQueues [as _onImmediate] (/home/user/Workspace/Projects/opencollective/opencollective-api/node_modules/bluebird/js/release/async.js:15:14)
    at processImmediate (internal/timers.js:439:21)
    at process.topLevelDomainCallback (domain.js:126:23)
```

We should await: 

```suggestion
  const adminUsers = await collective.getHostedCollectiveAdmins();
  return notifySubscribers(adminUsers, activity, options);
```

templates/emails/collective.update.published.hbs
snitin315: Oh! thanks :+1:

snitin315:
@Betree Thanks for the modifications.

PR #4135, opened 2020-06-30
update fund creation copy

piamancini:
just a copyright update


znarf:
@piamancini Maybe that's a spell checker issue, but should we actually say "copywrite"? "copy" for short?

PR #4170, opened 2020-07-03
Fix mark order as paid for Hosts

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3310

- Fix missing orders in the dashboard for orders made directly to the host
- Fix mark order as paid

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -867,9 +867,17 @@ export async function markOrderAsPaid(remoteUser, id) {
   if (order.status !== 'PENDING') {
     throw new ValidationFailed("The order's status must be PENDING");
   }
-  const HostCollectiveId = await models.Collective.getHostCollectiveId(order.CollectiveId);
-  if (!remoteUser.isAdmin(HostCollectiveId)) {
-    throw new Unauthorized('You must be logged in as an admin of the host of the collective');
+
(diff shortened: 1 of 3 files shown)


server/graphql/v1/mutations/orders.js
```diff
@@ -867,9 +867,17 @@ export async function markOrderAsPaid(remoteUser, id) {
   if (order.status !== 'PENDING') {
     throw new ValidationFailed("The order's status must be PENDING");
   }
-  const HostCollectiveId = await models.Collective.getHostCollectiveId(order.CollectiveId);
-  if (!remoteUser.isAdmin(HostCollectiveId)) {
-    throw new Unauthorized('You must be logged in as an admin of the host of the collective');
+
+  const collective = await models.Collective.findByPk(order.CollectiveId);
+  if (collective.isHostAccount) {
+    if (!remoteUser.isAdmin(collective.id)) {
```
kewitz, line 873: :ice_cream: You can simplify scopes by combining these conditionals.

Betree: If I just combine them, I'll need to update the `else` to an `else if` or to create a new `if` block because we don't want it to run if the collective is not a host (ie. `if (isHost && !isAdmin) { ... } else if (!isHost) { ... }`)

kewitz: Oh, I get it! Thanks!

PR #4179, opened 2020-07-06
Tier: Remove deprecated fields

Betree:
Require https://github.com/opencollective/opencollective-frontend/pull/4596

Seems that `maxQuantityPerUser` and `password` were never fully implemented, we have no occurrences in the DB.


Betree:
@kewitz good catch!

> Also, should we update GraphQL schemas in the frontend?

Yes, I was thinking about creating a single PR to update them once we merge the two "deprecate" PRs as there's nothing breaking

PR #4193, opened 2020-07-09
Account for fees on top in our platform reports.

kewitz:
For context on the way the fees on top transactions work:
  - The original transaction has 0 `platformFee`.
  - We create a new pair of transactions between donor and OpenCollective (id: 1).
    - These transactions are always in USD.
    - These transactions have the fee on top amount as the `amount` and `data.isFeeOnTop = true`.

Updates in the monthly report:
Screenshot from 2020-07-09 18 55 29

Updates in the weekly report:
Screenshot from 2020-07-09 18 52 56

### DIFF
--- a/reports/platform-report.js
+++ b/reports/platform-report.js
@@ -18,7 +18,15 @@ const query = (query, values) => {
 const getPlatformRevenue = async (startDate, endDate) => {
   const res = await query(
     `
-  SELECT currency, -SUM("platformFeeInHostCurrency") as "amount"
+  SELECT
+    -- In the case this is a fees on top transaction, the currency will already be USD
+    currency,
+    SUM(
+      CASE
+        WHEN (t."data"->>'isFeesOnTop' = 'true' AND t."type" = 'CREDIT' AND t."CollectiveId" = 1) THEN t."amount"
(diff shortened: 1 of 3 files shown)


reports/platform-report.js
```diff
@@ -137,17 +145,63 @@ async function PlatformReport(year, month) {
...
+    hosts = await Promise.all(
+      hosts.map(async host => {
+        const isUSD = host.currency === 'USD';
+        return {
+          ...host,
+          isUSD,
+          totalFeesDue: isUSD ? (host.feesOnTopDue || 0) + host.platformFeesDue : null,
+        };
+      }),
+    );
```
Betree, line 203: I don't get why we need to make them promises here, can't we simply do:

```suggestion
    hosts = hosts.map(host => {
      const isUSD = host.currency === 'USD';
      return {
        ...host,
        isUSD,
        totalFeesDue: isUSD ? (host.feesOnTopDue || 0) + host.platformFeesDue : null,
      };
    });
```

kewitz: Perfect! I was using FXrate before but I decided to remove and just account the `feesOnTopDue` in the USD.

PR #4207, opened 2020-07-13
Add fees on top to Order for recurring contributions

sbinlondon:
Relates opencollective/opencollective#3290

### DIFF
--- a/server/graphql/v2/object/Order.js
+++ b/server/graphql/v2/object/Order.js
@@ -112,6 +112,17 @@ export const Order = new GraphQLObjectType({
           return models.PaymentMethod.findByPk(order.PaymentMethodId);
         },
       },
+      platformFee: {
+        type: Amount,
+        description: 'If an Order has fees on top, we should return the amount',
+        resolve(order) {
+          if (order.data.isFeesOnTop) {
+            return { value: order.data.platformFee };
+          } else {
+            return null;
+          }
+        },
+      },
     };
   },
 });


server/graphql/v2/object/Order.js
```diff
@@ -112,6 +112,17 @@ export const Order = new GraphQLObjectType({
           return models.PaymentMethod.findByPk(order.PaymentMethodId);
         },
       },
+      platformFee: {
+        type: Amount,
+        description: 'If an Order has fees on top, we should return the amount',
+        resolve(order) {
+          if (order.data.isFeesOnTop) {
+            return { value: order.data.platformFee };
```
znarf, line 120: We should add `currency`, next to `value` otherwise it will always default to USD.

server/graphql/v2/object/Order.js
```diff
@@ -112,6 +112,17 @@ export const Order = new GraphQLObjectType({
           return models.PaymentMethod.findByPk(order.PaymentMethodId);
         },
       },
+      platformFee: {
```
znarf, line 115: It's time for us to find a more positive way to express this than `platformFee`. Exposing this field in the API is the right moment to think about it.

`onTopAmount` ? Any other idea @kewitz and @sbinlondon ?

kewitz: I think that works. Perhaps it would be clearer to move towards something like `platformContributionAmount` since these are not properly fees and neither can be called donations.

server/graphql/v2/object/Order.js
sbinlondon: @kewitz is right, this is why I didn't include the currency

server/graphql/v2/object/Order.js
sbinlondon: Agreed to a separate PR for consistency

znarf:
@sbinlondon It's ok to work on that in another PR but it needs to be immediately tackled. Introducing properties in the GraphQL API v2 is an important matter and I don't want us to expose `platformFee`.

server/graphql/v2/object/Order.js
znarf: @sbinlondon @kewitz It's not because it's the same currency as the `Order` that we should not explicitly set the currency of the `Amount`. As a general rule, when exposing an `Amount` in the GraphQL API v2, it should always have a `currency`.

PR #4208, opened 2020-07-13
Forbid collectives from hiding their funds

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3324

Hiding the budget section is only enabled for Funds and Projects (which are a kind of fund sub-collective).

### DIFF
--- a/server/graphql/v2/mutation/AccountMutations.ts
+++ b/server/graphql/v2/mutation/AccountMutations.ts
@@ -4,6 +4,7 @@ import { cloneDeep, set } from 'lodash';
 
 import { sequelize } from '../../../models';
 import { Forbidden, Unauthorized } from '../../errors';
+import { AccountTypeToModelMapping } from '../enum/AccountType';
 import { AccountReferenceInput, fetchAccountWithReference } from '../input/AccountReferenceInput';
 import { Account } from '../interface/Account';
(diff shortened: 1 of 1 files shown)


server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -42,6 +43,16 @@ const accountMutations = {
           throw new Forbidden();
         }
 
+        if (
+          args.key === 'collectivePage' &&
+          ![AccountTypeToModelMapping.FUND, AccountTypeToModelMapping.PROJECT].includes(account.type)
+        ) {
+          const budgetSection = args.value.sections?.find(s => s.section === 'budget');
+          if (!budgetSection.isEnabled) {
```
Betree, line 51: Isn't this going to crash if you don't provide a `section` budget (because `budgetSection` will be null)?

kewitz: Yes, already fixed here: https://github.com/opencollective/opencollective-api/pull/4221/commits/ed5361b9021ff7528b83cbb6f6c6bad836c9b78a

kewitz:
@Betree we sent an email to these collectives, more information here:
https://opencollective.slack.com/archives/G46PNRCGP/p1594737738014600

PR #4243, opened 2020-07-20
GQLV2 endpoints for contribution flow

Betree:


sbinlondon:
Good work! I think the `Contributors` object will make things a lot clearer for v2.

PR #4253, opened 2020-07-21
Encrypt DB

kewitz:
Resolves https://github.com/opencollective/opencollective-security/issues/49.

Before merging:
- [ ] Rename `TWO_FACTOR_AUTH_CIPHER` and `TWO_FACTOR_AUTH_SECRET_KEY` env variables to `DB_ENCRYPTION_CIPHER` and `DB_ENCRYPTION_SECRET_KEY`.


migrations/20200721091300-encrypt-tokens.js
```diff
@@ -0,0 +1,57 @@
...
+import { crypto } from '../server/lib/encryption';
+
+module.exports = {
+  up: async queryInterface => {
+    let lastId = 0,
+      count = 0,
+      total = 0;
+    console.info('Encrypting ConnectedAccounts.');
+    do {
+      const accounts = await models.ConnectedAccount.findAll({
```
Betree, line 13: Using the model to retrieve/edit data in migrations is dangerous because it may be out of sync with the DB. Here is why:

<details>
1. You create a model like `Animal { name }`
2. You create a migration with `models.Animal.findAll(...)` in it.
3. You add a `type` column to Animal

When (2) is going to run for the first time (while (3) doesn't exist yet) your SQL query is going to look like this:

```sql
SELECT name FROM animal
```

No problem here. **But** when future contributors will run the migrations from scratch after (3) has been added to the codebase, the query from (2) will look like this:

```sql
SELECT name, type FROM animal
```

This is going to crash because `type` has not been created yet. 

The rule is: we shouldn't assume that the model and the database are synchronized when used in transactions (the goal of transactions being precisely to reconciliate the database with the model).
</details>

To fix it, you can either:
- Specify `attributes`, to make sure we only fetch the required attributes
- Or use raw SQL

server/lib/encryption.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+const SECRET_KEY = config.dbEncryption.secretKey;
+const CIPHER = config.dbEncryption.cipher;
+
+/**
+ * SecretKey based authentication.
+ * Used for DB encryption of tokens.
+ */
+export const crypto = {
```
Betree, line 50: Nice lib! This is going to be very useful.

migrations/20200721091000-create-hash-column.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('ConnectedAccounts', 'hash', { type: Sequelize.STRING });
```
Betree, line 5: Maybe we could enforce the uniqueness at the DB level for more safety? (except if we have duplicate tokens in the DB today)

```suggestion
    await queryInterface.addColumn('ConnectedAccounts', 'hash', { type: Sequelize.STRING, unique: true });
```

migrations/20200721091300-encrypt-tokens.js
znarf: Alternative, move the file to the `scripts` folder and run it manually.

migrations/20200721091000-create-hash-column.js
kewitz: I actually coded that way, the problem is that we still have duplicated tokens in our DB that'll require a manual process to solve.
I'll put that back in my priorities list.

kewitz:
I rewrote the migration using raw SQL.
I also wrapped the migration in a transaction, which is giving me the safety I needed to actually execute this.

PR #4255, opened 2020-07-21
Restrict access to ConnectedAccount username

kewitz:
Resolves https://github.com/opencollective/opencollective-security/issues/34

This property is only leaking on V1 and it is currently being used to display GitHub and Twitter usernames.
I think this is a genuine application and I'm keeping it enabled for these services.


kewitz:
@Betree, perfect!

PR #4211, opened 2020-07-14
Adds TransferWise Strong Customer Authentication for fundTransfer

kewitz:
Resolves: https://github.com/opencollective/opencollective/issues/3285

Implements https://api-docs.transferwise.com/#payouts-guide-strong-customer-authentication.

PR #4256, opened 2020-07-21
Catch and sanitize errors thrown in TW SCA request

kewitz:
Related to https://github.com/opencollective/opencollective-api/pull/4211#discussion_r457941590


kewitz:
@Betree I refactored `fundTransfer` to use `requestDataAndThrowParsedError`. It was my original intent, I just forgot to remove the development version from `fundTransfer` at the time.

PR #4276, opened 2020-07-27
Remove the term blacklist from the codebase

znarf:
It's now seen as a discriminating word that should be avoided.


Betree:
I have read multiple articles on this topic that were stating that the word had no racial origin. I'm not against the change, but would love to hear more arguments from people who feels offended by them.

znarf:
On a personal side, I agree with you and I'm not convinced that blacklist and whitelist are actually from a racial origin.

However, the industry is moving in this direction and we don't want to be on the wrong side of history. Major projects like Linux are adopting such policies, ie: https://www.zdnet.com/article/linux-team-approves-new-terminology-bans-terms-like-blacklist-and-slave/

PR #4281, opened 2020-07-28
Add Tier.type to GraphQL v2

znarf:


server/graphql/v2/object/Tier.js
```diff
@@ -64,6 +64,9 @@ export const Tier = new GraphQLObjectType({
           return { value: tier.amount, currency: tier.currency };
         },
       },
+      type: {
+        type: TierType,
```
Betree, line 68: Looks like this should be NonNull, but there's no DB constraint neither

znarf: Adding that

PR #4288, opened 2020-07-28
Transactions Redesign

kewitz:
Related to https://github.com/opencollective/opencollective/issues/2434


server/graphql/v2/interface/Transaction.js
```diff
@@ -151,5 +167,31 @@ export const TransactionFields = () => {
...
+        } else {
+          return null;
+        }
+      },
+    },
+    order: {
+      type: Order,
+      resolve(transaction) {
+        if (transaction.OrderId) {
+          return transaction.getOrder({ paranoid: false });
```
Betree, line 184: Let's use a loader to prevent N+1 queries

```suggestion
          return req.loaders.Order.byId.load(transaction.OrderId);
```

server/graphql/v2/query/TransactionsQuery.ts
```diff
@@ -0,0 +1,128 @@
...
+        { '$fromCollective.name$': { [Op.iLike]: ilikeQuery } },
+        { '$collective.slug$': { [Op.iLike]: ilikeQuery } },
+        { '$collective.name$': { [Op.iLike]: ilikeQuery } },
+      ];
+
+      include.push(
+        { association: 'fromCollective', attributes: [] },
+        { association: 'collective', attributes: [] },
+        // One-to-many relationships with limits are broken in Sequelize. Could be fixed by https://github.com/sequelize/sequelize/issues/4376
+        // { association: 'items', duplicating: false, attributes: [], separate: true },
```
Betree, line 90: This won't be required for transactions

```suggestion
```

server/graphql/v2/interface/Transaction.js
```diff
@@ -78,6 +91,9 @@ export const TransactionFields = () => {
         return idEncode(transaction.id, 'transaction');
       },
     },
+    uuid: {
+      type: GraphQLString,
```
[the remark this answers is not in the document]
Betree: @kewitz I just reviewed the implication and it seems ok. We however need to update https://github.com/opencollective/opencollective-api/blob/1180877ec550c619cd4fa0ad3aff7cd2aacb015e/server/graphql/v1/queries.js#L312 to change the comment and add a small check for permissions

PR #4297, opened 2020-07-29
Fix host fee with feesOnTop bug

kewitz:
Fix https://github.com/opencollective/opencollective/issues/3370

image


znarf:
Same feedback as https://github.com/opencollective/opencollective-api/pull/4253#discussion_r462157380

It's problematic the use the model in `migrations`, so I suggest to move it to the `scripts` folder and run it manually instead.

kewitz:
@znarf sorry, but I don't get it, there's no model being used in the migration.

znarf:
@kewitz ok, sorry, I got it wrong

PR #4308, opened 2020-07-31
Account: Add platformContributionAvailable field

Betree:
Requirement for https://github.com/opencollective/opencollective-frontend/pull/4707


znarf:
As I already stated in https://github.com/opencollective/opencollective-api/pull/4207 it's important to move away from the "fees on top" on term. I don't mind about internal variables but when it comes to the API it's really important to not introduce it because it's gonna cost energy to move away from it.

Let's replace with `platformContributionAvailable`.

Betree:
I'll wait for https://github.com/opencollective/opencollective-api/pull/4308 to re-implement on top of it

PR #4316, opened 2020-08-03
Account: Add more interfaces for accounts

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3378 (see issue for motivation behind this PR)

To test:

```gql
{
  account(slug: "babel") {
    id
    ... on AccountWithHost {
      host {
        id
      }
    }
  }
}
```

### DIFF
--- a/server/graphql/v2/interface/AccountWithHost.ts
+++ b/server/graphql/v2/interface/AccountWithHost.ts
@@ -1,0 +1,63 @@
+import { GraphQLBoolean, GraphQLFloat, GraphQLInterfaceType, GraphQLNonNull } from 'graphql';
+import { GraphQLDateTime } from 'graphql-iso-date';
+import { isNil } from 'lodash';
+
+import { HOST_FEE_STRUCTURE } from '../../../constants/host-fee-structure';
+import { hostResolver } from '../../common/collective';
+import { HostFeeStructure } from '../enum/HostFeeStructure';
+import { Host } from '../object/Host';
+
+export const AccountWithHostFields = {
+  host: {
+    description: 'Returns the Fiscal Host',
+    type: Host,
+    resolve: hostResolver,
+  },
+  hostFeesStructure: {
+    description: 'Describe how the host charges the collective',
+    type: HostFeeStructure,
+    resolve: async (account, _, req): Promise<HOST_FEE_STRUCTURE | null> => {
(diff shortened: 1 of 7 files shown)


server/graphql/v2/interface/AccountWithHost.ts
```diff
@@ -0,0 +1,63 @@
...
+    },
+  },
+  isApproved: {
+    description: "Returns whether it's approved by the Fiscal Host",
+    type: GraphQLNonNull(GraphQLBoolean),
+    resolve(account): boolean {
+      return account.isApproved();
+    },
+  },
+  isActive: {
```
znarf, line 50: We're using this on Host Organizations to know if they activated their budget.

server/graphql/v2/interface/AccountWithHost.ts
```diff
@@ -0,0 +1,63 @@
...
+        return HOST_FEE_STRUCTURE.DEFAULT;
+      } else {
+        const host = await req.loaders.Collective.byId.load(account.HostCollectiveId);
+        return account.hostFeePercent === host.hostFeePercent
+          ? HOST_FEE_STRUCTURE.DEFAULT
+          : HOST_FEE_STRUCTURE.CUSTOM_FEE;
+      }
+    },
+  },
+  hostFeePercent: {
```
znarf, line 32: That might be used to expose the `hostFeePercent` of the host itself.

server/graphql/v2/interface/AccountWithHost.ts
Betree: Should be alright, `Host` and `Organization` object are normally not impacted by this

server/graphql/v2/interface/AccountWithHost.ts
Betree: `Host` has its own implementation of this field, because it doesn't inherit from `AccountWithHost` there's no risk of overriding.

server/graphql/v2/interface/AccountWithHost.ts
```diff
@@ -0,0 +1,63 @@
...
+    description: 'Fees percentage that the host takes for this collective',
+    type: GraphQLFloat,
```
kewitz, line 34: Why are we using float here?

PR #4322, opened 2020-08-04
Refactor oppositeTransaction 

znarf:
so that it acknowledges better balance and currency

### DIFF
--- a/server/models/Transaction.js
+++ b/server/models/Transaction.js
@@ -411,19 +411,41 @@ export default (Sequelize, DataTypes) => {
       transaction.amountInHostCurrency = Math.round(transaction.amountInHostCurrency);
     }
 
-    const oppositeTransaction = {
+    // Is the target "collective" (account) "Active" (has an host, manage its own budget)
+    const fromCollective = await models.Collective.findByPk(transaction.FromCollectiveId);
+
+    let oppositeTransaction = {
       ...transaction,
       type: -transaction.amount > 0 ? TransactionTypes.CREDIT : TransactionTypes.DEBIT,
       FromCollectiveId: transaction.CollectiveId,
       CollectiveId: transaction.FromCollectiveId,
-      HostCollectiveId: await models.Collective.getHostCollectiveId(transaction.FromCollectiveId), // see https://github.com/opencollective/opencollective/issues/1154
-      amount: -transaction.netAmountInCollectiveCurrency,
-      netAmountInCollectiveCurrency: -transaction.amount,
       amountInHostCurrency: Math.round(-transaction.netAmountInCollectiveCurrency * transaction.hostCurrencyFxRate),
       hostFeeInHostCurrency: transaction.hostFeeInHostCurrency,
       platformFeeInHostCurrency: transaction.platformFeeInHostCurrency,
       paymentProcessorFeeInHostCurrency: transaction.paymentProcessorFeeInHostCurrency,
     };
+    if (!fromCollective.isActive) {
+      oppositeTransaction = {
+        ...oppositeTransaction,
+        HostCollectiveId: null,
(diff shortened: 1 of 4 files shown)


server/models/Transaction.js
```diff
@@ -411,19 +411,54 @@ export default (Sequelize, DataTypes) => {
...
-      netAmountInCollectiveCurrency: -transaction.amount,
-      amountInHostCurrency: Math.round(-transaction.netAmountInCollectiveCurrency * transaction.hostCurrencyFxRate),
-      hostFeeInHostCurrency: transaction.hostFeeInHostCurrency,
-      platformFeeInHostCurrency: transaction.platformFeeInHostCurrency,
-      paymentProcessorFeeInHostCurrency: transaction.paymentProcessorFeeInHostCurrency,
-    };
+    // Is the target "collective" (account) "Active" (has an host, manage its own budget)
+    const fromCollective = await models.Collective.findByPk(transaction.FromCollectiveId);
+
+    let oppositeTransaction;
```
kewitz, line 417: Start `oppositeTransaction` with the common fields already populated, this will simplify the following conditional outcomes.

server/models/Transaction.js
```diff
@@ -411,19 +411,54 @@ export default (Sequelize, DataTypes) => {
...
+        amount: -transaction.netAmountInCollectiveCurrency,
+        netAmountInCollectiveCurrency: -transaction.amount,
+
+        amountInHostCurrency: Math.round(-transaction.netAmountInCollectiveCurrency * transaction.hostCurrencyFxRate),
+        hostFeeInHostCurrency: transaction.hostFeeInHostCurrency,
+        platformFeeInHostCurrency: transaction.platformFeeInHostCurrency,
+        paymentProcessorFeeInHostCurrency: transaction.paymentProcessorFeeInHostCurrency,
+      };
+    } else {
+      const oppositeTransactionCurrencyFxRate = await getFxRate(
```
kewitz, line 436: I don't think calling `getFxRate` is that expensive, perhaps we should apply this logic even in the case the collective is not active and simplify this function even further. I'm fine with both.

server/models/Transaction.js
```diff
@@ -411,19 +411,54 @@ export default (Sequelize, DataTypes) => {
...
+        FromCollectiveId: transaction.CollectiveId,
+        CollectiveId: transaction.FromCollectiveId,
+
+        // Added
+        HostCollectiveId: fromCollective.HostCollectiveId,
+
+        // Converted to fromCollective currency
+        currency: fromCollective.currency,
+        amount: -Math.round(transaction.netAmountInCollectiveCurrency * oppositeTransactionCurrencyFxRate),
+        netAmountInCollectiveCurrency: -Math.round(transaction.amount * oppositeTransactionCurrencyFxRate),
```
kewitz, line 453: Nice! :+1:

server/models/Transaction.js
```diff
@@ -411,19 +411,54 @@ export default (Sequelize, DataTypes) => {
...
+        data: { oppositeTransactionCurrencyFxRate },
```
kewitz, line 454: Shouldn't we also persist the existing transaction data?
```suggestion
        data: { ...transaction.data, oppositeTransactionCurrencyFxRate },
```

znarf: I considered, but I thought it was already persisted in the original transaction.

kewitz: I'm worried about being able to fetch CREDITs or DEBITs where `data.isFeesOnTop` is true in a simple query.
I'm not sure if there's any other case where we would be interested in fetching by data.

znarf: Sorry, got it wrong, we should definitely do that.

PR #4334, opened 2020-08-06
Disable contentSecurityPolicy

znarf:
It's currently breaking GraphQL playgrounds.

Let's figure a policy that makes sense for us before activating it.


Betree:
Ah my bad, this is API

PR #4336, opened 2020-08-06
WIP: request funding expense and open collective payout method

znarf:


server/models/PayoutMethod.ts
```diff
@@ -12,6 +12,7 @@ export enum PayoutMethodTypes {
   OTHER = 'OTHER',
   PAYPAL = 'PAYPAL',
   BANK_ACCOUNT = 'BANK_ACCOUNT',
+  OPEN_COLLECTIVE = 'OPEN_COLLECTIVE',
```
Betree, line 15: That would be clearer to me

```suggestion
  COLLECTIVE_BALANCE = 'COLLECTIVE_BALANCE',
```

Betree: Or to fit our new GQLV2 namings

```suggestion
  ACCOUNT_BALANCE = 'ACCOUNT_BALANCE',
```

sbinlondon: > That would be clearer to me

I agree that `COLLECTIVE_BALANCE` is clearer, and I prefer that to `ACCOUNT_BALANCE`. But then I guess orgs and other types of 'collective' can have this as well?

znarf: Yeah, we have Events, Funds, Projects and also Organizations/Hosts. So, `COLLECTIVE_BALANCE` is not really possible.

`OPEN_COLLECTIVE_BALANCE` is a bit verbose but could be an alternative.

kewitz: I'm OK with using simply `BALANCE`, I think it's obvious what this implies and I don't think it could mean something else.
Also, as of now, the Account abstraction only exists from the GraphQL layer above, I think it is interesting to keep it this way or come up with a plan to rename our models and tables from "collective" to "account".

PR #4342, opened 2020-08-07
Host: Use the new flag for custom fees

Betree:
Require https://github.com/opencollective/opencollective-frontend/pull/4770

This will make sure we properly differentiate collectives with custom fees and that we don't override their custom values when changing the global host fee.

### DIFF
--- a/migrations/20200807083335-add-collectives-data-use-custom-host-fee.js
+++ b/migrations/20200807083335-add-collectives-data-use-custom-host-fee.js
@@ -1,0 +1,32 @@
+'use strict';
+
+module.exports = {
+  /**
+   * Set "Collectives" -> "data" -> "useCustomHostFee" to true when their hostFeePercent
+   * is different from the one of the host.
+   */
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(`
+      UPDATE ONLY "Collectives" c
+      SET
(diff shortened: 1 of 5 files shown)


migrations/20200807083335-add-collectives-data-use-custom-host-fee.js
```diff
@@ -0,0 +1,32 @@
...
+   * is different from the one of the host.
+   */
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(`
+      UPDATE ONLY "Collectives" c
+      SET
+        data = (
+          CASE WHEN c.data IS NULL 
+          THEN '{"useCustomHostFee": true}'::jsonb
+          ELSE c.data::jsonb || '{"useCustomHostFee": true}'::jsonb
```
kewitz, line 15: Is this else condition equivalent to this?
```
jsonb_set(data, '{useCustomHostFee}', 'true')
```

Betree: Does Jsonb_set creates the Jason object if the column is null ?

kewitz: I don't think it does. But I'm referring just to the ELSE condition here.

PR #4343, opened 2020-08-07
Host supported payment methods

sbinlondon:
For the new Contribution Flow, also addresses opencollective/opencollective#3159

### DIFF
--- a/server/graphql/v2/object/Host.js
+++ b/server/graphql/v2/object/Host.js
@@ -1,6 +1,7 @@
-import { GraphQLBoolean, GraphQLInt, GraphQLNonNull, GraphQLObjectType } from 'graphql';
-import { get } from 'lodash';
+import { GraphQLBoolean, GraphQLInt, GraphQLList, GraphQLNonNull, GraphQLObjectType } from 'graphql';
+import { find, get } from 'lodash';
 
+import { PaymentMethodType } from '../enum';
 import { Account, AccountFields } from '../interface/Account';
 import URL from '../scalar/URL';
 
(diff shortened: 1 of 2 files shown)


server/graphql/v2/object/Host.js
```diff
@@ -43,6 +44,32 @@ export const Host = new GraphQLObjectType({
...
       },
+      supportedPaymentMethods: {
+        type: new GraphQLList(PaymentMethodType),
+        description:
+          'The list of payment methods (Stripe, Paypal, manual bank transfer, etc ...) the Host can accept for its Collectives',
+        async resolve(collective, _, req) {
+          const supportedPaymentMethods = [];
+
+          // Paypal, Stripe = connected accounts
+          const connectedAccounts = await req.loaders.Collective.connectedAccounts.load(collective.id);
```
znarf, line 55: Nice that you used a loader here.

sbinlondon:
> I don't think there is a need to add other payment methods by default (`COLLECTIVE_BALANCE`, `PREPAID_BUDGET`, `GIFT_CARD`) but feel free to add them if's making things more easy and consistent on the frontend.

Are there any settings on the host that would preclude these payment methods, or is that generally down to the Collective type, balance, etc (not involving the host)

znarf:
- `COLLECTIVE_BALANCE` should only be available on the same host (for now, will change later)
- `PREPAID_BUDGET` is normally limited to a specific Host (see `data.HostCollectiveId`)
- `GIFT_CARD` can be limited to a specific Host (see `limitedToHostCollectiveIds`)

At the Host level, they should all be available in as supported payment methods (explicitely or implicitely).

They should be filtered in `stepPayment`, for some of these, you can have a look at prior implementation.

PR #4355, opened 2020-08-10
Add "limited to" ids for PaymentMethod v2

sbinlondon:
* Adds `HostCollectiveId` to `data` on `prepaid` payment methods
* Adds `limitedToCollectiveIds` and `limitedToHostCollectiveIds` for payment methods on a `Host`

### DIFF
--- a/server/graphql/v2/object/PaymentMethod.js
+++ b/server/graphql/v2/object/PaymentMethod.js
@@ -1,10 +1,11 @@
-import { GraphQLInt, GraphQLObjectType, GraphQLString } from 'graphql';
+import { GraphQLInt, GraphQLList, GraphQLObjectType, GraphQLString } from 'graphql';
 import GraphQLJSON from 'graphql-type-json';
-import { pick } from 'lodash';
+import { get, pick } from 'lodash';
 
 import { idEncode } from '../identifiers';
 import { Account } from '../interface/Account';
 import { Amount } from '../object/Amount';
+import { Host } from '../object/Host';
 
 export const PaymentMethod = new GraphQLObjectType({
   name: 'PaymentMethod',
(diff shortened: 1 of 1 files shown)


server/graphql/v2/object/PaymentMethod.js
```diff
@@ -62,13 +62,42 @@ export const PaymentMethod = new GraphQLObjectType({
...
+          let encodedIds;
+          if (paymentMethod.limitedToCollectiveIds !== null) {
+            encodedIds = paymentMethod.limitedToCollectiveIds.map(id => {
+              return idEncode(id, 'account');
+            });
+          }
+          return encodedIds;
+        },
+      },
+      limitedToHostCollectiveIds: {
```
znarf, line 89: To follow the direction of GraphQL API v2, this should return a List of `Host`s, not ids.

Frontend side GraphQL query would look something like:

```
paymentMethods {
  id
  limitedToHosts {
   id
   legacyId
   slug
  }
}
```

server/graphql/v2/object/PaymentMethod.js
```diff
@@ -62,13 +62,42 @@ export const PaymentMethod = new GraphQLObjectType({
             return pick(paymentMethod.data, ['email']);
           }
 
+          // Return HostCollectiveId for prepaid
+          if (paymentMethod.type === 'prepaid') {
+            return pick(paymentMethod.data, ['HostCollectiveId']);
+          }
```
znarf, line 68: I think we should expose that in `limitedToHostCollectiveIds` / `limitedToHosts`. Would be more consistent and better structured.

server/graphql/v2/object/PaymentMethod.js
```diff
@@ -62,13 +62,42 @@ export const PaymentMethod = new GraphQLObjectType({
...
+
           const data = paymentMethod.data;
           // white list fields to send back; removes fields like CustomerIdForHost
           const dataSubset = pick(data, ['fullName', 'expMonth', 'expYear', 'brand', 'country', 'last4']);
 
           return dataSubset;
         },
       },
+      limitedToCollectiveIds: {
```
znarf, line 77: I just reviewed this one and I'm wondering if we need to keep supporting it. I just checked and the API support for this property is flawed, proving that it's not much used. Decision tomorrow.

In case we decided to keep it, similar to my comment below in `limitedToHostCollectiveIds`, it should be exposed as a list of `Account`s.

Betree: As discussed, I'm removing this field in https://github.com/opencollective/opencollective-api/pull/4361

server/graphql/v2/object/PaymentMethod.js
```diff
@@ -69,6 +70,25 @@ export const PaymentMethod = new GraphQLObjectType({
...
+            const hostId = get(paymentMethod, 'data.HostCollectiveId', null);
+            if (!hostId) {
+              return;
+            }
+            const host = await req.loaders.Collective.byId.load(hostId);
+            hosts = [host];
+          } else if (paymentMethod.type === 'virtualcard' && paymentMethod.limitedToHostCollectiveIds) {
+            hosts = paymentMethod.limitedToHostCollectiveIds.map(id => {
+              return req.loaders.Collective.byId.load(id);
+            });
```
Betree, line 87: Not a request change, but this can be simplified with the `loadMany` helper

```es6
hosts = req.loaders.Collective.byId.loadMany(paymentMethod.limitedToHostCollectiveIds);
```

PR #4367, opened 2020-08-12
Fix inconsistent 2FA sign-in test

kewitz:
Related to https://github.com/opencollective/opencollective/issues/3347

I increase the time-frame our 2FA validator considers to make sure it still works when the time-frame moves 2 steps.
It is common to consider more than one window when using TOTP.

### DIFF
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -100,6 +100,7 @@ export const updateToken = async (req, res, next) => {
       secret: decryptedTwoFactorAuthToken,
       encoding: 'base32',
       token: twoFactorAuthenticatorCode,
+      window: 2,
     });
     if (!verified) {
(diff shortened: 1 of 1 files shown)


kewitz:
@znarf the signin with 2FA one:
```
1) signin with 2FA
       can signin with 2fa enabled:
     AssertionError: Timed out retrying: Expected to find element: `.LoginTopBarProfileButton-name`, but never found it.
      at Context.eval (http://localhost:3000/__cypress/tests?p=test/cypress/support/index.js:33568:27)
```

Mr1262:
> Related to [opencollective/opencollective#3347](https://github.com/opencollective/opencollective/issues/3347)
> 
> I increase the time-frame our 2FA validator considers to make sure it still works when the time-frame moves 2 steps.
> It is common to consider more than one window when using TOTP.

PR #4373, opened 2020-08-13
Add longDescription to expense model

znarf:
The idea is to have a clean generic column to contain a longer description, in the case of "Funding Request" the message sent while sending the request.

### DIFF
--- a/migrations/20200813085021-add-expense-longDescription.js
+++ b/migrations/20200813085021-add-expense-longDescription.js
@@ -1,0 +1,17 @@
+'use strict';
+
+module.exports = {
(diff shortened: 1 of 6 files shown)


znarf:
@Betree does that change something on the API / Model side to make it a HTML field?

Betree:
@znarf The only change on the API side is to add a custom setter like in https://github.com/opencollective/opencollective-api/blob/48d7f329130ff94afbb73e01786e4661aa72ac32/server/models/Expense.js#L132-L138

The rest of the changes are frontend only.

PR #4375, opened 2020-08-13
GQLV2: Add createOrder endpoint

Betree:
Followup issues:
- Remove `type`/`service` from GQLV2/`PaymentMethod` and rename `providerType`
- Rename `OrderFrequency` to `ContributionFrequency`


server/graphql/v2/input/OrderCreateInput.ts
```diff
@@ -0,0 +1,56 @@
...
+      description: 'The contribution amount for 1 quantity, without platform contribution and taxes',
+    },
+    frequency: {
+      type: new GraphQLNonNull(OrderFrequency),
+    },
+    fromAccount: {
+      type: new GraphQLNonNull(AccountReferenceInput),
+      description: 'The profile making the order',
+    },
+    account: {
```
znarf, line 31: In other places, it's named `toAccount`

server/graphql/v2/enum/PaymentMethodType.ts
```diff
@@ -0,0 +1,62 @@
...
+export const getPaymentMethodType = ({ service, type }: PaymentMethod): PaymentMethodTypeEnum => {
+  if (service === PAYMENT_METHOD_SERVICE.STRIPE) {
+    if (type === PAYMENT_METHOD_TYPE.CREDITCARD) {
+      return PaymentMethodTypeEnum.CREDIT_CARD;
+    }
+  } else if (service === PAYMENT_METHOD_SERVICE.OPENCOLLECTIVE) {
+    if (type === PAYMENT_METHOD_TYPE.VIRTUALCARD) {
+      return PaymentMethodTypeEnum.GIFT_CARD;
+    } else if (type === PAYMENT_METHOD_TYPE.COLLECTIVE) {
+      return PaymentMethodTypeEnum.COLLECTIVE_BALANCE;
```
znarf, line 32: `ACCOUNT_BALANCE` to match `PayoutMethod` and to hint that it's not only for Collectives anymore.

server/graphql/v2/input/PaymentMethodInput.ts
```diff
@@ -0,0 +1,37 @@
...
+    },
+    save: {
+      type: GraphQLBoolean,
+      description: 'Wether this payment method should be saved for future payments',
+    },
+    creditCardInfo: {
+      type: CreditCardCreateInput,
+      description: 'When creating a credit card, use this field to set its info',
+    },
+    paypalData: {
```
znarf, line 32: One is `info`, one is `data`, I get their purpose is a bit different but looks strange.

server/graphql/v2/input/PaymentMethodInput.ts
```diff
@@ -0,0 +1,37 @@
...
+    },
+    type: {
+      type: PaymentMethodType,
+      description: 'Type of this payment method',
+    },
+    name: {
+      type: GraphQLString,
+      description: 'Name of this payment method',
+    },
+    save: {
```
znarf, line 24: Maybe we could take the opportunity to make this one more explicit, ie: `saveForLater`.

server/graphql/v2/input/PaypalPaymentDataInput.ts
```diff
@@ -0,0 +1,10 @@
+import { GraphQLInputObjectType, GraphQLNonNull, GraphQLString } from 'graphql';
+import GraphQLJSON from 'graphql-type-json';
+
+export const PaypalPaymentDataInput = new GraphQLInputObjectType({
+  name: 'PaypalPaymentDataInput',
+  fields: {
+    token: { type: new GraphQLNonNull(GraphQLString) },
+    data: { type: GraphQLJSON },
```
znarf, line 8: Something named `data` should ideally not contains something else named `data`.

server/graphql/v2/object/PaymentMethod.js
```diff
@@ -31,9 +33,16 @@ export const PaymentMethod = new GraphQLObjectType({
       },
       service: {
         type: GraphQLString,
+        deprecationReason: '2020-08-18: This field is being deprecated in favor of providerType',
       },
       type: {
         type: GraphQLString,
+        deprecationReason: '2020-08-18: This field is being deprecated in favor of providerType',
+      },
+      providerType: {
```
znarf, line 42: maybe `paymentMethodType` instead if it's temporary

Betree: I hesitated between the two (tho this field should really be called `type`) and went with `providerType` because it's not used in our codebase today (= easier to refactor) and is, in my opinion, pretty explicit about what it is: the type of the payment method provider.

server/graphql/v2/input/PaymentMethodInput.ts
Betree: I'll rename this one to `paypalInfo` for consistency

server/graphql/v2/input/PaymentMethodInput.ts
Betree: The field is not yet defined on `PaymentMethod`, so now is a good time to rename it if we want to. I personally like to prefix booleans with `is`/`has`, so `isSavedForLater`.

scripts/encrypt.js
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env ./node_modules/.bin/babel-node
```
Betree, line 1: Added that to easily encrypt tokens, which helps to setup PayPal. Updated the docs accordingly on https://docs.opencollective.com/help/contributing/development/paypal#developing-testing-sandbox

cc @kewitz

PR #4376, opened 2020-08-13
Proportionally split the payment processor fees when collecting fees on top

kewitz:


Mr1262:
fix/fees-on-top-processor-fees

server/models/Transaction.js
```diff
@@ -498,6 +505,8 @@ export default (Sequelize, DataTypes) => {
     await Transaction.createDoubleEntry(donationTransaction);
 
     // Remove fees from main transaction
+    transaction.paymentProcessorFeeInHostCurrency =
+      transaction.paymentProcessorFeeInHostCurrency - paymentProcessorFeeInHostCurrency;
```
znarf, line 509: `paymentProcessorFeeInHostCurrency - paymentProcessorFeeInHostCurrency;` is difficult to read.

Can you consider renaming variables?

kewitz:
@Mr1262 ?

znarf:
I'm good with it, would be great to have a second review from @Betree.

server/models/Transaction.js
```diff
@@ -476,6 +476,13 @@ export default (Sequelize, DataTypes) => {
       return;
     }
 
+    // Calculate the paymentProcessorFee proportional to the feeOnTop amount
+    const feeOnTopPaymentProcessorFee = toNegative(
+      Math.round(
+        transaction.paymentProcessorFeeInHostCurrency *
+          Math.abs(transaction.platformFeeInHostCurrency / transaction.amountInHostCurrency),
+      ),
+    );
```
Betree, line 485: When doing math in code, I personally find that using as many variables as possibles improves clarity by naming things.  But that's a personal preference

```suggestion
    const feeOnTopPercent = Math.abs(transaction.platformFeeInHostCurrency / transaction.amountInHostCurrency);
    const feeOnTopPaymentProcessorFee = toNegative(
      Math.round(transaction.paymentProcessorFeeInHostCurrency * feeOnTopPercent),
    );
```

PR #4384, opened 2020-08-14
Update public message v2

sbinlondon:


server/middleware/authentication.js
```diff
@@ -359,6 +359,9 @@ export function authorizeClientApp(req, res, next) {
  * Makes sure that the user the is logged in and req.remoteUser is populated.
  * if we cannot authenticate the user, we directly return an Unauthorized error.
  */
+
+// modify to check that the token is acually a login link and not just pasted into local storage
+//
```
znarf, line 364: You might want to remove that.

server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -0,0 +1,54 @@
...
+import { NotFound, Unauthorized } from '../../errors';
+import { idDecode } from '../identifiers';
+import { Member } from '../object/Member';
+
+const memberMutations = {
+  editPublicMessage: {
+    type: new GraphQLNonNull(Member),
+    description: 'Edit the public message for the given Member of a Collective',
+    args: {
+      FromCollectiveLegacyId: {
```
znarf, line 13: Use a reference.

server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -0,0 +1,54 @@
...
+        type: GraphQLInt,
+        description: 'Legacy id for the donating Collective',
+      },
+      ToCollectiveId: {
```
znarf, line 17: Also use a reference here.

server/middleware/authentication.js
sbinlondon: 🙈

server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -0,0 +1,56 @@
...
+import { idDecode } from '../identifiers';
+import { AccountReferenceInput } from '../input/AccountReferenceInput';
+import { Member } from '../object/Member';
+
+const memberMutations = {
+  editPublicMessage: {
+    type: new GraphQLNonNull(Member),
+    description: 'Edit the public message for the given Member of a Collective',
+    args: {
+      FromCollectiveLegacyId: {
```
Betree, line 14: Should be `fromAccount`, and type should be non-nullable.

Also `description` doesn't match the `type`.

server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -0,0 +1,56 @@
...
+        type: AccountReferenceInput,
+        description: 'Legacy id for the donating Collective',
+      },
+      ToCollectiveId: {
```
Betree, line 18: Should be `toAccount`, and type should be non-nullable.

Also `description` doesn't match the `type`.

server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -0,0 +1,56 @@
...
+      },
+    },
+    async resolve(_, args, req) {
+      const { FromCollectiveLegacyId, ToCollectiveId, message } = args;
+
+      if (!req.remoteUser || !req.remoteUser.isAdmin(FromCollectiveLegacyId)) {
+        throw new Unauthorized("You don't have the permission to edit this member's public message");
+      }
+
+      const decodedToCollectiveId = idDecode(ToCollectiveId.id, 'account');
```
Betree, line 34: Use `fetchAccountWithReference` to retrieve an account from a reference

server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -0,0 +1,56 @@
...
+
+      if (!member) {
+        throw new NotFound('No member found');
+      }
+
+      await member.update({
+        publicMessage: message,
+      });
+
+      return member;
```
Betree, line 51: You can directly write

```suggestion
      return member.update({ publicMessage: message });
```

server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -0,0 +1,56 @@
...
+      },
+      message: {
+        type: GraphQLString,
+        description: 'New public message',
+      },
+    },
+    async resolve(_, args, req) {
+      const { FromCollectiveLegacyId, ToCollectiveId, message } = args;
+
+      if (!req.remoteUser || !req.remoteUser.isAdmin(FromCollectiveLegacyId)) {
```
Betree, line 30: `FromCollectiveLegacyId` is an account reference, doing `req.remoteUser.isAdmin(FromCollectiveLegacyId)` won't work

server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -0,0 +1,57 @@
...
+const memberMutations = {
+  editPublicMessage: {
+    type: new GraphQLNonNull(Member),
+    description: 'Edit the public message for the given Member of a Collective',
+    args: {
+      FromAccount: {
+        type: GraphQLNonNull(AccountReferenceInput),
+        description: 'Reference to an account for the donating Collective',
+      },
+      ToAccount: {
```
Betree, line 17: In other mutations for GQLV2, args are lowercase (ie https://github.com/opencollective/opencollective-api/blob/32f1ded723fbf354677c0fbca0eec3950d66e5e6/server/graphql/v2/mutation/AccountMutations.ts#L19). I think we should be consistent on this.

```suggestion
      fromAccount: {
        type: GraphQLNonNull(AccountReferenceInput),
        description: 'Reference to an account for the donating Collective',
      },
      toAccount: {
```

server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -0,0 +1,57 @@
...
+        throw new Unauthorized("You don't have the permission to edit this member's public message");
+      }
+
+      if (!decodedToAccount) {
+        throw new NotFound('Account not found');
+      }
+
+      const member = await models.Member.findOne({
+        where: {
+          MemberCollectiveId: FromAccount.legacyId,
```
Betree, line 41: What if someone uses this mutation by passing an `id` rather than a `legacyId` in `FromAccount`?

We need to do the same `fetchAccountWithReference` for `fromAccount`.

server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -0,0 +1,57 @@
...
+      },
+      message: {
+        type: GraphQLString,
+        description: 'New public message',
+      },
+    },
+    async resolve(_, args, req) {
+      const { FromAccount, ToAccount, message } = args;
+
+      const decodedToAccount = await fetchAccountWithReference(ToAccount);
```
Betree, line 29: Not a requested change, but the return of `fetchAccountWithReference` is a `models.Collective` so the `decoded` in variable name doesn't seem necessary to me.

```suggestion
      const toAccount = await fetchAccountWithReference(ToAccount);
```

PR #4392, opened 2020-08-17
Validate slugs do not start nor ends with dashes

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3356

- [ ] DO NOT MERGE THIS BEFORE FIXING EXISTING SLUGS

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -18,6 +18,7 @@ import {
   pickBy,
   sum,
   sumBy,
+  trim,
   uniqBy,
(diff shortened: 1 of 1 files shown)


znarf:
Have we already done the migration? 

Otherwise, this can of validation might prevent people from updating their Collective data.

kewitz:
@znarf, not yet, I plan to manually migrate this.

znarf:
Ok for merge then if you're aware that you need to do these migrations first.

PR #4393, opened 2020-08-17
Update 2FA test and remove unused package from repo

sbinlondon:
Found in the frontend tests that I can just use `speakeasy` rather than `otpauth` for the tests, which is the only reason it was installed. So have updated the tests here and removed the package.

### DIFF
--- a/test/server/routes/users.test.js
+++ b/test/server/routes/users.test.js
@@ -3,7 +3,6 @@ import config from 'config';
 import crypto from 'crypto-js';
 import moment from 'moment';
 import nodemailer from 'nodemailer';
-import { Secret, TOTP } from 'otpauth';
 import sinon from 'sinon';
 import speakeasy from 'speakeasy';
 import request from 'supertest';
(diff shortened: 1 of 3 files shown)


test/server/routes/users.test.js
```diff
@@ -145,12 +144,11 @@ describe('server/routes/users', () => {
...
       const currentToken = user.jwt();
-      const twoFactorAuthenticatorCode = new TOTP({
+      const twoFactorAuthenticatorCode = speakeasy.totp({
         algorithm: 'SHA1',
-        digits: 6,
-        period: 30,
-        secret: Secret.fromB32(secret.base32),
-      }).generate();
+        encoding: 'base32',
+        secret: secret.base32,
```
znarf, line 150: Should we use `window` here?

sbinlondon: It isn't a param for `speakeasy.totp`. It's a param on `totp.verify` to allow a bit of wiggle room when you're decrypting. The thing you set here on creating the TOTP would be 'period' which specifies how big the window is. But you don't set the amount of windows you can have on decryption.

sbinlondon:
@znarf Ok to merge?

test/server/routes/users.test.js
znarf: Ok, make sense.

It's not clear however using "decrypting" and "decryption" in this context (unless I'm wrong). We're just generating a token (and later verifying it), there is no encryption in the process (out of the DB encryption which is something separate).

znarf (approved):

PR #4394, opened 2020-08-17
Add 2FA for large payouts on host dashboard

sbinlondon:
If:

* the amount of the payment is over $1000
* the method is Paypal
* and the user has 2FA enabled on their account

Then we send back an "error" asking them to input their 2FA code. If it matches, we then continue on through the rest of the payout function.

Starting out just adding it to Paypal, can also add it to Transferwise.


server/graphql/v1/mutations/expenses.js
```diff
@@ -671,6 +673,27 @@ export async function payExpense(req, args) {
     );
   }
 
+  // If the amount of the expense is larger than $1000 enforce 2FA if the user has it turned on
+  if (payoutMethodType === PayoutMethodTypes.PAYPAL && !args.forceManual && expense.amount > 100000) {
```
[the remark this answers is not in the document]
sbinlondon: Good shout, will add these changes

sbinlondon:
> Looks good, would be great to export this to a separate function in another file as we'll probably want to protect other endpoints in the future.

Yes, I was also thinking of extracting the 2FA token decryption into our crypto lib as it's now used in `AccountMutations` and also here?

server/graphql/v1/mutations/expenses.js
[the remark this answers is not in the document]
sbinlondon: - [x] activate for Transferwise
- [x] enforce if it's in host settings
- [x] put `expenseAmount` in configuration

server/graphql/v1/mutations/expenses.js
```diff
@@ -671,6 +672,20 @@ export async function payExpense(req, args) {
     );
   }
 
+  // If the amount of the expense is larger than $1000 enforce 2FA if the user has it turned on
+  const acceptedPayoutMethods =
```
znarf, line 676: Can you use a more self explaining variable names? This is a bit too generic.

server/controllers/users.js
```diff
@@ -94,14 +92,7 @@ export const updateToken = async (req, res, next) => {
...
   if (twoFactorAuthenticatorCode) {
-    const encryptedTwoFactorAuthToken = req.remoteUser.twoFactorAuthToken;
-    const decryptedTwoFactorAuthToken = crypto.decrypt(encryptedTwoFactorAuthToken);
-    const verified = speakeasy.totp.verify({
-      secret: decryptedTwoFactorAuthToken,
-      encoding: 'base32',
-      token: twoFactorAuthenticatorCode,
-      window: 2,
-    });
+    const verified = auth.verifyTwoFactorAuthCode(req.remoteUser.twoFactorAuthToken, twoFactorAuthenticatorCode);
```
znarf, line 95: Could this be `verifyTwoFactorAuthenticatorCode` to match the variable name?

server/graphql/v1/mutations/expenses.js
```diff
@@ -671,6 +672,20 @@ export async function payExpense(req, args) {
...
+    payoutMethodType === PayoutMethodTypes.PAYPAL || payoutMethodType === PayoutMethodTypes.BANK_ACCOUNT;
+  const hostHasPayout2FAEnabled = get(host, 'settings.payoutsTwoFactorAuth.enabled', null);
+  const hostPayout2FAExpenseAmount = get(host, 'settings.payoutsTwoFactorAuth.expenseAmount', null);
+  if (
+    acceptedPayoutMethods &&
+    !args.forceManual &&
+    hostHasPayout2FAEnabled &&
+    expense.amount >= hostPayout2FAExpenseAmount
+  ) {
+    enforceTwoFactorAuthOnPayouts(req, args.twoFactorAuthenticatorCode);
```
znarf, line 686: Even if it's verbose, avoiding abbreviations is generally preferable: `enforceTwoFactorAuthenticationOnPayouts`

server/graphql/v1/mutations/expenses.js
```diff
@@ -671,6 +672,20 @@ export async function payExpense(req, args) {
     );
   }
 
+  // If the amount of the expense is larger than $1000 enforce 2FA if the user has it turned on
+  const acceptedPayoutMethods =
+    payoutMethodType === PayoutMethodTypes.PAYPAL || payoutMethodType === PayoutMethodTypes.BANK_ACCOUNT;
+  const hostHasPayout2FAEnabled = get(host, 'settings.payoutsTwoFactorAuth.enabled', null);
+  const hostPayout2FAExpenseAmount = get(host, 'settings.payoutsTwoFactorAuth.expenseAmount', null);
```
znarf, line 679: Default should be a number, 1000?

server/graphql/v1/mutations/expenses.js
```diff
@@ -671,6 +672,20 @@ export async function payExpense(req, args) {
     );
   }
 
+  // If the amount of the expense is larger than $1000 enforce 2FA if the user has it turned on
+  const acceptedPayoutMethods =
+    payoutMethodType === PayoutMethodTypes.PAYPAL || payoutMethodType === PayoutMethodTypes.BANK_ACCOUNT;
+  const hostHasPayout2FAEnabled = get(host, 'settings.payoutsTwoFactorAuth.enabled', null);
```
znarf, line 678: That will not change the behavior, but default should be a boolean here,  `false`.

server/graphql/v1/mutations/expenses.js
```diff
@@ -671,6 +672,20 @@ export async function payExpense(req, args) {
     );
   }
 
+  // If the amount of the expense is larger than $1000 enforce 2FA if the user has it turned on
+  const acceptedPayoutMethods =
+    payoutMethodType === PayoutMethodTypes.PAYPAL || payoutMethodType === PayoutMethodTypes.BANK_ACCOUNT;
```
znarf, line 677: Optional (style):

```suggestion
    [PayoutMethodTypes.PAYPAL, PayoutMethodTypes.BANK_ACCOUNT].includes(payoutMethodType);
```

server/graphql/v1/mutations/expenses.js
```diff
@@ -671,6 +672,22 @@ export async function payExpense(req, args) {
     );
   }
 
+  // If the amount of the expense is larger than i.e. $1000 enforce 2FA if the user has it turned on
+  const payoutMethodsToEnforceTwoFactorAuthenticationOn = [
```
Betree, line 676: This would be a good variable name for the array itself, but here I think we can simplify as `is2FARequiredForPayoutMethod`

Betree (commented):
This looks good.

Two things for a followup:
- On the UX side, asking for your 2FA for each expense is a bit extreme. If you already validated for another expense, your machine should be marked as trusted (like for 1h?). It's quite similar to Github, where you need to re-enter your password to edit sensitive information (like permissions) but once you've entered it you don't have to re-enter it again for a few hours.
- I don't think that enforcing 2FA only above a minimum amount is a good idea. If someone hacks into a host account, it's not complicated to submit multiple small expenses (especially with a script). That will totally bypass this security measure, and in the end attackers can steal the same amount.

sbinlondon:
> On the UX side, asking for your 2FA for each expense is a bit extreme.

Yep that is the next step. This was just proof of concept/step 1 for 2FA on payouts.

> I don't think that enforcing 2FA only above a minimum amount is a good idea. If someone hacks into a host account, it's not complicated to submit multiple small expenses (especially with a script). That will totally bypass this security measure, and in the end attackers can steal the same amount.

That's a really good point. I'm not sure what the solution would be - just random 2FA checks? What do you think @znarf?

znarf:
> On the UX side, asking for your 2FA for each expense is a bit extreme.

This is why we only do it on some, and decided to look at the amount to filter. When I do transfers on my personal and business bank accounts, I'm asked to enter a 2FA every single time so it's not completely crazy.

> I don't think that enforcing 2FA only above a minimum amount is a good idea.

While I understand your point, I'm confident this is a good start. We can improve and introduce more complex heuristics overtime. We could sum all the amounts in our Redis and use this to counter the problem of multiple expenses.

>  If you already validated for another expense, your machine should be marked as trusted (like for 1h?).

It's a good idea, we thought about it but don't have the technical solution at the moment.

sbinlondon:
> If you already validated for another expense, your machine should be marked as trusted (like for 1h?).

Totally didn't finish my sentence in the comment above 🤦‍♀️  Meant to say that was what we were going to add next, a 'remember for 30 days' type thing. But as @znarf mentioned we haven't quite got around to planning how to do that yet. A bit more research required 👍

Betree:
> When I do transfers on my personal and business bank accounts, I'm asked to enter a 2FA every single time so it's not completely crazy.

It's not the same use case, you usually do a few bank transfers at the same time while a host admin batches the payments and can do dozens in a few minutes.

PR #4404, opened 2020-08-18
Fix Monthly Platform Report

kewitz:
The problem is that we were missing platform fees from bank transfers.
image

### DIFF
--- a/reports/platform-report.js
+++ b/reports/platform-report.js
@@ -20,14 +20,13 @@ const getPlatformRevenue = async (startDate, endDate) => {
   const res = await query(
     `
   SELECT
-    -- In the case this is a fees on top transaction, the currency will already be USD
     currency,
     SUM(
       CASE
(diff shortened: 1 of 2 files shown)


znarf:
Doesn't seem to be it. No reason to have "Fees On Top" in "Manually Added Funds" right?

Screenshot 2020-08-19 at 10 58 09

kewitz:
@znarf yep, this is actually related to bank transfers but I'm still investigating. I want to make these queries are MECE.

znarf:
@kewitz No platform fees for Bank Transfers?

kewitz:
@znarf good catch.
We recently had a small discussion on our channel about that with Pia, we're currently not collecting those.
Should I include it anyway in this report?

znarf:
@kewitz We have one significant transaction in July for the OSC host (id: 493199) and I'm working on the long term implementation.

Yes, we need to track it.

PR #4410, opened 2020-08-20
Adjust host currency and amount in oppositeTransaction

znarf:
Follow up to https://github.com/opencollective/opencollective-api/pull/4322


server/models/Transaction.js
```diff
@@ -437,11 +439,22 @@ export default (Sequelize, DataTypes) => {
...
         ...oppositeTransaction,
-        HostCollectiveId: fromCollective.HostCollectiveId,
+        HostCollectiveId: fromCollectiveHost.id,
         currency: fromCollective.currency,
+        hostCurrency: fromCollectiveHost.currency,
+        hostCurrencyFxRate: oppositeTransactionHostCurrencyFxRate,
         amount: -Math.round(transaction.netAmountInCollectiveCurrency * oppositeTransactionCurrencyFxRate),
+        amountInHostCurrency: -Math.round(
+          transaction.netAmountInCollectiveCurrency * oppositeTransactionHostCurrencyFxRate,
+        ),
```
znarf, line 457: TODO: what about `*feeInHostCurrency`, what should we do?

kewitz: If we don't convert, they'll be in the original host currency, right?
I think it would be interesting to convert to the oppositeTransaction's host currency, it will be useful later on when we start to collect fees from the host.

server/models/Transaction.js
```diff
@@ -424,7 +425,8 @@ export default (Sequelize, DataTypes) => {
       platformFeeInHostCurrency: transaction.platformFeeInHostCurrency,
       paymentProcessorFeeInHostCurrency: transaction.paymentProcessorFeeInHostCurrency,
     };
-    if (!fromCollective.isActive) {
+
+    if (!fromCollective.isActive || !fromCollectiveHost) {
```
kewitz, line 429: :+1: `fromCollectiveHost` can be null.

PR #4412, opened 2020-08-20
Add supportedPayoutMethods to Host resolver

kewitz:
Related to https://github.com/opencollective/opencollective-frontend/pull/4877


znarf (changes requested):
Looks great.

Missing is `ACCOUNT_BALANCE`, which should be available for all Hosts, but not for all Payees.

PR #4422, opened 2020-08-21
Dashboard info card endpoints

Betree:
For https://github.com/opencollective/opencollective-frontend/pull/4880

- Host: Balances for expense payment methods. To implement that, I've made `AccountStats` an interface specialized by `HostStats`.

### DIFF
--- a/server/paymentProviders/paypal/index.js
+++ b/server/paymentProviders/paypal/index.js
@@ -125,23 +125,23 @@ export default {
             );
           }
 
+          const redirectUrl = new URL(paymentMethod.data.redirect);
+          redirectUrl.searchParams.set('paypalApprovalStatus', req.query.paypalApprovalStatus);
+
           if (req.query.paypalApprovalStatus !== 'success') {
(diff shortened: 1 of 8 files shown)


server/paymentProviders/paypal/index.js
```diff
@@ -125,23 +125,23 @@ export default {
...
+            return res.redirect(redirectUrl.toString());
           }
 
           return (
             getPreapprovalDetailsAndUpdatePaymentMethod(pm)
               .catch(e => {
                 debugPaypal('>>> paypal callback error:', e);
-                const redirect = `${
-                  paymentMethod.data.redirect
-                }?status=error&service=paypal&error=Error%20while%20contacting%20PayPal&errorMessage=${encodeURIComponent(
```
Betree, line 140: All these statuses were not used in the frontend (so it should not be breaking), and the way we generated the URL was breaking if the redirect already contains filter parameters (which the new page does for filtering).

PR #4423, opened 2020-08-21
Refund feesOnTop

kewitz:
Also refacts createRefundTransaction to include `associateTransactionRefundId`.

Since the stripe charge carries the full amount and the transaction split is something that exists only on our side, refunding the feeOnTop transaction works fine.
I suggest using https://github.com/opencollective/opencollective-frontend/pull/4897 when testing.

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -144,7 +144,7 @@ export function calcFee(amount, fee) {
 export async function createRefundTransaction(transaction, refundedPaymentProcessorFee, data, user) {
   /* If the transaction passed isn't the one from the collective
    * perspective, the opposite transaction is retrieved. */
-  const collectiveLedger =
+  const creditTransaction =
     transaction.type === 'CREDIT'
       ? transaction
       : await models.Transaction.findOne({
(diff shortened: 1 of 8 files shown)


server/lib/payments.js
```diff
@@ -144,7 +144,7 @@ export function calcFee(amount, fee) {
 export async function createRefundTransaction(transaction, refundedPaymentProcessorFee, data, user) {
   /* If the transaction passed isn't the one from the collective
    * perspective, the opposite transaction is retrieved. */
-  const collectiveLedger =
```
Betree, line 147: :+1:

server/lib/payments.js
```diff
@@ -154,49 +154,65 @@ export async function createRefundTransaction(transaction, refundedPaymentProces
...
 
-  /* Amount fields. Must be calculated after tweaking all the fees */
-  userLedgerRefund.amount = -collectiveLedger.amount;
-  userLedgerRefund.amountInHostCurrency = -collectiveLedger.amountInHostCurrency;
-  userLedgerRefund.netAmountInCollectiveCurrency = -libtransactions.netAmount(collectiveLedger);
+  const creditTransactionRefund = buildRefund(creditTransaction);
+
+  if (transaction.data?.isFeesOnTop) {
+    const feeOnTopTransaction = await models.Transaction.findOne({
+      where: { ...FEES_ON_TOP_TRANSACTION_PROPERTIES, type: 'CREDIT' },
```
Betree, line 207: This `FEES_ON_TOP_TRANSACTION_PROPERTIES` export is pretty handy. You may be interested in Sequelize [scopes](https://sequelize.org/master/manual/scopes.html) for similar needs in the future.

Betree (approved):
Great simplification with the variables names and `associateTransactionRefundId`.

PR #4424, opened 2020-08-23
fix(Collective): prefer name over twitter handle for slug generation

snitin315:
Fixes https://github.com/opencollective/opencollective/issues/2856

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -662,9 +662,9 @@ export default function (Sequelize, DataTypes) {
           } else {
             potentialSlugs = [
               instance.slug,
+              instance.name ? instance.name.replace(/ /g, '-') : null,
               instance.image ? userlib.getUsernameFromGithubURL(instance.image) : null,
               instance.twitterHandle ? instance.twitterHandle.replace(/@/g, '') : null,
-              instance.name ? instance.name.replace(/ /g, '-') : null,
             ];
           }
           return Collective.generateSlug(potentialSlugs, useSlugify).then(slug => {


server/models/Collective.js
```diff
@@ -662,9 +662,9 @@ export default function (Sequelize, DataTypes) {
           } else {
             potentialSlugs = [
               instance.slug,
+              instance.name ? instance.name.replace(/ /g, '-') : null,
               instance.image ? userlib.getUsernameFromGithubURL(instance.image) : null,
               instance.twitterHandle ? instance.twitterHandle.replace(/@/g, '') : null,
-              instance.name ? instance.name.replace(/ /g, '-') : null,
```
snitin315, line 667: https://github.com/opencollective/opencollective-api/blob/43993cc4243dbacb1c4d2119638ef7fb1764d94a/server/models/Collective.js#L2946-L2961

Betree (commented):
I don't get how this is solving the issue? My understanding is that we have an issue in `generateSlug`, it only looks at the first `suggestions` entry.

snitin315:
> My understanding is that we have an issue in generateSlug, it only looks at the first suggestions entry. 

from what I have understood I think it checks the first suggestion entry and if it's available it sets the slug and returns. It moves to the next suggestion only if the previous suggestion was already taken or wasn't available. Hence checking for `name` before `twitterHandle` should fix it.

Maybe it's not the case. I will debug more will add some tests for clarification.

Betree:
Closing because it's stale, feel free to re-open if you get more time to look at this

PR #4430, opened 2020-08-24
Refactor some GraphQL V2 types

Betree:
API part of https://github.com/opencollective/opencollective/issues/3413. All changes should be non-breaking.

- `TierInterval` + `OrderFrequency` => `ContributionFrequency`
- `addStripeCreditCard` => `addCreditCard`
- `addCreditCard.paymentMethod` => `addCreditCard.creditCardInfo` 
- Added `addCreditCard.isSavedForLater`


server/graphql/v2/mutation/PaymentMethodMutations.js
```diff
@@ -1,61 +1,76 @@
...
+    } else if (args.paymentMethod) {
+      Object.assign(newPaymentMethodData, pick(args.paymentMethod, ['data', 'name', 'token']));
+    } else {
+      throw new ValidationFailed('Either creditCardInfo or paymentMethod must be provided');
+    }
 
-      try {
-        pm = await setupCreditCard(pm, {
-          collective,
-          user: account,
```
Betree, line 45: We were passing a `collective` to the `user` field, but this was only used for the `user.email`. As a consequence, the "customers" created in https://github.com/opencollective/opencollective-api/blob/c54ac505ec3718ea6e0f4de2cabfdd6e9707805e/server/paymentProviders/stripe/creditcard.js#L28 have been setup without an email. @znarf is that an issue?

server/graphql/v2/mutation/PaymentMethodMutations.js
```diff
@@ -1,61 +1,76 @@
...
+    const pm = await models.PaymentMethod.create(newPaymentMethodData);
+    return setupCreditCard(pm, { collective, user: req.remoteUser });
+  },
+};
 
-        pm.stripeError = {
-          message: error.message,
-          response: error.stripeResponse,
-        };
-      }
```
Betree, line 56: Removed that as we're not supporting `stripeError` on GQLV2 at the moment, which is going to throw an error to users if stripe asks for 3D secure. We should obviously iterate on that, but in the mean time is that ok to merge @znarf?

server/graphql/v2/mutation/PaymentMethodMutations.js
znarf: Good to know. Should not cause any issue.

server/graphql/v2/mutation/PaymentMethodMutations.js
znarf: What about editing a payment method in Recurring Contributions? Does that mean Confirmation ("3Dsecure") is not supported there?

Betree: Yes, it's not supported currently. If we face a 3D secure issue, it will just be ignored an the update will succeed. But now that I think about it, https://github.com/opencollective/opencollective-api/blob/c27ed4d726d5794c7a3ccc1a47a017ee1a571c80/templates/emails/payment.creditcard.confirmation.hbs should be sent once the payment fails no (in such case, we should continue to ignore this error)?

PR #4431, opened 2020-08-24
Refactor PENDING status and introduce more precise statuses

znarf:
`PENDING` status for Orders is currently ambiguous. In this PR, I'm introducing new statuses to lift the ambiguity and only use the `PENDING` status for "Manual" payments. For that matter, I'm introducing 3 new types:

- `NEW` for the initial default state of an Order, only application crashes will lead having this type in the database. Entries should be soft deleted after some time has passed.
- `REQUIRE_ACTION` for Payments that requires confirmation through Stripe (Strong Customer Authentication / "3D Secure").
- `PLEDGED` for "Pledge" contributions to Pledged Collectives.

Requires matching Frontend PR: https://github.com/opencollective/opencollective-frontend/pull/4917

### DIFF
--- a/migrations/20200824143059-new-order-statuses.js
+++ b/migrations/20200824143059-new-order-statuses.js
@@ -1,0 +1,24 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(
+      `UPDATE "Orders"
+  SET "status" = 'PLEDGED'
+  FROM "Collectives"
+  WHERE "Orders"."status" = 'PENDING'
+  AND "Orders"."PaymentMethodId" IS NULL
+  AND "Collectives"."id" = "Orders"."CollectiveId"
+  AND "Collectives"."isActive" = FALSE
+  AND "Collectives"."isPledged" = TRUE`,
+    );
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(
+      `UPDATE "Orders"
+  SET "status" = 'PENDING'
+  WHERE "status" = 'PLEDGED'`,
+    );
+  },
+};
--- a/server/constants/order_status.js
+++ b/server/constants/order_status.js
@@ -5,16 +5,20 @@
  * pending -> active (subscription)
  * pending -> active -> cancelled (subscription)
  * pending -> cancelled
- * pending -> rejected
  * pending -> expired
  */
 
 export default {
-  PENDING: 'PENDING',
-  PAID: 'PAID',
-  ACTIVE: 'ACTIVE',
-  CANCELLED: 'CANCELLED',
-  REJECTED: 'REJECTED',
-  ERROR: 'ERROR',
-  EXPIRED: 'EXPIRED',
+  NEW: 'NEW', // New default state since August 2020
+  REQUIRE_ACTION: 'REQUIRE_ACTION', // For Strong Customer Authentication ("3D Secure")
+  PAID: 'PAID', // For One Time Contributions
+  ERROR: 'ERROR', // For One Time and Recurring Contribution
+  // This is only for "Recurring Contributions"
+  ACTIVE: 'ACTIVE', // Active Recurring contribution with up to date payments
+  CANCELLED: 'CANCELLED', // When it's Cancelled by contributors or automatically after X failures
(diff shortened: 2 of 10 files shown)


migrations/20200824143059-new-order-statuses.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(
+      `UPDATE "Orders"
+  SET "status" = 'PLEDGED'
+  FROM "Collectives"
+  WHERE "Orders"."status" = 'PENDING'
+  AND "Orders"."PaymentMethodId" IS NOT NULL
```
Betree, line 10: Shouldn't that be `IS NULL`? I don't think pledges have payment methods attached

```suggestion
  AND "Orders"."PaymentMethodId" IS NULL
```

migrations/20200824143059-new-order-statuses.js
```diff
@@ -0,0 +1,18 @@
...
+  AND "Collectives"."id" = "Orders"."CollectiveId"
+  AND "Collectives"."isActive" = FALSE
+  AND "Collectives"."isPledged" = TRUE`,
+    );
+  },
+
+  down: async (queryInterface, Sequelize) => {},
```
Betree, line 17: Rollback is pretty simple for that one

```suggestion
  down: async (queryInterface, Sequelize) => {
    queryInterface.sequelize.query(`
      UPDATE "Orders"
      SET "status" = 'PENDING'
      WHERE "status" = 'PLEDGED'
    `);
  },
```

server/constants/order_status.js
```diff
@@ -5,16 +5,20 @@
...
 export default {
-  PENDING: 'PENDING',
-  PAID: 'PAID',
-  ACTIVE: 'ACTIVE',
-  CANCELLED: 'CANCELLED',
-  REJECTED: 'REJECTED',
-  ERROR: 'ERROR',
-  EXPIRED: 'EXPIRED',
+  NEW: 'NEW', // New default state since August 2020
+  REQUIRE_ACTION: 'REQUIRE_ACTION', // For Strong Customer Authentication ("3D Secure")
```
Betree, line 13: Not a fan of this one because it's not really explicit, a pending order always require an action. For example, a pending manual bank transfer also requires an action (from the client - to pay, and from the host admin - to confirm the payment).

If it's about `3D_SECURE`, I think something like one of the following may be more explicit:
- `REQUIRE_3D_SECURE`
- `REQUIRE_3D_SECURE_CONFIRMATION`
- `REQUIRE_CLIENT_CONFIRMATION`

server/constants/order_status.js
```diff
@@ -5,16 +5,20 @@
...
+  PAID: 'PAID', // For One Time Contributions
+  ERROR: 'ERROR', // For One Time and Recurring Contribution
+  // This is only for "Recurring Contributions"
+  ACTIVE: 'ACTIVE', // Active Recurring contribution with up to date payments
+  CANCELLED: 'CANCELLED', // When it's Cancelled by contributors or automatically after X failures
+  // This is only for "Manual" payments
+  PENDING: 'PENDING', // Initial state
```
Betree, line 20: Not especially for this PR, but maybe we can take the opportunity to make it more explicit too with something like `WAITING_HOST_CONFIRMATION`. That might require more refactor tho.

Betree (commented):
Good clarification :+1:

server/constants/order_status.js
znarf: I picked this term because I followed Stripe's terminology https://stripe.com/docs/api/payment_intents/object#payment_intent_object-status I agree it doesn't look good out of this context.

- "Confirmation" means something else in the Stripe context.
- "3D_SECURE" seems a bit too specific, it's rarely used

In the end, `REQUIRE_CLIENT_CONFIRMATION` could work.

sbinlondon: +1 to `REQUIRE_CLIENT_CONFIRMATION`

server/constants/order_status.js
znarf: I don't want to touch too much the old "Transactions" tab.

Maybe we should do that when we work on the new "Transactions" tab in the Host Dashboard.

PR #4432, opened 2020-08-24
Fix GraphQL schema issue preventing playground docs/schema to load

sbinlondon:
Screenshot 2020-08-24 at 15 26 31

My GraphQL v2 playground would not load the docs or schema. There were no errors in the terminal. However opening the browsers I found this.

Screenshot 2020-08-24 at 15 46 06

So I added the `orderBy` field to `Account.memberOf` and now it's fixed.

### DIFF
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -143,6 +143,10 @@ const accountFieldsDefinition = () => ({
         type: new GraphQLList(AccountType),
         description: 'Type of accounts (BOT/COLLECTIVE/EVENT/ORGANIZATION/INDIVIDUAL)',
       },
+      orderBy: {
+        type: new GraphQLNonNull(ChronologicalOrderInput),
+        defaultValue: ChronologicalOrderInput.defaultValue,
+      },
     },
   },
   transactions: {


server/graphql/v2/interface/Account.js
```diff
@@ -143,6 +143,10 @@ const accountFieldsDefinition = () => ({
         type: new GraphQLList(AccountType),
         description: 'Type of accounts (BOT/COLLECTIVE/EVENT/ORGANIZATION/INDIVIDUAL)',
       },
+      orderBy: {
+        type: new GraphQLNonNull(ChronologicalOrderInput),
```
znarf, line 147: It should be nullable.

znarf (changes requested):

server/graphql/v2/interface/Account.js
znarf: I'm personally not clear on the behavior between `GraphQLNonNull` and `defaultValue`, so I'm gonna trust you.

znarf:
@sbinlondon I think you need to rebase that PR against master.

server/graphql/v2/interface/Account.js
sbinlondon: When I made it non-nullable it threw more errors.
Screenshot 2020-08-24 at 18 08 09

Betree (approved):

server/graphql/v2/interface/Account.js
znarf: > When I made it non-nullable it threw more errors.

@sbinlondon You mean, when you made it nullable?

PR #4435, opened 2020-08-24
Add OrderQuery and MemberQuery to GraphQL v2

sbinlondon:
* Adds `OrderQuery` and `MemberQuery` for v2
* Adds `fetchOrderWithReference`

### DIFF
--- a/server/graphql/v2/interface/IsMemberOf.js
+++ b/server/graphql/v2/interface/IsMemberOf.js
@@ -8,6 +8,7 @@ import { MemberOfCollection } from '../collection/MemberCollection';
 import { AccountType, AccountTypeToModelMapping } from '../enum/AccountType';
 import { HostFeeStructure } from '../enum/HostFeeStructure';
 import { MemberRole } from '../enum/MemberRole';
+import { AccountReferenceInput, fetchAccountWithReference } from '../input/AccountReferenceInput';
 import { ChronologicalOrderInput } from '../input/ChronologicalOrderInput';
 
 export const IsMemberOfFields = {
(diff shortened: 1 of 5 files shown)


znarf:
@sbinlondon What's the use case for MemberQuery? I'm not sure about it. `fromAccount` and `toAccount` are looking odd.

sbinlondon:
https://github.com/opencollective/opencollective-frontend/blob/471b43ab7e5fb9bdbd3f97b7763c06c4a85ba643/components/new-contribution-flow/ContributionFlowSuccess.js#L337

When we navigate to the order success page, we don't feed in any order information via props. We do have the `OrderId` in the router query though. So first we get the order info using the `OrderQuery` but then we also need the `Member` info to see if there is a public message from when the sponsor has donated to this collective before. So we make an additional `MemberQuery` using the info we get back from the `OrderQuery` (the `toAccount`, or the Collective that's being donated to, and the `fromAccount`, who is giving the donation).

There is a v1 query I could use from the old Order page but I thought we were going to update old queries as we went.

znarf:
I think that could be introduced as a property to the `Order` like `order.membership`. Would be interested to see @Betree's feedback on this.

Betree:
> I think that could be introduced as a property to the Order like order.membership. Would be interested to see @Betree's feedback on this.

I agree with this. We could also add a `account` field on `account.memberOf`:

```gql
order(order: $order) {
  fromAccount {
    memberOf(account: $account) {
      // Returns all the memberships for fromAccount <> account
    }
  }
}
```

Whatever we do, it would be elegant to solve this with just one query.

sbinlondon:
@Betree @znarf I've made some changes, let me know what you think

server/graphql/v2/interface/IsMemberOf.js
```diff
@@ -59,6 +61,10 @@ export const IsMemberOfFields = {
           [Op.in]: args.accountType.map(value => AccountTypeToModelMapping[value]),
         };
       }
+      if (args.account) {
+        const account = await fetchAccountWithReference(args.account);
```
Betree, line 65: Small perf improvement

```suggestion
        const account = await fetchAccountWithReference(args.account, { loaders: req.loaders });
```

Betree (approved):
Works for me!

PR #4438, opened 2020-08-25
Account: Add editAccount mutation

Betree:
I initially implemented this to fill a need for the host dashboard, but ended up simplifying the feature. However this will be needed at some point, so it's better to include it while it's there.


Betree:
Closing in favor of https://github.com/opencollective/opencollective-api/pull/5203

PR #4439, opened 2020-08-25
Tax Form: only exclude RECEIPTs from tax form calculations

znarf:
Fixing what seems to be a misunderstanding in tax form calculations.

### DIFF
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -509,7 +509,7 @@ const getTopSponsors = () => {
           AND t."createdAt" > :since
         GROUP BY t."CollectiveId"
         ORDER BY "totalDonationsLast3months" DESC, collectives DESC LIMIT 20
-      ), 
+      ),
       "topSponsorsTotalDonations" as (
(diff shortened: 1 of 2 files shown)


Betree (approved):
I missed that we should count `UNCLASSIFIED` too for tax forms. Seems that it matches the [initial behavior](https://github.com/opencollective/opencollective-api/blob/c94c366e2c2a8af2ad09315bcf1e4795629f1dd3/server/models/Collective.js#L2901), so all good for me.

server/lib/queries.js
```diff
@@ -941,11 +941,11 @@ const getTaxFormsRequiredForExpenses = expenseIds => {
       AND ld.year = date_part('year', all_expenses."incurredAt")
       AND ld."documentType" = 'US_TAX_FORM'
     WHERE analyzed_expenses.id IN (:expenseIds)
-    AND analyzed_expenses.type = 'INVOICE'
+    AND analyzed_expenses.type != 'RECEIPT'
```
Betree, line 944: Just thought about it: how is that going to play with `FUNDING_REQUEST`?

PR #4447, opened 2020-08-26
Add CurrencyExchangeRate model and cron worker

kewitz:
Refactor and reconcile Currency constants and library.
Translate currency library to TypeScript.
Add CurrencyExchangeModel and migration.
Add a daily worker to track supported currencies exchange rates.


kewitz:
Rebased and fixed. Ready to merge.

server/models/CurrencyExchangeRate.ts
```diff
@@ -0,0 +1,62 @@
+import { Model } from 'sequelize';
+
+import restoreSequelizeAttributesOnClass from '../lib/restore-sequelize-attributes-on-class';
+
+/**
+ * Sequelize model to represent an ExpenseAttachedFile, linked to the `ExpenseAttachedFiles` table.
```
Betree, line 6: I always forget them too when copy pasting the boilerplate :yum: 

```suggestion
 * Sequelize model to represent an CurrencyExchangeRate, linked to the `CurrencyExchangeRates` table.
```

Betree (commented):
If I understand correctly, `CurrencyExchangeRate` is mostly there for recording, we don't yet read values from there right?

Betree:
Code is much cleaner BTW, nice job moving to async/await!

kewitz:
@Betree exactly. We're basically recording in our services so we can use it in the dashboard (metabase) and possibly other reports.

Betree (approved):
:100:

PR #4449, opened 2020-08-26
Archive older migrations

znarf:
- Move everything older than 2020 into the "archives" folder. Should we also delete them? I would love to, but looking at old migrations can be sometime interesting. While they would be in Git history, it would be a bit complicated to access.
- Deactivate (comment) migrations containing imports, so we don't look at it for things like `eslint` and `depcheck`
- I updated the DB dump yesterday, so all migrations already ran on it up to this date

People who have a staled version of the database from 2019 will need to recreate it from the dump.


Betree (approved):
:100: :100: :100:

sbinlondon (approved):
Great idea! 

>People who have a staled version of the database from 2019 will need to recreate it from the dump.

Maybe worth adding a line in the DB docs letting people know this?

PR #4450, opened 2020-08-26
parseInt on Amount value only if not null

sbinlondon:
From a support ticket: someone was trying to update their tier on a recurring contribution and it got hung up on loading the tiers:

Screenshot 2020-08-26 at 13 24 27

Had a look at the tiers query for that Collective and found this error:

Screenshot 2020-08-26 at 13 27 42

I was able to reproduce locally:

Screenshot 2020-08-26 at 13 24 36

Looking into it, in the `Amount` object for v2, we parse the amount value as an integer without checking first whether it's `null`, which it could be.

```
export const Amount = new GraphQLObjectType({
  name: 'Amount',
  description: 'A financial amount.',
  fields: {
    value: {
      type: GraphQLFloat,
      resolve(amount) {
        return parseInt(amount.value, 10) / 100;
      },
    },
...
```

In order to fix this I added a check to see if the value is null before parsing it as an int

```
      resolve(amount) {
        if (amount.value === null) {
          return null;
        } else {
          return parseInt(amount.value, 10) / 100;
        }
      },
```

Running locally and setting the minimum amount to `null` on a tier for a recurring contribution I have fixed the issue.

### DIFF
--- a/server/graphql/v2/object/Amount.js
+++ b/server/graphql/v2/object/Amount.js
@@ -9,7 +9,11 @@ export const Amount = new GraphQLObjectType({
     value: {
       type: GraphQLFloat,
       resolve(amount) {
-        return parseInt(amount.value, 10) / 100;
+        if (amount.value === null) {
+          return null;
+        } else {
+          return parseInt(amount.value, 10) / 100;
+        }
       },
     },
     currency: {


znarf:
@sbinlondon so the issue was on a Tier with a minimumAmount=null ?

sbinlondon:
> @sbinlondon so the issue was on a Tier with a minimumAmount=null ?

Yep

sbinlondon:
@znarf Ok to merge?

znarf (approved):

znarf:
Sure

Betree (commented):
Shouldn't we do the same for the `valueInCents` field?

Betree:
Also I'd recommend using the `isNil` helper from `lodash` to protect against `undefined` too

znarf:
@sbinlondon follow up?

sbinlondon:
@znarf @Betree Making a new PR now

PR #4451, opened 2020-08-26
GQLV2: Add addFunds endpoint

Betree:
For https://github.com/opencollective/opencollective-frontend/pull/4930


server/graphql/v1/mutations/orders.js
```diff
@@ -989,11 +990,16 @@ export async function addFundsToCollective(order, remoteUser) {
   };
 
   // Handle specific fees
-  if (order.hostFeePercent) {
+  if (!isNil(order.hostFeePercent)) {
```
kewitz, line 993: :eye: :lips: :eye:

PR #4453, opened 2020-08-26
Add Transaction.isRefund property

kewitz:


Betree (approved):
Way better this way :100:

PR #4461, opened 2020-08-27
Fix Slack notifications

Betree:
- No change to user webhooks. If their `webhookUrl` is configured with a proper app, it should continue working flawlessly.
- Remove all messages sent to our `#activity-public` (was broken for a month, after Slack deprecated legacy API)
- Remove all messages sent to our `#activity-private` (couldn't find the channel on our Slack)
- Changed the config structure so that we can now configure one webhook per channel, migrated `#abuse` to this system
- Added a `logger.warn` when webhook fails (so that we can remove the deprecated ones, if any)
- Removed the `flat` library (was not used anymore)

### DIFF
--- a/config/ci.json
+++ b/config/ci.json
@@ -25,6 +25,11 @@
       "bucket": "opencollective-test"
     }
   },
+  "slack": {
+    "webhooks": {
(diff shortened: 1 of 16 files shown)


znarf (approved):
Looks great, like it!

PR #4464, opened 2020-08-27
Make Amount fields nullable but Amount non-nullable

sbinlondon:


server/graphql/v2/object/Amount.js
```diff
@@ -25,7 +26,11 @@ export const Amount = new GraphQLObjectType({
     valueInCents: {
       type: GraphQLInt,
       resolve(amount) {
-        return parseInt(amount.value, 10);
+        if (isNil(amount.valueInCents)) {
```
Betree, line 29: The object is never built from `valueInCents`, we always build it by passing a `value` 

```suggestion
        if (isNil(amount.value)) {
```

server/graphql/v2/object/Order.js
```diff
@@ -113,7 +113,7 @@ export const Order = new GraphQLObjectType({
         },
       },
       platformFee: {
-        type: Amount,
```
Betree, line 116: This one has a `return null;`, we can't make it nullable

server/graphql/v2/object/Order.js
```diff
@@ -124,7 +124,7 @@ export const Order = new GraphQLObjectType({
         },
       },
       platformContributionAmount: {
-        type: Amount,
```
Betree, line 127: This one has a `return null;`, we can't make it nullable

Betree (commented):
Small changes requested below, but I think this is a good change. Brings clarity on how we should check for null values.

Thanks @sbinlondon!

Betree (approved):

PR #4473, opened 2020-08-28
Persist and expose PayPal Payout errors

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3251

### DIFF
--- a/server/graphql/v2/object/Activity.ts
+++ b/server/graphql/v2/object/Activity.ts
@@ -51,10 +51,16 @@ export const Activity = new GraphQLObjectType({
     data: {
       type: new GraphQLNonNull(GraphQLJSON),
       description: 'Data attached to this activity (if any)',
-      resolve(activity): object {
-        const toPick = ['error'];
+      async resolve(activity, _, req): Promise<object> {
+        const toPick = [];
         if (activity.type === ACTIVITY.COLLECTIVE_EXPENSE_PAID) {
           toPick.push('isManualPayout');
+        }
+        if (activity.type === ACTIVITY.COLLECTIVE_EXPENSE_ERROR) {
+          const collective = await req.loaders.Collective.byId.load(activity.CollectiveId);
+          if (req.remoteUser?.isAdmin(collective.HostCollectiveId)) {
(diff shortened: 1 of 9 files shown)


server/paymentProviders/paypal/payouts.ts
```diff
@@ -50,16 +51,25 @@ export const payExpensesBatch = async (expenses: any[]): Promise<any[]> => {
...
-  });
-  return Promise.all(updateExpenses);
+  try {
+    const response = await paypal.executePayouts(connectedAccount, requestBody);
+    const updateExpenses = expenses.map(async e => {
+      await e.update({ data: response.batch_header, status: status.PROCESSING });
+      const user = await models.User.findByPk(e.lastEditedById);
+      await e.createActivity(activities.COLLECTIVE_EXPENSE_PROCESSING, user);
+    });
+    return Promise.all(updateExpenses);
```
Betree, line 61: Is is possible to have only one expense in the batch failing? Or do the entire batch fails if just one fails?

server/graphql/v2/object/Activity.ts
```diff
@@ -52,11 +52,11 @@ export const Activity = new GraphQLObjectType({
       type: new GraphQLNonNull(GraphQLJSON),
       description: 'Data attached to this activity (if any)',
       resolve(activity): object {
+        const toPick = ['error'];
```
Betree, line 55: Are we 100% that the error message can't contain sensitive or private info, such as accounts IDs or tokens?

Betree: If we're not sure, I think we should pick this field only if `remoteUser` is a host admin.

kewitz: Well, this is out of our control so there's no way to be certain. I'll validate `remoteUser` is host admin.

server/paymentProviders/paypal/payouts.ts
kewitz: That is a really nice question! I was taking for granted that everything fails.
I just tested to make sure that was the behavior and it is, the whole batch fails.

PR #4485, opened 2020-09-01
Include reference on transactions to RUB

kewitz:
TransferWise requires a reference for transactions to particular currencies.
We're facing some issues with a transaction to RUB complaining about this, although their documentation suggests being incapable of transferring from USD to RUB, their API returns USD -> RUB as a valid transaction.
This should fix this error and implement the base code to include reference when needed, if despite this refactor the transaction still doesn't work, I'll start blacklisting currency pairs internally.

### DIFF
--- a/server/lib/transferwise.ts
+++ b/server/lib/transferwise.ts
@@ -154,7 +154,7 @@ export const createRecipientAccount = async (
   };
 };
 
-interface CreateTransfer {
+export interface CreateTransfer {
   accountId: number;
   quoteId: number;
   uuid: string;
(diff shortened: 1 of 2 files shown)


Betree (commented):
Shouldn't we always include the reference of the expense? Sounds like an interesting info to include in the payment.

kewitz:
@Betree the problem is that some currencies do not support the reference and the length of the reference changes with the currency. To avoid complexity on our side I decided to leave this out but it seems we might need to start implementing this.

Betree:
@kewitz that makes sense, all good for me them

PR #4488, opened 2020-09-01
Include gift-card donations in card emitter Transactions

kewitz:


server/graphql/v2/interface/Transaction.js
```diff
@@ -80,6 +80,9 @@ export const Transaction = new GraphQLInterfaceType({
       toAccount: {
         type: Account,
       },
+      virtualCardEmitterAccount: {
```
Betree, line 83: In https://github.com/opencollective/opencollective-api/blob/f65084f876e15e3854d83745bd3f9b196af08a61/server/graphql/v2/enum/PaymentMethodType.ts we've started to call them "Gift card". I believe the reason why they're called "virtual cards" in the code is related to the fact that the ancestor of what we have today for virtual cards was called "gift cards", so the developers had to choose a different name. But now these old payment methods are all deleted and the public API could be a great place to start renaming them.

I therefore suggest:
```suggestion
      giftCardEmitterAccount: {
```

Though I'd be ok with just:
```suggestion
      giftCardEmitter: {
```

@znarf wdyt?

server/graphql/v2/interface/Transaction.js
```diff
@@ -240,5 +243,12 @@ export const TransactionFields = () => {
         return transaction; // Individual fields are set by TransactionPermissions's resolvers
       },
     },
+    virtualCardEmitterAccount: {
+      type: Account,
+      description: '',
+      async resolve(transaction) {
+        return transaction.getVirtualCardEmitterCollective();
```
Betree, line 250: We should use a GraphQL loader for this one, transactions are usually fetched in lists so it's quite important

server/graphql/v2/query/TransactionsQuery.ts
```diff
@@ -78,7 +78,7 @@ const TransactionsQuery = {
       where['FromCollectiveId'] = fromAccount.id;
     }
     if (account) {
-      where['CollectiveId'] = account.id;
+      where[Op.or] = [{ CollectiveId: account.id }, { UsingVirtualCardFromCollectiveId: account.id, type: 'DEBIT' }];
```
Betree, line 81: I think this condition will be overwritten in https://github.com/opencollective/opencollective-api/blob/d1a6c9d959906517d18acb03c879fdd0f02f21fb/server/graphql/v2/query/TransactionsQuery.ts#L86 if we provide a `searchTerm`.

server/graphql/v2/interface/Transaction.js
znarf: @Betree 100% agree, "virtual card" has no place in the API v2, it should be "Gift Card" all the way.

PR #4501, opened 2020-09-02
Add membership of an order to the Order object

sbinlondon:


server/graphql/v2/object/Order.js
```diff
@@ -142,6 +145,21 @@ export const Order = new GraphQLObjectType({
           }
         },
       },
+      fromAccountMember: {
```
znarf, line 148: This property name doesn't make much sense to me. I propose `membership` instead. 

To figure a good name, maybe you should add a `description` and try to explain what it's returning.

znarf (approved):
Good for me.

server/graphql/v2/object/Order.js
```diff
@@ -142,6 +145,23 @@ export const Order = new GraphQLObjectType({
...
+        description:
+          'This represents a MemberOf relationship (ie: Collective backed by an Individual) attached to the Order.',
+        async resolve(order) {
+          const fromAccount = await order.getFromCollective();
+          const toAccount = await order.getCollective();
+
+          return models.Member.findOne({
+            where: {
+              MemberCollectiveId: fromAccount.id,
+              CollectiveId: toAccount.id,
```
Betree, line 159: We don't have to fetch the accounts, we can directly use the order fields (1 query instead of 3):

```suggestion
              MemberCollectiveId: order.FromCollectiveId,
              CollectiveId: order.CollectiveId,
```

server/graphql/v2/object/Order.js
```diff
@@ -142,6 +145,23 @@ export const Order = new GraphQLObjectType({
...
+              role: roles.BACKER,
```
Betree, line 160: Members are also grouped by tier. To be accurate here we should to:

```suggestion
              role: roles.BACKER,
              TierId: order.TierId,
```

PR #4505, opened 2020-09-03
fix: remove related collectives from thankyou email

snitin315:
Fixes https://github.com/opencollective/opencollective/issues/2681


Betree (commented):
Good move @snitin315!

@znarf `relatedcollectives` is still used in customized thank your emails (ie. `/templates/emails/thankyou.brusselstogether.hbs`) and in the `order.processing.hbs` + `subscription.canceled.hbs` templates. Should we remove them as well?

znarf:
@Betree I personally think so because this kind of feature is not looking good when it's not maintained.

Betree:
@znarf I agree.
@snitin315 Let's remove it from everywhere, including the template https://github.com/opencollective/opencollective-api/blob/9469b3a0571e1648382dd074615bc2fe7cbd6f23/server/lib/emailTemplates.js#L127 and where we define the variable (ie https://github.com/opencollective/opencollective-api/blob/ad2e019bc6dcf4cbdac0b55f2a4f97cc0669cecb/cron/monthly/user-report.js#L203).

Let me know if you don't have the bandwidth to followup :slightly_smiling_face:

snitin315:
I will remove it from everywhere 👍🏻

Betree:
@snitin315 it seems the test was not passing for this one, we should rebase & check what's wrong. Le me know if you want me to followup.

snitin315:
CI seems happy now 😄 .

/cc @Betree

Betree (approved):
Perfect, thank you @snitin315! Let's ship this :rocket:

PR #4506, opened 2020-09-03
2FA login now works with one-time links enabled

sbinlondon:
Addresses the issue the our login process and its one-time links did not work with our 2FA login flow.

This PR:

* adjusts the `/users/update-token` route so that it is only used for normal logins. It only checks if 2FA is enabled on the user's account, but doesn't handle the verification

* adds a new route `/users/two-factor-auth` to handle the verification code received from the front end and allow these users to log in

### DIFF
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -9,6 +9,7 @@ import { isValidEmail } from '../lib/utils';
 import models from '../models';
 
 const { Unauthorized } = errors;
+const { User } = models;
 
 /**
  *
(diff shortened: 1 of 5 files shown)


server/controllers/users.js
```diff
@@ -72,36 +72,36 @@ export const signin = (req, res, next) => {
...
  */
-export const updateToken = async (req, res, next) => {
-  const { twoFactorAuthenticatorCode } = req.body;
-
-  // the first time we try, we need to just check if the user has 2FA
-  if (req.remoteUser.twoFactorAuthToken !== null && !twoFactorAuthenticatorCode) {
-    return next(new Unauthorized('Two-factor authentication is enabled on this account. Please enter the code'));
+export const updateToken = async (req, res) => {
+  if (req.remoteUser.twoFactorAuthToken !== null) {
+    const token = req.remoteUser.jwt({ scope: 'twofactorauth' }, auth.TOKEN_EXPIRATION_SESSION);
```
znarf, line 84: Seems to me this token has no limitation and can be used as a regular token.

If true, this defeats the 2FA and allow anyone with this token to perform actions without being confirmed.

znarf (changes requested):
I like the approach that I find cleaner than before.

Seems to me there is a security issue. Let me know if I'm missing something.

sbinlondon:
As per discussion with @znarf going to work on implementing security feature that means JWTs with `scope: login` can only be used on the `/users/update-token` route and JWTs with `scope: twofactorauth` can only be used on the `/users/two-factor-auth` route.

Also will consider splitting `/users/update-token` further into two separate functions where the first handles login tokens only and the second handles refreshing/exchanging for long-term tokens only.

sbinlondon:
Now has a separate middleware for making sure we're only using `scope: twofactorauth` on `/user/two-factor-auth` 

Still to do:

* splitting `/users/update-token` further into two separate functions where the first handles login tokens only and the second handles refreshing/exchanging for long-term tokens only

server/middleware/authentication.js
```diff
@@ -364,3 +374,28 @@ export function mustBeLoggedIn(req, res, next) {
     }
   });
 }
+
+export const checkTwoFactorAuthJWT = (req, res, next) => {
+  parseJwtNoExpiryCheck(req, res, e => {
```
znarf, line 379: Let's simplify and use directly `jwt.verify` instead of this `parseJwtNoExpiryCheck` + `checkJwtExpiry` which are a bit weird.

server/middleware/authentication.js
```diff
@@ -98,12 +98,22 @@ export const _authenticateUserByJwt = async (req, res, next) => {
...
+
   /**
    * Functionality for one-time login links. We check that the lastLoginAt
    * in the JWT matches the lastLoginAt in the db. If so, we allow the user
    * to log in, and update the lastLoginAt.
    */
   if (req.jwtPayload.scope === 'login') {
+    const path = req.path;
+
+    if (path !== '/users/update-token') {
```
znarf, line 113: Thank you, not beautiful but doing the job. Can you add some code comments explaining and future steps for this?

server/controllers/users.js
```diff
@@ -72,36 +73,44 @@ export const signin = (req, res, next) => {
...
+  const user = await User.findByPk(userId);
+  if (!user) {
+    logger.warn(`User id ${userId} not found`);
+    next();
+    return;
   }
 
-  // we process a new token the 1st time if no 2FA, 2nd time if there is
-  const token = req.remoteUser.jwt({}, auth.TOKEN_EXPIRATION_SESSION);
+  const token = user.jwt({}, auth.TOKEN_EXPIRATION_SESSION);
```
znarf, line 108: For slightly better code structure, I would move this line just before the `res.send({ token: token });`. No need to generate the token if we're about to send an error.

znarf:
> splitting /users/update-token further into two separate functions where the first handles login tokens only and the second handles refreshing/exchanging for long-term tokens only

This is only a nice to have, let's make sure we're not spending multiple days on this.

server/middleware/authentication.js
```diff
@@ -364,3 +383,36 @@ export function mustBeLoggedIn(req, res, next) {
...
+  if (!header) {
+    return next(new BadRequest('Requires Authorization headers.'));
+  }
+
+  const parts = header.split(' ');
+  const scheme = parts[0];
+  const token = parts[1];
+  if (!/^Bearer$/i.test(scheme) || !token) {
+    return next(new BadRequest('Format is Authorization: Bearer [token]'));
+  }
```
znarf, line 398: Could you extract this in an helper and share it with the other function from where it was copy/pasted?

server/middleware/authentication.js
```diff
@@ -364,3 +383,36 @@ export function mustBeLoggedIn(req, res, next) {
...
+
+  jwt.verify(token, jwtSecret, (err, decoded) => {
+    // JWT library either returns an error or the decoded version
+    if (err && err.name === 'TokenExpiredError') {
+      req.jwtExpired = true;
+      req.jwtPayload = jwt.decode(token, jwtSecret); // we need to decode again
```
znarf, line 404: We don't want this handling of expired tokens. They should just fail.

server/middleware/authentication.js
```diff
@@ -364,3 +381,28 @@ export function mustBeLoggedIn(req, res, next) {
...
+
+  jwt.verify(token, jwtSecret, (err, decoded) => {
+    // JWT library either returns an error or the decoded version
+    if (err) {
+      return next(new BadRequest(err.message));
+    } else {
+      req.jwtPayload = decoded;
+    }
+  });
+
```
znarf, line 401: I think there is a potential problem here, if `jwt.verify` is asynchronous then we should wait before moving to the next block. Currently, there is no guarantee that the `jwt.verify` is done.

Maybe by performing `jwt.verify` with async/await instead of a callback you'll see better the problem.

server/middleware/authentication.js
```diff
@@ -364,3 +381,28 @@ export function mustBeLoggedIn(req, res, next) {
...
+  // if token does not have scope of 'twofactorauth' we should reject it
+  if (!req.jwtPayload || req.jwtPayload.scope !== 'twofactorauth') {
+    return next(new Unauthorized('Cannot use this token on this route.'));
+  } else {
+    next();
+  }
```
znarf, line 407: We should make sure this block is executed once we're done with `jwt.verify`.

sbinlondon:
CI won't pass now because the frontend branch changed. Please see this PR: https://github.com/opencollective/opencollective-api/pull/4544

PR #4514, opened 2020-09-04
Inherit platformFeePercent unless covid-19 initiative

kewitz:
This means resetting the platformFeePercent when moving from a host with fees on top to a host without.

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1200,6 +1200,17 @@ export default function (Sequelize, DataTypes) {
     });
   };
 
+  Collective.prototype.getProjects = function (query = {}) {
+    return Collective.findAll({
+      ...query,
+      where: {
+        ...query.where,
+        ParentCollectiveId: this.id,
+        type: types.PROJECT,
+      },
+    });
+  };
+
   /**
    * Return stats about backers based on the Members table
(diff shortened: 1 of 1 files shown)


server/models/Collective.js
```diff
@@ -1800,16 +1800,14 @@ export default function (Sequelize, DataTypes) {
       throw new Error('This host is not open to applications');
     }
 
+    const platformFeePercent = this.tags?.includes('covid-19') ? 0 : hostCollective.platformFeePercent;
```
znarf, line 1803: What about `EVENT` and `PROJECT`? How do we make sure they inherit their parent?

znarf (changes requested):
I don't get where/how Events and Projects are inheriting the platformFeePercent of their parent.

kewitz:
@znarf I was counting on the whole creation/cascading flow for that.
I'll make it explicit.

znarf:
@kewitz case to support: a Covid-19 collective decide to join another Host and the Host have platformFeePercent=5, the Events of this Collective should have platformFeePercent=0 (from the parent) and not platformFeePercent=5 (from the host).

kewitz:
@znarf yep, in the case the event or project is not tagged with `covid-19`. Just pushed the fix.

PR #4516, opened 2020-09-04
Request to removed as admin

JoyLubega:
Resolve : opencollective/opencollective/issues/3141


server/graphql/v1/mutations.js
```diff
@@ -451,6 +451,37 @@ const mutations = {
       }
     },
   },
+  sendEmailToAdmins: {
```
JoyLubega, line 454: @Betree This is the mutation referenced in the frontend for the backend, the only difference is that `sendEmailToAdmins` sends an email, yet `editCoreContributors` doesn't.

Betree (changes requested):
Nice start @JoyLubega! It's definitely the feature that we want.

There are however a few blockers that we need to solve:
- The mutation should be on our API V2. I can suggest adding that to `memberMutations` in https://github.com/opencollective/opencollective-api/blob/cb90393f72a416c512b6f5d544f81ddbbb01c1f9/server/graphql/v2/mutation/MemberMutations.js#L7
- The mutation name should be more explicit. `sendEmailToAdmins` doesn't really tells you what it does. I can suggest `requestAdminRemoval`
- Taking the emails from the arguments is dangerous, because attackers could put any email they want in this parameter to trigger SPAM. We need to load the admins dynamically, you can use `collective.getAdminUsers` for that.
- There should be a link to the settings in the email for the admins to easily go and remove the other admin

JoyLubega:
> * There should be a link to the settings in the email for you to easily go and remove yourself as an admin

I ne

Betree:
Sorry @JoyLubega, we've decided to go with a simpler solution of letting admins remove themselves as long as they're not the last admin of the collective. Thank you for your contribution!

PR #4525, opened 2020-09-08
Transactions: Include incognito contributions

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3469

Also added to `Account.transactions` resolver for consistency.

**Request for feedback**
- Should there be a param to enable that (ie. `includeIncognitoContributions`), with a default to `false`?

---

image

### DIFF
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -166,6 +166,12 @@ const accountFieldsDefinition = () => ({
       orderBy: {
         type: ChronologicalOrderInput,
       },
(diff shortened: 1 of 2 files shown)


kewitz:
Very nice @Betree!
Related to your question, I'm OK with automatically including incognito transactions for the admin of the account, my only problem is that we should be clear in the frontend that we're displaying these items only for that user.
Perhaps something like the privacy icon like we have in the Expense page.
We could do that by fetching the `isIncognito` property from the account.

PR #4528, opened 2020-09-08
Implement graphql-cache POC

znarf:
I'm not familiar enough with Apollo Server, so rolling our own cache for this proof of concept.

- upgrade GraphQL v1 to Apollo Server (better middleware)
- replace `purgeCacheForPage` by a more generic `purgeCacheForCollective`
- introduce caching based on a specific list of GraphQL operation names
- use our caching layer: Redis in production
- disabled by default, opt in
- 5 minutes by default. configurable
- flushed whenever the previous Cloudflare flushing was implemented
- opt out for authenticated requests

### DIFF
--- a/server/graphql/cache.js
+++ b/server/graphql/cache.js
@@ -1,0 +1,101 @@
+import { difference } from 'lodash';
+
+export const purgeCacheForCollectiveOperationNames = [
+  'CollectivePage',
+  'BudgetSection',
+  'UpdatesSection',
+  'TransactionsSection',
+  'TransactionsPage',
+  'ExpensesPage',
+  'CollectiveBannerIframe',
+  'CollectiveCover',
+  'RecurringContributions',
+  'ContributionsSection',
+];
+
+export function checkSupportedVariables(req, variableNames) {
+  if (!req.body.variables) {
+    return false;
+  }
+  // Check missing variables
+  if (difference(variableNames, Object.keys(req.body.variables)).length > 0) {
+    return false;
+  }
+  // Check extra variables
+  if (difference(Object.keys(req.body.variables), variableNames).length > 0) {
+    return false;
+  }
+
+  return true;
+}
+
+export function getGraphqlCacheKey(req) {
+  if (req.remoteUser) {
+    return;
+  }
+  if (!req.body || !req.body.operationName) {
+    return;
+  }
+  switch (req.body.operationName) {
+    case 'CollectivePage':
+      if (!checkSupportedVariables(req, ['slug', 'nbContributorsPerContributeCard'])) {
+        return;
+      }
+      if (req.body.variables.nbContributorsPerContributeCard !== 4) {
+        return;
+      }
+      return `${req.body.operationName}_${req.body.variables.slug}`;
+    case 'BudgetSection':
+      if (!checkSupportedVariables(req, ['slug', 'limit'])) {
+        return;
+      }
+      if (req.body.variables.limit !== 3) {
+        return;
+      }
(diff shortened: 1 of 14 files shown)


server/graphql/cache.js
```diff
@@ -0,0 +1,37 @@
...
+    case 'TransactionsPage':
+      if (req.body.variables.offset === 0 && req.body.variables.limit === 15) {
+        return `${req.body.operationName}_${req.body.variables.slug}`;
+      }
+      break;
+    case 'ExpensesPage':
+      if (req.body.variables.offset === 0 && req.body.variables.limit === 10) {
+        return `${req.body.operationName}_${req.body.variables.slug}`;
+      }
+      break;
```
Betree, line 35: Many variables can influence the results on these pages (payout method type, status, amount, etc). How are we dealing with that if there's only one cache key?

server/routes.js
```diff
@@ -106,21 +107,47 @@ export default app => {
...
+      apiKey: get(config, 'graphql.apolloEngineAPIKey'),
+    },
+    // Align with behavior from express-graphql
+    context: ({ req }) => {
+      return req;
+    },
+    formatResponse: (response, ctx) => {
+      const req = ctx.context;
+      if (req.cacheKey) {
+        cache.set(req.cacheKey, response, Number(config.graphql.cache.ttl));
```
Betree, line 144: Is there no risk to set the response from a logged in user here? Query like `ExpensesPage` can contain private info (ie. payoutMethod, requiredLegalDocument, transaction.uuid, etc)

znarf: I opt out authenticated users in the latest version.

server/graphql/cache.js
znarf: True, it's not properly implemented, the idea was to only cache the default case where you have slug+offset=0+limit=10.

I'm gonna change that.

server/lib/hyperwatch.js
```diff
@@ -56,6 +56,9 @@ const load = async app => {
 
         if (req.body && req.body.query && req.body.variables) {
           log = log.set('graphql', req.body);
+          if (res.servedFromGraphqlCache) {
+            log = log.setIn(['graphql', 'servedFromCache'], true);
```
Betree, line 60: :+1:

Betree (approved):
On a general feedback, I think it's a good MVP.

My main concern for this in the long term is that there's no way for other developers to know that the cache defined in `server/graphql/cache.js` exists when they work on a feature. For example, if someone change the default number of contributors per tier card that is checked in https://github.com/opencollective/opencollective-api/blob/78dd201c8c1309f9a0bc1ac6b43e4ff30b5a2c98/server/graphql/cache.js#L44 or the default number of transactions in https://github.com/opencollective/opencollective-api/blob/78dd201c8c1309f9a0bc1ac6b43e4ff30b5a2c98/server/graphql/cache.js#L68 to make a design update, they'll break the cache without knowing it.

For this reason I like the idea of having the cache defined at the endpoint level.

znarf:
@Betree it's definitely a Proof of Concept and has good characteristics for such:

- easily disableable
- easily deletable
- opt in, only activated for selected cases

I'm gonna merge it because it's gonna protect our performance against a known pattern. When someone post a link to Open Collective on Twitter or Mastodon, we have suddenly a lot of traffic on a given page. In this case, the CDN is not protecting us because the robots (mostly, but also humans) are coming from many different regions. This will work well in this situation.

As a follow up, we should:
- invest time in the Apollo Server Cache configuration
- invest time in cache invalidation mechanisms

If we happen to have a good Apollo Server Cache support, we should definitely delete and remove this.

PR #4529, opened 2020-09-08
Allow manual payment when expense status is ERROR

kewitz:
This is a UX improvement, the host admin wants to be able to manually process failed transactions on certain errors.

### DIFF
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -602,7 +602,11 @@ export async function payExpense(req, args) {
       'Expense is currently being processed, this means someone already started the payment process',
     );
   }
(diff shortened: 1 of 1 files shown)


Betree (commented):
We should update the error message below (`Expense needs to be approved`). Also I don't know what the pay modal looks like in this state, but we should probably hide the "Payment - Automatic" option if not already done.

kewitz:
@Betree I think the error message is still valid since we're just adding an exception for errored expenses that are being manually paid.
What message would you suggest here?

Betree:
I was thinking about `Expense needs to be approved or in error state`, but now that I think more about it it will just be more confusing. Let's keep it as is.

PR #4535, opened 2020-09-09
Fix Host report

kewitz:
This should fix the case where a collective changes its host and/or is deactivated during the period covered by the report.
It also explains how the host balance is calculated.

### DIFF
--- a/server/lib/hostlib.js
+++ b/server/lib/hostlib.js
@@ -4,20 +4,28 @@ import _ from 'lodash';
 import { convertToCurrency } from '../lib/currency';
 import models, { Op, sequelize } from '../models';
 
-export function getHostedCollectives(hostid, endDate = new Date()) {
+export function getHostedCollectives(hostid, startDate, endDate = new Date()) {
   return sequelize.query(
     `
-    SELECT g.* FROM "Collectives" g
-    LEFT JOIN "Members" ug ON g.id = ug."CollectiveId"
-    WHERE ug.role='HOST'
-      AND ug."MemberCollectiveId"=:hostid
-      AND g."deletedAt" IS NULL
-      AND ug."deletedAt" IS NULL
-      AND ug."createdAt" < :endDate
-      AND g."createdAt" < :endDate
+    with "members" as (
+      SELECT m."CollectiveId"
+      FROM "Members" m
+      WHERE m.role='HOST'
+        AND m."MemberCollectiveId" = :hostid
+        AND (m."deletedAt" IS NULL OR m."deletedAt" > :startDate)
+        AND m."createdAt" < :endDate
+    ), "transactions" as (
+      SELECT DISTINCT t."CollectiveId" FROM "Transactions" t WHERE t."HostCollectiveId" = :hostid AND t."createdAt" < :endDate AND t."createdAt" > :startDate
+    )
+
+    SELECT DISTINCT *
+    FROM "Collectives" c
+    WHERE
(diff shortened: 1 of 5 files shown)


templates/emails/host.report.hbs
```diff
@@ -111,6 +111,10 @@ Subject: {{host.name}} Host {{reportName}}
       </tr>
     </table>
   </center>
+  <p>
+    * This is the total balance belonging to of all hosted collectives and it doesn't considers any host fees
+    that you have collected during this period.
```
znarf, line 116: >  This is the total balance belonging to of all hosted collectives

Ok.

> and it doesn't considers any host fees that you have collected during this period.

This is not making things clearer to me, with all the context I have! We should better explain what that means regarding Host Fees.

If this means Host Fees are not counted, this is not great, the balance should definitely includes the Host Fees.

Also, we should be aware of 2 different cases:
- Classic: Hosts without budget (we should simply add the host fees)
- New: Hosts with budget activated (we should simply make sure we account the Host own transactions, and we generated the report once Host Fees are added to the budget, at the end of the month currently)

znarf: In the end, ok to merge this PR with a clearer message about Host Fees.

kewitz: We calculate the host balance by summing the `netAmount` of every transaction, it is safe to say host fees are not counted. This is exactly what I replied to that ticket (9338) from Daniel (Club-A-Kitchen).

What I don't have context is if this is on purpose or just a naive implementation that should be fixed.
I understand why this could be done on purpose: it seems more useful to have the net balance of my collectives since I already know how much I have in the bank including host fees, while the opposite isn't true.
This value represents what I could consider the minimum amount I should have in the bank.
Also, there's the case the host fees were partially consumed by a transaction not tracked in our platform, in this case, the balance would never match the reality anyway.

Perhaps we should display both values separated: the net balance of your collectives and how much you collected in host fees during the period.
cc @piamancini @alanna

znarf: @kewitz about wording, sorry but "it doesn't considers" is really difficult to process, especially for someone not familiar with the platform. Are they in? Are they out? I still think it could be better explained.

> We calculate the host balance by summing the netAmount of every transaction, it is safe to say host fees are not counted. 

Have you taken in consideration Hosts with budget activated? Foundation or OSC for example. https://opencollective.com/opensource/transactions or https://opencollective.com/foundation/transactions You can see Host fees being added to their respective Budget.

kewitz: > We should display the amount the host is managing for its collectives and the amount they collected in host fees.
MANCINI, P. 2020

server/lib/hostlib.js
```diff
@@ -4,20 +4,28 @@ import _ from 'lodash';
...
+    with "members" as (
+      SELECT m."CollectiveId"
+      FROM "Members" m
+      WHERE m.role='HOST'
+        AND m."MemberCollectiveId" = :hostid
+        AND (m."deletedAt" IS NULL OR m."deletedAt" > :startDate)
+        AND m."createdAt" < :endDate
+    ), "transactions" as (
+      SELECT DISTINCT t."CollectiveId" FROM "Transactions" t WHERE t."HostCollectiveId" = :hostid AND t."createdAt" < :endDate AND t."createdAt" > :startDate
+    )
```
kewitz, line 19: I had to go through transactions since the membership record was hard-deleted when the collectives were moved to the new host.

PR #4544, opened 2020-09-10
2FA login now works with one-time links enabled

sbinlondon:
Last PR was broken on CI because the frontend branch changed.

Addresses the issue the our login process and its one-time links did not work with our 2FA login flow.

This PR:

    adjusts the /users/update-token route so that it is only used for normal logins. It only checks if 2FA is enabled on the user's account, but doesn't handle the verification

    adds a new route /users/two-factor-auth to handle the verification code received from the front end and allow these users to log in


znarf:
I'm gonna merge because seems E2E is now failing without it.

PR #4550, opened 2020-09-11
Allow transferwise.availableCurrencies blocklist to be bypassed

kewitz:
This is useful to generate a generic bank account form without the intent to be used in TransferWise.

### DIFF
--- a/server/graphql/v2/object/TransferWise.ts
+++ b/server/graphql/v2/object/TransferWise.ts
@@ -71,10 +71,16 @@ export const TransferWise = new GraphQLObjectType({
       },
     },
     availableCurrencies: {
+      args: {
+        ignoreBlockedCurrencies: {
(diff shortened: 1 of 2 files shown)


znarf:
Following Up on https://github.com/opencollective/opencollective-api/pull/4276 let's try to avoid the term "blacklist" from now on.

kewitz:
@znarf thanks for the heads up! I'll refactor these PRs and use the new terms.d

znarf (changes requested):
"blocklist" is a good replacement to "blacklist" as a noun.

However, I think "block" is better than "blocklist" as a verb, so "blockListedCurrencies" -> "blockedCurrencies".

Betree (commented):
`blockedCurrencies` sounds better to me too, the rest looks good :+1:

PR #4559, opened 2020-09-14
Replace isomorphic-fetch with node-fetch

znarf:


znarf:
Looks like we rely on it as a polyfill. Closing!

PR #4561, opened 2020-09-14
Reject transaction

sbinlondon:
Frontend PR https://github.com/opencollective/opencollective-frontend/pull/5070

Relates https://github.com/opencollective/opencollective/issues/3484

* Adds capability for Collective admins to refund transactions
* Adds 'isRejected' field to `Transaction` v2 object
* Adds `rejectTransaction` mutation which:
    * refunds the transaction
    * changes the related `Order` status to `REJECTED`
    * destroys the membership with the role `BACKER`

### DIFF
--- a/server/graphql/common/transactions.ts
+++ b/server/graphql/common/transactions.ts
@@ -1,4 +1,7 @@
+import moment from 'moment';
+
 import { roles } from '../../constants';
+import orderStatus from '../../constants/order_status';
 import { TransactionTypes } from '../../constants/transactions';
 
 const isRoot = async (req): Promise<boolean> => {
@@ -66,6 +69,16 @@ const isPayeeHostAdmin = async (req, transaction): Promise<boolean> => {
   return req.remoteUser.isAdmin(collective.HostCollectiveId);
 };
 
+const isPayeeCollectiveAdmin = async (req, transaction): Promise<boolean> => {
+  if (!req.remoteUser) {
+    return false;
+  }
+  const collective = await req.loaders.Collective.byId.load(
+    transaction.type === 'CREDIT' ? transaction.CollectiveId : transaction.FromCollectiveId,
+  );
+  return req.remoteUser.isAdmin(collective.id);
+};
+
 /**
  * Returns true if the transaction meets at least one condition.
  * Always returns false for unauthenticated requests.
@@ -89,14 +102,34 @@ export const canRefund = async (transaction, _, req): Promise<boolean> => {
   if (transaction.type !== TransactionTypes.CREDIT || transaction.OrderId === null) {
     return false;
   }
-  return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin]);
+  const timeLimit = moment().subtract(30, 'd');
+  const createdAtMoment = moment(transaction.createdAt);
+  const transactionIsOlderThanThirtyDays = createdAtMoment < timeLimit;
+  if (transactionIsOlderThanThirtyDays) {
+    return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin]);
+  } else {
+    return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin, isPayeeCollectiveAdmin]);
+  }
 };
 
 export const canDownloadInvoice = async (transaction, _, req): Promise<boolean> => {
+  if (transaction.OrderId) {
+    const order = await req.loaders.Order.byId.load(transaction.OrderId);
+    if (order.status === orderStatus.REJECTED) {
+      return false;
+    }
+  }
   return remoteUserMeetsOneCondition(req, transaction, [
     isPayerCollectiveAdmin,
     isPayeeHostAdmin,
     isPayerAccountant,
     isPayeeAccountant,
   ]);
+};
+
+export const canReject = async (transaction, _, req): Promise<boolean> => {
+  if (transaction.type !== TransactionTypes.CREDIT || transaction.OrderId === null) {
+    return false;
+  }
+  return remoteUserMeetsOneCondition(req, transaction, [isPayeeCollectiveAdmin]);
 };
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -12,6 +12,7 @@ import FEATURE from '../../../constants/feature';
 import status from '../../../constants/order_status';
 import roles from '../../../constants/roles';
 import { VAT_OPTIONS } from '../../../constants/vat';
+import { canRefund } from '../../../graphql/common/transactions';
 import cache, { purgeCacheForCollective } from '../../../lib/cache';
 import * as github from '../../../lib/github';
 import * as libPayments from '../../../lib/payments';
@@ -22,7 +23,7 @@ import { canUseFeature } from '../../../lib/user-permissions';
 import { capitalize, formatCurrency, md5 } from '../../../lib/utils';
 import models from '../../../models';
 import { setupCreditCard } from '../../../paymentProviders/stripe/creditcard';
-import { FeatureNotAllowedForUser, NotFound, Unauthorized, ValidationFailed } from '../../errors';
+import { FeatureNotAllowedForUser, Forbidden, NotFound, Unauthorized, ValidationFailed } from '../../errors';
 
 const oneHourInSeconds = 60 * 60;
 
@@ -782,20 +783,16 @@ export async function refundTransaction(_, args, req) {
     throw new NotFound('Transaction not found');
   }
 
-  const collective = await models.Collective.findByPk(transaction.CollectiveId);
-  const isHost = await collective.isHost();
-  const HostCollectiveId = isHost ? collective.id : await collective.getHostCollectiveId();
-
-  // 1. Verify user permission. User must be either
-  //   a. Admin of the collective that received the donation - disabled for now
+  // 1a. Verify user permission using canRefun. User must be either
+  //   a. Admin of the collective that received the donation
   //   b. Admin of the Host Collective that received the donation
   //   c. Admin of opencollective.com/opencollective
+  // 1b. Check transaction age - only Host admins can refund transactions older than 30 days
+  // 1c. The transaction type must be CREDIT to prevent users from refunding their own DEBITs
 
-  if (
-    !req.remoteUser.isAdmin(HostCollectiveId) &&
-    !req.remoteUser.isRoot() /* && !req.remoteUser.isAdmin(collective.id) */
-  ) {
-    throw new Unauthorized('Not a site admin or host collective admin');
+  const canUserRefund = await canRefund(transaction, undefined, req);
+  if (!canUserRefund) {
+    throw new Forbidden('Cannot refund this transaction');
   }
 
   // 2. Refund via payment method
--- a/server/graphql/v2/interface/Transaction.js
+++ b/server/graphql/v2/interface/Transaction.js
@@ -8,6 +8,7 @@ import {
 } from 'graphql';
 import { GraphQLDateTime } from 'graphql-iso-date';
 
+import orderStatus from '../../../constants/order_status';
 import models from '../../../models';
 import * as TransactionLib from '../../common/transactions';
 import { TransactionType } from '../enum/TransactionType';
@@ -32,6 +33,11 @@ const TransactionPermissions = new GraphQLObjectType({
       type: new GraphQLNonNull(GraphQLBoolean),
       description: "Whether the current user can download this transaction's invoice",
       resolve: TransactionLib.canDownloadInvoice,
+    },
+    canReject: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      description: 'Whether the current user can reject the transaction',
+      resolve: TransactionLib.canReject,
     },
   },
 });
@@ -112,6 +118,12 @@ export const Transaction = new GraphQLInterfaceType({
       },
       permissions: {
         type: TransactionPermissions,
+      },
+      isRejected: {
+        type: new GraphQLNonNull(GraphQLBoolean),
+      },
+      refundTransaction: {
+        type: Transaction,
       },
     };
   },
@@ -271,6 +283,23 @@ export const TransactionFields = () => {
         return transaction.UsingVirtualCardFromCollectiveId
           ? await req.loaders.Collective.byId.load(transaction.UsingVirtualCardFromCollectiveId)
           : null;
+      },
+    },
+    isRejected: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      async resolve(transaction, _, req) {
+        if (transaction.OrderId) {
+          const order = await req.loaders.Order.byId.load(transaction.OrderId);
+          return order.status === orderStatus.REJECTED;
+        } else {
+          return false;
+        }
+      },
+    },
+    refundTransaction: {
+      type: Transaction,
+      resolve(transaction) {
+        return transaction.getRefundTransaction();
       },
     },
   };
--- a/server/graphql/v2/mutation/TransactionMutations.ts
+++ b/server/graphql/v2/mutation/TransactionMutations.ts
@@ -1,6 +1,11 @@
-import { GraphQLNonNull } from 'graphql';
+import { GraphQLNonNull, GraphQLString } from 'graphql';
 
-import { Unauthorized } from '../../errors';
+import orderStatus from '../../../constants/order_status';
+import { canReject } from '../../../graphql/common/transactions';
+import { purgeCacheForCollective } from '../../../lib/cache';
+import { notifyAdminsOfCollective } from '../../../lib/notifications';
+import models from '../../../models';
+import { Forbidden, NotFound, Unauthorized } from '../../errors';
 import { refundTransaction as legacyRefundTransaction } from '../../v1/mutations/orders';
 import { fetchTransactionWithReference, TransactionReferenceInput } from '../input/TransactionReferenceInput';
 import { Transaction } from '../interface/Transaction';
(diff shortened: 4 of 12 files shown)


server/graphql/v1/mutations/orders.js
```diff
@@ -786,14 +786,11 @@ export async function refundTransaction(_, args, req) {
...
-  //   a. Admin of the collective that received the donation - disabled for now
+  //   a. Admin of the collective that received the donation
   //   b. Admin of the Host Collective that received the donation
   //   c. Admin of opencollective.com/opencollective
 
-  if (
-    !req.remoteUser.isAdmin(HostCollectiveId) &&
-    !req.remoteUser.isRoot() /* && !req.remoteUser.isAdmin(collective.id) */
-  ) {
+  if (!req.remoteUser.isAdmin(HostCollectiveId) && !req.remoteUser.isRoot() && !req.remoteUser.isAdmin(collective.id)) {
```
Betree, line 793: Shouldn't we introduce the logic about 30 days policy here?

server/graphql/v2/interface/Transaction.js
```diff
@@ -252,5 +261,20 @@ export const TransactionFields = () => {
           : null;
       },
     },
+    isRejected: {
+      type: GraphQLBoolean,
```
Betree, line 265:
```suggestion
      type: new GraphQLNonNull(GraphQLBoolean),
```

server/graphql/v2/interface/Transaction.js
```diff
@@ -252,5 +261,20 @@ export const TransactionFields = () => {
...
+      async resolve(transaction, _, req) {
+        if (transaction.OrderId) {
+          const order = await req.loaders.Order.byId.load(transaction.OrderId);
+          if (order.status === orderStatus.REJECTED) {
+            return true;
+          } else {
+            return false;
+          }
```
Betree, line 273: **[:ice_cream: Details & personal preferences]** 

```suggestion
          return order.status === orderStatus.REJECTED;
```

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -23,6 +26,56 @@ const transactionMutations = {
...
+          returning: true,
+        },
+      );
+      if (!orderToUpdate) {
+        throw new NotFound('Order not found');
+      }
+      // get membership info
+      const toAccount = await fetchAccountWithReference({ legacyId: transaction.CollectiveId });
+      const fromAccount = await fetchAccountWithReference({ legacyId: transaction.FromCollectiveId });
+      const membership = await models.Member.findOne({
```
Betree, line 61: Someone can have multiple memberships to the same collective if they donate through multiple tiers. So either:
- We want to delete them all, then we need to use a `findAll`
- We want to delete just the one matching this contribution, then we need to add `TierId: order.TierId` to the query

Betree (commented):
The code looks super nice, great job @sbinlondon!

I wonder if we shouldn't implement https://github.com/opencollective/opencollective/issues/1656 first (or prevent users from rejecting PayPal contributions)? If they try to reject a PayPal contribution today, they'll get an error.

server/graphql/v1/mutations/orders.js
sbinlondon: Yes! I hadn't quite gotten around to that yet, but thanks for pointing out where it should go. I wasn't sure.

server/graphql/v2/interface/Transaction.js
sbinlondon: Omg of course 🤦‍♀️

server/graphql/v2/mutation/TransactionMutations.ts
sbinlondon: Interesting. I think we'd want to delete them all right? Because the point is that the person rejecting *doesn't* want this backer in their Collective at all. 

Would that be an issue to delete the memberships associated with orders/transactions that haven't also been rejected?

server/graphql/common/transactions.ts
```diff
@@ -89,7 +99,7 @@ export const canRefund = async (transaction, _, req): Promise<boolean> => {
   if (transaction.type !== TransactionTypes.CREDIT || transaction.OrderId === null) {
     return false;
   }
-  return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin]);
+  return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin, isPayeeCollectiveAdmin]);
```
znarf, line 102: I'm surprised by the wording here `payee`, it's something that I expect for Expenses and not for Contributions. No need to take action now but just wanted to mention that. cc @Betree

server/graphql/v1/mutations/orders.js
sbinlondon: This is the function for all refunds as well as now being used for rejects.

Do we want to enforce the 30 day policy on all these? cc @znarf @kewitz @piamancini

server/graphql/v2/mutation/TransactionMutations.ts
Betree: > I think we'd want to delete them all right

Seems like a good call. Then we shouldn't throw if the membership doesn't exist.

Also, may be worth using `models.Member.delete({ where ... })` to make only one SQL query, and to invalidate the cache only one time (must be done manually though `invalidateContributorsCache`, as models `hooks` are not triggered when using the `Model.Member.update` syntax.

server/graphql/v1/mutations/orders.js
sbinlondon: @Betree What's stopping us from implementing Paypal refunds? Have we just not got around to it or is there a blocker?

server/graphql/common/transactions.ts
Betree: @kewitz was the one introducing the term in this file. I stayed consistent when I added resolvers here because even though we don't use it in other places, the term `payee` seemed pretty appropriate to me - it's the person who receives the money from the transaction (while a `toAccount` can have different meanings if it's a CREDIT or DEBIT).

server/graphql/v2/mutation/TransactionMutations.ts
sbinlondon: I've just seen this comment, I will update the code after standup

server/graphql/common/transactions.ts
kewitz: The logic is quite simple, it is just the composition between transaction type and collective. The problem is that we're abstracting that into a function if we want to keep consistency and the function will require a name.
I also considered `beneficiary` at the time, but since we already have payee elsewhere, it seemed simpler.

server/graphql/v2/mutation/TransactionMutations.ts
sbinlondon: Sorry, have looked at the Sequelize docs and tried `models.Member.delete` and I get an error it isn't a function and it doesn't say this exists in the docs. It says to use `destroy()`.

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -23,6 +27,92 @@ const transactionMutations = {
...
+      if (memberships) {
+        memberships.forEach(membership => {
+          membership.destroy();
+        });
+      }
+
+      // TO DO: email user to let them know their transaction has been rejected
+      const fromAccountEmail = await req.loaders.getUserDetailsByCollectiveId
+        .load(fromAccount.id)
+        .then(user => user.email);
```
znarf, line 95: We don't want to use this strategy for emailing.

Email should go to all collective admins using the standard mechanism we have for that.

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -23,6 +27,92 @@ const transactionMutations = {
...
+          returning: true,
+        },
+      );
+      if (!orderToUpdate) {
+        throw new NotFound('Order not found');
+      }
+
+      // get membership info
+      const toAccount = await fetchAccountWithReference({ legacyId: transaction.CollectiveId });
+      const fromAccount = await fetchAccountWithReference({ legacyId: transaction.FromCollectiveId });
```
znarf, line 76: I don't see the case to use `fetchAccountWithReference` here. This is only meant to be used with a Reference object coming from the user. Use plain sequelize instead (findByPk)

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -23,6 +27,92 @@ const transactionMutations = {
...
+        const contributor = {
+          name: fromAccount.name,
+        };
+        const rejectionReason = args.message || null;
+
+        emailLib.send(
+          'contribution.rejected',
+          fromAccountEmail,
+          { collective, contributor, rejectionReason },
+          { sendEvenIfNotProduction: true },
```
znarf, line 109: Let's keep `sendEvenIfNotProduction` only for things like signup / login.

We don't want people to receive such emails when we test on staging.

sbinlondon: Ah right, I misunderstood what that was for.

server/graphql/v1/mutations/orders.js
```diff
@@ -785,19 +785,24 @@ export async function refundTransaction(_, args, req) {
...
+  // 1a. Verify user permission. User must be either
+  //   a. Admin of the collective that received the donation
   //   b. Admin of the Host Collective that received the donation
   //   c. Admin of opencollective.com/opencollective
 
-  if (
-    !req.remoteUser.isAdmin(HostCollectiveId) &&
-    !req.remoteUser.isRoot() /* && !req.remoteUser.isAdmin(collective.id) */
-  ) {
+  if (!req.remoteUser.isAdmin(HostCollectiveId) && !req.remoteUser.isRoot() && !req.remoteUser.isAdmin(collective.id)) {
```
Betree, line 797: Here we don't check wether the transaction is a CREDIT or DEBIT. It means that, by passing the proper id, contributors could be able to refund their own transactions.

Betree (changes requested):

server/graphql/v1/mutations/orders.js
sbinlondon: Can you clarify what you mean? Do I need to make sure the transaction is only refunded if it isn't CREDIT or DEBIT? Or only if it isn't? Or to make sure that it's one or the other?

sbinlondon: Looking at `canRefund` in `transactions.ts`, we check whether it's CREDIT or not and if it isn't, we can't refund. So does that take care of your issue? Or do you also want to add that to the line you've highlighted here?

```
export const canRefund = async (transaction, _, req): Promise<boolean> => {
  if (transaction.type !== TransactionTypes.CREDIT || transaction.OrderId === null) {
    return false;
  }
  return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin, isPayeeCollectiveAdmin]);
};
```

server/graphql/v2/interface/Transaction.js
```diff
@@ -113,6 +119,12 @@ export const Transaction = new GraphQLInterfaceType({
       permissions: {
         type: TransactionPermissions,
       },
+      isRejected: {
+        type: GraphQLBoolean,
```
Betree, line 123: Should match the one below

```suggestion
        type: new GraphQLNonNull(GraphQLBoolean),
```

server/graphql/v2/interface/Transaction.js
```diff
@@ -273,5 +285,27 @@ export const TransactionFields = () => {
...
+          return false;
+        }
+      },
+    },
+    refundTransaction: {
+      type: Transaction,
+      resolve(transaction) {
+        // If it's a sequelize model transaction, it means it has the method getRefundTransaction
+        // otherwise we just null
+        if (transaction && transaction.getRefundTransaction) {
```
Betree, line 304: I'm not sure how can we end up with a `null` transaction here, or why we would end up with something else than a sequelize transaction?

server/graphql/v1/mutations/orders.js
Betree: > Looking at canRefund in transactions.ts, we check whether it's CREDIT or not and if it isn't, we can't refund. So does that take care of your issue?

But unless I'm mistaking `canRefund` is not called from this `refundTransaction`.

The idea is that if I (collective 1) make a contribution to RandomCollective (collective 2), the transactions pair will look like that:
> { id: 1, type: DEBIT, FromCollectiveId: 2, CollectiveID: 1 }
> { id: 2, type: CREDIT, FromCollectiveID: 1, CollectiveId: 2 }

The happy path to refund the transaction is: `refundTransaction(2)`. We'll check that you're an admin of collective 2, and if yes than everything's fine.

But let's say I'm a malicious user, and I call this endpoint as `refundTransaction(1)`. Now the collective ID (1) will be my user profile, so I'll be allowed to refund my own transaction.

Note: I haven't tested this to confirm, it's only theoretical.

sbinlondon: Ok, thanks for explaining! I understand better now. I didn't know about the DEBIT/CREDIT model of transactions, I thought that meant the card type 🤦‍♀️ I will add that check in `refundTransaction`

templates/emails/contribution.rejected.hbs
```diff
@@ -0,0 +1,19 @@
+Subject: Your contribution to {{collective.name}}
+
+{{> header}}
+
+<h1>Hey {{contributor.name}},</h1>
```
znarf, line 5: Maybe this doesn't make sense if it's an Organization. Should we skip it?

server/graphql/common/transactions.ts
```diff
@@ -86,17 +99,37 @@ const remoteUserMeetsOneCondition = async (req, transaction, conditions): Promis
 
 /** Checks if the user can see transaction's attachments (items URLs, attached files) */
 export const canRefund = async (transaction, _, req): Promise<boolean> => {
-  if (transaction.type !== TransactionTypes.CREDIT || transaction.OrderId === null) {
+  const timeLimit = moment().subtract(30, 'd');
```
Betree, line 102: `canRefund` will return false for host admins after 30 days too, is this what we want?

server/graphql/v1/mutations/orders.js
```diff
@@ -785,19 +786,29 @@ export async function refundTransaction(_, args, req) {
   const collective = await models.Collective.findByPk(transaction.CollectiveId);
   const isHost = await collective.isHost();
   const HostCollectiveId = isHost ? collective.id : await collective.getHostCollectiveId();
+  const timeLimit = moment().subtract(30, 'd');
+  const createdAtMoment = moment(transaction.createdAt);
+  const transactionIsOlderThanThirtyDays = createdAtMoment < timeLimit;
```
Betree, line 791: We could also unify and use the `canRefund` helper directly

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -23,6 +26,89 @@ const transactionMutations = {
...
+          role: 'BACKER',
+        },
+      });
+
+      // remove member from Collective
+      if (memberships) {
+        memberships.forEach(membership => {
+          membership.destroy();
+        });
+      }
```
Betree, line 89: Not a blocker, but did you try using a batch function as recommended in https://github.com/opencollective/opencollective-api/pull/4561#discussion_r489347538? I realized I made a mistake there (it should be `destroy`, not `delete`) but the idea is to do something tike:

```suggestion
      await models.Member.destroy({
        where: {
          MemberCollectiveId: fromAccount.id,
          CollectiveId: toAccount.id,
          role: 'BACKER',
        },
      });

      purgeCacheForCollective(fromAccount);
      purgeCacheForCollective(toAccount);
```

The main advantage of this would be to make only one SQL query, and to purge the cache only one time (vs doing it once per membership).

sbinlondon: Right, sorry, I didn't use that since I thought you needed to use `delete` with what you suggested. I can add this change.

server/graphql/common/transactions.ts
```diff
@@ -89,14 +102,34 @@ export const canRefund = async (transaction, _, req): Promise<boolean> => {
...
   }
-  return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin]);
+  const timeLimit = moment().subtract(30, 'd');
+  const createdAtMoment = moment(transaction.createdAt);
+  const transactionIsOlderThanThirtyDays = createdAtMoment < timeLimit;
+  if (transactionIsOlderThanThirtyDays) {
+    return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin]);
+  } else {
+    return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeCollectiveAdmin]);
+  }
```
Betree, line 112: This condition is problematic:
- If transaction is older than 30 days, only host admins can refund (good)
- Otherwise only collective admins can refund (we want host admins there too)

server/graphql/v2/interface/Transaction.js
Betree: This was marked as resolved but there's no change, you may have forgotten to push your changes

Betree (changes requested):
We have a security issue on `rejectTransaction`: if you call the endpoint with an already refunded transaction that you don't own, we won't enter `legacyRefundTransaction` and thus there will be no check on permissions. As a consequence, malicious users may be able to mark any order as rejected and to remove the linked memberships.

Betree:
We can iterate in another PR, but it would be great to have some tests to make sure permissions are properly covered. Things that would be interesting to test:
- canReject: Collective admins can only refund during 30 days
- rejectTransaction
  - Can't reject if not root, host admin or collective admin
  - Refunds the transaction, removes all memberships and mark the order as rejected

sbinlondon:
I have added the permissions check for `canReject` then in the condition that it doesn't go on to `legacyRefundTransaction`, do you think that's sufficient?

Betree:
@sbinlondon I'd do the check just after `fetchTransactionWithReference`, it's not a big deal if we check two times (in case we're refunding) and that way we can provide an unified error message (without relying on the one from `legacyRefundTransaction`). But that's details, no requested change.

PR #4571, opened 2020-09-16
Introduce categories on accounts to help moderation

znarf:


server/graphql/v2/interface/Account.js
```diff
@@ -259,6 +259,10 @@ const accountFieldsDefinition = () => ({
     type: Location,
     description: 'The address associated to this account. This field is always public for collectives and events.',
   },
+  categories: {
+    type: new GraphQLList(GraphQLString),
```
Betree, line 263: We can probably do that since there's a default

```suggestion
    type: new GraphQLNonNull(new GraphQLList(GraphQLString)),
```

Betree (commented):
Using an `AccountCategory` enum could be helpful for users, to know the values that they can handle

PR #4595, opened 2020-09-21
Make ExpenseItem.incurredAt optional

kewitz:
Related to https://github.com/opencollective/opencollective/issues/3512

Not required, but technically, grant items do not have an incurred date.


kewitz:
Ignoring this refactor for now.

PR #4612, opened 2020-09-24
Make reject work with recurring contributions

sbinlondon:
Checks if the `Order` attached to a `Transaction` has a `Subscription` attached and if so, will do the work to cancel that recurring contribution.

### DIFF
--- a/server/graphql/v2/mutation/TransactionMutations.ts
+++ b/server/graphql/v2/mutation/TransactionMutations.ts
@@ -49,6 +49,11 @@ const transactionMutations = {
       // get transaction info
       const transaction = await fetchTransactionWithReference(args.transaction);
 
+      const canUserReject = await canReject(transaction, undefined, req);
+      if (!canUserReject) {
+        throw new Forbidden('Cannot reject this transaction');
+      }
+
       /** refund transaction and set status - - if the transaction has already been
        * refunded we don't want to try and do it again, but we will continue with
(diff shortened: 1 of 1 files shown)


server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -64,19 +71,43 @@ const transactionMutations = {
...
+      if (orderToUpdate.SubscriptionId) {
+        if (!req.remoteUser.isAdmin(orderToUpdate.FromCollectiveId)) {
+          throw new Unauthorized("You don't have permission to cancel this recurring contribution");
+        }
+        if (!orderToUpdate.Subscription.isActive && orderToUpdate.status === orderStatus.CANCELLED) {
+          throw new Error('Recurring contribution already canceled');
+        }
+        await orderToUpdate.update({ status: orderStatus.REJECTED });
+        await orderToUpdate.Subscription.deactivate();
+        await models.Activity.create({
```
znarf, line 93: If we don't use this Activity for something, we can just remove it.

znarf: If we include it, it should be REJECTED, not CANCELLED

sbinlondon: Don't we have to create the `Activity` for the subscription being cancelled so that is recorded in the history?

PR #4621, opened 2020-09-25
Review manual payment recipients

znarf:


Betree (approved):

server/lib/payments.js
```diff
@@ -498,20 +498,20 @@ export const sendOrderProcessingEmail = async order => {
...
     fromCollective: fromCollective.activity,
     pendingOrderLink: `${config.host.website}/${collective.slug}/orders/${order.id}`,
   };
 
-  return emailLib.send('order.new.pendingFinancialContribution', user.email, data, {
-    from: `${collective.name} <no-reply@${collective.slug}.opencollective.com>`,
-  });
+  const adminUsers = await host.getAdminUsers();
+  for (const adminUser of adminUsers) {
+    await emailLib.send('order.new.pendingFinancialContribution', adminUser.email, data);
```
Betree, line 513: Maybe we could send just one email with everyone as `CCI`?

znarf: Not the right moment to do this, because it's following our general practice.

However, I think we should use `notifyAdminsOfCollective` instead. If we want to support, `CCI` in the future, we will just have to modify `notifyAdminsOfCollective`.

Betree: *Update: BCC, not CCI. CCI is the French term*

Many emails start with `Hi {name}` so they're not easily converted.

We already have some places using `bcc` like https://github.com/opencollective/opencollective-api/blob/f2e241a75343c247b6e6ef3af3fe691fb975dd47/server/controllers/services/email.js#L54 or https://github.com/opencollective/opencollective-api/blob/15f81efadaaccd13daa921671b1b82bbf5f51c4a/server/models/Collective.js#L1650.

I'm going to push for this approach in https://github.com/opencollective/opencollective/issues/3626 to prevent re-uploading the PDF for each email.

PR #4633, opened 2020-09-28
HTML content: replace all external links by redirects

HADES-01:
Replace all external links by redirects
Fixes opencollective/opencollective#3536

### DIFF
--- a/server/lib/sanitize-html.ts
+++ b/server/lib/sanitize-html.ts
@@ -1,3 +1,4 @@
+import config from 'config';
 import { truncate } from 'lodash';
 import LibSanitize from 'sanitize-html';
 
@@ -89,6 +90,19 @@ export const buildSanitizerOptions = (allowedContent: AllowedContentType = {}): 
     transformTags: {
       h1: 'h3',
       h2: 'h3',
+      a: function (tagName, attribs) {
+        return {
+          tagName: 'a',
+          attribs: {
+            ...attribs,
+            href: `${
+              !attribs.href.startsWith(config.host.website)
+                ? `${config.host.website}/redirect?url=${encodeURIComponent(attribs.href)}`
+                : attribs.href
+            }`,
+          },
+        };
+      },
     },
   };
 };
--- a/test/server/lib/sanitize-html.test.js
+++ b/test/server/lib/sanitize-html.test.js
@@ -1,4 +1,5 @@
 import { expect } from 'chai';
+import config from 'config';
 
 import {
   buildSanitizerOptions,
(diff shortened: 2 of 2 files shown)


server/lib/sanitize-html.ts
```diff
@@ -89,6 +89,14 @@ export const buildSanitizerOptions = (allowedContent: AllowedContentType = {}):
     transformTags: {
       h1: 'h3',
       h2: 'h3',
+      a: function (tagName, attribs) {
+        const newAttribs = attribs;
+        newAttribs.href = `/redirect?url=${encodeURIComponent(attribs.href)}`;
+        return {
+          tagName: 'a',
+          attribs: newAttribs,
+        };
```
Betree, line 98: Using a different variable (`newAttribs`) has no effect, as mutating `newAttribs` will still mutate `attribs`. So we can simplify (and make sure we don't mutate) like that:

```suggestion
        return {
          tagName: 'a',
          attribs: {
            ...attribs,
            href: `/redirect?url=${encodeURIComponent(attribs.href)}`,
          },
        };
```

test/server/lib/sanitize-html.test.js
```diff
@@ -135,7 +135,7 @@ describe('server/lib/sanitize-html', () => {
...
+        'Si qua in iis corrigere voluit, deteriora fecit. Nec enim, dum metuit, iustus est, et certe, si metuere destiterit, non erit; Unum nescio, quo modo possit, si luxuriosus sit, finitas cupiditates habere. Illud vero minime consectarium, sed in primis hebes, illorum scilicet, non tuum, gloriatione dignam esse beatam vitam. Entry Header 1 Entry Header 2 Entry Header 3 Entry Line 1 Entry Line 2 Entry Line 3 Quid affers, cur Thorius, cur Caius Postumius, cur omnium horum magister, Orata, non iucundissime vixerit? Non laboro, inquit, de nomine. <i>Ex rebus enim timiditas, non ex vocabulis nascitur.</i> Hoc loco discipulos quaerere videtur, ut, qui asoti esse velint, philosophi ante fiant. <a href="/redirect?url=%2Fredirect%3Furl%3Dhttp%253A%252F%252Floripsum.net%252F" target="_blank">Dicimus aliquem hilare vivere;</a> Quae quidem sapientes sequuntur duce natura tamquam videntes; Nulla erit controversia. Si verbum sequimur, primum longius verbum praepositum quam bonum. Eorum enim omnium multa praetermittentium, dum eliga...',
```
Betree, line 138: We should ideally have the domain in the URL (`config.host.website`). Also this URL looks a bit strange, I would have expected:

```
href="https://opencollective.com/redirect?url=http%253A%252F%252Floripsum.net%252F"
```

Betree (commented):
Looks good!

One last thing to add to the review below: we should only do the redirect if we're facing an external link (`!href.startsWith(config.host.website)`).

test/server/lib/sanitize-html.test.js
HADES-01: @Betree I am not sure what is `config.host.website` as it is not defined in this file.

Betree: It's an import that contains all the config for the API, including the ones added from env variables. You can import it with `import config from 'config'`.

Betree:
@HADES-01 the code looks good to me, but the tests seem stuck. Could you rebase your branch on `master` to see if it helps?

HADES-01:
@Betree I did what you asked but it the `helloworks` test is still stuck.

Betree:
The failing test on HelloWork seemed related to the fact that the PR is opened on a fork. I've moved your commits to an internal branch: https://github.com/opencollective/opencollective-api/pull/4728

There was also a related E2E test that I fixed in https://github.com/opencollective/opencollective-api/pull/4728.

PR #4640, opened 2020-09-29
Add contribution policy and moderation categories

sbinlondon:
Relates https://github.com/opencollective/opencollective-frontend/pull/5171
Relates https://github.com/opencollective/opencollective/issues/3466

* Adds `moderation` as an allowed collective setting
* Adds `contributionPolicy` field on `COLLECTIVE`
* Adds moderation categories


server/graphql/v2/interface/Account.js
```diff
@@ -269,6 +269,9 @@ const accountFieldsDefinition = () => ({
       return collective;
     },
   },
+  contributionPolicy: {
```
Betree, line 272: Can we rather add that on `AccountWithContributions`? Unless there's a purpose having this field available on Users/Orgs

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -55,6 +56,17 @@ const accountMutations = {
           }
         }
 
+        if (args.key === 'moderation') {
```
Betree, line 59: I understand that you followed the pattern introduced above with the `collectivePage` settings, but the right place to check that should be in [validateSettings](https://github.com/opencollective/opencollective-api/blob/9aaa9a9393c8c7c720614aa08b19249b9ebe776d/server/lib/collectivelib.js#L103).

The main reason is that this function is called from the model, so regardless of where we try to edit the settings (GQLV1, V2, ForestAdmin, etc.) it will always be called.

server/models/Collective.js
```diff
@@ -276,6 +276,8 @@ export default function (Sequelize, DataTypes) {
 
       expensePolicy: DataTypes.TEXT, // markdown
 
+      contributionPolicy: DataTypes.TEXT, // markdown
```
Betree, line 279: Maybe we should limit the size?

Betree (commented):
Looks good :+1:

server/graphql/v2/mutation/AccountMutations.ts
sbinlondon: I did add it there and thru logs found that it called the function, but it didn't throw the error when it wasn't valid when called from this mutation for some reason. I can add it back and maybe you can help troubleshoot.

server/models/Collective.js
sbinlondon: With varchar?

server/graphql/v2/mutation/AccountMutations.ts
sbinlondon: I used `forEach` instead of `for x of x` and that was the problem. Fixed now!

server/models/Collective.js
Betree: Just with a sequelize validation like https://github.com/opencollective/opencollective-api/blob/cc3f8d8946559d3c62140b103c9f3ce4d392ee89/server/models/Collective.js#L199, changing the type is more complex to update if we need to in the future

Betree (approved):

PR #4641, opened 2020-09-29
PayPal payouts checking race-condition

kewitz:
There were two episodes where a batch was processed twice: September 28, 2020, 4:04 AM, and September 7, 2020, 9:34 PM.
I'm implementing some updates here, including delaying the transaction check worker to give 30 minutes so the webhook can have priority over it.
Duplicated transactions were manually soft-deleted.

### DIFF
--- a/cron/hourly/11-check-pending-paypal-expenses.js
+++ b/cron/hourly/11-check-pending-paypal-expenses.js
@@ -20,10 +20,13 @@ export async function run() {
         {
           status: { [Op.notIn]: [status.PAID, status.ERROR] },
           updatedAt: {
-            [Op.gte]: moment().subtract(15, 'days').toDate(),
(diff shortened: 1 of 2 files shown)


Betree (commented):
Seems like a good fix. If I understand correctly, `cron/hourly/11-check-pending-paypal-expenses.js` is only useful in case we didn't receive the webhook request from PayPal?

kewitz:
@Betree exactly, this is the fallback and I'm giving 30 minutes to prioritize the webhook.

Betree (approved):
Then this looks good to me :+1: 

There's still a very tiny chance of race condition for example if PayPal is overloaded and starts calling webhooks after a 30 minutes delay, but that's very unlikely. Locking could solve that for good, but I don't think we need to do it here & now.

PR #4648, opened 2020-10-01
Moderation features testing

sbinlondon:
Testing new moderation features

* tests Collective settings validation relating to `settings.moderation.rejectedCategories`
* `canReject` function: no tests because for now it's the exact same permissions as `canRefund`
* tests `rejectTransaction` v2 mutation

### DIFF
--- a/test/server/graphql/common/transactions.test.js
+++ b/test/server/graphql/common/transactions.test.js
@@ -1,7 +1,8 @@
 import { expect } from 'chai';
+import sinon from 'sinon';
 
 import { roles } from '../../../../server/constants';
-import { canDownloadInvoice, canRefund } from '../../../../server/graphql/common/transactions';
+import { canDownloadInvoice, canRefund, canReject } from '../../../../server/graphql/common/transactions';
 import { fakeCollective, fakeOrder, fakeTransaction, fakeUser } from '../../../test-helpers/fake-data';
 import { makeRequest } from '../../../utils';
 
@@ -24,7 +25,9 @@ describe('server/graphql/common/transactions', () => {
     hostAdminReq,
     hostAccountantReq,
     rootAdminReq,
-    contributorReq;
+    contributorReq,
+    timer,
+    oldTransaction;
 
   before(async () => {
     randomUser = await fakeUser();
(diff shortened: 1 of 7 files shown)


Betree (commented):
Looking good!

> canReject function: no tests because for now it's the exact same permissions as canRefund

The main purpose of tests is to detect breaking changes, so in such situation it's fine (and better) to copy paste and adapt the tests from `canRefund` - not to make sure that the feature works today, but to make sure it won't break in the future (ie. if we refactor the code for `canReject`).

Moreover, I see that `canRefund` and `canReject` are not using a common helper (they're duplicated code), so it would be easy to break one and not the other.

Finally, no update have been made on `canRefund` for testing the "30 days" limit. We should add a `Collective admins can only refund during 30 days` test on `canRefund` + `canReject`

sbinlondon:
> Finally, no update have been made on canRefund for testing the "30 days" limit. We should add a Collective admins can only refund during 30 days test on canRefund + canReject

I thought I had added that, my apologies. Will do now!

znarf (approved):

test/server/graphql/common/transactions.test.js
```diff
@@ -78,9 +89,35 @@ describe('server/graphql/common/transactions', () => {
...
       expect(await canRefund(transaction, undefined, fromCollectiveAccountantReq)).to.be.false;
+      expect(await canRefund(transaction, undefined, hostAdminReq)).to.be.true;
+      expect(await canRefund(transaction, undefined, rootAdminReq)).to.be.true;
+      expect(await canRefund(oldTransaction, undefined, hostAdminReq)).to.be.true;
+      expect(await canRefund(oldTransaction, undefined, rootAdminReq)).to.be.true;
+    });
+
+    it('can refund as admin of the receiving collective only if transaction < 30 days old', async () => {
       expect(await canRefund(transaction, undefined, collectiveAdminReq)).to.be.true;
+      expect(await canRefund(oldTransaction, undefined, collectiveAdminReq)).to.be.false;
```
Betree, line 100: (Not a requested change) Would have been great to test that host admins are not impacted by this

Betree (commented):
Looks good :+1:

PR #4651, opened 2020-10-01
Fix missing hosts with 0 platformFees bug

kewitz:
We were missing a couple of hosts because when we give them `feesOnTop`, we'll technically collect 0 `platformFees`.


kewitz:
Sorry, I just noticed I forgot to assign someone for review.

PR #4658, opened 2020-10-05
Add rejected moderation categories to Collective

sbinlondon:


znarf (changes requested):
The design is not bad itself but this was introduced as loosely typed strings in GraphQL v2 and we should follow the same here. https://github.com/opencollective/opencollective-api/pull/4571

znarf (approved):

PR #4662, opened 2020-10-06
TaxForm: Ignore tax forms from past years

Betree:
Fix https://opencollective.slack.com/archives/GFH4N961L/p1601498433020900

Also did some cleanup and added a test.

### DIFF
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -979,7 +979,7 @@ const getTaxFormsRequiredForAccounts = async (accountIds = [], date = new Date()
       AND d."documentType" = 'US_TAX_FORM'
     LEFT JOIN "LegalDocuments" ld
       ON ld."CollectiveId" = account.id
-      AND ld.year = date_part('year', all_expenses."incurredAt")
+      AND ld.year = :year
       AND ld."documentType" = 'US_TAX_FORM'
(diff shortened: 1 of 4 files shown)


server/lib/queries.js
```diff
@@ -979,7 +979,7 @@ const getTaxFormsRequiredForAccounts = async (accountIds = [], date = new Date()
       AND d."documentType" = 'US_TAX_FORM'
     LEFT JOIN "LegalDocuments" ld
       ON ld."CollectiveId" = account.id
-      AND ld.year = date_part('year', all_expenses."incurredAt")
```
Betree, line 982: This is just to get cleaner / more performant results (was not a bug)

server/lib/tax-forms.js
```diff
@@ -22,7 +22,7 @@ export async function findAccountsThatNeedToBeSentTaxForm(year) {
   } else {
     return models.Collective.findAll({
       where: { id: { [Op.in]: results.map(result => result.collectiveId) } },
-      include: [{ association: 'legalDocuments', required: false }],
+      include: [{ association: 'legalDocuments', required: false, where: { year } }],
```
Betree, line 25: This :arrow_up: was the root cause

PR #4682, opened 2020-10-08
Stop using redirects in oAuth flow, use default URLs or context

znarf:
Fix: https://github.com/opencollective/opencollective-security/issues/58

### DIFF
--- a/server/controllers/connectedAccounts.js
+++ b/server/controllers/connectedAccounts.js
@@ -1,134 +1,109 @@
-import Promise from 'bluebird';
 import config from 'config';
 import { get } from 'lodash';
 
 import { mustBeLoggedInTo } from '../lib/auth';
 import errors from '../lib/errors';
 import * as github from '../lib/github';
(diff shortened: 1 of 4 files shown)


Betree (approved):
LGTM

server/controllers/connectedAccounts.js
```diff
@@ -1,134 +1,109 @@
...
-            : `${config.host.website}/github/apply/${token}?utm_source=${utmSource}`;
-
-          res.redirect(newLocation);
-        })
-        .catch(next);
+
+      break;
     }
 
     case 'meetup':
```
Betree, line 61: I think we can remove the `meetup` integration there, it's already removed from the frontend since the release of the new CCF

Betree (commented):
The refactor looks good :+1:

server/controllers/connectedAccounts.js
znarf: Yep, will remove in a follow up PR.

PR #4683, opened 2020-10-08
Guest contributions

Betree:
Part of https://github.com/opencollective/opencollective/issues/3443

A first draft of the work for guest contributions that:
- Removes support for passing a `user` in `createOrder` V1. Shouldn't break anything as we deprecated this behavior a long time ago.
- Implements the basis for guest contributions

*Todo*
- [x] Frontend
- [x] Return the guest tokens from `createOrder` mutation
- [x] Make sure 3D secure works with guest accounts

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -23,7 +23,7 @@ import { handleHostPlanAddedFundsLimit, handleHostPlanBankTransfersLimit } from 
 import recaptcha from '../../../lib/recaptcha';
 import { getChargeRetryCount, getNextChargeAndPeriodStartDates } from '../../../lib/recurring-contributions';
 import { canUseFeature } from '../../../lib/user-permissions';
-import { capitalize, formatCurrency, md5 } from '../../../lib/utils';
+import { capitalize, formatCurrency, md5, parseToBoolean } from '../../../lib/utils';
 import models from '../../../models';
 import { setupCreditCard } from '../../../paymentProviders/stripe/creditcard';
 import { FeatureNotAllowedForUser, Forbidden, NotFound, Unauthorized, ValidationFailed } from '../../errors';
@@ -100,7 +100,9 @@ async function checkOrdersLimit(order, reqIp) {
 }
 
 const checkGuestContribution = order => {
-  if (order.interval) {
+  if (!parseToBoolean(config.guestContributions.enable)) {
+    throw new Error('Guest contributions are not enabled yet');
+  } else if (order.interval) {
     throw new Error('You need to sign up to create a recurring contribution');
   } else if (order.guestInfo?.email && !isEmail(order.guestInfo.email)) {
     throw new Error('You need to provide a valid email');
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -112,7 +112,8 @@ const checkGuestContribution = order => {
 };
 
 async function checkRecaptcha(order, remoteUser, reqIp) {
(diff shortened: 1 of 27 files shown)


server/models/GuestToken.ts
```diff
@@ -0,0 +1,62 @@
+import { Model } from 'sequelize';
```
sbinlondon, line 1: Why are we adding a new Model here, as opposed to a table or something else?

server/graphql/v2/scalar/EmailAddress.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  return value;
+};
+
+/**
+ * A field whose value conforms to the standard internet email address format as specified in RFC822: https://www.w3.org/Protocols/rfc822/.
+ * Inspired by https://github.com/Urigo/graphql-scalars/blob/master/src/scalars/EmailAddress.ts
+ */
+const EmailAddress = new GraphQLScalarType({
```
sbinlondon, line 20: Good idea! Will help us check emails elsewhere as well, not just with guest contributions.

server/graphql/v1/mutations/orders.js
```diff
@@ -29,7 +32,7 @@ const oneHourInSeconds = 60 * 60;
 
 const debug = debugLib('orders');
 
-async function checkOrdersLimit(order, remoteUser, reqIp) {
+async function checkOrdersLimit(order, reqIp) {
   if (['ci', 'test'].includes(config.env)) {
```
sbinlondon, line 36: I know this isn't part of your PR but I just noticed this and didn't realise we could include special return cases for CI! 🤯

server/graphql/v1/mutations/orders.js
```diff
@@ -96,8 +99,29 @@ async function checkOrdersLimit(order, remoteUser, reqIp) {
...
+    if (!guestInfo.name || !guestInfo.location?.address || !guestInfo.location.country) {
+      throw new Error('Contributions that are more than $5000 must have an address attached');
+    }
+  }
+};
+
 async function checkRecaptcha(order, remoteUser, reqIp) {
-  if (['ci', 'test'].includes(config.env)) {
+  // Disabled for all environments
+  if (['ci', 'test', 'development', 'production'].includes(config.env)) {
```
sbinlondon, line 124: I just noticed this and didn't realise we could include special return cases for CI! 🤯

server/models/GuestToken.ts
Betree: In the DB, it is a table (`GuestTokens`), the model is just it's translation to sequelize so we can easily use it in the code

server/graphql/v1/mutations/orders.js
Betree: It's better to not have conditional behaviors based on the env when possible, for reproducibility and testability. But sometimes it's very convenient!

server/models/GuestToken.ts
Betree: As for the reason to add it as a separate table rather than a column:
- Most profiles won't have a guest token. Having a guest token on verified organizations for example doesn't make much sense. When we do `models.Collective.findAll` it loads all fields by default, but 99% of the time we don't care about the guest token, so loading them is not useful.
- It's good to have some metadata about the tokens, like their creation date. With that in place we'll be able to expire them after a certain period if needed. Related benefit: it's easier to audit.
- `Collectives` table is already too big, and contains too much info. It's slow to edit. An alternative is sometimes to add the data to the `data` field, but querying it takes a long time (because it's JSON) and it's not typed, so it's easy to make mistakes when working with it.

PR #4684, opened 2020-10-08
feat(paymentReceipts): add the number of total transactions to invoice query

flickz:
Following up https://github.com/opencollective/opencollective-frontend/pull/5201 in order to get the right number of transactions for a specific receipt, I'm adding `totalTransactions` to `InvoiceType` query, v1.


znarf (approved):
Ok for API v1 code if that helps move forward.

Test needs to pass.

Betree (approved):

PR #4694, opened 2020-10-09
Move googleapis to dependencies

Betree:
A fix to followup on https://github.com/opencollective/opencollective-api/pull/4462

It was part of `devDependencies` so the lib was not found when running the script on the API.


znarf (approved):

snitin315 (approved):
My Bad. Thanks for the fix.

Betree:
@snitin315 Np, I didn't spot it either :slightly_smiling_face: Tagged you just to let you know.

PR #4710, opened 2020-10-13
fix(Recurring contributions): Add logs to migration

sbinlondon:


kewitz (approved):
Everything looks OK and the migration works locally.

Betree (commented):
A simpler approach to add logs while making sure we keep everything in a migration would be to do:

```es6
const result = await queryInterface.sequelize.query(...); // The full query
console.log(result);
```

All that's needed is to make sure the last statement of the query returns what you want to see. You can use the `RETURNING` statement on Updates to get their results.

But no blocker for me here.

PR #4711, opened 2020-10-13
Fix missing manual transfer feesOnTop

kewitz:
We had a case where the report was pointing out a higher value on total fees that were being displayed in the details.
Turns out, a collective tipped a donation to another collective using its balance (`pm.service = 'opencollective'`).

### DIFF
--- a/reports/platform-report.js
+++ b/reports/platform-report.js
@@ -152,6 +152,7 @@ async function PlatformReport(year, month) {
           max((CASE WHEN pm."service" = 'paypal' AND t."CollectiveId" = 1 THEN t."amount" ELSE 0 END)::float) as "feeOnTopPaypal",
           max((CASE WHEN (pm."service" = 'stripe' OR spm.service = 'stripe') AND t."CollectiveId" = 1 THEN t."amount" ELSE 0 END)::float) as "feeOnTopStripe",
           max((CASE WHEN t."PaymentMethodId" is NULL AND t."CollectiveId" = 1 THEN t."amount" ELSE 0 END)::float) as "feeOnTopBankTransfers",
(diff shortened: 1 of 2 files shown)


reports/platform-report.js
```diff
@@ -152,6 +152,7 @@ async function PlatformReport(year, month) {
           max((CASE WHEN pm."service" = 'paypal' AND t."CollectiveId" = 1 THEN t."amount" ELSE 0 END)::float) as "feeOnTopPaypal",
           max((CASE WHEN (pm."service" = 'stripe' OR spm.service = 'stripe') AND t."CollectiveId" = 1 THEN t."amount" ELSE 0 END)::float) as "feeOnTopStripe",
           max((CASE WHEN t."PaymentMethodId" is NULL AND t."CollectiveId" = 1 THEN t."amount" ELSE 0 END)::float) as "feeOnTopBankTransfers",
+          max(CASE WHEN t."CollectiveId" = 1 and pm."service" != 'stripe' AND pm."service" != 'paypal' AND (spm.service IS NULL OR spm.service != 'stripe') THEN t."amount" ELSE 0 END)::float as "feeOnTopManual",
```
sbinlondon, line 155: Looks good to me! Do you mind explaining the logic here just to double check?

sbinlondon (approved):

reports/platform-report.js
kewitz: Yes.
These lines are moving the fee amount to a different column based on its payment method.
We do this because we want to categorize and sum these values at the end.
This is the case where the payment method doesn't fit Stripe nor PayPal, meaning it was an internal operation.
This could be more explicit, but for now, I'm OK with just copying the same conditional we're using to calculate revenue.

PR #4712, opened 2020-10-14
Fix small errors in reject contributions cron

sbinlondon:
* Adds second argument for `user` to `libPayments.refundTransaction` so it executes
* Adds full path for contributor rejected categories (`collective.data.categories` instead of `collective.categories`)

### DIFF
--- a/cron/disabled/reject-contributions.js
+++ b/cron/disabled/reject-contributions.js
@@ -23,7 +23,7 @@ const query = `SELECT "Orders"."id"
 
 const getContributorRejectedCategories = (fromCollective, collective) => {
   const rejectedCategories = get(collective, 'settings.moderation.rejectedCategories', []);
-  const contributorCategories = get(fromCollective, 'categories', []);
+  const contributorCategories = get(fromCollective, 'data.categories', []);
 
   if (rejectedCategories.length === 0 || contributorCategories.length === 0) {
     return [];
@@ -103,7 +103,7 @@ async function run({ dryRun = false } = {}) {
           await subscription.deactivate();
         }
       } else {
-        logger.info(`  - Subsription not found`);
+        logger.info(`  - Subscription not found`);
       }
     } else {
       logger.info(`  - No subscription to deactivate`);
--- a/server/models/Transaction.js
+++ b/server/models/Transaction.js
@@ -1,6 +1,6 @@
 import Promise from 'bluebird';
 import debugLib from 'debug';
-import { defaultsDeep, get, isUndefined } from 'lodash';
+import { defaultsDeep, get, isNull, isUndefined } from 'lodash';
 import moment from 'moment';
 import { v4 as uuid } from 'uuid';
 
@@ -47,7 +47,14 @@ export default (Sequelize, DataTypes) => {
         },
         onDelete: 'SET NULL',
         onUpdate: 'CASCADE',
-        allowNull: false,
+        allowNull: true, // we allow CreatedByUserId to be null but only on refund transactions
+        validate: {
+          isValid(value) {
+            if (isNull(value) && this.isRefund === false) {
+              throw new Error('Only refund transactions can have null user.');
+            }
+          },
+        },
       },
 
       // Source of the money for a DEBIT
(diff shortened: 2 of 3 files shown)


cron/disabled/reject-contributions.js
```diff
@@ -50,6 +50,7 @@ async function run({ dryRun = false } = {}) {
     const order = await models.Order.findByPk(row['id']);
     const collective = await models.Collective.findByPk(order.CollectiveId);
     const fromCollective = await models.Collective.findByPk(order.FromCollectiveId);
+    const fromCollectiveUser = await models.User.findByPk(fromCollective.CreatedByUserId);
```
Betree, line 53: `CreatedByUserId` is meant for reporting/logging purposes and should be avoided for this kind of things. Plus with this code we'd be recording in the database that the user who created the collective is the one who triggered the refund, which is not true. A lot of collectives have Pia or Xavier as the `CreatedByUserId` because they helped collectives to get setup, we don't want their names to be associated with the refunds.

What would make sense for the data model would be to allow user to be `null` in `refundTransaction`, which would mean a transaction "created by no user in particular", or rather "created by Open Collective's system". We may need to convert some `CreatedByUserId` columns to nullable to achieve that.

Betree (changes requested):

cron/disabled/reject-contributions.js
sbinlondon: Ok, will change that today.

sbinlondon: This seems like it could be quite a large change for how `refundTransaction` works. @znarf would there be any issue with this?

cron/disabled/reject-contributions.js
```diff
@@ -78,7 +79,7 @@ async function run({ dryRun = false } = {}) {
       if (!transaction.RefundTransactionId) {
         if (!dryRun) {
           logger.info(`  - Refunding transaction`);
-          await libPayments.refundTransaction(transaction);
+          await libPayments.refundTransaction(transaction, fromCollectiveUser);
```
znarf, line 82: @sbinlondon How were you able to catch this error?

cron/disabled/reject-contributions.js
znarf: I agree with @Betree that it should be ok to have a refund without a `CreatedByUserId`

cron/disabled/reject-contributions.js
```diff
@@ -23,7 +23,7 @@ const query = `SELECT "Orders"."id"
 
 const getContributorRejectedCategories = (fromCollective, collective) => {
   const rejectedCategories = get(collective, 'settings.moderation.rejectedCategories', []);
-  const contributorCategories = get(fromCollective, 'categories', []);
+  const contributorCategories = get(fromCollective, 'data.categories', []);
```
znarf, line 26: Good catch!

cron/disabled/reject-contributions.js
sbinlondon: `refundTransaction` doesn't execute if you don't include the user. It does the refund and then calls `createRefundTransaction` which requires the `user` argument to be filled in.

This is why I wanted input on @Betree's suggestion to make `user` nullable https://github.com/opencollective/opencollective-api/pull/4712#discussion_r505210882

znarf: Good catch! Let's make `user` nullable.

sbinlondon:
* `CreatedByUserId` is now nullable for the Transaction model but only for refund transactions
* `refundTransaction` doesn't need a user arg now

server/models/Transaction.js
```diff
@@ -47,7 +47,14 @@ export default (Sequelize, DataTypes) => {
         },
         onDelete: 'SET NULL',
         onUpdate: 'CASCADE',
-        allowNull: false,
+        allowNull: true, // we allow CreatedByUserId to be null but only on refund transactions
+        validate: {
+          isValid(value) {
```
znarf, line 52: I was trying to find the documentation for `isValid` and was unable to find it. I just wanted to check if it was the good practice to throw an exception there.
 
Let me know if you have any link. Otherwise, looks good.

znarf (approved):

Betree (approved):

server/models/Transaction.js
sbinlondon: Not really a Sequelize method but instead a customer validator (bottom of the table): https://sequelize.org/master/manual/validations-and-constraints.html#validators

PR #4713, opened 2020-10-14
Implement rolling host payout limit with 2FA

sbinlondon:
Limit right now is hard coded at $1000.

This implements a rolling limit for each host admin user doing payouts once they're authenticated with 2FA.

```
When processing a payment (without the 2FA code passed as parameter)
 - Check in the "cache" if the key exists. cacheKey=`${user.id}_2fa_payment_limit`;
  - If it doesn't -> 2FA
  - If it does, check its value.
    - If the payment fits in the limit -> process and decrease the limit
    - If the payment doesn't fit in the limit -> 2FA
When processing a payment (with a valid 2FA code passed as parameter):
 - Check in the "cache" if the key exists. cacheKey=`${user.id}_2fa_payment_limit`;
   - If it exists -> reset the limit
   - If it doesn't -> initialize the limit
```

### DIFF
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -5,6 +5,7 @@ import emailLib from '../lib/email';
 import errors from '../lib/errors';
 import logger from '../lib/logger';
 import RateLimit, { ONE_HOUR_IN_SECONDS } from '../lib/rate-limit';
+import { verifyTwoFactorAuthenticatorCode } from '../lib/two-factor-authentication';
 import { isValidEmail } from '../lib/utils';
 import models from '../models';
 
@@ -106,7 +107,7 @@ export const twoFactorAuthAndUpdateToken = async (req, res, next) => {
   }
 
   // we need to verify the 2FA code before returning the token
-  const verified = auth.verifyTwoFactorAuthenticatorCode(user.twoFactorAuthToken, twoFactorAuthenticatorCode);
+  const verified = verifyTwoFactorAuthenticatorCode(user.twoFactorAuthToken, twoFactorAuthenticatorCode);
   if (!verified) {
     return next(new Unauthorized('Two-factor authentication code failed. Please try again'));
   }
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -8,7 +8,8 @@ import statuses from '../../../constants/expense_status';
 import expenseType from '../../../constants/expense_type';
 import FEATURE from '../../../constants/feature';
 import roles from '../../../constants/roles';
-import { enforceTwoFactorAuthenticationOnPayouts } from '../../../lib/auth';
+import { rollingPayoutLimitTwoFactorAuthentication } from '../../../lib/auth';
+import cache from '../../../lib/cache';
 import { getFxRate } from '../../../lib/currency';
 import { floatAmountToCents } from '../../../lib/math';
 import * as libPayments from '../../../lib/payments';
@@ -23,6 +24,7 @@ import * as ExpenseLib from '../../common/expenses';
 import { BadRequest, FeatureNotAllowedForUser, NotFound, Unauthorized, ValidationFailed } from '../../errors';
 
 const debug = debugLib('expenses');
+const CACHE_VALIDITY = 3600; // 1h in secs for cache to expire
 
 export async function updateExpenseStatus(req, expenseId, status) {
   const { remoteUser } = req;
@@ -703,67 +705,94 @@ export async function payExpense(req, args) {
       );
     }
 
-    // If the amount of the expense is larger than i.e. $1000 enforce 2FA if the user has it turned on
-    const is2FARequiredForPayoutMethod = [PayoutMethodTypes.PAYPAL, PayoutMethodTypes.BANK_ACCOUNT].includes(
-      payoutMethodType,
+    // 2FA for payouts
+    const isTwoFactorAuthenticationRequiredForPayoutMethod = [
+      PayoutMethodTypes.PAYPAL,
+      PayoutMethodTypes.BANK_ACCOUNT,
+    ].includes(payoutMethodType);
+    const hostHasPayoutTwoFactorAuthenticationEnabled = get(host, 'settings.payoutsTwoFactorAuth.enabled', false);
+    const hostPayoutTwoFactorAuthenticationRollingLimit = get(
+      host,
+      'settings.payoutsTwoFactorAuth.rollingLimit',
+      1000000,
     );
-    const hostHasPayout2FAEnabled = get(host, 'settings.payoutsTwoFactorAuth.enabled', false);
-    const hostPayout2FAExpenseAmount = get(host, 'settings.payoutsTwoFactorAuth.expenseAmount', 100000);
-    if (
-      is2FARequiredForPayoutMethod &&
+    const useTwoFactorAuthentication =
+      isTwoFactorAuthenticationRequiredForPayoutMethod &&
       !args.forceManual &&
-      hostHasPayout2FAEnabled &&
-      expense.amount >= hostPayout2FAExpenseAmount
-    ) {
-      enforceTwoFactorAuthenticationOnPayouts(req, args.twoFactorAuthenticatorCode);
+      hostHasPayoutTwoFactorAuthenticationEnabled;
+    let currentRollingPayoutLimit;
+    if (useTwoFactorAuthentication) {
+      currentRollingPayoutLimit = await rollingPayoutLimitTwoFactorAuthentication(
+        req,
+        args.twoFactorAuthenticatorCode,
+        hostPayoutTwoFactorAuthenticationRollingLimit,
+        expense.amount,
+      );
     }
 
-    // Pay expense based on chosen payout method
-    if (payoutMethodType === PayoutMethodTypes.PAYPAL) {
-      const paypalEmail = payoutMethod.data.email;
-      let paypalPaymentMethod = null;
-      try {
-        paypalPaymentMethod = await host.getPaymentMethod({ service: 'paypal' });
-      } catch {
-        // ignore missing paypal payment method
-      }
-      // If the expense has been filed with the same paypal email than the host paypal
-      // then we simply mark the expense as paid
-      if (paypalPaymentMethod && paypalEmail === paypalPaymentMethod.name) {
-        feesInHostCurrency.paymentProcessorFeeInHostCurrency = 0;
+    // immediately set the cache
+    if (currentRollingPayoutLimit) {
+      const { cacheKey, cacheValue } = currentRollingPayoutLimit;
+      cache.set(cacheKey, cacheValue, CACHE_VALIDITY);
+    }
+
+    try {
+      // Pay expense based on chosen payout method
+      if (payoutMethodType === PayoutMethodTypes.PAYPAL) {
+        const paypalEmail = payoutMethod.data.email;
+        let paypalPaymentMethod = null;
+        try {
+          paypalPaymentMethod = await host.getPaymentMethod({ service: 'paypal' });
+        } catch {
+          // ignore missing paypal payment method
+        }
+        // If the expense has been filed with the same paypal email than the host paypal
+        // then we simply mark the expense as paid
+        if (paypalPaymentMethod && paypalEmail === paypalPaymentMethod.name) {
+          feesInHostCurrency.paymentProcessorFeeInHostCurrency = 0;
+          await createTransactions(host, expense, feesInHostCurrency);
+        } else if (args.forceManual) {
+          await createTransactions(host, expense, feesInHostCurrency);
+        } else if (paypalPaymentMethod) {
+          await payExpenseWithPayPal(remoteUser, expense, host, paypalPaymentMethod, paypalEmail, feesInHostCurrency);
+        } else {
+          throw new Error('No Paypal account linked, please reconnect Paypal or pay manually');
+        }
+      } else if (payoutMethodType === PayoutMethodTypes.BANK_ACCOUNT) {
+        if (args.forceManual) {
+          feesInHostCurrency.paymentProcessorFeeInHostCurrency = 0;
+          await createTransactions(host, expense, feesInHostCurrency);
+        } else {
+          await payExpenseWithTransferwise(host, payoutMethod, expense, feesInHostCurrency, remoteUser);
+          await expense.setProcessing(remoteUser.id);
+          // Early return, we'll only mark as Paid when the transaction completes.
+          return expense;
+        }
+      } else if (payoutMethodType === PayoutMethodTypes.ACCOUNT_BALANCE) {
+        const payee = expense.fromCollective;
+        const payeeHost = await payee.getHostCollective();
+        if (!payeeHost) {
+          throw new Error('The payee needs to have an Host to able to be paid on its Open Collective balance.');
+        }
+        if (host.id !== payeeHost.id) {
+          throw new Error(
+            'The payee needs to be on the same Host than the payer to be paid on its Open Collective balance.',
+          );
+        }
         await createTransactions(host, expense, feesInHostCurrency);
-      } else if (args.forceManual) {
+      } else if (expense.legacyPayoutMethod === 'manual' || expense.legacyPayoutMethod === 'other') {
+        // note: we need to check for manual and other for legacy reasons
         await createTransactions(host, expense, feesInHostCurrency);
-      } else if (paypalPaymentMethod) {
-        await payExpenseWithPayPal(remoteUser, expense, host, paypalPaymentMethod, paypalEmail, feesInHostCurrency);
-      } else {
-        throw new Error('No Paypal account linked, please reconnect Paypal or pay manually');
       }
-    } else if (payoutMethodType === PayoutMethodTypes.BANK_ACCOUNT) {
-      if (args.forceManual) {
-        feesInHostCurrency.paymentProcessorFeeInHostCurrency = 0;
-        await createTransactions(host, expense, feesInHostCurrency);
-      } else {
-        await payExpenseWithTransferwise(host, payoutMethod, expense, feesInHostCurrency, remoteUser);
-        await expense.setProcessing(remoteUser.id);
-        // Early return, we'll only mark as Paid when the transaction completes.
-        return expense;
+    } catch (error) {
+      // if the payment doesn't go through we want to reset the cache amount
+      if (currentRollingPayoutLimit) {
+        const { cacheKey, cacheValue } = currentRollingPayoutLimit;
+        if (cacheValue !== 0) {
+          cache.set(cacheKey, cacheValue - expense.amount, CACHE_VALIDITY);
+        }
       }
-    } else if (payoutMethodType === PayoutMethodTypes.ACCOUNT_BALANCE) {
-      const payee = expense.fromCollective;
-      const payeeHost = await payee.getHostCollective();
-      if (!payeeHost) {
-        throw new Error('The payee needs to have an Host to able to be paid on its Open Collective balance.');
-      }
-      if (host.id !== payeeHost.id) {
-        throw new Error(
-          'The payee needs to be on the same Host than the payer to be paid on its Open Collective balance.',
-        );
-      }
-      await createTransactions(host, expense, feesInHostCurrency);
-    } else if (expense.legacyPayoutMethod === 'manual' || expense.legacyPayoutMethod === 'other') {
-      // note: we need to check for manual and other for legacy reasons
-      await createTransactions(host, expense, feesInHostCurrency);
+      return error;
     }
 
     return markExpenseAsPaid(expense, remoteUser);
--- a/server/graphql/v1/mutations/expenses.js
+++ b/server/graphql/v1/mutations/expenses.js
@@ -8,13 +8,15 @@ import statuses from '../../../constants/expense_status';
 import expenseType from '../../../constants/expense_type';
 import FEATURE from '../../../constants/feature';
 import roles from '../../../constants/roles';
-import { rollingPayoutLimitTwoFactorAuthentication } from '../../../lib/auth';
-import cache from '../../../lib/cache';
 import { getFxRate } from '../../../lib/currency';
 import { floatAmountToCents } from '../../../lib/math';
 import * as libPayments from '../../../lib/payments';
 import { handleTransferwisePayoutsLimit } from '../../../lib/plans';
 import { createFromPaidExpense as createTransactionFromPaidExpense } from '../../../lib/transactions';
+import {
+  handleTwoFactorAuthenticationPayoutLimit,
+  resetRollingPayoutLimitOnFailure,
+} from '../../../lib/two-factor-authentication';
 import { canUseFeature } from '../../../lib/user-permissions';
 import { formatCurrency } from '../../../lib/utils';
 import models, { sequelize } from '../../../models';
@@ -24,7 +26,6 @@ import * as ExpenseLib from '../../common/expenses';
 import { BadRequest, FeatureNotAllowedForUser, NotFound, Unauthorized, ValidationFailed } from '../../errors';
 
 const debug = debugLib('expenses');
-const CACHE_VALIDITY = 3600; // 1h in secs for cache to expire
 
 export async function updateExpenseStatus(req, expenseId, status) {
   const { remoteUser } = req;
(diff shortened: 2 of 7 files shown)


sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3516

server/lib/auth.js
```diff
@@ -93,16 +95,60 @@ export function verifyTwoFactorAuthenticatorCode(encryptedTwoFactorAuthToken, tw
...
+  twoFactorAuthenticatorCode,
+  hostRollingLimit,
+  expenseAmount,
+) {
   if (req.remoteUser.twoFactorAuthToken !== null) {
+    let cacheObject;
+    // 1. we check the 'cache' if the key exists: cacheKey=${user.id}_2fa_payment_limit
+    const userId = req.remoteUser.id;
+    const cacheKey = getCacheKey(userId);
+    const keyInCache = await checkCacheForKey(cacheKey);
```
Betree, line 116: Why not directly?

```suggestion
    const keyInCache = await cache.get(key);
```

server/lib/auth.js
```diff
@@ -93,16 +95,60 @@ export function verifyTwoFactorAuthenticatorCode(encryptedTwoFactorAuthToken, tw
...
+
     if (twoFactorAuthenticatorCode) {
       const verified = verifyTwoFactorAuthenticatorCode(req.remoteUser.twoFactorAuthToken, twoFactorAuthenticatorCode);
       if (!verified) {
         throw new Error('Two-factor authentication failed: invalid code. Please try again.');
       }
-      return;
+      // With 2FA code
+      // 2. if limit key exists, reset the limit
+      // 3. if limit key doesn't exist, initialise the limit - are these the same???
```
Betree, line 125: It seems so?

```suggestion
      // 3. if limit key doesn't exist, initialise the limit
```

Betree (changes requested):
I think adding some tests on that part is pretty important.

My main concern right now is that this limitation can easily be bypassed by just submitting all the payment requests at the same time (concurrency issue). There will usually be a few seconds between the moment when you check the cache and when you set the updated value (because calling the payment processor takes time), this can be used to trick the limitation.

I tried to represent the attack in the diagram below. Scenario: 
- Still a host admin account
- Wait for on of the admins to start paying an expense
- Once they've paid at least one, send multiple requests for payment at once
- Requests run in parallel, the cache doesn't account for the other expenses being paid at the same time when it checks the amount


Untitled Diagram

sbinlondon:
@Betree Interesting. So would I need to do a lock thing like @kewitz wrote recently for another part of the API (transactions I think?)

Betree:
Introducing a lock is indeed the proper way to solve this problem. Most of the locks that we have in our code today (that I'm aware of) are database locks, which are not what we want here. 

It may be worth looking at libraries like https://github.com/rogierschouten/async-lock or similar, would be great to do a bit of research to see what are the best practices in this field nowadays.

znarf:
@betree @sbinlondon If we check **and update** the rolling limit "as fast as possible" ("as soon as the request arrives" and before we trigger the payment) then we don't need a lock to handle the concurrency issue. It's really unlikely that we will face a race condition when checking the limit.

znarf:
@Betree Ok, if locking, I recommend to lock using a Redis key using our "cache" library. 

I originally was about to just write that, then realized the timing issues would be exactly the same as using the rolling limit.

Just wanted to give that input, trusting you do to the right choice!

Betree (changes requested):
To solve the problem, the `cache.get(...)` (`rollingPayoutLimitTwoFactorAuthentication`) neds to be included in the lock. Taking the example above, this is what we have now:

```es6
// With base cache = $50, each req is a payment for 400$
rollingPayoutLimitTwoFactorAuthentication(); // Req #1 - => $50
rollingPayoutLimitTwoFactorAuthentication(); // Req #2 - => $50
rollingPayoutLimitTwoFactorAuthentication(); // Req #3 - => $50

// Lock for req #1
lock(() => {
  pay();
  setCache($450);
})

// Lock for req #2
lock(() => {
  pay();
  setCache($450);
})

// Lock for req #3
lock(() => {
  pay();
  setCache($450);
})
```

sbinlondon:
Ok, I'll work on that today! Does the code look okay otherwise? i.e. the generator function for a cache lock in the cache library

sbinlondon:
Ok, I'll work on that today! Does the code look okay otherwise? i.e. the generator function for a cache lock in the cache library @Betree?

Betree (commented):
I just checked the code for https://github.com/rogierschouten/async-lock and I actually think it won't work. We're now using multiple workers in production, which means they all have their own internal memory and don't share variables. So each instance will have its own lock, which defeats the purpose of the lib for us.

After some research, it seems that the best practice nowadays is to use a redis lock: https://github.com/mike-marcacci/node-redlock.

It seems a bit more complex to introduce though, so I suggest:
- Implementing the mitigation proposed by @znarf [above](https://github.com/opencollective/opencollective-api/pull/4713#issuecomment-709340395)
- Opening an issue in https://github.com/opencollective/opencollective-security to document what we've talked about here

sbinlondon:
I will do that.

The reason I didn't trigger setting the cache immediately, and waited til after the payment, is in case the payment doesn't go through for some reason. Are we okay with that for now while we sort out the Redis lock?

Betree:
We can always restore the previous cache if something goes wrong

```es6
cache.set(cache.get() + amount);
try {
   pay();
} catch {
  cache.set(cache.get() - amount);
}
```

sbinlondon:
@kewitz All the tests passed fine locally but on the CI I am getting a Transferwise error. Do you know what the issue could be?

Screenshot 2020-10-16 at 13 37 59

server/graphql/v1/mutations/expenses.js
```diff
@@ -703,67 +705,93 @@ export async function payExpense(req, args) {
...
-      if (host.id !== payeeHost.id) {
-        throw new Error(
-          'The payee needs to be on the same Host than the payer to be paid on its Open Collective balance.',
-        );
+    } catch (error) {
+      // if the payment doesn't go through we want to reset the cache amount
+      if (currentRollingPayoutLimit) {
+        const { cacheKey, cacheValue } = currentRollingPayoutLimit;
+        const resetCacheValue = cacheValue === 0 ? cacheValue : cacheValue - expense.amount;
+        cache.set(cacheKey, resetCacheValue, CACHE_VALIDITY);
```
kewitz, line 792: Maybe we should simplify the ternary with an explicit conditional:
```suggestion
        if (cacheValue !== 0) {
          cache.set(cacheKey, cacheValue - expense.amount, CACHE_VALIDITY);
        }
```

server/lib/auth.js
```diff
@@ -93,16 +95,56 @@ export function verifyTwoFactorAuthenticatorCode(encryptedTwoFactorAuthToken, tw
...
+};
+
+export async function rollingPayoutLimitTwoFactorAuthentication(
+  req,
+  twoFactorAuthenticatorCode,
+  hostRollingLimit,
+  expenseAmount,
+) {
   if (req.remoteUser.twoFactorAuthToken !== null) {
+    let cacheObject;
```
kewitz, line 109: If we're returning this object on every conditional, we don't need to flag `cacheObject`.

server/lib/auth.js
```diff
@@ -93,16 +95,56 @@ export function verifyTwoFactorAuthenticatorCode(encryptedTwoFactorAuthToken, tw
...
+        const runningTotal = keyInCache + expenseAmount;
+        const expenseExceedsRollingLimit = runningTotal > hostRollingLimit;
+        if (expenseExceedsRollingLimit) {
+          throw new Error('Two-factor authentication payout limit exceeded: please re-enter your code.');
+        } else {
+          cacheObject = {
+            cacheKey,
+            cacheValue: runningTotal,
+          };
+          return cacheObject;
```
kewitz, line 145: Perhaps just return the object here:
```suggestion
          return { cacheKey, cacheValue: runningTotal };
```

test/server/graphql/v2/mutation/ExpenseMutations.test.js
```diff
@@ -753,4 +782,158 @@ describe('server/graphql/v2/mutation/ExpenseMutations', () => {
...
     });
   });
+
+  describe('processExpense > PAY > with 2FA payouts', () => {
+    const fee = 1.74;
+    let collective, host, collectiveAdmin, hostAdmin, sandbox, expense1, expense2, expense3, expense4, user;
+
+    before(() => {
+      sandbox = sinon.createSandbox();
+      sandbox.stub(paymentProviders.transferwise, 'payExpense').resolves({ quote: { fee } });
```
kewitz, line 792: You'll also need to stub `paymentProviders.transferwise.getTemporaryQuote` because of line `server/graphql/v1/mutations/expenses.js:672`:
```suggestion
      sandbox.stub(paymentProviders.transferwise, 'payExpense').resolves({ quote: { fee } });
      sandbox.stub(paymentProviders.transferwise, 'getTemporaryQuote').resolves({ fee });

**Review by kewitz — COMMENTED** [review]
Looks good, nice work!
I added a few suggestions related to simplification, feel free to skip if you don't agree.
Also, you forgot to stub one function used in the `payExpense` mutation, that should fix your tests.

File: server/lib/auth.js [hunk 7]
```diff
@@ -93,16 +95,53 @@ export function verifyTwoFactorAuthenticatorCode(encryptedTwoFactorAuthToken, tw
...
       const verified = verifyTwoFactorAuthenticatorCode(req.remoteUser.twoFactorAuthToken, twoFactorAuthenticatorCode);
       if (!verified) {
         throw new Error('Two-factor authentication failed: invalid code. Please try again.');
       }
-      return;
+      // With 2FA code
+      // 2. if limit key exists, reset the limit; if limit key doesn't exist, initialise the limit
+      return {
+        cacheKey,
+        cacheValue: 0,
```
Betree, line 123: Shouldn't that be `expenseAmount`? Otherwise the cache will always be at `0` when you pay the first expense

```suggestion
        cacheValue: expenseAmount,
```

Betree (commented):
The new approach looks good, it would be interesting to try an exploit the concurrency issue once we have it on staging to make sure that it's indeed prevented (or at least limited).

I'm good to merge the PR because it's a nice improvement right now, but I want to bring attention on the fact that this issue may lower the security of 2FA in the future. 2FA is there to protect you in case your account get stolen, and this feature provides a way to bypass this validation.

For example, let's imagine that an attacker finds a way to create and pay expense without having them appearing in the dashboard (ie. by having a script that approves + pays really quick). All they'd have to do is to synchronize their attack to do it when the host admin is paying the expenses, and they'll rely on the fact that host admins will keep entering the tokens when asked.

server/lib/auth.js
sbinlondon: That's my understanding of what should happen from @znarf's details about how to write this: https://github.com/opencollective/opencollective/issues/3516

"allow payment and allow a follow up budget of $10,0000"

server/lib/auth.js
```diff
@@ -93,16 +95,53 @@ export function verifyTwoFactorAuthenticatorCode(encryptedTwoFactorAuthToken, tw
   return verified;
 }
 
-export function enforceTwoFactorAuthenticationOnPayouts(req, twoFactorAuthenticatorCode) {
+const getCacheKey = userId => {
```
znarf, line 98: If it's used only in one place, maybe we can just construct the key without having a function to do it.

znarf (changes requested):
I would avoid naming variables such as `cacheKey` and `cacheAmount`, they're not technically caching anything.

Names such as `2faLimitKey` and `2faLimitAmount` would make more sense for me.

server/lib/auth.js
znarf: I'm not following. 

If the user enter a correct 2FA code, then the limit should be initialized to `hostRollingLimit`. Do I get something wrong?

server/lib/auth.js
```diff
@@ -93,16 +95,53 @@ export function verifyTwoFactorAuthenticatorCode(encryptedTwoFactorAuthToken, tw
...
+  req,
+  twoFactorAuthenticatorCode,
+  hostRollingLimit,
+  expenseAmount,
+) {
   if (req.remoteUser.twoFactorAuthToken !== null) {
+    // 1. we check the 'cache' if the key exists: cacheKey=${user.id}_2fa_payment_limit
+    const userId = req.remoteUser.id;
+    const cacheKey = getCacheKey(userId);
+    const keyInCache = await cache.get(cacheKey);
```
znarf, line 112: Use a more describing variable name (ideally without cache), as it's used later in `const runningTotal = keyInCache + expenseAmount;`

server/lib/auth.js
znarf: Scratch my previous comment. I was confused by the meaning of `cacheValue` (this is why we should have a better name for it).

Because it's containing the sum of all expenses processed, setting it to `0` make sense to me.

server/lib/auth.js
```diff
@@ -93,16 +95,53 @@ export function verifyTwoFactorAuthenticatorCode(encryptedTwoFactorAuthToken, tw
   return verified;
 }
 
-export function enforceTwoFactorAuthenticationOnPayouts(req, twoFactorAuthenticatorCode) {
+const getCacheKey = userId => {
+  return `${userId}_2fa_payment_limit`;
+};
+
+export async function rollingPayoutLimitTwoFactorAuthentication(
+  req,
```
znarf, line 103: Maybe it's clearer to have `user` (remoteUser) instead of `req` as argument.

server/graphql/v1/mutations/expenses.js
```diff
@@ -703,67 +705,94 @@ export async function payExpense(req, args) {
...
-        // ignore missing paypal payment method
-      }
-      // If the expense has been filed with the same paypal email than the host paypal
-      // then we simply mark the expense as paid
-      if (paypalPaymentMethod && paypalEmail === paypalPaymentMethod.name) {
-        feesInHostCurrency.paymentProcessorFeeInHostCurrency = 0;
-        await createTransactions(host, expense, feesInHostCurrency);
-      } else if (args.forceManual) {
+    // immediately set the cache
+    if (currentRollingPayoutLimit) {
```
znarf, line 734: If we're there, it means the payment is 2FA authorized.

Why are we not setting the value already within the `rollingPayoutLimitTwoFactorAuthentication` method?

server/graphql/v1/mutations/expenses.js
```diff
@@ -703,67 +705,94 @@ export async function payExpense(req, args) {
...
-      if (!payeeHost) {
-        throw new Error('The payee needs to have an Host to able to be paid on its Open Collective balance.');
       }
-      if (host.id !== payeeHost.id) {
-        throw new Error(
-          'The payee needs to be on the same Host than the payer to be paid on its Open Collective balance.',
-        );
+    } catch (error) {
+      // if the payment doesn't go through we want to reset the cache amount
+      if (currentRollingPayoutLimit) {
```
znarf, line 789: Maybe create an abstraction for that next to the `rollingPayoutLimitTwoFactorAuthentication` method.

znarf: Maybe it would make sense to have something like `lib/2fa.js` instead of putting these methods in `lib/auth.js`.

server/lib/auth.js
Betree: But the value returned by this helper is directly passed to `cache.set(cacheKey, cacheValue, CACHE_VALIDITY);`. So that means if I pay an expense for $500 while providing a valid 2FA code, the state of the cache after paying the expense will be $0.

Betree: Bigger problem: if I try to pay a $500 expense while providing a valid 2FA code, then it fails, my cache state will be -$500

znarf: I guess we need to be more precise in the specification and expectations. To make sure the code is doing what we actually want.

sbinlondon: > Bigger problem: if I try to pay a $500 expense while providing a valid 2FA code, then it fails, my cache state will be -$500

Yes, that's why I initially wrote the code so that the cache was not updated until *after* the payment was processed. Then we did all the cache locking. Then we discarded that.

So now based on the fact that we aren't going to implement cache locking now, we update the cache immediately and in the case the payment fails, we reset it to what it was before: https://github.com/opencollective/opencollective-api/blob/4b5ef9f6cbcb677e62c96ffbf7c4d6d72449e6e9/server/graphql/v1/mutations/expenses.js#L788

sbinlondon: > But the value returned by this helper is directly passed to cache.set(cacheKey, cacheValue, CACHE_VALIDITY);. So that means if I pay an expense for $500 while providing a valid 2FA code, the state of the cache after paying the expense will be $0.

> I guess we need to be more precise in the specification and expectations. To make sure the code is doing what we actually want.

Ok my understanding (again from @znarf's comment in the issue linked above) is that, given 4 payments to process that are $100, $200, $250, $300:

* I log in to make payments as a host admin. The host collective rolling limit is $500.
* I go to make my first payment, which is $100. I am asked for my 2FA code, and I enter it. The cache is created with a value of $0 out of $500 and the payment is processed. This first payment is **not** included in the limit.
* I make payment 2 of $200, the cache is now at $200 out of $500, the payment is processed.
* I make payment 3 of $250, the cache is now at $450 out of $500, the payment is processed.
* I make payment 4 of $300, the cache is now at $750 out of $500 - so we ask for 2FA again. I enter the code, the cache is reset to $0 out of $500, the payment is processed, etc

If this isn't everyone's understanding then can we use that example above and describe how you would expect it to work?

server/graphql/v1/mutations/expenses.js
sbinlondon: Probably a leftover from when the PR had cache locking. I can move it back.

sbinlondon: A leftover from when there was cache locking in the PR. There was a big overhaul, maybe some of the structure doesn't make sense anymore. I'll move it.

server/graphql/v1/mutations/expenses.js
sbinlondon: An abstraction how? It's just setting the cache amount again.

server/lib/auth.js
znarf: Thank you for taking the time to write this example, the behavior it describes looks good to me.

server/controllers/users.js
```diff
@@ -5,6 +5,7 @@ import emailLib from '../lib/email';
 import errors from '../lib/errors';
 import logger from '../lib/logger';
 import RateLimit, { ONE_HOUR_IN_SECONDS } from '../lib/rate-limit';
+import { verifyTwoFactorAuthenticatorCode } from '../lib/twoFactorAuth';
```
znarf, line 8: We generally use kebab case for file names.
https://github.com/opencollective/opencollective/blob/master/rfcs/004-file-naming-convention.md

server/graphql/v1/mutations/expenses.js
```diff
@@ -703,67 +704,91 @@ export async function payExpense(req, args) {
...
-    } else if (payoutMethodType === PayoutMethodTypes.ACCOUNT_BALANCE) {
-      const payee = expense.fromCollective;
-      const payeeHost = await payee.getHostCollective();
-      if (!payeeHost) {
-        throw new Error('The payee needs to have an Host to able to be paid on its Open Collective balance.');
-      }
-      if (host.id !== payeeHost.id) {
-        throw new Error(
-          'The payee needs to be on the same Host than the payer to be paid on its Open Collective balance.',
+    } catch (error) {
```
znarf, line 780: I still think an abstraction can help better structure the code here. That way, key construction and cache validity are kept out in the twoFactorAuthentication library.

Something like:

```
if (useTwoFactorAuthentication) {
  await twoFactorAuthentication.updatePayoutLimitOnFailure(user, expense)
}
```

"updatePayoutLimitOnFailure" might not be the best name but it's late I'm not inspired.

server/controllers/users.js
znarf: Also let's generally keep names consistent, it can't be "twoFactorAuth" then "twoFactorAuthentication" elsewhere.

server/graphql/v1/mutations/expenses.js
```diff
@@ -703,67 +704,91 @@ export async function payExpense(req, args) {
...
-      expense.amount >= hostPayout2FAExpenseAmount
-    ) {
-      enforceTwoFactorAuthenticationOnPayouts(req, args.twoFactorAuthenticatorCode);
+      hostHasPayoutTwoFactorAuthenticationEnabled;
+
+    if (useTwoFactorAuthentication) {
+      await rollingPayoutLimitTwoFactorAuthentication(
+        req.remoteUser,
+        args.twoFactorAuthenticatorCode,
+        hostPayoutTwoFactorAuthenticationRollingLimit,
```
znarf, line 727: I think we might be able to figure this value (`hostPayoutTwoFactorAuthenticationRollingLimit`) within the `rollingPayoutLimitTwoFactorAuthentication`.

```
const host = expense.collective.getHostCollective();
const hostPayoutTwoFactorAuthenticationRollingLimit = get(host, 'settings.payoutsTwoFactorAuth.rollingLimit', 1000000);
```

server/graphql/v1/mutations/expenses.js
```diff
@@ -703,67 +704,91 @@ export async function payExpense(req, args) {
...
+        expense.amount,
```
znarf, line 728: Passing the `expense` instead of `expense.amount` is more generic and future proof.

This would also help figure the `host` and compute `hostPayoutTwoFactorAuthenticationRollingLimit` within `rollingPayoutLimitTwoFactorAuthentication`

server/graphql/v1/mutations/expenses.js
```diff
@@ -703,67 +704,91 @@ export async function payExpense(req, args) {
...
+      isTwoFactorAuthenticationRequiredForPayoutMethod &&
       !args.forceManual &&
-      hostHasPayout2FAEnabled &&
-      expense.amount >= hostPayout2FAExpenseAmount
-    ) {
-      enforceTwoFactorAuthenticationOnPayouts(req, args.twoFactorAuthenticatorCode);
+      hostHasPayoutTwoFactorAuthenticationEnabled;
+
+    if (useTwoFactorAuthentication) {
+      await rollingPayoutLimitTwoFactorAuthentication(
```
znarf, line 724: Not a blocker but I have a hard time understanding the purpose of this method by looking at its name. I think a verb is missing.

Maybe `handleTwoFactorAuthenticationPayoutLimit`. Or `twoFactorAuthentication.handlePayoutLimit` when using another import/export style.

package-lock.json
```diff
@@ -6989,7 +6989,7 @@
     },
     "csv-stringify": {
       "version": "1.0.4",
-      "resolved": "http://registry.npmjs.org/csv-stringify/-/csv-stringify-1.0.4.tgz",
+      "resolved": "https://registry.npmjs.org/csv-stringify/-/csv-stringify-1.0.4.tgz",
```
znarf, line 6992: Should not be modified.

znarf (changes requested):
Thank you for your patience and for implementing the latest batch of changes. It's looking clearer now.

A few more changes and we should be good to go!

package-lock.json
sbinlondon: Ahhhhhh my bad, it's late, didn't mean to commit that!

server/lib/auth.js
sbinlondon: No worries, just wanted to make sure we were on the same page! I know that it could work a few different ways.

server/graphql/v1/mutations/expenses.js
sbinlondon: Yes, good point! Will move.

server/lib/two-factor-authentication.js
```diff
@@ -0,0 +1,80 @@
...
+        }
+      }
+    }
+  } else {
+    throw new Error('Host has two-factor authentication enabled for large payouts.');
+  }
+}
+
+export async function resetRollingPayoutLimitOnFailure(user, expense) {
+  const twoFactorAuthLimitKey = `${user.id}_2fa_payment_limit`;
```
znarf, line 72: I've given the feedback earlier to remove the cache key generation utility because it was only used once.

Now that the `twoFactorAuthLimitKey` is used in 2 different places, I think it's a cleaner practice to re-introduce it.

server/lib/two-factor-authentication.js
```diff
@@ -0,0 +1,80 @@
...
+  const twoFactorAuthLimitAmountForUser = await cache.get(twoFactorAuthLimitKey);
+
+  if (twoFactorAuthLimitAmountForUser !== 0) {
```
znarf, line 75: What if `twoFactorAuthLimitAmountForUser` is `null`? It's a possibility.

server/lib/two-factor-authentication.js
```diff
@@ -0,0 +1,80 @@
...
+  return verified;
+}
+
+export async function handleTwoFactorAuthenticationPayoutLimit(user, twoFactorAuthenticatorCode, expense) {
+  if (user.twoFactorAuthToken !== null) {
+    const host = expense.collective.getHostCollective();
+    const hostPayoutTwoFactorAuthenticationRollingLimit = get(
+      host,
+      'settings.payoutsTwoFactorAuth.rollingLimit',
+      1000000,
```
znarf, line 31: The PR says that the default is 1,000, here it's 10,000. Not sure which one is right.

sbinlondon: The default for our basic "if payment is > $1000, trigger 2FA".

On the issue https://github.com/opencollective/opencollective/issues/3516 we say rolling limit of $10,000.

server/lib/two-factor-authentication.js
```diff
@@ -0,0 +1,87 @@
...
+        }
+      }
+    }
+  } else {
+    throw new Error('Host has two-factor authentication enabled for large payouts.');
+  }
+}
+
+export async function resetRollingPayoutLimitOnFailure(user, expense) {
+  const twoFactorAuthenticationLimitKey = getTwoFactorAuthenticationLimitKey(user.id);
```
znarf, line 75: It doesn't need to be changed here, but in the future you will generally make your code more future proof by passing full object `user` vs just `user.id`.

znarf (approved):
Looks good!

PR #4716, opened 2020-10-16
Platform tips refunds

Betree:
Followup for another PR:
- Create a migration to mark tips as refunded (and create the opposite transaction) if the base transaction has been refunded

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -3,7 +3,6 @@ import Promise from 'bluebird';
 import config from 'config';
 import debugLib from 'debug';
 import { find, get, includes, omit, pick } from 'lodash';
-import { Op } from 'sequelize';
 
 import activities from '../constants/activities';
 import status from '../constants/order_status';
@@ -13,7 +12,7 @@ import { FEES_ON_TOP_TRANSACTION_PROPERTIES, OC_FEE_PERCENT } from '../constants
 import { createPrepaidPaymentMethod, isPrepaidBudgetOrder } from '../lib/prepaid-budget';
 import { formatAccountDetails } from '../lib/transferwise';
 import { formatCurrency } from '../lib/utils';
-import models from '../models';
+import models, { Op } from '../models';
 import paymentProviders from '../paymentProviders';
 
 import emailLib from './email';
@@ -175,7 +174,7 @@ export async function createRefundTransaction(transaction, refundedPaymentProces
     ]);
     refund.CreatedByUserId = user.id;
     refund.description = `Refund of "${t.description}"`;
-    refund.data = data;
+    refund.data = { ...refund.data, ...data };
 
     /* The refund operation moves back fees to the user's ledger so the
      * fees there should be positive. Since they're usually in negative,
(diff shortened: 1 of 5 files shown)


kewitz:
@Betree I changed to a relation to the `TransactionGroup`, I think it represents a better relationship since the donation is tied to the transaction pair.

server/lib/payments.js
```diff
@@ -205,7 +204,11 @@ export async function createRefundTransaction(transaction, refundedPaymentProces
 
   if (transaction.data?.isFeesOnTop) {
     const feeOnTopTransaction = await models.Transaction.findOne({
-      where: { ...FEES_ON_TOP_TRANSACTION_PROPERTIES, type: 'CREDIT' },
```
Betree, line 208: Are there any erroneous transactions introduced by this query that we would need to fix in the DB?

Betree (commented):
I agree with the transactionGroup change, it's better!

Updating the existing migration rather than writing a new one would have been a better strategy, only two lines needed to be changed to achieve the new behavior and it was properly tested.

*Edit: pasting here the code from the previous migration since the code has been amended:*

<details>

```sql
UPDATE
  "Transactions"
SET
  "PlatformTipForTransactionId" = transaction_links.transaction_id
FROM (
  SELECT
    t.id AS transaction_id,
    platform_tip.id AS platform_tip_id
  FROM
    "Transactions" t
  INNER JOIN "Transactions" platform_tip ON (
    t."OrderId" = platform_tip."OrderId"
    AND t."type" = platform_tip."type"
    AND t.id != platform_tip.id 
  )
  INNER JOIN "PaymentMethods" pm ON
    t."PaymentMethodId" = pm.id
  LEFT JOIN "PaymentMethods" spm ON
    pm."SourcePaymentMethodId" = spm.id
  WHERE
    -- platform_tip: Only transactions on "opencollective" (#1)
    ((platform_tip.type = 'CREDIT' AND platform_tip."CollectiveId" = 1) OR (platform_tip.type = 'DEBIT' AND platform_tip."FromCollectiveId" = 1))
    AND (t."FromCollectiveId" != 1 AND t."CollectiveId" != 1)
    AND (t.data ->> 'isFeesOnTop')::boolean = TRUE
    AND (
      -- Stripe
      (
        ((pm.service = 'stripe' AND pm.type = 'creditcard') OR (spm.service = 'stripe' AND spm.type = 'creditcard'))
        AND (platform_tip.data -> 'charge' ->> 'id') = (t.data -> 'charge' ->> 'id')
      )
      -- PayPal
      OR (pm.service = 'paypal' AND pm.type = 'payment' AND (platform_tip.data ->> 'id') = (t.data ->> 'id'))
      -- Collective PM
      OR (pm.service = 'opencollective' AND pm.type = 'collective' )
    )
  GROUP BY t.id, platform_tip.id
) AS transaction_links
WHERE
  "Transactions".id = transaction_links.platform_tip_id
```

</details>

server/lib/payments.js
kewitz: Nope, just the one we manually fixed last week (currently soft-deleted in our DB).

PR #4720, opened 2020-10-19
Improve updateOrder v2 mutation

sbinlondon:
* Update Subscription attached to Order when changing the Order amount
* Remove unused code
* Make sure that if changing to a FIXED Tier, the Order amount _must_ match the Tier's fixed amount (source of the bug last week)
* Adds tests for Order v2 mutations


server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -287,9 +235,15 @@ const orderMutations = {
           throw new Error('Amount is less than minimum value allowed for this Tier.');
         }
 
-        // check if the amount is different from the previous amount
+        // If using a FIXED tier, amount cannot be different from the tier's amount
+        if (tierInfo && tierInfo.amountType === 'FIXED' && amountInCents < tierInfo.amount) {
```
Betree, line 239: Sounds like this should be:

```suggestion
        if (tierInfo && tierInfo.amountType === 'FIXED' && amountInCents !== tierInfo.amount) {
```

However, I'm wondering how this would play with platform tips and taxes (VAT, GST). So maybe your solution is a good way to solve that without too much complication.

test/test-helpers/fake-data.js
```diff
@@ -360,6 +360,16 @@ export const fakeOrder = async (orderData = {}) => {
     order.paymentMethod = await models.PaymentMethod.findByPk(order.PaymentMethodId);
   }
 
+  if (withSubscription) {
```
Betree, line 363: Nice change!

Betree (approved):
Seeing old code go away always feels oddly satisfying

server/graphql/v2/mutation/OrderMutations.js
sbinlondon: Good catch, will change!

Betree: @sbinlondon If you do, make sure you properly test platform tips and tax form. Otherwise we can keep it as it is.

PR #4723, opened 2020-10-19
Attach single transaction PDF to thank you email

sbinlondon:
Fetches the PDF of an Order's Transaction from the PDF service and attaches it to the Order thank you email.

### DIFF
--- a/config/production.json
+++ b/config/production.json
@@ -20,7 +20,7 @@
     "frontend": "https://frontend.opencollective.com",
     "website": "https://opencollective.com",
     "images": "https://images.opencollective.com",
-    "invoices": "https://invoices.opencollective.com"
+    "pdf": "https://pdf.opencollective.com"
   },
   "log": {
     "slowRequest": true
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -11,11 +11,12 @@ import tiers from '../constants/tiers';
 import { FEES_ON_TOP_TRANSACTION_PROPERTIES, OC_FEE_PERCENT } from '../constants/transactions';
 import { createPrepaidPaymentMethod, isPrepaidBudgetOrder } from '../lib/prepaid-budget';
 import { formatAccountDetails } from '../lib/transferwise';
-import { formatCurrency } from '../lib/utils';
+import { formatCurrency, toIsoDateStr } from '../lib/utils';
 import models, { Op } from '../models';
 import paymentProviders from '../paymentProviders';
 
 import emailLib from './email';
+import { getTransactionPdf } from './pdf';
 import { subscribeOrUpgradePlan, validatePlanRequest } from './plans';
 import { getNextChargeAndPeriodStartDates } from './recurring-contributions';
 import { netAmount } from './transactions';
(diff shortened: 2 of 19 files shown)


sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/2867

config/production.json
```diff
@@ -20,7 +20,7 @@
     "frontend": "https://frontend.opencollective.com",
     "website": "https://opencollective.com",
     "images": "https://images.opencollective.com",
-    "invoices": "https://invoices.opencollective.com"
+    "invoices": "https://pdf.opencollective.com"
```
Betree, line 23: If it's not disruptive (otherwise we can keep that for later) we should also rename the service itself in the config:

```suggestion
    "pdf": "https://pdf.opencollective.com"
```

server/lib/pdf.js
```diff
@@ -0,0 +1,24 @@
...
+export const getTransactionPdf = async (transaction, user) => {
+  if (['ci', 'test'].includes(config.env)) {
+    return;
+  }
+  const pdfUrl = `${config.host.invoices}/transactions/${transaction.uuid}/invoice.pdf`;
+  const accessToken = user.jwt({}, TOKEN_EXPIRATION_LOGIN);
+  const headers = {
+    Authorization: `Bearer ${accessToken}`,
+  };
+  return await fetch(pdfUrl, { method: 'get', headers }).then(response => {
```
Betree, line 15: If the remote server doesn't reply, `fetch` will throw an error (try `fetch('https://xxxxxjuk846yk78y4k6i84k354etw5t4xx.com/qqqqq').then(resp => console.log(resp))`). 

We should properly handle that to make sure we still send the thank you email, even if there's to PDF attached.

server/lib/payments.js
```diff
@@ -426,6 +425,23 @@ const sendOrderConfirmedEmail = async order => {
...
+    pdf = await getTransactionPdf(transaction, user);
+
+    // attach pdf
+    if (pdf) {
+      const createdAtString = toIsoDateStr(transaction.createdAt ? new Date(transaction.createdAt) : new Date());
+      attachments.push({
+        filename: `transaction_${collective.slug}_${createdAtString}_${transaction.uuid}.pdf`,
+        content: pdf,
+      });
+      data.transactionPdf = true;
```
Betree, line 438: Good idea to set a flag for this to make it optional in the template :100:

Betree (commented):
Works great!

The `thankyou` email has been customized for multiple collectives: `ls templates/emails/thankyou.*`, in particular `foundation` and `opensource`. We should update these templates as well. For `templates/emails/thankyou.fr.hbs`, you can use:
```hbs
{{#if transactionPdf}}
<h2>📎 Pièces jointes</h2>
<ul>
  <li>Un reçu de votre transaction au format PDF</li>
</ul>
{{/if}}
```

My main worry right now is that the PDF service may not be able to catch up with the load. I'll monitor to react quickly, but we must make sure that the email is still sent even if there's a problem with the `fetch`.

server/lib/pdf.js
```diff
@@ -0,0 +1,28 @@
...
+      const { status } = response;
+      if (status >= 200 && status < 300) {
+        return response.body;
+      } else {
+        console.warn('Failed to fetch PDF');
+        return null;
+      }
+    })
+    .catch(error => {
+      console.error(`Error fetching PDF: ${error}`);
```
Betree, line 26: It's better to use `logger` on the API. You can import it from `server/lib/logger`.

```suggestion
      logger.error(`Error fetching PDF: ${error}`);
```

Betree (commented):
Just a small feedback about using `logger`, the rest looks perfect

PR #4726, opened 2020-10-20
Add HOST membership when activating budget

znarf:
Related: https://github.com/opencollective/opencollective/issues/3584


Betree (approved):
Nice!

PR #4728, opened 2020-10-20
HTML content: replace all external links by redirects

Betree:
An alternative to https://github.com/opencollective/opencollective-api/pull/4633 as the first keeps failing for unclear reasons


znarf (commented):
Can we have a list of domains we want to allow by default?

Betree:
@znarf which ones do we want to whitelist?

PR #4729, opened 2020-10-20
Integrate 3D secure into recurring contributions

sbinlondon:
Relates https://github.com/opencollective/opencollective-frontend/pull/5252

* Removes deprecated field `paymentMethod` on `addCreditCard`
* Adds `confirmCreditCard` mutation
* Breaks `addCreditCard` into two parts if the credit card requires strong customer authentication to prevent unauthenticated credit cards being saved to an account (this was happening in edit collective)
* Also removes some unused code (cancelSubscription)


server/paymentProviders/stripe/creditcard.js
```diff
@@ -175,13 +175,6 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
     throw new Error(UNKNOWN_ERROR_MSG);
   }
 
-  // Success: delete reference to paymentIntent
-  // TODO: paymentIntent should be later removed anyway on success, see subscriptions.js and orders.js
-  if (order.data.paymentIntent) {
-    delete order.data.paymentIntent;
-    await order.update({ data: order.data });
-  }
```
znarf, line 183: Why are we removing that code now?

sbinlondon: There was a note saying it should be taken care of in orders.js and subscriptions.js. 

Subscriptions.js doesn't exist anymore, and in orders.js (v1 mutations) we call `libPayments.executeOrder` in both `createOrder` and `confirmOrder`, which has a line to take care of deleting `paymentIntent`: https://github.com/opencollective/opencollective-api/blob/c831017ae342df590dab217c3df5d07a0ed39bd7/server/lib/payments.js#L329

So it's not necessary anymore.

sbinlondon:
Relates opencollective/opencollective#3429

server/graphql/v2/mutation/PaymentMethodMutations.js
```diff
@@ -1,39 +1,53 @@
...
+    shouldBeSaved: {
+      type: GraphQLBoolean,
+      description: 'If the Payment Method needs to be marked as saved again during confirmCreditCard',
+    },
+  },
+});
 
 const addCreditCard = {
-  type: new GraphQLNonNull(PaymentMethod),
+  type: GraphQLNonNull(CreditCardWithStripeError),
```
Betree, line 32: Just flagging that this is a breaking change.
- Either we deploy frontend & API at the same time (but there's still a risk)
- Or we preserve the legacy behavior & return type in `addStripeCreditCard`, so people with an outdated frontend will not see any crash

server/graphql/v2/mutation/PaymentMethodMutations.js
```diff
@@ -50,30 +64,69 @@ const addCreditCard = {
...
+      // unsave payment method if saved, we will resave it in confirmCreditCard
+      await pm.update({ saved: false });
+
+      pm.stripeError = {
+        message: error.message,
+        account: error.stripeAccount,
+        response: error.stripeResponse,
+      };
+
+      return { paymentMethod: pm, stripeError: pm.stripeError, shouldBeSaved: args.isSavedForLater };
```
Betree, line 89: I don't get the goal of this `shouldBeSaved` since we're just returning the passed argument? The frontend is already supposed to know which value was passed here right?

server/graphql/v2/mutation/PaymentMethodMutations.js
```diff
@@ -50,30 +64,69 @@ const addCreditCard = {
...
+    }
+
+    // Success: delete reference to setupIntent and mark again as saved if needed
+    if (pm.data.setupIntent) {
+      delete pm.data.setupIntent;
     }
 
-    return pm;
+    await pm.update({ confirmedAt: new Date(), data: pm.data });
```
Betree, line 97: The `save` flag doesn't seems to be updated as the comment above is suggesting it should be

server/graphql/v2/mutation/PaymentMethodMutations.js
```diff
@@ -50,30 +64,69 @@ const addCreditCard = {
...
+    paymentMethod: {
+      type: new GraphQLNonNull(PaymentMethodReferenceInput),
+    },
+    shouldBeSaved: {
+      type: GraphQLBoolean,
+      description: 'Whether this payment method should be marked as saved for future payments',
+    },
+  },
+  async resolve(_, args, req) {
+    const paymentMethod = await fetchPaymentMethodWithReference(args.paymentMethod);
```
Betree, line 116: Let's add some permissions check here, otherwise someone could use this to mark all payment methods on the service as saved/unsaved.

Betree (changes requested):

server/graphql/v2/mutation/PaymentMethodMutations.js
sbinlondon: You're right, sorry! Got a little too into the code cleaning 😳 I will add the deprecation back.

server/graphql/v2/mutation/PaymentMethodMutations.js
sbinlondon: The comment was copied from the other mutation, sorry. The code here is correct. We set `saved: false` in the case of an error, then the resetting happens in `confirmCreditCard`.

server/graphql/v2/mutation/PaymentMethodMutations.js
sbinlondon: Changing as discussed on Slack:

* adding back `addStripeCreditCard` as the mutation used to be, just so nothing breaks coming from the frontend when we deploy the API PR
* keeping these changes to `addCreditCard`
* then will write a follow up PR to remove `addStripeCreditCard` again

Both mutations are only used in the recurring contributions page now so this will be fine.

server/graphql/v2/mutation/PaymentMethodMutations.js
```diff
@@ -50,39 +118,82 @@ const addCreditCard = {
...
+
+    if (!paymentMethod || !req.remoteUser.isAdmin(paymentMethod.CollectiveId)) {
```
Betree, line 174: User might be unauthenticated

```suggestion
    if (!paymentMethod || !req.remoteUser?.isAdmin(paymentMethod.CollectiveId)) {
```

znarf (changes requested):
Not sure about `shouldBeSaved` / `isSavedForLater` as names. Have you considered alternatives?

Instead of `shouldBeSaved` as extra param, why not have `savedForLater` in the paymentMethod input and expose that also as `savedForLater` in the paymentMethod object?

server/graphql/v2/mutation/PaymentMethodMutations.js
```diff
@@ -1,39 +1,107 @@
...
-      deprecationReason: '2020-08-24: Please use creditCardInfo',
     },
+    account: {
+      type: AccountReferenceInput,
+      description: 'Account to add Payment Method to',
+    },
+  },
+  async resolve(_, args) {
+    const account = await fetchAccountWithReference(args.account);
+    const collective = await models.Collective.findByPk(account.id);
```
znarf, line 47: `account` is already a `collective`, you're fetching it 2 times.

server/graphql/v2/mutation/PaymentMethodMutations.js
```diff
@@ -1,39 +1,107 @@
...
+    if (!collective) {
+      throw Error('This collective does not exist');
```
znarf, line 49: We should say "account"? Not collective.

sbinlondon: Sorry, re: `addStripeCreditCard`, we're only adding back a deprecated mutation briefly to make sure there are no breaking changes. Then in a follow up API PR we will remove it (once the frontend is deployed using `addCreditCard`) immediately. I can make the changes to the code but we shouldn't dwell on 'perfecting' it.

sbinlondon:
> Not sure about `shouldBeSaved` / `isSavedForLater` as names. Have you considered alternatives?

`isSavedForLater` is from @Betree's changes to `addCreditCard` so can't speak to that one. I think it makes sense though and don't see the point in changing it.

> Instead of `shouldBeSaved` as extra param, why not have `savedForLater` in the paymentMethod input and expose that also as `savedForLater` in the paymentMethod object?

It wouldn't work in this case. If we are calling `confirmCreditCard` it's because we have to authenticate with 3D secure. If the `paymentMethod` has been created with `saved: true` in `addCreditCard`, but then we are asked to authenticate with 3D secure, we temporarily mark the `paymentMethod` as `saved: false` so it doesn't still show up on the account if we fail authentication. We can't add it as an input field because we aren't passing a `PaymentMethodCreateInput` to `confirmCreditCard`, we're only fetching the `paymentMethod` using the `PaymentMethodReferenceInput` once we are in the mutation, and of course as far as the database knows, `saved = false`. So we have to pass `shouldBeSaved` in the case that we need to change that again.

This is how `createOrder`/`confirmOrder` works, except it doesn't have to pass an extra argument because in `confirmOrder` you fetch the Order and the PaymentMethod and update the PaymentMethod based on `order.data.savePaymentMethod`. But here we're only dealing with the fetched PaymentMethod so we have to have this extra field.

server/graphql/v2/mutation/PaymentMethodMutations.js
znarf: Ok

znarf:
@sbinlondon seems complicated, why not simply rely on `order.data.savePaymentMethod`?

sbinlondon:
> Maybe can we save something in `paymentMethod.data` similar to `order.data.savePaymentMethod`?

That's a good idea! I will update that.

znarf (approved):
Looks good to me.

PR #4738, opened 2020-10-21
Submit on Behalf

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3487
Resolves https://github.com/opencollective/opencollective/issues/3541
Resolves https://github.com/opencollective/opencollective/issues/3567
Resolves https://github.com/opencollective/opencollective/issues/3568
Resolves https://github.com/opencollective/opencollective/issues/3617

### DIFF
--- a/server/graphql/v2/input/ExpenseInviteDraftInput.ts
+++ b/server/graphql/v2/input/ExpenseInviteDraftInput.ts
@@ -1,0 +1,77 @@
+import {
+  GraphQLBoolean,
+  GraphQLInputObjectType,
+  GraphQLInt,
+  GraphQLList,
+  GraphQLNonNull,
+  GraphQLString,
+} from 'graphql';
+import GraphQLJSON from 'graphql-type-json';
+
+import { ExpenseType } from '../enum/ExpenseType';
+
+import { LocationInput } from './LocationInput';
+
+const ExpenseInvitee = new GraphQLInputObjectType({
+  name: 'ExpenseInvitee',
+  fields: {
+    id: { type: GraphQLInt },
+    slug: { type: GraphQLString },
+    name: { type: GraphQLString },
+    email: { type: GraphQLString },
+    isInvite: { type: GraphQLBoolean },
+  },
+});
+
+/**
+ * Input type to use as the type for the expense input in createExpense mutation.
+ */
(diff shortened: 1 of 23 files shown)


server/models/User.js
```diff
@@ -189,6 +200,13 @@ export default (Sequelize, DataTypes) => {
           };
         },
       },
+      setterMethods: {
+        name(value) {
+          const { firstName, lastName } = splitName(value);
+          this.setDataValue('firstName', firstName);
```
Betree, line 206: I'm not against this change (can't hurt) but just mentioning that we're aiming to remove these fields in the future

Betree (commented):
This PR is going to crash Github with resolved issues when merging :smile: 

Good job over here! The architecture design is clear and well structured.

Some additional feedback:
- Because they're sending emails, `draftExpenseAndInviteUser` and `resendDraftExpenseInvite` should be rate limited. I recommend looking at `server/lib/rate-limit.ts` if you're looking for a simpler way to add that.
- We should update `getTaxFormsRequiredForExpenses` and `getTaxFormsRequiredForAccounts` to add `DRAFT` and `UNVERIFIED` to the `AND all_expenses.status NOT IN ('ERROR', 'REJECTED')` conditions => don't warn users about tax form until the expense is fully ready

kewitz:
Force pushed the resolution of conflicts, nothing actually really changed from the last review.

server/graphql/v2/input/ExpenseInviteDraftInput.ts
```diff
@@ -0,0 +1,77 @@
...
+import { LocationInput } from './LocationInput';
+
+const ExpenseInvitee = new GraphQLInputObjectType({
+  name: 'ExpenseInvitee',
+  fields: {
+    id: { type: GraphQLInt },
+    slug: { type: GraphQLString },
+    name: { type: GraphQLString },
+    email: { type: GraphQLString },
+    isInvite: { type: GraphQLBoolean },
```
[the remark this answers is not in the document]
kewitz: Not exactly because you can send either just the ID or all other parameters.

server/models/User.js
kewitz: This was breaking several tests due to `createUserWithCollective`. I rolled back.

PR #4742, opened 2020-10-21
Update PDF util

sbinlondon:
* Adds new util `fetchWithTimeout` which takes a URL and options to pass to `node-fetch` and a timeout in ms, and returns a thenable promise
* Uses the timeout when fetching PDFs for thank you emails
* Does not send PDF receipt for gift card transactions

### DIFF
--- a/server/lib/fetch.js
+++ b/server/lib/fetch.js
@@ -1,0 +1,44 @@
+import { get } from 'lodash';
+import fetch from 'node-fetch';
+
+import logger from './logger';
+
+/**
+ * Make a fetch call with a timeout. Returns a thenable Promise.
+ * @param {String} url The url to fetch.
+ * @param {Object} [fetchOptions] The method, headers, and other options for node-fetch. Default method is GET.
+ * @param {Number} [fetchOptions.timeoutInMs] The timeout in ms. Default is 5000.
+ */
+
+export const fetchWithTimeout = (url, fetchOptions) => {
+  const timeoutInMs = get(fetchOptions, 'timeoutInMs', 5000);
+
+  return new Promise((resolve, reject) => {
+    let timedOut = false;
+
+    const timer = setTimeout(() => {
+      timedOut = true;
+      logger.warn(`Fetch request to ${url} timed out`);
+      return reject(new Error(`Fetch request to ${url} timed out`));
+    }, timeoutInMs);
+
+    fetch(url, fetchOptions)
+      .then(
+        response => {
+          if (!timedOut) {
+            logger.info(`Fetch request to ${url} successful`);
+            return resolve(response);
+          }
+        },
+        err => {
+          if (timedOut) {
+            return;
+          }
+          return reject(new Error(`Fetch error: ${err.message}`));
+        },
+      )
+      .finally(() => {
+        clearTimeout(timer);
+      });
+  });
+};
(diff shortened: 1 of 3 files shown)


server/lib/fetch.js
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Make a fetch call with a timeout. Returns a thenable Promise.
+ * @param {Object} fetchParams The object of options to pass into the fetch request.
+ * @param {String} fetchParams.url The url to fetch.
+ * @param {Object} [fetchParams.options] The method, headers, and other options for node-fetch. Default method is GET.
+ * @param {Number} [timeoutInMs] The callback to call with the authorized client. Default is 5000 ms.
+ */
+
+export const fetchWithTimeout = (fetchParams, timeoutInMs = 5000) => {
```
znarf, line 13: I think it would look better by following `fetch` API: first argument `url`, second argument `options`. `timeoutInMs` could be read from `options` or kept as third argument.

server/lib/payments.js
```diff
@@ -428,8 +429,10 @@ const sendOrderConfirmedEmail = async (order, transaction) => {
       subscriptionsLink: interval && `${config.host.website}/${fromCollective.slug}/recurring-contributions`,
     };
 
-    // hit PDF service and get PDF
-    pdf = await getTransactionPdf(transaction, user);
+    // hit PDF service and get PDF (unless payment method type is gift card)
+    if (paymentMethod.type !== PAYMENT_METHOD_TYPES.VIRTUALCARD) {
```
Betree, line 433: `paymentMethod` can be null (ie. if order amount is `0`).

```suggestion
    if (paymentMethod?.type !== PAYMENT_METHOD_TYPES.VIRTUALCARD) {
```

server/lib/pdf.js
```diff
@@ -13,7 +13,8 @@ export const getTransactionPdf = async (transaction, user) => {
   const headers = {
     Authorization: `Bearer ${accessToken}`,
   };
-  return await fetch(pdfUrl, { method: 'get', headers })
+
+  return fetchWithTimeout(pdfUrl, { method: 'get', headers })
```
Betree, line 17: Maybe we could pass a different timeout here? PDF service can sometimes be slow, I think 10 seconds would be a good start.

Betree (commented):
The helper looks good to me, using async/await may have been cleaner - promises chains are a bit difficult to read. But that's not a blocker.

sbinlondon:
> The helper looks good to me, using async/await may have been cleaner - promises chains are a bit difficult to read. But that's not a blocker.

I'll keep this in mind and maybe refactor next week when I have a bit more time. Was good to practice promises at least 😅

server/lib/fetch.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Make a fetch call with a timeout. Returns a thenable Promise.
+ * @param {String} url The url to fetch.
+ * @param {Object} [fetchOptions] The method, headers, and other options for node-fetch. Default method is GET.
+ * @param {Number} [fetchOptions.timeoutInMs] The timeout in ms. Default is 5000.
+ */
+
+export const fetchWithTimeout = (url, fetchOptions) => {
+  const timeoutInMs = get(fetchOptions, 'timeoutInMs', null) || 5000;
```
znarf, line 14:
```suggestion
  const timeoutInMs = get(fetchOptions, 'timeoutInMs', 5000);
```

znarf (approved):
Looks good. I'm not sure if that's needed but we could also add Keep-Alive.

Similar to https://github.com/opencollective/opencollective-frontend/blob/master/lib/apollo-client.js#L49

PR #4753, opened 2020-10-22
Default messages when rejecting

znarf:


sbinlondon (approved):
Great idea - looks good to me 👍

PR #4786, opened 2020-10-29
Add Updates to GraphQL v2

sbinlondon:
Necessary PR for https://github.com/opencollective/opencollective-frontend/pull/5319

* Adds Updates object to v2
* Allows you to query Updates on an Account

### DIFF
--- a/server/graphql/v2/object/Update.js
+++ b/server/graphql/v2/object/Update.js
@@ -1,0 +1,74 @@
+import { GraphQLBoolean, GraphQLInt, GraphQLList, GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
+import { GraphQLDateTime } from 'graphql-iso-date';
+
+import { stripHTML } from '../../../lib/sanitize-html';
+import { UpdateAudienceType } from '../enum/UpdateAudienceType';
+import { getIdEncodeResolver, IDENTIFIER_TYPES } from '../identifiers';
+import { Account } from '../interface/Account';
+
+const Update = new GraphQLObjectType({
+  name: 'Update',
+  description: 'This represents an Update',
+  fields: () => {
+    return {
+      id: {
+        type: new GraphQLNonNull(GraphQLString),
+        resolve: getIdEncodeResolver(IDENTIFIER_TYPES.UPDATE),
+      },
+      legacyId: { type: GraphQLInt },
+      slug: { type: new GraphQLNonNull(GraphQLString) },
+      userCanSeeUpdate: {
+        description: 'Indicates whether or not the user is allowed to see the content of this update',
+        type: new GraphQLNonNull(GraphQLBoolean),
+        resolve(update, _, req) {
+          if (!update.isPrivate) {
+            return true;
+          }
+          return req.remoteUser && req.remoteUser.canSeeUpdates(update.CollectiveId);
+        },
+      },
+      isPrivate: { type: new GraphQLNonNull(GraphQLBoolean) },
+      title: { type: new GraphQLNonNull(GraphQLString) },
+      createdAt: { type: new GraphQLNonNull(GraphQLDateTime) },
+      updatedAt: { type: new GraphQLNonNull(GraphQLDateTime) },
+      publishedAt: { type: GraphQLDateTime },
+      notificationAudience: { type: UpdateAudienceType },
+      makePublicOn: { type: GraphQLDateTime },
+      summary: {
+        type: GraphQLString,
+        resolve(update, _, req) {
+          if (update.isPrivate && !(req.remoteUser && req.remoteUser.canSeeUpdates(update.CollectiveId))) {
+            return null;
+          }
+
+          return update.summary || '';
+        },
+      },
+      html: {
+        type: GraphQLString,
+        resolve(update, _, req) {
+          if (update.isPrivate && !(req.remoteUser && req.remoteUser.canSeeUpdates(update.CollectiveId))) {
+            return null;
+          }
+
+          return stripHTML(update.html || '');
+        },
+      },
+      tags: { type: new GraphQLList(GraphQLString) },
+      fromAccount: {
+        type: Account,
+        resolve(update, args, req) {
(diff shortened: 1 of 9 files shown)


server/graphql/v2/interface/Account.js
```diff
@@ -269,6 +270,14 @@ const accountFieldsDefinition = () => ({
       return collective;
     },
   },
+  updates: {
+    type: UpdateCollection,
```
Betree, line 274:
```suggestion
    type: new GraphQLNonNull(UpdateCollection),
```

server/graphql/v2/interface/Account.js
```diff
@@ -562,6 +571,28 @@ export const AccountFields = {
       return get(collective.data, 'categories', []);
     },
   },
+  updates: {
+    type: UpdateCollection,
+    args: {
+      limit: { type: GraphQLInt },
+      offset: { type: GraphQLInt },
```
Betree, line 578: Can be replaced by:

```suggestion
      ...CollectionArgs,
```

Benefits:
- Add default value on `limit` => we prefer that people don't fetch all expenses at once
- Add `description`s (better docs)

server/graphql/v2/object/Update.js
```diff
@@ -0,0 +1,118 @@
...
+          }
+          return req.remoteUser && req.remoteUser.canSeeUpdates(update.CollectiveId);
+        },
+      },
+      isPrivate: { type: new GraphQLNonNull(GraphQLBoolean) },
+      title: { type: new GraphQLNonNull(GraphQLString) },
+      createdAt: { type: new GraphQLNonNull(DateString) },
+      updatedAt: { type: new GraphQLNonNull(DateString) },
+      publishedAt: { type: DateString },
+      image: { type: GraphQLString },
```
Betree, line 61: This field is never set in the DB, looks like a feature that was never fully implemented. Let's not expose it

```suggestion
```

server/graphql/v2/object/Update.js
```diff
@@ -0,0 +1,118 @@
...
+      views: { type: GraphQLInt },
```
Betree, line 62: This one is not defined on the model

```suggestion
```

server/graphql/v2/object/Update.js
```diff
@@ -0,0 +1,118 @@
...
+      markdown: {
+        type: GraphQLString,
+        resolve(update, _, req) {
+          if (update.isPrivate && !(req.remoteUser && req.remoteUser.canSeeUpdates(update.CollectiveId))) {
+            return null;
+          }
+
+          return stripHTML(update.markdown || '');
+        },
+      },
```
Betree, line 94: This one is deprecated, let's not use it

```suggestion
```

server/graphql/v2/object/Update.js
```diff
@@ -0,0 +1,118 @@
...
+      tags: { type: new GraphQLList(GraphQLString) },
+      fromCollective: {
```
Betree, line 96:
```suggestion
      fromAccount: {
```

server/graphql/v2/object/Update.js
```diff
@@ -0,0 +1,118 @@
...
+        type: Account,
+        resolve(update, args, req) {
+          return req.loaders.Collective.byId.load(update.FromCollectiveId);
+        },
+      },
+      collective: {
```
Betree, line 102:
```suggestion
      account: {
```

server/graphql/v2/object/Update.js
```diff
@@ -0,0 +1,118 @@
...
+        resolve(update, args, req) {
+          return req.loaders.Collective.byId.load(update.CollectiveId);
+        },
+      },
+      tier: {
+        type: Tier,
+        resolve(update, args, req) {
+          return req.loaders.Tier.byId.load(update.TierId);
+        },
+      },
```
Betree, line 113: This field is never set in the DB, looks like a feature that was never fully implemented. Let's not expose it

```suggestion
```

server/graphql/v2/object/Update.js
```diff
@@ -0,0 +1,118 @@
...
+  GraphQLInt,
+  GraphQLList,
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLString,
+} from 'graphql';
+import { GraphQLDateTime } from 'graphql-iso-date';
+
+import { stripHTML } from '../../../lib/sanitize-html';
+import { DateString } from '../../v1/types';
```
Betree, line 13: We should rather use the `GraphQLDateTime` standard, `DateString` is a type from v1

server/graphql/v2/object/Update.js
```diff
@@ -0,0 +1,118 @@
...
+import { getIdEncodeResolver, IDENTIFIER_TYPES } from '../identifiers';
+import { Account } from '../interface/Account';
+
+import { Tier } from './Tier';
+
+const UpdateAudienceTypeEnum = new GraphQLEnumType({
```
Betree, line 19: We should move that to the `enum` folder, it's meant to be used by the `createMutation` too when we'll have it on v2

Betree (commented):
This will also resolve #4786 :rocket:

server/graphql/v2/interface/Account.js
sbinlondon: 1. Do you have an example of this in use?
2. Why do I keep seeing the word 'expenses' used with Updates? How are they related? I don't see the connection.

server/graphql/v2/object/Update.js
sbinlondon: Ok, it formats it a bit different from the v1 but that's not a problem. I'll just need to double check that it still plays nice with the frontend.

server/graphql/v2/interface/Account.js
Betree: https://github.com/opencollective/opencollective-api/blob/5cd771d0436ab06fe5ef34eaca7bd7ef2ab517ec/server/graphql/v2/query/TransactionsQuery.ts#L14

Betree (approved):
Looks good!

PR #4802, opened 2020-11-02
Consolidated monthly transaction PDFs for users and orgs

sbinlondon:
This PR:

* moves consolidated PDF functionality out of GraphQL v1 queries and into `lib/pdf`
* adds call for consolidated PDFs of contributions per host for a user collective to the `cron/monthly/user-report`
* attaches PDFs to user monthly report if any, and adds a line in the email about it
* adds to org monthly report

### DIFF
--- a/server/lib/pdf.js
+++ b/server/lib/pdf.js
@@ -1,6 +1,10 @@
 import config from 'config';
+import { get } from 'lodash';
+import moment from 'moment';
 
-import { TOKEN_EXPIRATION_LOGIN } from './auth';
+import models, { Op } from '../models';
+
+import { TOKEN_EXPIRATION_PDF } from './auth';
 import { fetchWithTimeout } from './fetch';
 import logger from './logger';
 
@@ -9,7 +13,7 @@ export const getTransactionPdf = async (transaction, user) => {
     return;
   }
   const pdfUrl = `${config.host.pdf}/transactions/${transaction.uuid}/invoice.pdf`;
-  const accessToken = user.jwt({}, TOKEN_EXPIRATION_LOGIN);
+  const accessToken = user.jwt({}, TOKEN_EXPIRATION_PDF);
   const headers = {
     Authorization: `Bearer ${accessToken}`,
   };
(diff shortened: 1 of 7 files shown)


sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3633

server/lib/pdf.js
```diff
@@ -28,3 +33,118 @@ export const getTransactionPdf = async (transaction, user) => {
...
+    const fromCollectiveSlug = invoiceInfo[2];
+
+    // Get user so we can generate access token
+    const fromCollectiveUser = await models.User.findOne({
+      where: { CollectiveId: invoice.FromCollectiveId },
+    });
+
+    // Call PDF service for the invoice
+    const pdfUrl = `${config.host.pdf}/collectives/${fromCollectiveSlug}/${toOrgCollectiveSlug}/${startOfMonth}/${endOfMonth}.pdf`;
+    const accessToken = fromCollectiveUser.jwt({}, TOKEN_EXPIRATION_LOGIN);
```
znarf, line 120: We should review this `TOKEN_EXPIRATION_LOGIN`, doesn't look good in this context.

server/lib/pdf.js
```diff
@@ -28,3 +33,118 @@ export const getTransactionPdf = async (transaction, user) => {
...
+
+    let invoicePdf;
+    await fetchWithTimeout(pdfUrl, { method: 'get', headers, timeoutInMs: 10000 })
+      .then(response => {
+        const { status } = response;
+        if (status >= 200 && status < 300) {
+          invoicePdf = response.body;
+        } else {
+          logger.warn('Failed to fetch PDF');
+          invoicePdf = null;
```
znarf, line 133: `invoicePdf = null;` Seems useless

server/lib/pdf.js
```diff
@@ -28,3 +33,118 @@ export const getTransactionPdf = async (transaction, user) => {
...
+        }
+      })
+      .catch(error => {
+        logger.error(`Error fetching PDF: ${error.message}`);
+      });
```
znarf, line 138: The whole block should be restructured with async/await.

server/lib/pdf.js
```diff
@@ -28,3 +33,118 @@ export const getTransactionPdf = async (transaction, user) => {
...
+      [Op.or]: [
+        { FromCollectiveId: fromCollective.id, UsingVirtualCardFromCollectiveId: null },
+        { UsingVirtualCardFromCollectiveId: fromCollective.id },
+      ],
+    },
+  });
+
+  const hostsById = {};
+  const invoicesByKey = {};
+  await Promise.map(transactions, async transaction => {
```
znarf, line 51: We should take an opportunity to restructure this with a cleaner/simpler approach.

server/lib/pdf.js
```diff
@@ -28,3 +33,118 @@ export const getTransactionPdf = async (transaction, user) => {
...
+  return invoices;
+};
+
+export const getConsolidatedInvoicePdfs = async fromCollective => {
+  if (['ci', 'test'].includes(config.env)) {
+    return;
+  }
+
+  // Get invoices
+  const invoices = await createConsolidatedInvoices(fromCollective);
```
znarf, line 96: `createConsolidatedInvoices` doesn't seem an appropriate name for the function, we're not creating anything. The name should answer the question "what are we actually getting?"?

server/lib/pdf.js
```diff
@@ -28,3 +33,118 @@ export const getTransactionPdf = async (transaction, user) => {
...
+  const invoices = [];
+  Object.keys(invoicesByKey).forEach(key => invoices.push(invoicesByKey[key]));
+  invoices.sort((a, b) => {
+    return a.slug > b.slug ? -1 : 1;
+  });
+  return invoices;
+};
+
+export const getConsolidatedInvoicePdfs = async fromCollective => {
+  if (['ci', 'test'].includes(config.env)) {
```
znarf, line 91: @Betree had some comments about this in the similar PR?

server/lib/pdf.js
```diff
@@ -28,3 +33,118 @@ export const getTransactionPdf = async (transaction, user) => {
...
+      month,
+      totalAmount,
+      totalTransactions,
+      currency: transaction.hostCurrency,
+    };
+  });
+  const invoices = [];
+  Object.keys(invoicesByKey).forEach(key => invoices.push(invoicesByKey[key]));
+  invoices.sort((a, b) => {
+    return a.slug > b.slug ? -1 : 1;
```
znarf, line 85: If it's to sort alphabetically, we should use `string.localeCompare`

server/lib/pdf.js
```diff
@@ -28,3 +33,118 @@ export const getTransactionPdf = async (transaction, user) => {
...
+      slug,
+      year,
+      month,
+      totalAmount,
+      totalTransactions,
+      currency: transaction.hostCurrency,
+    };
+  });
+  const invoices = [];
+  Object.keys(invoicesByKey).forEach(key => invoices.push(invoicesByKey[key]));
```
znarf, line 83: Looks weird. Isn't that `Object.values(invoicesByKey)` ?

znarf (changes requested):
Good progress.

Sorry if some comments are about old code that was only moved, but I think it's a good opportunity to refactor it to higher standards.

server/lib/pdf.js
sbinlondon: Yes he did - I had not had a chance to update it here, I'll make sure I do.

server/lib/pdf.js
sbinlondon: I'm not sure if it is - the code is taken from `server/graphql/v1/queries.js`. I'll update it here and see if it fits.

server/lib/pdf.js
sbinlondon: I'll have a closer look at the old code and see what all I can do to update it!

sbinlondon: No, I'm pretty sure it's `key` not `value`. If it was `value` I don't think the old code would have worked.

sbinlondon:
Not sure about the lines in the emails... could use some feedback on the best place to include it, especially in the org report

server/lib/pdf.js
```diff
@@ -1,15 +1,21 @@
...
+import moment from 'moment';
+
+import models, { Op } from '../models';
 
-import { TOKEN_EXPIRATION_LOGIN } from './auth';
 import { fetchWithTimeout } from './fetch';
 import logger from './logger';
 
+const minutesToSeconds = minutes => moment.duration({ minutes }).asSeconds();
+const TOKEN_EXPIRATION_PDF = minutesToSeconds(75);
```
znarf, line 11: Maybe move in the auth.js file next to the similar constants.

server/lib/pdf.js
znarf: I insist, I think it's the same as `invoices =  Object.values(invoicesByKey)` which would be much cleaner.

server/lib/pdf.js
```diff
@@ -28,3 +34,122 @@ export const getTransactionPdf = async (transaction, user) => {
...
+      year,
+      month,
+      totalAmount,
+      totalTransactions,
+      currency: transaction.hostCurrency,
+    };
+  }
+
+  Object.keys(invoicesByKey).forEach(key => invoices.push(invoicesByKey[key]));
+  console.log('invoices', invoices);
```
znarf, line 88: console.log

znarf (changes requested):
Looks much cleaner. A few details to review.

PR #4803, opened 2020-11-02
Add transaction PDFs to recurring contributions thank you emails

sbinlondon:
Fixes https://github.com/opencollective/opencollective/issues/3651

### DIFF
--- a/config/ci.json
+++ b/config/ci.json
@@ -42,5 +42,8 @@
   },
   "fixer": {
     "accessKey": "UNDEFINED_FIXER_ACCESS_KEY"
+  },
+  "pdfService": {
+    "fetchTransactionsReceipts": false
   }
 }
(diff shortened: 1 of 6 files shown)


Betree (commented):
Looking good! Did you try locally?

Betree (commented):
Also, rather than https://github.com/opencollective/opencollective-api/blob/50850ab5f724a99ebc89bd2663ee99fd8ad4a5c8/server/lib/pdf.js#L8, can we have a flag (ie `FETCH_TRANSACTIONS_RECEIPTS`) (set to false by default on CI & test)? That will:
- Give us more flexibility if we want to add some tests for this
- Allow us to temporarily disable the feature if we notice that it's slowing/blocking the recurring contributions job

sbinlondon:
> Looking good! Did you try locally?

Yes, all working fine!

PR #4813, opened 2020-11-03
Bayes classifier to flag collectives

znarf:


Betree (approved):
Looking good!

As an enhancement, maybe we should maybe add a way to ignore already classified collectives - either through a configurable `createdAt` param or via a flag set on collective's data. The second sounds better cause we'll be able to go back in time if we detect a new type of spam that was ignored before.

PR #4841, opened 2020-11-09
Only validate balance if dealing with Borderless Account

kewitz:
Some accounts are not Borderless Accounts and since we only use it to validate if the account has enough balance, I'll skip and let the transaction fail if we don't have the balance in their account.


kewitz:
Not needed. After a second investigation, I isolated the problem to the used token.

PR #3992, opened 2020-06-04
Collectives 'features' endpoint

Betree:
:point_right: Please read https://github.com/opencollective/opencollective/issues/2807 for the motivations and specs of this feature.

A proposal for a new `features` endpoint for Collectives, currently on API V1 only but meant to be copied for API V2. Some features are not entirely covered yet (ie. we don't check the data for conversations) because I'd like to get some feedback from the team before going further.

PR #4855, opened 2020-11-12
Rebased Features endpoint

sbinlondon:
Building on top of @Betree's [work from June](https://github.com/opencollective/opencollective-api/pull/3992) and updating it for how the API has changed since then and new ideas we have for this feature.

* Expands on `FEATURES` to include collective page features
* Implements `FEATURE_STATUS` to tell us which collective types support which features; and if they are supported, whether they're enabled (and further if they are, whether they have data or not) or disabled
* Adds a `features` field on `COLLECTIVE` (v1) and `ACCOUNT` (v2)  to query what features are available for a Collective
* Adds tests for the new features endpoint


server/graphql/v1/CollectiveInterface.js
```diff
@@ -870,6 +877,19 @@ export const CollectiveInterfaceType = new GraphQLInterfaceType({
   },
 });
 
+const FeaturesFields = () => {
+  return FeaturesList.reduce(
+    (obj, feature) =>
+      Object.assign(obj, {
+        [feature]: {
+          type: CollectiveFeatureStatus,
+          resolve: getFeatureStatusResolver(`${feature}`),
```
Betree, line 886: Why not directly:

```suggestion
          resolve: getFeatureStatusResolver(feature),
```

?

server/lib/allowed-features.ts
```diff
@@ -3,8 +3,30 @@ import { get } from 'lodash';
...
+  [FEATURE.EXPENSES]: [types.ORGANIZATION, types.COLLECTIVE, types.EVENT, types.FUND, types.PROJECT],
+  [FEATURE.COLLECTIVE_GOALS]: [types.COLLECTIVE, types.ORGANIZATION],
+  [FEATURE.TOP_FINANCIAL_CONTRIBUTORS]: [types.COLLECTIVE, types.ORGANIZATION, types.FUND],
+  [FEATURE.ALL_FINANCIAL_CONTRIBUTORS]:[types.ORGANIZATION, types.COLLECTIVE, types.EVENT, types.FUND, types.PROJECT],
+  // 4. Connect
+  [FEATURE.CONNECT]: [types.ORGANIZATION, types.COLLECTIVE],
   [FEATURE.CONVERSATIONS]: [types.COLLECTIVE, types.ORGANIZATION],
+  [FEATURE.UPDATES]: [types.COLLECTIVE, types.ORGANIZATION],
+  // 5. About
+  [FEATURE.ABOUT]: [types.USER, types.ORGANIZATION, types.COLLECTIVE, types.EVENT, types.FUND, types.PROJECT],
```
Betree, line 27: If it's allowed for all types, we don't need to set any flags here. `isFeatureAllowedForCollectiveType` will return `true` by default if there's not flag set.

```suggestion
```

server/constants/feature.ts
```diff
@@ -1,11 +1,34 @@
...
+  PROJECTS = 'PROJECTS',
+  // 3. Budget
+  BUDGET = 'BUDGET',
+  TRANSACTIONS = 'TRANSACTIONS',
+  USE_EXPENSES = 'USE_EXPENSES',
+  COLLECTIVE_GOALS = 'COLLECTIVE_GOALS',
+  TOP_FINANCIAL_CONTRIBUTORS = 'TOP_FINANCIAL_CONTRIBUTORS',
+  ALL_FINANCIAL_CONTRIBUTORS = 'ALL_FINANCIAL_CONTRIBUTORS',
+  // 4. Connect
+  CONNECT = 'CONNECT',
```
Betree, line 18: Sounds to me that `CONNECT` is not a feature, it's a section from the collective page.

server/lib/allowed-features.ts
```diff
@@ -38,22 +92,33 @@ export const hasOptedInForFeature = (collective, feature): boolean => {
...
+  }
+
+  // Check if collective feature is disabled
+  if (feature in OPTIONAL_COLLECTIVE_FEATURES) {
+    const collectivePageSections = get(collective, 'settings.collectivePage.sections');
+    if (collectivePageSections) {
+      const featureSettings = collectivePageSections.find(s => s.section === OPTIONAL_COLLECTIVE_FEATURES[feature]);
+      return featureSettings.isEnabled;
+    }
+  }
```
Betree, line 119: This is a good approach toward solving https://github.com/opencollective/opencollective/issues/3396, though as mentioned in my comment there I wonder how this will play with the `CONNECT` section. Let's discuss this point later.

server/lib/allowed-features.ts
```diff
@@ -17,14 +40,45 @@ export const OPT_OUT_FEATURE_FLAGS = {
...
-export const isFeatureAllowedForCollectiveType = (collectiveType: keyof types, feature: FEATURE): boolean => {
+export const isFeatureAllowedForCollectiveType = (
+  collectiveType: types,
+  isHost: boolean,
+  feature: FEATURE,
+): boolean => {
   const allowedTypes = FeatureAllowedForTypes[feature];
-  return allowedTypes ? allowedTypes.includes(collectiveType) : true;
+  const allowedForType = allowedTypes ? allowedTypes.includes(collectiveType) : true;
+  const orgFeaturesOnlyAllowedForHostOrgs = [
```
Betree, line 61: Maybe we can define this one as a constant outside of the function, as `FEATURES_ONLY_FOR_HOST_ORGS` (similar to what we do with `OPT_IN_FEATURE_FLAGS`)

server/lib/allowed-features.ts
```diff
@@ -38,22 +92,33 @@ export const hasOptedInForFeature = (collective, feature): boolean => {
...
+ * If a given feature is allowed for the collective type, check if it is activated for collective.
  */
-const hasFeature = (collective, feature: FEATURE): boolean => {
+export const hasFeature = (collective, feature: FEATURE): boolean => {
   if (!collective) {
     return false;
   }
 
-  // Check type
-  if (!isFeatureAllowedForCollectiveType(collective.type, feature)) {
```
Betree, line 49: No longer checking the type? This is breaking for https://github.com/opencollective/opencollective-api/blob/d64f4f69eb8ede92ad3cf16ba83f1b811f574177/server/graphql/common/conversations.ts#L36, as people will be able to submit conversations to user profiles for example. I think we should keep this check.

server/constants/feature.ts
```diff
@@ -1,11 +1,28 @@
...
+  // Collective page features
+  RECEIVE_FINANCIAL_CONTRIBUTIONS = 'RECEIVE_FINANCIAL_CONTRIBUTIONS',
+  RECURRING_CONTRIBUTIONS = 'RECURRING_CONTRIBUTIONS',
+  EVENTS = 'EVENTS',
+  PROJECTS = 'PROJECTS',
+  USE_EXPENSES = 'USE_EXPENSES',
+  RECEIVE_EXPENSES = 'RECEIVE_EXPENSES',
+  COLLECTIVE_GOALS = 'COLLECTIVE_GOALS',
+  TOP_FINANCIAL_CONTRIBUTORS = 'TOP_FINANCIAL_CONTRIBUTORS',
+  // ALL_FINANCIAL_CONTRIBUTORS = 'ALL_FINANCIAL_CONTRIBUTORS',
```
Betree, line 11:
```suggestion
```

test/server/lib/allowed-features.test.js
```diff
@@ -1,39 +1,180 @@
...
+
+    // it('ALL_FINANCIAL_CONTRIBUTORS', () => {
+    //   expect(isFeatureAllowedForCollectiveType('COLLECTIVE', false, FEATURE.ALL_FINANCIAL_CONTRIBUTORS)).to.be.true;
+    //   expect(isFeatureAllowedForCollectiveType('ORGANIZATION', false, FEATURE.ALL_FINANCIAL_CONTRIBUTORS)).to.be.true;
+    //   expect(isFeatureAllowedForCollectiveType('ORGANIZATION', true, FEATURE.ALL_FINANCIAL_CONTRIBUTORS)).to.be.true;
+    //   expect(isFeatureAllowedForCollectiveType('USER', false, FEATURE.ALL_FINANCIAL_CONTRIBUTORS)).to.be.false;
+    //   expect(isFeatureAllowedForCollectiveType('EVENT', false, FEATURE.ALL_FINANCIAL_CONTRIBUTORS)).to.be.true;
+    //   expect(isFeatureAllowedForCollectiveType('FUND', false, FEATURE.ALL_FINANCIAL_CONTRIBUTORS)).to.be.true;
+    //   expect(isFeatureAllowedForCollectiveType('PROJECT', false, FEATURE.ALL_FINANCIAL_CONTRIBUTORS)).to.be.true;
+    // });
```
Betree, line 94:
```suggestion
```

Betree (commented):
A few suggestions, but the overall looks good!

In the future we'll probably want to optimize the way we query active features in `server/graphql/common/features.ts`, but that's something for later.

server/lib/allowed-features.ts
sbinlondon: As discussed, have put it back to `settings.features.FEATURE` for now

PR #4866, opened 2020-11-16
Update contributionPolicy to have 3000 char limit

sbinlondon:
Prereq to https://github.com/opencollective/opencollective-frontend/pull/5398

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -271,12 +271,17 @@ export default function (Sequelize, DataTypes) {
         },
       },
 
-      expensePolicy: DataTypes.TEXT, // markdown
+      expensePolicy: {
+        type: DataTypes.TEXT, // markdown
+        validate: {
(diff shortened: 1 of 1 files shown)


Betree (approved):
Lgtm!

server/models/Collective.js
```diff
@@ -271,12 +271,17 @@ export default function (Sequelize, DataTypes) {
         },
       },
 
-      expensePolicy: DataTypes.TEXT, // markdown
+      expensePolicy: {
+        type: DataTypes.TEXT, // markdown
+        validate: {
+          len: [0, 16000], // largest policy is 15,500, all the rest are under 10,000
```
znarf, line 277: Maybe we need more space for safety? Like 32000.

sbinlondon: Would we want it getting much longer than 16,000 though?

PR #4867, opened 2020-11-16
Remove deprecated stripTags and replace it with lib/sanitize-html

sbinlondon:
We weren't using deprecated `stripTags` in many places so replaced it with the favoured `strip-html` (for stripping all tags) or `sanitize-html` where certain tags are allowed

### DIFF
--- a/server/graphql/common/comment.js
+++ b/server/graphql/common/comment.js
@@ -1,7 +1,7 @@
 import { pick } from 'lodash';
 
 import { mustBeLoggedInTo } from '../../lib/auth';
-import { stripTags } from '../../lib/utils';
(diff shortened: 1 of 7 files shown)


Betree (changes requested):
These two functions don't have the same usage:
- `stripTags`: sanitize html, but **allow some** tags (ie. `a`, `img`, etc.)
- `sanitizeHTML`: must be provided a sanitize config. By default `optsStripAll` is used, which **strips all** HTML.

So we can't just replace one by another, we need to make sure the config we pass to `sanitizeHTML` makes sense.

sbinlondon:
Resolved in https://github.com/opencollective/opencollective-api/pull/5173

PR #4883, opened 2020-11-18
Search: deprecate Algolia

Betree:
Related to https://github.com/opencollective/opencollective/issues/3688, https://github.com/opencollective/opencollective/issues/3247, https://github.com/opencollective/opencollective/issues/3233, https://github.com/opencollective/opencollective/issues/2605

This PR:
- Deprecates the `useAlgolia` param on `seach`
- Makes it default to `false`, with a feature flag to re-enable it quickly if needed

@znarf do you think we should add more metrics to keep track of the performances?

### DIFF
--- a/config/custom-environment-variables.json
+++ b/config/custom-environment-variables.json
@@ -129,7 +129,8 @@
   "algolia": {
     "appId": "ALGOLIA_APP_ID",
     "appKey": "ALGOLIA_KEY",
-    "index": "ALGOLIA_INDEX"
+    "index": "ALGOLIA_INDEX",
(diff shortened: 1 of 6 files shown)


znarf:
@Betree With production data, what is the difference in speed for a simple search?

Betree:
With:

```sql
{
  search(term: "test", useAlgolia: $useAlgolia) {
    total
    collectives {
      id
      slug
    }
  }
}
```

I go from ~500ms to ~3000ms when using SQL. So the cost is quite huge, we can/should probably optimize the query before deploying that.

znarf:
We could live with it. But if it's possible to go under 1sec, would be great.

PR #4896, opened 2020-11-20
Add default true pdfService variable to staging/prod config

sbinlondon:


znarf (changes requested):
As discussed, please convert to boolean when checking pdfService.fetchTransactionsReceipts === false.

server/lib/pdf.js
```diff
@@ -9,7 +9,7 @@ import { fetchWithTimeout } from './fetch';
 import logger from './logger';
 
 export const getTransactionPdf = async (transaction, user) => {
-  if (config.pdfService.fetchTransactionsReceipts === false) {
+  if (Boolean(config.pdfService.fetchTransactionsReceipts) === false) {
```
znarf, line 12: That doesn't work, `Boolean('false') is `true`. You have to use our `parseToBoolean` like in similar places.

znarf (changes requested):

znarf (approved):

PR #4919, opened 2020-11-25
migrate collective's expense & contribution policy to HTML

snitin315:
Refers https://github.com/opencollective/opencollective/issues/3713

Related Frontend PR - https://github.com/opencollective/opencollective-frontend/pull/5466

- [x] Have a migration that converts the content to HTML with `showdown` & sanitizes it
- [x] Use `server/lib/sanitize-html.ts` with `basicTextFormatting` + `links` to sanitize, sanitize in model's field with a `set`

### DIFF
--- a/migrations/20201125113750-migrate-collective-policies-to-html.js
+++ b/migrations/20201125113750-migrate-collective-policies-to-html.js
@@ -1,0 +1,74 @@
+'use strict';
+
+import showdown from 'showdown';
+import { buildSanitizerOptions, sanitizeHTML } from '../server/lib/sanitize-html';
+
+const converter = new showdown.Converter();
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    const collectivesWithExpensePolicy = await queryInterface.sequelize.query(
+      `
+        SELECT "id", "expensePolicy" FROM "Collectives"
+        WHERE LENGTH("expensePolicy") > 0 AND "expensePolicy" NOT LIKE '<%'
+        AND "deletedAt" IS NULL;
+      `,
+      { type: Sequelize.QueryTypes.SELECT },
+    );
+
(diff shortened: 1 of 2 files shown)


migrations/20201125113750-migrate-collective-policies-to-html.js
```diff
@@ -0,0 +1,73 @@
...
+      `
+        SELECT "id", "contributionPolicy" FROM "Collectives"
+        WHERE LENGTH("contributionPolicy") > 0 AND "contributionPolicy" NOT LIKE '<%'
+        AND "deletedAt" IS NULL;
+      `,
+      { type: Sequelize.QueryTypes.SELECT },
+    );
+
+    const sanitizeOptions = buildSanitizerOptions({
+      basicTextFormatting: true,
```
Betree, line 29: I forgot to mention it on the issue, but we also need `multilineTextFormatting` since the editor allows lists

Betree (changes requested):
The second item:
> Use server/lib/sanitize-html.ts with basicTextFormatting + links to sanitize, sanitize in model's field with a set

Hasn't been addressed, we need to add that in the model (`server/models/Collective.js`).

Also, we need to update the `len` validation since the string now contains HTML tags. Two choices here:
1. Strip all HTML, check only the text
2. Use a bigger value for `len` to just prevent people from putting a 5Mo text in there. I suggest `len: [0, 50000]`

I personally prefer (2), it's simpler & safer.

migrations/20201125113750-migrate-collective-policies-to-html.js
snitin315: updated 👍🏻

Betree (approved):
Perfect! Nor merging now because we want to deploy the frontend at the same time, but it's ready to be :+1:

PR #4924, opened 2020-11-26
Spam: Don't prevent collective creation

Betree:
Drawback of the current fix: there will be more spammy profiles created.

Alternatives:
1. Have a special flag to "whitelist" user, collective creation will not be prevented if flag is set
2. Only limit user and block account creation if using a domain from block list (`spam.scrore === 1 && spam.usesBlockedDomain`)
3. Don't limit the user, but still block collective creation


znarf (approved):
Looks good!

PR #4946, opened 2020-11-30
Add PDF of transaction to ticket confirmation email

sbinlondon:
Fixes https://github.com/opencollective/opencollective/issues/3726

### DIFF
--- a/server/lib/notifications.js
+++ b/server/lib/notifications.js
@@ -10,8 +10,10 @@ import activitiesLib from '../lib/activities';
 import emailLib from '../lib/email';
 import models from '../models';
 
+import { getTransactionPdf } from './pdf';
 import slackLib from './slack';
 import twitter from './twitter';
+import { toIsoDateStr } from './utils';
 import { enrichActivity, sanitizeActivity } from './webhooks';
 
 const debug = debugLib('notifications');
@@ -109,6 +111,7 @@ async function notifySubscribers(users, activity, options = {}) {
 }
 
 async function notifyUserId(UserId, activity, options = {}) {
+  let pdf;
   const user = await models.User.findByPk(UserId);
   debug('notifyUserId', UserId, user && user.email, activity.type);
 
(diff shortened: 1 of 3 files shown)


znarf (changes requested):
In other similar places, we have the ability to disable the feature, we should have it here too.

Also, seems there is a legit problem with a GraphQL test.

sbinlondon:
> In other similar places, we have the ability to disable the feature, we should have it here too.

We call `getTransactionPdf` here https://github.com/opencollective/opencollective-api/blob/f54f71aeb2de5b81585b3c913c09bd6bf6121103/server/lib/notifications.js#L125 and within `getTransactionPdf` we have the config check for `config.pdfService.fetchTransactionsReceipts`.

server/lib/notifications.js
```diff
@@ -117,6 +120,19 @@ async function notifyUserId(UserId, activity, options = {}) {
     const parentCollective = await event.getParentCollective();
     const ics = await event.getICS();
     options.attachments = [{ filename: `${event.slug}.ics`, content: ics }];
+    const transaction = await models.Transaction.findOne({
+      where: { OrderId: activity.data.order.id, type: 'DEBIT' },
```
Betree, line 124: Not sure if it has any impact, but in other places we take the credit transaction

```suggestion
      where: { OrderId: activity.data.order.id, type: 'CREDIT' },
```

Betree (commented):
:ice_cream: Not a problem because tickets cannot have recurring contributions, but fetching the transaction by order ID can bring bad surprises with recurring contributions. If there's a transaction ID in `activity.data` it could be nice to use it.

server/lib/notifications.js
sbinlondon: I wondered that. The DEBIT amount looks weird but in my mind it made sense that the DEBIT was coming out of the user's account, vs the CREDIT being applied to the collective... will fix!

sbinlondon:
> If there's a transaction ID in `activity.data` it could be nice to use it.

Not that I saw but I can look again

Betree (approved):
Haven't tested, but the code looks good to me

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -164,11 +164,11 @@ const accountMutations = {
   },
   removeTwoFactorAuthTokenFromIndividual: {
     type: new GraphQLNonNull(Individual),
-    description: 'Add 2FA to the Account if it does not have it',
+    description: 'Remove 2FA from the Account if it has been enabled',
```
znarf, line 167: Not related to the PR.

znarf (approved):
Good for me!

PR #4949, opened 2020-12-01
Fix data-export script not exiting

znarf:
Today in production. `No token set for Google Drive, skipping data export upload`

Could you check @Betree ?


Betree (approved):
Yes, it needs a first manual run to set the token then it will run automatically for the next ones. But this `process.exit` is definitely welcome!

Betree:
It's done, the script should run on its own in the future

PR #4958, opened 2020-12-03
use isAdminOfCollective instead of isAdmin

znarf:
works better for EVENTs and PROJECTs


Betree (approved):
Nice change. I'll take note to always use `isAdminOfCollective` as the default in the future.

PR #4959, opened 2020-12-03
Features: Fix checks & add new ones

Betree:


server/lib/allowed-features.ts
```diff
@@ -5,7 +5,6 @@ import FEATURE from '../constants/feature';
 
 // Please refer to and update https://docs.google.com/spreadsheets/d/15ppKaZJCXBjvY7-AjjCj3w5D-4ebLQdEowynJksgDXE/edit#gid=0
 const FeatureAllowedForTypes = {
-  // Collective page features
```
Betree, line 8: @sbinlondon removed this comment because these features are not only for the collective page (updates for example has standalone pages). Let me know if you still think we want to make this distinction

server/lib/allowed-features.ts
```diff
@@ -14,17 +15,18 @@ const FeatureAllowedForTypes = {
     types.PROJECT,
   ],
   [FEATURE.RECURRING_CONTRIBUTIONS]: [types.USER, types.ORGANIZATION, types.COLLECTIVE, types.FUND],
+  [FEATURE.RECEIVE_HOST_APPLICATIONS]: HOST_TYPES,
+  [FEATURE.HOST_DASHBOARD]: HOST_TYPES,
   [FEATURE.EVENTS]: [types.ORGANIZATION, types.COLLECTIVE],
   [FEATURE.PROJECTS]: [types.FUND],
   [FEATURE.USE_EXPENSES]: [types.ORGANIZATION, types.COLLECTIVE, types.EVENT, types.FUND, types.PROJECT],
-  [FEATURE.RECEIVE_EXPENSES]: [types.ORGANIZATION],
+  [FEATURE.RECEIVE_EXPENSES]: [types.ORGANIZATION, types.COLLECTIVE, types.EVENT, types.FUND, types.PROJECT],
```
Betree, line 23: @sbinlondon was this limitation to organizations made on purpose?

sbinlondon: Yes, I remember there being a conversation about how USE_EXPENSES vs RECEIVE_EXPENSES and only host orgs can RECEIVE_EXPENSES in the sense that they are the only ones who can pay them out?

sbinlondon (approved):
Looks good! Apologies for misunderstanding some of the features/who they are for previously.

server/lib/allowed-features.ts
Betree: I see, it's true that they are the only ones who can pay them out but all accounts with a budget (collective, event, etc.) can receive them. In fact, most of the time you will submit your expense to a collective (ie. Babel) rather than a fiscal host (ie. OSC).

PR #4975, opened 2020-12-04
Refactor Host Report to be async and to process transactions in JS to keep transaction consistency

kewitz:
Review with care, I suggest cloning production DB and running:
```
PG_DATABASE=opencollective_prod_snapshot HOST_ID=98478 DEBUG=hostreport TZ=utc npm run script cron/monthly/host-report.js
```

### DIFF
--- a/migrations/20201210150000-fix-host-payment-methods-and-transactions.js
+++ b/migrations/20201210150000-fix-host-payment-methods-and-transactions.js
@@ -1,0 +1,22 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, DataTypes) => {
+    const [, updateTs] = await queryInterface.sequelize.query(`
+      UPDATE "Transactions"
+      SET "amountInHostCurrency" = "amount", "hostFeeInHostCurrency" = ROUND("hostFeeInHostCurrency"/"hostCurrencyFxRate"), "hostCurrencyFxRate" = 1
+      FROM "PaymentMethods" pm, "Collectives" c
+      WHERE
+        "Transactions"."currency" = "Transactions"."hostCurrency"
+        AND "Transactions"."hostCurrencyFxRate" != 1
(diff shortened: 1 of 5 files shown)


migrations/20201210150000-fix-host-payment-methods-and-transactions.js
```diff
@@ -0,0 +1,32 @@
...
+
+    const [, updatePms] = await queryInterface.sequelize.query(`
+      UPDATE "PaymentMethods"
+      SET "currency" = c."currency"
+      FROM "Collectives" c
+      WHERE
+        c.id = "CollectiveId" 
+        AND "PaymentMethods"."type" = 'host'
+        AND "PaymentMethods"."currency" != c."currency"
+    `);
```
znarf, line 27: This is normal in the history of an account to have different currencies, even host account. They will always start with one in USD. Maybe we should add a JOIN on Transactions to make sure we're only updating the bogus one.

migrations/20201210150000-fix-host-payment-methods-and-transactions.js
```diff
@@ -0,0 +1,32 @@
...
+      UPDATE "Transactions"
+      SET "amountInHostCurrency" = "amount", "hostCurrencyFxRate" = 1
+      FROM "PaymentMethods" pm, "Collectives" c
+      WHERE
+        "Transactions"."currency" = "Transactions"."hostCurrency"
+        AND "Transactions"."hostCurrencyFxRate" != 1
+        AND pm.id = "Transactions"."PaymentMethodId"
+        AND c.id = pm."CollectiveId"
+        AND pm."type" = 'host'
+        AND pm."currency" != c."currency"
```
znarf, line 15: I think the condition should be on "Transactions"."currency" to make sure we don't have side effects in cases we're not interested.

PR #4983, opened 2020-12-07
Add host.settings.disablePaypalDonations

kewitz:
Also fixes disablePaypalPayouts logic.

### DIFF
--- a/server/graphql/v2/object/Host.js
+++ b/server/graphql/v2/object/Host.js
@@ -89,7 +89,7 @@ export const Host = new GraphQLObjectType({
             supportedPaymentMethods.push('CREDIT_CARD');
           }
 
-          if (find(connectedAccounts, ['service', 'paypal'])) {
+          if (find(connectedAccounts, ['service', 'paypal']) && !collective.settings?.disablePaypalDonations) {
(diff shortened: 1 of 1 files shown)


server/graphql/v2/object/Host.js
```diff
@@ -130,7 +130,7 @@ export const Host = new GraphQLObjectType({
           if (connectedAccounts?.find?.(c => c.service === 'transferwise')) {
             supportedPayoutMethods.push(PayoutMethodTypes.BANK_ACCOUNT);
           }
-          if (connectedAccounts?.find?.(c => c.service === 'paypal') || !collective.settings?.disablePaypalPayouts) {
+          if (connectedAccounts?.find?.(c => c.service === 'paypal') && !collective.settings?.disablePaypalPayouts) {
```
znarf, line 133: Are you sure about this?

With the old PayPal payouts, a Connected Account was not necessarily needed?

znarf: Unless 100% sure, I would not touch this now.

kewitz: You're correct, I'll fix this.

PR #4993, opened 2020-12-09
Price and Fees fixes and updates

kewitz:
pepesilvia

Add tests to ensure we're properly charging the host revenue share in Stripe.
Fix combined case of Platform Tip over Host Revenue share.
Deduct already collected Host Revenue Share (Stripe) from Host Fees at the end of the month.
Fix getHostMetrics platform tip claculation (use net amount).
Reimburse owed Payment Processor Fees from collected platform tips (stripe) in the monthly settlement invoice.

### DIFF
--- a/cron/monthly/host-fees.js
+++ b/cron/monthly/host-fees.js
@@ -2,9 +2,10 @@
 import '../../server/env';
 
 import config from 'config';
+import { round, sumBy } from 'lodash';
 import { v4 as uuid } from 'uuid';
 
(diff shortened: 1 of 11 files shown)


kewitz:
@znarf I added a migration to try and fix the broken records on our ledger.
Although this migration won't fix the bug that allows this to happen, I need a bit of context on how we should fix:
https://github.com/opencollective/opencollective-api/blob/master/server/paymentProviders/opencollective/host.js#L36-L42

kewitz:
Rebased.

PR #5023, opened 2020-12-15
Optimize contributors query

kewitz:
Since tipping the platform will enlist the user as a backer of OpenCollective, we have a problem fetching all contributors with we keep the subquery.

### DIFF
--- a/server/lib/contributors.ts
+++ b/server/lib/contributors.ts
@@ -97,17 +97,6 @@ const loadContributors = async (collectiveId: number): Promise<ContributorsCache
     return fromCache;
   }
 
-  // Subquery to get the total amount contributed for each member
-  const TotalAmountContributedQuery = `
-    SELECT  COALESCE(SUM(amount), 0)
-    FROM    "Transactions"
-    WHERE   "CollectiveId" = :collectiveId
-    AND     TYPE = 'CREDIT'
-    AND     "deletedAt" IS NULL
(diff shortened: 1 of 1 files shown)


server/lib/contributors.ts
```diff
@@ -131,19 +135,20 @@ const loadContributors = async (collectiveId: number): Promise<ContributorsCache
         ARRAY_AGG(DISTINCT m."role") AS roles,
         ARRAY_AGG(DISTINCT tier."id") AS "tiersIds",
         COALESCE(MAX(m.description), MAX(tier.name)) AS description,
-        (${TotalAmountContributedQuery}) AS "totalAmountDonated"
+        COALESCE(tc."totalAmountDonated", 0) AS "totalAmountDonated"
```
Betree, line 138: If the same profile donated both directly **and** through gift cards, my understanding of this query is that it will return two entries. Shouldn't we remove `totalAmountDonated` from the `group by` and set this to:

```suggestion
        COALESCE(SUM(tc."totalAmountDonated"), 0) AS "totalAmountDonated"
```

kewitz: I tested that and It was returning a bit more than expected in some cases, and I decided to rewrite this focusing on keeping the same value.

Betree: The issue described here did materialize in the issue, we should go back to this query to make sure we group entries properly: https://github.com/opencollective/opencollective/issues/3804

PR #5033, opened 2020-12-16
Generate and store 2FA recovery codes on setup

sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3709

Frontend PR: https://github.com/opencollective/opencollective-frontend/pull/5588

* When the user sets up 2FA on their account, generates recovery codes, returns them to the frontend to be displayed once during the setup process, then they are stored securely (encrypted with the 2FA token and also hashed) in the Users table
* When the user removes 2FA from their account, it removes the recovery codes

### DIFF
--- a/migrations/20210208130000-add-2fa-recovery-codes-to-user.js
+++ b/migrations/20210208130000-add-2fa-recovery-codes-to-user.js
@@ -1,0 +1,21 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('Users', 'twoFactorAuthRecoveryCodes', {
+      type: Sequelize.ARRAY(Sequelize.STRING),
+      allowNull: true,
+      defaultValue: null,
+    });
+    await queryInterface.addColumn('UserHistories', 'twoFactorAuthRecoveryCodes', {
+      type: Sequelize.ARRAY(Sequelize.STRING),
+      allowNull: true,
+      defaultValue: null,
+    });
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.removeColumn('Users', 'twoFactorAuthRecoveryCodes');
+    await queryInterface.removeColumn('UserHistories', 'twoFactorAuthRecoveryCodes');
+  },
+};
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "cookie-parser": "1.4.5",
     "cors": "2.8.5",
     "crypto-js": "4.0.0",
+    "crypto-random-string": "3.3.0",
     "dataloader": "2.0.0",
     "dataloader-sequelize": "2.3.1",
     "debug": "4.3.1",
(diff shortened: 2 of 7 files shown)


Betree (approved):
Looks good to me :+1:

sbinlondon:
Thanks @Betree - will try and get the E2E/GraphQL tests passing while I wait on @znarf's review.

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -159,9 +174,17 @@ const accountMutations = {
...
       const encryptedText = crypto.encrypt(args.token);
 
-      await user.update({ twoFactorAuthToken: encryptedText });
+      /** Generate recovery codes, hash and store them in the table, and return them to the user to write down */
+      const recoveryCodesArray = Array.from({ length: 6 }, () =>
+        cryptoRandomString({ length: 16, type: 'distinguishable' }),
+      );
+      const hashedRecoveryCodesArray = recoveryCodesArray.map(code => {
+        return crypto.hash(code);
+      });
```
kewitz, line 183: I don't think we need to hash them, we might want to encrypt the DB field tho.
Also, there's no purpose in using distinguishable codes if we're going to hash it later on.

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -17,6 +18,20 @@ import { Host } from '../object/Host';
 import { Individual } from '../object/Individual';
 import AccountSettingsKey from '../scalar/AccountSettingsKey';
 
+const AccountWithRecoveryCodes = new GraphQLObjectType({
```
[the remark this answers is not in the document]
sbinlondon: You're right, I'll update it to `Individual`

server/graphql/v2/mutation/AccountMutations.ts
sbinlondon: We were going to just encrypt them but @znarf has requested for them to be hashed for extra security. https://github.com/opencollective/opencollective/issues/3709#issuecomment-746120617

znarf: I haven't asked for **extra** hashing, I just asked for hashing. Why?

Hashing is more secure than encrypting, as it's one way and can't be decoded.

Whenever you can use hashing instead of encrypting always use hashing. We're currently using encryption for cases when hashing is not practical, when data needs to be decrypted.

There is no need to add encryption on top of hashing, it's a useless layer showing you're not 100% comfortable with hashing as a security measure.

sbinlondon: The code only hashes them now anyways. Was momentarily confused by the old discussion on the issue.

package.json
```diff
@@ -39,6 +39,7 @@
     "cookie-parser": "1.4.5",
     "cors": "2.8.5",
     "crypto-js": "4.0.0",
+    "crypto-random-string": "^3.3.0",
```
znarf, line 42: We use fixed dependencies.

znarf (approved):
Looks good for me.

migrations/20210208130000-add-2fa-recovery-codes-to-user.js
```diff
@@ -0,0 +1,21 @@
...
+      type: Sequelize.ARRAY(Sequelize.STRING),
+      allowNull: true,
+      defaultValue: null,
+    });
+    await queryInterface.addColumn('UserHistories', 'twoFactorAuthRecoveryCodes', {
+      type: Sequelize.ARRAY(Sequelize.STRING),
+      allowNull: true,
+      defaultValue: null,
+    });
+  },
```
znarf, line 15: Lets's look at it closely when this is deployed to production, @Betree or myself should be aware there is a migration going on.

PR #5052, opened 2020-12-23
Attach platform tip receipt and charge details in emails

znarf:
Fix https://github.com/opencollective/opencollective/issues/3598

Screenshot 2020-12-23 at 14 35 59

### DIFF
--- a/server/lib/notifications.js
+++ b/server/lib/notifications.js
@@ -111,7 +111,6 @@ async function notifySubscribers(users, activity, options = {}) {
 }
 
 async function notifyUserId(UserId, activity, options = {}) {
-  let pdf;
   const user = await models.User.findByPk(UserId);
   debug('notifyUserId', UserId, user && user.email, activity.type);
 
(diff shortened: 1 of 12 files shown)


server/lib/notifications.js
```diff
@@ -120,19 +119,38 @@ async function notifyUserId(UserId, activity, options = {}) {
...
       activity.data.transactionPdf = true;
     }
+
+    if (transaction.hasPlatformTip()) {
+      const platformTipTransaction = await transaction.getPlatformTipTransaction();
+      if (platformTipTransaction) {
+        const platformTipPdf = await getTransactionPdf(platformTipTransaction, user);
+        if (platformTipPdf) {
+          const createdAtString = toIsoDateStr(
+            platformTipTransaction.createdAt ? new Date(platformTipTransaction.createdAt) : new Date(),
```
Betree, line 143: This ternary condition was there in the previous version but I didn't pay attention to it. I'm a bit surprised now, because `createdAt` is not nullable.

server/models/Transaction.js
```diff
@@ -305,6 +305,26 @@ export default (Sequelize, DataTypes) => {
     return Transaction.findByPk(this.RefundTransactionId);
   };
 
+  Transaction.prototype.getFeeOnTopTransaction = function () {
+    return this.getPlatformTipTransaction();
+  };
```
Betree, line 310: I don't get the benefit of providing an alias, shouldn't we just call it "platform tip" and always use `getPlatformTipTransaction`?

server/lib/payments.js
```diff
@@ -432,18 +425,34 @@ const sendOrderConfirmedEmail = async (order, transaction) => {
...
+              platformTipTransaction.createdAt ? new Date(platformTipTransaction.createdAt) : new Date(),
+            );
+            attachments.push({
+              filename: `transaction_opencollective_${createdAtString}_${platformTipTransaction.uuid}.pdf`,
+              content: platformTipPdf,
+            });
+            data.platformTipPdf = true;
+          }
+        }
+      }
```
Betree, line 453: We duplicate the same code multiple times here, would be great to put that in a common helper like `attachPdfReceiptsToEmail`

znarf: Agree and thought about it.

But it's a bit of a mess of variables and side effects and I was not able to find an elegant API for it.

server/models/Transaction.js
znarf: To have a smooth transition, I wanted to express they're the same thing. But you might be right.

PR #5063, opened 2020-12-30
fix yearly report for hosts

xdamman:
I tried to run the yearly report for the allforclimate host and it wasn't working.
Took me a bit of time to figure out what was going on because it was not returning any error.

Turns out if SKIP_PDF is true or if the host report is the yearly report, you are returning before the end of the function instead of simply skipping PDF generation.

This fixes it. You may want to merge this before Jan 1 (but if I'm not mistaken you need to manually trigger this yearly report).

### DIFF
--- a/reports/host-report.js
+++ b/reports/host-report.js
@@ -248,8 +248,7 @@ async function HostReport(year, month, hostId) {
       data.transactions = transactions;
       // Don't generate PDF in email if it's the yearly report
       let pdf;
-      if (yearlyReport || process.env.SKIP_PDF) {
-      } else {
+      if (!yearlyReport && !process.env.SKIP_PDF) {
         pdf = await exportToPDF('expenses', data, {
           paper: host.currency === 'USD' ? 'Letter' : 'A4',
         }).catch(error => {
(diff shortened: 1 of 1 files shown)


znarf:
Seems it was broken in a recent refactor from @kewitz here: https://github.com/opencollective/opencollective-api/commit/4186198206e92116e2c06d486a56c04d326a5951 (PR: https://github.com/opencollective/opencollective-api/pull/4975)

Can you have a look when you're back @kewitz ?

Not urgent because all reports are currently turned off. They will be reactivated around January 4th when everyone is back on board.

reports/host-report.js
```diff
@@ -361,15 +362,14 @@ async function HostReport(year, month, hostId) {
...
-
       // Don't send transactions in email if there is more than 1000
       if (data.transactions.length > 1000) {
         delete data.transactions;
       }
       const admins = await getHostAdminsEmails(host);
       await sendEmail(admins, data, attachments);
     } catch (e) {
-      console.error(`Error in processing host ${host.slug}:`);
+      console.error(`Error in processing host ${host.slug}:`, e);
```
kewitz, line 372: :+1:

kewitz (approved):
LGTM.

kewitz:
Sorry about the trouble, @xdamman!

PR #5075, opened 2021-01-04
feat(ocf-application): handle user creation in createCollective for OCF host application

flickz:
Following up https://github.com/opencollective/opencollective-frontend/pull/5623

### DIFF
--- a/server/graphql/v2/input/UserCreateInput.js
+++ b/server/graphql/v2/input/UserCreateInput.js
@@ -1,0 +1,9 @@
+import { GraphQLInputObjectType, GraphQLNonNull, GraphQLString } from 'graphql';
+
+export const UserCreateInput = new GraphQLInputObjectType({
+  name: 'UserCreateInput',
+  fields: () => ({
+    name: { type: new GraphQLNonNull(GraphQLString) },
+    email: { type: new GraphQLNonNull(GraphQLString) },
+  }),
+});
--- a/server/graphql/v2/mutation/CreateCollectiveMutation.js
+++ b/server/graphql/v2/mutation/CreateCollectiveMutation.js
@@ -25,7 +25,7 @@ async function createCollective(_, args, req) {
 
   let user = remoteUser;
 
-  if (!user && args.user && args.host.slug === 'foundation') {
+  if (!user && args.user && args.host.legacyId === defaultHostCollective('foundation').CollectiveId) {
     user = await models.User.findByEmail(args.user.email);
     if (!user) {
       user = await models.User.createUserWithCollective(args.user);
--- a/server/graphql/v2/mutation/CreateCollectiveMutation.js
+++ b/server/graphql/v2/mutation/CreateCollectiveMutation.js
@@ -27,9 +27,13 @@ async function createCollective(_, args, req) {
   let user = remoteUser,
     host;
 
+  if (args.host) {
+    host = await fetchAccountWithReference(args.host, { loaders });
+  }
+
   return sequelize
     .transaction(async transaction => {
-      if (!user && args.user && args.host.legacyId === defaultHostCollective('foundation').CollectiveId) {
+      if (!user && args.user && host.id === defaultHostCollective('foundation').CollectiveId) {
         user = await models.User.findByEmail(args.user.email, transaction);
         if (!user) {
           user = await models.User.createUserWithCollective(args.user, transaction);
@@ -46,7 +50,6 @@ async function createCollective(_, args, req) {
         isActive: false,
         CreatedByUserId: user.id,
         settings: { ...DEFAULT_COLLECTIVE_SETTINGS, ...args.collective.settings },
-        ...(args.applicationData && { data: args.applicationData }),
       };
 
       if (isCollectiveSlugReserved(collectiveData.slug)) {
@@ -95,7 +98,6 @@ async function createCollective(_, args, req) {
           collectiveData.tags.push('open source');
         }
       } else if (args.host) {
-        host = await fetchAccountWithReference(args.host, { loaders });
         if (!host) {
           throw new ValidationFailed('Host Not Found');
         }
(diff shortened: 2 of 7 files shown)


server/graphql/v2/input/UserCreateInput.js
```diff
@@ -0,0 +1,9 @@
+import { GraphQLInputObjectType, GraphQLNonNull, GraphQLString } from 'graphql';
+
+export const UserCreateInput = new GraphQLInputObjectType({
```
Betree, line 3: @znarf from the new nomenclature, shouldn't that be `IndividualCreateInput`?

Betree (changes requested):

server/graphql/v2/input/UserCreateInput.js
znarf: Yeah, `IndividualCreateInput` is better.

server/graphql/v2/mutation/CreateCollectiveMutation.js
```diff
@@ -22,96 +24,119 @@ async function createCollective(_, args, req) {
...
-        where: { CollectiveId: remoteUser.CollectiveId, service: 'github' },
-      });
-      if (!githubAccount) {
-        throw new Error('You must have a connected GitHub Account to create a collective with GitHub.');
+  let user = remoteUser,
+    host;
+
+  return sequelize
+    .transaction(async transaction => {
+      if (!user && args.user && args.host.legacyId === defaultHostCollective('foundation').CollectiveId) {
```
Betree, line 32: We should not check against `args.host.legacyId`, because it's a valid usage to specify the host by `id` rather than `legacyId` (ie. `host: { id: 'xxxxx-yyyyy' }`). We should fetch the host from the DB `fetchAccountWithReference` (that can be done outside of the transaction) and the check should be:

```suggestion
      if (!user && args.user && host.id === defaultHostCollective('foundation').CollectiveId) {
```

Betree (commented):
Great job on the SQL transaction. Those are not easy to setup with sequelize but the code you implemented looks clean.

server/graphql/v2/mutation/CreateCollectiveMutation.js
```diff
@@ -22,96 +24,119 @@ async function createCollective(_, args, req) {
...
-    await collective.addHost(host, remoteUser, { shouldAutomaticallyApprove, message: args.message });
-    purgeCacheForCollective(host.slug);
-  }
+      const collectiveData = {
+        slug: args.collective.slug.toLowerCase(),
+        ...pick(args.collective, ['name', 'description', 'tags']),
+        isActive: false,
+        CreatedByUserId: user.id,
+        settings: { ...DEFAULT_COLLECTIVE_SETTINGS, ...args.collective.settings },
+        ...(args.applicationData && { data: args.applicationData }),
```
Betree, line 49: As per https://github.com/opencollective/opencollective-api/pull/5075#discussion_r552475720, we want to store the application data in the `HostApplication`, not the collective `data`. We should pass `applicationData` to `addHost` in https://github.com/opencollective/opencollective-api/blob/master/server/graphql/v2/mutation/CreateCollectiveMutation.js#L92 then pass it down to `recordApplication` through the `customData` key in https://github.com/opencollective/opencollective-api/blob/master/server/models/Collective.js#L1929.

server/graphql/v2/mutation/CreateCollectiveMutation.js
```diff
@@ -22,96 +24,125 @@ async function createCollective(_, args, req) {
...
-    }
-    if (!host.isHostAccount) {
-      throw new ValidationFailed('Host account is not activated as Host.');
-    }
   }
 
-  const collective = await models.Collective.create(collectiveData);
+  return sequelize
+    .transaction(async transaction => {
+      if (!user && args.user && host.id === defaultHostCollective('foundation').CollectiveId) {
```
Betree, line 36: Otherwise it will crash if no host is specified

```suggestion
      if (!user && args.user && host?.id === defaultHostCollective('foundation').CollectiveId) {
```

server/graphql/v2/mutation/CreateCollectiveMutation.js
```diff
@@ -22,96 +24,125 @@ async function createCollective(_, args, req) {
...
+      }
 
-  return collective;
+      // Will send an email to the authenticated user OR newly created user (For OCF application)
+      // - tell them that their collective was successfully created
+      // - tell them that their collective is pending validation (which might be wrong if it was automatically approved)
+      const remoteUserCollective = remoteUser
+        ? await loaders.Collective.byId.load(user.CollectiveId)
+        : await models.Collective.findByPk(user.CollectiveId);
+
```
Betree, line 130: I don't get this, `loaders.Collective.byId.load` and `models.Collective.findByPk` are supposed to have the same effect

flickz: @Betree Oh yeah, thanks for picking up on that. When I was experimenting with the transaction method, `loaders.Collective.byId.load`  was failing for new users because it was within the transaction process, I used `models.Collective.findByPk` to handle the case. But I forgot to remove it after realizing the method call shouldn't be used inside a transaction.

server/graphql/v2/mutation/CreateCollectiveMutation.js
```diff
@@ -22,96 +24,122 @@ async function createCollective(_, args, req) {
...
+        purgeCacheForCollective(host.slug);
+      }
 
-  return collective;
+      // Will send an email to the authenticated user OR newly created user (For OCF application)
+      // - tell them that their collective was successfully created
+      // - tell them that their collective is pending validation (which might be wrong if it was automatically approved)
+      const remoteUserCollective = await loaders.Collective.byId.load(user.CollectiveId);
+      models.Activity.create({
+        type: activities.COLLECTIVE_CREATED,
```
znarf, line 129: I just checked and as of today this email template only makes sense when there is already an host attached to collective creation. I suggest to only trigger it when there is a host.

znarf: The problem is not in this PR so no need to do anything ...

PR #5082, opened 2021-01-06
Add Support to Post Comments on Updates

SudharakaP:
This adds the api support to post comments on updates

Related to https://github.com/opencollective/opencollective/issues/3053

### DIFF
--- a/server/graphql/common/update.js
+++ b/server/graphql/common/update.js
@@ -1,10 +1,11 @@
 import { get } from 'lodash';
 
-import { mustHaveRole } from '../../../lib/auth';
-import { purgeCacheForCollective } from '../../../lib/cache';
-import { stripTags } from '../../../lib/utils';
-import models from '../../../models';
-import { NotFound, ValidationFailed } from '../../errors';
+import { mustHaveRole } from '../../lib/auth';
+import { purgeCacheForCollective } from '../../lib/cache';
+import { stripTags } from '../../lib/utils';
+import models from '../../models';
+import { NotFound, ValidationFailed } from '../errors';
+import { idDecode, IDENTIFIER_TYPES } from '../v2/identifiers';
 
 function require(args, path) {
   if (!get(args, path)) {
@@ -13,7 +14,10 @@ function require(args, path) {
 }
 
 export async function createUpdate(_, args, req) {
-  const CollectiveId = get(args, 'update.collective.id');
+  let CollectiveId = get(args, 'update.collective.id');
+  if (!CollectiveId) {
+    CollectiveId = get(args, 'update.account.legacyId');
+  }
   mustHaveRole(req.remoteUser, 'ADMIN', CollectiveId, 'create an update');
   require(args, 'update.title');
 
@@ -42,7 +46,12 @@ export async function createUpdate(_, args, req) {
 }
 
 async function fetchUpdate(id) {
-  const update = await models.Update.findByPk(id);
+  let update;
+  if (typeof id == 'string') {
+    update = await models.Update.findByPk(idDecode(id, IDENTIFIER_TYPES.UPDATE));
+  } else {
+    update = await models.Update.findByPk(id);
+  }
   if (!update) {
     throw new NotFound(`Update with id ${id} not found`);
   }
--- a/server/graphql/v1/mutations.js
+++ b/server/graphql/v1/mutations.js
@@ -539,7 +539,13 @@ const mutations = {
         type: new GraphQLNonNull(CommentInputType),
       },
     },
-    resolve: commentMutations.createComment,
+    resolve: (_, args, req) => {
+      if (args['UpdateId']) {
+        throw new Error('Use QPI V2 to post comments on updates');
+      }
+
+      return commentMutations.createComment(_, args, req);
+    },
   },
   editComment: {
     type: CommentType,
--- a/server/graphql/v1/mutations.js
+++ b/server/graphql/v1/mutations.js
@@ -1,14 +1,10 @@
-import config from 'config';
 import { GraphQLBoolean, GraphQLInt, GraphQLList, GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
-import { pick } from 'lodash';
 
 import statuses from '../../constants/expense_status';
-import roles from '../../constants/roles';
-import emailLib from '../../lib/email';
-import logger from '../../lib/logger';
-import models, { sequelize } from '../../models';
+import models from '../../models';
 import { bulkCreateVirtualCards, createVirtualCardsForEmails } from '../../paymentProviders/opencollective/virtualcard';
 import { editPublicMessage } from '../common/members';
+import * as updateMutations from '../common/update';
 import { createUser } from '../common/user';
 import { Forbidden, NotFound, Unauthorized, ValidationFailed } from '../errors';
 
@@ -52,7 +48,6 @@ import {
 } from './mutations/orders';
 import * as paymentMethodsMutation from './mutations/paymentMethods';
 import { editTier, editTiers } from './mutations/tiers';
-import * as updateMutations from './mutations/updates';
 import { confirmUserEmail, updateUserEmail } from './mutations/users';
 import { ApplicationInputType, ApplicationType } from './Application';
 import { CollectiveInterfaceType } from './CollectiveInterface';
@@ -480,6 +475,7 @@ const mutations = {
   },
   createUpdate: {
     type: UpdateType,
+    deprecationReason: 'This endpoint has been moved to GQLV2',
     args: {
       update: {
         type: new GraphQLNonNull(UpdateInputType),
--- a/server/graphql/v2/input/CommentCreateInput.js
+++ b/server/graphql/v2/input/CommentCreateInput.js
@@ -1,6 +1,7 @@
 import { GraphQLInputObjectType, GraphQLInt, GraphQLString } from 'graphql';
 
 import { ExpenseReferenceInput } from './ExpenseReferenceInput';
+import { UpdateReferenceInput } from './UpdateReferenceInput';
 
 /**
  * Input type to use as the type for the comment input in createComment mutation.
@@ -18,5 +19,6 @@ export const CommentCreateInput = new GraphQLInputObjectType({
       deprecationReason: '2019-02-26: Please use the expense field',
     },
     ConversationId: { type: GraphQLString },
+    update: { type: UpdateReferenceInput },
   }),
 });
--- a/server/graphql/v2/mutation/CommentMutations.js
+++ b/server/graphql/v2/mutation/CommentMutations.js
@@ -5,6 +5,7 @@ import { getDecodedId, idDecode, IDENTIFIER_TYPES } from '../identifiers';
 import { CommentCreateInput } from '../input/CommentCreateInput';
 import { CommentUpdateInput } from '../input/CommentUpdateInput';
 import { fetchExpenseWithReference } from '../input/ExpenseReferenceInput';
+import { fetchUpdateWithReference } from '../input/UpdateReferenceInput';
 import { Comment } from '../object/Comment';
 
 const commentMutations = {
@@ -44,10 +45,18 @@ const commentMutations = {
         args.comment.ConversationId = idDecode(args.comment.ConversationId, IDENTIFIER_TYPES.CONVERSATION);
       }
 
+      if (args.comment.update) {
+        const update = await fetchUpdateWithReference(args.comment.update, {
+          loaders: req.loaders,
+          throwIfMissing: true,
+        });
+        args.comment.UpdateId = update.id;
+      }
+
       if (args.comment.expense) {
         const expense = await fetchExpenseWithReference(args.comment.expense, req);
         if (!expense) {
-          throw new Error('This expense does not exists');
+          throw new Error('This expense does not exist');
         }
         args.comment.ExpenseId = expense.id;
       }
--- a/server/graphql/v2/mutation/CommentMutations.js
+++ b/server/graphql/v2/mutation/CommentMutations.js
@@ -1,6 +1,7 @@
 import { GraphQLNonNull, GraphQLString } from 'graphql';
 
 import { createCommentResolver, deleteComment, editComment } from '../../common/comment';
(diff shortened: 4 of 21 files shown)


SudharakaP:
@Betree : This is ready for review as well. 😄

server/graphql/v1/mutations.js
```diff
@@ -478,64 +473,6 @@ const mutations = {
       return addFundsToCollective(args.order, req.remoteUser);
     },
   },
-  createUpdate: {
```
Betree, line 481: It's fantastic that you implemented these mutations in V2, but we should not remove them from V1 directly - otherwise the people running an "old" version of the frontend (because we haven't deployed it before the API update, or because they haven't refreshed their browser yet) will get a crash when using them.

The way to go with this kind of changes is:
1. Implement the new mutation on V2
2. Mark the one on V1 as deprecated with `deprecationReason` 
3. Deploy everything
4. Remove mutation from V1 after a few days

server/graphql/v2/input/CommentCreateInput.js
```diff
@@ -18,5 +18,6 @@ export const CommentCreateInput = new GraphQLInputObjectType({
       deprecationReason: '2019-02-26: Please use the expense field',
     },
     ConversationId: { type: GraphQLString },
+    UpdateId: { type: GraphQLString },
```
Betree, line 21: The pattern we used for `ConversationId` is not the right one. The latest best practice is to have a `*ReferenceInput` (ie. `ExpenseReferenceInput`) so that we can refer the entity either by it's old ID, new ID, or slug if it has one.

We should create an `UpdateReferenceInput`, and have this here instead:
```suggestion
    update: { type: UpdateReferenceInput },
```

server/graphql/v2/input/UpdateAttributesInput.js
```diff
@@ -0,0 +1,19 @@
+import { GraphQLBoolean, GraphQLInputObjectType, GraphQLString } from 'graphql';
+
+import { IsoDateString } from '../../v1/types';
+
+import { AccountReferenceInput } from './AccountReferenceInput';
+
+export const UpdateAttributesInput = new GraphQLInputObjectType({
+  name: 'UpdateAttributesInput',
```
Betree, line 8: Naming convention for this one would be:
```suggestion
export const UpdateUpdateInput = new GraphQLInputObjectType({
  name: 'UpdateUpdateInput',
```

I agree it looks really strange :smile:

server/graphql/v2/input/UpdateAttributesInput.js
```diff
@@ -0,0 +1,19 @@
...
+  description: 'Input type for UpdateType',
+  fields: () => ({
+    id: { type: GraphQLString },
+    slug: { type: GraphQLString },
+    title: { type: GraphQLString },
+    isPrivate: { type: GraphQLBoolean },
+    makePublicOn: { type: IsoDateString },
```
Betree, line 15: From `server/graphql/v2/scalar/ISODateTime.ts`

```suggestion
    makePublicOn: { type: ISODateTime },
```

server/graphql/v2/input/UpdateAttributesInput.js
```diff
@@ -0,0 +1,19 @@
...
+
+import { IsoDateString } from '../../v1/types';
+
+import { AccountReferenceInput } from './AccountReferenceInput';
+
+export const UpdateAttributesInput = new GraphQLInputObjectType({
+  name: 'UpdateAttributesInput',
+  description: 'Input type for UpdateType',
+  fields: () => ({
+    id: { type: GraphQLString },
```
Betree, line 11:
```suggestion
    id: { type: new GraphQLNonNull(GraphQLString) },
```

server/graphql/v2/input/UpdateInput.js
```diff
@@ -0,0 +1,23 @@
+import { GraphQLBoolean, GraphQLInputObjectType, GraphQLInt, GraphQLNonNull, GraphQLString } from 'graphql';
+
+import { CollectiveAttributesInputType, TierInputType } from '../../v1/inputTypes';
+import { IsoDateString } from '../../v1/types';
+
+export const UpdateInput = new GraphQLInputObjectType({
+  name: 'UpdateInput',
```
Betree, line 7: Naming convention

```suggestion
export const UpdateCreateInput = new GraphQLInputObjectType({
  name: 'UpdateCreateInput',
```

server/graphql/v2/input/UpdateInput.js
```diff
@@ -0,0 +1,23 @@
...
+  description: 'Input type for UpdateType',
+  fields: () => ({
+    id: { type: GraphQLInt },
```
Betree, line 10: No need for ID when creating

```suggestion
```

server/graphql/v2/input/UpdateInput.js
```diff
@@ -0,0 +1,23 @@
...
+    views: { type: GraphQLInt },
```
Betree, line 11: Not used anymore

```suggestion
```

server/graphql/v2/input/UpdateInput.js
```diff
@@ -0,0 +1,23 @@
...
+    slug: { type: GraphQLString },
```
Betree, line 12: I think it's auto-generated

```suggestion
```

server/graphql/v2/input/UpdateInput.js
```diff
@@ -0,0 +1,23 @@
...
+    title: { type: GraphQLString },
```
Betree, line 13:
```suggestion
    title: { type: new GraphQLNonNull(GraphQLString) },
```

server/graphql/v2/input/UpdateInput.js
```diff
@@ -0,0 +1,23 @@
...
+    image: { type: GraphQLString },
```
Betree, line 14: I don't think it's used today

```suggestion
```

server/graphql/v2/input/UpdateInput.js
```diff
@@ -0,0 +1,23 @@
...
+    isPrivate: { type: GraphQLBoolean },
+    makePublicOn: { type: IsoDateString },
+    markdown: { type: GraphQLString },
```
Betree, line 17: deprecated

```suggestion
```

server/graphql/v2/input/UpdateInput.js
```diff
@@ -0,0 +1,23 @@
...
+    html: { type: GraphQLString },
+    fromCollective: { type: CollectiveAttributesInputType },
+    collective: { type: new GraphQLNonNull(CollectiveAttributesInputType) },
```
Betree, line 20: Should be `fromAccount`/`account`, with `AccountReferenceInput` as types

server/graphql/v2/input/UpdateInput.js
```diff
@@ -0,0 +1,23 @@
...
+    tier: { type: TierInputType },
```
Betree, line 21: Not used

```suggestion
```

server/graphql/v2/mutation/UpdateMutations.js
```diff
@@ -0,0 +1,70 @@
...
+
+import { createUpdate, deleteUpdate, editUpdate, publishUpdate, unpublishUpdate } from '../../common/update';
+import { UpdateAudienceType } from '../enum';
+import { UpdateAttributesInput } from '../input/UpdateAttributesInput';
+import { UpdateInput } from '../input/UpdateInput';
+import Update from '../object/Update';
+
+const updateMutations = {
+  createUpdate: {
+    type: Update,
```
Betree, line 11:
```suggestion
    type: new GraphQLNonNull(Update),
```

server/graphql/v2/mutation/UpdateMutations.js
```diff
@@ -0,0 +1,70 @@
...
+      update: {
+        type: new GraphQLNonNull(UpdateInput),
+      },
+    },
+    resolve(_, args, req) {
+      return createUpdate(_, args, req);
+    },
+  },
+  editUpdate: {
+    type: Update,
```
Betree, line 22:
```suggestion
    type: new GraphQLNonNull(Update),
```

server/graphql/v2/mutation/UpdateMutations.js
```diff
@@ -0,0 +1,70 @@
...
+      update: {
+        type: new GraphQLNonNull(UpdateAttributesInput),
+      },
+    },
+    resolve(_, args, req) {
+      return editUpdate(_, args, req);
+    },
+  },
+  publishUpdate: {
+    type: Update,
```
Betree, line 33:
```suggestion
    type: new GraphQLNonNull(Update),
```

server/graphql/v2/mutation/UpdateMutations.js
```diff
@@ -0,0 +1,70 @@
...
+      notificationAudience: {
+        type: new GraphQLNonNull(UpdateAudienceType),
+      },
+    },
+    resolve(_, args, req) {
+      return publishUpdate(_, args, req);
+    },
+  },
+  unpublishUpdate: {
+    type: Update,
```
Betree, line 47:
```suggestion
    type: new GraphQLNonNull(Update),
```

server/graphql/v2/mutation/UpdateMutations.js
```diff
@@ -0,0 +1,70 @@
...
+      id: {
+        type: new GraphQLNonNull(GraphQLString),
+      },
+    },
+    resolve(_, args, req) {
+      return unpublishUpdate(_, args, req);
+    },
+  },
+  deleteUpdate: {
+    type: Update,
```
Betree, line 58:
```suggestion
    type: new GraphQLNonNull(Update),
```

server/graphql/v2/object/Update.js
```diff
@@ -67,6 +69,29 @@ const Update = new GraphQLObjectType({
           return req.loaders.Collective.byId.load(update.CollectiveId);
         },
       },
+      comments: {
+        type: CommentCollection,
```
Betree, line 73: Non null (we return an empty list if no entry)

server/graphql/v2/object/Update.js
```diff
@@ -19,7 +21,7 @@ const Update = new GraphQLObjectType({
       slug: { type: new GraphQLNonNull(GraphQLString) },
       userCanSeeUpdate: {
         description: 'Indicates whether or not the user is allowed to see the content of this update',
-        type: new GraphQLNonNull(GraphQLBoolean),
+        type: GraphQLBoolean,
```
Betree, line 24: Having that non-null was actually welcome, there's no value in having that nullable. If it's crashing you can do:

```
return Boolean(req.remoteUser && req.remoteUser.canSeeUpdates(update.CollectiveId));
```

Betree (commented):
Nice start, and great job updating the tests.

SudharakaP:
@Betree : Thanks much for the detailed review. I've made all the changes you suggested along with the relevant front end changes. When this is deployed and the V1 mutations needs to be removed please ping me and I'll open a new PR. 😄

server/graphql/v2/mutation/CommentMutations.js
```diff
@@ -44,10 +45,18 @@ const commentMutations = {
         args.comment.ConversationId = idDecode(args.comment.ConversationId, IDENTIFIER_TYPES.CONVERSATION);
       }
 
+      if (args.comment.update) {
+        const update = await fetchUpdateWithReference(args.comment.update, req);
+        if (!update) {
+          throw new Error('This update does not exist');
+        }
```
Betree, line 52: Thank to the code you've implemented we can simplify this as:

```suggestion
        const update = await fetchUpdateWithReference(args.comment.update, { loaders: req.loaders, throwIfMissing: true });
```

server/graphql/common/update.js
```diff
@@ -0,0 +1,84 @@
...
+    FromCollectiveId: req.remoteUser.CollectiveId,
+    makePublicOn: args.update.makePublicOn,
+  });
+
+  purgeCacheForCollective(collective.slug);
+  return update;
+}
+
+async function fetchUpdate(id) {
+  const update = await models.Update.findByPk(idDecode(id, IDENTIFIER_TYPES.UPDATE));
```
Betree, line 46: If I understand correctly, since this function is called from both GQLV1 & GQLV2, `id` can be received as either a `string` or an `integer`. We should make sure it works with both.

Betree (commented):
Great update!

server/graphql/common/update.js
SudharakaP: @Betree : I've kept `v1/mutations/updates.js` as it is; https://github.com/opencollective/opencollective-api/blob/comments-on-updates/server/graphql/v1/mutations/updates.js

This one is the one that is triggered for the V1 endpoint. So my understanding is that it will be working for both. I'll remove `v1/mutations/updates.js` in the next PR after this is deployed and we are ready to deprecate the V1 endpoints. 😄

server/graphql/v2/mutation/CommentMutations.js
SudharakaP: @Betree : Thanks, I've committed this change. 👍🏽

server/graphql/common/update.js
Betree: But if you call this `fetchUpdate` from V1, won't it be called with the integer ID like that? 

```
fetchUpdate(4242);
```

SudharakaP: @Betree : Yes you are correct indeed. Sorry my mistake, was confused with this one. Anyways, I made the change to handle both. 😄 Let me know if you see any issues.

server/graphql/v2/object/Update.js
Betree: I ended up making it nullable in case user don't have permissions to see the comments

Betree (approved):
Made some minor update to the email templates to include the new info and improved the permissions (ie. can't access comments if udpate is private). Feature itself worked great, good job @SudharakaP!

SudharakaP:
> Made some minor update to the email templates to include the new info and improved the permissions (ie. can't access comments if udpate is private). Feature itself worked great, good job @SudharakaP!

Thanks for the updates and you are welcome. 😄

PR #5085, opened 2021-01-06
Fix: broken amount bug affecting Transactions with tips

kewitz:


znarf (approved):

kewitz:
Related to https://github.com/opencollective/opencollective/issues/3810.

PR #5097, opened 2021-01-09
Prevents the GraphQL Public API Being Accessed without an API Key

SudharakaP:
Resolves https://github.com/opencollective/opencollective/issues/3817


SudharakaP:
I misunderstood. As per, https://github.com/opencollective/opencollective/issues/3817#issuecomment-757136934 this PR isn't needed.

PR #5110, opened 2021-01-13
createOrder: Return error code when account already exists

Betree:
Fix https://sentry.io/organizations/open-collective/issues/2112154894

This error code will be handy for detecting & i18n the error.

Also fixed an error with "Unique UserId" that appeared when contributing from a second device with the same email.

### DIFF
--- a/migrations/20210114082711-remove-unique-guestoken-constraint.js
+++ b/migrations/20210114082711-remove-unique-guestoken-constraint.js
@@ -1,0 +1,19 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(`
+      ALTER TABLE "GuestTokens"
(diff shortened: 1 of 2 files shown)


migrations/20210114082711-remove-unique-guestoken-constraint.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(`
+      ALTER TABLE "GuestTokens"
+      DROP CONSTRAINT "GuestTokens_UserId_key";
```
Betree, line 7: This is the fix for the `UserId` issue

kewitz: Agreed! 👍

PR #5112, opened 2021-01-13
Fix repeated contributors bug

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3804

I had to do it a bit different from what was suggested because that would end up multiplying the total amount donated by the number of memberships, effectively showing that Triplebyte had donated $230 instead of $115.

Also, I think bringing the tier query to `member_collectives_matching_roles` and removing some conditionals from the main select actually simplify it further.

Let me know if you think there's an even simpler way to accomplish this.

### DIFF
--- a/server/lib/contributors.ts
+++ b/server/lib/contributors.ts
@@ -100,14 +100,27 @@ const loadContributors = async (collectiveId: number): Promise<ContributorsCache
   const allContributors = await sequelize.query(
     `
       WITH member_collectives_matching_roles AS (
-        SELECT      c.*
-        FROM        "Collectives" c
-        INNER JOIN  "Members" m ON m."MemberCollectiveId" = c.id
-        WHERE       m."CollectiveId" = :collectiveId
-        AND         c."id" != :collectiveId
-        AND         m."deletedAt" IS NULL AND c."deletedAt" IS NULL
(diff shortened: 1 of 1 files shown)


server/lib/contributors.ts
```diff
@@ -119,36 +132,37 @@ const loadContributors = async (collectiveId: number): Promise<ContributorsCache
...
       GROUP BY
-        mc.id, "totalAmountDonated"
+        mc.id,
+        mc.since,
+        mc.roles,
+        mc."tiersIds",
+        mc."publicMessage",
+        mc."isIncognito",
+        mc."isGuest",
+        mc."memberDescription"
```
Betree, line 165: :ice_cream: I feel like we shouldn't be grouping on these fields, because they're not what makes the entry "unique". I'd rather have them in aggregate functions.

Betree (commented):
Small comment below but the query looks good. It was a mistake from me to not implement a test for this query when I first implemented it, it would have helped to spot regressions. We do not need to do it now as the priority is just to fix it, but something to keep in mind for the future.

server/lib/contributors.ts
kewitz: Agree.

PR #5117, opened 2021-01-14
addFunds: Allow for all hosts if source has no budget

Betree:
Fix https://github.com/opencollective/opencollective/issues/3837

Followup on https://github.com/opencollective/opencollective-api/pull/5103 by relaxing the condition if the `source` doesn't have a budget (users or hosts orgs without a budget activated).

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -968,6 +968,16 @@ export default function (Sequelize, DataTypes) {
     return this;
   };
 
+  Collective.prototype.hasBudget = () => {
+    if ([types.COLLECTIVE, types.EVENT, types.PROJECT, types.FUND].includes(this.type)) {
+      return true;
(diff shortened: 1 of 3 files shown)


server/models/Collective.js
```diff
@@ -968,6 +968,16 @@ export default function (Sequelize, DataTypes) {
...
 
+  Collective.prototype.hasBudget = () => {
+    if ([types.COLLECTIVE, types.EVENT, types.PROJECT, types.FUND].includes(this.type)) {
+      return true;
+    } else if (this.type === types.ORGANIZATION) {
+      return this.isHostAccount && this.isActive;
+    } else {
+      return false;
+    }
+  };
```
znarf, line 979: Looks good!

znarf (approved):

PR #5125, opened 2021-01-18
GraphQL: Remove Update deprecations

SudharakaP:
This removes the v1 mutations related to the update now that both https://github.com/opencollective/opencollective-frontend/pull/5669 and https://github.com/opencollective/opencollective-api/pull/5082 are merged.

Related to https://github.com/opencollective/opencollective/issues/3053

### DIFF
--- a/server/graphql/common/update.js
+++ b/server/graphql/common/update.js
@@ -47,75 +47,52 @@ export async function createUpdate(_, args, req) {
   return update;
 }
 
-async function fetchUpdate(id) {
-  const update = await models.Update.findByPk(idDecode(id, IDENTIFIER_TYPES.UPDATE));
+/**
+ * Fetches the update. Throws if the update does not exists or if user is not allowed to edit it.
+ */
+async function fetchUpdateForEdit(id, remoteUser) {
+  if (!remoteUser) {
+    throw new Unauthorized('You must be logged in to edit this update');
+  } else if (!id) {
+    throw new ValidationFailed(`Update ID is required`);
+  }
+
+  const update = await models.Update.findByPk(idDecode(id, IDENTIFIER_TYPES.UPDATE), {
+    include: { association: 'collective' },
+  });
   if (!update) {
     throw new NotFound(`Update with id ${id} not found`);
+  } else if (!remoteUser.isAdminOfCollective(update.collective)) {
+    throw new Forbidden("You don't have sufficient permissions to edit this update");
   }
+
   return update;
 }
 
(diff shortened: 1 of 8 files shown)


SudharakaP:
@Betree : We've missed a GraphQL schema changed as per your comment; https://github.com/opencollective/opencollective-api/pull/5082/commits/c35dfa22d0f2b18c3ce7ad0c49239dd3df3d946f#r558238704. I've added that to this PR as well.

SudharakaP:
Just to note I've added deprecation stuff from https://github.com/opencollective/opencollective-api/pull/5139 to this as well.

Betree (commented):
We should also remove the code used to handle different arg types. For example, `id` will always be a string from now on so we don't need the handling in `fetchUpdate` anymore. We can also remove the part about markdown:

```
const markdown = args.update.markdown ? stripTags(args.update.markdown) : '';
```

SudharakaP:
> We should also remove the code used to handle different arg types. For example, `id` will always be a string from now on so we don't need the handling in `fetchUpdate` anymore. We can also remove the part about markdown:
> 
> ```
> const markdown = args.update.markdown ? stripTags(args.update.markdown) : '';
> ```

@Betree : Ops, I think I forgot to remove that. Just done that. Let me know if that looks good now. 😄

Betree (commented):
@SudharakaP It's a good practice to mark mutations as deprecated before removing them (allowing a certain grace period). I did so for the update mutations in https://github.com/opencollective/opencollective-api/pull/5204, we'll merge this one a little bit later.

SudharakaP:
> @SudharakaP It's a good practice to mark mutations as deprecated before removing them (allowing a certain grace period). I did so for the update mutations in #5204, we'll merge this one a little bit later.

Thanks a lot for letting me know. I'll keep that in mind for the future. So I've went ahead and rebased this PR. 😺

SudharakaP:
@Betree : `test-graphql` seems failing and I don't think it's related to this PR as it's also failing with the same error at https://github.com/opencollective/opencollective-api/pull/5373 🤔

SudharakaP:
> @Betree : `test-graphql` seems failing and I don't think it's related to this PR as it's also failing with the same error at #5373 thinking

Interesting, after restarting several times it seems to pass now. 🤔 🤔 🤔

Betree:
@SudharakaP Yep this one randomly fails. It seems that the `transactions` field is not used anymore, I'm going to deprecate it.

PR #5134, opened 2021-01-18
Automatically invoice Fixed Fee per Collective

kewitz:
The fixed fee per hosted collective can be set in `host.data.plan.pricePerCollective`.
Its value represents the **price in cents and in host currency** charged monthly per active hosted collective.


znarf (approved):
Looks good, let's have a dry run or test run when this reaches production.

PR #5141, opened 2021-01-19
Display the total amount of filtered expenses in the Dashboard #3613

leonardohofling:


server/graphql/v2/query/ExpensesQuery.ts
```diff
@@ -225,9 +225,14 @@ const ExpensesQuery = {
     const order = [[args.orderBy.field, args.orderBy.direction]];
     const { offset, limit } = args;
     const result = await models.Expense.findAndCountAll({ include, where, order, offset, limit });
+    let totalAmount = 0;
+    result.rows.map(expense => {
+      totalAmount += expense.amount;
+    });
```
kewitz, line 231: Since this `totalAmount` is still affected by pagination, I want to keep it simple for now and calculate this in the frontend.

PR #5148, opened 2021-01-20
search: skipRecentAccounts parameter

znarf:
We would like this parameter to avoid featuring spam accounts that just were created and haven't got the time to be deleted by spam handlers.

Also deprecate `onlyActive` as it doesn't seem implemented anymore.

### DIFF
--- a/server/lib/search.js
+++ b/server/lib/search.js
@@ -87,7 +87,7 @@ export const searchCollectivesInDB = async (
   term,
   offset = 0,
   limit = 100,
-  { types, hostCollectiveIds, isHost, onlyActive } = {},
+  { types, hostCollectiveIds, isHost, onlyActive, skipRecentAccounts } = {},
 ) => {
   // Build dynamic conditions based on arguments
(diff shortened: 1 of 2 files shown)


server/lib/search.js
```diff
@@ -108,6 +108,10 @@ export const searchCollectivesInDB = async (
     dynamicConditions += 'AND "isActive" = TRUE ';
   }
 
+  if (skipRecentAccounts) {
+    dynamicConditions += `AND createdAt" < (NOW() - interval '2 day')`;
```
Betree, line 112:
```suggestion
    dynamicConditions += `AND "createdAt" < (NOW() - interval '2 day')`;
```

Betree (commented):
If we intend to use this parameter on the search page, I'm afraid that we'll run into issues like https://github.com/opencollective/opencollective/issues/3535. It's mainly a problem for support though, so an URL parameter to change that could do the job. The gain seems worth it.

PR #5152, opened 2021-01-21
Host: Remove isTypeOf in favor of host field

Betree:
Rollback on the approach taken in https://github.com/opencollective/opencollective-api/pull/5146.

Because an account can both be an `ORGANIZATION` and a `HOST`, `isTypeOf` was causing [some trouble](https://sentry.io/organizations/open-collective/issues/2164707214/?project=5199682&query=is%3Aunresolved+Expected+value+of+type+%22Host%22&statsPeriod=14d). I've changed the approach to instead use the workaround that we've been using until now: have a `host` field on organizations to retrieve the host fields if they're one. Also added this field on `INDIVIDUAL` to be able to do the same.

### DIFF
--- a/server/graphql/v2/object/Host.js
+++ b/server/graphql/v2/object/Host.js
@@ -23,7 +23,6 @@ export const Host = new GraphQLObjectType({
   name: 'Host',
   description: 'This represents an Host account',
   interfaces: () => [Account, AccountWithContributions],
-  isTypeOf: collective => collective.isHostAccount,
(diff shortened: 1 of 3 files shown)


znarf (approved):
Looks good. Sorry, I should have spotted this.

Betree:
> Looks good. Sorry, I should have spotted this.

The bug was difficult to spot, it only happened for some profiles

PR #5153, opened 2021-01-21
enhancement(Update): free up slug before deletion

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/3876

- [x] When deleting an update, add a `-${Date.now()}` suffix to the existing slug (ideally in an `beforeDestroy` sequelize hook)
- [x] Add a test in `test/server/models/Update.test.js` to make sure the slug is freed up

### DIFF
--- a/test/server/models/Update.test.js
+++ b/test/server/models/Update.test.js
@@ -27,6 +27,7 @@ describe('server/models/Update', () => {
         { id: 2, title: 'update 2 - today', isPrivate: true, makePublicOn: today },
         { id: 3, title: 'update 3 - tomorrow', isPrivate: true, makePublicOn: tomorrow },
         { id: 4, title: 'update 4', isPrivate: true, makePublicOn: null },
+        { id: 5, title: 'unique-slug', isPrivate: false, makePublicOn: null },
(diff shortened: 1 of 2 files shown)


test/server/models/Update.test.js
```diff
@@ -58,4 +59,14 @@ describe('server/models/Update', () => {
...
     });
   });
+
+  describe('delete update', () => {
+    it('frees up current slug when deleted', async () => {
+      const slug = 'unique-slug';
+      const update = await models.Update.findByPk(5);
+      await update.destroy();
+
+      expect(slug).to.not.be.equal(update.slug);
```
Betree, line 69: If I add a `console.log(update.slug);` after this line, I get `unique-slug1` as a result, which is not what we want. Can we improve the test to make sure the output slug is formatted as we expect? If you need to reload soft-deleted data at some point, you can pass `paranoid: false` to sequelize

snitin315: fixed and improved test too.

Screenshot at 2021-01-22 16-54-20

Betree (approved):

PR #5162, opened 2021-01-22
feat(hostApplication): add customData to host application query

flickz:
Resolves https://github.com/opencollective/opencollective/issues/3844


Betree (approved):
@flickz I added a small commit to your PR to limit the access to pending applications data to hosts only. I should have done this initially, now that we're adding more info in these it's better if we keep it private.

PR #5165, opened 2021-01-24
New Backer email: Display amount in title

Yetizone:
Fixes https://github.com/opencollective/opencollective/issues/3823

Besides adding the suffix described in the issue this PR also adds some mock data for the email template in `compile-email.js`
I used the mock data for testing, but since there was no other previous data I thought I might as well leave it in the commit.

Used MailDev for testing: 
MailDev_Screenshot

### DIFF
--- a/server/lib/handlebars.js
+++ b/server/lib/handlebars.js
@@ -2,7 +2,7 @@ import handlebars from 'handlebars';
 import { lowercase } from 'lodash';
 import moment from 'moment-timezone';
 
-import { capitalize, formatCurrencyObject, pluralize, resizeImage } from './utils';
+import { capitalize, formatCurrency, formatCurrencyObject, pluralize, resizeImage } from './utils';
 
 // from https://stackoverflow.com/questions/8853396/logical-operator-in-a-handlebars-js-if-conditional
 handlebars.registerHelper('ifCond', function (v1, operator, v2, options) {
@@ -159,6 +159,26 @@ handlebars.registerHelper('encodeURIComponent', str => {
 });
 
 handlebars.registerHelper('formatCurrencyObject', (obj, props) => formatCurrencyObject(obj, props.hash));
+
+handlebars.registerHelper('formatOrderAmountWithInterval', order => {
+  if (!order.currency || !order.totalAmount) {
+    return null;
+  }
+
+  const formattedAmount = formatCurrency(order.totalAmount, order.currency);
+  const subscription = order.subscription;
+  const interval = subscription?.interval || order.interval;
+
+  if (interval !== null) {
+    if (interval === 'month') {
+      return `(${formattedAmount}/m)`;
+    } else if (interval === 'year') {
+      return `(${formattedAmount}/y)`;
+    }
+  } else {
+    return `(${formattedAmount})`;
+  }
+});
 
 handlebars.registerHelper('debug', console.log);
 
--- a/test/server/graphql/v1/mutation.test.js
+++ b/test/server/graphql/v1/mutation.test.js
@@ -576,7 +576,7 @@ describe('server/graphql/v1/mutation', () => {
           expect(emailSendMessageSpy.firstCall.args[0]).to.equal('user2@opencollective.com');
           expect(emailSendMessageSpy.firstCall.args[1]).to.equal('Your Organization on Open Collective');
           expect(emailSendMessageSpy.secondCall.args[0]).to.equal('user1@opencollective.com');
-          expect(emailSendMessageSpy.secondCall.args[1]).to.equal("Google joined Scouts d'Arlon as backer");
+          expect(emailSendMessageSpy.secondCall.args[1]).to.equal("Google joined Scouts d'Arlon as backer ($20/m)");
           expect(emailSendMessageSpy.secondCall.args[2]).to.contain('Looking forward!'); // publicMessage
           expect(emailSendMessageSpy.secondCall.args[2]).to.contain(
             '@google thanks for your financial contribution to @scouts',
--- a/scripts/compile-email.js
+++ b/scripts/compile-email.js
@@ -485,6 +485,24 @@ data['user.monthlyreport'] = {
     },
   ],
 };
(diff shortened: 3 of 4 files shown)


Yetizone:
Ah, the changes broke some of the tests. I'll mark this as Draft while I look into the failing tests.

Yetizone:
Just finished fixing the `test-graphql` test which was broken by the changes in this PR. 

Looking at the output of the other failing tests, they don't seem to be related to the patch though so I think I'll mark this as ready. Do let me know if I'm mistaken and there's anything else that needs to be done.

test/server/graphql/v1/mutation.test.js
```diff
@@ -575,7 +575,7 @@ describe('server/graphql/v1/mutation', () => {
           expect(emailSendMessageSpy.firstCall.args[0]).to.equal('user2@opencollective.com');
           expect(emailSendMessageSpy.firstCall.args[1]).to.equal('Your Organization on Open Collective');
           expect(emailSendMessageSpy.secondCall.args[0]).to.equal('user1@opencollective.com');
-          expect(emailSendMessageSpy.secondCall.args[1]).to.equal("Google joined Scouts d'Arlon as backer");
+          expect(emailSendMessageSpy.secondCall.args[1]).to.equal("Google joined Scouts d'Arlon as backer $20/m");
```
Betree, line 578: Would be great to have that in parenthesis as proposed in the issue

```suggestion
          expect(emailSendMessageSpy.secondCall.args[1]).to.equal("Google joined Scouts d'Arlon as backer ($20/m)");
```

server/lib/handlebars.js
```diff
@@ -152,6 +152,30 @@ handlebars.registerHelper('encodeURIComponent', str => {
 
 handlebars.registerHelper('formatCurrencyObject', (obj, props) => formatCurrencyObject(obj, props.hash));
 
+handlebars.registerHelper('formatOrderSummary', order => {
```
Betree, line 155: Something a bit more explicit there would be welcome. I can propose:

```suggestion
handlebars.registerHelper('formatOrderAmountWithInterval', order => {
```

server/lib/handlebars.js
```diff
@@ -152,6 +152,30 @@ handlebars.registerHelper('encodeURIComponent', str => {
...
+  if (order.currency === null || order.totalAmount === null) {
+    return null;
+  }
+
+  const currencyObj = new Object();
+  currencyObj[order.currency] = order.totalAmount;
+
+  let ret = formatCurrencyObject(currencyObj);
```
Betree, line 163: Simpler alternative using `formatCurrency` from `server/lib/utils.js` + explicit variable name:

```suggestion
  const formattedAmount = formatCurrency(order.totalAmount, order.currency);
```

Betree (commented):
Looking good so far! Some improvement suggestions below

Yetizone:
Done! I've pushed an updated version of the patch with the recent suggestions.

server/lib/handlebars.js
```diff
@@ -152,6 +152,28 @@ handlebars.registerHelper('encodeURIComponent', str => {
 
 handlebars.registerHelper('formatCurrencyObject', (obj, props) => formatCurrencyObject(obj, props.hash));
 
+handlebars.registerHelper('formatOrderAmountWithInterval', order => {
+  if (order.currency === null || order.totalAmount === null) {
```
Betree, line 156: In case the fields are undefined

```suggestion
  if (!order.currency || !order.totalAmount) {
```

Betree (changes requested):
I just tried locally and it doesn't seem to be working. Here's how I tested:
- Go to /apex/donate (logged out)
- Make a contribution
- Check email on http://localhost:1080

image

Yetizone:
I've just pushed an updated version of the patch.

When donating from `/apex/donate`, `order.subscription.interval` exists but it's set to null. Instead there's `order.interval` which is set to the proper value. 

The new logic now checks if the fields exist and are not null. While writing this I was wondering if both `order.subscription.interval` and `order.interval` should exist, considering that they are very similar, but that would be a separate issue.

server/lib/handlebars.js
```diff
@@ -152,6 +152,36 @@ handlebars.registerHelper('encodeURIComponent', str => {
...
+
+  const formattedAmount = formatCurrency(order.totalAmount, order.currency);
+  const subscription = order.subscription;
+  let interval = null;
+
+  if (subscription !== undefined && subscription.interval !== null) {
+    interval = subscription.interval;
+  } else if (order.interval !== undefined && order.interval !== null) {
+    interval = order.interval;
+  }
```
Betree, line 168: This can be simplified as:

```suggestion
  const interval = subscription?.interval || order.interval;
```

server/lib/handlebars.js
```diff
@@ -152,6 +152,36 @@ handlebars.registerHelper('encodeURIComponent', str => {
...
+  if (interval !== null) {
+    let formattedAmountWithInterval = null;
+
+    if (interval === 'month') {
+      formattedAmountWithInterval = `(${formattedAmount}/m)`;
+    } else if (interval === 'year') {
+      formattedAmountWithInterval = `(${formattedAmount}/y)`;
+    }
+
+    return formattedAmountWithInterval;
```
Betree, line 179: We can simplify here too

```suggestion
    if (interval === 'month') {
      return `(${formattedAmount}/m)`;
    } else if (interval === 'year') {
      return `(${formattedAmount}/y)`;
    }
```

Betree (commented):
Looks good! Tests are not passing, but it seems unrelated to the changes here. I'll try to rebase to see if that helps.

Yetizone:
Done! Just pushed an updated version of the patch with the recent suggestions.

Betree:
I'm moving this to https://github.com/opencollective/opencollective-api/pull/5437, tests are usually not running great on forks

Yetizone:
> tests are usually not running great on forks

Oh alright, that's good to know.

PR #5173, opened 2021-01-25
Use the same sanitizer lib everywhere

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3536
API part for https://github.com/opencollective/opencollective/issues/2043
Related to https://github.com/opencollective/opencollective/issues/1805

Also enforced the deprecation of `markdown` in some places


server/lib/utils.js
```diff
@@ -59,105 +58,6 @@ export function getDomain(url = '') {
...
-        return '(token obfuscated)';
-      }
-    }
-    return value;
-  };
-
-  return sanitizeObject(cloneDeep(obj), Object.keys(obj), sanitizer);
-};
-
-String.prototype.trunc = function (n, useWordBoundary) {
```
Betree, line 127: Also removed this one as it seemed unused

PR #5178, opened 2021-01-25
Remove View All Expenses button from notification emails

kewitz:
This was requested by @piamancini due to privacy concerns.
I don't think these buttons are particularly useful since we're already linking the expense.


Betree (approved):
No blocker on my side

znarf (approved):

PR #5185, opened 2021-01-26
Remove guest tokens

Betree:
These guest tokens used to be a way to prove that you owned a guest profile. With the latest changes changes on guest contributions, it doesn't make sense to keep them.

A notion of "guest token" still exists in this refactored code, but instead of storing it in the `GuestTokens` column we store it in `order.data.guestToken`. It's a way to prove that you own the **order** that was created. That is useful when users call `confirmOrder`, because we need to make sure they're the order creator without asking them to authenticate. It will also be used in the future for https://github.com/opencollective/opencollective/issues/3834, for the same reasons.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -90,7 +90,7 @@ jobs:
 
       - name: Check matching branch
         id: check-matching-branch
-        uses: octokit/request-action@v2.x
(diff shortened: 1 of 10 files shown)


znarf (approved):
Looks good!

Betree:
Going back to draft to debut E2E test 18, that is passing locally but randomly failing on CI

PR #5187, opened 2021-01-26
Purge cache to refresh after adding funds

JoyLubega:
## Description
- This Pr adds purge cache when adding funds


znarf:
There are some problems with matching branch in CI. 

Either rebase from master on API and Frontend, or use different branch names.

JoyLubega:
> There are some problems with matching branch in CI.
> 
> Either rebase from master on API and Frontend, or use different branch names.

Okay. on it

JoyLubega:
@znarf I just rebased the API and all passed

znarf (approved):

PR #5201, opened 2021-01-29
SPAM: Unredirect URLs for domain detection

Betree:
Haven't applied the fix to `getSpamDomains` since URL domain stays in clear in redirect:

```es6
<a href="http://localhost:3000/redirect?url=https%3a%2f%2fdasilex.co.uk%2ftest">our website</a>!
```

### DIFF
--- a/server/lib/spam.ts
+++ b/server/lib/spam.ts
@@ -2,6 +2,7 @@ import fs from 'fs';
 import path from 'path';
 
 import bayes from 'bayes';
+import config from 'config';
 import { clamp } from 'lodash';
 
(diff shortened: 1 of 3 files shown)


server/lib/spam.ts
```diff
@@ -394,8 +395,9 @@ const getSpamDomains = (content: string): string[] => {
     return [];
   }
 
+  const lowerCaseContent = content.toLowerCase();
```
Betree, line 398: Moved this one up to avoid calling `toLowerCase` on each loop

znarf (approved):
Thank you. Looks great!

PR #5202, opened 2021-01-29
fix(paymentReceipts): remove consolidated receipts for current month

flickz:
Resolves https://github.com/opencollective/opencollective/issues/3684

- Removes transactions from current months in consolidated reciepts
- Removes transactions to self. i.e where `transaction.fromCollectiveId !== transaction.collectiveId`

### DIFF
--- a/server/lib/pdf.js
+++ b/server/lib/pdf.js
@@ -36,7 +36,7 @@ export const getTransactionPdf = async (transaction, user) => {
 
 export const getConsolidatedInvoicesData = async fromCollective => {
   const transactions = await models.Transaction.findAll({
-    attributes: ['createdAt', 'HostCollectiveId', 'amountInHostCurrency', 'hostCurrency'],
+    attributes: ['createdAt', 'HostCollectiveId', 'amountInHostCurrency', 'hostCurrency', 'CollectiveId'],
     where: {
       type: 'CREDIT',
       [Op.or]: [
@@ -55,6 +55,15 @@ export const getConsolidatedInvoicesData = async fromCollective => {
     if (!HostCollectiveId) {
       continue;
     }
+
+    if (fromCollective.id === transaction.CollectiveId) {
+      continue;
+    }
+
+    if (moment(transaction.createdAt).isSame(new Date(), 'month')) {
+      continue;
+    }
+
     if (!hostsById[HostCollectiveId]) {
       hostsById[HostCollectiveId] = await models.Collective.findByPk(HostCollectiveId, {
         attributes: ['id', 'slug'],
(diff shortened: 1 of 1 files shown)


server/lib/pdf.js
```diff
@@ -55,6 +55,15 @@ export const getConsolidatedInvoicesData = async fromCollective => {
     if (!HostCollectiveId) {
       continue;
     }
+
+    if (fromCollective.id === transaction.CollectiveId) {
```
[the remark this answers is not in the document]
znarf: @flickz I haven't tested but Ben might be right. Better put that in SQL `CollectiveId != FromCollectiveId`

Not sure how to construct this using Sequelize.

server/lib/pdf.js
```diff
@@ -55,6 +56,11 @@ export const getConsolidatedInvoicesData = async fromCollective => {
     if (!HostCollectiveId) {
       continue;
     }
+
+    if (moment(transaction.createdAt).isSame(new Date(), 'month')) {
+      continue;
+    }
+
```
Betree, line 63: This one also seems to be something we can easily check with SQL: 
`createdAt: { [Op.lt]: moment().startOf('month') }`

server/lib/pdf.js
```diff
@@ -36,9 +36,10 @@ export const getTransactionPdf = async (transaction, user) => {
 
 export const getConsolidatedInvoicesData = async fromCollective => {
   const transactions = await models.Transaction.findAll({
-    attributes: ['createdAt', 'HostCollectiveId', 'amountInHostCurrency', 'hostCurrency'],
+    attributes: ['createdAt', 'HostCollectiveId', 'amountInHostCurrency', 'hostCurrency', 'CollectiveId'],
     where: {
       type: 'CREDIT',
+      CollectiveId: { [Op.not]: fromCollective.id },
```
Betree, line 42: Are we sure that this does not filter out refunds?

flickz: > Are we sure that this does not filter out refunds?

@Betree I'm having trouble testing for individual refund locally, do you have any recommended way I can do that?

flickz: Nvm, I found a way. 🙂

server/lib/pdf.js
```diff
@@ -55,6 +59,11 @@ export const getConsolidatedInvoicesData = async fromCollective => {
     if (!HostCollectiveId) {
       continue;
     }
+
+    // if (moment(transaction.createdAt).isSame(new Date(), 'month')) {
+    //   continue;
+    // }
+
```
Betree, line 66:
```suggestion
```

server/lib/pdf.js
Betree: I think this one is still an issue

flickz: @Betree This does not filter out refund. I tested it.

Betree: All good then, let's try it

PR #5203, opened 2021-01-29
Improve setCurrency

znarf:
- Cascade to transactions, transaction.setCurrency
- GraphQL mutation to do it for admins and root admins
- Fix issue for hosted collectives (will be backported)
- Allow hosts and self hosted to update (if hostCurrency is already good in the ledger)
- Skip Tiers if they have attached orders

### DIFF
--- a/server/graphql/v2/input/AccountUpdateInput.js
+++ b/server/graphql/v2/input/AccountUpdateInput.js
@@ -1,0 +1,14 @@
+import { GraphQLInputObjectType, GraphQLNonNull, GraphQLString } from 'graphql';
+
+import { Currency } from '../enum/Currency';
+
+export const AccountUpdateInput = new GraphQLInputObjectType({
+  name: 'AccountUpdateInput',
+  fields: () => ({
+    id: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'The public id identifying the account (ie: dgm9bnk8-0437xqry-ejpvzeol-jdayw5re)',
+    },
+    currency: { type: Currency },
+  }),
+});
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1835,7 +1835,10 @@ export default function (Sequelize, DataTypes) {
     const isHost = await this.isHost();
     if (isHost) {
       // We only expect currency change at the beginning of the history of the Host
-      const transactionCount = await models.Transaction.count({ where: { HostCollectiveId: this.id } });
+      // We're however good with it if currency is already recorded as hostCurrency in the ledger
+      const transactionCount = await models.Transaction.count({
+        where: { HostCollectiveId: this.id, hostCurrency: { [Op.not]: currency } },
+      });
       if (transactionCount > 0) {
         throw new Error(
           'You cannot change the currency of an Host with transactions. Please contact support@opencollective.com.',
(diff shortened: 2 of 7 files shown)


znarf:
@Betree Interesting given that you wrote this earlier https://github.com/opencollective/opencollective-api/pull/4438

server/models/Tier.js
```diff
@@ -378,6 +379,24 @@ export default function (Sequelize, DataTypes) {
...
+    }
+
+    // Check if Orders
+    const count = await models.Order.count({ where: { TierId: this.id } });
+    if (count > 0) {
+      logger.warning(
+        `Can't change currency of tier ${this.id}: ${this.currency} -> ${currency} (total orders: ${count})`,
+      );
+      return this;
+    }
```
Betree, line 395: This should probably return an error, since it means the function failed to complete

server/models/Transaction.js
```diff
@@ -321,6 +322,34 @@ export default (Sequelize, DataTypes) => {
...
+    // Nothing to do
+    if (currency === this.currency) {
+      return this;
+    }
+
+    const hostCurrencyFxRate = await getFxRate(currency, this.hostCurrency, this.createdAt);
+    // TODO: if Stripe transaction, we should better check if we have the rate stored there
+
+    // REMINDER: currency * hostCurrencyFxRate = amountInHostCurrency
+    // so: currency = amountInHostCurrency / hostCurrencyFxRate
```
Betree, line 335:
```suggestion
    // REMINDER: amount * hostCurrencyFxRate = amountInHostCurrency
    // so: amount = amountInHostCurrency / hostCurrencyFxRate
```

server/graphql/v2/input/AccountUpdateInput.js
```diff
@@ -0,0 +1,18 @@
+import { GraphQLInputObjectType, GraphQLString } from 'graphql';
+
+import { Currency } from '../enum/Currency';
+
+export const AccountUpdateInput = new GraphQLInputObjectType({
+  name: 'AccountUpdateInput',
+  fields: () => ({
+    id: {
+      type: GraphQLString,
```
Betree, line 9: I'm wondering if we shouldn't make this `id` mandatory and the only way to reference the account for this mutation, since it will eventually let you edit the `slug` in the future that would make things clearer - otherwise how to know if the provided `slug` should be used to fetch the account or to update it?

Betree (commented):
The way we compute the values seems good. This is going to be super helpful to implement https://github.com/opencollective/opencollective/issues/3460. 

Since it's a bit sensitive, some unit tests to prevent regressions would be a nice addition.

Betree:
> @Betree Interesting given that you wrote this earlier #4438

I'm closing it since the mutation is going to be implemented in this PR. We'll add the other fields when needed.

server/models/Collective.js
```diff
@@ -1835,30 +1835,42 @@ export default function (Sequelize, DataTypes) {
...
-      await Promise.map(collectives, collective => collective.setCurrency(currency), { concurrency: 3 });
+      if (collectives.length > 0) {
+        await Promise.map(collectives, collective => collective.setCurrency(currency), { concurrency: 3 });
+      }
+    }
+
+    // Update all transactions
+    const transactions = await models.Transaction.findAll({ where: { CollectiveId: this.id } });
+    if (transactions.length > 0) {
+      await Promise.map(transactions, transaction => transaction.setCurrency(currency), { concurrency: 3 });
```
kewitz, line 1858: This query will only return one transaction of the pair, we should make sure to update both, either here or in `transaction.setCurrency()`.

server/models/Transaction.js
```diff
@@ -321,6 +322,34 @@ export default (Sequelize, DataTypes) => {
...
+    const amount = Math.round(this.amountInHostCurrency / hostCurrencyFxRate);
+    const netAmountInCollectiveCurrency = Math.round(
+      (this.amountInHostCurrency +
+        this.platformFeeInHostCurrency +
+        this.hostFeeInHostCurrency +
+        this.paymentProcessorFeeInHostCurrency) /
+        hostCurrencyFxRate,
```
znarf, line 342: @Betree how should we handle taxAmount there?

In which currency is it stored? And should it be removed from netAmountInCollectiveCurrency?

Betree: `taxAmount` is always expressed with the same currency than `amount` (`transaction.currency`). We probably want to remove it from `netAmountInCollectiveCurrency`, I just checked and that's what we do in https://github.com/opencollective/opencollective-api/blob/master/server/models/Transaction.js#L624

server/models/Collective.js
znarf: No, the goal is to have a consistent currency regarding the CollectiveId. We should not necessarily update the other one because it has a different CollectiveId and maybe a different desired currency.

server/models/Transaction.js
znarf: Are you sure it's not in hostCurrency? Because, the code at https://github.com/opencollective/opencollective-api/blob/master/server/models/Transaction.js#L624 assume it is.

Betree: Affirmative. The way this amount is calculated in the contribution flow then in `createOrder` is `order.amount * taxPercent`. The code in https://github.com/opencollective/opencollective-api/blob/master/server/models/Transaction.js#L624 is therefor wrong to assume it's expressed in host currency, I missed that when I reviewed it.

I just checked in the DB and the ledger should be fine, we didn't face the case of a collective having a different currency than the host with VAT yet. That however surfaced 4 platform tips transactions that show platform tips and tax amounts don't interact well with each others. I'll create an issue for this, mentioning that we should update https://github.com/opencollective/opencollective-api/blob/master/server/models/Transaction.js#L624 too.

```sql
SELECT *
FROM "Transactions" t 
WHERE "taxAmount" < 0
AND "deletedAt" IS NULL
AND currency != 'EUR'
```

znarf: @Betree thank you, I will then take this into account in my refactoring.

server/models/Collective.js
kewitz: I don't think allowing a transaction pair to have a different currency is a good idea. Shouldn't the Credit/Debit pair always be settled in the same currency?

znarf: Within the current design, I think we have to, because we don't want to allow the "oppositeTransaction" to be in a different currency than the collective profile. The risk is to make balance computation impossible. See current versions of `getBalances` which is not handling multi-currency.

Happy to discuss alternatives, this is the right moment.

znarf:
I'm happy with the current state and I'm gonna merge. Following up in:

- https://github.com/opencollective/opencollective-api/pull/5233
- https://github.com/opencollective/opencollective-api/pull/5214
- Discussion about currency expectations in the Ledger. And upcoming issues based on the discussion.

PR #5206, opened 2021-01-29
Fix Expense varchar (255) size error

kewitz:
Both `Expense.privateMessage` and `ExpenseItem.description` accept HTML content.
Our tables are apparently correct in the case of `Expense.privateMessage` but the Sequelize model is not.
Either way, I'm updating both models and writing a migration to make it consistent.

### DIFF
--- a/migrations/20210129133000-switch-expense-text-types.js
+++ b/migrations/20210129133000-switch-expense-text-types.js
@@ -1,0 +1,27 @@
+'use strict';
+
+module.exports = {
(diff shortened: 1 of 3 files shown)


Betree (approved):
Nice change. HTML fields should definitely have `TEXT` type in all cases since it's difficult to properly estimate their length (because of the HTML tags)

znarf (commented):
Good stuff.

PR #5207, opened 2021-01-29
Consider transaction currency when calculating total money managed

kewitz:


server/lib/queries.js
```diff
@@ -1049,6 +1049,24 @@ const getBalances = async (collectiveIds, until = new Date()) =>
...
+const getBalancesInHostCurrency = async (collectiveIds, until = new Date()) =>
+  sequelize.query(
+    `
+      SELECT
+        t."CollectiveId",
+        COALESCE(SUM(ROUND(t."netAmountInCollectiveCurrency" * t."hostCurrencyFxRate")), 0) AS "balance"
+      FROM
+        "Transactions" t
+      WHERE
+        t."CollectiveId" IN (:ids)
```
znarf, line 1061: While we're at it, we should also join on `HostCollectiveId` to make sure we're only counting Transactions related to the host.

server/lib/queries.js
```diff
@@ -1049,6 +1049,24 @@ const getBalances = async (collectiveIds, until = new Date()) =>
     { type: sequelize.QueryTypes.SELECT, replacements: { ids: collectiveIds, until } },
   );
 
+const getBalancesInHostCurrency = async (collectiveIds, until = new Date()) =>
+  sequelize.query(
+    `
+      SELECT
+        t."CollectiveId",
+        COALESCE(SUM(ROUND(t."netAmountInCollectiveCurrency" * t."hostCurrencyFxRate")), 0) AS "balance"
```
Betree, line 1057: I just checked and we have some transactions with a null `hostCurrencyFxRate`:

```sql
SELECT * FROM "Transactions" t WHERE "hostCurrencyFxRate" IS NULL AND "deletedAt" IS NULL
```

We probably want to fix these entries and add a non-null constraint on `hostCurrencyFxRate`, but if we're not ready to do it now we could do:
```suggestion
        COALESCE(SUM(ROUND(t."netAmountInCollectiveCurrency" * COALESCE(t."hostCurrencyFxRate", 1))), 0) AS "balance"
```

Betree (commented):
Looks good

server/lib/queries.js
znarf: Can we run a migration and default hostCurrencyFxRate to 1?

server/lib/queries.js
kewitz: Awesome!

server/lib/queries.js
kewitz: I fixed the recent existing records manually, they were created in the invoice worker.
Also, I added set default value of hostCurrencyFxRate to 1.

PR #5212, opened 2021-02-01
ApplyToHost: Fix mutation for applying as an org

Betree:
Fix https://sentry.io/organizations/open-collective/issues/2191467101/
Regression introduced in https://github.com/opencollective/opencollective-api/pull/5196

### DIFF
--- a/server/graphql/v2/mutation/HostApplicationMutations.ts
+++ b/server/graphql/v2/mutation/HostApplicationMutations.ts
@@ -55,17 +55,18 @@ const HostApplicationMutations = {
       const collective = await fetchAccountWithReference(args.collective);
       if (!collective) {
         throw new NotFound('Collective not found');
-      }
-      if (![CollectiveType.COLLECTIVE, CollectiveType.FUND].includes(collective.type)) {
-        throw new Error('Account must be a collective or a fund');
-      }
-      if (!req.remoteUser.isAdminOfCollective(collective)) {
+      } else if (!req.remoteUser.isAdminOfCollective(collective)) {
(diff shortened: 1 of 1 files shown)


server/graphql/v2/mutation/HostApplicationMutations.ts
```diff
@@ -68,6 +64,11 @@ const HostApplicationMutations = {
         throw new NotFound('Host not found');
       }
 
+      const isSelfHosted = collective.type === CollectiveType.ORGANIZATION && collective.id === host.id;
```
znarf, line 67: `collective.type === CollectiveType.COLLECTIVE` ? Otherwise, I don't get it.

znarf (changes requested):
Sorry, I don't get the case where someone applies as an Org.

Betree:
@znarf
- Create an org
- Go to http://localhost:3000/myorg/accept-financial-contributions
- Click on "Connect stripe"
- Pick "Ourselves"
- See `applyToHost` mutation triggered with `applyToHost({ collective: organization, host: organization })`
 
Code: https://github.com/opencollective/opencollective-frontend/blob/11474acef561e6b2908a88006f60ec1d8d50b296/components/accept-financial-contributions/AcceptContributionsOurselvesOrOrg.js#L185

znarf:
It should not be possible to land on this path. We should not link at any time to http://localhost:3000/myorg/accept-financial-contributions and if necessary we should block that page for Organizations.

There is no such thing as Hosting an Organization and that would be confusing to allow it.

Is there a way to reproduce today without hacking the URL?

PR #5222, opened 2021-02-02
Prevent SCHEDULE_FOR_PAYMENT expenses from being edited

kewitz:


Betree (approved):
Nice :+1: Thanks for making that fix. Good job with the tests

PR #5230, opened 2021-02-03
Remove legacy expense code

Betree:
Taking out the code related to expenses deprecation from https://github.com/opencollective/opencollective-api/pull/5204 as it's a big piece on its own. This PR removes deprecated expense queries and mutations from GQLV1.

Also moved E2E tests to `GQLV2`'s `createExpense` in https://github.com/opencollective/opencollective-frontend/pull/5832.


test/utils.js
```diff
@@ -142,7 +142,7 @@ export const graphqlQuery = async (query, variables, remoteUser, schema = schema
  * @param {object} variables - Variables to use in the queries and mutations. Example: { id: 1 }
  * @param {object} remoteUser - The user to add to the context. It is not required.
  */
-export async function graphqlQueryV2(query, variables, remoteUser) {
+export async function graphqlQueryV2(query, variables, remoteUser = null) {
```
sbinlondon, line 145: Why add a default `null` for this argument?

sbinlondon (approved):
Amazing work! Always good to clean up code. I like that you've done the new mutations in Typescript so there's info about the arguments and returns.

test/utils.js
Betree: In some places we call this function with only two arguments `graphqlQueryV2(query, variables)`. It's a way making the query without logging in so it's valid, but Typescript would complain that this function expects a third arguments. This `= null` is there to tell Typescript that this third argument is optional.

PR #5235, opened 2021-02-03
GQLV2/Transactions: Include gift card transactions

Betree:


znarf:
Seems CI is broken by what I'm attempting to fix in https://github.com/opencollective/opencollective-api/pull/5232

PR #5237, opened 2021-02-03
Invalidate cache when removing team members

sbinlondon:
Fixes https://github.com/opencollective/opencollective/issues/3932

According to [`server/graphql/common/members.js`](https://github.com/opencollective/opencollective-api/blob/5412649b0f6a5aef4d79979f8a54fdad804e1908/server/graphql/common/members.js#L34), if you call `models.Member.update` and don't use it directly, i.e. using `models.Member.update(..., { where: ...})`, then the `afterUpdate` hook isn't triggered and the collective page cache isn't automatically cleared. 

In `Collective.prototype.editMembers`, because we [update the collective's members indirectly](https://github.com/opencollective/opencollective-api/blob/5412649b0f6a5aef4d79979f8a54fdad804e1908/server/models/Collective.js#L2208), this means that removed team members might persist for a bit on the collective page.

This PR adds the `invalidateContributorsCache` to `Collective.prototype.editMembers` to fix this (tested before and after and it works).

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -47,6 +47,7 @@ import {
   isCollectiveSlugReserved,
   validateSettings,
 } from '../lib/collectivelib';
+import { invalidateContributorsCache } from '../lib/contributors';
 import { getFxRate } from '../lib/currency';
 import emailLib from '../lib/email';
 import logger from '../lib/logger';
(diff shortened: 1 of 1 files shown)


Betree (approved):
Great investigation, nice job with explaining the root cause in your PR description - that's super clear. All good with the fix!

PR #5246, opened 2021-02-04
feat(ocf-application): allow existing collective admin to apply to OCF host

flickz:
Following up https://github.com/opencollective/opencollective-frontend/pull/5833 on the recent [update](https://github.com/opencollective/opencollective/issues/3683#issuecomment-769566325) to OCF onboarding flow, this PR allows authenticated admin of existing collective to apply to Open collective foundation.

### DIFF
--- a/server/graphql/v2/mutation/HostApplicationMutations.ts
+++ b/server/graphql/v2/mutation/HostApplicationMutations.ts
@@ -1,4 +1,5 @@
 import { GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
+import GraphQLJSON from 'graphql-type-json';
 
 import { activities } from '../../../constants';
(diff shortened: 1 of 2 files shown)


znarf:
Sorry, but I dislike the approach.

The `createCollective` mutation is about creating a collective and not applying with an existing collective.

We should apply with a dedicated mutation. @Betree do you have a suggestion how best to handle this?

Betree:
> Sorry, but I dislike the approach.
> 
> The `createCollective` mutation is about creating a collective and not applying with an existing collective.
> 
> We should apply with a dedicated mutation. @Betree do you have a suggestion how best to handle this?

I agree. We have an `applyToHost` mutation on V2 that should do the job just fine, we just need to add the `applicationData` there

znarf (approved):
Looks good.

Betree (approved):
:+1:

PR #5252, opened 2021-02-05
Rename VirtualCards to GiftCards

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3820

Require:
- https://github.com/opencollective/opencollective-rest/pull/227
- https://github.com/opencollective/opencollective-frontend/pull/5861

Given that https://github.com/opencollective/opencollective-frontend/pull/5861 gets deployed first, all the changes from this PR should be retro-compatible.


server/graphql/v1/mutations.js
```diff
@@ -505,37 +505,6 @@ const mutations = {
       return applicationMutations.deleteApplication(_, args, req);
     },
   },
-  createPaymentMethod: {
```
Betree, line 508: See opencollective/opencollective-rest#227

PR #5339, opened 2021-02-08
Refact Balance Queries

kewitz:
Resolves https://github.com/opencollective/opencollective-security/issues/69


kewitz:
Solved by https://github.com/opencollective/opencollective-api/pull/5439

PR #5346, opened 2021-02-09
Keep ExpenseId in its refund transactions

kewitz:
For debugging reasons.


Betree (commented):
Should we update existing transactions for data consistency? (or create an issue to do it later)

znarf (approved):

znarf:
Would be a nice touch to migrate past issues or to document the problem in an issue at least.

kewitz:
@Betree @znarf I'll add a migration to this PR.

Betree (approved):
The migration looks good!

PR #5354, opened 2021-02-11
Allow use of recovery codes during login if 2FA is unavailable

sbinlondon:
Relates https://github.com/opencollective/opencollective/issues/3709

### DIFF
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -13,7 +13,8 @@ import {
 import { isValidEmail } from '../lib/utils';
 import models from '../models';
 
-const { Unauthorized, ValidationFailed } = errors;
+const { Unauthorized, ValidationFailed, TooManyRequests } = errors;
+
 const { User } = models;
 
 /**
@@ -111,7 +112,7 @@ export const twoFactorAuthAndUpdateToken = async (req, res, next) => {
   };
 
   if (await rateLimit.hasReachedLimit()) {
-    return next(new Unauthorized('Too many attempts. Please try again in an hour'));
+    return next(new TooManyRequests('Too many attempts. Please try again in an hour'));
   }
 
   const user = await User.findByPk(userId);
@@ -125,19 +126,19 @@ export const twoFactorAuthAndUpdateToken = async (req, res, next) => {
     // if there is a 2FA code, we need to verify it before returning the token
     const verified = verifyTwoFactorAuthenticatorCode(user.twoFactorAuthToken, twoFactorAuthenticatorCode);
     if (!verified) {
-      return await fail(new Unauthorized('Two-factor authentication code failed. Please try again'));
+      return fail(new Unauthorized('Two-factor authentication code failed. Please try again'));
     }
   } else if (twoFactorAuthenticationRecoveryCode) {
     // or if there is a recovery code try to verify it
     if (typeof twoFactorAuthenticationRecoveryCode !== 'string') {
-      return await fail(new ValidationFailed('2FA recovery code must be a string'));
+      return fail(new ValidationFailed('2FA recovery code must be a string'));
     }
     const verified = verifyTwoFactorAuthenticationRecoveryCode(
       user.twoFactorAuthRecoveryCodes,
       twoFactorAuthenticationRecoveryCode,
     );
     if (!verified) {
-      return await fail(new Unauthorized('Two-factor authentication recovery code failed. Please try again'));
+      return fail(new Unauthorized('Two-factor authentication recovery code failed. Please try again'));
     }
     const remainingRecoveryCodes = user.twoFactorAuthRecoveryCodes.filter(
       code => crypto.hash(twoFactorAuthenticationRecoveryCode.toUpperCase()) !== code,
--- a/server/controllers/users.js
+++ b/server/controllers/users.js
@@ -5,7 +5,10 @@ import emailLib from '../lib/email';
 import errors from '../lib/errors';
 import logger from '../lib/logger';
 import RateLimit, { ONE_HOUR_IN_SECONDS } from '../lib/rate-limit';
-import { verifyTwoFactorAuthenticatorCode } from '../lib/two-factor-authentication';
+import {
+  verifyTwoFactorAuthenticationRecoveryCode,
+  verifyTwoFactorAuthenticatorCode,
+} from '../lib/two-factor-authentication';
 import { isValidEmail } from '../lib/utils';
 import models from '../models';
 
@@ -92,11 +95,10 @@ export const updateToken = async (req, res) => {
 };
 
 /**
- * Verify the 2FA code the user has entered when
- * logging in and send back another JWT.
+ * Verify the 2FA code or recovery code the user has entered when logging in and send back another JWT.
  */
 export const twoFactorAuthAndUpdateToken = async (req, res, next) => {
-  const { twoFactorAuthenticatorCode } = req.body;
+  const { twoFactorAuthenticatorCode, recoveryCode } = req.body;
 
   const userId = Number(req.jwtPayload.sub);
   const user = await User.findByPk(userId);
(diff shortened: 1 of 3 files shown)


server/lib/two-factor-authentication.js
```diff
@@ -22,6 +22,20 @@ export function verifyTwoFactorAuthenticatorCode(encryptedTwoFactorAuthToken, tw
...
 }
 
+/** Verifies a submitted recovery code against the hashed ones saved for them in the DB.
+ * If the recovery code is valid, it returns the remaining recovery codes to update the user with.
+ * If the recovery code is not valid, it returns null so we can throw an error.
+ */
+export function verifyTwoFactorAuthenticationRecoveryCode(recoveryCode, hashedRecoveryCodes) {
+  const verified = hashedRecoveryCodes.includes(crypto.hash(recoveryCode));
+  if (verified) {
+    return hashedRecoveryCodes.filter(code => crypto.hash(recoveryCode) !== code);
```
znarf, line 32: It's not the job of this function to filter the codes, this should be moved elsewhere.

It should return `true`/`false`. Or alternatively, do nothing/throw an exception.

server/controllers/users.js
```diff
@@ -106,11 +108,25 @@ export const twoFactorAuthAndUpdateToken = async (req, res, next) => {
...
+  if (twoFactorAuthenticatorCode) {
+    // if there is a 2FA code, we need to verify it before returning the token
+    const verified = verifyTwoFactorAuthenticatorCode(user.twoFactorAuthToken, twoFactorAuthenticatorCode);
+    if (!verified) {
+      return next(new Unauthorized('Two-factor authentication code failed. Please try again'));
+    }
+  } else if (recoveryCode) {
+    // or if there is a recovery code try to verify it
+    const { twoFactorAuthRecoveryCodes } = user;
+    const validatedRemainingRecoveryCodes = verifyTwoFactorAuthenticationRecoveryCode(
```
znarf, line 120: This function call is similar to `verifyTwoFactorAuthenticatorCode`, you should then be consistent in what are the parameters and their respective order (1. stored code, 2. user submitted code)

server/controllers/users.js
```diff
@@ -106,11 +108,25 @@ export const twoFactorAuthAndUpdateToken = async (req, res, next) => {
...
-    return next(new Unauthorized('Two-factor authentication code failed. Please try again'));
+  if (twoFactorAuthenticatorCode) {
+    // if there is a 2FA code, we need to verify it before returning the token
+    const verified = verifyTwoFactorAuthenticatorCode(user.twoFactorAuthToken, twoFactorAuthenticatorCode);
+    if (!verified) {
+      return next(new Unauthorized('Two-factor authentication code failed. Please try again'));
+    }
+  } else if (recoveryCode) {
+    // or if there is a recovery code try to verify it
+    const { twoFactorAuthRecoveryCodes } = user;
```
znarf, line 119: In `verifyTwoFactorAuthenticatorCode`, you're not extracting the value from the user object. You should stay consistent in your style and use `user.twoFactorAuthRecoveryCodes` directly.

server/controllers/users.js
```diff
@@ -92,11 +95,10 @@ export const updateToken = async (req, res) => {
 };
 
 /**
- * Verify the 2FA code the user has entered when
- * logging in and send back another JWT.
+ * Verify the 2FA code or recovery code the user has entered when logging in and send back another JWT.
  */
 export const twoFactorAuthAndUpdateToken = async (req, res, next) => {
-  const { twoFactorAuthenticatorCode } = req.body;
+  const { twoFactorAuthenticatorCode, recoveryCode } = req.body;
```
znarf, line 101: There is one parameter prefixed with `twoFactorAuthenticator` and the other not. You should try to be consistent in your variable naming.

znarf (changes requested):
Functionality looks good, just some relatively minor code style issues.

server/lib/two-factor-authentication.js
```diff
@@ -22,6 +22,20 @@ export function verifyTwoFactorAuthenticatorCode(encryptedTwoFactorAuthToken, tw
   return verified;
 }
 
+/** Verifies a submitted recovery code against the hashed ones saved for them in the DB.
+ * If the recovery code is valid, it returns the remaining recovery codes to update the user with.
+ * If the recovery code is not valid, it returns null so we can throw an error.
+ */
+export function verifyTwoFactorAuthenticationRecoveryCode(recoveryCode, hashedRecoveryCodes) {
+  const verified = hashedRecoveryCodes.includes(crypto.hash(recoveryCode));
```
Betree, line 30: This will crash if `user.twoFactorAuthRecoveryCodes` is not defined. We should handle this with a proper error message.

Betree (commented):
I think it's important to rate limit this kind of feature from day 1, because of bruteforce attacks but also because endpoints using crypto functions are good targets for CPU exhaustion in DDOS attacks. 

The helper in `server/lib/rate-limit.ts` can provide a simple way to achieve that, see https://github.com/opencollective/opencollective-api/blob/b133f81132b6a65603aca8ff187a8db33677c680/server/graphql/v2/mutation/ExpenseMutations.ts#L298-L301 for an example. 10 tries per hour sound like a good limit to me.

server/lib/two-factor-authentication.js
sbinlondon: I put it here because I was trying to keep stuff relating to `crypto` in one file (this one), and returned the recovery codes to the other file so that I didn't have to import the `User` model multiple places. There was actually some thought behind the way I did this and I did in fact consider the way you are suggesting now (but went with this way for the mentioned reasons). Happy to change it but it's not from lack of consideration about the issue.

znarf: Good to understand the reasons behind this decision but I'm standing by my comment, `verifyTwoFactorAuthenticationRecoveryCode` should only verify.

Maybe you can find another design, for example exposing a function that only compare 1 single recovery code against user input, and directly use this function in the filter.

server/lib/two-factor-authentication.js
sbinlondon: Have fixed this by checking in `server/controllers/users.js` that the recovery code is a string (since due to PR changes we now also use `crypto.hash` in there too).

server/controllers/users.js
```diff
@@ -92,25 +96,50 @@ export const updateToken = async (req, res) => {
...
+  const { twoFactorAuthenticatorCode, twoFactorAuthenticationRecoveryCode } = req.body;
 
   const userId = Number(req.jwtPayload.sub);
+
+  // Both 2FA and recovery codes rate limited to 10 tries per hour
+  const rateLimit = new RateLimit(`user_2FA_endpoint_${userId}`, 10, ONE_HOUR_IN_SECONDS);
+  if (!(await rateLimit.registerCall())) {
+    return next(new Unauthorized('Too many attempts. Please try again in an hour'));
+  }
+
```
Betree, line 111: :ice_cream: Since admins can be forced to update their tokens multiple time when they pay expenses, doing that here might be a too strong limitation. I would only count in the rate limit. One simple way to achieve that could be to have something like this:

```suggestion
  // Both 2FA and recovery codes rate limited to 10 tries per hour
  const rateLimit = new RateLimit(`user_2FA_endpoint_${userId}`, 10, ONE_HOUR_IN_SECONDS);
  const fail = (exception) => {
    await rateLimit.registerCall();
    next(exception);
  }

  if (await rateLimit.hasReachedLimit()) {
    return next(new Unauthorized('Too many attempts. Please try again in an hour'));
  }
```

Then replace all the `return next(new Exception(...))` by `return fail(new Exception(...))`. For example:
```es6
return fail(new Unauthorized('Two-factor authentication code failed. Please try again'));
```

server/controllers/users.js
```diff
@@ -92,25 +96,55 @@ export const updateToken = async (req, res) => {
...
+  // Both 2FA and recovery codes rate limited to 10 tries per hour
+  const rateLimit = new RateLimit(`user_2FA_endpoint_${userId}`, 10, ONE_HOUR_IN_SECONDS);
+  const fail = async exception => {
+    await rateLimit.registerCall();
+    next(exception);
+  };
+
+  if (await rateLimit.hasReachedLimit()) {
+    return next(new Unauthorized('Too many attempts. Please try again in an hour'));
```
Betree, line 114: Seems more appropriate

```suggestion
    return next(new RateLimitExceeded('Too many attempts. Please try again in an hour'));
```

server/controllers/users.js
```diff
@@ -92,25 +96,55 @@ export const updateToken = async (req, res) => {
...
 
-  // we need to verify the 2FA code before returning the token
-  const verified = verifyTwoFactorAuthenticatorCode(user.twoFactorAuthToken, twoFactorAuthenticatorCode);
-  if (!verified) {
-    return next(new Unauthorized('Two-factor authentication code failed. Please try again'));
+  if (twoFactorAuthenticatorCode) {
+    // if there is a 2FA code, we need to verify it before returning the token
+    const verified = verifyTwoFactorAuthenticatorCode(user.twoFactorAuthToken, twoFactorAuthenticatorCode);
+    if (!verified) {
+      return await fail(new Unauthorized('Two-factor authentication code failed. Please try again'));
```
Betree, line 128: :ice_cream: I don't think these `await` in the return are necessary

Betree (commented):
Looks good to me. I wonder what is the expected behavior when there's no recovery code left? That's not something we need to resolve now, it's an edge case.

Betree (approved):

PR #5370, opened 2021-02-15
Add total net amount received

znarf:


server/models/Collective.js
```diff
@@ -2513,6 +2513,24 @@ export default function (Sequelize, DataTypes) {
...
+    };
+    if (startDate) {
+      where.createdAt[Op.gte] = startDate;
+    }
+    return models.Transaction.findOne({
+      attributes: [
+        [Sequelize.fn('COALESCE', Sequelize.fn('SUM', Sequelize.col('netAmountInCollectiveCurrency')), 0), 'total'],
+      ],
+      where,
+    }).then(result => Promise.resolve(parseInt(result.toJSON().total, 10)));
```
Betree, line 2531: It's not new code but I'm surprised by this `.toJSON()`, it doesn't seem efficient. `Promise.resolve` also does not seem necessary.

Betree (approved):

server/models/Collective.js
znarf: Agree. Will keep it like this for now.

PR #5371, opened 2021-02-15
Add orderBy to account.members

znarf:
Similar to what we have in `memberOf`.

Was asked on support by an API user.


Betree (approved):
No blocker, it's good to have an `orderBy` to make sure results are always returned in the same order.

PR #5372, opened 2021-02-15
Exclude refunded transactions from totalAmountReceived

znarf:
- remove useless Promise.resolve
- use dataValues instead of toJSON


Betree (approved):
Nice, this should cover everything

PR #5397, opened 2021-02-22
Various enhancements for pending bank transfers

Betree:
Fix https://github.com/opencollective/opencollective/issues/3916

- Mark unpaid crash when already refunded in the past
- Bold reference for pending orders
- Add info about the processing time for the "expense paid" email

image

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -1005,7 +1005,7 @@ export async function markExpenseAsUnpaid(req, ExpenseId, processorFeeRefunded):
     }
 
     const transaction = await models.Transaction.findOne({
-      where: { ExpenseId },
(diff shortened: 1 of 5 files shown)


server/graphql/common/expenses.ts
```diff
@@ -1005,7 +1005,7 @@ export async function markExpenseAsUnpaid(req, ExpenseId, processorFeeRefunded):
     }
 
     const transaction = await models.Transaction.findOne({
-      where: { ExpenseId },
+      where: { ExpenseId, RefundTransactionId: null },
```
Betree, line 1008: Made this quick fix while I was there, steps were:
- Pay expense
- Mark it as unpaid
- Pay again
- Mark it as unpaid
- ==> See crash, "Transaction already refunded"

kewitz (approved):

PR #5402, opened 2021-02-22
Add Platform Tips Support for Funds Modal

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/3652


server/graphql/v1/mutations/orders.js
```diff
@@ -946,6 +947,12 @@ export async function addFundsToCollective(order, remoteUser) {
     orderData.data.platformFeePercent = order.platformFeePercent;
   }
 
+  if (!isNil(order.platformTipPercent)) {
```
Betree, line 950: We don't do it for `hostFeePercent` and `platformFeePercent`, but it would actually be nice to make sure this number is between 0 and 100 (I'm not sure if we verify that later in the process).

server/paymentProviders/opencollective/host.js
```diff
@@ -57,6 +65,9 @@ paymentMethodProvider.processOrder = async order => {
     platformFeeInHostCurrency,
     paymentProcessorFeeInHostCurrency: 0,
     description: order.description,
+    data: {
+      isFeesOnTop: order.data?.isFeesOnTop,
```
znarf, line 69: Should ideally be a boolean, instead of being loosely false (null/undefined).

server/paymentProviders/opencollective/host.js
```diff
@@ -24,7 +26,13 @@ paymentMethodProvider.processOrder = async order => {
...
 
   const hostFeePercent = await getHostFeePercent(order);
-  const platformFeePercent = await getPlatformFeePercent(order);
+
+  let platformFeePercent;
+  if (order.data.platformTipPercent) {
+    platformFeePercent = order.data.platformTipPercent;
+  } else {
+    platformFeePercent = await getPlatformFeePercent(order);
+  }
```
znarf, line 35: The goal of the `getPlatformFeePercent` utility is to centralize all the complexity. 

I think we should move `data.platformTipPercent` handling is the method itself.

server/graphql/v1/mutations/orders.js
SudharakaP: @Betree : Thanks. 👍🏽 . Done, I've added the validations for `hostFeePercent` and `platformFeePercent` as well. 😄

server/paymentProviders/opencollective/host.js
SudharakaP: Done. 👍🏽

server/paymentProviders/opencollective/host.js
SudharakaP: Done. 👍🏽

SudharakaP:
@Betree @znarf : Changes have been made based on the reviews. I'll work on adding some tests to this today as discussed. Let me know if you see anything else. 😄

SudharakaP:
> @Betree @znarf : Changes have been made based on the reviews. I'll work on adding some tests to this today as discussed. Let me know if you see anything else. smile

I've created some tests for the `AddFundsMutation`. Please take a look and let me know if there's anything missing or if you see any other issues. 😄

server/graphql/v2/mutation/AddFundsMutations.ts
```diff
@@ -16,6 +16,7 @@ export const addFundsMutation = {
     description: { type: new GraphQLNonNull(GraphQLString) },
     hostFeePercent: { type: new GraphQLNonNull(GraphQLInt) },
     platformFeePercent: { type: GraphQLInt, description: 'Can only be set if root' },
+    platformTipPercent: { type: GraphQLInt },
```
kewitz, line 19: I don't like this because it limits the tip to increments of 1% of the total value.
This also increases the complexity of the logic down below.
My suggestion is to always pass the `platformTip` as an absolute value in cents.

server/graphql/v1/mutations/orders.js
```diff
@@ -947,6 +948,12 @@ export async function addFundsToCollective(order, remoteUser) {
...
+    orderData.data.isFeesOnTop = true;
+    orderData.data.platformFee = calcFee(order.totalAmount, order.platformTipPercent);
+    orderData.data.platformTipPercent = order.platformTipPercent;
```
kewitz, line 954: Can be simplified if we pass the absolute `platformFee` amount.

server/graphql/v1/mutations/collectives.js
```diff
@@ -196,6 +196,7 @@ export async function createCollectiveFromGithub(_, args, req) {
     const host = await models.Collective.findByPk(defaultHostCollective('opensource').CollectiveId);
     const promises = [
       collective.addUserWithRole(user, roles.ADMIN),
+      host.addUserWithRole(user, roles.ADMIN),
```
kewitz, line 199: :red_circle: this is not safe and can't be merged. Set up the E2E test to bypass the limitation.

kewitz (changes requested):

server/graphql/v2/mutation/AddFundsMutations.ts
SudharakaP: @kewitz : I was wondering an alternative which is to make the `platformTipPercent` a float. The reason is that even if we remove the below logic, the `calcFee` will still need to be done within the (`getPlatformFeePercent`)[https://github.com/opencollective/opencollective-api/blob/059780c1a791adfb16380d3f9c7ce4d2e752fac1/server/lib/payments.js#L589] method. WDYT? 🤔

kewitz: You can just ignore the `getPlatformFeePercent` if you have the `platformTip`.
Why? Well, despite their name, basically they're too different things. Also, due to our business logic, we're deprecating platform fees in favor of tips and it is currently not possible to collect both on the same transaction.

SudharakaP: @kewitz : Thanks that makes sense. 👍🏽

server/graphql/v1/mutations/collectives.js
SudharakaP: @kewitz : Maybe I am missing something here, but my thinking is that the preceding `if` block; `if (config.env !== 'production' && user.email.match(/.*test.*@opencollective.com$/))` makes sure this only executes for E2E tests. 

The problem I am trying to solve here is that our E2E tests doesn't support logging in as a host admin (only just a collective admin). Maybe thinking further I think we could add a condition such that when a user passes a test email of the form `testuser+host@opencollective.com` where we can check for the `host` keyword within the test email here and only execute this condition for that particular case. Something like, 

```
if (user.email.match(<host-keyword-match-regex>)){
  host.addUserWithRole(user, roles.ADMIN),
}
```
WDYT? 🤔

SudharakaP: As discussed I've added a condition to support emails such as `testuser+host@opencollective.com` here. We probably need to update the [documentation](https://docs.opencollective.com/help/contributing/development/testing-with-cypress) as well, I'll do that in a separate PR. 😄

SudharakaP:
@kewitz : I've resolved all the reviews you pointed out; also note that I've added this host user creation instruction at https://github.com/opencollective/documentation/pull/103. Let me know if you have any further suggestions. 😄

server/graphql/v2/mutation/AddFundsMutations.ts
```diff
@@ -26,6 +27,19 @@ export const addFundsMutation = {
...
+    } else if (!(args.platformFeePercent >= 0 && args.platformFeePercent <= 100)) {
+      percentValidationFailPoint = 'platformFeePercent';
+    }
+
+    if (percentValidationFailPoint) {
+      throw new ValidationFailed(
+        `${percentValidationFailPoint} is a percentage and therefore should be between 0 and 100.`,
+      );
+    }
+
```
kewitz, line 42: For readability purposes, I suggest you to avoid negated conditionals and use two simple throws:
```suggestion
    if (args.hostFeePercent < 0 || args.hostFeePercent > 100) {
      throw new ValidationFailed('hostFeePercent should be a value between 0 and 100.');
    } else if (args.platformFeePercent < 0 || args.platformFeePercent > 100)) {
      throw new ValidationFailed('platformFeePercent should be a value between 0 and 100.');
    }

```

server/models/Transaction.js
```diff
@@ -435,6 +435,20 @@ export default (Sequelize, DataTypes) => {
     return exportToCSV(transactions, attributes, getColumnName, processValue);
   };
 
+  Transaction.createPlatformTipTransaction = async transaction => {
```
kewitz, line 438: I think we should keep logic within the `processOrder` function.
My reasoning here is that this special condition only applies to funds for now and this function can cause some confusion with the `feesOnTop` one.

server/graphql/v2/mutation/AddFundsMutations.ts
SudharakaP: Thanks. Done. 👍🏽

server/models/Transaction.js
SudharakaP: @kewitz : Thanks. Done, let me know if you see any further issues. 😄

kewitz (approved):

SudharakaP:
Probably needs one more review as this is changed to support the [new design](https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/OC.com-Production-Ready-%E2%9C%85?node-id=10518%3A29405); now I am sending two requests. The first one for the addition of the fund and the second one to add the platform tips. 😄

server/paymentProviders/opencollective/host.js
```diff
@@ -16,6 +16,20 @@ paymentMethodProvider.getBalance = () => {
...
 
+paymentMethodProvider.createPlatformTipTransaction = async transaction => {
+  const platformTipTransaction = {
+    ...transaction,
+    amount: transaction.platformTipInHostCurrency,
+    description: 'Financial contribution (Platform Tip) to Open Collective',
+    netAmountInCollectiveCurrency: transaction.platformTipInHostCurrency,
+    FromCollectiveId: transaction.HostCollectiveId,
+    CollectiveId: 8686, // Open Collective (Organization)
+    platformTipInHostCurrency: undefined,
```
znarf, line 27: Was this checked with @kewitz. Will this be taken in consideration in the settlement invoice?

znarf: If this is similar to a Bank Transfer then it should be good.

znarf (commented):
Looks good to me. A small simplification would be great as we don't have to support amount=0.

Let's make sure with @kewitz that it's properly accounted in the settlement invoice for the Host.

server/paymentProviders/opencollective/host.js
SudharakaP: @kewitz : I think you reviewed this before; but if you don't mind can you do a review again as we changed the code a bit (mostly on the front-end) with the new design updates. 😄

server/paymentProviders/opencollective/host.js
```diff
@@ -16,6 +16,20 @@ paymentMethodProvider.getBalance = () => {
   return Promise.resolve(maxInteger);
 };
 
+paymentMethodProvider.createPlatformTipTransaction = async transaction => {
```
kewitz, line 19: We already have a function for that called `Transaction.createFeesOnTopTransaction`.
That function is automatically called by `Transaction.createFromPayload` if `transaction.data?.isFeesOnTop && transaction.platformFeeInHostCurrency`.

Do we really need this logic here?

server/paymentProviders/opencollective/host.js
kewitz: @SudharakaP
This is missing `PlatformTipForTransactionGroup` and the value should be calculated and stored in USD, OpenCollective's currency.
`data.hostToPlatformFxRate` will also need to be calculated in the case host's currency is not USD.
I strongly recommend using `Transaction.createFeesOnTopTransaction` through `Transaction.createFromPayload`.
Check `createFeesOnTopTransaction` out and let me know how this differs from this one and we can discuss the best way forward.

kewitz (changes requested):

server/lib/payments.js
```diff
@@ -389,7 +390,7 @@ const validatePayment = payment => {
     throw new Error('Interval should be null, month or year.');
   }
 
-  if (!payment.amount) {
+  if (!payment.amount && !payment.platformTip) {
```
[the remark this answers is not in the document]
SudharakaP: @kewitz : The `platformTip` might not be populated on every case; but this is done so that after the `payment.amount` is paid the user adds the platform tip (in the second modal) and in that case the `payment.amount` is will be null or undefined whereas we do send the platform tip. So the payment has to go through with only the `platformTip`. 😄

server/paymentProviders/opencollective/host.js
SudharakaP: @kewitz : My understanding is that the `createFeesOnTopTransaction` is done when we have a transaction and some fee on top of it. So we send the amount and the fee in a single transaction. However as per the new changes to the front-end what we have now is two sections of the modal. Where the first request sends the transaction amount which goes from the source to the collective. And the platform tip is sent separately in the second part of the modal. In this case the `amount` is undefined and the platform tip should go from the host to opencollective which I think feesOnTop doesn't handle). Which is why I am having this function. Does that make sense? Maybe we'll arrange a call to clarify this further. 😄

SudharakaP:
@kewitz : I've made some small differences to your code review (had to add one field for the transaction to make it work correctly; 24836d6) and re-based. I think now this is good. Let me know if you have any further reviews. 😄

server/paymentProviders/opencollective/host.js
```diff
@@ -16,6 +18,24 @@ paymentMethodProvider.getBalance = () => {
   return Promise.resolve(maxInteger);
 };
 
+paymentMethodProvider.createPlatformTipTransaction = async payload => {
+  const transaction = payload.transaction;
+  const TransactionGroup = transaction.TransactionGroup || uuid();
+  const donationTransaction = {
+    ...transaction,
+    amount: transaction.data.platformTipInHostCurrency,
```
Betree, line 26: This was added recently

```suggestion
    amount: transaction.data.platformTipInHostCurrency,
    kind: TransactionKind.PLATFORM_TIP,
```

server/graphql/v1/mutations/collectives.js
```diff
@@ -200,6 +200,9 @@ export async function createCollectiveFromGithub(_, args, req) {
     collectiveData.LastEditedByUserId = user.id;
     collective = await models.Collective.create(collectiveData);
     const host = await models.Collective.findByPk(defaultHostCollective('opensource').CollectiveId);
+    if (user.email.match(/.*test.*host@opencollective.com$/)) {
```
Betree, line 203: I suppose that this was added for `test/cypress/integration/04-collective.test.js`? Even if it's not for production, adding you as an admin of the host when you create a collective under this host is a surprising side-effect. It's usually beneficial to keep the differences between prod & non-prod env as tiny as possible.

For this one, I would suggest using the default test account (`defaultTestUserEmail`) in Cypress. This account is already an admin of BrusselsTogetherASBL, as you can see in `test/cypress/integration/29-host.dashboard.test.js`.

server/paymentProviders/opencollective/host.js
```diff
@@ -16,6 +18,24 @@ paymentMethodProvider.getBalance = () => {
...
+    description: 'Financial contribution (Platform Tip) to Open Collective',
+    netAmountInCollectiveCurrency: transaction.data.platformTipInHostCurrency,
+    FromCollectiveId: payload.FromCollectiveId,
+    hostCurrencyFxRate: 1,
+    TransactionGroup,
+    PlatformTipForTransactionGroup: TransactionGroup,
```
Betree, line 32: We're [changing](https://github.com/opencollective/opencollective-api/pull/5900) how things are working here: 
- `TransactionGroup` should have the same `TransactionGroup` than original transactions (added funds)
- `PlatformTipForTransactionGroup` does not exists anymore

server/paymentProviders/opencollective/host.js
```diff
@@ -16,6 +18,24 @@ paymentMethodProvider.getBalance = () => {
   return Promise.resolve(maxInteger);
 };
 
+paymentMethodProvider.createPlatformTipTransaction = async payload => {
+  const transaction = payload.transaction;
+  const TransactionGroup = transaction.TransactionGroup || uuid();
```
Betree, line 23: :warning: We should not have this `|| uuid()`. If there's no `TransactionGroup` we can't create the platform tip, as platform tips need to be attached to an existing transaction. This is especially true since the `TransactionGroup` is not passed in the payload from https://github.com/opencollective/opencollective-api/pull/5402/files#diff-1ffaf5a7e2c0ad8ce588e2ebdcb1ddf56a3c2fe93e7f52043c2b0082412fe6b7R76, so the platform tip won't be linked to the original transaction.

Betree (changes requested):
Sorry for the very late review on this @SudharakaP, it slipped my attention. I added some feedback below on handling the platform tip (that has recently evolved).

Apart from that, I'm quite concerned about re-using the `addFunds` mutation to add the platform tip, for two reasons:
1. The second call to the mutation is not adding funds, it's creating a platform tip on an existing transaction. The name and the arguments seem unadapted.
2. Platform tips **must** belong to a transaction. The current mutation doesn't link any existing transaction, and it would be difficult to do so since it's not taking any transaction ID as a parameter. How can we therefore know from there which transactions the platform tip should be attached to?

My recommendation would be to create a new mutation like `addPlatformTipToTransaction` (@znarf @kewitz @SudharakaP feel free to find a better name):
- `transaction`: `GraphQLNonNull(TransactionReferenceInput)`
- `amount`: `GraphQLNonNull(AmountInput)`

Behavior:
- Must be an admin
- Can't add a platform tip if there's already a platform tip
- (For now) Must be a collective to collective contribution (no payment method like Stripe)

znarf (changes requested):
@SudharakaP could you integrate the latest feedback from @Betree ?

Things changed a bit with the `kind` refactor to the ledger.

server/graphql/v1/mutations/collectives.js
SudharakaP: @Betree : Please see my comment at, https://github.com/opencollective/opencollective-frontend/pull/5922#discussion_r620624265

server/paymentProviders/opencollective/host.js
SudharakaP: Done. 👍🏽

server/paymentProviders/opencollective/host.js
SudharakaP: Done. 👍🏽

server/paymentProviders/opencollective/host.js
SudharakaP: Done 👍🏽

SudharakaP:
> Sorry for the very late review on this @SudharakaP, it slipped my attention. I added some feedback below on handling the platform tip (that has recently evolved).

No worries. 😄 

> Apart from that, I'm quite concerned about re-using the `addFunds` mutation to add the platform tip, for two reasons:
> 
>     1. The second call to the mutation is not adding funds, it's creating a platform tip on an existing transaction. The name and the arguments seem unadapted.
> 
>     2. Platform tips **must** belong to a transaction. The current mutation doesn't link any existing transaction, and it would be difficult to do so since it's not taking any transaction ID as a parameter. How can we therefore know from there which transactions the platform tip should be attached to?
> 
> 
> My recommendation would be to create a new mutation like `addPlatformTipToTransaction` (@znarf @kewitz @SudharakaP feel free to find a better name):
> 
>     * `transaction`: `GraphQLNonNull(TransactionReferenceInput)`
> 
>     * `amount`: `GraphQLNonNull(AmountInput)`
> 
> 
> Behavior:
> 
>     * Must be an admin
> 
>     * Can't add a platform tip if there's already a platform tip
> 
>     * (For now) Must be a collective to collective contribution (no payment method like Stripe)

@Betree : I've done the changes except for one; which I have a doubt; https://github.com/opencollective/opencollective-frontend/pull/5922#discussion_r620624265. Basically I don't see how the `Add Funds` modal can be accessed from the dashboard. 🤔 Anyways, let me know if you see any other issues with the refactoring. 😄 

One thing I am not sure is the following; 

>     * Can't add a platform tip if there's already a platform tip

So does this mean that each user can only submit one platform tip? 🤔

Betree:
> > Can't add a platform tip if there's already a platform tip
>
> So does this mean that each user can only submit one platform tip?

There can be only one platform tip per transaction group. The mutation should fail if there's already a platform tip attached to the transaction you're submitting.

SudharakaP:
> > > Can't add a platform tip if there's already a platform tip
> > 
> > 
> > So does this mean that each user can only submit one platform tip?
> 
> There can be only one platform tip per transaction group. The mutation should fail if there's already a platform tip attached to the transaction you're submitting.

@Betree : Thanks much for the clarification. I've added that as well. 😄

server/graphql/v2/mutation/AddFundsMutations.ts
```diff
@@ -27,6 +27,12 @@ export const addFundsMutation = {
       throw new ValidationFailed(`Adding funds is only possible for the following types: ${allowedTypes.join(',')}`);
     }
 
+    if (args.hostFeePercent < 0 || args.hostFeePercent > 100) {
```
Betree, line 30: It's nice to have these checks in place :+1:

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -2,16 +2,59 @@ import express from 'express';
...
+    type: new GraphQLNonNull(Transaction),
+    description: 'Add platform tips to a transaction',
+    args: {
+      transaction: {
+        type: new GraphQLNonNull(TransactionReferenceInput),
+        description: 'Reference to the transaction in the platform tip',
+      },
+      amount: {
+        type: GraphQLNonNull(AmountInput),
+        description: 'Reference to the amount of input for the platform tip',
```
Betree, line 26:
```suggestion
        description: 'Amount of the platform tip',
```

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -2,16 +2,59 @@ import express from 'express';
...
+
+      const isPlatformTipAvailable = await models.Transaction.findOne({
+        where: {
+          TransactionGroup: transaction.TransactionGroup,
+          kind: 'PLATFORM_TIP',
+        },
+      });
+
+      if (isPlatformTipAvailable) {
+        throw new Error('Platform tip is already available for this transaction group');
```
Betree, line 48:
```suggestion
        throw new Error('Platform tip is already set for this transaction group');
```

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -2,16 +2,59 @@ import express from 'express';
...
+    async resolve(_: void, args, req: express.Request): Promise<typeof Transaction> {
+      if (!req.remoteUser) {
+        throw new Unauthorized('You need to be logged in to add a platform tip');
+      }
+
+      const transaction = await fetchTransactionWithReference(args.transaction);
+
+      if (!req.remoteUser.isAdmin(transaction.HostCollectiveId)) {
+        throw new Unauthorized('Only host admins can add platform tips');
+      }
```
Betree, line 38: We must implement the condition mentioned in https://github.com/opencollective/opencollective-api/pull/5402#pullrequestreview-642129612: (For now) Must be a collective to collective contribution (no payment method like Stripe).

The simplest way I can think of is:

```suggestion
      } else if (transaction.kind !== TransactionKind.ADDED_FUNDS) {
      	throw new ValidationFailed('Platform tips can only be added on added funds at the moment')
      }
```

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -2,16 +2,59 @@ import express from 'express';
...
+        type: GraphQLNonNull(AmountInput),
+        description: 'Reference to the amount of input for the platform tip',
+      },
+    },
+    async resolve(_: void, args, req: express.Request): Promise<typeof Transaction> {
+      if (!req.remoteUser) {
+        throw new Unauthorized('You need to be logged in to add a platform tip');
+      }
+
+      const transaction = await fetchTransactionWithReference(args.transaction);
```
Betree, line 34:
```suggestion
      const transaction = await fetchTransactionWithReference(args.transaction, { throwIfMissing: true });
```

test/server/graphql/v2/mutation/AddFundsMutations.test.js
```diff
@@ -0,0 +1,114 @@
...
+          description: 'add funds as admin',
+          hostFeePercent: 6,
+          platformFeePercent: 0,
+        },
+        collectiveAdmin,
+      );
+      expect(result.errors).to.exist;
+      expect(result.errors[0].message).to.match(
+        /You don't have enough permissions to use this payment method \(you need to be an admin of the collective that owns this payment method\)/,
+      );
```
Betree, line 94: Nice to have this test :+1:

server/models/Transaction.js
```diff
@@ -632,7 +632,7 @@ function defineModel() {
     transaction,
     PaymentMethodId,
   }) => {
-    if (!transaction.amount) {
+    if (!transaction.amount && !transaction.platformTipInHostCurrency) {
```
Betree, line 635: Sounds like we don't need this anymore?

```suggestion
    if (!transaction.amount) {
```

server/paymentProviders/opencollective/host.js
```diff
@@ -69,9 +70,9 @@ paymentMethodProvider.processOrder = async order => {
     },
   };
 
-  const transactions = await models.Transaction.createFromPayload(payload);
-
-  return transactions;
+  if (payload.transaction.amount > 0) {
+    return await models.Transaction.createFromPayload(payload);
+  }
```
Betree, line 75: Sounds like we don't need this anymore?

```suggestion
  return models.Transaction.createFromPayload(payload);
```

server/lib/payments.js
```diff
@@ -389,7 +389,7 @@ const validatePayment = payment => {
...
```
Betree, line 392: Sounds like we don't need this anymore?

```suggestion
  if (!payment.amount) {
```

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -2,16 +2,59 @@ import express from 'express';
...
+
+      if (isPlatformTipAvailable) {
+        throw new Error('Platform tip is already available for this transaction group');
+      }
+
+      const payload = {
+        transaction,
+        amount: args.amount.valueInCents,
+      };
+      return await createPlatformTipTransaction(payload);
```
Betree, line 55: Let's rather use https://github.com/opencollective/opencollective-api/blob/main/server/models/Transaction.js#L567. You'll probably need to update the transaction with `isFeesOnTop` + `platformFeeInHostCurrency` first.

Betree (commented):
Nice update :+1: Having some tests on `addPlatformTipToTransaction` would be nice.

server/graphql/v2/mutation/TransactionMutations.ts
Betree: Note: we'll also need to update the opposite transaction (CREDIT/DEBIT)

SudharakaP: @Betree : Done. 😄 So my thinking is opposite transaction is auto updated within https://github.com/opencollective/opencollective-api/blob/60088ad8718da967b78491b7f2f4a25d946d3672/server/models/Transaction.js#L608 ? Sorry, maybe I am confused by this function. 🤔 

On another note, so I've got rid of my custom function `createPlatformTipTransaction` and simplified the logic at, https://github.com/opencollective/opencollective-api/pull/5402/commits/5de069717a2d3a6c7fa033771884c4f179b74393#diff-37b360415f67c48408202a4754178f4bd2d792651489f40714fc6cff77e04a65R55-R66. Let me know if that looks good. 😃

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -2,16 +2,69 @@ import express from 'express';
...
+      transaction = {
+        ...transaction,
+        data: { isFeesOnTop: true },
+        platformFeeInHostCurrency: args.amount.valueInCents,
+        FromCollectiveId: transaction.HostCollectiveId,
+        kind: TransactionKind.PLATFORM_TIP,
+        ...FEES_ON_TOP_TRANSACTION_PROPERTIES,
+      };
+
+      await models.Transaction.createFeesOnTopTransaction({ transaction, host: transaction.HostCollectiveId });
```
Betree, line 64: It seems that `createFeesOnTopTransaction` expects `host` to be a Collective model, not an integer: https://github.com/opencollective/opencollective-api/blob/69fb382095a65afe6d6da2554b5c2b7a83bed66a/server/models/Transaction.js#L587

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -2,16 +2,69 @@ import express from 'express';
...
+      return transaction.dataValues;
```
Betree, line 65: Why return the `dataValues` rather than the model? This will break if we try to fetch files like `fromAccount` on this transaction like in https://github.com/opencollective/opencollective-api/blob/69fb382095a65afe6d6da2554b5c2b7a83bed66a/server/graphql/v2/object/Debit.js#L17

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -2,16 +2,69 @@ import express from 'express';
...
+      });
+
+      if (isPlatformTipAvailable) {
+        throw new Error('Platform tip is already set for this transaction group');
+      }
+
+      transaction = {
+        ...transaction,
+        data: { isFeesOnTop: true },
+        platformFeeInHostCurrency: args.amount.valueInCents,
```
Betree, line 58: The goal of `AmountInput` is to offer a flexible API where people can use either float amounts or integers (cents). Using `getValueInCentsFromAmountInput` ensures that we support that flexible API, regardless of what parameters were given.

```suggestion
        platformFeeInHostCurrency: getValueInCentsFromAmountInput(args.amount),
```

Betree (changes requested):

server/graphql/v2/mutation/TransactionMutations.ts
SudharakaP: Ops I missed that; corrected. 👍🏽

server/graphql/v2/mutation/TransactionMutations.ts
SudharakaP: Thanks. 👍🏽

server/graphql/v2/mutation/TransactionMutations.ts
SudharakaP: @Betree : I did this because otherwise trying to return the transaction I get the below error. Maybe because I am using the spread operator above to add some fields. But couldn't figure out exactly why; do you have any ideas? 🤔 

```
Error: Abstract type "Transaction" must resolve to an Object type at runtime for field "Mutation.addPlatformTipToTransaction" with value { dataValues: { id: 53096, type: "CREDIT", kind: "ADDED_FUNDS", uuid: "7f94aa79-46c0-4a1a-830a-22bf58185caa", description: "test", amount: 1800, currency: "USD", CreatedByUserId: 9476, FromCollectiveId: 2600, CollectiveId: 10886, HostCollectiveId: 10885, UsingGiftCardFromCollectiveId: null, OrderId: 6432, ExpenseId: null, hostCurrency: "USD", hostCurrencyFxRate: 1, amountInHostCurrency: 1800, platformFeeInHostCurrency: 0, hostFeeInHostCurrency: -72, paymentProcessorFeeInHostCurrency: 0, taxAmount: null, netAmountInCollectiveCurrency: 1728, data: [Object], TransactionGroup: "d1e9083e-4f22-4e6a-9b19-183f3a6d6b0d", RefundTransactionId: null, isRefund: false, createdAt: {}, updatedAt: {}, deletedAt: null, PaymentMethodId: 8768 }, _previousDataValues: { id: 53096, type: "CREDIT", kind: "ADDED_FUNDS", uuid: "7f94aa79-46c0-4a1a-830a-22bf58185caa", description: "test", amount: 1800, currency: "USD", CreatedByUserId: 9476, FromCollectiveId: 2600, CollectiveId: 10886, HostCollectiveId: 10885, UsingGiftCardFromCollectiveId: null, OrderId: 6432, ExpenseId: null, hostCurrency: "USD", hostCurrencyFxRate: 1, amountInHostCurrency: 1800, platformFeeInHostCurrency: 0, hostFeeInHostCurrency: -72, paymentProcessorFeeInHostCurrency: 0, taxAmount: null, netAmountInCollectiveCurrency: 1728, data: [Object], TransactionGroup: "d1e9083e-4f22-4e6a-9b19-183f3a6d6b0d", RefundTransactionId: null, isRefund: false, createdAt: {}, updatedAt: {}, deletedAt: null, PaymentMethodId: 8768 }, _changed: {}, _options: { isNewRecord: false, _schema: null, _schemaDelimiter: "", raw: true, attributes: [Array] }, isNewRecord: false, data: { isFeesOnTop: true }, platformFeeInHostCurrency: 0, FromCollectiveId: 10885, kind: "PLATFORM_TIP", CollectiveId: 8686, HostCollectiveId: 8686, hostCurrency: "USD", currency: "USD", paymentProcessorFeeInHostCurrency: NaN, amountInHostCurrency: NaN, amount: NaN }, received "undefined". Either the "Transaction" type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.
```

SudharakaP: @Betree : I've refactored the code at 695a02b to make sure we return the `Transaction` object. Let me know if this is good. 😄

SudharakaP:
@Betree : I've refactored the code to make sure all the issues you raised are corrected; also I want to point your attention to the following piece of code; 

https://github.com/opencollective/opencollective-api/blob/1067fc85f86d1d66202c7e6751420bec567ddc44/server/graphql/v2/mutation/TransactionMutations.ts#L55-L72

where we define the platform tip transaction fields. Note that the function, `createFeesOnTopTransaction` doesn't return a model.Transaction object rather it takes it's own payload, so I just used another query to get the platform tip transaction back so that we could return the transaction object. I don't know if this is the best way of doing things, maybe we could later extract and refactor `createFeesOnTopTransaction` to return the inserted transactions; but I didn't want to do that as it seems a bit complicated and I don't know what side effects it will have. Let me know your thoughts. 🤔

Betree:
> maybe we could later extract and refactor createFeesOnTopTransaction to return the inserted transactions; but I didn't want to do that as it seems a bit complicated and I don't know what side effects it will have. Let me know your thoughts. thinking

That sounds like the right thing to do. If we don't want to take the risk of breaking the existing code, an option is to return both: `return { transaction, platformTipTransaction }`, that way it's easy to keep the compatibility with existing code.

SudharakaP:
> > maybe we could later extract and refactor createFeesOnTopTransaction to return the inserted transactions; but I didn't want to do that as it seems a bit complicated and I don't know what side effects it will have. Let me know your thoughts. thinking
> 
> That sounds like the right thing to do. If we don't want to take the risk of breaking the existing code, an option is to return both: `return { transaction, platformTipTransaction }`, that way it's easy to keep the compatibility with existing code.

@Betree : Ah, thanks. I've made that change as well. Let me know if you see any further issues. 😄

server/graphql/v2/mutation/TransactionMutations.ts
```diff
@@ -2,16 +2,70 @@ import express from 'express';
...
+        throw new Error('Platform tip is already set for this transaction group');
+      }
+
+      const transaction = {
+        data: { isFeesOnTop: true },
+        TransactionGroup: fundsTransaction.TransactionGroup,
+        platformFeeInHostCurrency: getValueInCentsFromAmountInput(args.amount),
+        FromCollectiveId: fundsTransaction.HostCollectiveId,
+        kind: TransactionKind.PLATFORM_TIP,
+        ...FEES_ON_TOP_TRANSACTION_PROPERTIES,
```
Betree, line 61: If we don't set `amountInHostCurrency` here, my understanding is that `feeOnTopPercent` will be `NaN`, which can impact other values: https://github.com/opencollective/opencollective-api/blob/606b7c02928f22df8ec7b9269faf6b50778f0023/server/models/Transaction.js#L586

SudharakaP: @Betree : Great catch; thanks a lot. 😄 I just pushed a change (86d90a4) to set this amount to the platform tip amount. My understanding is that the conversion rate multiplier should be 1 here as the open collective hosts currency is always 'USD'. Let me know if that's not the case.

Betree (commented):
I did a final test, everything works great! Comparing the platform tip transactions created here with the ones created in the contribution flow, I noticed some minor differences:
- `CreatedByUserId`: should be the user who added the tip
- `PaymentMethodId`: should maybe be the collective payment method? I'm not sure if we're still using this. @znarf?

znarf:
Regarding `PaymentMethodId` I'm not sure what make sense. I think that might be the right call to use the `collective` Payment Method there.

SudharakaP:
> I did a final test, everything works great! Comparing the platform tip transactions created here with the ones created in the contribution flow, I noticed some minor differences:
> 
>     * `CreatedByUserId`: should be the user who added the tip (`req.remoteUser.id`)
> 
>     * `PaymentMethodId`: should maybe be the collective payment method? I'm not sure if we're still using this. @znarf?

@Betree : I've added both of these to the transactions object. Thanks for pointing it out. 😄

Betree (approved):
Lgtm! :rocket:

PR #5411, opened 2021-02-23
Exclude unclaimed giftcards

znarf:


Betree:
Good hotfix. Would be great to have some SQL filtering on that in the future.

PR #5424, opened 2021-02-24
Behavior consistency when an expense gets paid

Betree:
Require https://github.com/opencollective/opencollective-frontend/pull/5938

This PR addresses behavior inconsistencies when expenses get paid, mainly in the `CONTRIBUTOR` member creation and the "Expense paid" activity creation.

- Fix https://github.com/opencollective/opencollective/issues/2753: The new code now checks if a member exists before creating it and the migration makes sure we delete the duplicates

- Fix the missing member issue reported in [Slack](https://opencollective.slack.com/archives/C0RMV6F8C/p1614133975019300)
We were only adding users as contributors when the expense was paid manually. I've moved the member creation in `setPaid` to make sure we always create the entry, regardless of how the expense was paid. The migration takes care of creating all the entries that we missed.

- Fix missing webhook notifications: there was no `createActivity` when paying an expense with PayPal

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -684,21 +684,10 @@ export async function deleteExpense(req, expenseId): Promise<typeof models.Expen
 }
 
 /** Helper that finishes the process of paying an expense */
-async function markExpenseAsPaid(expense, remoteUser): Promise<typeof models.Expense> {
+async function markExpenseAsPaid(expense, remoteUser, isManualPayout = false): Promise<typeof models.Expense> {
   debug('update expense status to PAID', expense.id);
   await expense.setPaid(remoteUser.id);
-
-  // Add user as CONTRIBUTOR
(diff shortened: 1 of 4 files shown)


server/graphql/common/expenses.ts
```diff
@@ -736,7 +725,7 @@ async function payExpenseWithPayPal(remoteUser, expense, host, paymentMethod, to
       fees['hostFeeInHostCurrency'],
       fees['platformFeeInHostCurrency'],
     );
-    expense.setPaid(remoteUser.id);
+    await markExpenseAsPaid(expense, remoteUser);
```
Betree, line 728: This is what fixes https://opencollective.freshdesk.com/a/tickets/13223: `markExpenseAsPaid` also creates the activity.

snitin315 (approved):
Thanks 🚀

PR #5430, opened 2021-02-25
Expense: Create activity when draft expense gets verified

Betree:
Fix https://github.com/opencollective/opencollective/issues/4020 (https://opencollective.freshdesk.com/a/tickets/13223)

This PR will trigger a `COLLECTIVE_EXPENSE_CREATED` activity when a draft is verified, but it won't send the associated email (only the webhooks will be triggered).

@kewitz I'm wondering a few things:
1. Are collective admins notified in some way when there's a new draft? Can they see them listed somewhere?
2. Don't we want to send a "New expense" email notification too when the expense gets verified?
3. If the answer to (1) is yes, shouldn't we generate the `COLLECTIVE_EXPENSE_CREATED` activity when the draft gets created?

### DIFF
--- a/server/graphql/v2/mutation/ExpenseMutations.ts
+++ b/server/graphql/v2/mutation/ExpenseMutations.ts
@@ -440,7 +440,7 @@ const expenseMutations = {
       // Technically the expense was already created, but it was a draft. It truly becomes visible
       // for everyone (especially admins) at this point, so it's the right time to trigger `COLLECTIVE_EXPENSE_CREATED`
       await expense
-        .createActivity(activities.COLLECTIVE_EXPENSE_CREATED, req.remoteUser, { skipEmail: true })
+        .createActivity(activities.COLLECTIVE_EXPENSE_CREATED, req.remoteUser)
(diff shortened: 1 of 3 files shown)


kewitz (approved):
:cherries: :cake: :ok_hand:.

kewitz:
@Betree only the invited person is notified through COLLECTIVE_EXPENSE_INVITE_DRAFTED.
My rationale was that since the draft is temporary, I wouldn't like to bother collective admins if this draft never becomes an expense.
This is a good catch, and I think it makes sense to notify admins when the draft becomes a pending expense. This part was missing.

Betree:
> @Betree only the invited person is notified through COLLECTIVE_EXPENSE_INVITE_DRAFTED.
> My rationale was that since the draft is temporary, I wouldn't like to bother collective admins if this draft never becomes an expense.
> This is a good catch, and I think it makes sense to notify admins when the draft becomes a pending expense. This part was missing.

Ok, I'll remove the `skipEmail` to make sure they get an email notification then. Thanks!

PR #5431, opened 2021-02-25
Update currency handling for PayPal payments

znarf:
It's currently stored in a `hostCurrency` that doesn't necessarily match the host currency.

Can be reproduced when the Order currency doesn't match the host currency: "I have a Collective in USD on a Fiscal Host in EUR, when someone make a PayPal payment initiated in USD, it will be stored in USD instead of EUR".

Only one occurrence:

```
SELECT * FROM "Transactions", "PaymentMethods", "Collectives" as "Hosts"
WHERE "Transactions"."PaymentMethodId" = "PaymentMethods"."id"
AND "Transactions"."deletedAt" IS NULL
AND "PaymentMethods"."service" = 'paypal' AND "PaymentMethods"."type" = 'payment'
AND "Hosts"."id" = "Transactions"."HostCollectiveId"
AND "Hosts"."currency" != "Transactions"."hostCurrency";
```

### DIFF
--- a/server/paymentProviders/paypal/payment.js
+++ b/server/paymentProviders/paypal/payment.js
@@ -3,6 +3,7 @@ import { get } from 'lodash';
 import fetch from 'node-fetch';
 
 import * as constants from '../../constants/transactions';
+import { getFxRate } from '../../lib/currency';
 import logger from '../../lib/logger';
 import { floatAmountToCents } from '../../lib/math';
 import { getHostFee, getPlatformFee } from '../../lib/payments';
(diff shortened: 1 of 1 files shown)


server/paymentProviders/paypal/payment.js
```diff
@@ -123,40 +124,51 @@ export async function executePayment(order) {
...
-  const paypalFee = parseFloat(get(transaction, 'related_resources.0.sale.transaction_fee.value', '0.0'));
-  const amountFromPayPalInCents = floatAmountToCents(amountFromPayPal);
-  const paypalFeeInCents = floatAmountToCents(paypalFee);
-  const currencyFromPayPal = transaction.amount.currency;
 
-  // TODO: Double check why in one case we're using amountFromPayPalInCents and in the other order.totalAmount
-  const hostFeeInHostCurrency = await getHostFee(amountFromPayPalInCents, order);
-  const platformFeeInHostCurrency = await getPlatformFee(order.totalAmount, order);
+  const amount = floatAmountToCents(parseFloat(transaction.amount.total));
+  const currency = transaction.amount.currency;
```
kewitz, line 129: The assumption here is that this currency is equal to the collective currency.
It is implicit in the API, but from the frontend, this currency is set to be `tier.amount.currency || collective.currency`, which validates the assumption.

server/paymentProviders/paypal/payment.js
```diff
@@ -123,40 +124,51 @@ export async function executePayment(order) {
...
+
+  const paypalTransactionFee = parseFloat(get(transaction, 'related_resources.0.sale.transaction_fee.value', '0.0'));
+  const paymentProcessorFee = floatAmountToCents(paypalTransactionFee);
```
kewitz, line 132: :ice_cream:  Could join and make it a single assignment.

server/paymentProviders/paypal/payment.js
```diff
@@ -123,40 +124,51 @@ export async function executePayment(order) {
...
+
+  const host = await order.collective.getHostCollective();
+  if (!host) {
+    throw new Error(`Cannot create transaction: collective id ${order.collective.id} doesn't have a host`);
+  }
+  const hostCurrency = host.currency;
```
kewitz, line 138: :ok_hand:

kewitz (approved):
Nice, LGTM!

PR #5439, opened 2021-02-26
Update balance handling in GraphQL

znarf:
Expose new balance handling in GraphQL. Use loaders whenever possible.


kewitz (approved):
Nice work!
Thanks for getting rid of custom SQL queries, I think this is an improvement towards readability but I'm not sure how this will affect performance.

PR #5561, opened 2021-03-01
Members: Add email argument

Betree:
Athens collective needs (see https://github.com/athensresearch/athens/issues/694) to be able to see if someone is a contributor using an email address on GQLV2. I've added that to the `member` field, so the query looks pretty much like that:

```graphql
{
  account(slug: "athens") {
    members(email: "searched@gmail.com") {
      nodes {
        id
        role
        account {
          id
          slug
          ... on Individual {
            email
          }
        }
      }
    }
  }
}
```

This argument can only be used by collective admins.

### DIFF
--- a/server/graphql/v2/interface/HasMembers.js
+++ b/server/graphql/v2/interface/HasMembers.js
@@ -1,10 +1,12 @@
 import { GraphQLInt, GraphQLList, GraphQLNonNull } from 'graphql';
 
 import models, { Op } from '../../../models';
+import { BadRequest } from '../../errors';
 import { MemberCollection } from '../collection/MemberCollection';
 import { AccountType, AccountTypeToModelMapping } from '../enum/AccountType';
 import { MemberRole } from '../enum/MemberRole';
 import { ChronologicalOrderInput } from '../input/ChronologicalOrderInput';
+import EmailAddress from '../scalar/EmailAddress';
 
 export const HasMembersFields = {
   members: {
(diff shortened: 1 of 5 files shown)


znarf (approved):
Works for me!

server/graphql/v2/interface/HasMembers.js
```diff
@@ -38,6 +45,14 @@ export const HasMembersFields = {
         };
       }
 
+      if (args.email) {
+        if (!req.remoteUser.isAdminOfCollective(collective)) {
+          throw new BadRequest('Only admins can lookup for members using the "email" argument');
+        } else {
+          collectiveInclude.push({ association: 'user', required: true, where: { email: args.email.toLowerCase() } });
```
kewitz, line 52: I think we should filter out incognito collectives.

kewitz (approved):
LGTM.
I like the fact that only the collective admin has access to this feature, and also only to its contributors.
Perhaps a simple test to prevent regressions in the future?

server/graphql/v2/interface/HasMembers.js
Betree: I was wondering about this one, but it seems that historically collective admins have access to incognito user's emails: https://github.com/opencollective/opencollective-api/blob/661638353b7287fd5c590393592ec5a6946244a5/server/graphql/v1/types.js#L368.

Keeping things this way is actually interesting for Athens: they should be able to verify if someone contributed - regardless of whether it was made as an incognito contribution or not.

tangjeff0: In our case, users want us, the collective they are sponsoring to have their email address. This makes sense if you think about Substack's case about helping creators (collectives?) maintain a direct relationship with their readers.

tangjeff0:
Getting 

```
{
  "error": {
    "errors": [
      {
        "message": "Unknown argument \"email\" on field \"Account.members\".",
        "locations": [
          {
            "line": 4,
            "column": 13
          }
        ],
        "extensions": {
          "code": "GRAPHQL_VALIDATION_FAILED"
        }
      }
    ]
  }
}
```

with your query https://github.com/opencollective/opencollective-api/pull/5561#issue-582161147 @Betree.

Betree:
@tangjeff0 it's not deployed yet. I'll ping you once it's live, probably tomorrow

Betree:
@tangjeff0 It's live

tangjeff0:
@Betree looks like it works!

PR #5600, opened 2021-03-03
Typescript: Include all files and fix ESLint issues

Betree:
Some files (tests, scripts, etc...) were not included in the scope of typescript. This PR fixes that, along with one small typescript error that was generated after this change.

This will also improve auto-completion for typescript files on VSCode.

After this PR, there are 66 ESLint warnings left:
- 349 have been resolved
- 28 GraphQL deprecation warnings have been hidden

Most of the warnings left could be resolved by https://github.com/opencollective/opencollective/issues/4054.


server/paymentProviders/transferwise/index.ts
```diff
@@ -54,17 +64,17 @@ async function quoteExpense(connectedAccount, payoutMethod, expense): Promise<Qu
   const quote = await transferwise.createQuote(token, {
     profileId: connectedAccount.data.id,
     sourceCurrency: expense.currency,
-    targetCurrency: payoutMethod.data.currency,
+    targetCurrency: <string>payoutMethod.unfilteredData.currency,
```
Betree, line 67: @kewitz There are a few places where I replaced `data` by `unfilteredData`. My understanding is that `data` will never contain the currency because it's filtered out in https://github.com/opencollective/opencollective-api/blob/a23c49e9a1557fb214e512e7692748b183beb813/server/models/PayoutMethod.ts#L56. 

Let me know if this is a mistake.

kewitz: @Betree makes sense, but I think we can safely expose currency in data to make it simpler.

PR #5641, opened 2021-03-04
Minor fixes

Betree:


server/graphql/v1/mutations/paymentMethods.js
```diff
@@ -195,7 +195,7 @@ export async function updatePaymentMethod(args, remoteUser) {
     throw PaymentMethodPermissionError;
   }
 
-  return models.PaymentMethod.update(pick(args, allowedFields), { where: { id: paymentMethod.id } });
+  return paymentMethod.update(pick(args, allowedFields));
```
Betree, line 198: @znarf I don't know if it was working before the Sequelize update, but we were not returning a proper payment method here.

PR #5649, opened 2021-03-06
feat(Events): add support for online events

snitin315:
Refers https://github.com/opencollective/opencollective/issues/3439 

Related FE PR -  https://github.com/opencollective/opencollective-frontend/pull/6008

## Screenshot

Screenshot 2021-03-11 at 4 33 27 PM

### DIFF
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -652,6 +652,10 @@ export const CollectiveInterfaceType = new GraphQLInterfaceType({
         type: GraphQLJSON,
         deprecationReason: '2020-10-08: This field is not provided anymore and will return an empty object',
       },
+      privateInstructions: {
+        type: GraphQLString,
+        description: 'Private instructions related to an event',
+      },
       githubContributors: { type: new GraphQLNonNull(GraphQLJSON) },
       slug: { type: GraphQLString },
       path: { type: GraphQLString },
@@ -1105,6 +1109,18 @@ const CollectiveFields = () => {
       deprecationReason: '2020-10-08: This field is not provided anymore and will return an empty object',
       resolve() {
         return {};
+      },
+    },
+    privateInstructions: {
+      type: GraphQLString,
+      description: 'Private instructions related to an event',
+      resolve(collective, _, req) {
+        if (
+          collective.type === types.EVENT &&
+          (req.remoteUser?.isAdminOfCollective(collective) || req.remoteUser?.hasRole(roles.PARTICIPANT, collective))
+        ) {
+          return collective.data?.privateInstructions;
+        }
       },
     },
     githubContributors: {
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -56,6 +56,12 @@ export async function createCollective(_, args, req) {
       };
     }
   }
+  // Set private instructions
+  if (args.collective.privateInstructions) {
+    collectiveData.data = {
+      privateInstructions: args.collective.privateInstructions,
+    };
+  }
 
   collectiveData.isActive = false;
   if (args.collective.ParentCollectiveId) {
@@ -291,7 +297,7 @@ export function editCollective(_, args, req) {
   }
 
   const newCollectiveData = {
-    ...omit(args.collective, ['location', 'type', 'ParentCollectiveId', 'data']),
+    ...omit(args.collective, ['location', 'type', 'ParentCollectiveId', 'data', 'privateInstructions']),
     LastEditedByUserId: req.remoteUser.id,
   };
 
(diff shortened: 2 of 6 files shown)


server/graphql/v1/CollectiveInterface.js
```diff
@@ -650,7 +650,7 @@ export const CollectiveInterfaceType = new GraphQLInterfaceType({
       },
       data: {
         type: GraphQLJSON,
-        deprecationReason: '2020-10-08: This field is not provided anymore and will return an empty object',
+        description: 'Data related to events like privateInstructions',
```
Betree, line 653: :warning: We should keep this deprecation. We don't want to expose the full collective `data`, especially as it might contain some private information that is not meant to be exposed. We should rather expose a new GraphQL field like `privateInstructions`, that exposes just this information.

The permission check also needs to be a bit more permissive: all event participants must be allowed to see this. We need something like this (needs to be tested):
```es6
if (collective.type === types.EVENT && (
	req.remoteUser?.isAdminOfCollective(collective)
	|| req.remoteUser?.hasRole(roles.PARTICIPANT, collective)
) {
  return collective.data?.privateInstructions
}
```

server/graphql/v1/mutations/collectives.js
```diff
@@ -291,7 +291,7 @@ export function editCollective(_, args, req) {
   }
 
   const newCollectiveData = {
-    ...omit(args.collective, ['location', 'type', 'ParentCollectiveId', 'data']),
```
Betree, line 294: We should pick just the fields that we want, users should not be able to override the full `data` object

templates/emails/ticket.confirmed.hbs
```diff
@@ -13,7 +13,20 @@ Subject: {{ order.quantity }} {{pluralize "ticket" n=order.quantity}} confirmed
 
 <p>This confirms your {{ order.quantity }} {{pluralize "ticket" n=order.quantity}}{{#if order.totalAmount}} for a total of {{currency order.totalAmount currency=order.currency}}{{/if}}. </p>
 
-<p>We are excited that you'll be joining us at <a href="{{config.host.website}}/{{collective.slug}}/events/{{event.slug}}">{{ event.name }}</a> on {{ moment event.startsAt timezone=event.timezone format="MMMM Do YYYY" }} at {{ moment event.startsAt timezone=event.timezone format="h:mma" }} at {{ event.locationName }} ({{event.address}}).</p>
+<p>We are excited that you'll be joining us at <a href="{{config.host.website}}/{{collective.slug}}/events/{{event.slug}}">{{ event.name }}</a> on {{ moment event.startsAt timezone=event.timezone format="MMMM Do YYYY" }} at {{ moment event.startsAt timezone=event.timezone format="h:mma" }}
+{{#if isOffline }}
+  at {{ event.locationName }} ({{event.address}}).
+{{else}}
```
Betree, line 19: Maybe we can make it optional

```suggestion
{{else if event.address}}
```

Betree (changes requested):
The code looks good, most of my feedback is about `collective.data` that we must treat like something the user should not be able to read or write in its entirety.

snitin315:
@Betree I have applied the suggestion.

server/graphql/v1/mutations/collectives.js
```diff
@@ -381,7 +381,16 @@ export function editCollective(_, args, req) {
         return collective.updateCurrency(newCollectiveData.currency, req.remoteUser);
       }
     })
-    .then(() => collective.update(omit(newCollectiveData, ['HostCollectiveId', 'hostFeePercent', 'currency']))) // we omit those attributes that have already been updated above
+    .then(() => {
+      // Set private instructions
+      const privateInstructions = args.collective.data?.privateInstructions;
```
Betree, line 386: `collective.data` is deprecated on the `input` too: https://github.com/opencollective/opencollective-api/blob/1d321f312479e98cfd99dcfd34d20223fe31fdeb/server/graphql/v1/inputTypes.js#L171

We need to add a new `privateInstructions` field on `CollectiveInputType` as well.

snitin315: Updated 👍

server/graphql/v1/mutations/collectives.js
```diff
@@ -313,6 +319,13 @@ export function editCollective(_, args, req) {
     newCollectiveData.countryISO = location.country;
   }
 
+  // Set private instructions value
+  if (args.collective.privateInstructions) {
+    newCollectiveData.data = {
+      privateInstructions: args.collective.privateInstructions,
+    };
```
Betree, line 326: This will erase any existing data from the collective. We need to do:

```es6
    newCollectiveData.data = {
      ...collective.data,
      privateInstructions: args.collective.privateInstructions,
    };
```

Betree (changes requested):

server/graphql/v1/mutations/collectives.js
snitin315: It should be good now.

Betree (approved):

PR #5691, opened 2021-03-12
Only ignore suspicious account 48h from search

znarf:


Betree (commented):
Any idea of the impact in terms of performance? Filtering by JSON is usually not free

znarf:
@Betree good feedback, I haven't benchmarked, I'll check.

znarf:
I quickly checked and it didn't seem to be slower. Shipping this.

PR #5697, opened 2021-03-12
Updates: provide info about audience (+ optimizations)

Betree:
Basis for https://github.com/opencollective/opencollective/issues/2998

- Optimized the queries
- Added a decent test coverage of the functions used to retrieve update targets
- Fixed an issue where someone with multiple roles (ie. backer of the host and admin of a hosted collective) received multiple emails


Betree:
I actually need to update the code to better handle updates for hosted collectives

PR #5699, opened 2021-03-12
Not include image loading in e2e tests

SudharakaP:
Fixes https://github.com/opencollective/opencollective/issues/4083


server/models/Collective.js
```diff
@@ -876,18 +876,20 @@ function defineModel() {
...
   Collective.prototype.checkAndUpdateImage = async function (image) {
-    try {
-      const response = await fetch(image);
-      if (response.status !== 200) {
-        throw new Error(`status=${response.status}`);
-      }
-      const body = await response.text();
-      if (body.length === 0) {
-        throw new Error(`length=0`);
+    if (process.env.OC_ENV !== 'e2e') {
```
Betree, line 879: It's not only for E2E, unit tests are affected too

SudharakaP: Thanks for the review; I've corrected that. 👍🏽

server/models/Collective.js
```diff
@@ -876,18 +876,20 @@ function defineModel() {
...
   Collective.prototype.checkAndUpdateImage = async function (image) {
-    try {
-      const response = await fetch(image);
-      if (response.status !== 200) {
-        throw new Error(`status=${response.status}`);
-      }
-      const body = await response.text();
-      if (body.length === 0) {
-        throw new Error(`length=0`);
+    if (process.env.OC_ENV !== 'e2e' && process.env.OC_ENV !== 'ci') {
```
Betree, line 879: We also want this for local testing

```suggestion
    if (!['e2e', 'ci', 'test'].includes(process.env.OC_ENV)) {
```

SudharakaP:
@Betree : Done the changes you suggested including the local testing change; btw notice that there are two tests in the unit test which we need the image loading for which I've added exceptions. Hope that approach is okay with you. 😄

test/server/models/Collective.test.js
```diff
@@ -428,7 +428,10 @@ describe('server/models/Collective', () => {
       website: 'https://opencollective.com',
     });
     // Make sure clearbit image is fetched (done automatically and async in normal conditions)
+    const originalValue = process.env.OC_ENV;
+    process.env.OC_ENV = 'ci-exception';
     await collective.findImage();
+    process.env.OC_ENV = originalValue;
```
Betree, line 434: Changing the env dynamically is a bit dangerous: if `findImage` throws an exception the env will never be restored and other tests in the chain will break because of that. Plus, we don't know what kind of side-effect it can introduces if part of the targetted code starts to rely on the env too.

I would prefer having a `force` parameter (boolean) on `findImage` / `findImageForUser`. 

Note that though we don't respect it in the actual codebase, tests frameworks usually discourage checking the environment in the code like we do here. The best practice is usually to mock the function (or the [call](https://github.com/nock/nock) to the HTTP server) - globally if necessary.

SudharakaP: @Betree : Yes, I agree. Thanks for the information. I've corrected it by adding a force parameter. 👍🏽

Betree (approved):

PR #5701, opened 2021-03-13
feat: send email when an expense is rejected

bolariin:
Resolves https://github.com/opencollective/opencollective/issues/4081

Email content is based on approved expense email.

### Screenshots

* Reject email for host Reject email for host

* Reject email for expense submitter reject expense for expense submitter


Betree (approved):
Looks great @bolariin, thank you for taking the time to iterate on this :raised_hands: 

I've pushed two small changes, one to simplify a bit the email and the other to remove the email sent to hosts. 

In terms of process, an expense usually goes like: Pending -> Approved -> Paid. Host admins are only involved with the expense once it's approved, before that it's the responsibility of the **collective** admins to deal with it. We'll see if we get any additional feedback, but that should work great for the first version.

PR #5702, opened 2021-03-13
fix url scalar test case to match description

bolariin:
This is not related to a ticket but noticed that the test didn't match the description.


Betree (approved):
Awesome, good catch @bolariin!

PR #5709, opened 2021-03-15
Tier: allow flexible interval

snitin315:
Refers https://github.com/opencollective/opencollective/issues/3720

Related FE PR - https://github.com/opencollective/opencollective-frontend/pull/6047

### DIFF
--- a/server/graphql/v1/types.js
+++ b/server/graphql/v1/types.js
@@ -1262,6 +1262,8 @@ export const TierStatsType = new GraphQLObjectType({
             return req.loaders.Tier.totalMonthlyDonations.load(tier.id);
           } else if (tier.interval === intervals.YEAR) {
             return req.loaders.Tier.totalYearlyDonations.load(tier.id);
+          } else if (tier.interval === intervals.FLEXIBLE) {
+            return req.loaders.Tier.totalDonated.load(tier.id);
           } else {
             return 0;
           }
--- a/server/graphql/v1/types.js
+++ b/server/graphql/v1/types.js
@@ -1255,7 +1255,8 @@ export const TierStatsType = new GraphQLObjectType({
         },
       },
       totalRecurringDonations: {
-        description: 'How much money is given for this tier for each tier.interval (monthly/yearly)',
+        description:
+          'How much money is given for this tier for each tier.interval (monthly/yearly). For flexible tiers, this amount is a monthly average of contributions amount, taking into account both yearly and monthly subscriptions.',
         type: GraphQLInt,
         resolve(tier, args, req) {
           if (tier.interval === intervals.MONTH) {
@@ -1263,7 +1264,7 @@ export const TierStatsType = new GraphQLObjectType({
           } else if (tier.interval === intervals.YEAR) {
             return req.loaders.Tier.totalYearlyDonations.load(tier.id);
           } else if (tier.interval === intervals.FLEXIBLE) {
-            return req.loaders.Tier.totalDonated.load(tier.id);
+            return req.loaders.Tier.totalRecurringDonations.load(tier.id);
           } else {
             return 0;
           }
--- a/server/models/Member.js
+++ b/server/models/Member.js
@@ -131,6 +131,8 @@ function defineModel() {
    * @param {*} member { tier: { interval }, lastDonation}
    */
   Member.isActive = member => {
+    const { models } = sequelize;
+
     if (!member.tier || !member.tier.interval) {
       return true;
     }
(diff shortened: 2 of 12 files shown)


templates/emails/user.monthlyreport.hbs
```diff
@@ -226,7 +226,7 @@ Subject: {{capitalize month}} Report
           {{#if consolidatedPdfs}}<br />You can find PDFs of all your contributions, organized by fiscal host, attached to this email.{{/if}}</li>
           {{#each collectives}}
             {{#if nextGoal}}
-              <li><a href="{{../config.host.website}}/{{slug}}">{{name}}</a> needs {{currency nextGoal.missing.amount currency=currency}} more {{#if nextGoal.missing.interval}}per {{nextGoal.missing.interval}}{{/if}} to reach their next goal of {{nextGoal.title}} ({{nextGoal.percentage}} there). 
+              <li><a href="{{../config.host.website}}/{{slug}}">{{name}}</a> needs {{currency nextGoal.missing.amount currency=currency}} more {{#if nextGoal.missing.interval}} {{#ifCond nextGoal.missing.interval '!==' 'flexible'}}per {{nextGoal.missing.interval}} {{/ifCond}} {{/if}} to reach their next goal of {{nextGoal.title}} ({{nextGoal.percentage}} there).
```
Betree, line 229: My understanding from https://github.com/opencollective/opencollective-api/blob/87629f67ba47df6fad7dcb478f01a6fc9d803c51/server/models/Collective.js#L722 is that it can only be `month` or `year`, not `flexible`

Betree (commented):
Some paths that we still need to update (there may be others):
- https://github.com/opencollective/opencollective-api/blob/main/server/graphql/v1/types.js#L1257
- https://github.com/opencollective/opencollective-api/blob/87629f67ba47df6fad7dcb478f01a6fc9d803c51/server/models/Member.js#L133
- https://github.com/opencollective/opencollective-api/blob/87629f67ba47df6fad7dcb478f01a6fc9d803c51/server/models/Tier.js#L302
- https://github.com/opencollective/opencollective-api/blob/87629f67ba47df6fad7dcb478f01a6fc9d803c51/server/models/Tier.js#L275

server/graphql/v1/types.js
```diff
@@ -1262,6 +1262,8 @@ export const TierStatsType = new GraphQLObjectType({
             return req.loaders.Tier.totalMonthlyDonations.load(tier.id);
           } else if (tier.interval === intervals.YEAR) {
             return req.loaders.Tier.totalYearlyDonations.load(tier.id);
+          } else if (tier.interval === intervals.FLEXIBLE) {
+            return req.loaders.Tier.totalDonated.load(tier.id);
```
Betree, line 1266: According to the description, this does not seem like a correct definition:
> How much money is given for this tier for each tier.interval (monthly/yearly)

This will return the total amount donated for this tier, which is fundamentally different from the sum of the total active recurring contributions. To solve that, an option is to introduce a `totalRecurringDonations` loader that will do:
> totalMonthlyContributions + (totalYearlyContributions / 12)

We can then update the field's description to:
```diff
- How much money is given for this tier for each tier.interval (monthly/yearly)
+ How much money is given for this tier for each tier.interval (monthly/yearly). For flexible tiers, this amount is a monthly average of contributions amount, taking into account both yearly and monthly subscriptions.
```

server/models/Member.js
```diff
@@ -143,7 +145,32 @@ function defineModel() {
...
+        }
+        if (order.interval === 'month' && days(transaction.createdAt) <= 31) {
+          return true;
+        }
+        if (order.interval === 'year' && days(transaction.createdAt) <= 365) {
+          return true;
+        }
+        return false;
+      });
+    });
```
Betree, line 173: This will break [the place](https://github.com/opencollective/opencollective-api/blob/483a5ab7f182d419fb3f87772c7af46c3c87e7e7/server/graphql/v1/queries.js#L805) where this is used since we're now returning a promise instead of a boolean without awaiting it. Let's keep that simple and treat flexible as yearly there: members are active when they have contributed in the past 365 days.

snitin315: Much simpler 👍🏻

server/graphql/v1/types.js
snitin315: @Betree  Something like this for flexible - 

```
        SELECT SUM(total) from (
            SELECT o."TierId" AS "TierId", COALESCE(SUM(s."amount")/12, 0) AS "total"
            FROM "Orders" o
            INNER JOIN "Subscriptions" s ON o."SubscriptionId" = s.id
            WHERE "TierId" IN (?)
            AND s."isActive" = TRUE
            AND s."interval" = 'year'
            GROUP BY "TierId"
            UNION ALL
            SELECT o."TierId" AS "TierId", COALESCE(SUM(s."amount"), 0) AS "total"
            FROM "Orders" o
            INNER JOIN "Subscriptions" s ON o."SubscriptionId" = s.id
            WHERE "TierId" IN (?)
            AND s."isActive" = TRUE
            AND s."interval" = 'month'
            GROUP BY "TierId"
        );
```

snitin315: Added a new `totalRecurringDonations` loader 👍🏻

Betree (approved):

PR #5720, opened 2021-03-16
Attach Receipt to Paid Charge Expenses

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4067

### DIFF
--- a/migrations/20210316141000-add-missing-expense-type.js
+++ b/migrations/20210316141000-add-missing-expense-type.js
@@ -1,0 +1,11 @@
+'use strict';
+
+module.exports = {
+  up: async queryInterface => {
+    await queryInterface.sequelize.query(`ALTER TYPE "enum_ExpenseHistories_type" ADD VALUE 'CHARGE';`);
+  },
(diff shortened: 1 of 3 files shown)


Betree (approved):
Looking good :+1:

Betree:
@kewitz no emergency since the feature is not public yet, but I just realized we will need to update the code for tax forms to make sure charges are not counted in the calculation:
- https://github.com/opencollective/opencollective-api/blob/f9fc75b1e1885767e836a7259cc1c6dfb69503e8/server/lib/queries.js#L895
- https://github.com/opencollective/opencollective-api/blob/f9fc75b1e1885767e836a7259cc1c6dfb69503e8/server/lib/queries.js#L936

kewitz:
@Betree could you take a quick look at the tax query refact? https://github.com/opencollective/opencollective-api/pull/5720/commits/ef727d8245b88d946d75ae5943d4b96f43e650d4?

Betree (approved):
Looking good!

PR #5723, opened 2021-03-17
Contributors: Optimize query

Betree:
Tested on https://opencollective.com/opencollective

### DIFF
--- a/server/lib/contributors.ts
+++ b/server/lib/contributors.ts
@@ -99,66 +99,42 @@ const loadContributors = async (collectiveId: number): Promise<ContributorsCache
 
   const allContributors = await sequelize.query(
     `
-     WITH member_collectives_matching_roles AS (
-      SELECT
-        c.*,
-        ARRAY_AGG(DISTINCT m."role") AS "roles",
-        MIN(m."since") as "since",
-        ARRAY_AGG(DISTINCT m."TierId") as "tiersIds",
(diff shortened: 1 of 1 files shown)


znarf (approved):
Looks good!

Betree:
This doesn't seem to work well with people having multiple memberships

Betree:
I had to update my previous query because each member row was affecting the `totalAmountDonated` (eg. users with 3 memberships had their total amount x 3).

This new query is actually a rollback of https://github.com/opencollective/opencollective-api/pull/5023 since we basically go back to a subquery for transactions but an `EXPLAIN ANALYZE` seems to suggest that the first version of this query was 100x faster.

Updated query complexity (~0.226 ms)
> Sort  (cost=11667117.51..11667120.65 rows=6286 width=263) (actual time=225.098..226.041 rows=9380 loops=1)

kewitz (approved):
I'm surprised to see that subquery is more performant in this case but I trust the analysis.
Let's keep an eye open for timeouts on the collective page.

Betree:
> I'm surprised to see that subquery is more performant in this case but I trust the analysis.
> Let's keep an eye open for timeouts on the collective page.

@kewitz  According to the `EXPLAIN ANALYZE`, the expensive part was the join between `member_collectives_matching_roles` and `total_contributed`. I haven't found a way to optimize it, but there might be something to do on that level:
image

Long term solution IMO: https://github.com/opencollective/opencollective/issues/3941

PR #5724, opened 2021-03-17
Mark expense as Spam

znarf:
Frontend: https://github.com/opencollective/opencollective-frontend/pull/6065

As demoed today:
- Introduce a new SPAM status and new MARK_AS_SPAM action
- Hide from the expenses lists, a bit like DRAFT
- Send a notification to the abuse channel on slack
- Prevent user to submit more expenses

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -11,6 +11,7 @@ import { getFxRate } from '../../lib/currency';
 import { floatAmountToCents } from '../../lib/math';
 import * as libPayments from '../../lib/payments';
 import { handleTransferwisePayoutsLimit } from '../../lib/plans';
+import { notifyTeamAboutSpamExpense } from '../../lib/spam';
 import { createFromPaidExpense as createTransactionFromPaidExpense } from '../../lib/transactions';
 import {
   handleTwoFactorAuthenticationPayoutLimit,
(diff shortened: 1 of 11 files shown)


server/graphql/common/expenses.ts
```diff
@@ -342,6 +356,31 @@ export const rejectExpense = async (
...
+  }
+
+  const updatedExpense = await expense.update({ status: expenseStatus.SPAM, lastEditedById: req.remoteUser.id });
+
+  // Limit the user so they can't submit expenses in the future
+  const submittedByUser = await updatedExpense.getSubmitterUser();
+  await submittedByUser.limitFeature(FEATURE.USE_EXPENSES);
+
+  // We create the activity as a good practice but there is no email sent right now
+  const activity = await expense.createActivity(activities.COLLECTIVE_EXPENSE_MARKED_AS_SPAM, req.remoteUser);
```
Betree, line 376: We may want to add this activity in https://github.com/opencollective/opencollective-api/blob/1d321f312479e98cfd99dcfd34d20223fe31fdeb/server/graphql/loaders/expenses.ts#L61 to surface it in the interface

server/graphql/common/expenses.ts
```diff
@@ -251,6 +252,19 @@ export const canReject = async (req: express.Request, expense: typeof models.Exp
...
+/**
+ * Returns true if expense can be rejected by user
+ */
+export const canMarkAsSpam = async (req: express.Request, expense: typeof models.Expense): Promise<boolean> => {
+  if (![expenseStatus.REJECTED].includes(expense.status)) {
+    return false;
+  } else if (!canUseFeature(req.remoteUser, FEATURE.USE_EXPENSES)) {
+    return false;
+  } else {
+    return remoteUserMeetsOneCondition(req, expense, [isHostAdmin]);
```
Betree, line 264: We do want to enable that for collective admins no?

```suggestion
    return remoteUserMeetsOneCondition(req, expense, [isHostAdmin, isCollectiveAdmin]);
```

Betree (approved):
Works for me, code looks good. As proposed in https://github.com/opencollective/opencollective/issues/3588#issuecomment-765233255, I think that a generic `reportAbuse` mutation would have been welcome since we already know that we want to add the same kind of functionality for other entities. But it also makes sense to put it in `processExpense` since we're changing its status.

We should add these new statuses to `getTaxFormsRequiredForExpenses` and `getTaxFormsRequiredForAccounts` to make sure SPAM expenses won't be included in the tax form job.

server/graphql/common/expenses.ts
znarf: This morning, we decided to start with host admins only.

PR #5726, opened 2021-03-17
(Transfer)Wise QuoteV2 Refactor

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4008
Resolves https://github.com/opencollective/opencollective/issues/3965

### DIFF
--- a/server/lib/transferwise.ts
+++ b/server/lib/transferwise.ts
@@ -16,7 +16,7 @@ import {
   BorderlessAccount,
   CurrencyPair,
   Profile,
-  Quote,
+  QuoteV2,
   RecipientAccount,
   Transfer,
   Webhook,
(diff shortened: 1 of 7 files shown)


patcon:
Thanks so much @kewitz and OC team 🙌 

Is this now immediately on prod? (If so, I will happily drop my mentioning it as caveat during onboarding of collectives :)

cc @benhylau

server/lib/transferwise.ts
```diff
@@ -95,10 +100,11 @@ export const requestDataAndThrowParsedError = (
   },
   defaultErrorMessage?: string,
 ): Promise<any> => {
-  debug(`calling ${url}`);
+  debug(`calling ${url}: ${JSON.stringify({ data, params: options.params }, null, 2)}`);
   const pRequest = data ? fn(url, data, options) : fn(url, options);
   return pRequest
     .then(getData)
+    .then(tap(data => debug(JSON.stringify(data, null, 2))))
```
Betree, line 107: That looks good!

Betree (commented):
Good improvement!

@patcon I think this has not been deployed in production yet

kewitz:
@patcon this is going to be deployed soon but we're currently having trouble with a non-related dependency build process.

PR #5750, opened 2021-03-23
fix: space buttons on paid expense email

bolariin:
resolves https://github.com/opencollective/opencollective/issues/4103

### Screenshots
Paid Expense email


Betree (commented):
How does it look like in the desktop view? Don't we end up with a vertical misalignment because one button has a margin-bottom and not the other?

bolariin:
The buttons look aligned on desktop. 

Paid email view on desktop

Betree (approved):
Alright, thank you @bolariin!

PR #5758, opened 2021-03-23
Store `creationRequest` for guest users

Betree:
User-agent header not added to v1's `createOrder` since guests are not supposed to be created from there


znarf (commented):
Looks good!

PR #5766, opened 2021-03-24
Update deleteUserCollective mutation

znarf:


Betree (commented):
Just a note to mention https://github.com/opencollective/opencollective/issues/2751, the `deleteUserCollective` and `deleteCollective` mutations are missing some important stuff like the memberships.

PR #5797, opened 2021-03-30
Update hostFeePercent and platformFeePercent in added funds to GraphQLFloat

znarf:


Betree (commented):
Maybe we should limit the number of decimals (eg. allow `7.42` but not `3.333333`)? No strong opinion on this

PR #5808, opened 2021-04-01
Contributors: Optimized query

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4121

Another take of the query I tried to implement in https://github.com/opencollective/opencollective-api/pull/5723/commits/f36e8f76bd5aeeaa2dc7b22eb1e34974eece67d8, except this one has 100% match on the results.

Complexity with `opencollective`:
- Before: 11798636
- After: 45489

### DIFF
--- a/server/lib/contributors.ts
+++ b/server/lib/contributors.ts
@@ -87,7 +87,7 @@ const storeContributorsInCache = (collectiveId: number, allContributors: Contrib
   return cacheEntry;
 };
 
-const contributorsQueryForLargeCollectives = `
+const contributorsQuery = `
   SELECT
     c.id,
     c."name",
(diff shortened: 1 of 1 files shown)


kewitz (commented):
Nice. Are we keeping a tab on the time spent in this query?

Betree:
> Nice. Are we keeping a tab on the time spent in this query?

Time spent mostly matches the complexity described in https://github.com/opencollective/opencollective-api/pull/5808#issue-607117849. Basically, we were in a situation where:
- For small collectives, the query I initially implemented was sometimes around ~5+ seconds, vs ~0.2s for yours
- For large collectives (mostly `opencollective`), the query you initially implemented took ~20s vs ~0.4s for the one I initially implemented (but with a huge complexity added)

And things were pretty random for mid-sized collectives. 

This one takes between 0.2s and 0.6s depending on the collectives. The complexity described here (`45489`) is for `opencollective`. All other collectives will end up with something much smaller.

PR #5823, opened 2021-04-05
Refact: Add VirtualCards model

kewitz:


migrations/20210401200001-migrate-virtual-cards-from-payout-methods.js
```diff
@@ -0,0 +1,66 @@
...
+      INNER JOIN "Collectives" c
+        ON pm."CollectiveId" = c.id
+      WHERE
+        pm."deletedAt" IS NULL
+        AND pm."type" = 'CREDIT_CARD';
+    `);
+
+    await queryInterface.sequelize.query(`
+      UPDATE "Expenses"
+        SET "VirtualCardId" = "PayoutMethods"."data"->>'token'
```
Betree, line 43: Shouldn't we also set `PayoutMethodId = NULL`?

server/graphql/v2/object/VirtualCard.ts
```diff
@@ -0,0 +1,20 @@
...
+import { GraphQLDateTime } from 'graphql-iso-date';
+import { GraphQLJSONObject } from 'graphql-type-json';
+
+export const VirtualCard = new GraphQLObjectType({
+  name: 'VirtualCard',
+  description: 'VirtualCard related properties.',
+  fields: () => ({
+    id: { type: GraphQLString },
+    CollectiveId: { type: GraphQLInt },
+    HostCollectiveId: { type: GraphQLInt },
```
Betree, line 11: That doesn't look like things we want to expense, we can eventually replace them by `account` + `host`

server/graphql/v2/object/VirtualCard.ts
```diff
@@ -0,0 +1,20 @@
...
+    name: { type: GraphQLString },
+    last4: { type: GraphQLString },
+    data: { type: GraphQLJSONObject },
+    privateData: { type: GraphQLJSONObject },
+    createdAt: { type: GraphQLDateTime },
+    updatedAt: { type: GraphQLDateTime },
+    deletedAt: { type: GraphQLDateTime },
```
Betree, line 18: Not sure if it's intentional, deleted entries are usually not returned

```suggestion
```

Betree (commented):
Looking good!

cron/hourly/reconcile-privacy-transactions.js
```diff
@@ -46,7 +41,7 @@ export async function run() {
 
       const { data: transactions } = await privacyLib.listTransactions(
         connectedAccount.token,
-        card.data.token,
+        card.id,
```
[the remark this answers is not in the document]
kewitz: Actually, this is not a security token, it is how Privacy identifies the credit card.
I also want to make this model compatible with Stripe, that's why I'm using it as an id.

migrations/20210401200001-migrate-virtual-cards-from-payout-methods.js
kewitz: Definitely!

server/graphql/v2/object/VirtualCard.ts
kewitz: Cool, I'll implement this refactor in the next PR where I actually use this.

PR #5838, opened 2021-04-08
getFxRate: Make sure currencies are uppercase

Betree:
This PR fixes one of the root cause for [this support ticket](https://opencollective.freshdesk.com/a/tickets/14258). 

In short, user payment was displayed as failed so they kept retrying, but the Stripe charge actually succeded. We failed to record the transaction because of two things:
1. Fixer API was down, so we could not fetch the conversion rate from there
2. Thew currency passed to `getFxRate` was coming from Stripe's `balanceTransaction` as lowercase (see [docs](https://stripe.com/docs/api/balance_transactions/object))

Because there was no Fixer API and fallback could not retrieve the fallback (it was expecting uppercase currencies) we failed to retrieve the FX Rate and crashed. This PR will fix most of the similar cases in the future if it happens again, and we'll address the underlying issue with https://github.com/opencollective/opencollective/issues/4174.

### DIFF
--- a/server/lib/currency.ts
+++ b/server/lib/currency.ts
@@ -130,6 +130,8 @@ export async function getFxRate(
   date: string | Date = 'latest',
 ): Promise<number> {
   debug(`getFxRate for ${date} ${fromCurrency} -> ${toCurrency}`);
+  fromCurrency = fromCurrency?.toUpperCase();
+  toCurrency = toCurrency?.toUpperCase();
 
   if (fromCurrency === toCurrency) {
     return 1;
--- a/server/paymentProviders/stripe/creditcard.js
+++ b/server/paymentProviders/stripe/creditcard.js
@@ -215,7 +215,7 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
       OrderId: order.id,
       amount: order.totalAmount,
       currency: order.currency,
-      hostCurrency: balanceTransaction.currency,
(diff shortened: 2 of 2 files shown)


server/lib/currency.ts
```diff
@@ -130,6 +130,8 @@ export async function getFxRate(
   date: string | Date = 'latest',
 ): Promise<number> {
   debug(`getFxRate for ${date} ${fromCurrency} -> ${toCurrency}`);
+  fromCurrency = fromCurrency?.toUpperCase();
+  toCurrency = toCurrency?.toUpperCase();
```
Betree, line 134: This is to make `getFxRate` more resilient in the future, in case there are other places where we pass it lowercase currencies.

server/paymentProviders/stripe/creditcard.js
```diff
@@ -215,7 +215,7 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
       OrderId: order.id,
       amount: order.totalAmount,
       currency: order.currency,
-      hostCurrency: balanceTransaction.currency,
+      hostCurrency: balanceTransaction.currency?.toUpperCase(),
```
Betree, line 218: This is the root fix, see https://stripe.com/docs/api/balance_transactions/object. In the DB these values were properly recorded with uppercase thanks to the model's setter.

kewitz: Though one. :dart:

kewitz (approved):

PR #5854, opened 2021-04-12
editNotifications: Multiple fixes & enhancements

Betree:


migrations/20210412140907-remove-notifications-unique-type.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = {
+  up: async queryInterface => {
+    await queryInterface.sequelize.query(`
+      DROP INDEX IF EXISTS "Notifications_type_CollectiveId_UserId";
```
Betree, line 6: Removed this index as it's perfectly valid to define multiple webhooks for the same event (eg. one for Discord, one for Slack, one for your website, etc).

server/graphql/v1/mutations/notifications.js
```diff
@@ -47,13 +47,15 @@ export async function editWebhooks(args, remoteUser) {
...
   if (toCreate.length > 0) {
     promises.push(
-      toCreate.map(notification =>
-        models.Notification.create({
-          ...pick(notification, allowedFields),
-          CollectiveId: args.collectiveId,
-          UserId: remoteUser.id,
-          channel: channels.WEBHOOK,
-        }),
+      Promise.all(
```
Betree, line 50: Returning a list of promise (rather than a promise) here made the creation async, resulting in the functions not returning all created webhooks

migrations/20210412140907-remove-notifications-unique-type.js
```diff
@@ -0,0 +1,15 @@
...
+    `);
+
+    await queryInterface.addIndex('Notifications', ['CollectiveId', 'type', 'channel']);
```
Betree, line 9: This index will be more performant: `CollectiveId` is that main way to filter notifications

PR #5855, opened 2021-04-12
PayPal: Automatic webhook management

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4096

This PR implements automatic PayPal webhooks management through two separate mechanisms:

1. **`scripts/update-hosts-paypal-webhooks.ts`**
The script can be triggered on the API to update the webhooks for all existing PayPal accounts. It will basically go through each one of them, and either create, link, or update the webhook to make sure events match what we expect.
If we ever need to support (or remove) an event in our webhook for PayPal, all we have to do is to run `npm run script scripts/update-hosts-paypal-webhooks.ts` and it will take care of updating all PayPal integrations.

2. **Automatically create the webhook when a PayPal account is linked**
See screencast below. https://github.com/opencollective/opencollective-frontend/pull/6167 will take care of removing the field in the frontend.

Peek 2021-04-13 17-29

### DIFF
--- a/scripts/update-hosts-paypal-webhooks.ts
+++ b/scripts/update-hosts-paypal-webhooks.ts
@@ -1,0 +1,43 @@
+/**
+ * This script can be used whenever PayPal webhooks event types change to update
+ * Host's connected accounts.
+ */
+
+import '../server/env';
+
+import logger from '../server/lib/logger';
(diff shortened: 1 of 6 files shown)


kewitz (approved):
Looking good! Thanks for covering Payouts!

PR #5873, opened 2021-04-14
PayPal: Add support for refunds

Betree:
Resolve https://github.com/opencollective/opencollective/issues/1656

Also forwards the refund reason when provided (ie. when rejecting)

image


kewitz (approved):
Looks good!
We will need to think about what happens if there are fees on top, but I think this can wait until the ledger refactor.

Betree:
@kewitz For fees on top, my understanding is that this will act like other payment methods except for Stripe (mainly, manual). In its current state, the function will create refund transactions for the platform tips. What we're missing is:
- Ignore transaction in the settlement if it has been refunded
- If a settled transaction is refunded, refund the platform tip in the next settlement

PR #5879, opened 2021-04-15
Fix assign virtual card to own host bug

kewitz:


scripts/privacy-add-credit-card.js
```diff
@@ -5,71 +5,33 @@ import '../server/env';
...
-  await models.PayoutMethod.create({
-    CollectiveId,
-    data: { token: cardToken, details },
-    name: card.last_four,
-    type: PayoutMethodTypes.CREDIT_CARD,
-  });
-
-  console.log(`Successfully added ${cardToken} to collective ${CollectiveId}!`);
+  await privacy.assignCardToCollective({ cardNumber, expireDate, cvv }, collective, host);
+  console.log('Done!');
```
SudharakaP, line 16: Extraneous log? 🤔 

```suggestion
```

kewitz: It's a script so it is OK to log stuff.

PR #5884, opened 2021-04-16
Introduce `kind` column in Transaction table

Betree:
Part of https://github.com/opencollective/opencollective/issues/4190

- [x] Set `kind` when creating transactions
- [x] Migrate existing transactions (when possible)
- [x] Validate `kind` in existing tests

### DIFF
--- a/cron/monthly/invoice-platform-fees.js
+++ b/cron/monthly/invoice-platform-fees.js
@@ -10,6 +10,7 @@ import { v4 as uuid } from 'uuid';
 import expenseStatus from '../../server/constants/expense_status';
 import expenseTypes from '../../server/constants/expense_type';
 import { SHARED_REVENUE_PLANS } from '../../server/constants/plans';
+import { TransactionKind } from '../../server/constants/transaction-kind';
 import { SETTLEMENT_EXPENSE_PROPERTIES, TransactionTypes } from '../../server/constants/transactions';
 import { uploadToS3 } from '../../server/lib/awsS3';
 import { generateKey } from '../../server/lib/encryption';
@@ -384,6 +385,7 @@ export async function run() {
           netAmountInCollectiveCurrency: totalAmountCredited,
           type: TransactionTypes.CREDIT,
           TransactionGroup: uuid(),
+          kind: TransactionKind.PLATFORM_TIP,
         });
       }
 
--- a/cron/monthly/invoice-platform-fees.js
+++ b/cron/monthly/invoice-platform-fees.js
@@ -10,7 +10,6 @@ import { v4 as uuid } from 'uuid';
 import expenseStatus from '../../server/constants/expense_status';
 import expenseTypes from '../../server/constants/expense_type';
 import { SHARED_REVENUE_PLANS } from '../../server/constants/plans';
-import { TransactionKind } from '../../server/constants/transaction-kind';
 import { SETTLEMENT_EXPENSE_PROPERTIES, TransactionTypes } from '../../server/constants/transactions';
 import { uploadToS3 } from '../../server/lib/awsS3';
 import { generateKey } from '../../server/lib/encryption';
@@ -385,7 +384,7 @@ export async function run() {
           netAmountInCollectiveCurrency: totalAmountCredited,
           type: TransactionTypes.CREDIT,
           TransactionGroup: uuid(),
-          kind: TransactionKind.PLATFORM_TIP,
+          kind: null, // Keeping this one null on purpose, see https://github.com/opencollective/opencollective-api/pull/5884#discussion_r616440055
         });
       }
 
--- a/migrations/20210416143836-add-kind-to-transactions.js
+++ b/migrations/20210416143836-add-kind-to-transactions.js
@@ -1,0 +1,63 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    // 1. Add the 'kind' column on Transactions table
+    await queryInterface.addColumn('Transactions', 'kind', {
+      allowNull: true,
+      type: Sequelize.ENUM(
+        'ADDED_FUNDS',
+        'CONTRIBUTION',
+        'EXPENSE',
+        'HOST_FEE',
+        'HOST_FEE_SHARE',
+        'PAYMENT_PROCESSOR_FEE',
+        'PLATFORM_FEE',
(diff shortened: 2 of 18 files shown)


server/lib/prepaid-budget.js
```diff
@@ -24,6 +24,7 @@ export async function createPrepaidPaymentMethod(originalCreditTransaction) {
     paymentProcessorFeeInHostCurrency: 0,
     platformFeeInHostCurrency: 0,
     hostFeeInHostCurrency: 0,
+    kind: null, // We intentionally set an empty `kind` for prepaid
```
Betree, line 27: @znarf I wonder whether we shouldn't set `ADDED_FUNDS` or create a new `kind` for prepaids? Unless we want to deprecate this approach

znarf: I thought about it and think we should have a dedicated `kind` for this. `PREPAID_PAYMENT_METHOD` ?

It's something that we want to continue supporting in the future, at least on the mid-term.

cron/monthly/invoice-platform-fees.js
```diff
@@ -384,6 +385,7 @@ export async function run() {
           netAmountInCollectiveCurrency: totalAmountCredited,
           type: TransactionTypes.CREDIT,
           TransactionGroup: uuid(),
+          kind: TransactionKind.PLATFORM_TIP,
```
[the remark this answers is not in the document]
znarf: I would keep it `null` as this should normally be deprecated in the process of our refactoring.

This will be replaced by having `kind=PLATFORM_TIP` `isDebt=true` transactions one by one. Instead of batching them at the end of the month.

migrations/20210416143836-add-kind-to-transactions.js
```diff
@@ -0,0 +1,74 @@
...
+    // CONTRIBUTION
+    // This will set KIND=CONTRIBUTION on all Transactions with an order, except for the ones
+    // we already migrated above (added funds) and prepaid budgets (tier #15445).
+    await queryInterface.sequelize.query(`
+      UPDATE "Transactions" t
+      SET "kind" = 'CONTRIBUTION'
+      FROM "Orders" o
+      WHERE t."OrderId" IS NOT NULL
+      AND t."OrderId" = o.id
+      AND o."TierId" != 15445
```
znarf, line 41: I think it's ok if the initial payment for a pre-paid budget is a CONTRIBUTION.

What should be `PREPAID_PAYMENT_METHOD` is the transaction to give back and this one should not have have an order id?

migrations/20210416143836-add-kind-to-transactions.js
```diff
@@ -0,0 +1,74 @@
...
+      SET "kind" = 'EXPENSE'
+      WHERE "ExpenseId" IS NOT NULL
+    `);
+
+    // PLATFORM_TIP
+    await queryInterface.sequelize.query(`
+      UPDATE "Transactions"
+      SET "kind" = 'PLATFORM_TIP'
+      WHERE ("FromCollectiveId" = 8686 OR "CollectiveId" = 8686)
+      AND ("data" ->> 'isFeesOnTop')::boolean = TRUE
```
znarf, line 67: I noticed that Transactions `388057` and `388056` are matching, maybe this is not accurate.

Can't we look at `PlatformTipForTransactionGroup` ?

migrations/20210416143836-add-kind-to-transactions.js
Betree: You're right. Actually, transactions created like that will have kin=CONTRIBUTION so we should not check that in the migration. `PREPAID_PAYMENT_METHOD` is properly set below and in `server/lib/prepaid-budget.js`

migrations/20210416143836-add-kind-to-transactions.js
Betree: `388057` and `388056` seem like invalid transactions that we should fix. The order (#73185) is directed to Open Collective directly, so there's no reason to have `isFeesOnTop` in there (unless I'm missing something). I did not find similar occurrences in the pledged.

I'm afraid that we can miss some transactions if we look at `PlatformTipForTransactionGroup`, the following query returns a lot of results:
```sql
SELECT * 
FROM "Transactions" t 
WHERE ("FromCollectiveId" = 8686 OR "CollectiveId" = 8686)
AND ("data" ->> 'isFeesOnTop')::boolean = TRUE
AND "PlatformTipForTransactionGroup" IS NULL
```

znarf: Seems `isFeesOnTop` means mainly that Platform Tips were attached to the original transaction. Should have been named `hasPlatformTips` and should not have been propagated to the Platform Tip transactions themselves.

Anyway, in this context I think this doing the job, except for the transaction that I highlighted.

znarf (approved):
It seems good to go. We will need to have a close look at production when do the migration.

migrations/20210416143836-add-kind-to-transactions.js
Betree: I've manually updated `388057` and `388056`, so the migration should be safe to run.

I'm still a bit concerned by the query above, there are basically two cases where the `PlatformTipForTransactionGroup` is `null` for valid platform tips:
- Refunds (not a big deal in this case)
- Transactions created before `2020-10-15 18:55:09` (see the query below). My guess is that https://github.com/opencollective/opencollective-api/blob/5654a97eb4417b80ebb90aca07931f0f8961678d/migrations/20201016085526-add-platform-contribution-transaction-group.js#L34 didn't catch everything. I suspect the join on the date to be the root cause for this. I initially proposed [another migration](https://github.com/opencollective/opencollective-api/pull/4716#pullrequestreview-512415868) joining on the payment providers payment IDs that seems to catch these cases. I'll try fixing that with https://github.com/opencollective/opencollective/issues/4205

```sql
SELECT *
FROM "Transactions" t 
INNER JOIN "Orders" o ON t."OrderId" = o.id
WHERE (t."FromCollectiveId" = 8686 OR t."CollectiveId" = 8686)
AND o."CollectiveId" != 8686
AND (t."data" ->> 'isFeesOnTop')::boolean = TRUE
AND t."PlatformTipForTransactionGroup" IS NULL
AND t."isRefund" IS NOT TRUE
ORDER BY t."createdAt" DESC
```

PR #5888, opened 2021-04-19
Add support for expense submission restriction

SudharakaP:
This adds support for restricting expense submission for collective members and contributors who previously submitted expenses.

Related to https://github.com/opencollective/opencollective/issues/4196

Related front end PR: https://github.com/opencollective/opencollective-frontend/pull/6215


server/lib/collectivelib.ts
```diff
@@ -48,6 +48,7 @@ export const getCollectiveAvatarUrl = (
 
 export const COLLECTIVE_SETTINGS_KEYS_LIST = [
   'apply',
+  'disableExpenseSubmission',
```
znarf, line 51: Perhaps `disablePublicExpenseSubmission`

Betree (commented):
If we really want to enforce this limitation, we should add a small check in the `createExpense` mutation: https://github.com/opencollective/opencollective-api/blob/d2128098f20c4b285e4a9e6628a9192c2cfa23a6/server/graphql/v2/mutation/ExpenseMutations.ts

server/lib/collectivelib.ts
SudharakaP: Done. 👍🏽

SudharakaP:
> If we really want to enforce this limitation, we should add a small check in the `createExpense` mutation: https://github.com/opencollective/opencollective-api/blob/d2128098f20c4b285e4a9e6628a9192c2cfa23a6/server/graphql/v2/mutation/ExpenseMutations.ts

@Betree : Good idea, I've added that as well; 😄 

image

server/graphql/common/expenses.ts
```diff
@@ -510,7 +510,7 @@ type ExpenseData = {
   payeeLocation?: Record<string, unknown>;
   items?: Record<string, unknown>[];
   attachedFiles?: Record<string, unknown>[];
-  collective?: Record<string, unknown>;
+  collective?: Record<string, any>;
```
Betree, line 513: Any reason for doing this change? Using `any` is usually discouraged and generates an ESLint warning. If it's about accessing properties, you can usually solve it by using brackets (eg. `settings?.['disablePublicExpenseSubmission']`) for accessing the property.

server/graphql/common/expenses.ts
```diff
@@ -531,6 +531,13 @@ export async function createExpense(
     throw new Unauthorized('Missing expense.collective.id');
   }
 
+  if (remoteUser.rolesByCollectiveId) {
```
Betree, line 534: Why wouldn't that be set? Why skipping the check if it's not?

server/graphql/common/expenses.ts
```diff
@@ -531,6 +531,13 @@ export async function createExpense(
...
+    const isMember = Object.keys(remoteUser.rolesByCollectiveId).includes(String(expenseData.collective.id));
```
Betree, line 535: :ice_cream: Since `rolesByCollectiveId` is an object, it sounds simpler to check the key directly:
```suggestion
    const isMember = Boolean(remoteUser.rolesByCollectiveId[String(expenseData.collective.id)]);
```

or with lodash
```suggestion
    const isMember = has(remoteUser.rolesByCollectiveId, String(expenseData.collective.id));
```

server/graphql/common/expenses.ts
SudharakaP: @Betree : I did this change to bypass the typescript CI error we are getting; 

`server/graphql/common/expenses.ts:536:42 - error TS2339: Property 'disablePublicExpenseSubmission' does not exist on type 'unknown'.`

Let me think of an alternative approach. 🤔

server/graphql/common/expenses.ts
SudharakaP: Thanks. Done. 👍🏽

server/graphql/common/expenses.ts
SudharakaP: @Betree : I just removed this condition. 😄 

I just added this check for good measure, in case if it isn't set we skip gracefully without throwing a null pointer in the next line. But even for a user who doesn't have any roles in any collective it seems to me that `rolesByCollectiveId` returns an empty object. So I think it's okay to remove it. 👍🏽

server/graphql/common/expenses.ts
Betree: Try with brackets, as mentioned above that should work: 

```es6
settings?.['disablePublicExpenseSubmission']
```

server/graphql/common/expenses.ts
SudharakaP: @Betree : Actually I spoke too soon, it seems that when we remove it some tests are failing; https://github.com/opencollective/opencollective-api/runs/2401851404?check_suite_focus=true. 🤔

server/graphql/common/expenses.ts
SudharakaP: @Betree : Ops sorry, I somehow didn't try that. Yes it works, thanks. 👍🏽

server/graphql/common/expenses.ts
Betree: `rolesByCollectiveId` is not supposed to be `null` here, as it's normally populated before entering any query or mutation. We can eventually add an `await remoteUser?.populateRoles()` in `makeRequest` if some tests are inconsistent on this regard: https://github.com/opencollective/opencollective-api/blob/2d46ef85b6894c690704fcc7b32e5527f34253f3/test/utils.js#L62

SudharakaP: @Betree : Done. It was the graphql tests that were failing and I had to add the `populateRoles()` in two places to make it work. 😄

Betree (approved):
Looks good :+1:

SudharakaP:
> Looks good +1

Thanks. 😄

PR #5895, opened 2021-04-20
Add support for running workflows without restriction for dependabot

SudharakaP:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    branches:
-      - main
-  pull_request:
+    branches-ignore:
+      - 'dependabot/**'
+  pull_request_target:
     types:
       - opened
       - synchronize
@@ -214,6 +214,8 @@ jobs:
         run: npm run build
 
   test:
+    if: >-
+      github.event_name == 'push' || github.actor == 'dependabot[bot]'
     runs-on: ubuntu-latest
 
     timeout-minutes: 30
@@ -352,6 +354,8 @@ jobs:
         run: ./scripts/check-schemas.sh
 
   graphql-inspector:
+    if: >-
+      github.event_name == 'push' || github.actor == 'dependabot[bot]'
     needs: schema-update
     name: GraphQL Inspector
     runs-on: ubuntu-latest
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,6 +1,14 @@
 name: E2E
 
-on: [push]
+on:
+  push:
+    branches-ignore:
+      - 'dependabot/**'
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
 
 env:
   CI: true
@@ -29,6 +37,8 @@ env:
 
 jobs:
   e2e:
+    if: >-
+      github.event_name == 'push' || github.actor == 'dependabot[bot]'
     runs-on: ubuntu-latest
 
     timeout-minutes: 30


Betree:
@SudharakaP It seems that the tests are not running on this PR?

znarf (commented):
Let's check why tests are not running before merging?

SudharakaP:
@znarf @Betree : Ah, think I figured it out. It seems that using both `pull_request` and `pull_request_target` is having some problems. I've corrected that. It should be good now I think. Let me know if you see any issues. 😄

Betree (approved):
Looks good to me, we can try :+1:

tsimbalar:
⚠️ sorry, got to this PR by reading conversation in https://github.com/dependabot/dependabot-core/issues/3253 ... 

While your GitHub actions here work, they possibly don't do what you expect ⚠️  . 

WIth the `pull_request_target` event, the workflow run "in the context of the base branch" (i.e. the target branch where we'll want to merge eventually). 

This means that the `checkout` action will by default checkout a commit ... from `master` .... so for dependabot PRs your checks will run against master, therefore not validating whether updates of dependabot break anything. 

You probably need to do something like this : 

```yaml
      - name: Checkout
        uses: actions/checkout@v2-beta
        with:
          # we are in a pull_request_target event - we need to explicitly checkout the PR's head. Otherwise we would be running against main branch
          ref: ${{ github.event.pull_request.head.sha }}
```

instead of 

```yaml
      - name: Checkout
        uses: actions/checkout@v2-beta
```

... but then this will probably break your actions when triggered by a `push`, as  `github.event.pull_request.head.sha` won't be available 😭

SudharakaP:
@tsimbalar : Thanks much for reporting this issue; something I've overlooked completely. 😞 I've added three PRs (linked above this comment) where we branch out in the checkout action depending on whether it's a push or pull request target event. Let us know if you see any problems with the new changes. 😄

tsimbalar:
@SudharakaP glad I could help :) 

FYI this is exactly the pattern (or anti-pattern) mentioned at the end of the article https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ 

> We have also noticed a pattern that has a vulnerable intent of use, however due to misunderstanding of pull_request_target ends up being a broken, but not vulnerable, workflow. In these cases, repositories switched to using the pull_request_target trigger, but use the default values for the actions/checkout action, for example:
> ```
> # The workflow is broken. DO NOT use it in production.
> on: [push, pull_request_target]
> 
> jobs:
>   build:
>     runs-on: ubuntu-latest
>     steps:
>       - name: Checkout
>         uses: actions/checkout@v2
>       - name: Build and test
>         run: /bin/bash ./build.sh && /bin/bash ./runtests.sh
>       - name: Report
>         if: failure() && github.event.action != 'push'
>         with: fancy/commenter@v1
>           message: |
>             Some checks have failed.
> ```
> This doesn’t do what the authors most likely intended. In case of a PR, it builds the latest changeset from the target repository. The workflow is broken in the sense that it does not actually build the PR, but luckily is not vulnerable, since there is no explicit checkout of the actual PR contents or unsafe handling of those contents.

SudharakaP:
> @SudharakaP glad I could help :)
> 
> FYI this is exactly the pattern (or anti-pattern) mentioned at the end of the article https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
> 
> > We have also noticed a pattern that has a vulnerable intent of use, however due to misunderstanding of pull_request_target ends up being a broken, but not vulnerable, workflow. In these cases, repositories switched to using the pull_request_target trigger, but use the default values for the actions/checkout action, for example:
> > ```
> > # The workflow is broken. DO NOT use it in production.
> > on: [push, pull_request_target]
> > 
> > jobs:
> >   build:
> >     runs-on: ubuntu-latest
> >     steps:
> >       - name: Checkout
> >         uses: actions/checkout@v2
> >       - name: Build and test
> >         run: /bin/bash ./build.sh && /bin/bash ./runtests.sh
> >       - name: Report
> >         if: failure() && github.event.action != 'push'
> >         with: fancy/commenter@v1
> >           message: |
> >             Some checks have failed.
> > ```
> > 
> > 
> > This doesn’t do what the authors most likely intended. In case of a PR, it builds the latest changeset from the target repository. The workflow is broken in the sense that it does not actually build the PR, but luckily is not vulnerable, since there is no explicit checkout of the actual PR contents or unsafe handling of those contents.

@tsimbalar : Yes, exactly; I somehow missed that part. But I think the three new PRs above handles that as we now use checkout action based on either push or pull request target. Thanks again for your help. 😄

tsimbalar:
Looked good to me 👍

PR #5896, opened 2021-04-20
Virtual Card: Options, Policy and Request a Card

kewitz:


server/graphql/common/expenses.ts
```diff
@@ -676,14 +677,22 @@ export async function editExpense(
...
     throw new Unauthorized("You don't have permission to edit this expense");
   }
 
-  const isPaidCreditCardCharge = expense.type === expenseType.CHARGE && expense.status === expenseStatus.PAID;
+  const isPaidCreditCardCharge =
+    expense.type === expenseType.CHARGE && expense.status === expenseStatus.PAID && Boolean(expense.VirtualCardId);
+
+  if (isPaidCreditCardCharge && !hasItemChanges) {
+    throw new ValidationFailed(
+      'You need to include Expense Items when adding missing information to card charge expenes',
```
SudharakaP, line 693:
```suggestion
      'You need to include Expense Items when adding missing information to card charge expenses',
```

server/graphql/common/expenses.ts
```diff
@@ -769,21 +778,30 @@ export async function editExpense(
...
+      status: shouldUpdateStatus ? 'PENDING' : expense.status,
+      FromCollectiveId: fromCollective.id,
+      PayoutMethodId: PayoutMethodId,
+      legacyPayoutMethod: models.Expense.getLegacyPayoutMethodTypeFromPayoutMethod(payoutMethod),
+      tags: cleanExpenseData.tags,
+    };
+    if (isPaidCreditCardCharge) {
+      updatedExpenseProps['data'] = { ...expense.data, missingDetails: false };
+    }
+    return expense.update(updatedExpenseProps, { transaction: t });
```
SudharakaP, line 794: Maybe rename, `t` to transaction and just put, 🤔 

```suggestion
    return expense.update(updatedExpenseProps, { transaction });
```

kewitz: The problem here is that "Transaction" may suggest something very different given the context.

SudharakaP: Ah, I see; then we could leave it I guess. Just that looks odd to have a one letter argument. 😸 😸

PR #5899, opened 2021-04-21
fix(migrations/uppercase-vendor-slugs): Add empty down method

Betree:
Migrations must have a `down` key, otherwise rolling back produces a crash. We should probably add an ESLint rule for that.

```bash
➜  opencollective-api git:(main) npm run db:migrate:undo 
== 20210420162000-uppercase-vendor-slugs: reverting =======

ERROR: Could not find migration method: down
```


kewitz:
Ouch, it was removed because lint complained about the empty function.

PR #5900, opened 2021-04-21
PlatformTips: Deprecate `PlatformTipForTransactionGroup`

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4205

- [x] Migrate platform tips without `PlatformTipForTransactionGroup`
- [x] Add tests to make sure new platform tips have the same `TransactionGroup` than their original transactions
- [x] Migrate all existing platform tips to make sure their `TransactionGroup` are the same as the transactions they're referring to
- [x] Update all queries using `PlatformTipForTransactionGroup`
- [x] Create index on `TransactionGroup`
- [x] Remove the `PlatformTipForTransactionGroup` column

PR #5902, opened 2021-04-21
Addition of `if` conditions to make sure ci runs are not duplicated

SudharakaP:
The CI now runs fine but when doing pull requests for dependabot after, https://github.com/opencollective/opencollective-api/pull/5895 but there's a possibility of duplication of the CI runs (example: https://github.com/opencollective/opencollective-api/pull/5901). This PR takes care of that. 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,8 @@ env:
 
 jobs:
   lint:
+    if: >-
+      github.event_name == 'push' || github.actor == 'dependabot[bot]'
     runs-on: ubuntu-latest
 
     timeout-minutes: 15
@@ -59,6 +61,8 @@ jobs:
       - run: npm run lint:check
 
   prettier:
+    if: >-
+      github.event_name == 'push' || github.actor == 'dependabot[bot]'
     runs-on: ubuntu-latest
 
     timeout-minutes: 15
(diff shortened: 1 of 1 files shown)


znarf (approved):
Looks good.

Betree (approved):

Betree (commented):
Lint job still seems duplicated on this PR

znarf:
Let's check the duplication.

SudharakaP:
@Betree : My understanding is that we cannot just test a github action flow with a PR like this. Because I think GitHub actions defined on the PR doesn't override the ones in the repo. Only after merging it starts overriding it. 🤔

znarf:
Let's give it a try then.

Betree:
It seems to work fine after a rebase: https://github.com/opencollective/opencollective-api/pull/5900. Having all these "skipped" tests is a bit messy, but that's no blocker to me if it's difficult to solve.

SudharakaP:
> It seems to work fine after a rebase: #5900. Having all these "skipped" tests is a bit messy, but that's no blocker to me if it's difficult to solve.

@Betree : Tbh, this is a temporary solution in my opinion. I feel there's a lot of updates being rolled out on this regard; https://github.com/dependabot/dependabot-core/issues/3253 so we'll keep a tab on that and update if and when a better method is available. Maybe they'll add an exception to dependabot PRs. 😉

PR #5908, opened 2021-04-22
Implement TransactionSettlements

Betree:
Part of https://github.com/opencollective/opencollective/issues/4077

This PR implements the base system for settlements, without making any change to the way transactions are generated today (in other words, it's not active yet with real data). Tests in `test/server/models/TransactionSettlement.test.ts` are recreating the desired environment and are there to assert that the settlement system works, but this new system won't be active until we switch to the new data model proposed in https://github.com/opencollective/opencollective/issues/4124 where we actually have two transactions pairs for platform tips (one from the contributor to the host, and the other as a debt from the host to Open Collective).

It's also the first use case for ledger snapshots. [This file](https://github.com/opencollective/opencollective-api/blob/feat/transactions-settlement-table/test/server/models/TransactionSettlement.test.ts.snap) will show you the evolution of the ledger after calling the different methods. It's intended to match what's described in https://github.com/opencollective/opencollective/issues/4124.

Next steps after we merge this PR:
- Record platform tips with the debt transaction
- Update `cron/monthly/invoice-platform-fees.js` with this new settlement mechanism

### DIFF
--- a/server/models/TransactionSettlement.ts
+++ b/server/models/TransactionSettlement.ts
@@ -1,0 +1,188 @@
+import { groupBy } from 'lodash';
+
+import { TransactionKind } from '../constants/transaction-kind';
+import restoreSequelizeAttributesOnClass from '../lib/restore-sequelize-attributes-on-class';
+import sequelize, { DataTypes, Model, Op } from '../lib/sequelize';
+
+import Collective from './Collective';
+import Transaction from './Transaction';
+
+export enum TransactionSettlementStatus {
+  OWED = 'OWED',
+  INVOICED = 'INVOICED',
+  SETTLED = 'SETTLED',
+}
+
+interface TransactionSettlementAttributes {
+  TransactionGroup: string;
+  kind: TransactionKind;
+  status: TransactionSettlementStatus;
+  ExpenseId: number;
+  createdAt: Date;
+  updatedAt: Date;
+  deletedAt: Date;
+}
+
+interface TransactionSettlementCreateAttributes {
+  TransactionGroup: string;
+  kind: TransactionKind;
+  status: TransactionSettlementStatus;
+  ExpenseId?: number;
+}
+
+class TransactionSettlement
+  extends Model<TransactionSettlementAttributes, TransactionSettlementCreateAttributes>
+  implements TransactionSettlementAttributes {
+  TransactionGroup: string;
+  kind: TransactionKind;
+  status: TransactionSettlementStatus;
+  ExpenseId: number;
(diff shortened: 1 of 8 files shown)


test/server/models/TransactionSettlement.test.ts
```diff
@@ -0,0 +1,201 @@
...
+  // DEBIT Tip transaction (contributor -> OC)
+  await fakeTransaction({
+    TransactionGroup: transaction.TransactionGroup,
+    kind: TransactionKind.PLATFORM_TIP,
+    type: 'DEBIT',
+    amount: -200,
+    CollectiveId: transaction.FromCollectiveId,
+    FromCollectiveId: FEES_ON_TOP_TRANSACTION_PROPERTIES.CollectiveId,
+    HostCollectiveId: null,
+    description: 'Tip transaction form the contributor to Open Collective',
```
znarf, line 66: Replace form by from here and everywhere else?

znarf (approved):
Looking good and a solid foundation for the model.

I think I might have more feedback when I'll see how it's used in the codebase.

test/server/models/TransactionSettlement.test.ts
Betree: Good catch!

server/models/TransactionSettlement.ts
```diff
@@ -0,0 +1,188 @@
...
+      `
+        SELECT t.*, ts.status as "settlementStatus"
+        FROM "Transactions" t
+        INNER JOIN "TransactionSettlements" ts
+          ON t."TransactionGroup" = ts."TransactionGroup"
+          AND t."kind" = ts."kind"
+        WHERE t."type" = 'CREDIT'
+        AND t."isDebt" IS TRUE
+        AND t."deletedAt" IS NULL
+        AND ts."deletedAt" IS NULL
```
kewitz, line 86: Missing `hostId`?

test/test-helpers/fake-data.js
```diff
@@ -29,6 +29,14 @@ export const randArray = (generateFunc, min = 1, max = 1) => {
   return [...Array(arrayLength)].map((_, idx) => generateFunc(idx, arrayLength));
 };
 
+/**
+ * Allows to generate an UUID with the first 8 characters hardcoded. This is useful to generate
+ * random but identifiable valid UUIDs.
+ */
+export const fakeUUID = firstHeightChars => {
+  return `${firstHeightChars}-${uuid().substr(9)}`;
```
kewitz, line 37: `firstEightChars`.

PR #5913, opened 2021-04-23
Add snapshot testing for the ledger

Betree:
Base helper for using snapshot testing for the ledger. Will later introduce more specific ones that pre-filter the columns, eg:

```es6
snapshotTransactions(transactions, { view: 'amounts' }) // To get all amounts & currency rates
snapshotTransactions(transactions, { view: 'info' }) // Get transaction.info
snapshotTransactions(transactions, { view: 'platform-tips' })
...
```

See https://github.com/opencollective/opencollective-api/blob/test/transactions-snapshot-testing/test/server/models/Transaction.test.js.snap for the result.


znarf:
Looking good!

PR #5916, opened 2021-04-23
SanitizeHtml: Allow figure with content-type for iframes

Betree:
For https://github.com/opencollective/opencollective-frontend/pull/6042

### DIFF
--- a/server/lib/sanitize-html.ts
+++ b/server/lib/sanitize-html.ts
@@ -1,5 +1,5 @@
 import config from 'config';
-import { truncate } from 'lodash';
+import { truncate, uniq } from 'lodash';
 import prependHttp from 'prepend-http';
 import LibSanitize from 'sanitize-html';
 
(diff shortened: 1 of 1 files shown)


SudharakaP:
@Betree : Thanks for adding this. But something not working correctly before adding this `figure` for content types? 🤔

Betree:
@SudharakaP responsiveness. Before that, the height was fixed to 400px so on mobile you ended up with two black stripes above and below the video. Having this content-type allowed us to add [this CSS rule](https://github.com/opencollective/opencollective-frontend/pull/6042/commits/077ef4060b50ef57c3af8fba2de3dcac55970f28#diff-7c66bed1a72345f8e57ccf8d28acc894e870d12efc4045b9ab242b0a386b064cR110) which creates a real 16:9 container.

Betree:
@SudharakaP 

Before:
image

After:
image

SudharakaP:
> @SudharakaP responsiveness. Before that, the height was fixed to 400px so on mobile you ended up with two black stripes above and below the video. Having this content-type allowed us to add [this CSS rule](https://github.com/opencollective/opencollective-frontend/pull/6042/commits/077ef4060b50ef57c3af8fba2de3dcac55970f28#diff-7c66bed1a72345f8e57ccf8d28acc894e870d12efc4045b9ab242b0a386b064cR110) which creates a real 16:9 container.

@Betree : Thanks. Looks wonderful now. 😄

PR #5919, opened 2021-04-23
Point to the correct branch in the checkout action when using dependabot

SudharakaP:

Resolves the issue mentioned in, https://github.com/opencollective/opencollective-api/pull/5895#issuecomment-825621778

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -102,7 +102,7 @@ jobs:
         with:
           # We are in a pull_request_target event - we need to explicitly checkout the PR's head.
           # Otherwise we would be running against main branch
-          ref: ${{ github.event.pull_request.head.sha }}
+          ref: ${{ github.event.pull_request.head.ref }}
 
       - name: Set REF in env, removing the `refs/` part (Push)
         if: github.event_name == 'push'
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yml
```diff
@@ -93,9 +93,17 @@ jobs:
...
+        uses: actions/checkout@v2
 
-      - name: Checkout (api)
+      - name: Checkout (Pull Request Target)
+        if: github.event_name == 'pull_request_target'
         uses: actions/checkout@v2
+        with:
+          # We are in a pull_request_target event - we need to explicitly checkout the PR's head.
+          # Otherwise we would be running against main branch
+          ref: ${{ github.event.pull_request.head.sha }}
```
znarf, line 106: What about frontend matching branches lines 111 to 136?

Is this still gonna work properly with this change?

SudharakaP: @znarf : Yes, matching branch is explicitly set (https://github.com/opencollective/opencollective-frontend/blob/6a6b6a90da12ca43e180f49811d4ee03236f7f63/.github/workflows/e2e.yml#L107). Don't think we should change that. It's only when it's not explicitly set that it defaults to the main branch. 😄

PR #5926, opened 2021-04-27
GraphQL: Don't cache results if there is an error

Betree:
This morning, a migration that removed some columns caused requests made to the collective page to fail. This was resolved as soon as the deployment completed, but failed requests were kept in the cache and had to be cleaned manually.

It seems like a good practice to not cache responses if there's an error:
1. To have a faster refresh when the root error gets fixed
2. To prevent one-time errors from being sent for all future requests

The main downside for this can be the server load if we keep fetching a page that is erroring with an expensive query.

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -160,7 +160,7 @@ export default app => {
     formatResponse: (response, ctx) => {
       const req = ctx.context;
 
-      if (req.cacheKey) {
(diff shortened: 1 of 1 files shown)


znarf (approved):
I think it's worth to have a try.

PR #5961, opened 2021-05-03
updateOrder: Check order not found

Betree:
Resolve https://sentry.io/organizations/open-collective/issues/2374987288/

### DIFF
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -210,7 +210,9 @@ const orderMutations = {
         ],
       });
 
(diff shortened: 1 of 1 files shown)


znarf (approved):
Looks good. Do you have an idea how this could happen?

Betree:
@znarf I'm not yet 100% sure about the root cause, but it's for sure related to https://opencollective.slack.com/archives/GRN3WA3BM/p1618932868011200, the payment method has a `NULL` `CollectiveId` and therefore we fail to check the permissions for it.

I'm investigating that in a separate PR.

PR #5964, opened 2021-05-03
PaymentMethod: Always store collective id

Betree:
Experiment on reverting a change from https://github.com/opencollective/opencollective-api/pull/2428, where we decided to not store the `CollectiveId` for payment methods. This behavior is causing some complications like https://sentry.io/organizations/open-collective/issues/2374987288 and https://opencollective.slack.com/archives/GRN3WA3BM/p1618932868011200 - basically most of the places where we check payment methods permissions are not ready to handle this and will reject users that should be allowed.

I did not find much context about why the change was initially done, but it's possibly related to Stripe CSA / 3D secure.

### DIFF
--- a/migrations/20210503103943-fill-payment-methods-collective-id.js
+++ b/migrations/20210503103943-fill-payment-methods-collective-id.js
@@ -1,0 +1,35 @@
+'use strict';
+
+module.exports = {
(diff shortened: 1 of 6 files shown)


znarf (approved):
Looks good!

PR #5965, opened 2021-05-03
Remove Braintree

Betree:
Related to https://github.com/opencollective/opencollective/issues/3986
:warning: https://github.com/opencollective/opencollective-frontend/pull/6296 should be merged & deployed first.

We ended up going with PayPal implementation instead.


znarf:
Looks good. To be rebased and small lint issue to fix.

PR #5966, opened 2021-05-03
Transactions: Add debts for tip transactions

Betree:
Part of https://github.com/opencollective/opencollective/issues/4124

- ❌ Create tips in SQL transactions => see https://github.com/opencollective/opencollective-api/pull/5981
- [x] Create debt transaction
- [x] Create TransactionSettlement entry
- [x] Handle refunds
- [x] Migrate all platform tips created in May
- [x] Update host invoice script


cron/monthly/invoice-platform-fees.js
```diff
@@ -338,13 +342,8 @@ export async function run() {
...
     const transactionIds = transactions.map(t => t.TransactionId);
-    const totalAmountCredited = sumBy(
-      items
-        .filter(i => i.description != 'Shared Revenue')
-        .filter(i => i.description != 'Reimburse: Payment Processor Fee for collected Platform Tips')
-        .filter(i => i.description != 'Fixed Fee per Hosted Collective'),
-      'amount',
-    );
+    const itemsWithoutDebt = items.filter(i => i.description === 'Platform Fees');
+    const totalAmountCredited = sumBy(itemsWithoutDebt, 'amount');
```
znarf, line 346: It's a general comment about this file. I would like us to stop doing custom queries in SQL and use instead some helpers in the model or in separated library.

The goal would be to have a single source of truth, and share metrics with:
https://github.com/opencollective/opencollective-api/blob/main/server/models/Collective.js#L2946

Betree:
Some general considerations from yesterday's call:
- We may want to add `kind` to the CSV exports
- Better descriptions for platform tips
- Filter out debts on frontend by default?
- Ability to filter by `kind` in the frontend
- When running the `invoice-platform-fees` script early June, we shouldn't consider June transactions

Betree:
I moved the migration to a script, and added a `check` command to make sure the data is consistent.

Betree:
Renamed script to `cron/monthly/host-settlement.js`. Also made a copy of the legacy script that can only be run in dry mode to `scripts/legacy-host-settlement.js` to easily compare the results and make sure they're matching.

znarf (approved):

PR #5968, opened 2021-05-04
Privacy reconciliation worker bugs in the backend

kewitz:


server/paymentProviders/privacy/index.ts
```diff
@@ -44,7 +44,7 @@ const createExpense = async (
   const host = opts?.host || (await models.Collective.findByPk(virtualCard.HostCollectiveId));
   const hostCurrencyFxRate = opts?.hostCurrencyFxRate || (await getFxRate('USD', host.currency));
   const amount = privacyTransaction.settled_amount;
-  const UserId = collective.CreatedByUserId;
```
znarf, line 47: As improvement, let's assign an Individual account to the card (from the Collective admins) when we attach it to the Collective.

znarf (approved):

PR #5976, opened 2021-05-04
Fix Invoice worker $0 reimburse bug

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4248

This also fixes the frontend total amount issue.
Screenshot from 2021-05-04 16-27-47


znarf:
@kewitz Could you check and edit the problematic expenses?

kewitz:
@znarf sure, I'll do that manually.

PR #5981, opened 2021-05-05
Use SQL transactions for creating transactions

Betree:
Out of simplicity, this PR extracts the changes related to using SQL transactions to create transactions from https://github.com/opencollective/opencollective-api/pull/5966.

Because we will be creating up to 6 transactions per contribution, having SQL transactions is a way to make sure we won't end up with an inconsistent state (ie. having half of the transactions created) if something fails in the process.


Betree:
Closing this for now. We'll probably want to revisit later

PR #5982, opened 2021-05-05
Fix permissions for host bank account

Betree:
Related to https://opencollective.freshdesk.com/a/tickets/16630

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -2,7 +2,7 @@
 import Promise from 'bluebird';
 import config from 'config';
 import debugLib from 'debug';
-import { find, get, includes, isNumber, omit, pick } from 'lodash';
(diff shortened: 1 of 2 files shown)


server/lib/payments.js
```diff
@@ -494,7 +494,7 @@ export const sendOrderProcessingEmail = async order => {
       OrderId: order.id,
     };
     data.instructions = stripHTML(instructions).replace(/{([\s\S]+?)}/g, (match, key) => {
-      if (key && formatValues[key]) {
+      if (key && !isNil(formatValues[key])) {
```
Betree, line 497: This is to prevent displaying variable names when the value is missing (eg. `make a transfer to: {account}`)

I've made a similar fix in https://github.com/opencollective/opencollective-frontend/pull/6309.

znarf (approved):

PR #6002, opened 2021-05-07
Fix Wise webhook activity echo

kewitz:


Betree (commented):
The index looks good to me, @kewitz let me know how it performs

PR #6007, opened 2021-05-08
fixed language and simplified thank you notification

alanna:
Related front-end PR: https://github.com/opencollective/opencollective-frontend/pull/6491


alanna:
@SudharakaP what's blocking this one?

SudharakaP:
> @SudharakaP what's blocking this one?

@alanna : Ah, I think this missed my attention, I normally get a code review approval before merging; it's review is assigned to @znarf and I'll make sure we merge this on Monday. 😄

templates/emails/thankyou.hbs
```diff
@@ -16,43 +12,31 @@ Subject: Thank you for your {{currency order.totalAmount currency=order.currency
...
 
 {{#if transactionPdf}}
-<h2>📎 Attachments</h2>
-<ul>
-  <li>A PDF receipt of your contribution to {{collective.name}}</li>
-  {{#if platformTipPdf}}
-    <li>A PDF receipt of your contribution to Open Collective</li>
-  {{/if}}
-</ul>
+<h2>📎 Receipt attached</h2>
```
znarf, line 31: This was clearer before, explaining what were the different receipts. I would rollback in its previous state.

znarf (changes requested):
Let's rollback the changes about attached receipts. I personally think this section is making a good job at the moment and the proposed version doesn't improve experience for the user.

znarf:
@alanna [OSC](https://github.com/opencollective/opencollective-api/blob/main/templates/emails/thankyou.opensource.hbs) and [OCF](https://github.com/opencollective/opencollective-api/blob/main/templates/emails/thankyou.opensource.hbs) have dedicated templates we should make sure to update that one too.

There is also multiple other custom "Thank You" emails, if they're not maintained by their collective/host/community I think they should be deprecated. https://github.com/opencollective/opencollective-api/blob/main/templates/emails/

templates/emails/thankyou.hbs
SudharakaP: @znarf : I've reverted that section back to it's original state. 👍🏽

SudharakaP:
> Let's rollback the changes about attached receipts. I personally think this section is making a good job at the moment and the proposed version doesn't improve experience for the user.

@znarf : I've reverted that section back to it's original state. 👍🏽

znarf (approved):
Cool, let's follow up on opensource and foundation as soon as possible.

SudharakaP:
> Cool, let's follow up on opensource and foundation as soon as possible.

@alanna @znarf : Done at https://github.com/opencollective/opencollective-api/pull/6150.

alanna:
OK sure I agree about @znarf's comments, let's go with that.

We should make a clear decision about custom email templates - if we support them, we should build a way for the host admins to customise them directly. If we don't, we should deprecate them. Users *really* want these kinds of custom emails so I would advocate for properly supporting them if at all possible. It's hard to improve email templates if we're constantly trying to sync them up with the custom ones (maybe this is fine for you guys but I find it's confusing to keep track of).

SudharakaP:
@znarf : Noticed that we need to slightly change a test as well for this on front-end; https://github.com/opencollective/opencollective-frontend/pull/6491

PR #6023, opened 2021-05-12
Wise: Ignore payment fees and let Host absorb the costs

kewitz:
This was a feature requested by The Social Change Agency through Slack.

### DIFF
--- a/test/server/graphql/v2/mutation/ExpenseMutations.test.js
+++ b/test/server/graphql/v2/mutation/ExpenseMutations.test.js
@@ -2,7 +2,7 @@ import { expect } from 'chai';
 import config from 'config';
 import crypto from 'crypto-js';
 import gqlV2 from 'fake-tag';
-import { omit, pick } from 'lodash';
+import { defaultsDeep, omit, pick } from 'lodash';
 import sinon from 'sinon';
 import speakeasy from 'speakeasy';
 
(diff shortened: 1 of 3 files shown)


test/server/graphql/v2/mutation/ExpenseMutations.test.js
```diff
@@ -1086,6 +1086,19 @@ describe('server/graphql/v2/mutation/ExpenseMutations', () => {
...
           );
         });
+
+        it('should ignore payment processor fee if host.settings.transferwise.ignorePaymentProcessorFees is true', async () => {
+          await host.update({
+            settings: defaultsDeep(host.settings, { transferwise: { ignorePaymentProcessorFees: true } }),
+          });
+          const mutationParams = { expenseId: expense.id, action: 'PAY' };
+          const res = await graphqlQueryV2(processExpenseMutation, mutationParams, hostAdmin);
+          res.errors && console.error(res.errors);
```
SudharakaP, line 1096: Probably not needed, 😉 

```suggestion
```

SudharakaP (approved):
Looks great. 👍🏽

test/server/graphql/v2/mutation/ExpenseMutations.test.js
SudharakaP: @kewitz : Am I mistaken or does this line have some purpose. Just wondering. 😄

kewitz: @SudharakaP, a GraphQL call won't throw if there are any errors because the call is a bundle of resolvers.
It will return a property containing an array of every resolver that failed and its path.
We often include this line in our tests for debugging purposes, if there's an error we want to print it so we can properly see its stack trace.

PR #6024, opened 2021-05-12
Add period filter for virtual cards - collective settings

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4270

Related front end PR: https://github.com/opencollective/opencollective-frontend/pull/6360

### DIFF
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -34,6 +34,7 @@ import PayoutMethod from '../object/PayoutMethod';
 import { TagStats } from '../object/TagStats';
 import { TransferWise } from '../object/TransferWise';
 import EmailAddress from '../scalar/EmailAddress';
+import ISODateTime from '../scalar/ISODateTime';
 
 import { CollectionArgs } from './Collection';
 import { HasMembersFields } from './HasMembers';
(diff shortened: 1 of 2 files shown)


server/graphql/v2/interface/Account.js
```diff
@@ -351,6 +362,14 @@ const accountFieldsDefinition = () => ({
         offset: args.offset,
       };
 
+      if (args.dateFrom) {
+        query.where['createdAt'] = { [Op.gte]: args.dateFrom };
+      }
+      if (args.dateTo) {
+        query.where['createdAt'] = query.where['createdAt'] || {};
+        query.where['createdAt'][Op.lte] = args.dateTo;
```
kewitz, line 370: No need to update, this could also be:
```suggestion
        query.where['createdAt'] = Object.assign({}, query.where['createdAt'], { [Op.lte]: args.dateTo });
```

server/graphql/v2/query/TransactionsQuery.ts
```diff
@@ -52,7 +52,7 @@ const TransactionsQuery = {
     },
     dateTo: {
       type: ISODateTime,
-      description: 'Only return expenses that were created after this date',
+      description: 'Only return expenses that were created before this date',
```
kewitz, line 55: :dart:

kewitz (approved):

server/graphql/v2/interface/Account.js
SudharakaP: @kewitz : Ah, thanks. Seems more nicer this way. 😄

PR #6045, opened 2021-05-18
Support Zero Decimal Currencies on Stripe

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4297

**Screen-cast**

zero-decimal-final

### DIFF
--- a/server/paymentProviders/stripe/creditcard.js
+++ b/server/paymentProviders/stripe/creditcard.js
@@ -1,6 +1,7 @@
 import config from 'config';
 import { get, isNumber, result } from 'lodash';
 
+import { ZERO_DECIMAL_CURRENCIES } from '../../constants/currencies';
 import * as constants from '../../constants/transactions';
 import logger from '../../lib/logger';
 import { createRefundTransaction, getHostFee, getPlatformFee } from '../../lib/payments';
@@ -118,8 +119,11 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
   const isSharedRevenue = !!hostFeeSharePercent;
 
   // Read or compute Platform Fee
-  const platformFee = await getPlatformFee(order.totalAmount, order, host, { hostFeeSharePercent });
-  const platformTip = order.data?.platformFee;
+  const platformFee = zeroDecimalCurrencyValue(
+    order.currency,
+    await getPlatformFee(order.totalAmount, order, host, { hostFeeSharePercent }),
+  );
+  const platformTip = zeroDecimalCurrencyValue(order.currency, order.data?.platformFee);
 
   // Make sure data is available (breaking in some old tests)
   order.data = order.data || {};
@@ -129,7 +133,7 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
   let paymentIntent = order.data.paymentIntent;
   if (!paymentIntent) {
     const createPayload = {
-      amount: order.totalAmount,
+      amount: zeroDecimalCurrencyValue(order.currency, order.totalAmount),
       currency: order.currency,
       customer: hostStripeCustomer.id,
       description: order.description,
@@ -161,7 +165,6 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
       stripeAccount: hostStripeAccount.username,
     });
   }
-
   paymentIntent = await stripe.paymentIntents.confirm(paymentIntent.id, {
     stripeAccount: hostStripeAccount.username,
   });
@@ -231,7 +234,18 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
 };
 
 /**
- * Given a charge id, retrieves its correspind charge and refund data.
+ * Handles the zero-decimal currencies for Stripe; https://stripe.com/docs/currencies#zero-decimal
+ */
+const zeroDecimalCurrencyValue = (currency, amount) => {
+  if (ZERO_DECIMAL_CURRENCIES.includes(currency.toUpperCase())) {
+    return Math.floor(amount / 100);
+  } else {
+    return amount;
+  }
+};
+
+/**
+ * Given a charge id, retrieves its corresponding charge and refund data.
  */
 export const retrieveChargeWithRefund = async (chargeId, stripeAccount) => {
   const charge = await stripe.charges.retrieve(chargeId, {
--- a/server/paymentProviders/stripe/creditcard.js
+++ b/server/paymentProviders/stripe/creditcard.js
@@ -119,11 +119,8 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
   const isSharedRevenue = !!hostFeeSharePercent;
 
   // Read or compute Platform Fee
-  const platformFee = zeroDecimalCurrencyValue(
-    order.currency,
-    await getPlatformFee(order.totalAmount, order, host, { hostFeeSharePercent }),
-  );
-  const platformTip = zeroDecimalCurrencyValue(order.currency, order.data?.platformFee);
+  const platformFee = await getPlatformFee(order.totalAmount, order, host, { hostFeeSharePercent });
+  const platformTip = order.data?.platformFee;
 
   // Make sure data is available (breaking in some old tests)
   order.data = order.data || {};
@@ -133,7 +130,7 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
   let paymentIntent = order.data.paymentIntent;
   if (!paymentIntent) {
     const createPayload = {
-      amount: zeroDecimalCurrencyValue(order.currency, order.totalAmount),
+      amount: convertToStripeAmount(order.currency, order.totalAmount),
       currency: order.currency,
       customer: hostStripeCustomer.id,
       description: order.description,
@@ -146,7 +143,7 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
     };
     // We don't add a platform fee if the host is the root account
     if (platformFee && hostStripeAccount.username !== config.stripe.accountId) {
-      createPayload.application_fee_amount = platformFee;
+      createPayload.application_fee_amount = convertToStripeAmount(order.currency, platformFee);
     }
     if (order.interval) {
       createPayload.setup_future_usage = 'off_session';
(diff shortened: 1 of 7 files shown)


server/paymentProviders/stripe/creditcard.js
```diff
@@ -161,7 +162,7 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
       stripeAccount: hostStripeAccount.username,
     });
   }
-
+  paymentIntent = handleZeroDecimalCurrencies(paymentIntent);
```
znarf, line 165: I don't think this much, I would prefer if we create the `amount` and `application_fee_amount` straight with the right values above.

znarf (changes requested):
One change requested.

Also, could you verify that it works as expected with a test?

server/paymentProviders/stripe/creditcard.js
SudharakaP: @znarf : But notice that we are not always creating the `paymentIntent` right? On line 130; 

https://github.com/opencollective/opencollective-api/blob/e1e76bb9a46ecc4f9fab28f27c705b885337cf17/server/paymentProviders/stripe/creditcard.js#L165

it is possible that the object already contains the `paymentIntent`, in which I think we need to alter it after the creation. Am I missing something? 😄

SudharakaP:
> One change requested.
> 
> Also, could you verify that it works as expected with a test?

@znarf : You mean testing as in logging into a test stripe account? The only test I did here is to validate that the payment_intent object is correct as shown in the screenshot. 😃

znarf:
@SudharakaP In the screencast, if we're doing a 10JPY transaction, why don't we see 10JPY in the Transactions list? This is what should be returned right?

SudharakaP:
> @SudharakaP In the screencast, if we're doing a 10JPY transaction, why don't we see 10JPY in the Transactions list? This is what should be returned right?

@znarf: Seems like a caching issue; note that in the screen-cast I refresh the screen and it shows the 10 JPY transaction. 😄 

But I did found a small issue based on your previous review let me correct those and update the screen-cast.

server/paymentProviders/stripe/creditcard.js
znarf: We are creating it in this code and not anywhere else, so it's all fine.

We don't have to handle the case of such paymentIntent already existing in our db, because we assume such case is not existing and will not exist.

SudharakaP: @znarf : Thanks changed the code to make sure the the amount and application_fee_amount are calculated with the right amounts. 👍🏽

znarf:
Please look at `balanceTransaction` and properly handle the amounts generated from it: `hostFeeInHostCurrency`, `platformFeeInHostCurrency`, `paymentProcessorFeeInHostCurrency`, `amountInHostCurrency`, `hostCurrencyFxRate`.

Also, please add a GraphQL test to check that the returned transaction match what is expected.

server/paymentProviders/stripe/creditcard.js
```diff
@@ -231,7 +234,18 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
 };
 
 /**
- * Given a charge id, retrieves its correspind charge and refund data.
+ * Handles the zero-decimal currencies for Stripe; https://stripe.com/docs/currencies#zero-decimal
+ */
+const zeroDecimalCurrencyValue = (currency, amount) => {
```
znarf, line 239: It's not a good descriptive name for what it's doing.

What does it do? It converts an amount to Stripe's expected format. What is the most important thing? The amount?

I would then propose, `convertToStripeAmount = (amount, currency) =>`

You will later need the opposite, `convertFromStripeAmount = (amount, currency) =>`

SudharakaP: Thanks. I've renamed the functions. 👍🏽

SudharakaP:
> Please look at `balanceTransaction` and properly handle the amounts generated from it: `hostFeeInHostCurrency`, `platformFeeInHostCurrency`, `paymentProcessorFeeInHostCurrency`, `amountInHostCurrency`, `hostCurrencyFxRate`.
> 
> Also, please add a GraphQL test to check that the returned transaction match what is expected.

@znarf : Sorry, this is taking a bit of time as I am still trying out this handle the amounts that are generated from the `balanceTransactions` properly. One thing I noticed is that even before this PR the amount calculation seems off when we use a zero dollar transaction. For example see the screenshot below; this was taken from main branch. I think we need to fix this first but couldn't figure it out yet. 🤔 

image

znarf:
> My understanding is that it should show the full amount submitted (which is 100 YEN) and the platform fee in YEN (the same currency) and then the total amount. Is that correct? 

We don't have a Platform Fee here, you mean Payment Processor Fee? It should be in YEN, unless it's recorded in dollars in the database. It would be recorded in dollars if the Host Stripe account is configured in dollars, in YEN if the Host Stripe Account configured in YEN. We're normally ensuring that the Host currency on  Open Collective match the one on Stripe.

Why it's not displaying 100? You have to look at the raw record from Stripe to understand, we're storing it somewhere in the Transactions table.

SudharakaP:
> > My understanding is that it should show the full amount submitted (which is 100 YEN) and the platform fee in YEN (the same currency) and then the total amount. Is that correct?
> 
> We don't have a Platform Fee here, you mean Payment Processor Fee? It should be in YEN, unless it's recorded in dollars in the database. It would be recorded in dollars if the Host Stripe account is configured in dollars, in YEN if the Host Stripe Account configured in YEN. We're normally ensuring that the Host currency on Open Collective match the one on Stripe.
> 
> Why it's not displaying 100? You have to look at the raw record from Stripe to understand, we're storing it somewhere in the Transactions table.

@znarf : Thanks your reply makes perfect sense. 😄  Yes, I meant the payment processor fee. It seems that the test stripe account is configured in dollars which is the reason it's shown in dollars in my screen-cast. I am trying to find a method to change the currency of the test stripe account. Do you know a method of changing the currency of our test stripe account? 🤔

SudharakaP:
@znarf : I've found the remaining problems with this transaction calculations and corrected them. Now we should have support for the case where the host currency is different from the host's stripe account currency. To demonstrate I've updated the above screencast. 😄

server/paymentProviders/stripe/creditcard.js
```diff
@@ -203,7 +206,11 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
     hostFeeSharePercent,
   };
 
-  const platformFeeInHostCurrency = isSharedRevenue ? platformTip || 0 : fees.applicationFee;
+  const fxRate = convertFromStripeAmount(balanceTransaction.currency, balanceTransaction.amount) / order.totalAmount;
```
znarf, line 209: Why not name it `hostCurrencyFxRate` ?

znarf: What about computing the result of `convertFromStripeAmount(balanceTransaction.currency, balanceTransaction.amount)` before, storing it in a variable and naming it `amountInHostCurrency`?

znarf:
As asked yesterday:

> Also, please add a GraphQL test to check that the returned transaction match what is expected

server/paymentProviders/stripe/creditcard.js
SudharakaP: Thanks. 👍🏽 I've corrected those. 😄

SudharakaP:
> As asked yesterday:
> 
> > Also, please add a GraphQL test to check that the returned transaction match what is expected

@znarf: Ops, I missed this one. 😄 I've just added a new graphql test file to validate the zero decimal currencies. Let me know what you think. I was wondering about the best place to put this file; but decided on the `stripe` folder as it's related to that; but if you disagree we can move it to a different folder. 😄

server/paymentProviders/stripe/creditcard.js
```diff
@@ -191,7 +191,8 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
 
   // Create a Transaction
   const fees = extractFees(balanceTransaction);
```
znarf, line 193: Just noting that extractFees is not super solid, let's say that for some reason there is a currency mismatch, we will be wrong. It would be better to read fees directly from the Stripe object:

```
 {
        "type": "stripe_fee",
        "amount": 332,
        "currency": "usd",
        "application": null,
        "description": "Stripe processing fees"
      },
```

This doesn't need to be addressed.

server/paymentProviders/stripe/creditcard.js
```diff
@@ -216,9 +221,9 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
       amount: order.totalAmount,
       currency: order.currency,
       hostCurrency: balanceTransaction.currency?.toUpperCase(),
-      amountInHostCurrency: balanceTransaction.amount,
-      hostCurrencyFxRate: balanceTransaction.amount / order.totalAmount,
-      paymentProcessorFeeInHostCurrency: fees.stripeFee,
+      amountInHostCurrency: amountInHostCurrency,
```
znarf, line 224:
```suggestion
      amountInHostCurrency,
```

Doesn't have to be addressed.

znarf (approved):
Looks good to me.

server/paymentProviders/stripe/creditcard.js
SudharakaP: @znarf : I'll leave this as it is for now since the only purpose of this `extractFees` object is to get the `stripeFees` and `applicationFees` for now; both of which we convert to proper values in place. 😄

server/paymentProviders/stripe/creditcard.js
SudharakaP: Done. 👍🏽

SudharakaP:
@znarf : As discussed, I've added support for the refund transactions; here's a screen-cast of that working as well. This is actually part of your previous review where the `extractFees` function not being handling the currency conversion properly. I realized that adding that support is sufficient for the refunds to work since we are always using the `extractFees` function after getting the `balanceTransactions` object from Stripe. 😄 

refunds-with-zero-dollar-currencies

test/server/paymentProviders/stripe/zero-decimal-currencies.test.js
```diff
@@ -0,0 +1,72 @@
+import { expect } from 'chai';
+import gql from 'fake-tag';
+import { describe, it } from 'mocha';
+
+import * as store from '../../../stores';
+import * as utils from '../../../utils';
+
+describe('server/graphql/v1/transaction', () => {
```
znarf, line 8: We usually put GraphQL tests in the `test/server/graphql/v1` folder

We want to make the `describe` match the filename.

znarf:
Could you add a test for the refund part? GraphQL test is better than the other form because it covers more.

test/server/paymentProviders/stripe/zero-decimal-currencies.test.js
SudharakaP: Move to the `v1 folder and change the names. Sorry I missed this somehow. 😄

SudharakaP:
> Could you add a test for the refund part? GraphQL test is better than the other form because it covers more.

@znarf : Ah, good point. I've added some graphql testing for this as well. Let me know if you see any further issues. 😄

znarf (approved):
Looks good to me.

test/server/graphql/v1/zero-decimal-currencies.test.js
```diff
@@ -0,0 +1,72 @@
...
+        result.data.Collective.transactions[0].paymentProcessorFeeInHostCurrency;
+      const hostFeeInHostCurrency = result.data.Collective.transactions[0].hostFeeInHostCurrency;
+      const netAmountInCollectiveCurrency = result.data.Collective.transactions[0].netAmountInCollectiveCurrency;
+
+      expect(result.data.Collective).to.exist;
+      expect(result.data.Collective.transactions).to.have.length(1);
+      expect(amount).to.equal(10000);
+      expect(platformFeeInHostCurrency).to.equal(-25 * 100);
+      expect(paymentProcessorFeeInHostCurrency).to.equal(-35 * 100);
+      expect(hostFeeInHostCurrency).to.equal((-(10000 * 5) / 100) * 100);
```
Betree, line 66: This one looks strange, if the contribution amount is `10000` with a host fee of 5% then the fee should be `500`, not `50000`?

PR #6053, opened 2021-05-19
Ledger: rename "Fees On Top" "Platform Tips" in a few places

znarf:


server/models/Transaction.js
```diff
@@ -575,7 +575,7 @@ function defineModel() {
     return Promise.mapSeries(transactions, t => Transaction.create(t, opts)).then(results => results[index]);
   };
 
-  Transaction.createFeesOnTopTransaction = async ({ transaction, host }) => {
+  Transaction.createPlatformTipTransaction = async ({ transaction, host }) => {
```
Betree, line 578: In [my PR](https://github.com/opencollective/opencollective-api/pull/5966/files#diff-4cffa0c51e597183bea1e391ac7dde45b32da7acdd7a8cfc4d2c42443d704097R609), I called this one `createPlatformTipTransactions` (with a `s`) since we're creating a transaction pair, but also the debt transactions

server/models/Transaction.js
```diff
@@ -675,7 +675,9 @@ function defineModel() {
 
     // Separate donation transaction and remove platformFee from the main transaction
     if (transaction.data?.isFeesOnTop && transaction.platformFeeInHostCurrency) {
-      transaction = (await Transaction.createFeesOnTopTransaction({ transaction, host })).transaction;
+      const result = await Transaction.createPlatformTipTransaction({ transaction, host });
+      // Transaction was modified by createPlatformTipTransaction, we get it from the result
```
Betree, line 679: Nice to add a comment on that

Betree (approved):

server/models/Transaction.js
znarf: Ok, updated to plural

PR #6059, opened 2021-05-20
Fix the broken button on updates email template

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4304


znarf (approved):
Looks good.

PR #6060, opened 2021-05-20
Fix error introduced on winston logger after upgrading to v3.2.1

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4305

This solution is derived from the suggestion given in the following comment; https://github.com/winstonjs/winston/issues/1591#issuecomment-459335734

### DIFF
--- a/server/lib/logger.js
+++ b/server/lib/logger.js
@@ -1,8 +1,6 @@
 import config from 'config';
 import winston, { format } from 'winston';
 
-const logger = winston.createLogger();
-
 const winstonLevel = config.log.level;
 
 const winstonFormat = format.combine(format.colorize(), format.splat(), format.simple());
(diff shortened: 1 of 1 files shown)


znarf:
Update to winston 3.2.1 is 2 years old:
https://github.com/opencollective/opencollective-api/commit/cfbbc98d8254895c4e2254b4c45201fe8f8ffbc9

Why haven't we seen this before?

SudharakaP:
> Update to winston 3.2.1 is 2 years old: [cfbbc98](https://github.com/opencollective/opencollective-api/commit/cfbbc98d8254895c4e2254b4c45201fe8f8ffbc9)
> 
> Why haven't we seen this before?

@znarf : Oh, I didn't realize that it was updated so long ago. Hmm....I am not sure, it is probably the first time I was running `npm run test:update`. Wonder if this is machine dependent somehow. 🤔 🤔

znarf:
Let's close unless the problem is still going on.

SudharakaP:
> Let's close unless the problem is still going on.

Yes, I think this is something that only effected me. Thanks for closing it. 👍🏽

PR #6067, opened 2021-05-23
Add video embed support for updates

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4298

Depends on https://github.com/opencollective/opencollective-frontend/pull/6394

### DIFF
--- a/server/lib/notifications.js
+++ b/server/lib/notifications.js
@@ -241,26 +241,17 @@ const notifyUpdateSubscribers = async activity => {
 };
 
 function replaceVideosByImagePreviews(activity) {
-  const regexImg = new RegExp(`<iframe([\\w\\W]+?)[\\/]?>`, 'g');
-  const regexUrl = new RegExp(
-    `(https):\\/\\/([\\w_-]+(?:(?:\\.[\\w_-]+)+))([\\w.,@?^=%&:/~+#-]*[\\w@?^=%&/~+#-])?`,
-    'ig',
-  );
+  const iframePreviewRegex = /<iframe\s[^>]*src="([^"]+)"[^>]*><\/iframe>/gi;
   activity.data.update.html = activity.data.update.html
-    .replace(regexImg, match => {
-      let imageReplacement;
-      if (match.includes('youtube')) {
-        imageReplacement = match
-          .replace(new RegExp('<iframe', 'g'), '<img')
-          .replace(new RegExp('allowfullscreen', 'g'), '')
-          .replace(new RegExp('frameborder="0"', 'g'), '');
-
-        const { videoService, videoId } = getEmbedDetails(imageReplacement);
-        const previewImageURL = constructPreviewImageURL(videoService, videoId);
-        return imageReplacement.replace(regexUrl, previewImageURL);
+    .replace(iframePreviewRegex, (_, href) => {
+      const { service, id } = parseServiceLink(href);
+      const imgSrc = constructPreviewImageURL(service, id);
+      if (imgSrc) {
+        return `<img src="${imgSrc}" alt="${service} content">`;
+      } else {
+        return '';
       }
     })
-    .replace(new RegExp('<figure data-trix-content-type="([\\w\\W]+?)[\\/]?">', 'g'), '')
     .replace(new RegExp('</iframe><figcaption></figcaption></figure>', 'g'), '')
     .replace(new RegExp('width="100%" height="394"', 'g'), '');
   return activity;
@@ -271,15 +262,6 @@ function constructPreviewImageURL(service, id) {
     return `https://img.youtube.com/vi/${id}/0.jpg`;
   } else {
     return null;
-  }
-}
-
-function getEmbedDetails(iframe) {
-  const regex = new RegExp(`(https):\\/\\/([\\w_-]+(?:(?:\\.[\\w_-]+)+))([\\w.,@?^=%&:/~+#-]*[\\w@?^=%&/~+#-])?`, 'ig');
-  const match = regex.exec(iframe);
-  if (match[0].includes('youtube')) {
-    const { id } = parseServiceLink(match[0]);
-    return { videoService: 'youtube', videoId: id };
   }
 }
 
--- a/server/lib/notifications.js
+++ b/server/lib/notifications.js
@@ -12,9 +12,9 @@ import { TransactionTypes } from '../constants/transactions';
 import activitiesLib from '../lib/activities';
 import emailLib, { NO_REPLY_EMAIL } from '../lib/email';
 import models from '../models';
+import { sanitizerOptions } from '../models/Update';
 
 import { getTransactionPdf } from './pdf';
-import { buildSanitizerOptions } from './sanitize-html';
 import slackLib from './slack';
 import twitter from './twitter';
 import { parseToBoolean, toIsoDateStr } from './utils';
@@ -243,17 +243,8 @@ const notifyUpdateSubscribers = async activity => {
 };
 
 function replaceVideosByImagePreviews(activity) {
-  const sanitizeOptions = buildSanitizerOptions({
-    links: true,
-    mainTitles: true,
-    titles: true,
-    basicTextFormatting: true,
-    multilineTextFormatting: true,
-    images: true,
-    tables: true,
-  });
-  sanitizeOptions.transformTags = {
-    ...sanitizeOptions.transformTags,
+  sanitizerOptions.transformTags = {
+    ...sanitizerOptions.transformTags,
     iframe: (tagName, attribs) => {
       const { service, id } = parseServiceLink(attribs.src);
       const imgSrc = constructPreviewImageURL(service, id);
@@ -270,7 +261,7 @@ function replaceVideosByImagePreviews(activity) {
       }
     },
   };
-  activity.data.update.html = sanitizeHtml(activity.data.update.html, sanitizeOptions);
+  activity.data.update.html = sanitizeHtml(activity.data.update.html, sanitizerOptions);
   return activity;
 }
 
--- a/server/lib/notifications.js
+++ b/server/lib/notifications.js
@@ -260,6 +260,8 @@ function replaceVideosByImagePreviews(activity) {
 function constructPreviewImageURL(service, id) {
   if (service === 'youtube') {
     return `https://img.youtube.com/vi/${id}/0.jpg`;
+  } else if (service === 'anchorFm') {
+    return `https://theme.zdassets.com/theme_assets/2009830/ed34a3258bf8d79c3db30e4269dd95052481fc50.png`;
   } else {
     return null;
   }
--- a/server/lib/notifications.js
+++ b/server/lib/notifications.js
@@ -3,6 +3,7 @@ import Promise from 'bluebird';
 import config from 'config';
 import debugLib from 'debug';
 import { get, remove } from 'lodash';
+import sanitizeHtml from 'sanitize-html';
 
 import { activities, channels } from '../constants';
 import activityType from '../constants/activities';
@@ -13,6 +14,7 @@ import emailLib, { NO_REPLY_EMAIL } from '../lib/email';
 import models from '../models';
 
 import { getTransactionPdf } from './pdf';
+import { buildSanitizerOptions } from './sanitize-html';
 import slackLib from './slack';
 import twitter from './twitter';
 import { parseToBoolean, toIsoDateStr } from './utils';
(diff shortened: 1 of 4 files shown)


server/lib/notifications.js
```diff
@@ -215,9 +215,87 @@ const notifyUpdateSubscribers = async activity => {
...
+          .replace(new RegExp('frameborder="0"', 'g'), '');
+
+        const { videoService, videoId } = getEmbedDetails(imageReplacement);
+        const previewImageURL = constructPreviewImageURL(videoService, videoId);
+        return imageReplacement.replace(regexUrl, previewImageURL);
+      }
+    })
+    .replace(new RegExp('<figure data-trix-content-type="([\\w\\W]+?)[\\/]?">', 'g'), '')
+    .replace(new RegExp('</iframe><figcaption></figcaption></figure>', 'g'), '')
+    .replace(new RegExp('width="100%" height="394"', 'g'), '');
```
Betree, line 244: It's quite complex to undestand this code without proper context. It also looks like `.replace(new RegExp('<figure data-trix-content-type="([\\w\\W]+?)[\\/]?">', 'g'), '')` can have side effects that are not related to iframes.

I haven't tested the following code (other than the regex with https://regex101.com/r/OiYxtK/1), but I think using a slightly more advanced regex for the iframe can really simplify things here:

```es6
const iframePreviewRegex = /<iframe\s[^>]*src="([^"]+)"[^>]*><\/iframe>/ig;
activity.data.update.html = activity.data.update.html.replaceAll(iframePreviewRegex, (_, href) => {
  const { service, id } = parseServiceLink(href);
  const imgSrc = constructPreviewImageURL(service, id);
  if (imgSrc) {
    return `<img src="${imgSrc}" alt="${service} content">`
  } else {
    return ''
  }
})
```

Betree (commented):
Would be great to have a preview for anchor.fm, people will be surprised if it just disappears from the content when they send emails.

Also, we really need unit tests to make sure future regressions on this code don't break the emails.

server/lib/notifications.js
```diff
@@ -215,9 +215,87 @@ const notifyUpdateSubscribers = async activity => {
...
+  console.log(activity.data.update.html);
```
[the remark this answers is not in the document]
SudharakaP: Ops I've missed this 🤯

server/lib/notifications.js
SudharakaP: @Betree : Thanks much. I've refactored this based on your suggestion. Seems to work perfectly. Now as for the `.replace(new RegExp('<figure data-trix-content-type="([\\w\\W]+?)[\\/]?">', 'g'), '')`, yes you are right; this is an oversight on my part and it could have side effects. 🙄 I removed this line as it's really not needed to remove these `<figure` tags anyways. 🐨

SudharakaP:
> Would be great to have a preview for anchor.fm, people will be surprised if it just disappears from the content when they send emails.
> 
> Also, we really need unit tests to make sure future regressions on this code don't break the emails.

@Betree : Working on this. I don't think we could get a preview image for anchor.fm as in YouTube. So I am gonna put generic image there. Hope that's adequate for this purpose. 🤔 🐨

SudharakaP:
@Betree : I've added the anchor preview as well. It looks looks like this. Let me know your thoughts. I wonder if this is the best placeholder to show but I felt this is good for now? 🤔 🐨 

image

Betree:
> > Would be great to have a preview for anchor.fm, people will be surprised if it just disappears from the content when they send emails.
> > Also, we really need unit tests to make sure future regressions on this code don't break the emails.
> 
> @Betree : Working on this. I don't think we could get a preview image for anchor.fm as in YouTube. So I am gonna put generic image there. Hope that's adequate for this purpose. thinking koala

Yes, absolutely! Ideally, it should be something that lets you know you can click on it to read the track but for a first version any image just to make sure the content doesn't disappear is fine :wink:

server/lib/notifications.js
```diff
@@ -215,9 +215,62 @@ const notifyUpdateSubscribers = async activity => {
...
+  activity.data.update.html = activity.data.update.html
+    .replace(iframePreviewRegex, (_, href) => {
+      const { service, id } = parseServiceLink(href);
+      const imgSrc = constructPreviewImageURL(service, id);
+      if (imgSrc) {
+        return `<img src="${imgSrc}" alt="${service} content">`;
+      } else {
+        return '';
+      }
+    })
```
Betree, line 233: image


Regarding this warning, I would add a small line to remove all iframes that we haven't been able to convert

```suggestion
    }).replace(/<iframe\s[^>]><\/iframe>/gi, '')
```

server/lib/notifications.js
```diff
@@ -215,9 +215,62 @@ const notifyUpdateSubscribers = async activity => {
...
+    .replace(new RegExp('</iframe><figcaption></figcaption></figure>', 'g'), '')
+    .replace(new RegExp('width="100%" height="394"', 'g'), '');
```
Betree, line 235: I don't think we still need these two?

server/lib/notifications.js
Betree: We can also use `sanitize-html` for this

server/lib/notifications.js
SudharakaP: Nope, you are right. That's my mistake. 😞

server/lib/notifications.js
SudharakaP: Done. 👍🏽 Didn't notice this warning in CI. 🤔

SudharakaP:
@Betree : I've corrected the remaining issues and added some tests as well. Let me know how it looks and if anything else needs to be done. 😄

server/lib/notifications.js
```diff
@@ -215,9 +215,61 @@ const notifyUpdateSubscribers = async activity => {
...
+    })
+    .replace(/<iframe\s[^>]><\/iframe>/gi, '');
+  return activity;
+}
+
+function constructPreviewImageURL(service, id) {
+  if (service === 'youtube') {
+    return `https://img.youtube.com/vi/${id}/0.jpg`;
+  } else if (service === 'anchorFm') {
+    return `https://theme.zdassets.com/theme_assets/2009830/ed34a3258bf8d79c3db30e4269dd95052481fc50.png`;
```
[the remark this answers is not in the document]
SudharakaP: @Betree : Good point. This image I put comes actually from the anchor.fm help page; https://help.anchor.fm/hc/en-us/articles/360016501412-Creating-your-podcast-cover-art. 

image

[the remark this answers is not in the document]
Betree: Yes

server/lib/notifications.js
```diff
@@ -215,9 +215,61 @@ const notifyUpdateSubscribers = async activity => {
   const emailOpts = { from: activity.data.fromEmail };
   const update = await models.Update.findByPk(activity.data.update.id);
   const allUsers = await update.getUsersToNotify();
-  return notifySubscribers(allUsers, activity, emailOpts);
+  const modifiedActivity = replaceVideosByImagePreviews(activity);
+  return notifySubscribers(allUsers, modifiedActivity, emailOpts);
 };
 
+function replaceVideosByImagePreviews(activity) {
+  const iframePreviewRegex = /<iframe\s[^>]*src="([^"]+)"[^>]*><\/iframe>/gi;
```
Betree, line 223: I'm not comfortable with us using regexps. This one won't cover all use cases (what if the regex has no closing tags? or something between the tags (i.e. `<iframe>Hello</iframe>`)? I'm also uncertain about the `.replace`, it's difficult to rely on this.

I'd like us to use a real HTML parser to achieve this. Thankfully, `sanitize-html` has just what we need: https://github.com/apostrophecms/sanitize-html#transformations. By using the `transformTags` option we'll be able to replace iframes by images safely and the code should be easier to understand.

server/lib/notifications.js
```diff
@@ -215,9 +215,61 @@ const notifyUpdateSubscribers = async activity => {
...
+        return '';
+      }
+    })
+    .replace(/<iframe\s[^>]><\/iframe>/gi, '');
+  return activity;
+}
+
+function constructPreviewImageURL(service, id) {
+  if (service === 'youtube') {
+    return `https://img.youtube.com/vi/${id}/0.jpg`;
```
Betree, line 240: In the spirit of defensive programming, I think we should sanitize the `id` here. A YouTube id is `[a-zA-Z0-9_-]{11}`, if we don't get something like that we should return `null`.

Betree (changes requested):

server/lib/notifications.js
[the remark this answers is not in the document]
SudharakaP: Done. 👍🏽

server/lib/notifications.js
SudharakaP: Thanks. I've used the transformations from `sanitize-html`. Let me know how it looks. I've enabled all content types we support except video embeds; 

https://github.com/opencollective/opencollective-api/blob/b7b4aef1c118a594255e1bab1ba4c4702fe117f4/server/lib/notifications.js#L245-L275

since we area dealing with a rich text editor text. 🤔

server/lib/notifications.js
SudharakaP: Thanks. I've changed that to sanitize using the regex you suggested. 👍🏽

SudharakaP:
@Betree : Also in the final commit; 5d66ec0 I've removed one test which I've added. This isn't a YouTube link or a AnchorFM link and hence this kinda of iframe isn't valid and the API will not add such content at all. 

https://github.com/opencollective/opencollective-api/blob/f1648e68812a7d724e9bb7b6a29f744cd73debbc/server/lib/sanitize-html.ts#L90

Hence the test will always fail for such a test I think. 😉

Betree:
> @Betree : Also in the final commit; [5d66ec0](https://github.com/opencollective/opencollective-api/commit/5d66ec099401a097fe9b2f26a001650af02afe88) I've removed one test which I've added. This isn't a YouTube link or a AnchorFM link and hence this kinda of iframe isn't valid and the API will not add such content at all.
> 
> https://github.com/opencollective/opencollective-api/blob/f1648e68812a7d724e9bb7b6a29f744cd73debbc/server/lib/sanitize-html.ts#L90
> 
> Hence the test will always fail for such a test I think. wink

This test was there on purpose, the goal was to make sure iframe from other domains are stripped out. We should add the test back.

server/lib/notifications.js
```diff
@@ -236,9 +238,77 @@ const notifyUpdateSubscribers = async activity => {
   const emailOpts = { from: activity.data.fromEmail };
   const update = await models.Update.findByPk(activity.data.update.id);
   const allUsers = await update.getUsersToNotify();
-  return notifySubscribers(allUsers, activity, emailOpts);
+  const modifiedActivity = replaceVideosByImagePreviews(activity);
+  return notifySubscribers(allUsers, modifiedActivity, emailOpts);
 };
 
+function replaceVideosByImagePreviews(activity) {
+  const sanitizeOptions = buildSanitizerOptions({
```
Betree, line 246: Would be great to import this from https://github.com/opencollective/opencollective-api/blob/0bc1022162162c6b2621eb6c6df8e897145ae93f/server/models/Update.js#L15, to make sure changes to options there are added here as well.

Betree (commented):
Looking great! Small feedback below. Let's then add back the test and we'll be good to go.

server/lib/notifications.js
SudharakaP: Corrected this one. 👍🏽

SudharakaP:
> Looking great! Small feedback below. Let's then add back the test and we'll be good to go.

Thanks for the explanation, that makes sense. I've added the test back and adjusted the count here,

https://github.com/opencollective/opencollective-api/blob/5b31336033ee042a36a03c3b6e439f1aa37aee95/test/server/lib/notifications.test.ts#L170-L174

so that we don't consider the final test (with `www.test.com`) as that one is stripped out. Let me know if everything looks good now to merge. 👍🏽

server/lib/notifications.js
```diff
@@ -236,9 +238,68 @@ const notifyUpdateSubscribers = async activity => {
   const emailOpts = { from: activity.data.fromEmail };
   const update = await models.Update.findByPk(activity.data.update.id);
   const allUsers = await update.getUsersToNotify();
-  return notifySubscribers(allUsers, activity, emailOpts);
+  const modifiedActivity = replaceVideosByImagePreviews(activity);
+  return notifySubscribers(allUsers, modifiedActivity, emailOpts);
 };
 
+function replaceVideosByImagePreviews(activity) {
+  sanitizerOptions.transformTags = {
```
Betree, line 246: Doing so is mutating the `sanitizerOptions` from `../models/Update`.

I suggest to do something like:

```es6
import { sanitizerOptions as updateSanitizerOptions } from '../models/Update';

function replaceVideosByImagePreviews(activity) {
  const sanitizerOptions = {
    ...updateSanitizerOptions,
    ...
  }
```

Betree (changes requested):
> so that we don't consider the final test (with www.test.com) as that one is stripped out. Let me know if everything looks good now to merge.

The `iframe` should be stripped out from the content, but the mail should still be sent

server/lib/notifications.js
SudharakaP: Done. Thanks for pointing it out. 👍🏽

Betree (approved):
:rocket: :rocket: :rocket:

PR #6075, opened 2021-05-25
TaxForms: Truncate name

Betree:
Fix https://opencollective.slack.com/archives/GST5GVCQ7/p1621945053000400

Combined user name + collective name were too long


znarf (approved):
Perfect

PR #6090, opened 2021-05-27
Add Support for Structured Bank info Without Wise

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/3893

Related front end PR: https://github.com/opencollective/opencollective-frontend/pull/6418

### DIFF
--- a/server/graphql/v2/object/TransferWise.ts
+++ b/server/graphql/v2/object/TransferWise.ts
@@ -74,7 +74,7 @@ export const TransferWise = new GraphQLObjectType({
       args: {
         ignoreBlockedCurrencies: {
           type: GraphQLBoolean,
-          description: 'Ignores blacklisted currencies, used to generate the bank information form for manual payments',
+          description: 'Ignores blocked currencies, used to generate the bank information form for manual payments',
         },
       },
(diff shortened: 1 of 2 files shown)


server/graphql/v2/object/TransferWise.ts
```diff
@@ -74,7 +74,7 @@ export const TransferWise = new GraphQLObjectType({
       args: {
         ignoreBlockedCurrencies: {
           type: GraphQLBoolean,
-          description: 'Ignores blocklisted currencies, used to generate the bank information form for manual payments',
+          description: 'Ignores blacklisted currencies, used to generate the bank information form for manual payments',
```
kewitz, line 77: We're moving out of terms like blacklist/whitelist and master/slave.
```suggestion
          description: 'Ignores blocked currencies, used to generate the bank information form for manual payments',
```

kewitz (approved):
💯

server/graphql/v2/object/TransferWise.ts
SudharakaP: Ah, yes thanks for pointing it out. 🐼

PR #6095, opened 2021-05-28
Rewrite host-metrics with simple sql

znarf:


cron/monthly/host-settlement.js
```diff
@@ -320,7 +320,7 @@ export async function run() {
     const transactions = hostTransactions.map(t => {
       if (t.source === 'Shared Revenue') {
         // In this context, the original t.amount is actually -t.hostFeeInHostCurrency
-        t.amount = round(t.amount * ((t.data?.hostFeeSharePercent || hostFeeSharePercent) / 100));
+        t.amount = t.amount * ((t.data?.hostFeeSharePercent || hostFeeSharePercent) / 100);
```
Betree, line 323: If amount=333 and percentage=0.1, we'll get `33.300000000000004` here. We can eventually do `round(x, 2)`

```suggestion
        t.amount = round(t.amount * ((t.data?.hostFeeSharePercent || hostFeeSharePercent) / 100));
```

server/lib/host-metrics.js
```diff
@@ -0,0 +1,218 @@
...
+    END
+   ) as "_currency"
+FROM "Transactions" as t1
+INNER JOIN "Transactions" as t2
+ON t1."TransactionGroup" = t2."TransactionGroup"
+INNER JOIN "Collectives" as h
+ON t1."HostCollectiveId" = h."id"
+WHERE t1."HostCollectiveId" = :HostCollectiveId
+AND t1."createdAt" >= :startDate AND t1."createdAt" <= :endDate
+AND (t1."kind" IS NULL OR t1."kind" NOT IN ('PLATFORM_TIP'))
```
Betree, line 55: I realize we haven't added any index on `kind`. We could maybe look into that when we'll want to optimize these queries.

Betree (commented):
Looking good!

cron/monthly/host-settlement.js
znarf: This change is a bit out of context but I had problems re-conciliating some amounts and the rounding was ultimately the cause. This should ultimately be scrapped when we implement `HOST_FEE_SHARE`.

But something to keep in mind, when the `HOST_FEE_SHARE` is taken from Stripe with `application_fee`, it's rounded there and this is this precise value that we need to report.

Betree: Maybe we can avoid the previous rounding issue by setting a precision?

```suggestion
        t.amount = round(t.amount * ((t.data?.hostFeeSharePercent || hostFeeSharePercent) / 100), 2);
```

PR #6110, opened 2021-06-02
HostSettlement: Prevent invalid chars in CSV filename

Betree:
The previously used function, `generateKey`, can return some characters that are not valid for filenames:

```es6
> encodeBase64(randomBytes(8))
'WnX7MTjM8n4='
> encodeBase64(randomBytes(8))
'JYqo0MbPBs4='
> encodeBase64(randomBytes(8))
'qFe+iZF/X50='
```

We thus ended up with some files looking like `Open%20Collective%20Brussels-May-2021.xxxxxx/.csv` which, when downloaded, just appear as `.csv`.


kewitz:
Nice catch, thanks for the fix!

PR #6112, opened 2021-06-02
Add Support to Publish a Changelog Entry

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4325

Related Front end PR: https://github.com/opencollective/opencollective-frontend/pull/6447

### DIFF
--- a/server/graphql/v1/types.js
+++ b/server/graphql/v1/types.js
@@ -950,6 +950,12 @@ export const UpdateType = new GraphQLObjectType({
           return update.isPrivate;
         },
       },
+      isChangelog: {
+        type: GraphQLBoolean,
+        resolve(update) {
+          return update.isChangelog;
+        },
+      },
       notificationAudience: {
         type: UpdateAudienceTypeEnum,
         resolve(update) {
--- a/server/graphql/v1/types.js
+++ b/server/graphql/v1/types.js
@@ -951,7 +951,7 @@ export const UpdateType = new GraphQLObjectType({
         },
       },
       isChangelog: {
-        type: GraphQLBoolean,
+        type: new GraphQLNonNull(GraphQLBoolean),
         resolve(update) {
           return update.isChangelog;
         },
--- a/server/graphql/v2/input/UpdateCreateInput.js
+++ b/server/graphql/v2/input/UpdateCreateInput.js
@@ -10,6 +10,7 @@ export const UpdateCreateInput = new GraphQLInputObjectType({
   fields: () => ({
     title: { type: new GraphQLNonNull(GraphQLString) },
     isPrivate: { type: GraphQLBoolean },
+    isChangelog: { type: new GraphQLNonNull(GraphQLBoolean) },
     makePublicOn: { type: IsoDateString },
     html: { type: GraphQLString },
     fromAccount: { type: AccountReferenceInput },
--- a/server/graphql/v2/input/UpdateCreateInput.js
+++ b/server/graphql/v2/input/UpdateCreateInput.js
@@ -10,7 +10,7 @@ export const UpdateCreateInput = new GraphQLInputObjectType({
   fields: () => ({
     title: { type: new GraphQLNonNull(GraphQLString) },
     isPrivate: { type: GraphQLBoolean },
-    isChangelog: { type: new GraphQLNonNull(GraphQLBoolean) },
+    isChangelog: { type: GraphQLBoolean },
     makePublicOn: { type: IsoDateString },
     html: { type: GraphQLString },
     fromAccount: { type: AccountReferenceInput },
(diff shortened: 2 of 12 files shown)


server/graphql/v1/types.js
```diff
@@ -950,6 +950,12 @@ export const UpdateType = new GraphQLObjectType({
           return update.isPrivate;
         },
       },
+      isChangelog: {
+        type: GraphQLBoolean,
```
Betree, line 954:
```suggestion
        type: new GraphQLNonNull(GraphQLBoolean),
```

server/graphql/v2/input/UpdateCreateInput.js
```diff
@@ -10,6 +10,7 @@ export const UpdateCreateInput = new GraphQLInputObjectType({
   fields: () => ({
     title: { type: new GraphQLNonNull(GraphQLString) },
     isPrivate: { type: GraphQLBoolean },
+    isChangelog: { type: new GraphQLNonNull(GraphQLBoolean) },
```
Betree, line 13: I think it's fine to allow people to create updates without providing a value for `isChangelog` here since it only affects the team and defaults to `false`. That will also make sure that the form does not crash if someone tries to submit an update with an older frontend.

```suggestion
    isChangelog: { type: GraphQLBoolean },
```

Betree (approved):
Looking great! A unit test to assert that we don't send any emails when `isChangelog` is true would be nice

SudharakaP:
> Looking great! A unit test to assert that we don't send any emails when `isChangelog` is true would be nice

@Betree : Yes, I've added that and some more tests to cover the creation of changelogs. Let me know how it goes. Thanks a lot for all the reviewing. 😄

test/server/graphql/v2/mutation/UpdateMutations.test.js
```diff
@@ -109,9 +110,59 @@ describe('server/graphql/v2/mutation/UpdateMutations', () => {
...
+
+    it('fails if authenticated but cannot edit collective', async () => {
+      const result = await utils.graphqlQueryV2(createUpdateMutation, { update: changelogUpdate }, user1);
+      expect(result.errors).to.have.length(1);
+      expect(result.errors[0].message).to.equal("You don't have sufficient permissions to create an update");
+    });
+
+    it('creates a changelog update', async () => {
+      const result = await utils.graphqlQueryV2(createUpdateMutation, { update: changelogUpdate }, user3);
+      console.log(result);
```
Betree, line 151:
```suggestion
```

test/server/graphql/v2/mutation/UpdateMutations.test.js
```diff
@@ -109,9 +110,59 @@ describe('server/graphql/v2/mutation/UpdateMutations', () => {
...
+        isChangelog: true,
+        account: { legacyId: opencollective.id },
+      };
+    });
+    const createUpdateMutation = gqlV2/* GraphQL */ `
+      mutation CreateUpdate($update: UpdateCreateInput!) {
+        createUpdate(update: $update) {
+          id
+          slug
+          publishedAt
```
Betree, line 132:
```suggestion
          publishedAt
          isChangelog
```

We should make sure the created update has `isChangelog` === true

test/server/graphql/v2/mutation/UpdateMutations.test.js
```diff
@@ -109,9 +110,59 @@ describe('server/graphql/v2/mutation/UpdateMutations', () => {
...
+      const result = await utils.graphqlQueryV2(createUpdateMutation, { update: changelogUpdate }, user3);
+      console.log(result);
+      result.errors && console.error(result.errors[0]);
+      const createdUpdate = result.data.createUpdate;
+      expect(createdUpdate.slug).to.equal('monthly-changelog-update');
+      expect(createdUpdate.publishedAt).to.be.null;
+    });
+
+    it('do not send emails on changelog update', async () => {
+      expect(sendEmailSpy.callCount).to.equal(0);
```
Betree, line 159: Nice :+1:

Betree (approved):

test/server/graphql/v2/mutation/UpdateMutations.test.js
SudharakaP: Thanks. 😄

test/server/graphql/v2/mutation/UpdateMutations.test.js
SudharakaP: Ops, I've missed this. 😄

test/server/graphql/v2/mutation/UpdateMutations.test.js
SudharakaP: @Betree : Thanks for pointing it out. It's done. 😄

PR #6114, opened 2021-06-03
Separate Transactions for Host Fees: Core Support

znarf:
Behind a feature flag for now. Tests not updated.

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -3,6 +3,7 @@ import Promise from 'bluebird';
 import config from 'config';
 import debugLib from 'debug';
 import { find, get, includes, isNil, isNumber, omit, pick } from 'lodash';
+import { v4 as uuid } from 'uuid';
 
 import activities from '../constants/activities';
 import status from '../constants/order_status';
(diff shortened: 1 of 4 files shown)


server/models/Transaction.js
```diff
@@ -690,6 +688,42 @@ function defineModel() {
     return { transaction, platformTipTransaction, platformTipDebtTransaction };
   };
 
+  Transaction.createHostFeeTransactions = async ({ transaction, host }) => {
```
Betree, line 691: In `createPlatformTipTransactions`, we now pass individual params

```suggestion
  Transaction.createHostFeeTransactions = async (transaction, host) => {
```

server/lib/payments.js
```diff
@@ -122,6 +124,53 @@ export function calcFee(amount, fee) {
   return Math.round((amount * fee) / 100);
 }
 
+export const buildRefundForTransaction = (t, user, data, refundedPaymentProcessorFee) => {
```
Betree, line 127: Extracted this one since we're going to need it for the script

PR #6115, opened 2021-06-03
Introduce validateContributionPayload and related refactor

znarf:


Betree (commented):
Adding these validations really simplifies the function :+1:

Betree (approved):

PR #6123, opened 2021-06-07
Script to migrate host fee for existing transactions

Betree:
Require https://github.com/opencollective/opencollective-api/pull/6114


znarf (approved):
Good for me!

PR #6125, opened 2021-06-08
Make sure only root users can create changelog entries

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4325

Thanks for pointing this out @Betree. 😉

### DIFF
--- a/server/graphql/common/update.js
+++ b/server/graphql/common/update.js
@@ -30,7 +30,7 @@ export async function createUpdate(_, args, req) {
     throw new Error('This collective does not exist');
   } else if (!req.remoteUser.isAdminOfCollective(collective)) {
     throw new Forbidden("You don't have sufficient permissions to create an update");
-  } else if (args.update?.isChangelog && !req.remoteUser.isRoot()) {
+  } else if (args.update.isChangelog && !req.remoteUser.isRoot()) {
     throw new Forbidden('Only root users can create changelog updates.');
   }
 
--- a/server/graphql/common/update.js
+++ b/server/graphql/common/update.js
@@ -30,6 +30,8 @@ export async function createUpdate(_, args, req) {
     throw new Error('This collective does not exist');
   } else if (!req.remoteUser.isAdminOfCollective(collective)) {
     throw new Forbidden("You don't have sufficient permissions to create an update");
+  } else if (args.update?.isChangelog && !req.remoteUser.isRoot()) {
+    throw new Forbidden('Only root users can create changelog updates.');
   }
 
   const update = await models.Update.create({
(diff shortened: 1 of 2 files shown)


server/graphql/common/update.js
```diff
@@ -30,6 +30,8 @@ export async function createUpdate(_, args, req) {
     throw new Error('This collective does not exist');
   } else if (!req.remoteUser.isAdminOfCollective(collective)) {
     throw new Forbidden("You don't have sufficient permissions to create an update");
+  } else if (args.update?.isChangelog && !req.remoteUser.isRoot()) {
```
Betree, line 33: :ice_cream: We already made sure the update exists at this point

```suggestion
  } else if (args.update.isChangelog && !req.remoteUser.isRoot()) {
```

Betree (approved):
Looking great! Looks like we need to adapt a test as well

SudharakaP:
> Looking great! Looks like we need to adapt a test as well

@Betree : Thanks, I've missed the test. Just corrected that. 😄

Betree:
> Why decide that "only root users can create changelog entries" instead of "changelog entries can only be created on the opencollective account"?

@znarf We may want to enable posting changelog entries from https://opencollective.com/engineering/updates (and similar) in the future, so that makes things a bit more flexible.

SudharakaP:
> > Why decide that "only root users can create changelog entries" instead of "changelog entries can only be created on the opencollective account"?
> 
> @znarf We may want to enable posting changelog entries from https://opencollective.com/engineering/updates (and similar) in the future, so that makes things a bit more flexible.

@Betree : Correct me if I am wrong @Betree , but my thinking is that on the API side we didn't have a check; so when creating a changelog update we only used to check on the front-end side whether the collective is indeed `opencollective`. So adding root user on the api makes sense for now, until we add this support for adding changelog updates for other collectives such as engineering. So I'll keep this as is for now? 🤔

Betree:
> > > Why decide that "only root users can create changelog entries" instead of "changelog entries can only be created on the opencollective account"?
> > 
> > 
> > @znarf We may want to enable posting changelog entries from https://opencollective.com/engineering/updates (and similar) in the future, so that makes things a bit more flexible.
> 
> @Betree : Correct me if I am wrong @Betree , but my thinking is that on the API side we didn't have a check; so when creating a changelog update we only used to check on the front-end side whether the collective is indeed `opencollective`. So adding root user on the api makes sense for now, until we add this support for adding changelog updates for other collectives such as engineering. So I'll keep this as is for now? thinking

I think so. For me, the main goal of this was purely to have a security fix, we don't want anyone to be able to publish to our changelog. After that, I'm not strongly for one approach or the other.

SudharakaP:
> I think so. For me, the main goal of this was purely to have a security fix, we don't want anyone to be able to publish to our changelog. 

@Betree : Exactly my thoughts; 

So @znarf if you don't have any objections I'll be merging this one.

PR #6133, opened 2021-06-08
Hot Fix: make Platform Tip calculations more solid

znarf:


Betree (approved):
Looks good, would be great to consolidate old transactions affected by this issue in the future

PR #6137, opened 2021-06-09
PayPal: Correct currency for one-time contributions

Betree:
For recurring contributions, PayPal passes the currency as a `currency` parameter. But for one-time captures, they use the `currency_code` parameter. We've therefore recorded one-time contributions (~150) in `USD` rather than their real currency.

This is an immediate fix for the issue, as it's currently preventing the transactions from being recorded in the ledger because of the new checks we've put in place. I'm working on a migration to fix past data.

### DIFF
--- a/server/paymentProviders/paypal/webhook.ts
+++ b/server/paymentProviders/paypal/webhook.ts
@@ -83,9 +83,6 @@ const loadSubscriptionForWebhookEvent = async (req: Request, subscriptionId: str
 };
 
 async function handleSaleCompleted(req: Request): Promise<void> {
-  // TODO During the internal testing phase, we're logging all webhooks events to make debugging easier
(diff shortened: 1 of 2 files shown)


server/paymentProviders/paypal/webhook.ts
```diff
@@ -83,9 +83,6 @@ const loadSubscriptionForWebhookEvent = async (req: Request, subscriptionId: str
 };
 
 async function handleSaleCompleted(req: Request): Promise<void> {
-  // TODO During the internal testing phase, we're logging all webhooks events to make debugging easier
-  logger.info(`PayPal webhook (PAYMENT.SALE.COMPLETED): ${JSON.stringify(req.body)}`);
```
Betree, line 87: Took the opportunity to remove these logs, storing the object in `data` is now enough

PR #6140, opened 2021-06-09
Notify host admins when creating the settlement invoice

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4055


znarf (approved):
Looks good.

Betree (commented):
Looks good! Is the email looking ok, given that we don't provide a `remoteUser` to `createActivity`?

kewitz:
@betree, yes because `expense.createActivity()` actually uses `expense.UserId` to create the activity.

PR #6146, opened 2021-06-10
Fix Sequelize transaction error when creating Activities

kewitz:


server/models/Transaction.js
```diff
@@ -252,8 +252,8 @@ function defineModel() {
       },
 
       hooks: {
-        afterCreate: (transaction, options) => {
-          Transaction.createActivity(transaction, { transaction: options.transaction });
+        afterCreate: transaction => {
+          Transaction.createActivity(transaction);
```
kewitz, line 256: We can't use a SQL transaction here since it is an async call.

PR #6147, opened 2021-06-10
Add Support for Notifications for Changelog Updates

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4345

Related front-end PR: https://github.com/opencollective/opencollective-frontend/pull/6476


server/graphql/v1/mutations.js
```diff
@@ -208,6 +208,18 @@ const mutations = {
       return updateUserEmail(remoteUser, email);
     },
   },
+  setChangelogViewDate: {
+    type: UserType,
+    description: 'Update the time which the user viewed the changelog updates',
+    args: {
+      changelogViewDate: {
+        type: new GraphQLNonNull(GraphQLString),
```
Betree, line 216: To provide better validations

```suggestion
        type: new GraphQLNonNull(IsoDateString),
```

server/graphql/v1/types.js
```diff
@@ -241,6 +241,12 @@ export const UserType = new GraphQLObjectType({
           return user.data && user.data.features && user.data.features.ALL === false;
         },
       },
+      changelogViewDate: {
+        type: DateString,
```
Betree, line 245: Haven't used these in quite some time (GQLV1) but I think `IsoDateString` better respects standards

```suggestion
        type: IsoDateString,
```

server/graphql/v2/interface/Account.js
```diff
@@ -322,6 +323,24 @@ const accountFieldsDefinition = () => ({
       return { nodes: result.rows, totalCount: result.count, limit, offset };
     },
   },
+  latestChangelogPublishDate: {
```
Betree, line 326: In my opinion, having this on `Account` does not fit: 99,99% of accounts are not concerned by the changelog. It's a platform feature, so I would add this as a top-level query (in `server/graphql/v2/query/index.js`) rather than a field in Account. @znarf any opinion?

server/graphql/v2/interface/Account.js
```diff
@@ -322,6 +323,24 @@ const accountFieldsDefinition = () => ({
...
+      const latestChangelogPublishDate = await models.Update.findAll({
+        where: {
+          CollectiveId: collective.id,
+          isChangelog: true,
+        },
+        attributes: [[sequelize.fn('max', sequelize.col('publishedAt')), 'date']],
+        raw: true,
+      });
+
+      if (latestChangelogPublishDate && !latestChangelogPublishDate.isEmpty) {
```
Betree, line 338: Is this `isEmpty` provided by sequelize?

Betree (commented):
Looks good! Some minor feedback on the GraphQL types. We may also want to add some caching on `latestChangelogPublishDate` since it's going to be fetched intensively (at least each sign-in). I'm not sure what's the cost of this SQL query.

server/graphql/v2/interface/Account.js
SudharakaP: @Betree : Ops, my mistake. This should be, `if (latestChangelogPublishDate && latestChangelogPublishDate.length > 0) {` 😮

server/graphql/v2/interface/Account.js
SudharakaP: @Betree : Yes I was having doubts when I put this on Accounts.js, now it makes sense. I've moved it to the `index.js`. 😄

SudharakaP:
> Looks good! Some minor feedback on the GraphQL types. We may also want to add some caching on `latestChangelogPublishDate` since it's going to be fetched intensively (at least each sign-in). I'm not sure what's the cost of this SQL query.

@Betree : Caching added, thanks for pointing it out. 🐨 I've made it such that this value is retained in cache for 12 hours; I hope this is okay. 🤔 Let me know otherwise.

Betree:
> > Looks good! Some minor feedback on the GraphQL types. We may also want to add some caching on `latestChangelogPublishDate` since it's going to be fetched intensively (at least each sign-in). I'm not sure what's the cost of this SQL query.
> 
> @Betree : Caching added, thanks for pointing it out. koala I've made it such that this value is retained in cache for 12 hours; I hope this is okay. thinking Let me know otherwise.

Seems like a good value! A good strategy would be to invalidate this cache whenever a new changelog entry gets posted

SudharakaP:
> Seems like a good value! A good strategy would be to invalidate this cache whenever a new changelog entry gets posted

@Betree : Thanks. I've added a change to make sure the cache is deleted when a new change-log entry is published and also increased the interval of cache retention for this to 24 hours. Hope that's okay, since there's not much point of invalidating the cache at a lower interval since this value currently only changes when a new change-log entry is added. 🤔 🐨

migrations/dev-20210610-add-users-changelogViewDate-column.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, DataTypes) => {
+    await queryInterface.addColumn('Users', 'changelogViewDate', {
+      type: DataTypes.DATE,
+      defaultValue: DataTypes.NOW,
```
znarf, line 7: Why a `defaultValue` of NOW?

server/graphql/v1/mutations.js
```diff
@@ -208,6 +209,18 @@ const mutations = {
       return updateUserEmail(remoteUser, email);
     },
   },
+  setChangelogViewDate: {
```
znarf, line 212: Any reason to use API v1 for this? All new features should usually be on API v2.

server/graphql/v2/query/LatestChangelogPublishDateQuery.js
```diff
@@ -0,0 +1,41 @@
...
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'The slug identifying the collective that requested the changelog publish date',
+    },
+  },
+  async resolve(_, args) {
+    const cacheKey = 'latest_changelog_publish_date';
+    let latestChangelogPublishDate = await cache.get(cacheKey);
+    if (!latestChangelogPublishDate) {
+      const collectiveId = await fetchCollectiveId(args.collectiveSlug);
+      latestChangelogPublishDate = await models.Update.findAll({
```
znarf, line 21: `findOne` is what you want to use here, not `findAll`

server/graphql/v2/query/LatestChangelogPublishDateQuery.js
```diff
@@ -0,0 +1,41 @@
...
+import { GraphQLDateTime } from 'graphql-iso-date';
+
+import cache, { fetchCollectiveId } from '../../../lib/cache';
+import sequelize from '../../../lib/sequelize';
+import models from '../../../models';
+
+const LatestChangelogPublishDateQuery = {
+  type: GraphQLDateTime,
+  args: {
+    collectiveSlug: {
```
znarf, line 11: Doesn't make sense to have a parameter here (`collectiveSlug`) and then hardcode a cache key that doesn't take into account the parameter. This needs to be reviewed.

server/graphql/v2/query/index.js
```diff
@@ -32,6 +33,7 @@ const query = {
   transactions: TransactionsQuery,
   update: UpdateQuery,
   paypalPlan: PaypalPlanQuery,
+  latestChangelogPublishDate: LatestChangelogPublishDateQuery,
```
znarf, line 36: I find this query too specific and would like to challenge the design here.

Better use a generic query that returns a collection of updates and:
- make sure that there is a boolean filter to request changelog entries
- make sure that they are ordered ante-chronologically by default 
- make sure that there is pagination and one can ask a limit of 1

Then you have exactly what you need with a GraphQL query:

```
 query LatestChangelogPublishDateQuery($slug: String!) {
    account(slug: $slug) {
     updates(isChangelog: true, limit: 1) {
        nodes {
           id
          createdAt
    }
```

or something like:


```
 query LatestChangelogPublishDateQuery($account: AccountReferenceInput!) {
     updates(account: $account, isChangelog: true, limit: 1) {
        nodes {
           id
          createdAt
    }
```

znarf (changes requested):

server/graphql/v2/interface/Account.js
znarf: I proposed in another comment to rethink the design and try to better use our generic GraphQL model.

migrations/dev-20210610-add-users-changelogViewDate-column.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, DataTypes) => {
+    await queryInterface.addColumn('Users', 'changelogViewDate', {
```
znarf, line 5: I find it a bit overkill to add a plain SQL column to store this. Could have been stored in `data` at this stage.

Personal opinion, not asking for a change for now.

migrations/dev-20210610-add-users-changelogViewDate-column.js
SudharakaP: Not strictly needed, I just added it as I thought having a default value here makes sense than a null value, but thinking again now having a null makes sense. 👍🏽

server/graphql/v2/query/LatestChangelogPublishDateQuery.js
SudharakaP: @znarf : Ah yes, don't think there's a need to pass the parameter to grab the latest publish date. I've removed this and refactored the code. 👍🏽 🐨

server/graphql/v1/mutations.js
SudharakaP: @znarf : My thinking was that all the user related mutations are on v1. I thought it's best to keep this with that until we migrate all the v1 mutations to v2 related to the user. But I am okay with adding this mutation to v2 as well. 🤔

server/graphql/v2/query/index.js
SudharakaP: @znarf : Thanks for the input. A small doubt I have with this approach is that the caching of the latest change log publish date. With this approach, we could cache the result of this query based on the limit being equal to 1 and that it's a changelog update and also checking for the collective (should be `opencollective`). But down the line when we add the publishing updates as a collective feature (say `Open Collective Engineering` or any other collective) we will have to come back here and add more conditions for the caching to work. But I might be wrong here, in which case please let me know and I'll change it to this approach. 🤔 🐨

Betree: `latestChangelogPublishDate` is indeed a good way to optimize this query that is going to be called very often. Calling `updates` with a `limit: 1` would be more expensive, and more difficult in the Frontend to update Apollo's cache when the changelog entry is viewed by the user. Since the feature is going to stay, I don't think it's a bad thing to have specific queries for it.

As an alternative design, we can eventually move this field to `UserType`/`Individual`, as something like `hasSeenLatestChangelogEntry`. That way we can keep a caching strategy, while reducing the number of fields since we wouldn't need to have a `User.changelogViewDate` anymore on the GraphQL API.

```es6
const UserType = {
  ...,
  fields : {
    ...,
    hasSeenLatestChangelogEntry: {
      ...,
      resolve(user, _, req) {
        const latestChangelogPublishDate = getValueFromCache();
        return user.changelogViewDate >= latestChangelogPublishDate;
      }
    }
  }
}
```

SudharakaP: @znarf : So as per Benjamin's comment I will leave this as it is for now if you are okay with it. But let me know if you disagree and we can work towards a different approach. 😄 🐨

server/graphql/v1/mutations.js
znarf: Thank you for sharing the rationale.

I definitely advice to not introduce it in GraphQL v1 and properly implement it in GraphQL v2.

server/graphql/v2/query/LatestChangelogPublishDateQuery.js
```diff
@@ -0,0 +1,31 @@
...
+import models from '../../../models';
+
+const LatestChangelogPublishDateQuery = {
+  type: GraphQLDateTime,
+  async resolve() {
+    const cacheKey = 'latest_changelog_publish_date';
+    let latestChangelogPublishDate = await cache.get(cacheKey);
+    if (!latestChangelogPublishDate) {
+      const collectiveId = await fetchCollectiveId('opencollective');
+      latestChangelogPublishDate = await models.Update.findOne({
```
znarf, line 14: The variable should not be named `latestChangelogPublishDate` if it doesn't contain a date.

znarf: You see the problem when you read this `latestChangelogPublishDate?.date`. Why read the date of a date?

server/models/User.js
```diff
@@ -105,6 +105,11 @@ function defineModel() {
         type: DataTypes.ARRAY(DataTypes.STRING),
         allowNull: true,
       },
+      changelogViewDate: {
+        type: DataTypes.DATE,
+        defaultValue: DataTypes.NOW,
```
znarf, line 110: Why a defaultValue of NOW?

server/graphql/v2/query/LatestChangelogPublishDateQuery.js
```diff
@@ -0,0 +1,31 @@
...
+import sequelize from '../../../lib/sequelize';
+import models from '../../../models';
+
+const LatestChangelogPublishDateQuery = {
+  type: GraphQLDateTime,
+  async resolve() {
+    const cacheKey = 'latest_changelog_publish_date';
+    let latestChangelogPublishDate = await cache.get(cacheKey);
+    if (!latestChangelogPublishDate) {
+      const collectiveId = await fetchCollectiveId('opencollective');
```
znarf, line 13: I forgot, are we supporting posts to `engineering` and `design` collectives at this stage?

server/graphql/v1/mutations.js
SudharakaP: @znarf : Yep, just moved it to GraphQL v2. 😄

server/graphql/v2/mutation/UserMutations.js
```diff
@@ -0,0 +1,28 @@
+import { GraphQLNonNull } from 'graphql';
+import { GraphQLDateTime } from 'graphql-iso-date/dist';
```
znarf, line 2: Any reason for the `/dist`? We're not doing this elsewhere.

server/models/User.js
SudharakaP: Yes, this is also related to, https://github.com/opencollective/opencollective-api/pull/6147#discussion_r651190923. I've corrected it. 😯

server/graphql/v2/query/LatestChangelogPublishDateQuery.js
SudharakaP: Thanks. Corrected it. 🐨

server/graphql/v2/object/User.js
```diff
@@ -0,0 +1,130 @@
...
+
+import models, { Op } from '../../../models';
+
+import { Collective } from './Collective';
+import { Member } from './Member';
+
+export const User = new GraphQLObjectType({
+  name: 'UserDetails',
+  description: 'This represents the details of a User',
+  fields: () => {
```
znarf, line 12: Please don't introduce any field not required by the current work. Or the review process is going out of track.

server/graphql/v2/query/LatestChangelogPublishDateQuery.js
SudharakaP: @znarf : The plan is on this issue to support posts from `opencollective`; the task for supporting the `engineering` and `design` collectives will be handled at a later issue; this probably needs reviewing I think, but I'll create an issue for it soon. 🐨

server/graphql/v2/mutation/UserMutations.js
SudharakaP: @znarf : Yes, you are right. The `dist` shouldn't be there. I've corrected that. 😄

server/graphql/v2/object/User.js
SudharakaP: @znarf : Ah, I somehow initially thought we need the whole object, but I don't think we do; we only need the fields we try to query from the frontend. Thus I have adjusted this. 😄 Hope now it looks cleaner. 😄

SudharakaP:
@znarf : As per our discussion, I've changed the PR a bit to remove creating a `User` object in GraphQl v2 and to use the `Individual` instead. Let me know if everything looks good now. One point to mention is that in this PR, I've added the field; 

https://github.com/opencollective/opencollective-api/blob/28cb3073a04259c67856473436cf70aefc4d7f6f/server/graphql/v1/types.js#L244-L248

to GraphQl v1 side. I think this is necessary as the query we use to get the User object on the homepage is still using v1. Maybe in a future PR when we move the user object to v2, we could get rid of this. Hope that's okay? 🤔 🐨

server/graphql/v2/mutation/UserMutations.js
```diff
@@ -0,0 +1,30 @@
...
+    },
+    resolve: async (_, { changelogViewDate }, { remoteUser }) => {
+      if (!remoteUser) {
+        throw new Unauthorized();
+      } else if (!changelogViewDate) {
+        throw new ValidationFailed('The change log view date must be set');
+      }
+      const date = new Date(changelogViewDate);
+      const user = await remoteUser.update({ changelogViewDate: date });
+      return await fetchAccountWithReference({ id: user.CollectiveId });
```
Betree, line 25: Not needed

```suggestion
      return fetchAccountWithReference({ id: user.CollectiveId });
```

server/graphql/v2/query/index.js
Betree: Let's wait for @znarf's feedback before closing this one

Betree (commented):
Nice progress!

server/graphql/v2/query/index.js
znarf: @Betree @SudharakaP I generally prefer my proposed approach but you can move forward with the current one if you're unconvinced. Caching not that a big deal for me and could also be applied to the collection if it was necessary.

The other approach from Ben is interesting and could be a compromise.

Generally, as all options are not introducing much debt, we can later change our mind and iterate.

server/graphql/v2/mutation/UserMutations.js
znarf: Works but using fetchAccountWithReference is not proper, we can simply do Collective.findByPk.

`fetchAccountWithReference` main use case is when we pass it a `AccountReferenceInput`

znarf: Or even better `user.getCollective()`

server/graphql/v2/object/Individual.js
```diff
@@ -107,6 +108,15 @@ export const Individual = new GraphQLObjectType({
           }
         },
       },
+      changelogViewDate: {
+        type: GraphQLDateTime,
+        async resolve(collective) {
+          const user = await models.User.findOne({
+            where: { CollectiveId: collective.id },
+          });
```
znarf, line 116: `const user = await collective.getUser()`

server/graphql/v2/query/LatestChangelogPublishDateQuery.js
```diff
@@ -0,0 +1,31 @@
...
+    const cacheKey = 'latest_changelog_publish_date';
+    let latestChangelogUpdate = await cache.get(cacheKey);
+    if (!latestChangelogUpdate) {
+      const collectiveId = await fetchCollectiveId('opencollective');
+      latestChangelogUpdate = await models.Update.findOne({
+        where: {
+          CollectiveId: collectiveId,
+          isChangelog: true,
+        },
+        attributes: [[sequelize.fn('max', sequelize.col('publishedAt')), 'date']],
```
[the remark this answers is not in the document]
SudharakaP: Done. 👍🏽

server/graphql/v2/object/Individual.js
SudharakaP: Thanks. Done. 👍🏽

server/graphql/v2/mutation/UserMutations.js
SudharakaP: Thanks for the feedback. Done. 👍🏽

server/graphql/v2/mutation/UserMutations.js
```diff
@@ -0,0 +1,30 @@
...
+        type: new GraphQLNonNull(GraphQLDateTime),
+      },
+    },
+    resolve: async (_, { changelogViewDate }, { remoteUser }) => {
+      if (!remoteUser) {
+        throw new Unauthorized();
+      } else if (!changelogViewDate) {
+        throw new ValidationFailed('The change log view date must be set');
+      }
+      const date = new Date(changelogViewDate);
```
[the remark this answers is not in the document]
SudharakaP: Ops, I missed that. You are right. Done. 👍🏽

server/graphql/v2/mutation/UserMutations.js
```diff
@@ -0,0 +1,30 @@
...
+    description: 'Update the time which the user viewed the changelog updates',
+    args: {
+      changelogViewDate: {
+        type: new GraphQLNonNull(GraphQLDateTime),
+      },
+    },
+    resolve: async (_, { changelogViewDate }, { remoteUser }) => {
+      if (!remoteUser) {
+        throw new Unauthorized();
+      } else if (!changelogViewDate) {
```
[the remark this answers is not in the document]
SudharakaP: Thanks. Done. 👍🏽

server/graphql/v2/query/index.js
SudharakaP: @znarf @Betree : I tried moving the field `latestChangelogPublishDate` but it seems that this is not a straightforward change on the front-end side. The problem here is that we use GraphQl v1 for getting the `LoggedInUser`. And we need this to pass the `slug` (or some identifier such as `id`) for the `Individual` query. This creates some problems since we will have to first wait until the `LoggedInUser` is fully loaded and then try to get the `Individual`. I couldn't figure out a nice way of doing this. Ideally, I think we should get this user object through one single API in a single pass but that will involve migrating the User query in the front-end from GraphQl V1 to V2. Let me know what you think. 🤔 🐨 

I've opened two draft PR's to demonstrate what I mean; https://github.com/opencollective/opencollective-api/pull/6167 and https://github.com/opencollective/opencollective-frontend/pull/6501. 🤔

Note that, the problem here is, https://github.com/opencollective/opencollective-frontend/blob/eb17345c134bc0ab87af47fe9d38c2b76ec55788/components/TopBar.js#L344-L351 which depends on the LoggedInUser.

Betree: I think the simplest is to have this `hasSeenLatestChangelogEntry` field on both `Individual` and `UserType`. Retrieving the info from `LoggedInUser` will be much simpler. But in any case, you can fetch the logged-in individual from GQLV2 using `loggedInAccount`

SudharakaP: @Betree : Thanks. That was helpful. 🙏🏽  

@znarf @Betree : I've handled this in the other two PRs, https://github.com/opencollective/opencollective-api/pull/6167 and https://github.com/opencollective/opencollective-frontend/pull/6501 so that this don't get too much polluted and probably easier to review. Thanks and let me know if you find any other issues. 🐨 😄

SudharakaP:
I am closing this in favor of https://github.com/opencollective/opencollective-api/pull/6167 as per @Betree 's comment; https://github.com/opencollective/opencollective-api/pull/6167#pullrequestreview-687449715. We have addressed all code reviews that are mentioned here except of https://github.com/opencollective/opencollective-api/pull/6147#discussion_r651202662 which is addressed in the new PR. 😄

PR #6149, opened 2021-06-11
VirtualCard: Support refund transactions

kewitz:
This covers the missing edge case of refunded virtual card charges.
Notice I'm only creating the transaction pair here, I don't think it makes sense to create an expense with a negative value.

### DIFF
--- a/server/paymentProviders/privacy/index.ts
+++ b/server/paymentProviders/privacy/index.ts
@@ -154,7 +154,7 @@ const processTransaction = async (
   });
 
   if (
-    expense && // Expense is undefined if the the transaction is a refund
+    !isRefund && // There will be no expense for refunds
     collective.settings?.ignoreExpenseMissingReceiptAlerts !== true
   ) {
     expense
--- a/server/paymentProviders/privacy/index.ts
+++ b/server/paymentProviders/privacy/index.ts
@@ -14,10 +14,10 @@ import VirtualCardModel from '../../models/VirtualCard';
 import { Transaction } from '../../types/privacy';
 import { CardProviderService } from '../types';
 
-const createExpense = async (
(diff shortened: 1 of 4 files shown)


server/paymentProviders/privacy/index.ts
```diff
@@ -14,10 +14,10 @@ import VirtualCardModel from '../../models/VirtualCard';
 import { Transaction } from '../../types/privacy';
 import { CardProviderService } from '../types';
 
-const createExpense = async (
+const processTransaction = async (
   privacyTransaction: Transaction,
   opts?: { host?: any; collective?: any; hostCurrencyFxRate?: number },
-): Promise<any> => {
+): Promise<typeof models.Expense | undefined> => {
   const amount = privacyTransaction.settled_amount;
```
znarf, line 21: What is our guarantee that `settled_amount` is in USD here. Is it really always the case? Even if the charge is in another currency?

server/paymentProviders/privacy/index.ts
```diff
@@ -125,7 +153,10 @@ const createExpense = async (
     return expense;
   });
 
-  if (collective.settings?.ignoreExpenseMissingReceiptAlerts !== true) {
+  if (
+    expense && // Expense is undefined if the the transaction is a refund
```
znarf, line 157: You have `isRefund` available as a variable, maybe it would be more explicit to use it?

znarf (approved):
I'm good with the change.

@Betree have a look so you have the context about this behavior.

server/paymentProviders/privacy/index.ts
kewitz: Yes, they don't even list currency on their schema.

PR #6152, opened 2021-06-14
GQLV1: Add support for HOST_FEE as separate transactions

Betree:


znarf (approved):
Seems a lot of code for a small retro-compatibility but there was not other way I guess.

A bit concerned about the performance, but if it's only used in a few places then it's fine.

Hopefully we should be able to remove that support in the future.

server/graphql/v1/queries.js
```diff
@@ -295,6 +296,7 @@ const queries = {
         startDate: args.dateFrom,
         endDate: args.dateTo,
         includeExpenseTransactions: args.includeExpenseTransactions,
+        kinds: args.kinds,
```
znarf, line 299: If it was API v2, I think I would insist on making it singular `kind` (still allowing a list).

I'm more relax on GraphQL v1 and change is not required.

Betree:
> Seems a lot of code for a small retro-compatibility but there was not other way I guess.
> 
> A bit concerned about the performance, but if it's only used in a few places then it's fine.
> 
> Hopefully we should be able to remove that support in the future.

90% of the code is actually the loader, which we can probably reuse in the future as we'll probably want to be able to link transactions and their host fees in one way or another.

server/graphql/v1/queries.js
Betree: The same `kinds` field actually exists on GQLV2: https://github.com/opencollective/opencollective-api/blob/main/server/graphql/v2/query/TransactionsQuery.ts#L81. Having a plural here makes sense to me since we only accept a list. It's similar to what we do with the `tags` argument for example.

znarf: I think the DX is easier with singular, no need to think if it should be singular or plural, just use the name of field. That way, you can also accept a single element without feeling weird. In GraphQL, when the argument is a list, you can pass a single item, which is nice. In this case that would be`kinds: `PLATFORM_TIP`. With singular, it's also closer to its Sequelize / SQL equivalent.

This was a conscious design in the early days of GraphQL v2:

https://github.com/opencollective/opencollective-api/blob/main/server/graphql/v2/interface/Account.js#L155
https://github.com/opencollective/opencollective-api/blob/main/server/graphql/v2/interface/Account.js#L204

I haven't been vigilant enough and I should have catch `types` and `tags` when they were introduced.

PR #6156, opened 2021-06-15
Move to PLATFORM_TIP_DEBT

Betree:


znarf:
Let's try to ship this and start making debts visible on the frontend (hidden by default, I already have use case for it when interacting with Hosts on support).

Betree:
> Let's try to ship this and start making debts visible on the frontend (hidden by default, I already have use case for it when interacting with Hosts on support).

Alright, I'll prioritize next week

migrations/20210615074327-platform-tip-debt-kind.js
```diff
@@ -0,0 +1,20 @@
...
+  },
+
+  down: async queryInterface => {
+    await queryInterface.sequelize.query(`
+      UPDATE "Transactions"
+      SET "kind" = 'PLATFORM_TIP'
+      WHERE "kind" = 'PLATFORM_TIP_DEBT'
+    `);
+  },
+};
```
znarf, line 20: Where are we adding `PLATFORM_TIP_DEBT` to the `enum_Transactions_kind`?

Betree: It still WIP

znarf (approved):

PR #6157, opened 2021-06-15
PayPal: Handle pending captures

Betree:
Part of https://github.com/opencollective/opencollective/issues/4368

Handle `PAYMENT.CAPTURE.COMPLETED` webhook action to record payments that take some time to be confirmed (that seems to be the case mostly for bank transfer), and make sure we never record transactions unless a payment is indeed confirmed.

:warning: We'll need to run `scripts/paypal/update-hosts-webhooks.ts` to update webhooks with this new action after deploying


server/paymentProviders/paypal/webhook.ts
```diff
@@ -114,6 +114,47 @@ async function handleSaleCompleted(req: Request): Promise<void> {
...
+        association: 'paymentMethod',
+        required: true,
+        where: { service: 'paypal', type: 'payment' },
+      },
+    ],
+  });
+
+  if (!order) {
+    logger.debug(`No pending order found for capture ${capture.id}`);
+    return;
```
kewitz, line 139: 👍🏼

kewitz (approved):
Looks good! I think it covers everything. :shipit:

server/paymentProviders/paypal/webhook.ts
```diff
@@ -114,6 +114,47 @@ async function handleSaleCompleted(req: Request): Promise<void> {
...
+  }
+
+  // 2. Validate webhook event
+  const host = await order.collective.getHostCollective();
+  const paypalAccount = await getPaypalAccount(host);
+  await validateWebhookEvent(paypalAccount, req);
+
+  // 3. Record the transaction
+  const transaction = await recordPaypalCapture(order, capture);
+  await order.update({ processedAt: new Date(), status: 'PAID' });
```
SudharakaP, line 149: Minor issue, use `PAID` from `order_status.js` maybe? 🙂

server/paymentProviders/paypal/webhook.ts
```diff
@@ -114,6 +114,47 @@ async function handleSaleCompleted(req: Request): Promise<void> {
...
+
+  // 4. Send thankyou email
+  await sendThankYouEmail(order, transaction);
+
+  // 5. Register user as a member, since the transaction is not created in `processOrder`
+  await order.getOrCreateMembers();
```
SudharakaP, line 155: Maybe alternate the two calls, first add them as member and then send the thank you email? I normally like to send thank you email at the end since that way if the sending of email throws an exception for some reason the user will still be added as a member. 🤔

Betree: I would say the opposite: sending the thank you email is more important than adding the members

SudharakaP: Great. In that case it's fine. In my mind somehow I thought adding backers is more important. 😄

SudharakaP (approved):

PR #6160, opened 2021-06-15
Add a dedicated test for the ledger

Betree:


server/lib/budget.js
```diff
@@ -134,6 +134,7 @@ export async function getTotalMoneyManagedAmount(host, { startDate, endDate, cur
 
   const results = await sumCollectivesTransactions(ids, {
     startDate,
+    excludeRefunds: false,
```
Betree, line 137: This was needed to fix the following tests:

```es6
      expect(await collective.getBalance()).to.eq(0);
      expect(await host.getBalance()).to.eq(-1200); // Will be -200 after settlement (platform tip)
      expect(await host.getTotalMoneyManaged()).to.eq(-1200);
```

Without that, `host.getTotalMoneyManaged()` returned `-1000`

server/paymentProviders/opencollective/manual.js
```diff
@@ -53,7 +53,7 @@ async function processOrder(order) {
 
   const isFeesOnTop = order.data?.isFeesOnTop || false;
 
-  const paymentProcessorFeeInHostCurrency = 0;
+  const paymentProcessorFeeInHostCurrency = order.data?.paymentProcessorFeeInHostCurrency || 0;
```
Betree, line 56: Added this mainly for the tests, but it made sense to add the ability to set payment processor fees on manual payments (eg. to reflect bank transfer fees)

Betree:
Closing, I want to rebase in a separate PR and keep the existing history, esp. to not lose what was done in https://github.com/opencollective/opencollective-api/pull/6160/commits/9685dba647fb18aa5c36d56f432611a8199cb59e

PR #6162, opened 2021-06-15
Fix Privacy duplicated Transactions when reconciling refunds

kewitz:
Since we use expenses to detect if the transaction was already reconciled and we're not creating expenses for refunds, there's an edge case that will cause the worker to create multiple refund transactions on the ledger.

I'm solving this by updating the `processTransaction()` method to check our table for an existing transaction with repeated id (in Privacy's context defined by token) if the charge we're reconciling is a refund.

This could also be solved at the worker level by updating how we paginate the transaction request using information stored in the `Expense.data` and the reason why I didn't do that is due to the fact we are planning to implement a webhook interface to sync these transactions.

### DIFF
--- a/migrations/20210615130700-index-transaction-privacy-id.js
+++ b/migrations/20210615130700-index-transaction-privacy-id.js
@@ -1,0 +1,15 @@
+'use strict';
+
+module.exports = {
+  up: async queryInterface => {
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY privacy_transfer_id ON "Transactions" (((data ->> 'token')::text)) WHERE (data ->> 'token') IS NOT NULL;
+    `);
+  },
+
+  down: async queryInterface => {
+    await queryInterface.sequelize.query(`
+      DROP INDEX CONCURRENTLY IF EXISTS privacy_transfer_id;
+    `);
+  },
+};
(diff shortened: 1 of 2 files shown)


SudharakaP (approved):
Looks good. 💯

kewitz:
Rebased.

migrations/20210615130700-index-transaction-privacy-id.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = {
+  up: async queryInterface => {
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY privacy_transfer_id ON "Transactions" (((data ->> 'token')::text)) WHERE (data ->> 'token') IS NOT NULL;
```
Betree, line 6: Since the index is created concurrently there's no risk of blocking the DB, this migration does a great job in this regard.

The only downside of the current approach is that Postgres needs to go through all the existing transactions, which could explain why it took so long. A simple condition on the date could have to speed it up:

```es6
WHERE (data ->> 'token') IS NOT NULL AND "createdAt" > '2021-03-01';
```

(I'm not sure when is the first privacy transaction)

kewitz: Thanks Ben! I'll keep that in mind for the next time!

PR #6167, opened 2021-06-16
Add Support for Notifications for Changelog Updates

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4345

Front-End PR: https://github.com/opencollective/opencollective-frontend/pull/6510


server/graphql/v2/object/Individual.js
```diff
@@ -108,11 +108,28 @@ export const Individual = new GraphQLObjectType({
...
+                CollectiveId: collectiveId,
+                publishedAt: { [Op.ne]: null },
+                isChangelog: true,
+              },
+              order: [['publishedAt', 'DESC']],
+              limit: 1,
+            });
+
+            // keep the latest change log publish date for a day in cache
+            cache.set(cacheKey, latestChangelogUpdate, 24 * 60 * 60);
```
Betree, line 129: No need to store the full update in the cache

```es6
cache.set(cacheKey, latestChangelogUpdate?.publishedAt, 24 * 60 * 60);
```

SudharakaP: Thanks. Done. 👍🏽

server/graphql/v1/types.js
```diff
@@ -241,6 +242,12 @@ export const UserType = new GraphQLObjectType({
           return user.data && user.data.features && user.data.features.ALL === false;
         },
       },
+      hasSeenLatestChangelogEntry: {
+        type: GraphQLBoolean,
+        async resolve(user) {
+          return await hasSeenLatestChangelogEntry(user);
```
Betree, line 248: Any reason for adding this?

```suggestion
          return hasSeenLatestChangelogEntry(user);
```

server/graphql/v1/types.js
```diff
@@ -241,6 +242,12 @@ export const UserType = new GraphQLObjectType({
           return user.data && user.data.features && user.data.features.ALL === false;
         },
       },
+      hasSeenLatestChangelogEntry: {
+        type: GraphQLBoolean,
```
Betree, line 246:
```suggestion
        type: new GraphQLNonNull(GraphQLBoolean),
```

server/graphql/v2/mutation/index.js
```diff
@@ -17,6 +17,7 @@ import paymentMethodMutations from './PaymentMethodMutations';
 import payoutMethodMutations from './PayoutMethodMutations';
 import transactionMutations from './TransactionMutations';
 import updateMutations from './UpdateMutations';
+import userMutations from './UserMutations';
```
Betree, line 20: To be confirmed by @znarf, but I think the term "User" is banned in GQLV2. We rather refer to them as "Individuals"

Betree (commented):
I don't think merging https://github.com/opencollective/opencollective-api/pull/6147 in an incomplete state is a good idea. I would prefer if we can either merge this PR in `feat/add-support-for-notifications-for-changelogs` or close the other one, but make it only one PR. Easier to review, easier to rollback, easier to audit when looking at the history.

Other than that, some minor feedback below but the overall design looks good!

server/graphql/v1/types.js
SudharakaP: Ah yes, I think I was testing something; this isn't needed. 🤔

server/graphql/v1/types.js
SudharakaP: Corrected it. 👍🏽

server/graphql/v2/mutation/index.js
SudharakaP: @znarf : Should we change this you think? 😄

znarf: It's not a big problem because it's not user facing.

But it you ask me, it should not been named User but Account or Individual. There should be ideally no "User" in GraphQL API v2, it's confusing.

SudharakaP: Done. I've renamed those references. 😄

server/graphql/v2/object/Individual.js
```diff
@@ -107,6 +108,12 @@ export const Individual = new GraphQLObjectType({
           }
         },
       },
+      hasSeenLatestChangelogEntry: {
+        type: new GraphQLNonNull(GraphQLBoolean),
+        async resolve(individual) {
+          return await hasSeenLatestChangelogEntry(individual);
```
Betree, line 114: Did you try this one to confirm if it works? My understanding is that `individual` is a `Collective`, not a `User`

znarf: Good feedback. Do we have any test coverage on this?

server/graphql/v2/object/Individual.js
```diff
@@ -107,6 +108,12 @@ export const Individual = new GraphQLObjectType({
           }
         },
       },
+      hasSeenLatestChangelogEntry: {
+        type: new GraphQLNonNull(GraphQLBoolean),
+        async resolve(individual) {
```
znarf, line 113: In other resolvers we name it `collective` so it's clear it's a `Collective`.

server/graphql/v2/object/Individual.js
SudharakaP: @Betree : Ops, you are right, seems I've missed this, since we aren't using this field for now. Anyways, I've tested it and added a line to convert it to user first before passing. 👍🏽

server/graphql/v2/object/Individual.js
SudharakaP: Done. 😄

server/graphql/v2/object/Individual.js
```diff
@@ -107,6 +108,13 @@ export const Individual = new GraphQLObjectType({
           }
         },
       },
+      hasSeenLatestChangelogEntry: {
+        type: new GraphQLNonNull(GraphQLBoolean),
+        async resolve(collective) {
+          const user = collective.getUser();
+          return await hasSeenLatestChangelogEntry(user);
```
znarf, line 115: await is unnecessary

server/graphql/common/user.ts
```diff
@@ -64,3 +65,29 @@ export const createUser = (
...
+  if (!latestChangelogUpdatePublishDate) {
+    const collectiveId = await fetchCollectiveId('opencollective');
+    const latestChangelogUpdate = await models.Update.findOne({
+      where: {
+        CollectiveId: collectiveId,
+        publishedAt: { [Op.ne]: null },
+        isChangelog: true,
+      },
+      order: [['publishedAt', 'DESC']],
+      limit: 1,
```
znarf, line 81: No need to limit with findOne

server/graphql/v2/object/Individual.js
SudharakaP: Thanks. Corrected that. 😄

server/graphql/common/user.ts
SudharakaP: Done. My mistake. 👍🏽

SudharakaP:
@Betree @znarf : Are we good to merge this. Just wondering as although the reviews were done, there's no approval yet. 😉

PR #6170, opened 2021-06-17
fix: add replyTo header in new pending financial contribution email

MlkMahmud:
Add the fromCollective admins in the replyTo header

resolves https://github.com/opencollective/opencollective/issues/4100


Betree (approved):
Thanks @MlkMahmud! I've added a small commit to handle incognito profiles as they're a bit special, but your implementation worked and is pretty clean :+1:

Betree:
CI does not want to run properly, probably because this is a fork. I'm merging now and will follow up if there are issues.

PR #6171, opened 2021-06-18
Digital Wallet: Implement Alipay

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4355

Adds Alipay single-time payment and refund.

### DIFF
--- a/server/paymentProviders/stripe/alipay.ts
+++ b/server/paymentProviders/stripe/alipay.ts
@@ -141,7 +141,9 @@ const confirmOrder = async (req: Request, res: Response, next: NextFunction): Pr
     } else if (redirect_status === 'failed') {
       const id = idDecode(OrderId, IDENTIFIER_TYPES.ORDER);
       debug(`payment for order ${id} failed, deleting order`);
-      const order = await models.Order.findByPk(id);
+      const order = await models.Order.findByPk(id, {
+        include: [{ model: models.Collective, as: 'collective' }],
+      });
       if (order) {
         await order.destroy();
         res.redirect(
--- a/server/paymentProviders/stripe/alipay.ts
+++ b/server/paymentProviders/stripe/alipay.ts
@@ -1,0 +1,212 @@
+/* eslint-disable camelcase */
+import querystring from 'querystring';
+
+import config from 'config';
+import debugLib from 'debug';
+import { NextFunction, Request, Response } from 'express';
+
+import OrderStatus from '../../constants/order_status';
+import { TransactionTypes } from '../../constants/transactions';
+import { idDecode, IDENTIFIER_TYPES } from '../../graphql/v2/identifiers';
+import logger from '../../lib/logger';
+import { createRefundTransaction, getHostFee, getPlatformFee } from '../../lib/payments';
+import stripe, { convertFromStripeAmount, convertToStripeAmount, extractFees } from '../../lib/stripe';
+import models from '../../models';
+
+import { refundTransaction } from './common';
+
+const debug = debugLib('alipay');
+
+const compatibleCurrencies = ['cny', 'aud', 'cad', 'eur', 'gbp', 'hkd', 'jpy', 'myr', 'nzd', 'sgd', 'usd'];
+
+const processOrder = async (order: typeof models.Order): Promise<void> => {
+  const hostStripeAccount = await order.collective.getHostStripeAccount();
+  if (!hostStripeAccount) {
+    throw new Error('Host is not connected to Stripe');
+  }
+  if (!compatibleCurrencies.includes(order.currency.toLowerCase())) {
+    throw new Error(`We can not pay with Alipay in ${order.currency} currency`);
+  }
+
+  let intent;
+  if (!order.data?.paymentIntent) {
+    debug(`creating intent for order ${order.id}`);
+    intent = await stripe.paymentIntents.create(
+      {
+        payment_method_types: ['alipay'],
+        amount: convertToStripeAmount(order.currency, order.totalAmount),
+        currency: order.currency,
+      },
+      {
+        stripeAccount: hostStripeAccount.username,
+      },
+    );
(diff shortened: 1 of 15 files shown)


server/paymentProviders/stripe/alipay.ts
```diff
@@ -0,0 +1,214 @@
...
+        ],
+      });
+      if (order.status !== OrderStatus.REQUIRE_CLIENT_CONFIRMATION) {
+        logger.warn(
+          `Trying to confirm Alipay order but order status is not waiting for client confirmation: #${order.id}`,
+        );
+        res.sendStatus(200);
+        return;
+      }
+      debug(`confirming order ${order.id}`);
```
Betree, line 79: Haven't tested the following to confirm it's possible, but it sounds like we should check that `order.data.paymentIntent.id` matches `payment_intent.id` here, otherwise the following exploit seems possible:
1. Create a first order for $10 with Alipay
2. Create a second order for $1000 with Alipay
3. Confirm the first payment, but temper the payload in https://github.com/opencollective/opencollective-frontend/pull/6505/files#diff-506c494bf8c5f399011b0826e77f2c62689a91d228668c18972beb087f6d6923R210 to inject the `OrderId` of the second order
4. Your $10 payment confirms a $1000 order (which could be used to bypass tier min amounts for example)

server/paymentProviders/stripe/alipay.ts
```diff
@@ -0,0 +1,214 @@
...
+
+      res.redirect(`${config.host.website}/${order.collective.slug}/donate/success?OrderId=${OrderId}`);
+    } else if (redirect_status === 'failed') {
+      const id = idDecode(OrderId, IDENTIFIER_TYPES.ORDER);
+      debug(`payment for order ${id} failed, deleting order`);
+      const order = await models.Order.findByPk(id, {
+        include: [{ model: models.Collective, as: 'collective' }],
+      });
+      if (order) {
+        await order.destroy();
```
Betree, line 148: Not sure whether it should be destroyed or rather marked as `ERROR`? In both cases, I think it would be great to have some details about the error in `order.data`.

server/paymentProviders/stripe/alipay.ts
```diff
@@ -0,0 +1,214 @@
...
+  }
+};
+
+const webhook = async (_, event) => {
+  if (event.type === 'charge.refund.updated') {
+    const refund = event.data.object;
+    if (refund.status === 'succeeded') {
+      const transaction = await models.Transaction.findOne({
+        where: { type: 'CREDIT', isRefund: false, data: { charge: { id: refund.charge } } },
+        include: [{ model: models.Collective, as: 'collective' }, { model: models.PaymentMethod }],
```
Betree, line 170: :ice_cream: There's probably almost 0 chance of a conflict on a charge ID with another provider, but I find it cleaner to be explicit about what we're searching here (and it saved us a `if (transaction.PaymentMethod.type !== 'alipay')` below) 

```suggestion
        include: [{ model: models.Collective, as: 'collective' }, { model: models.PaymentMethod, require: true, where: { type: 'alipay' } }],
```

Betree (commented):
The code looks great! Nice to see that we're able to share so much code between creditcard and alipay.

server/paymentProviders/stripe/alipay.ts
```diff
@@ -0,0 +1,214 @@
...
+const compatibleCurrencies = ['cny', 'aud', 'cad', 'eur', 'gbp', 'hkd', 'jpy', 'myr', 'nzd', 'sgd', 'usd'];
+
+const processOrder = async (order: typeof models.Order): Promise<void> => {
+  const hostStripeAccount = await order.collective.getHostStripeAccount();
+  if (!hostStripeAccount) {
+    throw new Error('Host is not connected to Stripe');
+  }
+  if (!compatibleCurrencies.includes(order.currency.toLowerCase())) {
+    throw new Error(`We can not pay with Alipay in ${order.currency} currency`);
+  }
```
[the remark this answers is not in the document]
kewitz: That's a great idea! I just looked into that but for some reason, Stripe doesn't list AliPay nor WePay in the capabilities endpoint.
https://stripe.com/docs/connect/account-capabilities#payment-methods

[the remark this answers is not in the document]
Betree: I meant in the settings, as we do in the contribution flow. The main idea being to prevent malicious users to use Alipay on a host that disabled it.

server/paymentProviders/stripe/alipay.ts
```diff
@@ -0,0 +1,224 @@
...
+      }
+      debug(`confirming order ${order.id}`);
+      const host = await order.collective.getHostCollective();
+      const hostStripeAccount = await order.collective.getHostStripeAccount();
+      const hostPlan = await host.getPlan();
+      const hostFeeSharePercent = hostPlan?.hostFeeSharePercent;
+      const isSharedRevenue = !!hostFeeSharePercent;
+
+      // Read or compute Platform Fee
+      const platformFee = await getPlatformFee(order.totalAmount, order, host, { hostPlan, hostFeeSharePercent });
```
znarf, line 92: I thought we could not charge an application fee, what's the point of computing this?

server/paymentProviders/stripe/alipay.ts
```diff
@@ -0,0 +1,224 @@
...
+        charge,
+        balanceTransaction,
+        isFeesOnTop: order.data?.isFeesOnTop,
+        isSharedRevenue,
+        platformFee: platformFee,
+        platformTip,
+        hostFeeSharePercent,
+      };
+      const hostCurrencyFxRate = amountInHostCurrency / order.totalAmount;
+      const platformFeeInHostCurrency = isSharedRevenue ? platformTip * hostCurrencyFxRate || 0 : fees.applicationFee;
```
znarf, line 118: Why do we try to read fees.applicationFee ?

PR #6178, opened 2021-06-18
Migrate from ISODateTime to GraphQLDateTime

znarf:
https://github.com/opencollective/opencollective-frontend/pull/6511


Betree (approved):
I now realize that this can be a breaking change for mutations and queries using an `ISODateTime` for an argument. Unfortunately, without union input types or the Tagged type proposals that are not yet implemented in graphql-js I don't see a simple way to prevent that completely.

znarf:
@Betree It's definitely a breaking as we had to update frontend code. Not sure if other projects are also relying on this.

Betree:
The PDF service is, I've made a similar PR in https://github.com/opencollective/opencollective-pdf/pull/691

znarf:
@nicolo-ribaudo I'm sorry about it, we hoped it would be only used in our codebase and we missed that you and others were relying on it.

We're generally avoiding breaking changes here but this is a good lesson. We'll think about a mechanism to publish important changes to our API.

PR #6183, opened 2021-06-21
Add `decline_code` and Error Details for Stripe Errors

SudharakaP:
This adds some additional details when a Stripe error occurs in the order submission. 🐨 

Related to https://github.com/opencollective/opencollective/issues/4362


server/paymentProviders/stripe/creditcard.js
```diff
@@ -322,6 +322,9 @@ export default {
       ];
 
       if (knownErrors.includes(error.message)) {
+        logger.error(
+          `Type: ${error.type}, Message: ${error.message}, Decline Code: ${error.decline_code}, Code: ${error.code}`,
```
Betree, line 326: I would somehow indicate that it's a Stripe error in the logs:
```suggestion
          `Stripe Error (handled): ${error.type}, Message: ${error.message}, Decline Code: ${error.decline_code}, Code: ${error.code}`,
```

I also wonder if it is right to log known errors such as `Your card has expired as errors.

SudharakaP: @Betree : Yes, but I think we are missing this information only for known errors, we do print the whole error object for others; 🤔 

https://github.com/opencollective/opencollective-api/blob/54255052ca1ee24482e9e6d7f729f509838bfc15/server/paymentProviders/stripe/creditcard.js#L357-L359

PR #6187, opened 2021-06-21
Implement host fee share

znarf:


server/models/Transaction.js
```diff
@@ -754,14 +759,85 @@ function defineModel() {
...
+      );
+    }
+
+    return { hostFeeShareTransaction, hostFeeShareDebtTransaction };
+  };
+
+  Transaction.creatHostFeeShareDebtTransactions = async (hostFeeShareTransactionData, host) => {
+    // Create debt transaction
+    const hostFeeShareDebtTransactionData = {
+      ...omit(hostFeeShareTransactionData, ['data']), // TODO: remove the OrderId?
```
Betree, line 822: The more I think about it, the more I find it useful to be able to group all transactions (including refunds) by `OrderId`

server/models/Transaction.js
```diff
@@ -811,11 +887,17 @@ function defineModel() {
...
     // Create Host Fee transaction
     if (transaction.hostFeeInHostCurrency && parseToBoolean(config.ledger.separateHostFees) === true) {
-      // transaction.hostFeeInHostCurrency = 0;
       const result = await Transaction.createHostFeeTransactions(transaction, host);
-      // Transaction was modified by createHostFeeTransaction, we get it from the result
-      if (result && result.transaction) {
-        transaction = result.transaction;
+      if (result) {
+        if (result.hostFeeTransaction) {
+          // TODO: rename isPlatformTipDirectlyCollected
```
Betree, line 893: I agree, better have an unified flag for this

Betree (commented):
Great start!

PR #6188, opened 2021-06-21
Add Emoji Reactions to Updates

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4185

Related Front End PR: https://github.com/opencollective/opencollective-frontend/pull/6523


server/graphql/v2/mutation/EmojiReactionMutations.js
```diff
@@ -0,0 +1,130 @@
...
+import { canComment } from '../../common/expenses';
+import { Forbidden, NotFound, Unauthorized, ValidationFailed } from '../../errors';
+import { idDecode, IDENTIFIER_TYPES } from '../identifiers';
+import { CommentReferenceInput } from '../input/CommentReferenceInput';
+import { UpdateReferenceInput } from '../input/UpdateReferenceInput';
+import { EmojiReactionsOutput } from '../object/EmojiReactionsOutput';
+
+const emojiReactionMutations = {
+  addEmojiReaction: {
+    type: new GraphQLNonNull(EmojiReactionsOutput),
```
Betree, line 14: Since `EmojiReactionsOutput` is very specific to these mutations, I would keep the object declaration in the same file. The type name can also be more explicit, I can suggest `EmojiReactionResponse`

server/graphql/v2/mutation/EmojiReactionMutations.js
```diff
@@ -0,0 +1,130 @@
...
+      } else if (args.update) {
+        const updateId = idDecode(args.update.id, IDENTIFIER_TYPES.UPDATE);
+        const update = await models.Update.findByPk(updateId);
+
+        if (!update) {
+          throw new ValidationFailed('This update does not exist');
+        }
+
+        const reaction = await models.EmojiReaction.addReactionOnUpdate(req.remoteUser, updateId, args.emoji);
+        return { update: models.Update.findByPk(reaction.UpdateId), comment: null };
```
Betree, line 61: We already loaded the update above

```suggestion
        return { update, comment: null };
```

server/graphql/v2/mutation/EmojiReactionMutations.js
```diff
@@ -0,0 +1,130 @@
...
+        if (!comment) {
+          throw new ValidationFailed('This comment does not exist');
+        } else if (comment.ExpenseId) {
+          const expense = await models.Expense.findByPk(comment.ExpenseId);
+          if (!expense || !(await canComment(req, expense))) {
+            throw new Forbidden('You are not allowed to comment or add reactions on this expense');
+          }
+        }
+        const reaction = await models.EmojiReaction.addReactionOnComment(req.remoteUser, commentId, args.emoji);
+        return { comment: models.Comment.findByPk(reaction.CommentId), update: null };
```
Betree, line 51: Already loaded above

```suggestion
        return { comment, update: null };
```

server/graphql/v2/mutation/EmojiReactionMutations.js
```diff
@@ -0,0 +1,130 @@
...
+            emoji,
+          },
+        });
+
+        if (!reaction) {
+          throw new NotFound(`This update reaction does not exist or has been deleted.`);
+        }
+
+        // Check permissions
+        if (!remoteUser.isAdmin(reaction.FromCollectiveId)) {
```
Betree, line 119: Seems like we can simplify the code here, there are some duplications between the `if (comment)` and `if (update)`. Would be great to have something like:

```es6
const reaction = ...; // Load from either comment or update
// Check permissions, delete, etc. all common code
```

server/graphql/v2/object/Update.js
```diff
@@ -138,6 +139,22 @@ const Update = new GraphQLObjectType({
           return req.loaders.Collective.byId.load(update.CollectiveId);
         },
       },
+      reactions: {
+        type: GraphQLJSON,
+        description: 'Returns a map of reactions counts for this update',
+        async resolve(update, args, req) {
+          return await req.loaders.Update.reactionsByUpdateId.load(update.id);
```
Betree, line 146:
```suggestion
          return req.loaders.Update.reactionsByUpdateId.load(update.id);
```

Betree (commented):
That looks awesome! Some feedback below, but it's mostly cosmetic; the code itself seems robust.

migrations/dev-20210621-add-support-for-update-emojis.js
```diff
@@ -0,0 +1,61 @@
...
+      queryInterface.sequelize.query(
+        'ALTER TABLE "EmojiReactions" RENAME CONSTRAINT "CommentReactions_CommentId_fkey" TO "EmojiReactions_CommentId_fkey";',
+      );
+      queryInterface.sequelize.query(
+        'ALTER TABLE "EmojiReactions" RENAME CONSTRAINT "CommentReactions_FromCollectiveId_fkey" TO "EmojiReactions_FromCollectiveId_fkey";',
+      );
+      queryInterface.sequelize.query(
+        'ALTER TABLE "EmojiReactions" RENAME CONSTRAINT "CommentReactions_UserId_fkey" TO "EmojiReactions_UserId_fkey";',
+      );
+      queryInterface.sequelize.query('ALTER INDEX "CommentReactions_pkey" RENAME TO "EmojiReactions_pkey";');
```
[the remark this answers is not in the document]
SudharakaP: @Betree : My thinking is that for these renames we don't care if they run asynchronously or not. 🤔 

I put them in a `then` purely for clarity and organization. So these renames all correspond to the `CommentReactions` table rename. So I thought it would be clearer for the reader to bundle them together in a `then`. But let me know if you think otherwise. 😄 🤔

server/graphql/v2/mutation/EmojiReactionMutations.js
SudharakaP: @Betree : Yes, but note that we add the reaction to the update on the preceding line, so we need to fetch it again to get the updated `update`? 🤔

server/graphql/v2/mutation/EmojiReactionMutations.js
SudharakaP: @Betree : Yes that's correct but note that we add the reaction to the comment on the preceding line. So we need to get the updated comment, isn't? 🤔 😄

server/graphql/v2/mutation/EmojiReactionMutations.js
SudharakaP: Done, I've refactored both `addEmojiReaction` and `removeEmojiReaction` methods to avoid any duplicate logic. 😄

migrations/dev-20210621-add-support-for-update-emojis.js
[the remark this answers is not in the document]
Betree: I understand, the intent makes sense. I think it's better to always await on promises, but since this is a migration that should be fine.

server/graphql/v2/mutation/EmojiReactionMutations.js
Betree: I don't think the update itself gets updated in the process of adding a reaction, does it?

server/graphql/v2/mutation/EmojiReactionMutations.js
Betree: Same here, I don't think the comment gets mutated in the operation

server/graphql/v2/mutation/EmojiReactionMutations.js
```diff
@@ -0,0 +1,164 @@
...
+      },
+    });
+  }
+
+  if (!reaction) {
+    if (identifierType === IDENTIFIER_TYPES.COMMENT) {
+      throw new NotFound(`This comment reaction does not exist or has been deleted.`);
+    } else {
+      throw new NotFound(`This update reaction does not exist or has been deleted.`);
+    }
```
Betree, line 148: :ice_cream: I think it's fine to have only one message here

```suggestion
    throw new NotFound(`This reaction does not exist or has been deleted.`);
```

test/test-helpers/fake-data.js
```diff
@@ -308,18 +308,29 @@ export const fakeComment = async commentData => {
 /**
  * Creates a fake comment reaction. All params are optionals.
  */
-export const fakeCommentReaction = async (reactionData = {}) => {
+export const fakeEmojiReaction = async (isComment, reactionData = {}) => {
```
Betree, line 311: Let's keep a similar API to the other `fake*` helpers above.

```suggestion
export const fakeEmojiReaction = async (reactionData = {}, opts) => {
```

Betree (approved):
No blocker on my side!

test/test-helpers/fake-data.js
SudharakaP: @Betree : Did you mean passing an objection `opt` as second argument? I've done it that way, but let me know if you see any further issues. 😄

server/graphql/v2/mutation/EmojiReactionMutations.js
SudharakaP: Ah, yes you are right; 🤯 I've corrected this. 😄

server/graphql/v2/mutation/EmojiReactionMutations.js
SudharakaP: Yep, you are right. I've corrected this. 😄

migrations/dev-20210621-add-support-for-update-emojis.js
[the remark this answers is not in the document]
SudharakaP: I'll leave it as it is unless you like me to change it. I have no strong opinion either way on this. 😄

Betree (approved):
Lgtm!

server/graphql/v2/mutation/EmojiReactionMutations.js
```diff
@@ -0,0 +1,160 @@
...
+import { CommentReferenceInput } from '../input/CommentReferenceInput';
+import { UpdateReferenceInput } from '../input/UpdateReferenceInput';
+import { Comment } from '../object/Comment';
+import Update from '../object/Update';
+
+/**
+ * Object type for EmojiReaction mutation.
+ */
+const EmojiReactionsResponse = new GraphQLObjectType({
+  name: 'EmojiReactionOutput',
```
Betree, line 17: The names must match here

```suggestion
const EmojiReactionsResponse = new GraphQLObjectType({
  name: 'EmojiReactionResponse',
```

SudharakaP: Done. 👍🏽

SudharakaP:
> That looks awesome! Some feedback below, but it's mostly cosmetic; the code itself seems robust.

Thanks much. 😄

Betree (approved):

PR #6191, opened 2021-06-22
Misc improvements on settlement handling

Betree:
It seems cleaner to mark the transaction as settled rather than deleting the settlement

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -271,7 +271,7 @@ export async function createRefundTransaction(transaction, refundedPaymentProces
     const platformTipRefundTransaction = await models.Transaction.createDoubleEntry(platformTipRefund);
     await associateTransactionRefundId(platformTipTransaction, platformTipRefundTransaction, data);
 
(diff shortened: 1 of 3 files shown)


server/lib/payments.js
```diff
@@ -280,7 +280,7 @@ export async function createRefundTransaction(transaction, refundedPaymentProces
       },
     });
 
-    // Old tips did not have a "debt" transaction associated
+    // Tips directly collected (and legacy ones) do not have a "debt" transaction associated
```
znarf, line 283: So what's happening to them?

I was looking at that comment this morning and it feels like it's trying to communicate something but is failing short.

Betree: The goal of this comment is to explain the reason for this `if`, why do some platform tips don't have a debt? It seems explicit to me that we're not doing anything more if there's no debt associated (since we're not entering the `if`), but I can eventually add more text if it's not

znarf (approved):

PR #6204, opened 2021-06-24
More accurate handling of HOST_FEE_SHARE_DEBT

znarf:


Betree (approved):
The code looks good, adding these to the snapshot tests will help to better visualize

PR #6206, opened 2021-06-24
Refactor platform tip handling

znarf:


server/paymentProviders/opencollective/collective.js
```diff
@@ -53,42 +54,45 @@ paymentMethodProvider.processOrder = async order => {
...
+    hostCurrencyFxRate,
     amountInHostCurrency,
     hostFeeInHostCurrency,
-    platformFeeInHostCurrency,
     taxAmount: order.taxAmount,
-    paymentProcessorFeeInHostCurrency: 0,
     description: order.description,
     data: {
       isFeesOnTop: order.data?.isFeesOnTop,
+      hasPlatformTip: platformTip ? true : false,
```
Betree, line 91: Maybe we have an opportunity to solve https://github.com/opencollective/opencollective/issues/4236 here? By adding an additional flag like `isElligibleToPlatformTip`

server/paymentProviders/stripe/creditcard.js
```diff
@@ -197,22 +196,31 @@ const createChargeAndTransactions = async (hostStripeAccount, { order, hostStrip
   const hostFee = await getHostFee(order, host);
   const hostFeeInHostCurrency = Math.round(hostFee * hostCurrencyFxRate);
 
+  const fees = extractFees(balanceTransaction, balanceTransaction.currency);
+
+  let platformTipInHostCurrency, platformFeeInHostCurrency;
+  const platformTip = getPlatformTip(order);
+  if (platformTip) {
+    platformTipInHostCurrency = isSharedRevenue ? platformTip * hostCurrencyFxRate || 0 : fees.applicationFee;
```
Betree, line 204: It was there from before, but looks to me like this one should be rounded

```suggestion
    platformTipInHostCurrency = isSharedRevenue ? Math.round(platformTip * hostCurrencyFxRate) || 0 : fees.applicationFee;
```

Betree (approved):
The code is much clearer like that :ok_hand:

server/paymentProviders/opencollective/collective.js
znarf: I added support for this but no migration for past data at this point.

server/paymentProviders/stripe/creditcard.js
znarf: Added it back.

Betree: Sorry, my feedback was not very clear: the `Math.round` was not there in the previous version. Anyway, I think it's a good addition.

PR #6224, opened 2021-06-29
Remove CommentReactionMutations.js

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4393. 

Since we now use `emojiReactionMutations` in favour of `commentReactionMutations`, we should remove the comment reaction mutations. 😄 🐨


Betree (approved):
All good, let's wait for the feature to be fully deployed on prod before merging this. Could also be nice to update the GQL files, but not mandatory.

SudharakaP:
> All good, let's wait for the feature to be fully deployed on prod before merging this. It could also be nice to update the GQL files, but not mandatory.

Thanks. I've just updated the GQL files as well. 😄

SudharakaP:
@Betree : I'll be merging this one if no objections given that we have now merged the new emoji reaction mutations. 😄

Betree (approved):
Sounds good :+1:

PR #6226, opened 2021-06-30
Add Filtering Support for Updates

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4384

Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/6555

### DIFF
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -300,7 +300,10 @@ const accountFieldsDefinition = () => ({
       ...CollectionArgs,
       onlyPublishedUpdates: { type: GraphQLBoolean },
       onlyChangelogUpdates: { type: GraphQLBoolean },
-      orderBy: { type: ChronologicalOrderInput },
+      orderBy: {
+        type: new GraphQLNonNull(ChronologicalOrderInput),
+        defaultValue: ChronologicalOrderInput.defaultValue,
+      },
       searchTerm: { type: GraphQLString },
     },
     async resolve(collective, { limit, offset, onlyPublishedUpdates, onlyChangelogUpdates, orderBy, searchTerm }) {
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -300,20 +300,56 @@ const accountFieldsDefinition = () => ({
       ...CollectionArgs,
       onlyPublishedUpdates: { type: GraphQLBoolean },
       onlyChangelogUpdates: { type: GraphQLBoolean },
+      sortBy: { type: GraphQLString },
+      searchTerm: { type: GraphQLString },
     },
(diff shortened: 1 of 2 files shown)


server/graphql/v2/interface/Account.js
```diff
@@ -300,20 +300,56 @@ const accountFieldsDefinition = () => ({
       ...CollectionArgs,
       onlyPublishedUpdates: { type: GraphQLBoolean },
       onlyChangelogUpdates: { type: GraphQLBoolean },
+      sortBy: { type: GraphQLString },
```
Betree, line 303: In `ExpensesQuery`, `TransactionsQuery` and `OrdersQuery` it seems that we use `orderBy`, with the `ChronologicalOrderInput` type. Would be great to make it consistent. See https://github.com/opencollective/opencollective-api/blob/b1737a73c4c770a27eefd505749c45cb52d01e2d/server/graphql/v2/query/ExpensesQuery.ts#L105

server/graphql/v2/interface/Account.js
```diff
@@ -300,20 +300,56 @@ const accountFieldsDefinition = () => ({
...
+      searchTerm: { type: GraphQLString },
     },
-    async resolve(collective, { limit, offset, onlyPublishedUpdates, onlyChangelogUpdates }) {
+    async resolve(collective, { limit, offset, onlyPublishedUpdates, onlyChangelogUpdates, sortBy, searchTerm }) {
       let where = {
         CollectiveId: collective.id,
+        [Op.and]: [],
       };
+      const include = [{ association: 'fromCollective', required: true, attributes: [] }];
```
Betree, line 311: This `include` does not seem necessary if there's no `searchTerm`

server/graphql/v2/interface/Account.js
SudharakaP: Thanks. Fixed. 👍🏽

server/graphql/v2/interface/Account.js
SudharakaP: Thanks. I've refactored it. 👍🏽

server/graphql/v2/interface/Account.js
```diff
@@ -300,20 +300,52 @@ const accountFieldsDefinition = () => ({
       ...CollectionArgs,
       onlyPublishedUpdates: { type: GraphQLBoolean },
       onlyChangelogUpdates: { type: GraphQLBoolean },
+      orderBy: { type: ChronologicalOrderInput },
```
Betree, line 303: I would make it non-nullable with a default value (createdAt DESC). It's good to have a predictive order for things.

Plus, I'm not sure what happens if this is null line 317, probably a crash when we try to do `orderBy.field`?

SudharakaP: Yes, agreed it would crash. Seems I've missed this one. I've just corrected it. 🤔 🐨

Betree (approved):
Lgtm :+1:

PR #6234, opened 2021-07-01
Update postgres version from CI and Docs to match prod

kewitz:


SudharakaP (changes requested):
I think we also need to update the following lines to something like, 

```
You need to have PostgreSQL 9.x, 10.x or 11.x with the Postgis extension.

In production, we're currently running 9.6.17.
```
to something like,

```
You need to have PostgreSQL 13.x with the Postgis extension.

In production, we're currently running 13.3.
```

PR #6242, opened 2021-07-01
HostSettlement: Fix dates and filter inactive hosts

Betree:


znarf (approved):
Looks good, thank you for rewriting the mess with the dates. Let's double check it's consistent with what we want, it's easy to get it wrong.

Minor lint and typescript issues to fix according to CI.

PR #6248, opened 2021-07-02
Make Sure Dates Have Same Format Before Comparison - Changelog Updates

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4421

### DIFF
--- a/server/graphql/common/user.ts
+++ b/server/graphql/common/user.ts
@@ -87,6 +87,6 @@ export const hasSeenLatestChangelogEntry = async (user: typeof models.User): Pro
   if (!latestChangelogUpdatePublishDate) {
     return true;
   } else {
-    return user?.changelogViewDate >= latestChangelogUpdatePublishDate;
+    return new Date(user?.changelogViewDate) >= new Date(latestChangelogUpdatePublishDate);
   }
 };
--- a/server/graphql/common/user.ts
+++ b/server/graphql/common/user.ts
@@ -69,7 +69,9 @@ export const createUser = (
 export const hasSeenLatestChangelogEntry = async (user: typeof models.User): Promise<boolean> => {
   const cacheKey = 'latest_changelog_publish_date';
   let latestChangelogUpdatePublishDate = await cache.get(cacheKey);
-  if (!latestChangelogUpdatePublishDate) {
+  if (latestChangelogUpdatePublishDate) {
+    return user?.changelogViewDate >= new Date(latestChangelogUpdatePublishDate);
+  } else {
     const collectiveId = await fetchCollectiveId('opencollective');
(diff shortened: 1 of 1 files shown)


server/graphql/common/user.ts
```diff
@@ -87,6 +87,6 @@ export const hasSeenLatestChangelogEntry = async (user: typeof models.User): Pro
   if (!latestChangelogUpdatePublishDate) {
     return true;
   } else {
-    return user?.changelogViewDate >= latestChangelogUpdatePublishDate;
+    return new Date(user?.changelogViewDate) >= new Date(latestChangelogUpdatePublishDate);
```
Betree, line 90: My understanding is that we will end up with a string here (rather than a date) only when we load stuff from the cache (because it has been serialized). `user.changelogViewDate` is already a date, and `latestChangelogUpdate.publishedAt` too.

We can do something like:

```es6
const dateFromCache = await cache.get(cacheKey);
if (dateFromCache) {
  return user?.changelogViewDate >= new Date(dateFromCache);
} else {
  ...
}
```

SudharakaP: @Betree : But then the doubt I have is why does it only appear in staging not in our local? 🤔 That's the part that I don't quite get. However, let us try your solution. I have made a fix, let me know if it looks good. 😄

Betree (commented):
It seems that this `else` will never be reached 

image

SudharakaP:
> It seems that this `else` will never be reached
> 
> image

@Betree : Ops, I've just fixed that. Hope it looks good now. 😄

server/graphql/common/user.ts
```diff
@@ -81,12 +83,11 @@ export const hasSeenLatestChangelogEntry = async (user: typeof models.User): Pro
     });
 
     latestChangelogUpdatePublishDate = latestChangelogUpdate?.publishedAt;
+    if (latestChangelogUpdatePublishDate) {
+      return true;
+    }
```
Betree, line 88: I don't get this condition, why not check against the `user?.changelogViewDate` here?

SudharakaP: @Betree : If this condition is not preset the changelog notification will appear without any new changelogs present. This will happen when we enable the changelog feature flag, but before posting the first update. 😄 I know it's a kinda condition which only helps initially when there's no changelog updates in the db, but I want to handle that condition so users don't get unnecessary notifications.

Betree: Shouldn't that be:

```suggestion
    if (!latestChangelogUpdatePublishDate) {
      return true;
    }
```

SudharakaP: @Betree : Ops, yes you are right. 🤯

Betree (approved):
Looks good!

PR #6256, opened 2021-07-07
getTotalNetAmountReceivedAmount: Include new host fees

Betree:
Part of https://github.com/opencollective/opencollective/issues/4431

This PR changes the way `getTotalNetAmountReceivedAmount` calculates the amount, by including `DEBIT` transactions to make sure we count the `HOST_FEE` transactions as well. I've limited the kinds to `['CONTRIBUTION', 'ADDED_FUNDS', 'HOST_FEE', 'PAYMENT_PROCESSOR_FEE']` to avoid counting expenses or other unrelated transactions.


Betree:
@znarf we should review this one in a follow-up and find a better approach

znarf:
It's problematic because it's gonna include the `ADDED_FUNDS` DEBITs.

PR #6258, opened 2021-07-07
Don't show changelog update notification to new users

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4435

new-user

### DIFF
--- a/server/graphql/common/user.ts
+++ b/server/graphql/common/user.ts
@@ -69,8 +69,10 @@ export const createUser = (
 export const hasSeenLatestChangelogEntry = async (user: typeof models.User): Promise<boolean> => {
   const cacheKey = 'latest_changelog_publish_date';
   let latestChangelogUpdatePublishDate = await cache.get(cacheKey);
+  // Make sure we don't show the changelog notifications for newly confirmed users
+  const userChangelogViewDate = user.changelogViewDate || user.confirmedAt || user.createdAt;
   if (latestChangelogUpdatePublishDate) {
-    return user?.changelogViewDate >= new Date(latestChangelogUpdatePublishDate);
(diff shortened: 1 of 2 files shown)


Betree (commented):
Now that I see this, I realize that this approach has a few downsides:
- Having `null` has a default value here makes it clear & easy to track the changelog opening percentage. It's kind of false to say that they viewed the changelog on this date when they have not.
- To make this work, we'd need to adapt things for guest users. Guest users create an account when they contribute, but they often login only weeks or months later. This first sign-in should be the date used to compare against changelog.

Editing `hasSeenLatestChangelogEntry` seems simpler. We should use `user.confirmedAt` for our checks. Something as simple as:

```es6
export const hasSeenLatestChangelogEntry = (...) => {
  // Add a comment here to explain why we're doing that
  const userChangelogViewDate = user.changelogViewDate || user.confirmedAt || user.createdAt;

  (...)
}
```

SudharakaP:
> Now that I see this, I realize that this approach has a few downsides:
> 
>     * Having `null` has a default value here makes it clear & easy to track the changelog opening percentage. It's kind of false to say that they viewed the changelog on this date when they have not.
> 
>     * To make this work, we'd need to adapt things for guest users. Guest users create an account when they contribute, but they often login only weeks or months later. This first sign-in should be the date used to compare against changelog.

@Betree : Agree, I've adjusted the code based on your suggestion. 😄

Betree (approved):
Lgtm :+1: The previous version of the code assumed that `user` could be null, but if that's not breaking anything I'm all good.

SudharakaP:
> Lgtm +1 The previous version of the code assumed that `user` could be null, but if that's not breaking anything I'm all good.

@Betree : Thanks. No it seems to work fine without the null check. So I think it's fine. 😄

PR #6259, opened 2021-07-07
Get the Correct emoji reaction count when removing emojis

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4387

This is simpler that I thought, the root cause for the concurrently removing emoji issue was that we were not checking for the correct emoji when multiple emojis of the same kinda are present in the table. This PR corrects it. 😄 

emoji-addition-correction


Betree (commented):
Good fix! We can further simplify if we go with this approach. Rather than:
```es6
reaction = models.X.findOne(...);
if (!reaction) { ... }
reaction.destroy(...);
```

We can directly do:
```es6
const idColumn = identifierType === IDENTIFIER_TYPES.COMMENT ? 'CommentId' : 'UpdateId';
await models.EmojiReaction.destroy({
  where: {
    [idColumn]: commentOrUpdateId,
    UserId: remoteUser.id,
    emoji,
  }
})
```

SudharakaP:
> Good fix! We can further simplify if we go with this approach. Rather than:
> 
> ```js
> reaction = models.X.findOne(...);
> if (!reaction) { ... }
> reaction.destroy(...);
> ```
> 
> We can directly do:
> 
> ```js
> const idColumn = identifierType === IDENTIFIER_TYPES.COMMENT ? 'CommentId' : 'UpdateId';
> await models.EmojiReaction.destroy({
>   where: {
>     [idColumn]: commentOrUpdateId,
>     UserId: remoteUser.id,
>     emoji,
>   }
> })
> ```

@Betree : Thanks. I've refactored the code based on your suggestion, but kept the `findOne` method as if we use the `destroy` directly we lose access to the `reactions` object (the `destroy` method only returns a boolean) which we need to pass back to the front-end. 😄 🤔

Betree:
> @Betree : Thanks. I've refactored the code based on your suggestion, but kept the `findOne` method as if we use the `destroy` directly we lose access to the `reactions` object (the `destroy` method only returns a boolean) which we need to pass back to the front-end. smile thinking

I don't see where we're returning the `reaction` object? We're using it to load the update/comment, but we may as well use `commentOrUpdateId` for that.

SudharakaP:
> > @Betree : Thanks. I've refactored the code based on your suggestion, but kept the `findOne` method as if we use the `destroy` directly we lose access to the `reactions` object (the `destroy` method only returns a boolean) which we need to pass back to the front-end. smile thinking
> 
> I don't see where we're returning the `reaction` object? We're using it to load the update/comment, but we may as well use `commentOrUpdateId` for that.

@Betree : Ops, you are right. I missed that part. I've changed the code to fix that. 😄

SudharakaP:
@kewitz : I am assigning this review to you since Ben is on vacation. I think I've fixed everything suggested in the review. Let me know if everything looks good. 😄

server/graphql/v2/mutation/EmojiReactionMutations.js
```diff
@@ -123,37 +123,23 @@ const addReactionToCommentOrUpdate = async (id, req, emoji, identifierType) => {
...
-  if (!remoteUser.isAdmin(reaction.FromCollectiveId)) {
-    throw new Forbidden();
+  if (!emojiRemoved) {
+    throw new NotFound(`This reaction does not exist or has been deleted or you do not have permission to change it.`);
   }
 
-  await reaction.destroy();
   if (identifierType === IDENTIFIER_TYPES.COMMENT) {
-    return { comment: models.Comment.findByPk(reaction.CommentId), update: null };
+    return { comment: models.Comment.findByPk(commentOrUpdateId), update: null };
```
kewitz, line 140: How does this work exactly? Shouldn't we await for the promise to resolve?
```suggestion
    return { comment: await models.Comment.findByPk(commentOrUpdateId), update: null };
```

server/graphql/v2/mutation/EmojiReactionMutations.js
```diff
@@ -123,37 +123,23 @@ const addReactionToCommentOrUpdate = async (id, req, emoji, identifierType) => {
...
   } else {
-    return { update: models.Update.findByPk(reaction.UpdateId), comment: null };
+    return { update: models.Update.findByPk(commentOrUpdateId), comment: null };
```
kewitz, line 142:
```suggestion
    return { update: await models.Update.findByPk(commentOrUpdateId), comment: null };
```

server/graphql/v2/mutation/EmojiReactionMutations.js
SudharakaP: I think it works because according to the doc; https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#rewriting_a_promise_chain_with_an_async_function

```
The return value of an async function is implicitly wrapped in Promise.resolve - if it's not already a promise itself (as in this example).
```

So my understanding is that for the return value we don't need the `await`. Am I missing something? 🤔

server/graphql/v2/mutation/EmojiReactionMutations.js
SudharakaP: @kewitz : Same as my explanation above. 😄

server/graphql/v2/mutation/EmojiReactionMutations.js
kewitz: You're not returning a promise here, you're returning an object.

SudharakaP: @kewitz : Yes you are right. I was confused on this; we are returning an object; but I need to figure out why it was working without the await. Let me think more on it. 🤔

SudharakaP: @kewitz : My thinking is we are seeing this as working because here we aren't waiting for the value in the front-end but using the optimistic response. 

https://github.com/opencollective/opencollective-frontend/blob/6d05f55b84c378f7d716c8033fbba0c9a4da6430/components/conversations/EmojiReactionPicker.js#L75

So I agree that here we should have an `await`.

Betree: I think it worked because GraphQL will await each `resolve` of the `EmojiReactionsResponse` object's fields. 

I also prefer the `await` approach though.

Betree (approved):
Lgtm :+1:

PR #6260, opened 2021-07-07
VirtualCards: Add webhook handler for Privacy

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4390

### DIFF
--- a/server/paymentProviders/privacy/webhook.ts
+++ b/server/paymentProviders/privacy/webhook.ts
@@ -1,0 +1,43 @@
+import { Request } from 'express';
+
+import logger from '../../lib/logger';
+import * as privacyLib from '../../lib/privacy';
+import models from '../../models';
+import { Transaction } from '../../types/privacy';
+
+import privacy from './index';
+
+async function webhook(req: Request & { body: Transaction; rawBody: string }): Promise<void> {
+  const virtualCard = await models.VirtualCard.findOne({
+    where: {
+      id: req.body.card.token,
+    },
+    include: [
+      { association: 'collective', required: true },
+      { association: 'host', required: true },
+    ],
+  });
+  if (!virtualCard) {
+    logger.error('privacy/webhook: could not find VirtualCard', { body: req.body });
+    return;
+  }
+
+  const host = virtualCard.host;
+  const collective = virtualCard.collective;
+  const [connectedAccount] = await host.getConnectedAccounts({
+    where: { service: 'privacy', deletedAt: null },
+  });
+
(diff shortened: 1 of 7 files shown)


server/models/Transaction.js
```diff
@@ -262,10 +262,14 @@ function defineModel() {
       },
 
       hooks: {
-        afterCreate: transaction => {
-          Transaction.createActivity(transaction);
-          // intentionally returns null, needs to be async (https://github.com/petkaantonov/bluebird/blob/master/docs/docs/warning-explanations.md#warning-a-promise-was-created-in-a-handler-but-was-not-returned-from-it)
-          return null;
+        afterCreate: (transaction, options) => {
+          if (options?.transaction) {
+            return Transaction.createActivity(transaction, { transaction: options.transaction });
```
Betree, line 267: My personal experience with working on that in https://github.com/opencollective/opencollective-api/pull/5981 is that this will not play well in some cases because `createActivity` then gets null results when it tries to fetch the transaction and its associations by ID. If it's not essential for this PR, I would prefer to wait for a proper revamp of this code.

server/paymentProviders/privacy/webhook.ts
```diff
@@ -0,0 +1,43 @@
...
+  });
+  if (!virtualCard) {
+    logger.error('privacy/webhook: could not find VirtualCard', { body: req.body });
+    return;
+  }
+
+  const host = virtualCard.host;
+  const collective = virtualCard.collective;
+  const [connectedAccount] = await host.getConnectedAccounts({
+    where: { service: 'privacy', deletedAt: null },
```
Betree, line 28: :ice_cream: Unless I'm mistaking, `deletedAt: null` is always included unless you specify `paranoid: false`

```suggestion
    where: { service: 'privacy' },
```

Betree (approved):
If possible I really think we should remove the thingy on `afterCreate` for Transactions. It's a touchy topic and we need to iterate on it, but I'm afraid of the side effects.

The rest looks really great :100:

server/models/Transaction.js
kewitz: My issue is that I want to create expenses and transactions within the same postgres transaction in case it fails and `Transaction.createActivity` is async.
Being async is not a problem when you're not operating within a postgres transaction, but if you are, it means that both Transaction and Activity need to be awaited for (that's why I'm also returning the `createActivity`) here.

The alternative is not using a postgres transaction in the function that processes privacy transactions, it would necessary means that we need to manually roll back created instances if things go wrong.

PR #6270, opened 2021-07-13
Refact: Wise batch optimizations

kewitz:


server/paymentProviders/transferwise/index.ts
```diff
@@ -240,27 +239,38 @@ async function createExpensesBatchGroup(
...
+    await Promise.all(
+      expenses.map(expense =>
+        expense.update({
+          data: { ...expense.data, batchGroup },
+        }),
+      ),
+    );
+  } catch (e) {
+    logger.error(e);
+    await transferwise.cancelBatchGroup(token, profileId, batchGroup.id, batchGroup.version).catch(console.error);
```
SudharakaP, line 272: @kewitz : I am not entirely sure, but doesn't it make more sense to send any exception to the logger itself rather than `console.error`? 🤔 

```suggestion
    await transferwise.cancelBatchGroup(token, profileId, batchGroup.id, batchGroup.version).catch(logger.error);
```

SudharakaP (approved):

PR #6274, opened 2021-07-13
Use Commit Hash as hash value for `.next` and build Cache

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4322

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -169,17 +169,16 @@ jobs:
         id: next-build
         with:
           path: opencollective-frontend/dist
-          key: ${{ runner.os }}-next-build-${{ hashFiles('opencollective-frontend/**') }}
+          key: ${{ runner.os }}-next-build-${{ env.FRONTEND_COMMIT_HASH }}
 
-      - name: Restore .next cache (frontend)
-        if: steps.next-build.outputs.cache-hit != 'true'
+      - name: Restore .next folder (frontend)
         uses: actions/cache@v1
         with:
-          path: opencollective-frontend/.next/cache
-          key: ${{ runner.os }}-next-cache-${{ hashFiles('opencollective-frontend/**') }}
+          path: opencollective-frontend/.next
+          key: ${{ runner.os }}-nextjs-${{ env.FRONTEND_COMMIT_HASH }}
           restore-keys: |
-            ${{ runner.os }}-next-cache-${{ hashFiles('opencollective-frontend/**') }}
-            ${{ runner.os }}-next-cache-
+            ${{ runner.os }}-nextjs-${{ env.FRONTEND_COMMIT_HASH }}
+            ${{ runner.os }}-nextjs-
 
       - name: Build (frontend)
         if: steps.next-build.outputs.cache-hit != 'true'
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -164,21 +164,22 @@ jobs:
         working-directory: opencollective-frontend
         run: echo "FRONTEND_COMMIT_HASH=$(git rev-parse HEAD)" >> $GITHUB_ENV
 
-      - name: Restore build (frontend)
+      - name: Restore .next build (frontend)
         uses: actions/cache@v1
         id: next-build
         with:
-          path: opencollective-frontend/dist
+          path: opencollective-frontend/.next
           key: ${{ runner.os }}-next-build-${{ env.FRONTEND_COMMIT_HASH }}
 
-      - name: Restore .next folder (frontend)
+      - name: Restore .next cache (frontend)
+        if: steps.next-build.outputs.cache-hit != 'true'
         uses: actions/cache@v1
         with:
-          path: opencollective-frontend/.next
-          key: ${{ runner.os }}-nextjs-${{ env.FRONTEND_COMMIT_HASH }}
+          path: opencollective-frontend/.next/cache
+          key: ${{ runner.os }}-next-cache-${{ env.FRONTEND_COMMIT_HASH }}
           restore-keys: |
-            ${{ runner.os }}-nextjs-${{ env.FRONTEND_COMMIT_HASH }}
-            ${{ runner.os }}-nextjs-
+            ${{ runner.os }}-next-cache-${{ env.FRONTEND_COMMIT_HASH }}
+            ${{ runner.os }}-next-cache-
 
       - name: Build (frontend)
         if: steps.next-build.outputs.cache-hit != 'true'
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -160,22 +160,23 @@ jobs:
         working-directory: opencollective-frontend
         run: CYPRESS_INSTALL_BINARY=0 npm ci --prefer-offline --no-audit
 
+      - name: Set commit hash (frontend)
+        working-directory: opencollective-frontend
+        run: echo "FRONTEND_COMMIT_HASH=$(git rev-parse HEAD)" >> $GITHUB_ENV
+
       - name: Restore build (frontend)
         uses: actions/cache@v1
         id: next-build
         with:
           path: opencollective-frontend/dist
-          key: ${{ runner.os }}-next-build-${{ hashFiles('opencollective-frontend/**') }}
+          key: ${{ runner.os }}-next-build-${{ env.FRONTEND_COMMIT_HASH }}
 
       - name: Restore .next cache (frontend)
         if: steps.next-build.outputs.cache-hit != 'true'
         uses: actions/cache@v1
         with:
           path: opencollective-frontend/.next/cache
-          key: ${{ runner.os }}-next-cache-${{ hashFiles('opencollective-frontend/**') }}
-          restore-keys: |
-            ${{ runner.os }}-next-cache-${{ hashFiles('opencollective-frontend/**') }}
-            ${{ runner.os }}-next-cache-
+          key: ${{ runner.os }}-next-cache-${{ env.FRONTEND_COMMIT_HASH }}
 
       - name: Build (frontend)
         if: steps.next-build.outputs.cache-hit != 'true'
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -172,10 +172,14 @@ jobs:
           key: ${{ runner.os }}-next-build-${{ env.FRONTEND_COMMIT_HASH }}
 
       - name: Restore .next cache (frontend)
+        if: steps.next-build.outputs.cache-hit != 'true'
         uses: actions/cache@v1
         with:
-          path: opencollective-frontend/.next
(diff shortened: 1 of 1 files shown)


znarf:
You're writing it's for "next cache". But it's also used for "build cache", right ?

znarf (approved):
Great if it works.

SudharakaP:
> You're writing it's for "next cache". But it's also used for "build cache", right ?

@znarf : Yes, I did some more simplifications to the action; let me test this a bit more thoroughly with some more CI restarts. 🤔

.github/workflows/e2e.yml
```diff
@@ -160,22 +160,23 @@ jobs:
...
 
       - name: Restore .next cache (frontend)
-        if: steps.next-build.outputs.cache-hit != 'true'
         uses: actions/cache@v1
         with:
           path: opencollective-frontend/.next/cache
-          key: ${{ runner.os }}-next-cache-${{ hashFiles('opencollective-frontend/**') }}
-          restore-keys: |
-            ${{ runner.os }}-next-cache-${{ hashFiles('opencollective-frontend/**') }}
-            ${{ runner.os }}-next-cache-
```
znarf, line 178: Why removing this line?

znarf:
Can you change the title of the PR then? It's misleading.

.github/workflows/e2e.yml
znarf: Do you understand what it does?

SudharakaP: @znarf : My understanding is if the key wasn't hit, then the `restore-key` is used and but in this case both are the same which didn't make sense to me. So I was trying out removing it to see what happens. 🤔

SudharakaP:
> Can you change the title of the PR then? It's misleading.

@znarf : Done. 

Also without the `restore-key` lines you've mentioned above; https://github.com/opencollective/opencollective-api/pull/6274#discussion_r669137110, the caching seems to work fine. However I need to test this after the next merge to frontend to be certain. Because I think we are hitting https://github.com/actions/cache/issues/191; when the cache is outdated in a workflow run it doesn't seem to update and correct it in subsequent runs. 🤔

.github/workflows/e2e.yml
znarf: In case the first restore key is not found, it fall back on the partial one. This is on purpose and should not be removed if there is no explicit reason.

https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows

znarf: (Please don't mark such ongoing discussions as resolved, it's misleading.)

SudharakaP:
@znarf @Betree : This is ready to be merged. I tested this be restarting the CI several times. Note that now the cache restore time is a few seconds. 

image

I've removed and simplified some things and my thinking is it's more cleaner now. We probably shouldn't rely on hashing the directory `opencollective-frontend` especially since after the build process, or even after the cache restore the directory changes (`dist`, `.next` directories are added) and thus the next calls to `hashFiles` will be unreliable I think; anyways let me know if you see any issues.

.github/workflows/e2e.yml
znarf: I can't accept the `restore-keys` to be removed if there is no good reason for it, it was built like this on purpose.

SudharakaP: @znarf : That makes sense, I've re-added the cache keys back. Will think more on the reasoning behind the addition of these fallback keys. 🤔

znarf: @SudharakaP The reason is giving there: https://github.com/opencollective/opencollective-api/pull/6274#discussion_r669380270

.github/workflows/e2e.yml
```diff
@@ -160,19 +160,22 @@ jobs:
...
       - name: Restore build (frontend)
         uses: actions/cache@v1
         id: next-build
         with:
           path: opencollective-frontend/dist
-          key: ${{ runner.os }}-next-build-${{ hashFiles('opencollective-frontend/**') }}
+          key: ${{ runner.os }}-next-build-${{ env.FRONTEND_COMMIT_HASH }}
 
       - name: Restore .next cache (frontend)
-        if: steps.next-build.outputs.cache-hit != 'true'
```
znarf, line 171: Why are you removing this?

.github/workflows/e2e.yml
```diff
@@ -160,19 +160,22 @@ jobs:
...
         uses: actions/cache@v1
         with:
-          path: opencollective-frontend/.next/cache
-          key: ${{ runner.os }}-next-cache-${{ hashFiles('opencollective-frontend/**') }}
+          path: opencollective-frontend/.next
+          key: ${{ runner.os }}-nextjs-${{ env.FRONTEND_COMMIT_HASH }}
```
znarf, line 178: If you modify the `key`, you should modify the `restore-keys` similarly.

.github/workflows/e2e.yml
znarf: This should not be changed if there is not good reason for it.

.github/workflows/e2e.yml
```diff
@@ -160,19 +160,22 @@ jobs:
...
-          key: ${{ runner.os }}-next-build-${{ hashFiles('opencollective-frontend/**') }}
+          key: ${{ runner.os }}-next-build-${{ env.FRONTEND_COMMIT_HASH }}
 
       - name: Restore .next cache (frontend)
-        if: steps.next-build.outputs.cache-hit != 'true'
         uses: actions/cache@v1
         with:
-          path: opencollective-frontend/.next/cache
-          key: ${{ runner.os }}-next-cache-${{ hashFiles('opencollective-frontend/**') }}
+          path: opencollective-frontend/.next
```
znarf, line 177: Why did you decide to cache the whole .next directory and not the cache directory like before?

.github/workflows/e2e.yml
znarf: If we're able to restore the build, no reason to restore .next cache.

.github/workflows/e2e.yml
SudharakaP: @znarf : Sorry, this is a mistake on my part. I have readjusted this to match the key. 👍🏽

.github/workflows/e2e.yml
SudharakaP: @znarf : If we don't remove this when we restart the github action builds it will fail with; https://github.com/opencollective/opencollective-api/runs/3077008271. My understanding is that the reason it was working before is actually because even when the build is restarted (with no changes) the build step will always run. To demonstrate this I have created the commit; 6e37d9c (which is what we had before this PR). I've manually restarted the build for this commit. Note that in that build run (https://github.com/opencollective/opencollective-api/runs/3077512247), the cache was not found; although we expect it to be found since it already ran the first time. 

My thinking is that when we restart the build the value `hashFiles('opencollective-frontend/**')` is different that what we stored the cache for. The previous build steps post cache operation stores the cache value but at that time the whole `opencollective-frontend` directory undergoes changes and hence the hash value is not accurate. I couldn't find any previous builds where this build step was skipped tbh. So this is why I thought that this cache hit should be removed. Does that make sense?

.github/workflows/e2e.yml
SudharakaP: @znarf : Basically, if we don't cache the whole .next directory then when the e2e tests are run we get the following error due to the `.next` directory missing some things.

https://github.com/opencollective/opencollective-api/runs/3077008271

I've explained my rational for doing this at, https://github.com/opencollective/opencollective-api/pull/6274#discussion_r670552256. Let me know what you think on this. 🙂

znarf: It was working before and for months. How do you explain that it would now suddenly failing?

.github/workflows/e2e.yml
znarf: @SudharakaP Thank you for the explanations.

I'm afraid that if we remove the "skip", some steps are losing their purpose.

We'll need to look deeper into this before merging.

znarf:
Let's review the build steps and follow the same steps as the E2E on frontend side:

1) Restore .next build
2) Restore .next cache (if no match on .next build)
3) Build (if no match on .next build)

.github/workflows/e2e.yml
SudharakaP: @znarf : Yes, it was working, but each time it was working, the build step was run. Note that we had the same condition `if: steps.next-build.outputs.cache-hit != 'true'` for both this and the build step. So essentially, after this `.next/cache` is restored the build step was run again. The build step replaces the whole `.next` folder. Thus I think previously there was no point of running this cache restore at all; we could have just removed this `.next/cache` restore and it would still work the same. Does that make sense? 🙂

SudharakaP:
> Let's review the build steps and follow the same steps as the E2E on frontend side:
> 
>     1. Restore .next build
> 
>     2. Restore .next cache (if no match on .next build)
> 
>     3. Build (if no match on .next build)

@znarf : Yes we can try this way, and I think it should work. I've adjusted the code to mimic the front-end. We will have to wait again for something to be merged in the front-end to fully test this (as otherwise it will use the previous `opencollective-frontend/dist` cache).

znarf:
@SudharakaP Thank you, where is the PR that needs to be merged on the frontend?

SudharakaP:
> @SudharakaP Thank you, where is the PR that needs to be merged on the frontend?

Any PR will be fine. Here the thing is to run a complete test we need the `env.FRONTEND_COMMIT_HASH` to change so that it will run a cache invalidated build; then I will restart the github actions CI manually to make sure the caching work as well. That doesn't happen unless we merge something on the front-end. Does that make sense? Hope I explained this correctly 🙂

znarf:
@SudharakaP To avoid cache conflict, just rename the keys, add something like `-v2` in it.

SudharakaP:
> @SudharakaP To avoid cache conflict, just rename the keys, add something like `-v2` in it.

@znarf : I am doubtful that will do anything. It will invalidate the cache in the sense it will trigger the build step in the first run (when you commit the change). But when you manually restart the GitHub action (or push a subsequent commit) it will not do anything and use the old cache. Note that the cache key is dependent upon `env.FRONTEND_COMMIT_HASH` so unless we change that the test isn't complete. 

Sorry for taking time on this; but it seems nothing was merged on frontend today; maybe if you are okay I can merge a small dependency update such as, https://github.com/opencollective/opencollective-frontend/pull/6627. It's just that I love doing complete tests to the extent possible when doing this kinda change; and my understanding is that just changing the cache as you suggested doesn't complete the testing fully; it doesn't do anything to the front-end repo and doesn't cover what happens when a user pushes a PR here and then the front-end changes in the interim and then the user pushes another commit or manually restart the build. Hope I explained this correctly. 😄 😄

SudharakaP:
@znarf : Okay I've tested this as we just merged something on frontend. Now I think we are good. The e2e tests seems to be working alright. 👍🏽

znarf (approved):
Ok, let's try this!

PR #6286, opened 2021-07-16
Wise batch payment to fulfill expenses that did work

kewitz:
Don't scratch out the whole batch if some expense fails to be created.
There's still and edge-case related to validations that happen only when funding the batch or transaction but I think this is a step in the right direction.

### DIFF
--- a/server/paymentProviders/transferwise/index.ts
+++ b/server/paymentProviders/transferwise/index.ts
@@ -2,10 +2,12 @@ import crypto from 'crypto';
 
 import config from 'config';
 import express from 'express';
-import { compact, find, has, pick, split, toNumber } from 'lodash';
+import { compact, find, has, isEmpty, pick, split, toNumber } from 'lodash';
 import moment from 'moment';
 import { v4 as uuid } from 'uuid';
 
+import activities from '../../constants/activities';
+import status from '../../constants/expense_status';
 import { BadRequest, TransferwiseError } from '../../graphql/errors';
 import cache from '../../lib/cache';
 import logger from '../../lib/logger';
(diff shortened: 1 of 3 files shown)


server/paymentProviders/transferwise/index.ts
```diff
@@ -243,24 +248,25 @@ async function createExpensesBatchGroup(
...
           token,
-        });
-        return transfer.id;
-      }),
+        })
+          .then(({ transfer }) => transfer.id)
+          .catch(() => {
+            // createTransfer will automatically mark the Expense as error and comment the issue
+            // Here we just ignore so we can pay the rest of the batch.
+          }),
```
SudharakaP, line 261: @kewitz : Just a suggestion, but my thinking here is to just log the exception (without throwing it) as having this empty block might cause some hard to find bugs down the pipe. For example; if an error happens at `Expense.prototype.createActivity` without this logging it will not show the real error anywhere but rather just the, `All expenses failed, there is no batch to pay.`

kewitz: This catch belongs to a single expense promise chain. Notice that createTransfer is already taking care of logging its errors.

SudharakaP: @kewitz : Yes, `createTransfer` is already taking care of most of the logging, except within the `createTransfer` catch block; 

https://github.com/opencollective/opencollective-api/blob/5a577324a84a056f7e2a7bc398219e1773603d4a/server/paymentProviders/transferwise/index.ts#L184-L189

we have `createActivity`. If that throws the error it won't be printed anywhere I think. Hope that makes sense. Let me know if I am missing something here. 😄

PR #6291, opened 2021-07-19
Prevent Sending Email for Transfers from Event to Parent Collective

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4382

Peek 2021-07-19 11-02

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -578,6 +578,7 @@ const sendOrderConfirmedEmail = async (order, transaction) => {
       }
     }
 
+    // Don't send a "thank you" email when moving funds between a sub-collective and its parent
     if (fromCollective.ParentCollectiveId !== collective.id) {
       const emailOptions = {
         from: `${collective.name} <no-reply@${collective.slug}.opencollective.com>`,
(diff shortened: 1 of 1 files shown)


server/lib/payments.js
```diff
@@ -578,12 +578,14 @@ const sendOrderConfirmedEmail = async (order, transaction) => {
       }
     }
 
-    const emailOptions = {
-      from: `${collective.name} <no-reply@${collective.slug}.opencollective.com>`,
-      attachments,
-    };
+    if (fromCollective.ParentCollectiveId !== collective.id) {
```
znarf, line 581: You should add a comment to explain why we're doing this.

SudharakaP: @znarf : I've added a comment explaining the logic. Thanks. 👍🏽

znarf (approved):

Betree (commented):
No blocker, but it's a bit sad that we check this at the end of the function because we still fetch data and call the PDF service to generate the receipts while it's not needed. It might have been better to check this in https://github.com/opencollective/opencollective-api/blob/d0f7ce9b3a8f48424764eb5312e936275c84f294/server/lib/payments.js#L397

PR #6332, opened 2021-07-22
Fix getTotalNetAmountReceivedAmount: Separate fees calculation

Betree:


server/lib/budget.js
```diff
@@ -105,19 +105,31 @@ export function getTotalAmountReceivedAmount(collective, { startDate, endDate, c
...
     excludeInternals: true,
   });
+
+  const totalFees = await sumCollectiveTransactions(collective, {
+    startDate,
+    endDate,
+    currency,
+    column: ['v0', 'v1'].includes(version) ? 'netAmountInCollectiveCurrency' : 'netAmountInHostCurrency',
+    transactionType: DEBIT,
+    kind: ['HOST_FEE'],
```
SudharakaP, line 128: Maybe better to refer this string from,

https://github.com/opencollective/opencollective-api/blob/7dacf7cdc7f7b45b5a9cc01149589e62630bb8ab/server/constants/transaction-kind.ts#L9 

so that we have a single place where we can change things in the future. 😄

Betree: Correct, it's good to use the enums when they exist

PR #6336, opened 2021-07-22
Add `hostSlug` and `kind` to `BudgetSection` Cache

SudharakaP:
It was pointed out in the following comment that we might be missing some values in the `BudgetSection` cache. This PR corrects it. Thanks, @Betree 😸 

https://github.com/opencollective/opencollective-frontend/pull/6631#pullrequestreview-712401571

I realize that its missing the `kind` as well, and I've added that as well. Let me know if it looks good. 

Related to https://github.com/opencollective/opencollective/issues/4453

### DIFF
--- a/server/graphql/cache.js
+++ b/server/graphql/cache.js
@@ -56,7 +56,7 @@ export function getGraphqlCacheKey(req) {
       }
       return `${req.body.operationName}_${queryHash}_${req.body.variables.slug}`;
     case 'BudgetSection':
-      if (!checkSupportedVariables(req, ['slug', 'limit'])) {
+      if (!checkSupportedVariables(req, ['slug', 'limit', 'hostSlug', 'kind'])) {
         return;
       }
       if (req.body.variables.limit !== 3) {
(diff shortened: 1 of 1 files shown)


Betree (commented):
What about `BudgetSectionWithHost`? We renamed the query in https://github.com/opencollective/opencollective-frontend/pull/6631, so this cache won't apply to it, yet accounts with hosts are the most important to cache since they're usually the ones that have transactions.

znarf:
Let's test that it works as expected in the development environment before merging.

We don't want to to do back and forth there.

Betree (commented):
Also, I already renamed `kinds` to `kind` in the frontend (https://github.com/opencollective/opencollective/issues/4378), so feel free to update `req.body.variables.kinds` below.

SudharakaP:
@Betree @znarf : I've changed the code to add `BudgetSectionWithHost` as well as renamed `kinds` to `kind`. I did some cache refreshes on my dev environment to see if it is correctly printed out. And it seems to be working correctly now. Let me know if you see any issues.

Betree (approved):
I haven't tested, but the code looks good. Let's ship this ASAP if it works, we got some performances issues this weekend.

znarf:
I tested locally and it seemed to work properly.

znarf (approved):

PR #6337, opened 2021-07-22
Do not Generate Unnecessary pdf receipts when email isn't sent

SudharakaP:
It was pointed out in https://github.com/opencollective/opencollective-api/pull/6291#pullrequestreview-712400504 (thanks @Betree 😄 ) that we generate an unnecessary pdf receipts in the api. I think this needs to be corrected and hence this PR. 

Related to https://github.com/opencollective/opencollective/issues/4382


Betree (approved):
:+1:

PR #6343, opened 2021-07-24
Add support to move transfer funds related queries to GraphQL V2

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4487

Front End PR: https://github.com/opencollective/opencollective-frontend/pull/6675


SudharakaP:
This is no longer needed as per https://github.com/opencollective/opencollective-frontend/pull/6675#discussion_r676039116. We can filter by `type` instead, so no need to add the `service`.

SudharakaP:
> This is no longer needed as per [opencollective/opencollective-frontend#6675 (comment)](https://github.com/opencollective/opencollective-frontend/pull/6675#discussion_r676039116). We can filter by `type` instead, so no need to add the `service`.

Actually we need this as we are using both filters, `type` and `service` as per, https://github.com/opencollective/opencollective-frontend/pull/6675#discussion_r676112597 😉

server/graphql/v2/interface/Account.js
```diff
@@ -751,8 +755,8 @@ export const AccountFields = {
   paymentMethods: {
     type: new GraphQLNonNull(new GraphQLList(PaymentMethod)),
     args: {
-      // TODO: Should filter by providerType
       types: { type: new GraphQLList(GraphQLString) },
+      service: { type: GraphQLString },
```
znarf, line 759: Make it a list `type: new GraphQLList(GraphQLString)` and keep it singular `service`.

Also, according to https://github.com/opencollective/opencollective/issues/4378 `types` should be **deprecated** and turned to `type`.

SudharakaP: @znarf : Done, let me create a new PR for the deprecation as it's not related to this issue.

SudharakaP: The deprecation is done at, https://github.com/opencollective/opencollective-api/pull/6344 and https://github.com/opencollective/opencollective-frontend/pull/6679

server/graphql/v2/interface/Account.js
```diff
@@ -274,6 +274,10 @@ const accountFieldsDefinition = () => ({
         type: new GraphQLList(GraphQLString),
         description: 'Filter on given types (creditcard, giftcard...)',
       },
+      service: {
+        type: new GraphQLList(GraphQLString),
```
Betree, line 278: Shouldn't we use the `PaymentMethodService` enum?

SudharakaP: @Betree : Yep, that makes sense. I've changed that to use the `PaymentMethodService` enum. 👍🏽

server/graphql/v2/interface/Account.js
```diff
@@ -756,9 +761,9 @@ export const AccountFields = {
   paymentMethods: {
     type: new GraphQLNonNull(new GraphQLList(PaymentMethod)),
     args: {
-      // TODO: Should filter by providerType
       types: { type: new GraphQLList(GraphQLString), deprecationReason: '2020-07-26: Please use type (singular)' },
       type: { type: new GraphQLList(GraphQLString) },
+      service: { type: PaymentMethodService },
```
znarf, line 766: It's expected to be a list right?

server/graphql/v2/interface/Account.js
znarf: Should still be a GraphQLList.

server/graphql/v2/interface/Account.js
SudharakaP: @znarf : Thanks. My mistake. Corrected it. 👍🏽

znarf (approved):

PR #6363, opened 2021-07-28
Rollback redis to 3.0.2

znarf:
problem related to https://github.com/NodeRedis/node-redis/pull/1508


znarf:
Actually https://devcenter.heroku.com/changelog-items/1932

PR #6369, opened 2021-07-29
CollectivePage: Move contributions section to GQLV2 and add Pagination

Betree:
Fix for https://github.com/opencollective/opencollective/issues/3788

WIP


server/graphql/v2/collection/MemberCollection.js
```diff
@@ -27,6 +28,9 @@ export const MemberOfCollection = new GraphQLObjectType({
       nodes: {
         type: new GraphQLList(MemberOf),
       },
+      roles: {
+        type: new GraphQLList(GraphQLJSON),
```
Betree, line 32: Why JSON rather than a strong type?

kewitz: I wasn't sure if I should present the collective type in the same array at the beginning so I left this as a prototype.
Now that I know we require roles per collective type I can add a proper type to this.

PR #6372, opened 2021-07-29
Ensure all `payExpense` paths update the settlements

Betree:
Moved the settlement update to `setPaid` to make sure all paths are updated regardless of the payout method, since in some cases we're calling the code to pay expenses from elsewhere (eg. for Transferwise, in the `handleTransferStateChange` webhook). 

Also added a migration to fix one invalid expense in the DB (97 transaction pairs).

### DIFF
--- a/server/models/Expense.js
+++ b/server/models/Expense.js
@@ -365,6 +365,11 @@ function defineModel() {
   Expense.prototype.setPaid = async function (lastEditedById) {
     await this.update({ status: status.PAID, lastEditedById });
 
+    // Update transactions settlement
+    if (this.data?.['isPlatformTipSettlement']) {
+      await models.TransactionSettlement.markExpenseAsSettled(this);
+    }
+
     try {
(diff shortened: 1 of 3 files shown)


server/models/Expense.js
```diff
@@ -365,6 +365,11 @@ function defineModel() {
   Expense.prototype.setPaid = async function (lastEditedById) {
     await this.update({ status: status.PAID, lastEditedById });
 
+    // Update transactions settlement
+    if (this.data?.['isPlatformTipSettlement']) {
```
znarf, line 369: Note for the future, this value `isPlatformTipSettlement` doesn't look good now.

We should move to something cleaner and explicit like Expense.type = 'SETTLEMENT'

znarf (approved):

server/models/Expense.js
Betree: I agree

PR #6377, opened 2021-07-30
Remove support for addFundsToOrg

znarf:


server/constants/activities.js
```diff
@@ -43,7 +43,6 @@ export default {
   COLLECTIVE_MONTHLY: 'collective.monthly',
   ORDERS_SUSPICIOUS: 'orders.suspicious',
   BACKYOURSTACK_DISPATCH_CONFIRMED: 'backyourstack.dispatch.confirmed',
-  ADDED_FUND_TO_ORG: 'added.fund.to.org',
```
Betree, line 46: I would rather move it under `// Not used anymore, leaving for historical reference` since we have occurrences for it in the DB

Betree (approved):

server/constants/activities.js
znarf: Ok!

PR #6402, opened 2021-08-04
Introduce Expense.type=SETTLEMENT

znarf:
Fix https://github.com/opencollective/opencollective/issues/4505


znarf:
@kewitz What was your experience when introducing `CHARGE`, do we have something to update in Frontend also?

kewitz:
@znarf yes, you need to add the constant and configure the default filtering actions for this type.

PR #6411, opened 2021-08-05
Add Support for Crypto Pending Orders to Show up in Dashboard

SudharakaP:
⚠️ Depends on https://github.com/opencollective/opencollective-api/pull/6378

Related to https://github.com/opencollective/opencollective/issues/4526

Front End PR: https://github.com/opencollective/opencollective-frontend/pull/6725

### DIFF
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -97,8 +97,8 @@ const orderMutations = {
         platformFee,
       };
 
-      const userArgent = req.header('user-agent');
-      const result = await createOrderLegacy(legacyOrderObj, req.loaders, req.remoteUser, req.ip, userArgent, req.mask);
+      const userAgent = req.header('user-agent');
+      const result = await createOrderLegacy(legacyOrderObj, req.loaders, req.remoteUser, req.ip, userAgent, req.mask);
       return { order: result.order, stripeError: result.stripeError, guestToken: result.order.data?.guestToken };
     },
   },
--- a/server/paymentProviders/thegivingblock/index.js
+++ b/server/paymentProviders/thegivingblock/index.js
@@ -90,14 +90,14 @@ export const processOrder = async order => {
   // TODO: update currency?
   await order.paymentMethod.update({ data: { ...order.paymentMethod.data, depositAddress } });
 
-  // Update order with pledgeId
-  await order.update({ data: { ...order.data, pledgeId } });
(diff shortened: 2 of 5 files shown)


server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -89,7 +90,7 @@ const orderMutations = {
         paymentMethod,
         fromCollective: fromCollective && { id: fromCollective.id },
         collective: { id: collective.id },
-        totalAmount: getOrderTotalAmount(order),
+        totalAmount: paymentMethod.type === PAYMENT_METHOD_TYPE.CRYPTO ? 1 : getOrderTotalAmount(order), // insert 1 as a dummy value for the pending order for crypto
```
znarf, line 93: That should not be needed. Frontend can simply pass a random value. No need to pollute this.

SudharakaP: Makes sense, I've corrected this. Actually I don't think this is needed anymore. 👍🏽

server/paymentProviders/thegivingblock/index.js
```diff
@@ -96,7 +97,7 @@ export const processOrder = async order => {
   const cryptoToFiatFxRate = await getFxRate(order.data.customData.pledgeCurrency, order.currency);
   if (cryptoToFiatFxRate) {
     const totalAmount = Math.round(order.data.customData.pledgeAmount * cryptoToFiatFxRate);
-    await order.update({ totalAmount });
+    await order.update({ totalAmount, status: orderStatus.PENDING });
```
znarf, line 100: Update to PENDING before or after or differently, as there is no guarantee to enter this condition at this stage.

server/paymentProviders/thegivingblock/index.js
```diff
@@ -147,9 +148,7 @@ export const confirmOrder = async order => {
     },
   };
 
-  const creditTransaction = await models.Transaction.createFromContributionPayload(transactionPayload);
-
-  return creditTransaction;
+  return await models.Transaction.createFromContributionPayload(transactionPayload);
```
znarf, line 151: No need to await.

server/paymentProviders/thegivingblock/index.js
SudharakaP: Ah, yes, I've changed this to update the order status before the `if` statement.

znarf (approved):
Looks good!

PR #6421, opened 2021-08-09
Correct .nvmrc to use node `14.x`

SudharakaP:
This was a mistake introduced in a previous commit; rather we should force the use of node 14.x for `nvm` just like we do in the front-end.

Related Slack thread: https://opencollective.slack.com/archives/C0RMV6F8C/p1628517217051100

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-engine-strict=true
+v14.x


.nvmrc
```diff
@@ -1 +1 @@
-engine-strict=true
\ No newline at end of file
+v14.x
```
znarf, line 1: It's best practice to have line break to avoid the warning you're seeing upper.

SudharakaP: I've corrected that, but where did you see this warning? I didn't see any warning in prettier or lint. 🤔

znarf: Screenshot 2021-08-09 at 16 56 33

Similar to this.

SudharakaP: Ah thanks, corrected that. 👍🏽

znarf (approved):

PR #6431, opened 2021-08-10
Chore: Type V2 Queries

kewitz:


server/graphql/v2/query/ExpensesQuery.ts
```diff
@@ -74,70 +74,72 @@ const updateFilterConditionsForReadyToPay = async (where, include): Promise<void
...
+    type: GraphQLDateTime,
+    description: 'Only return expenses that were created after this date',
+  },
+  searchTerm: {
+    type: GraphQLString,
+    description: 'The term to search',
+  },
+};
+
+const ExpensesQuery: Query<typeof ExpenseQueryArgs> = {
```
znarf, line 139: If `ExpenseQueryArgs` is an object, `typeof ExpenseQueryArgs` returns `object`. What's the point of typing with such a generic type?

kewitz: It is not that generic in this case. And after you get the `keyof` you'll be able to get all the argmuent keys you can have:
Screenshot from 2021-08-11 09-43-15

znarf: Just read about the following, I was originally reading it in a JavaScript way
https://www.typescriptlang.org/docs/handbook/2/typeof-types.html

Still looking like a hack, because only the keys are similar not the structure of the object itself. I'm no typescript expert.

How would we validate such argument? `fromAccount: {slug: "opensource"}`

kewitz: You're correct. This is just a naive solution to implement quick key-level arg validation without requiring manual typing of our objects.
The ideal solution would be integrating GraphQL and TS, which is quite more costly, both for refactoring existing code and for writing new code.

Betree:
Related: https://github.com/opencollective/opencollective/issues/4054

I'd be curious about experimenting with TypeGraphQL. It's quite different from what we do today so we should be careful, but if there's a way to make it work with our current queries and mutations that could be an interesting option to explore.

kewitz:
Closing this workaround PR so we can properly spec a better solution.

PR #6432, opened 2021-08-10
Add Confirmation Email for Crypto Deposit

SudharakaP:
⚠️ Should be merged only after https://github.com/opencollective/opencollective-api/pull/6411

Related to https://github.com/opencollective/opencollective/issues/4527

image

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -397,6 +397,8 @@ export const sendEmailNotifications = (order, transaction) => {
   // Don't send a "thank you" email when moving funds between a sub-collective and its parent
   if (transaction && order.fromCollective?.ParentCollectiveId !== order.collective?.id) {
     sendOrderConfirmedEmail(order, transaction); // async
+  } else if (order.status === status.PENDING && order.paymentMethod?.type === 'crypto') {
+    sendCryptoOrderProcessingEmail(order);
   } else if (order.status === status.PENDING) {
     sendOrderProcessingEmail(order); // This is the one for the Contributor
     sendManualPendingOrderEmail(order); // This is the one for the Host Admins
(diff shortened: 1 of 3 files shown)


znarf (changes requested):
Don't reuse the function and the template. Use a separated function and a separated template.

SudharakaP:
> Don't reuse the function and the template. Use a separated function and a separated template.

Thanks. I've refactored it to a separate function and a separate template. 👍🏽

server/lib/payments.js
```diff
@@ -590,6 +592,28 @@ const sendOrderConfirmedEmail = async (order, transaction) => {
...
+  const { collective, fromCollective } = order;
+  const user = order.createdByUser;
+  const host = await collective.getHostCollective();
+
+  const data = {
+    order: order.info,
+    collective: collective.info,
+    host: host.info,
+    fromCollective: fromCollective.activity,
+    isCrypto: true,
```
znarf, line 607: I don't think that's needed.

templates/emails/order.crypto.processing.hbs
```diff
@@ -0,0 +1,54 @@
...
+{{/if}}
+
+<p>Thank you for your crypto contribution to {{collective.name}} in the amount of {{pledgeAmount}} {{pledgeCurrency}}, made on
+  {{moment order.createdAt}}.</p>
+
+<p>Your contribution is currently pending. Once you transmit the funds,
+   and they have been confirmed as received by <a href="https://opencollective.com/{{host.slug}}">{{host.name}}</a>,
+   the amount will be credited to {{collective.name}}. </p>
+
+<p>Note that we use <a href="https://www.thegivingblock.com/">Giving Block</a>
```
znarf, line 30: It's not necessary to mention The Giving Block.

server/lib/payments.js
SudharakaP: Yep, my mistake, I forgot to remove this after refactoring. Done. 👍🏽

templates/emails/order.crypto.processing.hbs
SudharakaP: Thanks, I've removed this paragraph. 👍🏽

znarf (changes requested):
We talked about having the deposit address or the QR code in the email.

QR code might be difficult but can we add the address somewhere?

SudharakaP:
> We talked about having the deposit address or the QR code in the email.
> 
> QR code might be difficult but can we add the address somewhere?

@znarf : Ops, something I've missed. Yes I've added the deposit address in the above email. Let me know if it looks good. The screenshot above is updated. 🤔

SudharakaP:
Merged as per https://opencollective.slack.com/archives/C0RMV6F8C/p1629475803098700

PR #6433, opened 2021-08-11
Updating recurring contribution should update Members #4519

leonardohofling:
Resolve opencollective/opencollective#4519


znarf (commented):
Thank you. Looking good.

About the migration, we want it to be a script (in the scripts/fixes folder) that we can use from the command line.

Ideally, we would first fetch the rows we want to fix and process them one by one with a sequelize update through the model. That way we could easily add some logging and default as a "dry run".

leonardohofling:
> Thank you. Looking good.
> 
> About the migration, we want it to be a script (in the scripts/fixes folder) that we can use from the command line.
> 
> Ideally, we would first fetch the rows we want to fix and process them one by one with a sequelize update through the model. That way we could easily add some logging and default as a "dry run".

Could you check again, please? I did some changes...

znarf (approved):
Looks good!

PR #6436, opened 2021-08-11
Prevent Sending Emails for Contributions from Host to its Own Collective

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4382#issuecomment-878711106

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -395,7 +395,11 @@ export const sendEmailNotifications = (order, transaction) => {
   debug('sendEmailNotifications');
   // for gift cards and manual payment methods
   // Don't send a "thank you" email when moving funds between a sub-collective and its parent
-  if (transaction && order.fromCollective?.ParentCollectiveId !== order.collective?.id) {
+  if (
+    transaction &&
+    order.fromCollective?.ParentCollectiveId !== order.collective?.id &&
+    order.fromCollective.id !== order.collective?.HostCollectiveId
+  ) {
     sendOrderConfirmedEmail(order, transaction); // async
   } else if (order.status === status.PENDING) {
     sendOrderProcessingEmail(order); // This is the one for the Contributor
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -407,9 +407,17 @@ export const sendEmailNotifications = (order, transaction) => {
   debug('sendEmailNotifications');
   if (
     transaction &&
+    // Check if transaction is from child collective to parent collective.
+    // fromCollective: child collective, collective: parent collective
     order.fromCollective?.ParentCollectiveId !== order.collective?.id &&
-    order.fromCollective?.id !== order.collective?.HostCollectiveId &&
+    // Check if transaction is from parent collective to child collective.
+    // fromCollective: parent collective, collective: child collective
     order.fromCollective?.id !== order.collective?.ParentCollectiveId &&
+    // Check if transaction is from host to one of its own collective
+    // fromCollective: host, collective: a collective whose host is fromCollective
+    order.fromCollective?.id !== order.collective?.HostCollectiveId &&
+    // Check is transaction is from a collective to its host
+    // fromCollective: a collective, collective: host of fromCollective
     order.fromCollective?.HostCollectiveId !== order.collective?.id
   ) {
     sendOrderConfirmedEmail(order, transaction); // async
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -398,7 +398,7 @@ export const sendEmailNotifications = (order, transaction) => {
   if (
     transaction &&
     order.fromCollective?.ParentCollectiveId !== order.collective?.id &&
-    order.fromCollective.id !== order.collective?.HostCollectiveId
+    order.fromCollective?.id !== order.collective?.HostCollectiveId
   ) {
     sendOrderConfirmedEmail(order, transaction); // async
   } else if (order.status === status.PENDING) {
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -391,14 +391,26 @@ export async function associateTransactionRefundId(transaction, refund, data) {
   return find([refundCredit, refundDebit, debit, credit], { id: transaction.id });
 }
 
+/*
+ * Send email notifications.
+ *
+ * Don't send emails when moving funds between a sub-collective(event/project) and its parent or
+ * from between a host and one of its collectives. Transaction types for each of these are:
+ *
+ * Child to Parent Collective: CREDIT
+ * Host to its own collective: CREDIT
+ * Parent Collective to Child: CREDIT
+ * Collective to its Host: CREDIT
+ *
+ */
 export const sendEmailNotifications = (order, transaction) => {
   debug('sendEmailNotifications');
-  // Don't send a "thank you" email when moving funds from a sub-collective to its parent or from a host to
-  // one of its collectives.
   if (
     transaction &&
     order.fromCollective?.ParentCollectiveId !== order.collective?.id &&
-    order.fromCollective?.id !== order.collective?.HostCollectiveId
+    order.fromCollective?.id !== order.collective?.HostCollectiveId &&
+    order.fromCollective?.id !== order.collective?.ParentCollectiveId &&
+    order.fromCollective?.HostCollectiveId !== order.collective?.id
   ) {
     sendOrderConfirmedEmail(order, transaction); // async
   } else if (order.status === status.PENDING) {
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -395,29 +395,26 @@ export async function associateTransactionRefundId(transaction, refund, data) {
  * Send email notifications.
  *
  * Don't send emails when moving funds between a sub-collective(event/project) and its parent or
- * from between a host and one of its collectives. Transaction types for each of these are:
+ * from between a host and one of its collectives.
  *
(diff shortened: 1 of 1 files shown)


server/lib/payments.js
```diff
@@ -395,7 +395,11 @@ export const sendEmailNotifications = (order, transaction) => {
   debug('sendEmailNotifications');
   // for gift cards and manual payment methods
   // Don't send a "thank you" email when moving funds between a sub-collective and its parent
-  if (transaction && order.fromCollective?.ParentCollectiveId !== order.collective?.id) {
+  if (
+    transaction &&
+    order.fromCollective?.ParentCollectiveId !== order.collective?.id &&
+    order.fromCollective.id !== order.collective?.HostCollectiveId
```
kewitz, line 401:
```suggestion
    order.fromCollective?.id !== order.collective?.HostCollectiveId
```

znarf (changes requested):
Can you update the code comment explaining the block?

SudharakaP:
> Can you update the code comment explaining the block?

Done, it seems that there was a older comment coming from a time where the block of code is doing something related to gift cards which I've removed as well.

server/lib/payments.js
```diff
@@ -393,9 +393,13 @@ export async function associateTransactionRefundId(transaction, refund, data) {
 
 export const sendEmailNotifications = (order, transaction) => {
   debug('sendEmailNotifications');
-  // for gift cards and manual payment methods
-  // Don't send a "thank you" email when moving funds between a sub-collective and its parent
-  if (transaction && order.fromCollective?.ParentCollectiveId !== order.collective?.id) {
+  // Don't send a "thank you" email when moving funds between a sub-collective and its parent or between host
+  // and one of its collectives.
```
znarf, line 397: So, it's a DEBIT or a CREDIT we're looking at? Assuming DEBIT.

So, case 1) let's say order.fromCollective (the recipient) = {slug: sustainoss-2020, type: event, parentCollective = sustainoss} and order.collective (the sender) = {slug: sustainoss, type: collective}. That would work.

So, case 2) let's say order.fromCollective (the recipient) = {slug: sustainoss, type: collective, hostCollective = opensource} and order.collective (the sender) = {slug: opensource, type: organization}. That doesn't work.

There is something wrongs it seems. Maybe by explaining this in comments the cases we would be able to read the code and make sure it's right?

By the way, have you tested the code?

SudharakaP: In your cases is the recipient and sender correct? My understanding is that, `order.fromCollective` should be the sender whereas the `order.collective` should be the recipient? 🤔 

Yes I have tested this. I think here what we are trying to fix is the contribution from host to one of it's collective part (the second point you mentioned). This one I tested with a org profile (host) contributing to one of it's collectives. Not sure if I missed anything, but it didn't work for you? I've just changed the comment to make it a bit more readable and to show from where to where the funds are going. 🤔

znarf: > So, it's a DEBIT or a CREDIT we're looking at? Assuming DEBIT.

Could you answer this question? I'm not working on the issue and, without this, reviewing is difficult. Can you make sure to double check for case 1) and case 2) and tell us what it is? Can you also add the information in the code comments?

SudharakaP: So I checked the transaction types and they are CREDIT for all cases, 

1) Child to Parent Collective: CREDIT
2) Host to its own collective: CREDIT
3) Parent Collective to Child: CREDIT
4) Collective to its Host: CREDIT

I've added some code comments, and also made sure to support the back transfers from the parent collective to the child and from the collective to its host. 

>Why would we be checking order.fromCollective?.ParentCollectiveId if order.fromCollective is the sender? We're lost and that's a problem. Let's make things clear.

This check in particular for the case where we transfer from the child collective (event/project) to it's parent. I've also added the condition `order.fromCollective?.id !== order.collective?.ParentCollectiveId` to support the case where the parent collective transfers to the child collective. 

Let me know if things look better now. 🤔

server/lib/payments.js
```diff
@@ -391,11 +391,27 @@ export async function associateTransactionRefundId(transaction, refund, data) {
...
+ *
+ */
 export const sendEmailNotifications = (order, transaction) => {
   debug('sendEmailNotifications');
-  // for gift cards and manual payment methods
-  // Don't send a "thank you" email when moving funds between a sub-collective and its parent
-  if (transaction && order.fromCollective?.ParentCollectiveId !== order.collective?.id) {
+  if (
+    transaction &&
+    order.fromCollective?.ParentCollectiveId !== order.collective?.id &&
```
znarf, line 410: I was expecting to see some comments before each case. It's not really readable having the comments on top. 

I was also expecting detailed information about what is `fromCollective` and `collective` in each case.

We're talking for days about making this code block understandable and we're still not there. Can you make sure it's understandable by someone not familiar with the code base?

SudharakaP: @znarf : Ops, sorry for the trouble. Let me try to make it better. I've added comments at each case explaining `fromCollective` and `collective` in each case. Let me know if you think if there's anything to expand on in the comments.

server/lib/payments.js
```diff
@@ -391,11 +391,35 @@ export async function associateTransactionRefundId(transaction, refund, data) {
...
+ *
+ */
 export const sendEmailNotifications = (order, transaction) => {
   debug('sendEmailNotifications');
-  // for gift cards and manual payment methods
-  // Don't send a "thank you" email when moving funds between a sub-collective and its parent
-  if (transaction && order.fromCollective?.ParentCollectiveId !== order.collective?.id) {
+  if (
+    transaction &&
+    // Check if transaction is from child collective to parent collective.
```
znarf, line 410:
```suggestion
    // Check if transaction is from child (event/project) to parent (collective/fund/host).
```

server/lib/payments.js
```diff
@@ -391,11 +391,35 @@ export async function associateTransactionRefundId(transaction, refund, data) {
...
+    // fromCollective: child collective, collective: parent collective
```
znarf, line 411:
```suggestion
    // fromCollective: child (event/project), collective: parent (collective/fund/host)
```

server/lib/payments.js
```diff
@@ -391,11 +391,35 @@ export async function associateTransactionRefundId(transaction, refund, data) {
...
+    order.fromCollective?.ParentCollectiveId !== order.collective?.id &&
+    // Check if transaction is from parent collective to child collective.
```
znarf, line 413:
```suggestion
    // Check if transaction is from parent (collective/fund/host) to child (event/project)
```

server/lib/payments.js
```diff
@@ -391,11 +391,35 @@ export async function associateTransactionRefundId(transaction, refund, data) {
...
+    // fromCollective: parent collective, collective: child collective
```
znarf, line 414:
```suggestion
    // fromCollective: parent (collective/fund/host) , collective: child (event/project)
```

server/lib/payments.js
```diff
@@ -391,11 +391,35 @@ export async function associateTransactionRefundId(transaction, refund, data) {
...
+    order.fromCollective?.id !== order.collective?.ParentCollectiveId &&
+    // Check if transaction is from host to one of its own collective
```
znarf, line 416:
```suggestion
    // Check if transaction is from host to one of its hosted collective/fund/project/event
```

server/lib/payments.js
```diff
@@ -391,11 +391,35 @@ export async function associateTransactionRefundId(transaction, refund, data) {
...
+    // fromCollective: host, collective: a collective whose host is fromCollective
```
znarf, line 417:
```suggestion
    // fromCollective: host, collective: a collective/fund/project/event
 whose host is fromCollective
```

server/lib/payments.js
```diff
@@ -391,11 +391,35 @@ export async function associateTransactionRefundId(transaction, refund, data) {
...
+    order.fromCollective?.id !== order.collective?.HostCollectiveId &&
+    // Check is transaction is from a collective to its host
```
znarf, line 419:
```suggestion
    // Check is transaction is from a collective/fund/project/event to its host
```

server/lib/payments.js
```diff
@@ -391,11 +391,35 @@ export async function associateTransactionRefundId(transaction, refund, data) {
...
+    // fromCollective: a collective, collective: host of fromCollective
```
znarf, line 420:
```suggestion
    // fromCollective: a collective/fund/project/event, collective: host of fromCollective
```

server/lib/payments.js
```diff
@@ -391,11 +391,35 @@ export async function associateTransactionRefundId(transaction, refund, data) {
...
+/*
+ * Send email notifications.
+ *
+ * Don't send emails when moving funds between a sub-collective(event/project) and its parent or
+ * from between a host and one of its collectives. Transaction types for each of these are:
+ *
+ * Child to Parent Collective: CREDIT
+ * Host to its own collective: CREDIT
+ * Parent Collective to Child: CREDIT
+ * Collective to its Host: CREDIT
```
znarf, line 403:
```suggestion
 * from between a host and one of its collectives.
 *
 * In all cases, transaction.type is CREDIT.
```

znarf (commented):
Thank you. Finally looking good!

Added a bunch of suggestions, feel free to iterate on it.

SudharakaP:
Merged as per https://opencollective.slack.com/archives/C0RMV6F8C/p1629475803098700

PR #6439, opened 2021-08-12
fix(collective): change slug generation to prioritise name over twitter handle

FluxZA:
Previously  generateSlug function chose the first non-null suggested slug, but the Collective name was passed last in the array of suggestions. Update generateSlug to clarify importance of suggestion order and change Collective name to be the first suggestion passed in.

MVP for opencollective/opencollective#2856

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -617,15 +617,15 @@ function defineModel() {
           }
           let potentialSlugs,
             useSlugify = true;
+          // Populate potentialSlugs, priority of choices is the same as order in the array
           if (instance.isIncognito) {
             useSlugify = false;
             potentialSlugs = [`incognito-${uuid().split('-')[0]}`];
           } else {
             potentialSlugs = [
-              instance.slug,
+              instance.name ? instance.name.replace(/ /g, '-') : null,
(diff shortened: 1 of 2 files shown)


SudharakaP:
This is interesting, it seems that looking at the CI, https://github.com/opencollective/opencollective/issues/4305#issue-896290059 has resurfaced. 🤔

Betree:
> This is interesting, it seems that looking at the CI, [opencollective/opencollective#4305 (comment)](https://github.com/opencollective/opencollective/issues/4305#issue-896290059) has resurfaced. thinking

I see this error almost every time the CI runs on a fork PR. Not sure why

FluxZA:
I re-based my branch and added a test for the correct behaviour. Sorry that this was not in the original PR.

Betree (approved):
Sorry for taking so long on the review!

The code looks amazing, thanks for adding these rich comments :pray: These functions are probably as old as Open Collective itself and really deserve a refresh. Your update is a great step forward!

As mentioned, tests are flawky when ran on forks but I tested `test/server/models/Collective.test.js` locally (which does most of the testing on slugs) and everything seems to work.

I'm merging now, and sending you an invite to join https://github.com/orgs/opencollective/teams/contributors/members.

PR #6440, opened 2021-08-12
Account.childrenAccounts resolver

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4383

### DIFF
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -578,11 +578,7 @@ const accountFieldsDefinition = () => ({
         type: new GraphQLList(AccountType),
       },
     },
-    async resolve(account, args, req) {
-      if (!req.remoteUser?.isAdminOfCollective(account)) {
-        throw new Unauthorized('You need to be logged in as an admin of the account to see its children accounts');
-      }
-
+    async resolve(account, args) {
       const where = {
         ParentCollectiveId: account.id,
       };
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -185,10 +185,11 @@ const accountFieldsDefinition = () => ({
         query.where.role = { [Op.in]: roles };
       }
 
-      let conditionOnMemberCollective;
-      if (args.type) {
-        const types = args.type.split(',');
-        conditionOnMemberCollective = { type: { [Op.in]: types } };
+      const conditionOnMemberCollective = {};
+      if (args.accountType && args.accountType.length > 0) {
+        conditionOnMemberCollective.type = {
+          [Op.in]: args.accountType.map(value => AccountTypeToModelMapping[value]),
+        };
       }
 
       query.include = [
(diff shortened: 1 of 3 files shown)


server/graphql/v2/interface/Account.js
```diff
@@ -568,6 +569,43 @@ const accountFieldsDefinition = () => ({
...
+      limit: { type: GraphQLInt, defaultValue: 100 },
+      offset: { type: GraphQLInt, defaultValue: 0 },
+      accountType: {
+        type: new GraphQLList(AccountType),
+      },
+    },
+    async resolve(account, args, req) {
+      if (!req.remoteUser?.isAdminOfCollective(account)) {
+        throw new Unauthorized('You need to be logged in as an admin of the account to see its children accounts');
+      }
```
znarf, line 584: Why did you go for this? Doesn't seem necessary.

server/graphql/v2/interface/Account.js
```diff
@@ -185,10 +185,11 @@ const accountFieldsDefinition = () => ({
...
 
-      let conditionOnMemberCollective;
-      if (args.type) {
-        const types = args.type.split(',');
-        conditionOnMemberCollective = { type: { [Op.in]: types } };
+      const conditionOnMemberCollective = {};
+      if (args.accountType && args.accountType.length > 0) {
+        conditionOnMemberCollective.type = {
+          [Op.in]: args.accountType.map(value => AccountTypeToModelMapping[value]),
+        };
```
znarf, line 192: That's right. Useful for for Individual -> User.

kewitz: Keeping it consistent with the existing behaviour for V2 account querying.

server/graphql/v2/interface/Account.js
kewitz: I'll remove this.

znarf:
That was merged a bit fast. I wish Ben would have had the opportunity to look at this.

kewitz:
@Betree, could you take a look at this PR? I underrated this issue a bit and merged it last week after François' review.
The main discussion I had with François after was related to if we should have this abstract children resolver or implement "events" and "projects" resolvers.

server/graphql/v2/interface/Account.js
```diff
@@ -568,6 +569,39 @@ const accountFieldsDefinition = () => ({
       };
     },
   },
+  childrenAccounts: {
+    type: AccountCollection,
```
Betree, line 573: We can make this non-nullable

server/graphql/v2/interface/Account.js
```diff
@@ -568,6 +569,39 @@ const accountFieldsDefinition = () => ({
...
+    args: {
+      limit: { type: GraphQLInt, defaultValue: 100 },
+      offset: { type: GraphQLInt, defaultValue: 0 },
```
Betree, line 576: These ones too, since they have a defaultValue. It's good practice to enforce a `limit`

Betree (commented):
This looks great, I like the fact that we're implementing this as a generic field. 

Some minor feedback below but nothing blocking.

PR #6443, opened 2021-08-13
Generate New Access Tokens Automatically

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4546

### DIFF
--- a/server/paymentProviders/thegivingblock/index.js
+++ b/server/paymentProviders/thegivingblock/index.js
@@ -17,24 +17,45 @@ const AES_ENCRYPTION_METHOD = config.thegivingblock.aesEncryptionMethod;
 const API_URL = config.thegivingblock.apiUrl;
 const GIVINGBLOCK_USERNAME = config.thegivingblock.username;
 const GIVINGBLOCK_PASSWORD = config.thegivingblock.password;
+const MAX_REQUEST_RETRIES = 3;
 
-async function apiRequest(path, options = {}) {
+async function apiRequest(path, options = {}, account) {
   const response = await fetch(`${API_URL}${path}`, options);
   const result = await response.json();
 
-  // Whenever and api request is made we check if access token is expired and if
-  // so we login again (https://app.gitbook.com/@the-giving-block/s/public-api-documentation/#authentication-flow).
-  // Access tokens are only valid for 2 hours.
+  return await handleErrorsAndRetry(result, path, options, account);
+  // console.log(result);
+}
+
+/*
+ * Whenever and api request is made we check if access token is expired and if so we login again.
+ * Refer: https://app.gitbook.com/@the-giving-block/s/public-api-documentation/#authentication-flow.
+ * Access tokens are only valid for 2 hours.
+ */
+async function handleErrorsAndRetry(result, path, options = {}, account = null) {
   if (result.data.errorMessage) {
-    if (result.data.meta.errorCode === 'INVALID_JWT_TOKEN') {
+    if (result.data.meta.errorCode === 'INVALID_JWT_TOKEN' && account) {
       logger.debug('Access token is invalid. Requesting a new one.');
-      return await login(GIVINGBLOCK_USERNAME, GIVINGBLOCK_PASSWORD);
+      let error;
+      for (let retries = 0; retries < MAX_REQUEST_RETRIES; retries++) {
+        try {
+          const { accessToken, refreshToken } = await login(GIVINGBLOCK_USERNAME, GIVINGBLOCK_PASSWORD);
+          await account.update({ data: { ...account.data, accessToken, refreshToken } });
+          const headers = {
+            Authorization: `Bearer ${accessToken}`,
+          };
+          options.body.set('refreshToken', refreshToken);
+          const response = await fetch(`${API_URL}${path}`, { headers, ...options });
+          const result = await response.json();
+          return result.data;
+        } catch (err) {
+          error = err;
+        }
+      }
+      throw error;
     }
     throw new Error(`The Giving Block: ${result.data.errorMessage} ${result.data.meta.errorCode}`);
   }
-
-  // console.log(result);
-
   return result.data;
 }
 
@@ -46,24 +67,24 @@ export async function login(login, password) {
   return apiRequest(`/login`, { method: 'POST', body });
 }
 
-export async function refresh(refreshToken) {
+export async function refresh(account) {
   const body = new URLSearchParams();
-  body.set('refreshToken', refreshToken);
+  body.set('refreshToken', account.data.refreshToken);
 
-  return apiRequest(`/refresh-tokens`, { method: 'POST', body });
+  return apiRequest(`/refresh-tokens`, { method: 'POST', body }, account);
 }
 
-export async function getOrganizationsList(accessToken) {
+export async function getOrganizationsList(account) {
   const headers = {
-    Authorization: `Bearer ${accessToken}`,
+    Authorization: `Bearer ${account.data.accessToken}`,
   };
 
   return apiRequest(`/organizations/list`, { headers });
 }
 
-export async function createDepositAddress(accessToken, { organizationId, pledgeAmount, pledgeCurrency } = {}) {
+export async function createDepositAddress(account, { organizationId, pledgeAmount, pledgeCurrency } = {}) {
   const headers = {
-    Authorization: `Bearer ${accessToken}`,
+    Authorization: `Bearer ${account.data.accessToken}`,
   };
 
   const body = new URLSearchParams();
@@ -72,7 +93,7 @@ export async function createDepositAddress(accessToken, { organizationId, pledge
   body.set('pledgeAmount', pledgeAmount);
   body.set('pledgeCurrency', pledgeCurrency);
 
-  return apiRequest(`/deposit-address`, { method: 'POST', body, headers });
+  return apiRequest(`/deposit-address`, { method: 'POST', body, headers }, account);
 }
 
 export const processOrder = async order => {
@@ -85,11 +106,11 @@ export const processOrder = async order => {
 
   // refresh credentials
   // TODO: we normally have to do it only every 2 hours but this handy for now
-  const { accessToken, refreshToken } = await refresh(account.data.refreshToken);
+  const { accessToken, refreshToken } = await refresh(account);
   await account.update({ data: { ...account.data, accessToken, refreshToken } });
 
   // create wallet address
-  const { depositAddress, pledgeId } = await createDepositAddress(account.data.accessToken, {
+  const { depositAddress, pledgeId } = await createDepositAddress(account, {
     organizationId: account.data.organizationId,
     pledgeAmount: order.data.customData.pledgeAmount,
     pledgeCurrency: order.data.customData.pledgeCurrency,
--- a/server/paymentProviders/thegivingblock/index.js
+++ b/server/paymentProviders/thegivingblock/index.js
@@ -36,13 +36,12 @@ async function handleErrorsAndRetry(result, path, options = {}, account = null) 
       logger.debug('Access token is invalid. Requesting a new one.');
       let error;
       try {
-        const { accessToken, refreshToken } = await login(USERNAME, PASSWORD);
-        await account.update({ data: { ...account.data, accessToken, refreshToken } });
+        await login(USERNAME, PASSWORD, account);
         if (options.body?.get('refreshToken')) {
-          options.body.set('refreshToken', refreshToken);
+          options.body.set('refreshToken', account.data.refreshToken);
         }
         if (options.headers?.Authorization) {
-          options.headers.Authorization = `Bearer ${accessToken}`;
+          options.headers.Authorization = `Bearer ${account.data.accessToken}`;
         }
         const response = await fetch(`${API_URL}${path}`, options);
         const result = await response.json();
@@ -57,12 +56,13 @@ async function handleErrorsAndRetry(result, path, options = {}, account = null) 
   return result.data;
 }
 
-export async function login(login, password) {
+export async function login(login, password, account) {
   const body = new URLSearchParams();
   body.set('login', login);
   body.set('password', password);
 
-  return apiRequest(`/login`, { method: 'POST', body });
+  const { accessToken, refreshToken } = await apiRequest(`/login`, { method: 'POST', body });
+  return account.update({ data: { ...account.data, accessToken, refreshToken } });
 }
 
 export async function refresh(account) {
--- a/server/paymentProviders/thegivingblock/index.js
+++ b/server/paymentProviders/thegivingblock/index.js
@@ -15,8 +15,8 @@ const AES_ENCRYPTION_KEY = config.thegivingblock.aesEncryptionKey;
 const AES_ENCRYPTION_IV = config.thegivingblock.aesEncryptionIv;
 const AES_ENCRYPTION_METHOD = config.thegivingblock.aesEncryptionMethod;
 const API_URL = config.thegivingblock.apiUrl;
-const GIVINGBLOCK_USERNAME = config.thegivingblock.username;
-const GIVINGBLOCK_PASSWORD = config.thegivingblock.password;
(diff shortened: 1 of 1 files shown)


znarf (changes requested):
We're not supposed to do this and use the password, `refreshToken` should be enough.

Investigate and understand why it's failing before using this approach as last resort.

znarf (changes requested):
Also, the title of the PR is misleading, we're not refreshing, we're `login` again.

There is a `refresh` method and it's not the same at all.

SudharakaP:
> Also, the title of the PR is misleading, we're not refreshing, we're `login` again.
> 
> There is a `refresh` method and it's not the same at all.

Yes, I agree, re-titled the PR. 

> We're not supposed to do this and use the password, `refreshToken` should be enough.
> 
> Investigate and understand why it's failing before using this approach as last resort.

Makes sense, I did some more testing and observed that when we run `RefreshTokens` method although it says in documentation we get a `accessToken` and a `refreshToken`, I only see the `refreshToken` being refreshed. Never have I seen the `accessToken` being renewed. I think this is the root cause of the problem. And after some time, say 2 hours when the access token expires, we get this `INVALID_JWT_TOKEN` response. 🤔 

Do you think we should close this PR or use this as a workaround for now? Maybe I'll send an email to The Giving Block on this note? 🤔

SudharakaP:
Moreover I think maybe we should use this even if the `accessToken` do get refreshed. Note that, [in the doc](https://app.gitbook.com/@the-giving-block/s/public-api-documentation/#authentication-flow) it's mentioned to use this login method as a fallback and I think that is what we are doing here. Ideally it should not throw the `INVALID_JWT_TOKEN` but only if it does we login again. 🤔

znarf (changes requested):
We need something a bit more sophisticated. We would need to login AND retry the failing request. If necessary, we also want a retry attempts protection to avoid infinite loops.

SudharakaP:
> We need something a bit more sophisticated. We would need to login AND retry the failing request. If necessary, we also want to count and limit retry attempts, as a protection to avoid infinite loops.

@znarf : Thanks for the feedback. I've added some code to retry 3 times before giving up and throwing the error. Let me know if it looks good. 🤔

server/paymentProviders/thegivingblock/index.js
```diff
@@ -7,27 +7,52 @@ import fetch from 'node-fetch';
...
+ */
+async function handleErrors(result, path, options = {}) {
+  if (result.data.errorMessage) {
+    if (result.data.meta.errorCode === 'INVALID_JWT_TOKEN') {
+      logger.debug('Access token is invalid. Requesting a new one.');
+      let error;
+      for (let retries = 0; retries < MAX_REQUEST_RETRIES; retries++) {
+        try {
+          await login(GIVINGBLOCK_USERNAME, GIVINGBLOCK_PASSWORD);
+          await apiRequest(path, options);
```
znarf, line 45: You're not refreshing the accessToken here ...

server/paymentProviders/thegivingblock/index.js
```diff
@@ -7,27 +7,53 @@ import fetch from 'node-fetch';
...
+async function handleErrors(result, path, options = {}, account = null) {
+  if (result.data.errorMessage) {
+    if (result.data.meta.errorCode === 'INVALID_JWT_TOKEN' && account) {
+      logger.debug('Access token is invalid. Requesting a new one.');
+      let error;
+      for (let retries = 0; retries < MAX_REQUEST_RETRIES; retries++) {
+        try {
+          const { accessToken, refreshToken } = await login(GIVINGBLOCK_USERNAME, GIVINGBLOCK_PASSWORD);
+          await account.update({ data: { ...account.data, accessToken, refreshToken } });
+          return await apiRequest(path, options);
```
znarf, line 46: You're still not updating the accessToken in the apiRequest, how do you expect this to work?

SudharakaP: @znarf : I am was trying some things here. This isn't done yet; should have marked this PR as draft earlier. I'll ping you once completed. I am trying to find a way to update the accessToken as well as resend the request. It's a bit tricky as in some requests we send the refresh token whereas in the deposit request we send the access token.

server/paymentProviders/thegivingblock/index.js
```diff
@@ -36,11 +62,11 @@ export async function login(login, password) {
   return apiRequest(`/login`, { method: 'POST', body });
 }
 
-export async function refresh(refreshToken) {
+export async function refresh(refreshToken, account) {
```
znarf, line 65: The way you're adding `account` everywhere, without refactoring more, that doesn't looks good.

If going with this approach, you have to realize that `refreshToken` is already available in `account` and take the necessary steps to simplify the code.

That is basically valid for every function where you added the account parameter.

server/paymentProviders/thegivingblock/index.js
SudharakaP: Yep, my mistake. I've refactored the code.

server/paymentProviders/thegivingblock/index.js
SudharakaP: Thanks. Yes, I've refactored the code to use `account` instead of passing `refreshToken`. The testing of this is a bit tricky so what I did was to go to the contribution flow, change the `accessToken` and then progress in the contribution flow to the QR code section. Then since the `accessToken` is not valid anymore it will throw the `INVALID_JWT_TOKEN` and hence do the re-login. This is also valid if we "lose" the refresh token for some reason. So if we change the `refreshToken` in the middle and make a request this could will handle that as well. Let me know what you think. 🤔

server/paymentProviders/thegivingblock/index.js
```diff
@@ -7,24 +7,55 @@ import fetch from 'node-fetch';
...
+ * Access tokens are only valid for 2 hours.
+ */
+async function handleErrorsAndRetry(result, path, options = {}, account = null) {
+  if (result.data.errorMessage) {
+    if (result.data.meta.errorCode === 'INVALID_JWT_TOKEN' && account) {
+      logger.debug('Access token is invalid. Requesting a new one.');
+      let error;
+      for (let retries = 0; retries < MAX_REQUEST_RETRIES; retries++) {
+        try {
+          const { accessToken, refreshToken } = await login(GIVINGBLOCK_USERNAME, GIVINGBLOCK_PASSWORD);
```
znarf, line 42: When`login` in, you're going to `apiRequest`. If it fails, you're going to `handleErrorsAndRetry` and you then go to `login`. If it keeps failing, this is an infinite loop.

server/paymentProviders/thegivingblock/index.js
```diff
@@ -7,24 +7,55 @@ import fetch from 'node-fetch';
...
+          await account.update({ data: { ...account.data, accessToken, refreshToken } });
+          const headers = {
+            Authorization: `Bearer ${accessToken}`,
+          };
+          options.body.set('refreshToken', refreshToken);
```
znarf, line 47: That's a bit brute force and could lead to problems.

server/paymentProviders/thegivingblock/index.js
SudharakaP: @znarf : But according to the docs, the `login` request does not throw `INVALID_JWT_TOKEN` error. So if the login fails, it will throw an error. I've tested this with changing the login credentials along with the access token. What I am thinking is if the login fails the `result.data.meta.errorCode === 'INVALID_JWT_TOKEN'` is always false. 🤔

server/paymentProviders/thegivingblock/index.js
SudharakaP: Sorry, I am a bit confused on this one. Could you elaborate on what problems this could lead to? 🤔 🙂

SudharakaP:
@znarf : Done some changes to remove the retry loop and add some conditionals around the token resetting. Let me know if all looks good now. 👍🏽

server/paymentProviders/thegivingblock/index.js
```diff
@@ -7,24 +7,55 @@ import fetch from 'node-fetch';
...
+    if (result.data.meta.errorCode === 'INVALID_JWT_TOKEN' && account) {
+      logger.debug('Access token is invalid. Requesting a new one.');
+      let error;
+      try {
+        const { accessToken, refreshToken } = await login(GIVINGBLOCK_USERNAME, GIVINGBLOCK_PASSWORD);
+        await account.update({ data: { ...account.data, accessToken, refreshToken } });
+        if (options.body.get('refreshToken')) {
+          options.body.set('refreshToken', refreshToken);
+        }
+        if (options.headers) {
```
znarf, line 44:
```suggestion
        if (options.headers?.Authorization) {
```

server/paymentProviders/thegivingblock/index.js
```diff
@@ -7,24 +7,55 @@ import fetch from 'node-fetch';
...
+          options.headers = {
+            Authorization: `Bearer ${accessToken}`,
+          };
```
znarf, line 47:
```suggestion
          options.headers.Authorization: `Bearer ${accessToken}`;
```

server/paymentProviders/thegivingblock/index.js
```diff
@@ -75,11 +106,11 @@ export const processOrder = async order => {
 
   // refresh credentials
   // TODO: we normally have to do it only every 2 hours but this handy for now
-  const { accessToken, refreshToken } = await refresh(account.data.refreshToken);
+  const { accessToken, refreshToken } = await refresh(account);
```
znarf, line 109: refresh should take care of updating the account and return the updated account

server/paymentProviders/thegivingblock/index.js
```diff
@@ -7,24 +7,55 @@ import fetch from 'node-fetch';
...
+ */
+async function handleErrorsAndRetry(result, path, options = {}, account = null) {
+  if (result.data.errorMessage) {
+    if (result.data.meta.errorCode === 'INVALID_JWT_TOKEN' && account) {
+      logger.debug('Access token is invalid. Requesting a new one.');
+      let error;
+      try {
+        const { accessToken, refreshToken } = await login(GIVINGBLOCK_USERNAME, GIVINGBLOCK_PASSWORD);
+        await account.update({ data: { ...account.data, accessToken, refreshToken } });
+        if (options.body.get('refreshToken')) {
```
znarf, line 41: what if body is not set?

server/paymentProviders/thegivingblock/index.js
```diff
@@ -7,24 +7,55 @@ import fetch from 'node-fetch';
...
 import { getHostFee, getHostFeeSharePercent } from '../../lib/payments';
 import models from '../../models';
 
 const AES_ENCRYPTION_KEY = config.thegivingblock.aesEncryptionKey;
 const AES_ENCRYPTION_IV = config.thegivingblock.aesEncryptionIv;
 const AES_ENCRYPTION_METHOD = config.thegivingblock.aesEncryptionMethod;
 const API_URL = config.thegivingblock.apiUrl;
-
-async function apiRequest(path, options = {}) {
+const GIVINGBLOCK_USERNAME = config.thegivingblock.username;
```
znarf, line 18:
```suggestion
const THEGIVINGBLOCK_USERNAME = config.thegivingblock.username;
```

server/paymentProviders/thegivingblock/index.js
```diff
@@ -7,24 +7,55 @@ import fetch from 'node-fetch';
...
+const GIVINGBLOCK_PASSWORD = config.thegivingblock.password;
```
znarf, line 19:
```suggestion
const THEGIVINGBLOCK_PASSWORD = config.thegivingblock.password;
```

Consistency please, we never name anything "Giving Block" anywhere.

server/paymentProviders/thegivingblock/index.js
znarf: Could actually be shorten to `USERNAME` in this context, like other environment variables.

server/paymentProviders/thegivingblock/index.js
znarf: Could actually be shorten to `PASSWORD` in this context, like other environment variables.

server/paymentProviders/thegivingblock/index.js
SudharakaP: Done. 👍🏽

server/paymentProviders/thegivingblock/index.js
SudharakaP: Done. 👍🏽

server/paymentProviders/thegivingblock/index.js
SudharakaP: I've added the `account` update login within the `refresh` function. I think now it looks better. Let me know. 👍🏽

server/paymentProviders/thegivingblock/index.js
SudharakaP: Ah, that's a good point. I've added a null check here. I think that will take care of it without an error. 👍🏽

server/paymentProviders/thegivingblock/index.js
SudharakaP: Done. 👍🏽

server/paymentProviders/thegivingblock/index.js
SudharakaP: Done. 👍🏽

server/paymentProviders/thegivingblock/index.js
```diff
@@ -36,15 +36,13 @@ async function handleErrorsAndRetry(result, path, options = {}, account = null)
       logger.debug('Access token is invalid. Requesting a new one.');
       let error;
       try {
-        const { accessToken, refreshToken } = await login(GIVINGBLOCK_USERNAME, GIVINGBLOCK_PASSWORD);
+        const { accessToken, refreshToken } = await login(USERNAME, PASSWORD);
```
znarf, line 39: You forgot to refactor here.

znarf: No need to account.update as it's already done.

SudharakaP: But after login don't we have to update the `account` so that future requests can use that? Note that this login problem can occur for any request. Only in the refresh method we update the `account`. 🤔

znarf: `login` should update the account like `refresh`, why do we have different strategies?

SudharakaP: @znarf : Ah, I get it now. I've corrected this to make it similar to `refresh`. Let me know if all looks good. 👍🏽

server/paymentProviders/thegivingblock/index.js
```diff
@@ -7,53 +7,83 @@ import fetch from 'node-fetch';
...
 import models from '../../models';
 
 const AES_ENCRYPTION_KEY = config.thegivingblock.aesEncryptionKey;
 const AES_ENCRYPTION_IV = config.thegivingblock.aesEncryptionIv;
 const AES_ENCRYPTION_METHOD = config.thegivingblock.aesEncryptionMethod;
 const API_URL = config.thegivingblock.apiUrl;
-
-async function apiRequest(path, options = {}) {
+const USERNAME = config.thegivingblock.username;
+const PASSWORD = config.thegivingblock.password;
```
znarf, line 19: Line break please.

znarf (approved):
Good for me if it works. Not tested.

server/paymentProviders/thegivingblock/index.js
```diff
@@ -7,53 +7,84 @@ import fetch from 'node-fetch';
...
 }
 
-export async function refresh(refreshToken) {
+export async function refresh(account) {
   const body = new URLSearchParams();
-  body.set('refreshToken', refreshToken);
+  body.set('refreshToken', account.data.refreshToken);
 
-  return apiRequest(`/refresh-tokens`, { method: 'POST', body });
+  const { accessToken, refreshToken } = apiRequest(`/refresh-tokens`, { method: 'POST', body }, account);
```
znarf, line 73: Missing await here.

PR #6447, opened 2021-08-16
Transactions: use dataloader generators

Betree:
This PR should hopefully fix the memory leak that we're facing for a few weeks. Since the data loaders were created at the top level, my guess is that they were never cleared and grew indefinitely.


znarf (approved):
Let's have a try!

PR #6448, opened 2021-08-16
Fix payment method type

znarf:
Frontend: https://github.com/opencollective/opencollective-frontend/pull/6760

This was a rabbit hole but I think that PR put us in the right direction to clean that technical debt.

### DIFF
--- a/server/constants/paymentMethods.ts
+++ b/server/constants/paymentMethods.ts
@@ -17,7 +17,7 @@ export enum PAYMENT_METHOD_TYPE {
   COLLECTIVE = 'collective',
   HOST = 'host',
   ADAPTIVE = 'adaptive',
-  GIFT_CARD = 'giftcard',
+  GIFTCARD = 'giftcard',
   MANUAL = 'manual',
   CRYPTO = 'crypto',
 }
--- a/server/graphql/v2/enum/PaymentMethodType.ts
+++ b/server/graphql/v2/enum/PaymentMethodType.ts
@@ -1,10 +1,17 @@
 import { GraphQLEnumType } from 'graphql';
 
-import { PAYMENT_METHOD_TYPES } from '../../../constants/paymentMethods';
+import { PAYMENT_METHOD_TYPE } from '../../../constants/paymentMethods';
 
 export const PaymentMethodType = new GraphQLEnumType({
   name: 'PaymentMethodType',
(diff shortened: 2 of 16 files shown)


server/graphql/v2/enum/PaymentMethodType.ts
```diff
@@ -1,10 +1,10 @@
...
 
-import { PAYMENT_METHOD_TYPES } from '../../../constants/paymentMethods';
+import { PAYMENT_METHOD_TYPE } from '../../../constants/paymentMethods';
 
 export const PaymentMethodType = new GraphQLEnumType({
   name: 'PaymentMethodType',
-  values: PAYMENT_METHOD_TYPES.reduce((values, key) => {
-    return { ...values, [key]: {} };
+  values: Object.keys(PAYMENT_METHOD_TYPE).reduce((values, key) => {
+    return { ...values, [key]: { value: PAYMENT_METHOD_TYPE[key] } };
```
Betree, line 8: Too keep it retro-compatible, I would suggest:

```suggestion  values: {
      ...PAYMENT_METHOD_TYPES.reduce(
      (values, key) => ({ ...values, [key]: { deprecationReason: 'Please use uppercase values' } }),
      {},
    ),
    ...Object.keys(PAYMENT_METHOD_TYPE).reduce(
      (values, key) => ({ ...values, [key]: { value: PAYMENT_METHOD_TYPE[key] } }),
      {},
    ),
```

znarf: Thank you for the tip, trying this.

znarf:
@Betree E2E tests are still failing, looking at it.

server/constants/paymentMethods.ts
```diff
@@ -17,7 +17,7 @@ export enum PAYMENT_METHOD_TYPE {
   COLLECTIVE = 'collective',
   HOST = 'host',
   ADAPTIVE = 'adaptive',
-  GIFT_CARD = 'giftcard',
```
Betree, line 20: I would tend to the opposite direction: all entries should have this`_` (eg. `CREDIT_CARD`) as that's how things are formatted in our other enums (snake uppercase).

znarf: I'm favoring consistency with the database values for now.

I get that constants are effectively here to help having nicer/different values than the database one.

But as we already suffered too much on this topic, I think we need to relax a bit with a simple mental model.

PR #6449, opened 2021-08-16
HostSettlement: Show amount as decimal

Betree:
Fix https://github.com/opencollective/opencollective/issues/4525

image


znarf (approved):
I want to replace this with the new CSV format when it will be ready.

In the meantime, good to go!

PR #6451, opened 2021-08-16
childrenAccounts resolver followup

kewitz:
Related to https://github.com/opencollective/opencollective/issues/4383.
From https://github.com/opencollective/opencollective-api/pull/6440 late review.

### DIFF
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -5,7 +5,7 @@ import { assign, get, invert, isEmpty } from 'lodash';
 
 import { types as CollectiveTypes } from '../../../constants/collectives';
 import models, { Op } from '../../../models';
-import { NotFound, Unauthorized } from '../../errors';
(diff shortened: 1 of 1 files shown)


server/graphql/v2/interface/Account.js
```diff
@@ -570,15 +570,19 @@ const accountFieldsDefinition = () => ({
...
-      offset: { type: GraphQLInt, defaultValue: 0 },
+      limit: { type: new GraphQLNonNull(GraphQLInt), defaultValue: 100 },
+      offset: { type: new GraphQLNonNull(GraphQLInt), defaultValue: 0 },
       accountType: {
         type: new GraphQLList(AccountType),
       },
     },
     async resolve(account, args) {
+      if (args.limit > 100) {
+        throw new BadRequest('Max limit size is 100');
```
znarf, line 583: In some other places, we use more friendly messages such as:

```'Cannot fetch more than 100 orders at the same time, please adjust the limit```

Betree (approved):

PR #6453, opened 2021-08-17
PayPal: Handle duplicated webhook events

Betree:
Prevent double-recording of transactions when PayPal calls multiple times our webhook for the same event

### DIFF
--- a/server/paymentProviders/paypal/webhook.ts
+++ b/server/paymentProviders/paypal/webhook.ts
@@ -93,6 +93,19 @@ async function handleSaleCompleted(req: Request): Promise<void> {
 
   const { order } = await loadSubscriptionForWebhookEvent(req, subscriptionId);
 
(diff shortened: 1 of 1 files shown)


server/paymentProviders/paypal/webhook.ts
```diff
@@ -93,6 +93,19 @@ async function handleSaleCompleted(req: Request): Promise<void> {
 
   const { order } = await loadSubscriptionForWebhookEvent(req, subscriptionId);
 
+  // Make sure the sale hasn't already been recorded
+  const existingTransaction = await models.Transaction.findOne({
+    where: {
+      OrderId: order.id, // Not necessary, but makes the query faster
```
kewitz, line 99: I'm all in for this type of optimization. 👍🏼

kewitz (approved):

PR #6454, opened 2021-08-17
Add support for TRANSACTION_CONVERTED event for the Giving Block

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4527

### DIFF
--- a/server/paymentProviders/thegivingblock/webhook.js
+++ b/server/paymentProviders/thegivingblock/webhook.js
@@ -15,18 +15,7 @@ export async function webhook(req) {
   const payload = JSON.parse(payloadString);
   logger.info(`payload: ${JSON.stringify(payload)}`);
 
-  if (req.body.eventType === 'DEPOSIT_TRANSACTION') {
-    const pledgeId = payload.pledgeId;
-    const valueAtDonationTimeUSD = payload.valueAtDonationTimeUSD;
-
-    const order = await models.Order.findOne({ where: { data: { pledgeId } } });
-    if (!order) {
-      throw new Error(`Could not find matching order. pledgeId=${pledgeId}`);
-    }
-
-    // update totalAmount with latest value
-    await order.update({ totalAmount: Number(valueAtDonationTimeUSD) * 100, currency: 'USD' });
-  } else if (req.body.eventType === 'TRANSACTION_CONVERTED') {
+  if (req.body.eventType === 'TRANSACTION_CONVERTED') {
     const pledgeId = payload.pledgeId;
     const netValueAmount = payload.netValueAmount;
 
@@ -36,9 +25,7 @@ export async function webhook(req) {
     }
 
     // update totalAmount with latest value
-    await order.update({ totalAmount: Number(netValueAmount) * 100, currency: 'USD' });
-
-    await order.update({ status: OrderStatus.PAID });
+    await order.update({ totalAmount: Number(netValueAmount) * 100, currency: 'USD', status: OrderStatus.PAID });
 
     // process as paid
     const transaction = await confirmOrder(order);
--- a/server/paymentProviders/thegivingblock/webhook.js
+++ b/server/paymentProviders/thegivingblock/webhook.js
@@ -1,4 +1,6 @@
+import OrderStatus from '../../../server/constants/order_status';
 import logger from '../../lib/logger';
+import { sendThankYouEmail } from '../../lib/recurring-contributions';
 import models from '../../models';
 
 import { confirmOrder, decryptPayload } from './index';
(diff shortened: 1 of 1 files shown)


server/paymentProviders/thegivingblock/webhook.js
```diff
@@ -17,18 +19,36 @@ export async function webhook(req) {
...
     }
 
     // update totalAmount with latest value
-    order = await order.update({ totalAmount: Number(valueAtDonationTimeUSD) * 100, currency: 'USD' });
+    await order.update({ totalAmount: Number(valueAtDonationTimeUSD) * 100, currency: 'USD' });
+
+    return;
+  }
+
+  if (req.body.eventType === 'TRANSACTION_CONVERTED') {
```
kewitz, line 33:
```suggestion
  else if (req.body.eventType === 'TRANSACTION_CONVERTED') {
```

server/paymentProviders/thegivingblock/webhook.js
```diff
@@ -17,18 +19,36 @@ export async function webhook(req) {
...
+    const order = await models.Order.findOne({ where: { data: { pledgeId } } });
     if (!order) {
       throw new Error(`Could not find matching order. pledgeId=${pledgeId}`);
     }
 
     // update totalAmount with latest value
-    order = await order.update({ totalAmount: Number(valueAtDonationTimeUSD) * 100, currency: 'USD' });
+    await order.update({ totalAmount: Number(valueAtDonationTimeUSD) * 100, currency: 'USD' });
+
+    return;
```
kewitz, line 30: You can drop the early `return` if all logic is contained within eventType conditionals.
You should also drop the "Event not supported" error, it is bad practice since this usually causes the service provider to keep retrying when an error is throwed.

server/paymentProviders/thegivingblock/webhook.js
```diff
@@ -17,18 +19,36 @@ export async function webhook(req) {
...
 
     // process as paid
-    await confirmOrder(order);
+    const transaction = await confirmOrder(order);
 
-    order = await order.update({ status: 'PAID' });
+    // send email confirmation
+    await sendThankYouEmail(order, transaction);
 
     return;
```
kewitz, line 53: You should also drop this.

server/paymentProviders/thegivingblock/webhook.js
SudharakaP: Done. 👍🏽

server/paymentProviders/thegivingblock/webhook.js
SudharakaP: Done. 👍🏽

server/paymentProviders/thegivingblock/webhook.js
SudharakaP: Ah, thanks for letting me know. Makes sense. I've remove this error. 👍🏽 💯

server/paymentProviders/thegivingblock/webhook.js
```diff
@@ -17,23 +19,33 @@ export async function webhook(req) {
...
+      throw new Error(`Could not find matching order. pledgeId=${pledgeId}`);
+    }
 
-    order = await order.update({ status: 'PAID' });
+    // update totalAmount with latest value
+    await order.update({ totalAmount: Number(netValueAmount) * 100, currency: 'USD' });
 
-    return;
-  }
+    await order.update({ status: OrderStatus.PAID });
```
znarf, line 41: you can group updates together

server/paymentProviders/thegivingblock/webhook.js
```diff
@@ -17,23 +19,33 @@ export async function webhook(req) {
...
 
-    let order = await models.Order.findOne({ where: { data: { pledgeId } } });
+    const order = await models.Order.findOne({ where: { data: { pledgeId } } });
     if (!order) {
       throw new Error(`Could not find matching order. pledgeId=${pledgeId}`);
     }
 
     // update totalAmount with latest value
-    order = await order.update({ totalAmount: Number(valueAtDonationTimeUSD) * 100, currency: 'USD' });
+    await order.update({ totalAmount: Number(valueAtDonationTimeUSD) * 100, currency: 'USD' });
```
znarf, line 28: You can drop the whole block.

server/paymentProviders/thegivingblock/webhook.js
SudharakaP: Ops, totally missed that one. Corrected it. 👍🏽

server/paymentProviders/thegivingblock/webhook.js
SudharakaP: I've dropped this. I was wondering if we should still use the `DEPOSIT_TRANSACTION` and maybe update the order to store that fact. I think this transaction acknowledges the fact the the deposit was done properly and we are waiting for the Giving Block to convert it. When we don't use this until the `TRANSACTION_CONVERTED` is received we might not have any guarantee that the giving block is working on the conversion (or if they have received it). 🤔

znarf (approved):

server/paymentProviders/thegivingblock/webhook.js
znarf: If we have something useful to do we can do it, perhaps updating order.data with the result. In another PR perhaps?

SudharakaP: One thing I was thinking is perhaps we can show the contributor that the transaction is under processing by the giving block (with some sort of tag or something in the UI). This way they have a bit more information than the `PENDING` status. But maybe we'll keep this in mind for the future as I don't know how useful a feature that is in the initial iteration. 🤔

PR #6455, opened 2021-08-17
PayPal: Add webhook to handle remote refunds

Betree:
Part of opencollective/opencollective#4368
Also includes first draft for script to check unsynchronized refunds (#6367)


server/paymentProviders/paypal/payment.js
```diff
@@ -186,12 +186,21 @@ export const refundPaypalCapture = async (transaction, captureId, user, reason)
...
   }
 
-  // eslint-disable-next-line camelcase
-  const payload = { note_to_payer: truncate(reason, { length: 255 }) || undefined };
-  const result = await paypalRequestV2(`payments/captures/${captureId}/refund`, host, 'POST', payload);
-  const rawRefundedPaypalFee = get(result, 'seller_payable_breakdown.paypal_fee.amount.value', '0.00');
-  const refundedPaypalFee = floatAmountToCents(parseFloat(rawRefundedPaypalFee));
-  return createRefundTransaction(transaction, refundedPaypalFee, { paypalResponse: result }, user);
+  // Add a flag on transaction to prevent multiple calls from the webhook
+  await transaction.update({ data: { ...transaction.data, isRefundedFromOurSystem: true } });
```
kewitz, line 190: This is smart, we already had echoes from Paypal before.
If the early return lives in the `webhook.ts:handleCaptureRefunded` I don't really see the benefit of having this lock/try/catch logic here. I think it would be simpler if we had the whole lock logic inside `handleCaptureRefunded`.

scripts/paypal/payment-reconciliator.ts
```diff
@@ -100,47 +102,85 @@ const markOrderAsError = async order => {
...
+  const response = await paypalRequest(apiUrl, null, host, 'GET');
+  // TODO: Handle pagination
+  return response['transaction_details'];
+};
 
-    if (fix && !isValid && !parseToBoolean(process.env.DRY)) {
-      await markOrderAsError(orderItem.value);
+const findRefundedContributions = async (hostSlugs: string[]) => {
+  for (const hostSlug of hostSlugs) {
+    console.log(`\nChecking host ${hostSlug} for refunded contributions not marked as such in our ledger...`);
```
kewitz, line 148: Aren't we just listing refunded transactions?

scripts/paypal/payment-reconciliator.ts
```diff
@@ -100,47 +102,85 @@ const markOrderAsError = async order => {
...
+  urlParams.append('fields', 'all');
+  urlParams.append('page_size', '100');
+  urlParams.append('page', '1');
+  urlParams.append('transaction_status', 'V');
+  urlParams.append('start_date', startDate.toISOString());
+  urlParams.append('end_date', endDate.toISOString());
+  const apiUrl = `reporting/transactions?${urlParams.toString()}`;
+  const response = await paypalRequest(apiUrl, null, host, 'GET');
+  // TODO: Handle pagination
+  return response['transaction_details'];
```
kewitz, line 143: We could just `console.warn` that the user should use a smaller time frame when the length of the transactions is 100.

server/routes.js
```diff
@@ -217,7 +217,7 @@ export default async app => {
   app.post('/webhooks/stripe', stripeWebhook); // when it gets a new subscription invoice
   app.post('/webhooks/transferwise', transferwiseWebhook); // when it gets a new subscription invoice
   app.post('/webhooks/privacy', privacyWebhook); // when it gets a new subscription invoice
-  app.post('/webhooks/paypal', paypalWebhook);
+  app.post('/webhooks/paypal/:hostId?', paypalWebhook);
```
kewitz, line 220: Nice solution.

kewitz (approved):
LGTM. Added a small suggestion for the locking logic.

scripts/paypal/payment-reconciliator.ts
Betree: That's a good option. For this first MVP, I expect the number of entries to be quite low. I'm still not done with this script and will need to complete it in another PR

scripts/paypal/payment-reconciliator.ts
Betree: For now, yes. Sorry, I should have indicated that this script was not complete in the PR description

server/routes.js
Betree: It was necessary for the refund part because the refund payload doesn't have any info about which transaction/order/host we should look at. But ultimately, this can help us optimize other endpoints as we'll only need to look at the orders/transactions that are related to that host

server/paymentProviders/paypal/payment.js
Betree: The main idea here is to ignore the webhook event if the transaction is refunded from Open Collective's interface (because the `refundTransaction` mutation already takes care of everything). It's not intended to prevent PayPal echoes, that part is taken care of in the webhook itself  (by making sure `RefundTransactionId: null`).

But I realize that the comment is ambiguous, I'm going to push an update for it

PR #6457, opened 2021-08-18
Prevent users from inviting existing members

kewitz:


Betree (approved):
Nice check

server/graphql/v2/mutation/MemberInvitationMutations.js
```diff
@@ -58,6 +58,17 @@ const memberInvitationMutations = {
...
+        where: {
+          role: args.role,
+          CollectiveId: account.id,
+          MemberCollectiveId: memberAccount.id,
+        },
+      });
+      if (existingMembership) {
+        throw new BadRequest('Invited user is already a member of the account');
+      }
+
```
znarf, line 71: Could you move this in the `MemberInvitation.invite` method?

znarf (changes requested):
Change requested.

znarf:
What issue is this fixing?

server/graphql/v2/mutation/MemberInvitationMutations.js
snitin315: I believe this check is already present https://github.com/opencollective/opencollective-api/blob/b9aeb1685c08290987a6a67e8e5fe7801d96e261/server/models/MemberInvitation.js#L159-L180

kewitz: Good catch.
There's no issue attached to this, I caught this while demoing last Friday and added to my to-do list.

kewitz:
Sorry for wasting your time. I believe what happened is that we had multiple users with the same name on staging when I tested during demos.

PR #6460, opened 2021-08-18
Add new `BALANCE_TRANSFER` kind to Transactions for Internal Transfers (Empty Balance)

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4489

Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/6761

### DIFF
--- a/migrations/dev-20210818-add-balance-transfer-transaction-kind.js
+++ b/migrations/dev-20210818-add-balance-transfer-transaction-kind.js
@@ -1,0 +1,38 @@
+'use strict';
+
+module.exports = {
+  up: async queryInterface => {
+    await queryInterface.sequelize.query(`
+      ALTER TYPE "enum_Transactions_kind"
+      ADD VALUE IF NOT EXISTS 'BALANCE_TRANSFER' AFTER 'PREPAID_PAYMENT_METHOD'
+    `);
+  },
+
+  down: async queryInterface => {
+    await queryInterface.sequelize.query(`
+      UPDATE "Transactions"
+      SET "kind" = 'CONTRIBUTION'
+      WHERE "kind" = 'BALANCE_TRANSFER'
+    `);
+
+    await queryInterface.sequelize.query(`
+      ALTER TYPE "enum_Transactions_kind" RENAME TO "enum_Transactions_kind_old";
+    `);
+
+    await queryInterface.sequelize.query(`
+      CREATE TYPE "enum_Transactions_kind" as enum ('ADDED_FUNDS', 'CONTRIBUTION', 'EXPENSE', 'HOST_FEE', 'HOST_FEE_SHARE', 'HOST_FEE_SHARE_DEBT', 'PAYMENT_PROCESSOR_FEE', 'PLATFORM_FEE', 'PLATFORM_TIP', 'PLATFORM_TIP_DEBT', 'PREPAID_PAYMENT_METHOD')
+    `);
+
+    await queryInterface.sequelize.query(`
+      ALTER TABLE "Transactions" ALTER COLUMN "kind" TYPE "enum_Transactions_kind" USING "kind"::text::"enum_Transactions_kind"
+    `);
+
+    await queryInterface.sequelize.query(`
+      ALTER TABLE "TransactionSettlements" ALTER COLUMN "kind" TYPE "enum_Transactions_kind" USING "kind"::text::"enum_Transactions_kind"
+    `);
+
+    await queryInterface.sequelize.query(`
+      DROP TYPE "enum_Transactions_kind_old"
+    `);
+  },
+};
--- a/migrations/dev-20210818-add-balance-transfer-transaction-kind.js
+++ b/migrations/dev-20210818-add-balance-transfer-transaction-kind.js
@@ -4,7 +4,7 @@ module.exports = {
   up: async queryInterface => {
     await queryInterface.sequelize.query(`
       ALTER TYPE "enum_Transactions_kind"
-      ADD VALUE IF NOT EXISTS 'BALANCE_TRANSFER' AFTER 'PREPAID_PAYMENT_METHOD'
+      ADD VALUE IF NOT EXISTS 'BALANCE_TRANSFER' AFTER 'ADDED_FUNDS'
     `);
   },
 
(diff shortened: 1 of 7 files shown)


server/paymentProviders/opencollective/collective.js
```diff
@@ -97,9 +97,7 @@ paymentMethodProvider.processOrder = async order => {
     },
   };
 
-  const transactions = await models.Transaction.createFromContributionPayload(transactionPayload);
-
-  return transactions;
+  return await models.Transaction.createFromContributionPayload(transactionPayload);
```
znarf, line 100:
```suggestion
  return models.Transaction.createFromContributionPayload(transactionPayload);
```

server/paymentProviders/opencollective/collective.js
```diff
@@ -76,7 +76,7 @@ paymentMethodProvider.processOrder = async order => {
     CollectiveId: order.CollectiveId,
     PaymentMethodId: order.PaymentMethodId,
     type: TransactionTypes.CREDIT,
-    kind: TransactionKind.CONTRIBUTION,
+    kind: order.data?.customData?.isBalanceTransfer ? TransactionKind.BALANCE_TRANSFER : TransactionKind.CONTRIBUTION,
```
znarf, line 79: Maybe it's time to introduce an `Order.kind`. The possible values would be `ADDED_FUNDS`, `CONTRIBUTION` and `BALANCE_TRANSFER`.

What do you think @Betree ?

server/constants/transaction-kind.ts
```diff
@@ -1,6 +1,8 @@
 export enum TransactionKind {
   /** Transactions coming from the "Add Funds" */
   ADDED_FUNDS = 'ADDED_FUNDS',
+  /** Transactions from transferring the remaining balance from a project/event **/
```
znarf, line 4: It's also for the balance of a collective to a fiscal host.

server/paymentProviders/opencollective/collective.js
Betree: That would be cleaner than a flag in `data`. I'm wondering though, do we really need an order for added funds and balance transfers? What would be the downsides of creating the transactions directly, without an order?

migrations/dev-20210818-add-balance-transfer-transaction-kind.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+module.exports = {
+  up: async queryInterface => {
+    await queryInterface.sequelize.query(`
+      ALTER TYPE "enum_Transactions_kind"
+      ADD VALUE IF NOT EXISTS 'BALANCE_TRANSFER' AFTER 'PREPAID_PAYMENT_METHOD'
```
Betree, line 7: Why not this, to keep it in alphabetical order?

```suggestion
      ADD VALUE IF NOT EXISTS 'BALANCE_TRANSFER' AFTER 'ADDED_FUNDS'
```

server/paymentProviders/opencollective/collective.js
znarf: I don't know, we're using Orders because this is the way it is until now.

Because of this, we can use the GraphQL `createOrder` mutation. If we're reconsidering this, we would have to create a separate mutation for `BALANCE_TRANSFER` (we already have one for ADDED_FUNDS).

I actually appreciate having an Order or an Expense for most Transactions. But looking at it more closely, I don't see a real value today.

znarf: Also, having an order seems required to use the complex PaymentMethod code which create the transactions. Maybe it's best to keep Order for now, out of simplicity.

migrations/dev-20210818-add-balance-transfer-transaction-kind.js
SudharakaP: Ops, yes that makes sense. Done. 👍🏽

server/constants/transaction-kind.ts
SudharakaP: Added `collective` to the comments as well. Thanks. 👍🏽

server/paymentProviders/opencollective/collective.js
SudharakaP: So I think for now we will keep this as it is. But note that I've removed the `isBalanceTransfer` field from `customData` as suggested at https://github.com/opencollective/opencollective-frontend/pull/6761#discussion_r691861458

server/graphql/v2/input/OrderCreateInput.ts
```diff
@@ -72,5 +72,9 @@ export const OrderCreateInput = new GraphQLInputObjectType({
       type: OrderContextInput,
       description: 'Some context about how this order was created',
     },
+    isBalanceTransfer: {
+      type: GraphQLBoolean,
+      description: 'Whether this is transferring the remaining balance from a project/event',
```
znarf, line 77: Or from a collective to its host.

SudharakaP: Thanks for pointing it out. Corrected that. 👍🏽

Betree (approved):
Looking good!

znarf (commented):
Missing is a migration of past entries. Could this be done in a separate PR?

server/graphql/v2/input/OrderCreateInput.ts
```diff
@@ -1,7 +1,7 @@
 import { GraphQLBoolean, GraphQLInputObjectType, GraphQLInt, GraphQLList, GraphQLNonNull } from 'graphql';
 import GraphQLJSON from 'graphql-type-json';
 
-import { ContributionFrequency } from '../enum/ContributionFrequency';
+import { ContributionFrequency } from '../enum';
```
znarf, line 4: What's the point?

SudharakaP: Well, I guess not much, but my IDE gives me these warnings that the import could be shortened and I think I just clicked on it. 🙂

SudharakaP:
> Missing is a migration of past entries. Could this be done in a separate PR?

@znarf: I am wondering whether this is possible? 🤔 Because I don't think prior to this we differentiate between a balance transfer vs a real contribution from the collective to the host for example, or from the child to the parent collective. But I think from child (event/project) to parent collective transfers we could consider all of them as balance transfers right? But for collective to host I am not sure if this is the case as one could also use the contribution flow. In that case the balance transfer button isn't clicked at all and the intention of the person is actually to "contribute" to the host. Correct me if I am wrong on this? 🤔

znarf:
> I am wondering whether this is possible? 

@SudharakaP ok then.

PR #6461, opened 2021-08-18
Contributions Pagination: Follow-up

kewitz:


server/graphql/v2/interface/IsMemberOf.js
```diff
@@ -48,8 +49,7 @@ export const IsMemberOfFields = {
         description: 'Filters on the Host fees structure applied to this account',
       },
       orderBy: {
-        type: new GraphQLNonNull(ChronologicalOrderInput),
-        defaultValue: { field: 'createdAt', direction: 'ASC' },
+        type: ChronologicalOrderInput,
```
Betree, line 52: It's good practice to force having a predictable order. I don't get the reason for this change?

server/graphql/v2/interface/IsMemberOf.js
```diff
@@ -151,6 +161,36 @@ export const IsMemberOfFields = {
             as: 'collective',
             where: collectiveConditions,
             required: true,
+            attributes: {
```
Betree, line 164: I don't get this part, where are the fields being used?

server/graphql/v2/interface/IsMemberOf.js
```diff
@@ -128,15 +136,17 @@ export const IsMemberOfFields = {
...
 
-      if (!isNil(args.isApproved)) {
-        collectiveConditions.approvedAt = { [args.isApproved ? Op.not : Op.is]: null };
-      }
-      if (!isNil(args.isArchived)) {
-        collectiveConditions.deactivatedAt = { [args.isArchived ? Op.not : Op.is]: null };
+      const order = [
+        [sequelize.literal('"collective.memberCount"'), 'DESC'],
+        [sequelize.literal('"collective.totalAmountDonated"'), 'DESC'],
+      ];
```
Betree, line 142: This should be an option of the `orderBy` argument. Or at least, we shouldn't fetch this heavy data when `orderBy` is referring to creation date.

server/graphql/v2/interface/IsMemberOf.js
```diff
@@ -151,6 +161,36 @@ export const IsMemberOfFields = {
...
+            attributes: {
+              include: [
+                [
+                  sequelize.literal(`(
+                    SELECT COUNT(*)
+                    FROM "Members" AS "collective->members"
+                    WHERE
+                        "collective->members"."CollectiveId" = collective.id
+                        AND
+                        "collective->members".role = 'BACKER'
```
Betree, line 173: Deleted members should be ignored

server/graphql/v2/interface/IsMemberOf.js
```diff
@@ -151,6 +161,36 @@ export const IsMemberOfFields = {
             as: 'collective',
             where: collectiveConditions,
             required: true,
+            attributes: {
+              include: [
+                [
+                  sequelize.literal(`(
+                    SELECT COUNT(*)
+                    FROM "Members" AS "collective->members"
```
Betree, line 169: I'm a bit afraid of the performance impact of summing members and transactions in these subqueries. Could we measure the complexity for `opensource`, `opencollective` and a smaller host to see how it performs?

server/graphql/v2/interface/IsMemberOf.js
kewitz: I implemented the default ordering in the resolver itself.
The reason behind this is that the default ordering we had before (number of contributors and amount contributed) is much more complex.

server/graphql/v2/interface/IsMemberOf.js
kewitz: Should we use pseudo ordering arguments? (`memberCount` and `totalAmoundDonated`)

Betree: Yes, it's not the easiest approach but I think that's what makes the most sense from a consumer point of view and gives flexibility for the future:

```gql
memberOf(orderBy: [MEMBER_COUNT, ASC])
```

server/graphql/v2/interface/IsMemberOf.js
Betree: Also sharing something I've been discussing with @znarf, I think we all see the need for something like that: for this kind of use case, it would be great to have a `popularity` score somewhere (based on the contributor's count, and eventually the amount of money received) that would be re-computed by a CRON job every hour.

That would also be a valuable data to improve search.

kewitz:
Rebased.

znarf (commented):
What about performance benchmark? This can't go live if performance is bad.

kewitz:
I have some logs related to the GraphQL query running locally. In practice, there's no notable difference because this component loads async after the page is loaded.

Improved ordering:
```
ContributionsSection {"slug":"opensource"} 1274ms
ContributionsSection {"slug":"opensource"} 1180ms
ContributionsSection {"slug":"opensource"} 1045ms
ContributionsSection {"slug":"opensource"} 1543ms
```

Existing ordering:
```
ContributionsSection {"slug":"opensource"} 1462ms
ContributionsSection {"slug":"opensource"} 1512ms
ContributionsSection {"slug":"opensource"} 1259ms
ContributionsSection {"slug":"opensource"} 1198ms
```

znarf:
@kewitz I'm a bit surprised by the results. Is there a Frontend PR that matches this so I can try with production data?

kewitz:
@znarf, yes, it has the same branch name.
I tested fully local but this will for sure depend on the production environment DB caching.

PR #6266, opened 2021-07-12
Update `ics` package to latest version: v2.29.0

SudharakaP:
This updates the `ics` package to the latest version. It seems that the `ics.createEvent` function now takes dates as arrays instead of `moment` objects. 🤔 

Related to https://github.com/opencollective/opencollective-api/pull/6212

PR #6462, opened 2021-08-19
Correct the Month in GCal Invites

SudharakaP:
Related to https://github.com/opencollective/opencollective-api/pull/6266

This was a mistake introduced in the above PR. The month here is zero indexed so we need to compensate by adding 1 to it. 😃

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -766,14 +766,14 @@ function defineModel() {
         const endDate = new Date(this.endsAt);
         const start = [
           startDate.getFullYear(),
-          startDate.getMonth(),
+          startDate.getMonth() + 1,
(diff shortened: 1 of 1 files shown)


znarf (commented):
Could you double check that timezone handling is how it should be?

SudharakaP:
> Could you double check that timezone handling is how it should be?

I did some checking on the event timezone vs the gcal time-zones and they seem to be in sync and correct. So I don't see anything wrong with the time-zones. 🤔

znarf (approved):
Cool, let's ship this asap.

SudharakaP:
> Cool, let's ship this asap.

Thanks for the review. Will do a deployment soon. 👍🏽

PR #6473, opened 2021-08-20
Correct event members registered as BACKERS to ATTENDEES

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4417


znarf (approved):
Looks good to me. @Betree should review too.

scripts/change-event-members-from-backers-to-attendees.js
```diff
@@ -0,0 +1,27 @@
...
+    {
+      type: sequelize.QueryTypes.SELECT,
+      model: models.Member,
+      mapToModel: true,
+    },
+  );
+
+  for (const member of members) {
+    await member.update({ role: MemberRoles.ATTENDEE });
+  }
```
Betree, line 22: This is quite heavy: we're going to load 3688 members then trigger 3688 update queries, while everything could be done in one single SQL statement. Not a requested change as the API can probably handle it, but in the future maybe try to do something like:

```sql
UPDATE "Members"
SET role = ...
FROM "Tiers", "Collectives"
WHERE ...
```

server/models/Order.js
```diff
@@ -292,7 +293,7 @@ function defineModel() {
     // Register user as collective backer
     const member = await this.collective.findOrAddUserWithRole(
       { id: this.CreatedByUserId, CollectiveId: this.FromCollectiveId },
-      roles.BACKER,
+      this.tier?.type === tiers.TICKET ? roles.ATTENDEE : roles.BACKER,
```
Betree, line 296: I'm not sure `tier` will always be loaded when calling this helper, like here for example: https://github.com/opencollective/opencollective-api/blob/488eec825a7cd78a59b8c25708c078ed73c42a96/server/paymentProviders/paypal/webhook.ts#L127

Betree (commented):
Why a script rather than a migration? This sounds like something we would like to fix in all environments (dev included).

scripts/change-event-members-from-backers-to-attendees.js
SudharakaP: @Betree : Yes, makes sense to create a migration script than a regular script. 👍🏽 Thanks for pointing it out. I've corrected that along with the SQL to use an update query.

server/models/Order.js
SudharakaP: @Betree : I've changed the code to obtain the tier information based on the `TierID` instead rather than relying on the `tier` object. I think this makes more sense. Let me know. 🤔

SudharakaP:
@Betree : Also want to point out that we need to align an e2e test in the front-end after running the migration as I think one of the `BACKER` entries are changed to `ATTENDEE`. Done this at, https://github.com/opencollective/opencollective-frontend/pull/6773

migrations/dev-20210823-change-event-members-from-backers-to-attendees.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+module.exports = {
+  up: async queryInterface => {
+    await queryInterface.sequelize.query(
+      `UPDATE "Members" SET "role" = 'ATTENDEE' FROM "Tiers" t, "Collectives" c WHERE c.type = 'EVENT' AND role != 'ATTENDEE' AND t.type = 'TICKET' AND "TierId" = t.id AND t."CollectiveId" = c.id`,
```
Betree, line 6: Some formatting could make this more readable, but the query itself seems good :+1:

server/models/Order.js
```diff
@@ -288,11 +289,11 @@ function defineModel() {
   Order.prototype.getOrCreateMembers = async function () {
     // Preload data
     this.collective = this.collective || (await this.getCollective());
-
-    // Register user as collective backer
+    const tier = await models.Tier.findOne({ where: { id: this.TierId } });
```
Betree, line 292: :no_entry_sign: `this.TierId` can be null

Betree (changes requested):

migrations/dev-20210823-change-event-members-from-backers-to-attendees.js
SudharakaP: Thanks. Corrected it. 👍🏽

server/models/Order.js
SudharakaP: @Betree : I've added a null check for this case so that if the `TierID` is null we don't take that into account. Let me know if you see any issues with this approach. 🤔

PR #6480, opened 2021-08-24
Finalized merge accounts

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4567

Add all related models to the merge accounts util, to make sure we don't miss any data.

The goal of the `MigrationLogs` table is to be re-used whenever we need to store information about migration results. For example, we could imagine doing something like this in a sequelize migration:

```es6
await models.MigrationLog.create({
  type: 'MIGRATION',
  name: '20210615074327-platform-tip-debt-kind',
  data: {
    updatedTransactionsIds: [45,88,77]
  }
})
```

The real stuff happens in https://github.com/opencollective/opencollective-api/pull/6480/files#diff-5b651c29bc4624a4f05eb6a0ab745a233063e46ab263aca619d3cbeaf87a78ed.


server/lib/collectivelib.ts
```diff
@@ -484,7 +484,35 @@ export const mergeCollectives = async (
...
+          giftCardTransactionsIds,
+          paymentMethodsIds,
+          membersIds,
+          membershipsIds,
+          memberInvitationIds,
+          membershipInvitationsIds,
+          activitiesIds,
+        },
       },
+      { transaction },
```
Betree, line 515: This is the part where we log the result

server/lib/merge-accounts.ts
```diff
@@ -0,0 +1,377 @@
...
+};
+
+// Defines the collective field names used in the DB. Useful to prevent typos in the config below
+type CollectiveField =
+  | 'CollectiveId'
+  | 'HostCollectiveId'
+  | 'ParentCollectiveId'
+  | 'FromCollectiveId'
+  | 'UsingGiftCardFromCollectiveId'
+  | 'MemberCollectiveId';
```
znarf, line 114: Nice.

scripts/merge-collectives.js
```diff
@@ -59,7 +59,7 @@ async function main() {
       `\nMerging ${fromCollective.slug} (#${fromCollective.id}) into ${intoCollective.slug} (#${intoCollective.id})...`,
     );
     await sleep(3000); // Wait some time so user can CTRL+C if there's a mistake
```
znarf, line 61: Nice!

znarf (approved):
Looks great! Let me know if you want to talk once last time before merging.

Betree:
@znarf Thanks for the review, I think it's good to go live for some real-life testing!

PR #6485, opened 2021-08-24
Handle order creation for adding funds to past events

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/3982

Related Front-End PR: https://github.com/opencollective/opencollective-frontend/pull/6772


znarf (changes requested):
That should not be needed, we can already pass the right Payment Method from the frontend I think.

PR #6486, opened 2021-08-25
feat: Add captcha to guest contributions

kewitz:
Related to https://github.com/opencollective/opencollective/issues/4329

Before merging:
- [x] Create a shared hCaptcha account
- [x] Set up hCaptcha on staging
- [x] Set up hCaptcha on production

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "graphql-type-json": "0.3.2",
     "handlebars": "4.7.7",
     "hashids": "2.2.8",
+    "hcaptcha": "0.1.0",
     "helloworks-sdk": "1.3.0",
     "helmet": "4.6.0",
     "html-pdf": "3.0.1",
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -169,8 +169,7 @@ const checkGuestContribution = async (order, loaders) => {
 
 async function checkCaptcha(order, remoteUser, reqIp) {
   const requestedProvider = order.guestInfo?.captcha?.provider;
-  const isCaptchaEnabled =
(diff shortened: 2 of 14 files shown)


server/graphql/v1/mutations/orders.js
```diff
@@ -164,20 +167,41 @@ const checkGuestContribution = async (order, loaders) => {
...
+    config.recaptcha &&
+    parseToBoolean(config.recaptcha.enable)
+  ) {
+    response = await recaptcha.verify(order.guestInfo.captcha.token, reqIp);
+  } else {
+    throw new BadRequest('Could not find requested Captcha provider', undefined, order.guestInfo?.captcha);
+  }
 
-  // TODO: check response and throw an error if needed
+  console.log(response);
```
SudharakaP, line 201:
```suggestion
```

package.json
```diff
@@ -79,6 +79,7 @@
     "graphql-type-json": "0.3.2",
     "handlebars": "4.7.7",
     "hashids": "2.2.8",
+    "hcaptcha": "^0.1.0",
```
SudharakaP, line 82:
```suggestion
    "hcaptcha": "0.1.0",
```

package.json
```diff
@@ -233,7 +234,7 @@
   },
   "engines": {
     "node": "14.x",
-    "npm": "6.x"
+    "npm": ">=6.x"
```
SudharakaP, line 237: Is this change related to this? 🤔 Also note that changing this might be a bit problematic since this will probably allow users to use npm v7 which will change the `package-lock.json` as we discussed on Slack; https://opencollective.slack.com/archives/C0RMV6F8C/p1629743374124200

SudharakaP (changes requested):
Some minor comments. 😄

PR #6487, opened 2021-08-25
GQLV2: Deprecate `Account.accountMembers`

Betree:
This should have been implemented with the existing `members` field. Also, the arguments don't match our GQLV2 conventions (`TierId`, `tierSlug`).


snitin315 (approved):
Still wrapping my brain around GQLV2.

znarf (approved):
Simple first step.

PR #6489, opened 2021-08-25
Member Invitations: Improve validations

Betree:
- Add a maximum limit on the number of members an account can have. The previous limit included only invitations, while this one is a sum of existing members + invitations.
- Prevent adding members on an Individual account

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -67,7 +67,7 @@
     "searchEmailPerHour": 150,
     "searchEmailPerHourPerIp": 50,
     "collectiveEmailMessagePerHour": 5,
-    "maxMemberInvitationsPerCollective": 50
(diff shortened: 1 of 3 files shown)


config/default.json
```diff
@@ -67,7 +67,7 @@
     "searchEmailPerHour": 150,
     "searchEmailPerHourPerIp": 50,
     "collectiveEmailMessagePerHour": 5,
-    "maxMemberInvitationsPerCollective": 50
+    "maxCoreContributorsPerAccount": 60
```
Betree, line 70: webpack-webinar has 51 admins, this value has been picked to not break anything

znarf (approved):

PR #6498, opened 2021-08-27
Expose the existing kinds matching args on Transactions query

znarf:
Part of https://github.com/opencollective/opencollective/issues/4590

### DIFF
--- a/server/graphql/v2/interface/Collection.ts
+++ b/server/graphql/v2/interface/Collection.ts
@@ -52,6 +52,7 @@ export interface CollectionReturnType {
   totalCount: number;
   limit: number;
   offset: number;
+  kinds?: Array<string>;
 }
 
 export { Collection, CollectionFields, CollectionArgs };
--- a/server/graphql/v2/interface/Collection.ts
+++ b/server/graphql/v2/interface/Collection.ts
@@ -52,6 +52,13 @@ export interface CollectionReturnType {
   totalCount: number;
   limit: number;
   offset: number;
+}
+
+export interface TransactionsCollectionReturnType {
+  nodes: Record<string, unknown>[];
+  totalCount: number;
+  limit: number;
+  offset: number;
   kinds?: Array<string>;
 }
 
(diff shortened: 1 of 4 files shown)


server/graphql/v2/interface/IsMemberOf.js
```diff
@@ -68,26 +68,28 @@ export const IsMemberOfFields = {
...
-          ],
-          group: ['role', 'collective.type'],
-          raw: true,
-        })
-      ).map(m => ({
-        role: m.role,
-        type: invert(AccountTypeToModelMapping)[m.type],
-      }));
+      // No await needed, GraphQL will take care of it
+      // TODO: try to skip if it's not a requested field
```
kewitz, line 72: 👍🏼

server/graphql/v2/interface/Collection.ts
```diff
@@ -54,4 +54,12 @@ export interface CollectionReturnType {
   offset: number;
 }
 
+export interface TransactionsCollectionReturnType {
```
kewitz, line 57: Extend `CollectionReturnType` with `kinds`.

kewitz (commented):
💯

server/graphql/v2/interface/Collection.ts
znarf: @kewitz What would be the syntax?

server/graphql/v2/query/TransactionsQuery.ts
```diff
@@ -152,6 +152,16 @@ const TransactionsQuery = {
     if (host) {
       where.push({ HostCollectiveId: host.id });
     }
+
+    // No await needed, GraphQL will take care of it
+    // TODO: try to skip if it's not a requested field
+    const existingKinds = models.Transaction.findAll({
+      attributes: ['kind'],
+      where,
```
kewitz, line 160: Add where `kind IS NOT NULL`.
Screenshot from 2021-09-06 11-11-04

Betree: Or a proper label in the frontend (eg. `Unknown`) if we still want the ability to filter them

PR #6510, opened 2021-08-30
Rename PAYMENT_PROCESSOR _FEE to _COVER

znarf:
Fix: https://github.com/opencollective/opencollective/issues/4544


Betree (approved):
Looks good!

PR #6511, opened 2021-08-30
Add Support to Show Warning for Past Events in Contribution Flow

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/3982

Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/6797

### DIFF
--- a/server/graphql/v2/object/Event.js
+++ b/server/graphql/v2/object/Event.js
@@ -52,6 +52,10 @@ export const Event = new GraphQLObjectType({
           }
         },
       },
+      startsAt: {
+        description: 'The Event start date and time',
+        type: GraphQLDateTime,
+      },
       endsAt: {
         description: 'The Event end date and time',
         type: GraphQLDateTime,
--- a/server/graphql/v2/object/Event.js
+++ b/server/graphql/v2/object/Event.js
@@ -1,4 +1,5 @@
 import { GraphQLBoolean, GraphQLNonNull, GraphQLObjectType } from 'graphql';
+import { GraphQLDateTime } from 'graphql-iso-date';
 
 import { Account, AccountFields } from '../interface/Account';
 import { AccountWithContributions, AccountWithContributionsFields } from '../interface/AccountWithContributions';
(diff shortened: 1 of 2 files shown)


server/graphql/v2/object/Event.js
```diff
@@ -51,6 +52,16 @@ export const Event = new GraphQLObjectType({
...
         },
       },
+      endsAt: {
+        description: 'The Event end date and time',
+        type: GraphQLDateTime,
+        async resolve(event) {
+          if (!event.endsAt) {
+            return null;
+          }
+          return new Date(event.endsAt);
```
znarf, line 62: What justifies a such complicated resolver? Is a custom resolver actually needed? See `createdAt` on most objects.

server/graphql/v2/object/Event.js
```diff
@@ -51,6 +52,16 @@ export const Event = new GraphQLObjectType({
           }
         },
       },
+      endsAt: {
```
znarf, line 55: Why not introduce `startsAt`? `endsAt` without `startsAt` looks a bit like a missed opportunity.

znarf (commented):
Changes requested.

server/graphql/v2/object/Event.js
SudharakaP: Yep, I thought we need it to handle the null case, but I am wrong. We don't need it. I've removed the resolver. Thanks for pointing it out. 👍🏽

server/graphql/v2/object/Event.js
SudharakaP: Well, my thinking was that we are not using `startsAt` at this stage. Only the `endsAt` is useful for this PR (to support the warning for past events). I normally go by the rule that I try not to introduce changes which are not relevant to the issue in question which the PR aims to solve. This is why I didn't introduce it. Anyways, do you think we should introduce `startsAt`? 🤔

znarf: It doesn't make sense to introduce `endsAt` alone. I'm asking you to introduce `startsAt` too.

SudharakaP: @znarf : I've added `startsAt` as well. 👍🏽

znarf (approved):

PR #6513, opened 2021-08-30
Add Data Field for Order object: For Crypto Contribution Flow

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4593

Related Front End PR: https://github.com/opencollective/opencollective-frontend/pull/6809

### DIFF
--- a/server/graphql/v2/object/Order.js
+++ b/server/graphql/v2/object/Order.js
@@ -207,7 +207,7 @@ export const Order = new GraphQLObjectType({
         type: GraphQLJSON,
         description: 'Data related to the order',
         resolve(order) {
-          return pick(order.data, ['customData.pledgeCurrency']);
+          return pick(order.data, ['customData.pledgeCurrency', 'customData.pledgeAmount']);
         },
       },
     };
--- a/server/graphql/v2/object/Order.js
+++ b/server/graphql/v2/object/Order.js
@@ -206,7 +206,7 @@ export const Order = new GraphQLObjectType({
       data: {
         type: GraphQLJSON,
         description: 'Data related to the order',
-        async resolve(order) {
+        resolve(order) {
           return pick(order.data, ['customData.pledgeCurrency']);
         },
       },
--- a/server/graphql/v2/object/Order.js
+++ b/server/graphql/v2/object/Order.js
@@ -1,6 +1,7 @@
 import { GraphQLInt, GraphQLList, GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
 import { GraphQLDateTime } from 'graphql-iso-date';
 import GraphQLJSON from 'graphql-type-json';
+import { pick } from 'lodash';
 
 import roles from '../../../constants/roles';
 import models from '../../../models';
@@ -205,6 +206,9 @@ export const Order = new GraphQLObjectType({
       data: {
         type: GraphQLJSON,
         description: 'Data related to the order',
+        async resolve(order) {
+          return pick(order.data, ['customData.pledgeCurrency']);
+        },
       },
     };
   },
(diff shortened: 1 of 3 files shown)


znarf (changes requested):
We absolutely don't want to expose the `data` field unfiltered.

SudharakaP:
> We absolutely don't want to expose the `data` field unfiltered.

@znarf : Makes sense. 👍🏽  I've only returned the relevant fields now; let me know how that looks.

server/graphql/v2/object/Order.js
```diff
@@ -201,6 +202,17 @@ export const Order = new GraphQLObjectType({
...
+        type: GraphQLJSON,
+        description: 'Data related to the order',
+        async resolve(order) {
+          return {
+            customData: {
+              pledgeCurrency: order?.data?.customData?.pledgeCurrency,
+            },
+          };
+        },
+      },
```
znarf, line 215: Can you propose something that is less **ad-hoc** and a bit more future proof? See for example: https://github.com/opencollective/opencollective-api/blob/main/server/graphql/v2/object/PaymentMethod.js#L87

```
// Protect and limit fields
const allowedFields = [];
return pick(order.data, allowedFields);
```

@Betree any other example of data field sanitization?

Betree: @znarf the example you gave is good, another one is the same one on GQLV1 https://github.com/opencollective/opencollective-api/blob/84b4118c6c63fdab8631b385bce533e94c330d88/server/graphql/v1/types.js#L1763.

Note that pick supports sub-fields, so we can do:

```suggestion
      data: {
        type: GraphQLJSON,
        description: 'Data related to the order',
        async resolve(order) {
          return pick(order.data, ['customData.pledgeCurrency']);
        },
      },
```

Betree: Btw why storing this in `customData`? This field is normally supposed to store values generated from tier's `customFields`, as per https://github.com/opencollective/opencollective-api/blob/84b4118c6c63fdab8631b385bce533e94c330d88/server/graphql/v2/input/OrderCreateInput.ts#L69.

If we want to store additional values on order, we can just store them directly in `order.data`.

SudharakaP: I've corrected the `data` field sanitation by using the `pick` method as suggested. 

As for the user of `order.data` instead of `customData` I agree and if no objections will do this in a separate PR at this might need changes both front and api. 🤔

SudharakaP: Issue opened at, https://github.com/opencollective/opencollective/issues/4607

server/graphql/v2/object/Order.js
```diff
@@ -201,6 +203,13 @@ export const Order = new GraphQLObjectType({
           return order; // Individual fields are set by OrderPermissions resolvers
         },
       },
+      data: {
+        type: GraphQLJSON,
+        description: 'Data related to the order',
+        async resolve(order) {
```
znarf, line 209:
```suggestion
        resolve(order) {
```

server/graphql/v2/object/Order.js
```diff
@@ -201,6 +203,13 @@ export const Order = new GraphQLObjectType({
           return order; // Individual fields are set by OrderPermissions resolvers
         },
       },
+      data: {
+        type: GraphQLJSON,
+        description: 'Data related to the order',
+        resolve(order) {
+          return pick(order.data, ['customData.pledgeCurrency']);
```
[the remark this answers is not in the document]
SudharakaP: Sorry, missed this one. I've added that as well. 👍🏽

znarf (approved):

PR #6514, opened 2021-08-30
Handle Error when the Deposit Address Creation Fails - Crypto Flow

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4602

image

### DIFF
--- a/server/paymentProviders/thegivingblock/index.js
+++ b/server/paymentProviders/thegivingblock/index.js
@@ -109,11 +109,14 @@ export const processOrder = async order => {
   await refresh(account);
 
   // create wallet address
-  const { depositAddress, pledgeId } = await createDepositAddress(account, {
+  const { depositAddress, pledgeId, errorMessage, meta } = await createDepositAddress(account, {
     organizationId: account.data.organizationId,
     pledgeAmount: order.data.customData.pledgeAmount,
     pledgeCurrency: order.data.customData.pledgeCurrency,
   });
+  if (errorMessage) {
+    throw new Error(`The Giving Block: ${errorMessage} ${meta.errorCode}`);
+  }
 
   // update payment method
   // TODO: update name?
--- a/server/paymentProviders/thegivingblock/index.js
+++ b/server/paymentProviders/thegivingblock/index.js
@@ -36,7 +36,6 @@ async function handleErrorsAndRetry(result, path, options = {}, account = null) 
   if (result.data.errorMessage) {
     if (result.data.meta.errorCode === 'INVALID_JWT_TOKEN' && account) {
       logger.debug('Access token is invalid. Requesting a new one.');
-      let error;
       try {
         await login(USERNAME, PASSWORD, account);
         if (options.body?.get('refreshToken')) {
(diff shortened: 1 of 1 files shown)


server/paymentProviders/thegivingblock/index.js
```diff
@@ -109,11 +109,14 @@ export const processOrder = async order => {
...
   // create wallet address
-  const { depositAddress, pledgeId } = await createDepositAddress(account, {
+  const { depositAddress, pledgeId, errorMessage, meta } = await createDepositAddress(account, {
     organizationId: account.data.organizationId,
     pledgeAmount: order.data.customData.pledgeAmount,
     pledgeCurrency: order.data.customData.pledgeCurrency,
   });
+  if (errorMessage) {
+    throw new Error(`The Giving Block: ${errorMessage} ${meta.errorCode}`);
+  }
```
znarf, line 119: Let's log the error and fail with a generic error message for now. Similar to Stripe:
https://github.com/opencollective/opencollective-api/blob/main/server/paymentProviders/stripe/creditcard.js#L18 

We'll later be able to have a list of errors that we actually want to share with the end users:
https://github.com/opencollective/opencollective-api/blob/main/server/paymentProviders/stripe/creditcard.js#L344

znarf (changes requested):

server/paymentProviders/thegivingblock/index.js
SudharakaP: Makes sense. 👍🏽 Corrected this. Now it shows as below.  

image

server/paymentProviders/thegivingblock/index.js
```diff
@@ -109,11 +112,15 @@ export const processOrder = async order => {
...
-  const { depositAddress, pledgeId } = await createDepositAddress(account, {
+  const { depositAddress, pledgeId, errorMessage, meta } = await createDepositAddress(account, {
     organizationId: account.data.organizationId,
     pledgeAmount: order.data.customData.pledgeAmount,
     pledgeCurrency: order.data.customData.pledgeCurrency,
   });
+  if (errorMessage) {
+    logger.error(`The Giving Block: ${errorMessage} ${meta.errorCode}`);
+    throw new Error(GENERIC_ERROR_MSG);
+  }
```
znarf, line 123: Why is this needed if we have `handleErrorsAndRetry`?

Add error handling there and refactor to avoid duplicate code.

SudharakaP: Yes, I've refactored code to handle the error within `handleErrorsAndRetry`. However it seems that there's cases where we could first have the `INVALID_JWT_TOKEN` error and then when we try to make the request again within the `handleErrorsAndRetry` then we could silently fail as we don't handle the response error there. I've handled this case as well. I could reproduce this in my local. 👍🏽 

Let me know how it looks now. 🤔

server/paymentProviders/thegivingblock/index.js
```diff
@@ -46,13 +47,18 @@ async function handleErrorsAndRetry(result, path, options = {}, account = null)
...
+        if (result.data.errorMessage) {
+          throw new Error(`The Giving Block: ${result.data.errorMessage} ${result.data.meta.errorCode}`);
+        }
         return result.data;
       } catch (err) {
         error = err;
       }
-      throw error;
+      logger.error(error.message);
+      throw new Error(GENERIC_ERROR_MSG);
```
znarf, line 58: I don't see the point having this out of the `catch`.

SudharakaP: Yes makes sense. I've put this within the catch and simplified. Thanks. 👍🏽

znarf (approved):

PR #6517, opened 2021-08-31
fix(graphql-cache): check variable is set before checking length

znarf:


server/graphql/cache.js
```diff
@@ -155,10 +155,10 @@ export function getGraphqlCacheKey(req) {
       if (req.body.variables.orderBy.field !== 'MEMBER_COUNT' || req.body.variables.orderBy.direction !== 'DESC') {
         return;
       }
-      if (req.body.variables.role.length !== 1) {
+      if (!req.body.variables.role || req.body.variables.role.length !== 1) {
```
Betree, line 158: Alternative

```suggestion
      if (req.body.variables.role?.length !== 1) {
```

Betree (approved):

PR #6519, opened 2021-08-31
Attach new csv to collective report

znarf:
Let's try to ship it today and gather some feedback.

### DIFF
--- a/cron/monthly/collective-report.js
+++ b/cron/monthly/collective-report.js
@@ -8,15 +8,19 @@ import moment from 'moment';
 
 import { TransactionKind } from '../../server/constants/transaction-kind';
 import { generateHostFeeAmountForTransactionLoader } from '../../server/graphql/loaders/transactions';
+import { getCollectiveTransactionsCsv } from '../../server/lib/csv';
 import { notifyAdminsOfCollective } from '../../server/lib/notifications';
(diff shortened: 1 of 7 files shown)


cron/monthly/collective-report.js
```diff
@@ -181,6 +185,20 @@ const processCollective = collective => {
         return collective;
       });
     })
+    .then(async collective => {
+      const transactionsCsvV2 = await getCollectiveTransactionsCsv(collective, { startDate, endDate });
+      if (transactionsCsvV2) {
+        const csvFilenameV2 = `${collective.slug}-${moment(d).format(dateFormat)}-transactions-v2.csv`;
+        options.attachments = [
```
kewitz, line 192: Actually, if we're including this a second report, we'll need to do:
```suggestion
        options.attachments.push(
```

kewitz (approved):
LGTM, we'll only need to make sure we're attaching both CSVs for now.

cron/monthly/collective-report.js
znarf: Thank you for that. When testing locally, we can't see the attachments in MailDev.

kewitz: You can change the view to the Attachments tab:
Screenshot from 2021-08-31 16-03-39

PR #6528, opened 2021-08-31
Move orders mutations to common

znarf:
stop calling them "legacy" as they are actively used and maintained

### DIFF
--- a/server/graphql/common/orders.js
+++ b/server/graphql/common/orders.js
@@ -7,32 +7,26 @@ import * as hcaptcha from 'hcaptcha';
 import { get, isNil, omit, pick, set } from 'lodash';
 import { isEmail } from 'validator';
 
-import activities from '../../../constants/activities';
-import CAPTCHA_PROVIDERS from '../../../constants/captcha-providers';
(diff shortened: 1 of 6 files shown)


znarf:
@Betree let me know what you think about this? One step before removing those mutations fully from GraphQL v1?

Betree (approved):
Definitely a good change. We should review the code for all the mutations when we'll drop support on V1; the reason why I kept naming things `legacyXxx` (eg. `legacyRefundTransaction`) was for us to remember that this function needs to be updated/optimized for V2 usage.

Taking `legacyRefundTransaction` as an example, we're loading the transactions two times (`fetchTransactionWithReference`, pass `id` then `models.Transaction.findByPk`). For all these mutations, we should review the arguments and return values.

It looks like there's not a lot of work to do to move everything to V2, `createOrder` V1 only seems to be used by pledges. That would be a good occasion to rather create a dedicated `createPledge` mutation on V2, it always annoyed me that we use the same mutation for these two very different actions.

znarf:
I noticed that `confirmOrder` v1 is also used in the Frontend right now (confirmOrder page)

PR #6543, opened 2021-09-02
Mutation for root users to connect accounts

Betree:
For https://github.com/opencollective/opencollective-frontend/pull/6825

### DIFF
--- a/server/graphql/v2/mutation/MemberInvitationMutations.js
+++ b/server/graphql/v2/mutation/MemberInvitationMutations.js
@@ -3,9 +3,7 @@ import { GraphQLDateTime } from 'graphql-iso-date';
 import { pick } from 'lodash';
 
 import { types as CollectiveTypes } from '../../../constants/collectives';
-import roles from '../../../constants/roles';
 import MemberRoles from '../../../constants/roles';
-import { purgeCacheForCollective } from '../../../lib/cache';
 import models from '../../../models';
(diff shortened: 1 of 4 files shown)


server/graphql/v2/mutation/MemberInvitationMutations.js
```diff
@@ -48,19 +50,24 @@ const memberInvitationMutations = {
...
-
-      if (!req.remoteUser.isAdminOfCollective(account)) {
+      const memberInfo = pick(args, ['role', 'description', 'since']);
+
+      if (args.role === roles.CONNECTED_COLLECTIVE && req.remoteUser.isRoot()) {
+        // Special case: when users are root, we allow them to create connected collectives directly
+        await models.Member.connectCollectives(memberAccount, account, req.remoteUser, memberInfo);
+        purgeCacheForCollective(account.slug);
+        purgeCacheForCollective(memberAccount.slug);
+        return null;
```
znarf, line 60: A bit weird to return `null` here? Should we use a different mutation instead?

Betree: As per https://github.com/opencollective/opencollective-frontend/pull/6825#issuecomment-913460330, I agree that using this mutation seems a bit hacky. Should I go with a (root-only) new `createMember` mutation?

PR #6545, opened 2021-09-02
Add support to validate the legal name against the payout method

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4581

Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/6818

image

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -23,6 +23,7 @@ import { ExpenseAttachedFile } from '../../models/ExpenseAttachedFile';
 import { ExpenseItem } from '../../models/ExpenseItem';
 import { PayoutMethodTypes } from '../../models/PayoutMethod';
 import paymentProviders from '../../paymentProviders';
+import { RecipientAccount as BankAccountPayoutMethodData } from '../../types/transferwise';
 import { BadRequest, FeatureNotAllowedForUser, Forbidden, NotFound, Unauthorized, ValidationFailed } from '../errors';
 
 const debug = debugLib('expenses');
@@ -627,9 +628,10 @@ export async function createExpense(
   // Create and validate TransferWise recipient
   let recipient;
   if (payoutMethod?.type === PayoutMethodTypes.BANK_ACCOUNT) {
+    const payoutMethodData = <BankAccountPayoutMethodData>payoutMethod.data;
     if (
       expenseData.fromCollective.legalName &&
-      payoutMethod.data?.accountHolderName !== expenseData.fromCollective.legalName
+      payoutMethodData?.accountHolderName !== expenseData.fromCollective.legalName
     ) {
       throw new Error('The legal name should match the bank account holder name');
     }
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -627,6 +627,9 @@ export async function createExpense(
   // Create and validate TransferWise recipient
   let recipient;
   if (payoutMethod?.type === PayoutMethodTypes.BANK_ACCOUNT) {
+    if (payoutMethod.data?.accountHolderName !== expenseData.fromCollective.legalName) {
+      throw new Error('The legal name should match the bank account holder name');
+    }
     const host = await collective.getHostCollective();
     const connectedAccounts = host && (await host.getConnectedAccounts({ where: { service: 'transferwise' } }));
     if (connectedAccounts?.[0]) {
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -629,12 +629,9 @@ export async function createExpense(
   let recipient;
   if (payoutMethod?.type === PayoutMethodTypes.BANK_ACCOUNT) {
     const payoutMethodData = <BankAccountPayoutMethodData>payoutMethod.data;
-    if (
-      !isAccountHolderNameAndLegalNameMatch(
-        payoutMethodData?.accountHolderName,
-        <string>expenseData.fromCollective.legalName,
-      )
-    ) {
+    const accountHolderName = payoutMethodData?.accountHolderName;
+    const legalName = <string>expenseData.fromCollective.legalName;
+    if (accountHolderName && legalName && !isAccountHolderNameAndLegalNameMatch(accountHolderName, legalName)) {
       throw new Error('The legal name should match the bank account holder name');
     }
     const host = await collective.getHostCollective();
@@ -728,28 +725,23 @@ export const getItemsChanges = async (
  * 4) If one of account holder name or legal name is not defined then this function returns true.
  */
 export const isAccountHolderNameAndLegalNameMatch = (accountHolderName: string, legalName: string): boolean => {
-  if (legalName && accountHolderName) {
-    const namesArray = legalName.split(' ');
-    let legalNameReversed;
-    if (namesArray.length === 2) {
-      const firstName = namesArray[0];
-      const lastName = namesArray[1];
-      legalNameReversed = `${lastName} ${firstName}`;
-    }
-    if (
-      accountHolderName.localeCompare(legalName, undefined, {
-        sensitivity: 'base',
-        ignorePunctuation: true,
-      }) &&
-      accountHolderName.localeCompare(legalNameReversed, undefined, {
-        sensitivity: 'base',
-        ignorePunctuation: true,
-      })
-    ) {
-      return false;
-    }
+  const namesArray = legalName.split(' ');
+  let legalNameReversed;
+  if (namesArray.length === 2) {
+    const firstName = namesArray[0];
+    const lastName = namesArray[1];
+    legalNameReversed = `${lastName} ${firstName}`;
   }
-  return true;
+  return !(
+    accountHolderName.localeCompare(legalName, undefined, {
+      sensitivity: 'base',
+      ignorePunctuation: true,
+    }) &&
+    accountHolderName.localeCompare(legalNameReversed, undefined, {
+      sensitivity: 'base',
+      ignorePunctuation: true,
+    })
+  );
 };
 
 export async function editExpense(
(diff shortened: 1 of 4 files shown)


server/models/PayoutMethod.ts
```diff
@@ -23,11 +23,13 @@ export enum PayoutMethodTypes {
 /** An interface for the values stored in `data` field for PayPal payout methods */
 export interface PaypalPayoutMethodData {
   email: string;
+  accountHolderName: string;
```
Betree, line 26: Why adding this to `PaypalPayoutMethodData` and `OtherPayoutMethodData`? Don't we want that just for `BankAccountPayoutMethodData`?

SudharakaP: @Betree : The reason this was added is due to typescript errors such as; 

```
server/graphql/common/expenses.ts:632:26 - error TS2339: Property 'accountHolderName' does not exist on type 'PayoutMethodDataType'.
  Property 'accountHolderName' does not exist on type 'PaypalPayoutMethodData'.
```

I couldn't find a more elegant way to do this. 🤔

Betree: The property doesn't exists on `PaypalPayoutMethodData` and `OtherPayoutMethodData`, so there's no reason to add it there. What we want is to make sure typescript understands that we're playing with a `BankAccountPayoutMethodData` in there. For example, by casting the value:

```ts
if (payoutMethod?.type === PayoutMethodTypes.BANK_ACCOUNT) {
  const payoutMethodData = <BankAccountPayoutMethodData>payoutMethod.data;
  if (payoutMethodData?.accountHolderName !== expenseData.fromCollective.legalName) {
    ...
  }
}
```

SudharakaP: Ah, that makes perfect sense. I need to brush up on my Typescript skills. 🤔

Betree (approved):
Looks good to me

Betree (commented):
Actually, can we make the check case-insensitive? Would be frustrating for users to be blocked because they've made their family name uppercase in one of the forms

SudharakaP:
> Actually, can we make the check case-insensitive? Would be frustrating for users to be blocked because they've made their family name uppercase in one of the forms

Thanks for pointing it out. I just made some checks to make sure we perform the check case-insensitively. 😄

Betree (commented):
I wanted to estimate how this change is going to impact users. I did that using this query:
```sql
SELECT MAX(e."createdAt") AS "createdAt", payee.name, pm."data" ->> 'accountHolderName'
FROM "Expenses" e
INNER JOIN "PayoutMethods" pm ON e."PayoutMethodId" = pm.id
INNER JOIN "Collectives" payee ON e."FromCollectiveId" = payee.id
WHERE pm."type" = 'BANK_ACCOUNT'
AND payee.name NOT ILIKE pm."data" ->> 'accountHolderName'
GROUP BY payee.name, pm."data" ->> 'accountHolderName'
ORDER BY MAX(e."createdAt") DESC
```

Not sharing the real names for privacy reasons, but here are a few cases that I think we should handle from the start:
- Trim `pm."data" ->> 'accountHolderName'`, as it's apparently not. Ideally we should do that on the model and have a migration, but if you want to go fast adding a `.trim()` before the comparison is also fine.
- Special characters: `François` VS `Francois`. Bank accounts sometimes don't support special characters like that, so we should ideally run both strings through `str.normalize("NFD").replace(/[\u0300-\u036f]/g, "")`
- `Evil Corp, Inc` -> `Evil Corp, Inc.` (the dot, doesn't sound like this should fail)
- First name/last name order: `Benjamin Piouffle` -> `Piouffle Benjamin`. Would be great to ignore the order of the words in the comparison.

Because of the added complexity, I would actually move all these checks to a separate function that loosely compares two names and returns a boolean to say whether they're egual.

Also, we're missing the validation when editing an expense. It should apply to both create and edit.

SudharakaP:
> I wanted to estimate how this change is going to impact users. I did that using this query:
> 
> ```sql
> SELECT MAX(e."createdAt") AS "createdAt", payee.name, pm."data" ->> 'accountHolderName'
> FROM "Expenses" e
> INNER JOIN "PayoutMethods" pm ON e."PayoutMethodId" = pm.id
> INNER JOIN "Collectives" payee ON e."FromCollectiveId" = payee.id
> WHERE pm."type" = 'BANK_ACCOUNT'
> AND payee.name NOT ILIKE pm."data" ->> 'accountHolderName'
> GROUP BY payee.name, pm."data" ->> 'accountHolderName'
> ORDER BY MAX(e."createdAt") DESC
> ```
> 
> Not sharing the real names for privacy reasons, but here are a few cases that I think we should handle from the start:
> 
>     * Trim `pm."data" ->> 'accountHolderName'`, as it's apparently not. Ideally we should do that on the model and have a migration, but if you want to go fast adding a `.trim()` before the comparison is also fine.
> 
>     * Special characters: `François` VS `Francois`. Bank accounts sometimes don't support special characters like that, so we should ideally run both strings through `str.normalize("NFD").replace(/[\u0300-\u036f]/g, "")`
> 
>     * `Evil Corp, Inc` -> `Evil Corp, Inc.` (the dot, doesn't sound like this should fail)
> 
>     * `Evil Corp Inc` -> `Evil Corp, Inc` (the comma, doesn't sound like this should fail). Maybe we should ignore punctuation altogether
> 
>     * First name/last name order: `Benjamin Piouffle` -> `Piouffle Benjamin`. Would be great to ignore the order of the words in the comparison.
> 
> 
> Because of the added complexity, I would actually move all these checks to a separate function that loosely compares two names and returns a boolean to say whether they're equal. We should also unit test it, using all the examples provided above.
> 
> Also, we're missing the validation when editing an expense. It should apply to both create and edit.

@Betree : Thanks for the detailed explanation. I've added support for all the points mentioned here using the [`localeCompare` ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) function. All punctuation is ignored and if the name consist only of two parts (first name and last name) we now make sure the order is ignored.

https://github.com/opencollective/opencollective-api/blob/55302337bf3a1fee17ea400b643dcfcdeeb2ec58/server/graphql/common/expenses.ts#L715-L744

Let me know how it looks. 😉

test/server/graphql/common/expenses.test.js
```diff
@@ -407,5 +408,13 @@ describe('server/graphql/common/expenses', () => {
       expect(await canUnschedulePayment(collectiveAccountantReq, expense)).to.be.false;
       expect(await canUnschedulePayment(hostAccountantReq, expense)).to.be.false;
     });
+
+    it('make sure legal name is validated against the account holder name', async () => {
+      expect(isAccountHolderNameAndLegalNameMatch('Evil Corp, Inc', 'Evil Corp, Inc.')).to.be.true;
+      expect(isAccountHolderNameAndLegalNameMatch('François', 'Francois')).to.be.true;
+      expect(isAccountHolderNameAndLegalNameMatch('Sudharaka Palamakumbura', 'Palamakumbura Sudharaka')).to.be.true;
```
Betree, line 415: Let's add some more tests to increase the resilience here
```suggestion
      expect(isAccountHolderNameAndLegalNameMatch('Evil Corp, Inc', 'Evil Corp, Inc.')).to.be.true;
      expect(isAccountHolderNameAndLegalNameMatch('Evil Corp Inc', 'Evil Corp, Inc.')).to.be.true;
      expect(isAccountHolderNameAndLegalNameMatch(' Evil   Corp,    Inc.', '   Evil Corp   Inc')).to.be.true;
      expect(isAccountHolderNameAndLegalNameMatch('François Dêaccènt', 'Francois DeAccEnt')).to.be.true;
      expect(isAccountHolderNameAndLegalNameMatch('Sudharaka Palamakumbura', 'Palamakumbura Sudharaka')).to.be.true;
      expect(isAccountHolderNameAndLegalNameMatch('Sudharaka Palamakumbura', 'Sudharaka Palamakumbura')).to.be.true;
```

server/graphql/common/expenses.ts
```diff
@@ -709,6 +719,39 @@ export const getItemsChanges = async (
...
+        sensitivity: 'base',
+        ignorePunctuation: true,
+      }) &&
+      accountHolderName.localeCompare(legalNameReversed, undefined, {
+        sensitivity: 'base',
+        ignorePunctuation: true,
+      })
+    ) {
+      return false;
+    }
```
Betree, line 750: :ice_cream: A small helper could improve readability here

```suggestion
    const localeCompare = (str1, str2) => str1.localeCompare(str2, undefined, { sensitivity: 'base', ignorePunctuation: true });
    if (localeCompare(accountHolderName, legalName) || localeCompare(accountHolderName, legalNameReversed)) {
      return false;
    }
```

server/graphql/common/expenses.ts
```diff
@@ -786,6 +829,19 @@ export async function editExpense(
...
+
+  // Validate bank account payout method
+  if (payoutMethod?.type === PayoutMethodTypes.BANK_ACCOUNT) {
+    const payoutMethodData = <BankAccountPayoutMethodData>payoutMethod.data;
+    if (
+      !isAccountHolderNameAndLegalNameMatch(
+        payoutMethodData?.accountHolderName,
+        <string>expenseData.fromCollective.legalName,
+      )
+    ) {
```
Betree, line 841: :ice_cream: I would personally keep the helper function just about comparing name (also to make it consistent with the frontend), and add a check here:


```suggestion
    const accountHolderName = payoutMethodData?.accountHolderName;
    const legalName = expenseData.fromCollective.legalName;
    if (accountHolderName && legalName && !namesAreEqual(accountHolderName, legalName)) {
```

Betree (commented):
Great tests here! I suggested some more to make it more resilient, the rest are just suggestions on code style that are personal preferences; feel free to ignore them.

server/graphql/common/expenses.ts
SudharakaP: Done. 👍🏽

test/server/graphql/common/expenses.test.js
SudharakaP: Done. 👍🏽

SudharakaP:
> Great tests here! I suggested some more to make it more resilient, the rest are just suggestions on code style that are personal preferences; feel free to ignore them.

Thanks for the review. I've addressed all of it. Let me know if anything further needs to be done. 👍🏽

Betree (approved):

PR #6550, opened 2021-09-03
refact: optimize Wise pay-batch endpoint

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/4728

This endpoint sometimes timeouts for specific hosts and I'm not completely sure why.
I'm trying to simplify the query here to see if I can get rid of this issue.


Betree (commented):
Fetching by `id` is supposedly super performant, I'm not convinced that this change will make a strong difference. Do you have more data (Sentry issue? Something to look for in the logs?) that we could exploit to better understand the problem?

server/controllers/transferwise.ts
```diff
@@ -50,14 +50,27 @@ export async function payBatch(
...
     const expenseIds = body?.expenseIds?.map(id => idDecode(id, IDENTIFIER_TYPES.EXPENSE));
     const expenses = await models.Expense.findAll({
       where: { id: { [Op.in]: expenseIds } },
-      include: [{ model: models.PayoutMethod, as: 'PayoutMethod' }],
+      include: [
+        { model: models.PayoutMethod, as: 'PayoutMethod', required: true },
+        {
+          model: models.Collective,
+          as: 'collective',
+          where: { HostCollectiveId: host.id },
```
Betree, line 61: Though we'll probably prevent that if there are pending expenses, collectives can change hosts and past expenses should be attached to the host of the collective at submission's time (https://github.com/opencollective/opencollective/issues/4271). I would add a small TODO here:

```suggestion
          where: { HostCollectiveId: host.id }, // TODO: We should ideally use the host attached to the expense. See https://github.com/opencollective/opencollective/issues/4271
```

server/controllers/transferwise.ts
```diff
@@ -50,14 +50,27 @@ export async function payBatch(
...
     }
-    // TODO: ensure expense ids are sent with the OTT retry request
     const expenseIds = body?.expenseIds?.map(id => idDecode(id, IDENTIFIER_TYPES.EXPENSE));
     const expenses = await models.Expense.findAll({
       where: { id: { [Op.in]: expenseIds } },
-      include: [{ model: models.PayoutMethod, as: 'PayoutMethod' }],
+      include: [
+        { model: models.PayoutMethod, as: 'PayoutMethod', required: true },
+        {
+          model: models.Collective,
```
Betree, line 59: I don't know whether we're later using `expense.collective`, but if the goal here is just to add an `INNER JOIN` to filter expenses then we can easily save some RAM (which should help if we're trying to address a timeout issue)

```suggestion
          model: models.Collective,
          attributes: [],
```

Betree (approved):
Looks solid :+1:

PR #6558, opened 2021-09-06
Accounts query

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4630

### DIFF
--- a/server/graphql/v2/query/collection/AccountsQuery.ts
+++ b/server/graphql/v2/query/collection/AccountsQuery.ts
@@ -1,9 +1,9 @@
 import { GraphQLBoolean, GraphQLList, GraphQLNonNull, GraphQLString } from 'graphql';
 
-import { Service } from '../../../../constants/connected_account';
 import models, { Op, sequelize } from '../../../../models';
 import { AccountCollection } from '../../collection/AccountCollection';
 import { AccountType } from '../../enum';
+import { PaymentMethodService } from '../../enum/PaymentMethodService';
 import { CHRONOLOGICAL_ORDER_INPUT_DEFAULT_VALUE, ChronologicalOrderInput } from '../../input/ChronologicalOrderInput';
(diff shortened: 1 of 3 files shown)


server/graphql/v2/query/collection/AccountsQuery.ts
```diff
@@ -0,0 +1,95 @@
...
+    hasCustomContributionsEnabled: {
+      type: GraphQLBoolean,
+      description: 'Only accounts with custom contribution (/donate) enabled',
+    },
+    onlyWithCreditCardSupport: {
+      type: GraphQLBoolean,
+      defaultValue: false,
+      description:
+        'If true, only accounts that accepts credit card contributions (either via Stripe or PayPal) will be returned',
+    },
```
znarf, line 35: PayPal is not really a credit card payment service for us.

Semantically, a different approach would be more future proof and precise e.g. `supportsPaymentMethod: [PayPal, CreditCard]` I understand this makes it more complicated but we have or need something similar in other places, so why not (`host { supportedPaymentMethods}`).

Betree: I initially went with this approach, but since we're deprecating the unified payment method type this kind of thing starts to be a bit tricky. Unless we introduce a new object type like `PaymentMethodProvider`, with a `service` + `type`. That way we'll be able to do:

```gql
account(supportedPaymentMethods: [{ service: STRIPE, type: CREDIT_CARD }]) {
  ...
}
```

znarf: We will need to migrate the following one to this new format (`PaymentMethodProvider`), so it's a matter of time before we start using this structure. https://github.com/opencollective/opencollective-api/blob/main/server/graphql/v2/object/Host.js#L85

I'm ok with `onlyWithCreditCardSupport` as an interim solution but let's add a comment to make clear the intent for the future.

PR #6563, opened 2021-09-06
fix: ignore bank account validation for hosts that are not connected to Wise

kewitz:
This is safe because we're using Wise only to supply the formatted bank account form and everything here should be manually paid by the Host that is not connected to Wise.


Betree (approved):
Nice fix :+1:

PR #6564, opened 2021-09-06
Flag to include/exclude GiftCard Transactions

znarf:
I decided to default to false because I think that despite being different than the current behavior, it's the most sensible choice. 

This is something that developers need to ask as an extra, not something that they need to filter. We can simply set the flag to true on the frontend where we need it.

Asking review from @Betree because this is related to Gift Cards. And @kewitz because it's related to the Transactions / CSV project.


Betree (commented):
Good to have this flag, the code looks good as well. If we default to `false`, let make sure we have matching frontend PRs where/if needed to not break anything.

Betree:
@znarf PDF PR to add the `includeGiftCardTransactions` flag: https://github.com/opencollective/opencollective-pdf/pull/768

znarf:
@Betree merging. Please merge PDF too.

PR #6573, opened 2021-09-09
createUser: Add support for legal name

Betree:
Fix https://github.com/opencollective/opencollective/issues/4651

The field was introduced in the form and in the `createCollective` mutation but not in the `createUser` mutation, breaking the invite user feature.

### DIFF
--- a/server/graphql/v1/types.js
+++ b/server/graphql/v1/types.js
@@ -3,7 +3,6 @@ import {
   GraphQLEnumType,
   GraphQLError,
   GraphQLFloat,
-  GraphQLInputObjectType,
   GraphQLInt,
   GraphQLList,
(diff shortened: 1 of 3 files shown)


server/graphql/v1/types.js
```diff
@@ -2101,25 +2100,6 @@ export const SubscriptionType = new GraphQLObjectType({
   },
 });
 
-export const UserInputType = new GraphQLInputObjectType({
```
Betree, line 2104: Also removed this type as it was not used anywhere

SudharakaP:
@Betree : Thanks much for making the change. Something I've missed during my testing of the legal names feature. 🤯 🤯

PR #6575, opened 2021-09-09
Payment receipt fixes

znarf:


server/graphql/v1/types.js
```diff
@@ -689,6 +690,7 @@ export const InvoiceType = new GraphQLObjectType({
       },
       currency: {
         type: GraphQLString,
+        deprecationReason: '2021-09-09: Not used, so we stop returning it.',
```
Betree, line 693: The currency is still used in https://github.com/opencollective/opencollective-pdf/blob/74c00ca5d73429235bfc81a3d5a1b532208058be/components/Receipt.js#L339. I can make a quick fix to use the transaction currency instead

Betree: https://github.com/opencollective/opencollective-pdf/pull/771

znarf: Seems you are still using `totalAmount` there? 

https://github.com/opencollective/opencollective-pdf/blob/74c00ca5d73429235bfc81a3d5a1b532208058be/components/Receipt.js#L339

Should I revert my changes? 

Happy to do it as I missed those fields were used in the PDF service.

Betree: No need to rollback, `totalAmount` is manually populated here: https://github.com/opencollective/opencollective-pdf/blob/e3ace4b96c6aeb97fe71733de16a7e26b2a0275e/pages/receipts/transactions/%5Buuid%5D/%5Bfilename%5D.js#L21

Betree: In https://github.com/opencollective/opencollective-pdf/pull/771/files, I removed the field from the query

PR #6576, opened 2021-09-09
Implement CSV v2 in host report

znarf:
Part of https://github.com/opencollective/opencollective/issues/4590

### DIFF
--- a/server/lib/csv.js
+++ b/server/lib/csv.js
@@ -6,12 +6,8 @@ import { fetchWithTimeout } from './fetch';
 import logger from './logger';
 import { parseToBoolean } from './utils';
 
-export const getCollectiveTransactionsCsv = async (collective, { startDate, endDate } = {}) => {
-  if (parseToBoolean(config.restService.fetchCollectiveTransactionsCsv) === false) {
-    return;
-  }
-
-  const url = new URL(`${config.host.rest}/v2/${collective.slug}/transactions.csv`);
+const getTransactionsCsv = async (type, collective, { startDate, endDate } = {}) => {
+  const url = new URL(`${config.host.rest}/v2/${collective.slug}/${type}.csv`);
   if (startDate) {
     url.searchParams.set('dateFrom', moment.utc(startDate).toISOString());
   }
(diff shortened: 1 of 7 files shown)


kewitz (approved):
Awesome. It works perfectly! 💯

templates/emails/host.report.hbs
```diff
@@ -446,8 +446,11 @@ Subject: {{#if erratumMessage}}Erratum{{#if erratumNumber}} {{erratumNumber}}{{/
   <h2>📎 Attachments</h2>
   <ul>
     <li>A CSV export of all the transactions for this {{#if month}}month{{else}}year{{/if}}</li>
+    {{#if csvV2}}
+      <li>NEW: A second CSV export in a different format (v2). Send us your feedback!</li>
```
SudharakaP, line 450: Maybe add information on how to send feedback. 🤔 

```suggestion
      <li>NEW: A second CSV export in a different format (v2). Feel free to send us your feedback. Just reply to this email!</li>
```

server/lib/csv.js
```diff
@@ -39,3 +37,19 @@ export const getCollectiveTransactionsCsv = async (collective, { startDate, endD
       logger.error(`Error fetching CSV: ${error.message}`);
     });
 };
+
+export const getCollectiveTransactionsCsv = async (collective, { startDate, endDate } = {}) => {
+  if (parseToBoolean(config.restService.fetchCollectiveTransactionsCsv) === false) {
```
SudharakaP, line 42: Maybe simplify a bit... 🤔 

```suggestion
  if (!parseToBoolean(config.restService.fetchCollectiveTransactionsCsv)) {
```

server/lib/csv.js
```diff
@@ -39,3 +37,19 @@ export const getCollectiveTransactionsCsv = async (collective, { startDate, endD
...
+    return;
+  }
+
+  return getTransactionsCsv('transactions', collective, { startDate, endDate });
+};
+
+export const getHostTransactionsCsv = async (collective, { startDate, endDate } = {}) => {
+  if (parseToBoolean(config.restService.fetchHostTransactionsCsv) === false) {
```
SudharakaP, line 50:
```suggestion
  if (!parseToBoolean(config.restService.fetchHostTransactionsCsv)) {
```

server/lib/csv.js
kewitz: If you're touching this, I would suggest to always test for the condition you want to happen and not the opposite condition:
```ts
if (parseToBoolean(config.restService.fetchCollectiveTransactionsCsv)) {
  return getTransactionsCsv('transactions', collective, { startDate, endDate });
}
```

znarf: I like the opposite. At the beginning, try to find all the reasons to not do it, return early or throw exceptions. Then, when you're good, proceed without any condition. No nested code and generally easier to read.

znarf: In different cases, generally a good idea to favor positive assertions versus negative one.

PR #6578, opened 2021-09-09
enhancement: sort projects by creation date

znarf:
https://github.com/opencollective/opencollective/issues/4650

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1186,6 +1186,10 @@ function defineModel() {
 
   Collective.prototype.getChildren = function (query = {}) {
     return Collective.findAll({
+      order: [
+        ['createdAt', 'DESC'],
+        ['id', 'DESC'],
+      ],
       ...query,
       where: { ...query.where, ParentCollectiveId: this.id },
     });
@@ -1193,6 +1197,11 @@ function defineModel() {
 
   Collective.prototype.getEvents = function (query = {}) {
     return this.getChildren({
+      order: [
+        ['startsAt', 'DESC'],
+        ['endsAt', 'DESC'],
+        ['id', 'DESC'],
+      ],
       ...query,
       where: { ...query.where, type: types.EVENT },
     });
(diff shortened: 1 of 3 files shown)


server/models/Collective.js
```diff
@@ -1186,21 +1186,26 @@ function defineModel() {
 
   Collective.prototype.getChildren = function (query = {}) {
     return Collective.findAll({
+      order: ['createdAt', 'DESC'],
```
Betree, line 1189: Following up on https://github.com/opencollective/opencollective/issues/4598 I want to update all places where we're ordering on non-unique fields to make sure we have a secondary condition. Though rare, it's not impossible to end up with transactions or collectives that have the exact same `createdAt`. In these cases, the order is unpredictable and can cause pagination issues.

The solution is quite simple, add a secondary condition on `id`. Here for example:
```suggestion
      order: [['createdAt', 'DESC'], ['id', 'DESC']],
```

Here there's almost no chance to see a problem happening (you don't create two projects at the exact same time), so no requested change.

In `getEvents`, though pretty edge case, it's also conceivable to have two events happening at the same time for a collective. We could make sure the order stays predictable with something like:

```es6
      order: [
        ['startsAt', 'DESC'],
        ['endsAt', 'DESC'],
        ['id', 'DESC'],
      ],
```

server/models/Collective.js
```diff
@@ -1186,21 +1186,26 @@ function defineModel() {
...
+      ],
+      ...omit(query, ['where']),
       where: { ...query.where, type: types.EVENT },
     });
   };
 
   Collective.prototype.getProjects = function (query = {}) {
     return this.getChildren({
-      ...query,
+      ...omit(query, ['where']),
```
Betree, line 1208: Not really necessary since `where` is overridden just after

Betree (approved):

server/models/Collective.js
znarf: I wanted to make the behavior more explicit. But I can rollback.

PR #6580, opened 2021-09-09
Discriminate PayPal when checking limits for tax forms

Betree:
Resolve https://github.com/opencollective/opencollective/issues/4645

- [x] Implement in checks
- [x] Add tests

### DIFF
--- a/server/models/LegalDocument.js
+++ b/server/models/LegalDocument.js
@@ -4,12 +4,14 @@ export const LEGAL_DOCUMENT_TYPE = {
   US_TAX_FORM: 'US_TAX_FORM',
 };
 
-function defineModel() {
-  const NOT_REQUESTED = 'NOT_REQUESTED';
-  const REQUESTED = 'REQUESTED';
-  const RECEIVED = 'RECEIVED';
-  const ERROR = 'ERROR';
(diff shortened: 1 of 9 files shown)


server/models/LegalDocument.js
```diff
@@ -4,12 +4,14 @@ export const LEGAL_DOCUMENT_TYPE = {
...
-  const NOT_REQUESTED = 'NOT_REQUESTED';
-  const REQUESTED = 'REQUESTED';
-  const RECEIVED = 'RECEIVED';
-  const ERROR = 'ERROR';
+export const LEGAL_DOCUMENT_REQUEST_STATUS = {
+  NOT_REQUESTED: 'NOT_REQUESTED',
+  REQUESTED: 'REQUESTED',
+  RECEIVED: 'RECEIVED',
+  ERROR: 'ERROR',
+};
```
znarf, line 12: When should we have constants here versus in the `constant` folder?

znarf (approved):
Looks good!

server/models/LegalDocument.js
Betree: I'm not sure, arguably all shared constants should go in the constants folder? Would be great to define some rules around this.

PR #6588, opened 2021-09-13
Add support for `order.data` field and migrate `pledgeCurrency` and `pledgeAmount` to `order.data`

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4607

Related Front-End PR: https://github.com/opencollective/opencollective-frontend/pull/6865

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -547,6 +547,7 @@ export async function createOrder(order, loaders, remoteUser, reqIp, userAgent, 
         isEmbed: Boolean(order.context?.isEmbed),
         isGuest,
         isBalanceTransfer: order.isBalanceTransfer,
+        ...order.data,
       },
       status: orderStatus,
     };
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -1,5 +1,5 @@
 import { GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
-import { isNil, isNull, isUndefined } from 'lodash';
+import { isNil, isNull, isUndefined, pick } from 'lodash';
 
 import activities from '../../../constants/activities';
 import status from '../../../constants/order_status';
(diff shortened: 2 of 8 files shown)


server/graphql/v1/mutations/orders.js
```diff
@@ -547,6 +547,7 @@ export async function createOrder(order, loaders, remoteUser, reqIp, userAgent,
         isEmbed: Boolean(order.context?.isEmbed),
         isGuest,
         isBalanceTransfer: order.isBalanceTransfer,
+        ...order.data,
```
Betree, line 550: This should be at the top, just after `data: {` because we don't want people to override the other fields, otherwise an attacker could do:

```es6
createOrder({ data: { reqIp: "You'll never now!", isFeesOnTop: false, isGuest: false })
```

If possible, we should also add a `pick` to only allow the fields that we want to keep.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -77,6 +77,14 @@ const orderMutations = {
       const collective = await loadAccount(order.toAccount);
       const paymentMethod = await getLegacyPaymentMethodFromPaymentMethodInput(order.paymentMethod);
 
+      /* Currently we only support the Giving Block related values in the data object
+       *  TODO: Maybe in future we could extend this to all existing data fields
+       */
+      let orderData;
+      if (order.data) {
+        orderData = pick(order.data, ['thegivingblock']);
```
Betree, line 85: We should move this `pick` to the common `createOrder` mutation, as we always want this sanitization to happen. Also, `pick` is powerful enough for us so sanitize with more granularity:

```suggestion
        orderData = pick(order.data, ['thegivingblock.pledgeCurrency', 'thegivingblock.pledgeAmount']);
```

Could be worth extracting this list to a shared `ORDER_PUBLIC_DATA_FIELDS` (or similar) since we're using it below.

Betree (commented):
Thanks for iterating on this one :+1: Are we planning to migrate past entries?

PR #6595, opened 2021-09-15
Add support for returning Contribution and Expense Stats related to Host

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4658


server/graphql/v2/object/Host.js
```diff
@@ -427,6 +442,113 @@ export const Host = new GraphQLObjectType({
...
       },
+      contributionStats: {
+        type: new GraphQLNonNull(ContributionStats),
+        args: {
+          accounts: {
+            type: new GraphQLList(new GraphQLNonNull(AccountReferenceInput)),
+            description: 'A collection of accounts for which the contribution stats should be returned.',
+          },
+          from: {
+            type: GraphQLString,
```
Betree, line 453:
```suggestion
            type: GraphQLDateTime,
```

server/graphql/v2/object/Host.js
```diff
@@ -427,6 +442,113 @@ export const Host = new GraphQLObjectType({
...
         },
       },
+      contributionStats: {
+        type: new GraphQLNonNull(ContributionStats),
+        args: {
+          accounts: {
+            type: new GraphQLList(new GraphQLNonNull(AccountReferenceInput)),
+            description: 'A collection of accounts for which the contribution stats should be returned.',
+          },
+          from: {
```
Betree, line 452: :ice_cream: We don't have a clear pattern for that in GQLV2, but I would personally opt for `dateFrom`/`dateTo` as we do in `TransactionsCollectionQuery`, since `from`/`to` could just as well refer to the accounts used for the contribution. @znarf any opinion on this?

server/graphql/v2/object/Host.js
```diff
@@ -427,6 +442,113 @@ export const Host = new GraphQLObjectType({
...
+            type: GraphQLString,
+            description: "Inferior date limit in which we're calculating the contribution stats",
+          },
+          to: {
+            type: GraphQLString,
```
Betree, line 457:
```suggestion
            type: GraphQLDateTime,
```

server/graphql/v2/object/Host.js
```diff
@@ -427,6 +442,113 @@ export const Host = new GraphQLObjectType({
...
+          if (!req.remoteUser?.isAdmin(host.id)) {
+            throw new Unauthorized('You need to be logged in as an admin of the host to see the contribution stats.');
+          }
+          const where = { HostCollectiveId: host.id, kind: 'CONTRIBUTION' };
+          const dateRange = dateRangeHelper(args.from, args.to);
+          if (dateRange) {
+            where.createdAt = dateRange;
+          }
+          if (args.accounts) {
+            const collectiveIds = args.accounts.map(account => account.legacyId);
```
Betree, line 471: What if someone makes a query like:

```gql
host(slug: "...") {
  contributionStats(accounts: [{ slug: "xxxx" }, { slug: "yyyy"}]) {
    ...
  }
}
```

server/graphql/v2/object/Host.js
```diff
@@ -427,6 +442,113 @@ export const Host = new GraphQLObjectType({
...
+            where.CollectiveId = { [Op.in]: collectiveIds };
+          }
+          const contributions = await models.Transaction.findAndCountAll({
```
Betree, line 474: :no_entry_sign: Here we're loading all transactions (without `SELECT`ing the fields) in RAM to sum them in JS. If we take OSC as a reference, without filtering by collectives (default), that's ~460,000 transactions. This will probably crash or timeout. Let's COUNT and SUM in SQL, even if that's mean having two queries.

server/graphql/v2/object/Host.js
```diff
@@ -427,6 +442,113 @@ export const Host = new GraphQLObjectType({
...
+            include: [{ model: models.Order, where: { interval: { [Op.ne]: null }, status: 'ACTIVE' } }],
+          });
+          const dailyAvgIncome =
+            contributions.rows.map(contribution => contribution.amount).reduce((result, item) => result + item, 0) /
+              contributions.count || 0;
+          return {
+            numContributions: contributions.count,
+            numOneTime,
+            numRecurring,
+            dailyAvgIncome,
```
Betree, line 492: Sounds like we could rely on GraphQL features to run the queries asynchronously:

```suggestion
            numContributions: models.Transaction.count(...),
            numOneTime: models.Transaction.count(...),
            numRecurring: models.Transaction.count(...),
            dailyAvgIncome: models.Transaction.count(...),
```

server/graphql/v2/object/ContributionStats.js
```diff
@@ -0,0 +1,12 @@
+import { GraphQLFloat, GraphQLInt, GraphQLObjectType } from 'graphql';
+
+export const ContributionStats = new GraphQLObjectType({
+  name: 'ContributionStats',
+  description: 'Contribution statistics related to the given accounts',
+  fields: () => ({
```
Betree, line 6: All these fields should be non-null

server/graphql/v2/object/ContributionStats.js
```diff
@@ -0,0 +1,12 @@
...
+    numContributions: { type: GraphQLInt, description: 'The total number of contributions' },
```
Betree, line 7: I personally find the `_Count` pattern clear than `num_` (eg. `numContributions` -> `contributionsCount`). And it's consistent with what we do on `Collections`, where we say `totalCount`.

@znarf any opinion?

server/graphql/v2/object/ContributionStats.js
```diff
@@ -0,0 +1,12 @@
...
+    numOneTime: { type: GraphQLInt, description: 'Number of one time contributions' },
+    numRecurring: { type: GraphQLInt, description: 'Number of recurring contributions' },
+    dailyAvgIncome: { type: GraphQLFloat, description: 'The daily average income' },
```
Betree, line 10:
```suggestion
    dailyAvgIncome: { type: new GraphQLNonNull(Amount), description: 'The daily average income' },
```

Betree (commented):
The general design looks pretty good to me.

I only provided feedback for `contributionStats`, but most comments also apply to `expenseStats`.

server/graphql/v2/object/Host.js
znarf: Yeah definitely `dateFrom` and `dateTo` like the transactions query. Never invent new names if they exist elsewhere.

After years in API design, my personal favorite is actually `since`/`before` but yeah, `dateFrom` and `dateTo` here.

server/graphql/v2/object/ContributionStats.js
znarf: Absolutely.

server/graphql/v2/object/ExpenseStats.js
```diff
@@ -0,0 +1,13 @@
+import { GraphQLFloat, GraphQLInt, GraphQLObjectType } from 'graphql';
+
+export const ExpenseStats = new GraphQLObjectType({
+  name: 'ExpenseStats',
+  description: 'Expense statistics related to the given accounts',
+  fields: () => ({
+    numExpenses: { type: GraphQLInt, description: 'The total number of expenses' },
+    dailyAverage: { type: GraphQLFloat, description: 'The daily average paid in expenses' },
```
znarf, line 8: Amount

znarf:
```suggestion
    dailyAverageAmount: { type: Amount, description: 'The daily average paid in expenses' },
```

server/graphql/v2/object/ContributionStats.js
znarf:
```suggestion
    dailyAverageIncome: { type: new GraphQLNonNull(Amount), description: 'The daily average income' },
```

Never abbreviations like 'avg' please, full words.

server/graphql/v2/object/Host.js
```diff
@@ -427,6 +442,113 @@ export const Host = new GraphQLObjectType({
...
+            type: new GraphQLList(new GraphQLNonNull(AccountReferenceInput)),
+            description: 'A collection of accounts for which the contribution stats should be returned.',
+          },
+          from: {
+            type: GraphQLString,
+            description: "Inferior date limit in which we're calculating the contribution stats",
+          },
+          to: {
+            type: GraphQLString,
+            description: "Superior date limit in which we're calculating the contribution stats",
```
znarf, line 458:
```suggestion
            description: "Calculate contribution stats until this date.",
```

server/graphql/v2/object/Host.js
znarf: Reference inputs are meant to be processed with reference helpers.

server/graphql/v2/object/Host.js
znarf: @Betree Should they be in separate resolvers?

Betree: That's an option, I don't have any strong opinion at this point since grouping everything in the same resolvers could also allow for some nice optimizations. For example, stuff like:

```es6
const stats = models.Transaction.({ ..., groupBy: 'order.interval' });
return {
  ...,
  numOneTime: stats.find(s => !s.interval).count,
  numRecurring: stats.find(s => s.interval).count,
}
```

server/graphql/v2/object/Host.js
```diff
@@ -23,12 +24,26 @@ import { CollectionArgs } from '../interface/Collection';
...
 import { Amount } from './Amount';
+import { ContributionStats } from './ContributionStats';
+import { ExpenseStats } from './ExpenseStats';
 import { HostMetrics } from './HostMetrics';
 import { HostMetricsTimeSeries } from './HostMetricsTimeSeries';
 import { HostPlan } from './HostPlan';
 import { PaymentMethod } from './PaymentMethod';
 import PayoutMethod from './PayoutMethod';
 
+const dateRangeHelper = (startDate, endDate) => {
```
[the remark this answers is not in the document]
SudharakaP: I've renamed this to `getFilterDateRange`; what do you think about that name? 🤔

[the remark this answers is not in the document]
Betree: Definitely more accurate :100:

server/graphql/v2/object/Host.js
SudharakaP: For this one, I've changed the queries to run asynchronously except the queries related to calculating the averages; since both the count and sum needs to be calculated before-handed. Do you think that's acceptable, or are there any further optimizations to be done? 🤔

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +452,124 @@ export const Host = new GraphQLObjectType({
...
+            where.createdAt = dateRange;
+          }
+          if (args.account) {
+            const collectiveIds = args.account.map(async account => {
+              const accountObj = await fetchAccountWithReference(account, { throwIfMissing: true });
+              return accountObj.id;
+            });
+            await Promise.all(collectiveIds).then(collectiveIds => {
+              where.CollectiveId = { [Op.in]: collectiveIds };
+            });
```
Betree, line 487: This seems like a need you could see in other places, worth adding a helper (eg. `fetchAccountsWithReferences`) in my opinion. This part is a personal preference.

However, the part I'm blocking on is the fetching of collectives one by one, it's necessarily expensive: try it with 10 accounts, you'll trigger 10 SQL queries.

```es6
export const fetchAccountsWithReferences = async (inputs, ...) => {
  const conditions = inputs.map(getSQLConditionFromAccountReferenceInput);
  const accounts = await models.Collective.findAll({ where: { [Op.or]: conditions } });
  if (account.length !== inputs.length && throwIfMissing) {
    ...
  }

  return accounts;
}
```

```suggestion
            const accounts = await fetchAccountsWithReferences(args.account, { ... });
            where.CollectiveId = accounts.map(a => a.id);
```

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +452,124 @@ export const Host = new GraphQLObjectType({
...
+          }
+          const contributionsCount = await models.Transaction.count({
+            where,
+          });
+          const contributionsAmountSum = await models.Transaction.sum('amount', where);
```
Betree, line 492: We're counting both `DEBIT` and `CREDIT` in there, which probably results in incorrect numbers. Same for `contributionsCount`, and probably the other stats too.

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +452,124 @@ export const Host = new GraphQLObjectType({
...
+          const dailyAverageIncomeAmount = contributionsCount !== 0 ? contributionsAmountSum / contributionsCount : 0;
```
Betree, line 493: Are you sure about this number? To me `contributionsAmountSum / contributionsCount` gives you the average amount **per contribution**, not the average amount **per day** (daily).

I would expect something called `dailyAverageIncomeAmount` to be equal to something like `contributionsAmountSum / numberOfDaysCoveredByPeriod`.

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +452,124 @@ export const Host = new GraphQLObjectType({
...
+          return {
+            contributionsCount,
+            oneTimeContributionsCount: models.Transaction.count({
+              where,
+              include: [{ model: models.Order, where: { interval: null } }],
+            }),
+            recurringContributionsCount: models.Transaction.count({
+              where,
+              include: [{ model: models.Order, where: { interval: { [Op.ne]: null }, status: 'ACTIVE' } }],
```
Betree, line 502: Having `status = 'ACTIVE'` means that we'll only return stats for active contributions. So if a recurring contribution gets canceled and you start looking at past data, it won't be surfaced. I think the status of the order doesn't matter here.

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +452,124 @@ export const Host = new GraphQLObjectType({
...
+            }),
+            dailyAverageIncomeAmount: {
+              valueInCents: dailyAverageIncomeAmount,
+              value: dailyAverageIncomeAmount / 100,
+              currency: host.currency,
```
Betree, line 507: From `server/graphql/v2/object/Amount.js`, this object only expects a `value` (in cents) and a `currency`. The properties themselves are generated by loaders. Here, all the amounts returned are divided by 100.

```suggestion
              value: dailyAverageIncomeAmount,
              currency: host.currency,
```

Betree (commented):
It's far better, great progress. We need to iterate on it before shipping, the feedback I've left below also applies to `expenseStats`.

server/graphql/v2/object/Host.js
znarf: What do we want to achieve there? Counting total contributions amount is a bit unclear to me?

When we're doing such things in other places, the query is usually **way** more complicated.

server/graphql/v2/object/Host.js
SudharakaP: Thanks for pointing it out. I've corrected this. 👍🏽

server/graphql/v2/object/Host.js
SudharakaP: Ah, thanks. I've removed the active status. 👍🏽

server/graphql/v2/object/Host.js
SudharakaP: Thanks for pointing this out. I've corrected the averages by calculating the number of days for the given periods. So if the `dateFrom` is `null` I've made sure to count the stats from `2015` as we don't have any data before 2015 if I am not mistaken. 🤔

server/graphql/v2/object/Host.js
SudharakaP: Ah, thanks for the example. Should have caught this yesterday. Now I've refactored the code and added a new function as you suggested which just gives the collection of collectives in one SQL. Let me know if there's any other suggestions on this. 😉

server/graphql/v2/object/Host.js
SudharakaP: @Betree : For this one, I've made the contributions query consider only `CREDITS` and the expense queries only consider `DEBITS`. Does that make sense? What I was thinking is that for contributions we are interested in the amounts that came into the collective whereas for expenses it's the other way around. 🤔 

@znarf : The idea here is to show the total contributions and expenses for all the collectives hosted by a given host. My understanding is that we need the total money in and the total money out (correct me if I am wrong @Betree). 🤔 

image

server/graphql/v2/input/AccountReferenceInput.js
```diff
@@ -93,3 +94,36 @@ export const fetchAccountWithReference = async (
...
+    const conditions = [];
+    inputs.forEach(input => {
+      if (input.id && typeof input.id === 'string') {
+        const id = idDecode(input.id, 'account');
+        conditions.push({ id });
+      } else if (input.legacyId || typeof input.id === 'number') {
+        conditions.push({ id: input.legacyId || input.id });
+      } else if (input.slug) {
+        conditions.push({ slug: input.slug.toLowerCase() });
+      }
```
Betree, line 116:
```suggestion
      } else {
        throw new Error('Please provide an id or a slug');
      }
```

server/graphql/v2/input/AccountReferenceInput.js
```diff
@@ -93,3 +94,36 @@ export const fetchAccountWithReference = async (
...
+ */
+export const fetchAccountsWithReferences = async (inputs, { throwIfMissing = false }) => {
+  const getSQLConditionFromAccountReferenceInput = inputs => {
+    const conditions = [];
+    inputs.forEach(input => {
+      if (input.id && typeof input.id === 'string') {
+        const id = idDecode(input.id, 'account');
+        conditions.push({ id });
+      } else if (input.legacyId || typeof input.id === 'number') {
+        conditions.push({ id: input.legacyId || input.id });
```
Betree, line 113: I don't see how we could end up with different types here. The expected input type is an `AccoutReferenceInput`, if we're passing something else then we're misusing the function.

```suggestion
      if (input.id) {
        conditions.push({ id: idDecode(input.id, 'account') });
      } else if (input.legacyId) {
        conditions.push({ id: input.legacyId });
```

server/graphql/v2/input/AccountReferenceInput.js
```diff
@@ -93,3 +94,36 @@ export const fetchAccountWithReference = async (
...
+        conditions.push({ slug: input.slug.toLowerCase() });
+      }
+    });
+
+    return conditions;
+  };
+
+  const conditions = getSQLConditionFromAccountReferenceInput(inputs);
+  const accounts = await models.Collective.findAll({ where: { [Op.or]: conditions } });
+  if (accounts.length !== inputs.length && throwIfMissing) {
```
Betree, line 124: :ice_cream: Edge case so feel free to ignore it (but maybe add a comment)?

If you pass `account: [{ slug: "opensource" }, { legacyId: 11004 }]` this will throw (since both input are referring to the same account). We may want our check to look like:

```es6
const accountLoadedForInput = input => accounts.some(account => accountMatchesInput(account, input));
if (throwIfMissing && !inputs.every(accountLoadedForInput)) {
```

But the current code is an ok solution for a first version in my opinion.

server/graphql/v2/object/Host.js
```diff
@@ -15,20 +19,53 @@ import { HostApplicationCollection } from '../collection/HostApplicationCollecti
...
+  } else if (startDate) {
+    dateRange = { [Op.gte]: startDate };
+  } else if (endDate) {
+    dateRange = { [Op.lt]: endDate };
+  }
+  return dateRange;
+};
+
+const getNumberOfDays = (startDate, endDate) => {
+  const startTimeOfStatistics = new Date(2015, 0, 1);
```
Betree, line 55: This looks arbitrary. Maybe a better default value could be:

```suggestion
  const startTimeOfStatistics = host.createdAt;
```

server/graphql/v2/object/Host.js
```diff
@@ -15,20 +19,53 @@ import { HostApplicationCollection } from '../collection/HostApplicationCollecti
...
+  if (startDate && endDate) {
+    numberOfDays = days(startDate, endDate);
+  } else if (startDate) {
+    numberOfDays = days(startDate);
+  } else if (endDate) {
+    numberOfDays = days(startTimeOfStatistics, endDate);
+  } else {
+    numberOfDays = days(startTimeOfStatistics);
+  }
+  return numberOfDays;
```
Betree, line 66: :ice_cream: Looks like this can be simplified:

```suggestion
  const since = startDate || host.createdAt;
  return days(since, endDate);
```

server/graphql/v2/input/AccountReferenceInput.js
SudharakaP: Thanks for pointing it out. I've done this as well. Let me know how it looks. 😄

SudharakaP:
@Betree : I've incorporated all the reviews above. Hope it looks better now Thanks for the reviews. 👍🏽

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +464,120 @@ export const Host = new GraphQLObjectType({
...
+          const dailyAverageIncomeAmount = contributionsAmountSum ? contributionsAmountSum / numberOfDays : 0;
+          return {
+            contributionsCount,
+            oneTimeContributionsCount: models.Transaction.count({
+              where,
+              include: [{ model: models.Order, where: { interval: null } }],
+            }),
+            recurringContributionsCount: models.Transaction.count({
+              where,
+              include: [{ model: models.Order, where: { interval: { [Op.ne]: null } } }],
```
Betree, line 515: I think this one will return erroneous values: if you have one monthly recurring contribution with 8 transactions, it will return `recurringContributionsCount=8`. We should group by `OrderId`. Same with `oneTimeContributionsCount`, though we're not supposed to have multiple transactions for the same order for this one, but just in case...

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +464,120 @@ export const Host = new GraphQLObjectType({
...
+          },
+        },
+        async resolve(host, args, req) {
+          if (!req.remoteUser?.isAdmin(host.id)) {
+            throw new Unauthorized('You need to be logged in as an admin of the host to see the contribution stats.');
+          }
+          const where = {
+            HostCollectiveId: host.id,
+            kind: TransactionKind.CONTRIBUTION,
+            type: TransactionTypes.CREDIT,
```
Betree, line 490: There may be a better way to handle these, but for now I would just ignore them:

```suggestion
            type: TransactionTypes.CREDIT,
            isRefund: false,
            RefundTransactionId: null
```

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +464,120 @@ export const Host = new GraphQLObjectType({
...
+          if (dateRange) {
+            where.createdAt = dateRange;
+          }
+          if (args.account) {
+            const collectives = await fetchAccountsWithReferences(args.account, { throwIfMissing: true });
+            const collectiveIds = collectives.map(collective => collective.id);
+            where.CollectiveId = { [Op.in]: collectiveIds };
+          }
+          const expensesCount = await models.Transaction.count({
+            where,
```
Betree, line 556: Should group by `ExpenseId` because some expenses can have multiple transactions (they get marked as paid, then marked as unpaid, then re-marked as paid). Also for `invoicesCount` and the other below.

server/graphql/v2/object/Host.js
SudharakaP: Thanks for the catch. I've corrected it for both `recurringContributionsCount` and `oneTimeContributionsCount`. 👍🏽

server/graphql/v2/object/Host.js
SudharakaP: Thanks. I've corrected this. 👍🏽

SudharakaP:
@Betree : I've added the `GroupId` for both the expenses and the contributions counts. Now I think it's better. Let me know if you see anything else. 😃

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +464,148 @@ export const Host = new GraphQLObjectType({
...
+          const contributionsCount = await models.Transaction.count({
+            where,
+          });
+
+          const oneTimeContributionCountByOrderId = await models.Transaction.count({
+            where,
+            include: [{ model: models.Order, where: { interval: null } }],
+            group: ['OrderId'],
+          });
+          const oneTimeContributionsCount = oneTimeContributionCountByOrderId.length;
```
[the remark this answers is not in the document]
SudharakaP: It's not returning a number when we group it by `OrderId`. What it returns is an array with one time contribution counts by order id (hence the name `oneTimeContributionCountByOrderId`). Something like,

`[{orderId: 10882, count: 2}, {orderId: 10883, count: 5}, ...]`

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +464,148 @@ export const Host = new GraphQLObjectType({
...
+            where,
+            include: [{ model: models.Order, where: { interval: { [Op.ne]: null } } }],
+            group: ['OrderId'],
+          });
+          const recurringContributionsCount = recurringContributionCountByOrderId.length;
+
+          const contributionsAmountSum = await models.Transaction.sum('amount', { where });
+          const dailyAverageIncomeAmount = contributionsAmountSum ? contributionsAmountSum / numberOfDays : 0;
+          return {
+            contributionsCount,
```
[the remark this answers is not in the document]
SudharakaP: Thanks. That makes sense, I've refactored all of them to use functions. 😉

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +464,148 @@ export const Host = new GraphQLObjectType({
...
+            where.createdAt = dateRange;
+          }
+          if (args.account) {
+            const collectives = await fetchAccountsWithReferences(args.account, { throwIfMissing: true });
+            const collectiveIds = collectives.map(collective => collective.id);
+            where.CollectiveId = { [Op.in]: collectiveIds };
+          }
+          const group = ['ExpenseId'];
+
+          const expenseCountByExpenseId = await models.Transaction.count({
```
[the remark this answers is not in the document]
SudharakaP: I've refactored each of these to return using the function notation as per the previous review, so we don't need these variable names anyways. 👍🏽

server/graphql/v2/object/Host.js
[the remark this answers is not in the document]
Betree: Then it sounds like this query is far from being optimal. For large collectives, it's going to return an array with 5000+ objects while we're only interested in one integer value.

In raw SQL, we could write something like:
```sql
SELECT COUNT (DISTINCT "OrderId") FROM "Transactions" t 
```

It would be great if we could something like that with Sequelize. Otherwise we can still fallback on raw SQL.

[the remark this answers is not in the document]
SudharakaP: @Betree : I should have caught this in the beginning. We could use the `distinct` parameter; https://sequelize.org/master/class/lib/model.js~Model.html#static-method-count. I've changed that and now it should look better. 😉

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +464,135 @@ export const Host = new GraphQLObjectType({
...
+          const dateRange = getFilterDateRange(args.dateFrom, args.dateTo);
+          if (dateRange) {
+            where.createdAt = dateRange;
+          }
+          if (args.account) {
+            const collectives = await fetchAccountsWithReferences(args.account, { throwIfMissing: true });
+            const collectiveIds = collectives.map(collective => collective.id);
+            where.CollectiveId = { [Op.in]: collectiveIds };
+          }
+          const distinct = { distinct: true, col: 'OrderId' };
```
Betree, line 504: Is this option documented? I can't find it on https://sequelize.org/master/. I'm also a bit confused by the usage, in some places you pass it as `options.distinct = true` while others are `option.distinct.distinct = true`

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +464,135 @@ export const Host = new GraphQLObjectType({
...
+            isRefund: false,
+            RefundTransactionId: null,
+          };
+          const numberOfDays = getNumberOfDays(args.dateFrom, args.dateTo, host);
+          const dateRange = getFilterDateRange(args.dateFrom, args.dateTo);
+          if (dateRange) {
+            where.createdAt = dateRange;
+          }
+          if (args.account) {
+            const collectives = await fetchAccountsWithReferences(args.account, { throwIfMissing: true });
```
Betree, line 500: Small optimization if you rebase on `main`

```suggestion
            const collectives = await fetchAccountsWithReferences(args.account, { throwIfMissing: true, attributes: ['id'] });
```

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +464,135 @@ export const Host = new GraphQLObjectType({
...
+                include: [{ model: models.Order, where: { interval: null } }],
+                ...distinct,
+              }),
+            recurringContributionsCount: () =>
+              models.Transaction.count({
+                where,
+                include: [{ model: models.Order, where: { interval: { [Op.ne]: null } } }],
+                ...distinct,
+              }),
+            dailyAverageIncomeAmount: {
```
Betree, line 525: Looks like we could put this one in a function too and do the `contributionsAmountSum`/dailyAverage there

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +464,135 @@ export const Host = new GraphQLObjectType({
...
+          return {
+            expensesCount: () =>
+              models.Transaction.count({
+                where,
+                ...distinct,
+              }),
+            invoicesCount: models.Transaction.count({
+              where,
+              include: [{ model: models.Expense, where: { type: expenseType.INVOICE } }],
+              distinct,
```
Betree, line 576: Here and in other places I think we're missing the `...`, we're currently passing options as `options.distinct.distinct = true`.

Betree (commented):
We should not do it now, but thinking about it these stats should not be limited to hosts. Ideally, I'd want to be able to fetch these stats for all account types. Examples:

```gql
account(slug: "opensource") {
  contributionsStats(includeHostedAccounts: true) {
    contributionsCount
  }
}
```

```gql
account(slug: "babel") {
  contributionsStats {
    contributionsCount
  }
}
```

Let's keep that in mind for a V2.

server/graphql/v2/object/Host.js
SudharakaP: Done. 👍🏽

server/graphql/v2/object/Host.js
SudharakaP: My mistake; thanks for pointing it out. 👍🏽

SudharakaP:
> We should not do it now, but thinking about it these stats should not be limited to hosts. Ideally, I'd want to be able to fetch these stats for all account types. Examples:
> 
> ```graphql
> account(slug: "opensource") {
>   contributionsStats(includeHostedAccounts: true) {
>     contributionsCount
>   }
> }
> ```
> 
> ```graphql
> account(slug: "babel") {
>   contributionsStats {
>     contributionsCount
>   }
> }
> ```
> 
> Let's keep that in mind for a V2.

Just so we don't miss this, I've created a separate issue for this; https://github.com/opencollective/opencollective/issues/4781

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +464,142 @@ export const Host = new GraphQLObjectType({
...
+            throw new Unauthorized('You need to be logged in as an admin of the host to see the contribution stats.');
+          }
+          const where = {
+            HostCollectiveId: host.id,
+            kind: TransactionKind.CONTRIBUTION,
+            type: TransactionTypes.CREDIT,
+            isRefund: false,
+            RefundTransactionId: null,
+          };
+          const numberOfDays = getNumberOfDays(args.dateFrom, args.dateTo, host);
```
Betree, line 494: If dateFrom & dateTo are the same day (which happens if you pick today), this will equal 0.

`contributionsAmountSum / 0` = 

```suggestion
          const numberOfDays = getNumberOfDays(args.dateFrom, args.dateTo, host) || 1;
```

Betree (commented):
The current version of the code looks great to me, good job leading this forward @SudharakaP. Let's just be safe on the division by 0 issue mentioned below and we'll be good to go!

server/graphql/v2/object/Host.js
SudharakaP: Thanks for pointing it out. I've handled this both here as well as in the expense stats. 👍🏽

SudharakaP:
> The current version of the code looks great to me, good job leading this forward @SudharakaP. Let's just be safe on the division by 0 issue mentioned below and we'll be good to go!

Thanks. 👍🏽 I've handed the division by 0 issue as well. Let me know if you see anything else on this one. 👍🏽

znarf:
@SudharakaP This was not deployed.

SudharakaP:
> @SudharakaP This was not deployed.

Sorry, I seem to have missed this although I thought I did this on Friday. Anyways this has happened a couple of occasions so I've added a reminder for myself to check on this everyday from now on just so to prevent this from happening again. 🤔

PR #6601, opened 2021-09-16
fix: use DEBIT transaction when refunding expenses

kewitz:
Related to https://opencollective.freshdesk.com/a/tickets/35682

Otherwise we get the error:
```
TypeError: Cannot read property 'getHostCollective' of null
    at Function.<anonymous> (/home/leo/Code/opencollective/api/server/models/Transaction.js:522:53),    
    at Function.createDoubleEntry (/home/leo/Code/opencollective/api/server/models/Transaction.js:510:3),
    at /home/leo/Code/opencollective/api/server/lib/payments.js:195:28,
    at refundPaymentProcessorFeeToCollective (/home/leo/Code/opencollective/api/server/lib/payments.js:187:51),
    at Object.<anonymous> (/home/leo/Code/opencollective/api/server/lib/payments.js:305:13),
    at Object.createRefundTransaction (/home/leo/Code/opencollective/api/server/lib/payments.js:319:35),
    at /home/leo/Code/opencollective/api/server/graphql/common/expenses.ts:1284:23,
    at /home/leo/Code/opencollective/api/server/graphql/common/expenses.ts:994:12
```

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -301,8 +301,12 @@ export async function createRefundTransaction(transaction, refundedPaymentProces
         `Partial processor fees refunds are not supported, got ${refundedPaymentProcessorFee} for #${transaction.id}`,
       );
     } else if (transaction.paymentProcessorFeeInHostCurrency) {
+      // When refunding an Expense, we need to use the DEBIT transaction which is attached to the Collective and its Host.
+      const transactionToRefundPaymentProcessorFee = transaction.ExpenseId
+        ? await transaction.getRelatedTransaction({ type: DEBIT })
+        : transaction;
       // Host take at their charge the payment processor fee that is lost when refunding a transaction
-      await refundPaymentProcessorFeeToCollective(transaction, transactionGroup);
+      await refundPaymentProcessorFeeToCollective(transactionToRefundPaymentProcessorFee, transactionGroup);
     }
   }
 
--- a/test/stories/ledger.test.ts
+++ b/test/stories/ledger.test.ts
@@ -13,10 +13,17 @@ import {
   PLATFORM_TIP_TRANSACTION_PROPERTIES,
   SETTLEMENT_EXPENSE_PROPERTIES,
 } from '../../server/constants/transactions';
-import { payExpense } from '../../server/graphql/common/expenses';
+import { markExpenseAsUnpaid, payExpense } from '../../server/graphql/common/expenses';
 import { createRefundTransaction, executeOrder } from '../../server/lib/payments';
 import models from '../../server/models';
-import { fakeCollective, fakeHost, fakeOrder, fakePayoutMethod, fakeUser } from '../test-helpers/fake-data';
+import {
+  fakeCollective,
+  fakeExpense,
+  fakeHost,
+  fakeOrder,
+  fakePayoutMethod,
+  fakeUser,
+} from '../test-helpers/fake-data';
 import { nockFixerRates, resetTestDB, snapshotLedger } from '../utils';
 
 const SNAPSHOT_COLUMNS = [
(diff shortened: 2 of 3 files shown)


znarf:
I don't have enough to understand the problem and review the PR.

What is the case we're looking at? Why was getHostCollective() failing here? Is there something special? Is this new?

Please details in the PR or code comments.

kewitz:
@znarf there's a link to the support ticket in the PR with more context on the real problem.

To summarize, at the beginning of `createRefundTransaction` we do:
```js
if (transaction.type === DEBIT) {
  transaction = await transaction.getRelatedTransaction({ type: CREDIT });
}
```
And if the we're also refunding payment processor fees, in `refundPaymentProcessorFeeToCollective()` we'll create the refund transactions with:
```js
 await models.Transaction.createDoubleEntry({
    type: CREDIT,
    kind: TransactionKind.PAYMENT_PROCESSOR_COVER,
    CollectiveId: transaction.CollectiveId,
    FromCollectiveId: transaction.HostCollectiveId,
    HostCollectiveId: transaction.HostCollectiveId,
    ...
});
```

The issue is that, if we're dealing with an Expense, the credit transaction is attributed to the user and not the collective.
What we really want is to have the Host of the collective covering for the payment processor fees charged when paying for the expense.
That's why we should use the debit transaction when dealing with an expense.

znarf:
> Why was getHostCollective() failing here? Is there something special? Is this new?

kewitz:
> Why was getHostCollective() failing here? Is there something special? Is this new?
It fails because, on an expense transaction, we'll create a double transaction with
```js
const fromCollective = await models.Collective.findByPk(transaction.FromCollectiveId);
const fromCollectiveHost = await fromCollective.getHostCollective();
```
and we call that with the "FromCollectiveId" being the host of the receiving user (who usually does not have a host).
More importantly, it makes zero sense to refund the payment processor fee of an expense to the user.

znarf:
> const fromCollective = await models.Collective.findByPk(transaction.FromCollectiveId);

No reason for me to have `fromCollective = null`. I don't get it.

> More importantly, it makes zero sense to refund the payment processor fee of an expense to the user.

Happy so see this challenged, I don't see the justification for this.

znarf:
I now understand that `transaction.FromCollectiveId = otherTransaction.HostCollectiveId`.

I think we should fail earlier because we're not supposed to even try to create a Transaction with `FromCollectiveId=null`.

kewitz:
> No reason for me to have `fromCollective = null`. I don't get it.

The CREDIT transaction of an expense will have no `HostCollectiveId` because it credits the `ColletiveId` of the user.
Therefore, when you call `createDoubleEntry` on `refundPaymentProcessorFeeToCollective` you're calling
```js
await models.Transaction.createDoubleEntry({
    type: CREDIT,
    kind: TransactionKind.PAYMENT_PROCESSOR_COVER,
    CollectiveId: transaction.CollectiveId,
    FromCollectiveId: null,
    HostCollectiveId: null,
    ...
});
```

> Happy so see this challenged, I don't see the justification for this.
When you pay for an expense, you're debiting credit from the collective and paying the user. If you're refunding and the host is covering for the payment fees, the cover should go from the host to the collective.

znarf:
>>> More importantly, it makes zero sense to refund the payment processor fee of an expense to the user.

>> Happy so see this challenged, I don't see the justification for this.

> When you pay for an expense, you're debiting credit from the collective and paying the user. If you're refunding and the host is covering for the payment fees, the cover should go from the host to the collective.

Do we really want to do a `PAYMENT_PROCESSOR_COVER` scheme for Expenses? 

Let's say, yes. I would like to see some test coverage for that case in: https://github.com/opencollective/opencollective-api/blob/main/test/stories/ledger.test.ts

znarf (changes requested):
As written before, please add test coverage for the case in:
https://github.com/opencollective/opencollective-api/blob/main/test/stories/ledger.test.ts

server/lib/payments.js
```diff
@@ -301,8 +301,12 @@ export async function createRefundTransaction(transaction, refundedPaymentProces
         `Partial processor fees refunds are not supported, got ${refundedPaymentProcessorFee} for #${transaction.id}`,
       );
     } else if (transaction.paymentProcessorFeeInHostCurrency) {
+      // When refunding an Expense, we need to use the DEBIT transaction which is attached to the Collective and its Host.
+      const transactionToRefundPaymentProcessorFee = transaction.ExpenseId
```
znarf, line 305: Maybe `transaction.kind === EXPENSE` would have been more semantical.

znarf (approved):

PR #6607, opened 2021-09-20
Rename period args to dateFrom/dateTo

Betree:
This is a breaking change for the host reports page, but since it's not been exposed publicly yet it's ok to take this shortcut.

### DIFF
--- a/server/graphql/v2/object/Host.js
+++ b/server/graphql/v2/object/Host.js
@@ -73,14 +73,24 @@ export const Host = new GraphQLObjectType({
           from: {
             type: GraphQLString,
             description: "Inferior date limit in which we're calculating the metrics",
+            deprecationReason: '2020-09-20: Use dateFrom',
           },
           to: {
             type: GraphQLString,
             description: "Superior date limit in which we're calculating the metrics",
+            deprecationReason: '2020-09-20: Use dateTo',
+          },
+          dateFrom: {
+            type: GraphQLDateTime,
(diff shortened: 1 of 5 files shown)


server/graphql/v2/object/Host.js
```diff
@@ -73,26 +73,36 @@ export const Host = new GraphQLObjectType({
...
+            description: 'The start date of the time series',
+          },
+          dateTo: {
+            type: GraphQLDateTime,
+            description: 'The end date of the time series',
           },
         },
         async resolve(host, args) {
-          const metrics = await host.getHostMetrics(args?.from, args?.to);
+          const metrics = await host.getHostMetrics(args.dateFrom || args.from, args.dateTo || args.to);
```
SudharakaP, line 93: Given that the arguments `from`, `to`, `dateFrom`, `dateTo` can be null; shouldn't we handle that? Either using the `?` or some other explicit handling? I am not quite understanding why you removed the `?` marks here. 🤔

SudharakaP (commented):
@Betree : Everything looks good except for my small doubt below. 😉

server/graphql/v2/object/Host.js
Betree: If all arguments are missing, `args` will just result in an empty object, it won't be nullish. The `?` were not needed here.

SudharakaP: Thanks for the explanation. 👍🏽

PR #6612, opened 2021-09-21
Authenticate the request to fetch Host report CSV

Betree:
For https://github.com/opencollective/opencollective-rest/pull/390

### DIFF
--- a/server/lib/auth.js
+++ b/server/lib/auth.js
@@ -14,7 +14,8 @@ const minutesToSeconds = minutes => moment.duration({ minutes }).asSeconds();
 export const TOKEN_EXPIRATION_LOGIN = minutesToSeconds(75);
 export const TOKEN_EXPIRATION_CONNECTED_ACCOUNT = daysToSeconds(1);
 export const TOKEN_EXPIRATION_SESSION = daysToSeconds(90);
-export const TOKEN_EXPIRATION_PDF = minutesToSeconds(75);
+export const TOKEN_EXPIRATION_PDF = minutesToSeconds(5);
+export const TOKEN_EXPIRATION_CSV = minutesToSeconds(5);
 
 const ALGORITHM = 'HS256';
 const KID = 'HS256-2019-09-02';
--- a/server/lib/csv.js
+++ b/server/lib/csv.js
@@ -1,12 +1,12 @@
 import config from 'config';
 import moment from 'moment';
 
-// import { TOKEN_EXPIRATION_CSV } from './auth';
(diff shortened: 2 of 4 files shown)


server/lib/auth.js
```diff
@@ -14,7 +14,8 @@ const minutesToSeconds = minutes => moment.duration({ minutes }).asSeconds();
 export const TOKEN_EXPIRATION_LOGIN = minutesToSeconds(75);
 export const TOKEN_EXPIRATION_CONNECTED_ACCOUNT = daysToSeconds(1);
 export const TOKEN_EXPIRATION_SESSION = daysToSeconds(90);
-export const TOKEN_EXPIRATION_PDF = minutesToSeconds(75);
+export const TOKEN_EXPIRATION_PDF = minutesToSeconds(5);
```
Betree, line 17: Also changed the value for PDF service here, it's a single-use token so 5 minutes should be large enough.

server/lib/csv.js
```diff
@@ -1,12 +1,12 @@
...
 import moment from 'moment';
 
-// import { TOKEN_EXPIRATION_CSV } from './auth';
+import { TOKEN_EXPIRATION_CSV } from './auth';
 import { fetchWithTimeout } from './fetch';
 import logger from './logger';
 import { parseToBoolean } from './utils';
 
-export const getTransactionsCsvUrl = async (type, collective, { startDate, endDate, kind } = {}) => {
+export const getTransactionsCsvUrl = (type, collective, { startDate, endDate, kind, add } = {}) => {
```
Betree, line 9: Removed the `async` here as it was crashing the query below

server/lib/csv.js
```diff
@@ -56,10 +64,15 @@ export const getCollectiveTransactionsCsv = async (collective, { startDate, endD
...
   if (parseToBoolean(config.restService.fetchHostTransactionsCsv) === false) {
     return;
   }
 
-  return getTransactionsCsv('hostTransactions', collective, { startDate, endDate });
+  return getTransactionsCsv('hostTransactions', collective, {
+    startDate,
+    endDate,
+    useAdminAccount: true,
+    add: ['accountName', 'oppositeAccountName'],
```
znarf, line 76: I would prefer to have it as defaults on the REST service side. So there is only 1 default format we  promote and spread.

Betree: I moved that to https://github.com/opencollective/opencollective-rest/pull/390/files#diff-275afb3c816085cf369751b24d6b5354d4fc2d4fa194d6a979a9fe1692ca0115R272

PR #6613, opened 2021-09-21
Use new description generation

znarf:
- Remove hardcoded descriptions and use the generator in Sequelize hooks
- Fix a few typos in generator
- Fix directly related tests (snapshots and looking at description string)
- Fix indirectly broken tests (somehow the returned order is changing for Transactions)


server/models/Transaction.js
```diff
@@ -257,6 +258,16 @@ function defineModel() {
       },
 
       hooks: {
+        beforeCreate: async transaction => {
+          if (!transaction.description) {
+            transaction.description = await generateDescription(transaction);
```
Betree, line 263: This centralized approach will make it more difficult to wrap the creation of new transactions in SQL transactions (what was started in https://github.com/opencollective/opencollective-api/pull/5981). We can eventually take the one from `beforeCreate` arguments and pass it down to `generateDescription` but that will make things more complex (what if you're using loaders?).

In this regard, the existing approach of setting the description where the transaction gets created makes more sense as these parts of the code already have the data needed to generate the descriptions, they don't need to preload any additional information.

We can still add some helpers when they're used in multiple places, for example:

```
models.Transaction.create({
  isRefund: true,
  description: generateRefundTransactionDescription(originalTransaction),
})
```

Betree (commented):
Do we want to migrate past data?

znarf:
> Do we want to migrate past data?

For now, I don't think so. Main reason is that the generator is software based and we would need to iterate 1 by 1 on all transactions in the table.

server/models/Transaction.js
znarf: Ok, let's take some time to think about this.

PR #6616, opened 2021-09-21
Stop treating host admins as admins of the collective they're hosting

Betree:
This was introduced in https://github.com/opencollective/opencollective-api/pull/1095, but it's unclear why. We ideally want to be able to set a different set of permissions between host admins and collective admins.

- [x] Remove the code from `models.User`
- [x] Update tests
- [x] Re-enable dedicated test from https://github.com/opencollective/opencollective/issues/4677


znarf (approved):
Good for me!

PR #6617, opened 2021-09-21
 Make Currency Format Precision Consistent in Email Templates

SudharakaP:
Not sure if this is useful or not, but I thought I'll make the currency format precision consistent across all the email templates rather than just fixing the pending financial contribution template only. 😄 

Resolve https://github.com/opencollective/opencollective/issues/4713

### DIFF
--- a/server/lib/handlebars.js
+++ b/server/lib/handlebars.js
@@ -125,7 +125,7 @@ handlebars.registerHelper('currency', (value, props) => {
     return value.toLocaleString(locale, {
       style: 'currency',
       currency,
-      minimumFractionDigits: precision || 0,
+      minimumFractionDigits: precision || 2,
       maximumFractionDigits: !isNil(precision) ? precision : getDefaultCurrencyPrecision(currency),
     });
   })();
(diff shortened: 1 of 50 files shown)


znarf:
Why not make precision=2 the default in the `currency` helper?

SudharakaP:
> Why not make precision=2 the default in the `currency` helper?

That makes more sense. I've gone ahead and made the change so that in the email templates as well as the `formatCurrency` helpers the default precision is set to 2 decimals.

server/lib/handlebars.js
```diff
@@ -125,7 +125,7 @@ handlebars.registerHelper('currency', (value, props) => {
     return value.toLocaleString(locale, {
       style: 'currency',
       currency,
-      minimumFractionDigits: precision || 0,
+      minimumFractionDigits: precision || 2,
```
Betree, line 128: Let's not impact the currencies without decimals

```suggestion
      minimumFractionDigits: precision || getDefaultCurrencyPrecision(currency),
```

SudharakaP: Thanks for pointing it out. Just corrected it. 👍🏽

Betree (approved):

PR #6624, opened 2021-09-23
Support for extra env in development environment

znarf:
I keep editing my `.env` file to switch between different environment variables and it's time to optimize this process for me.

This feature is completely optional but can be used like this.

1) Create an extra `.env` file named `.env.staging`, `.env.production` or `.env.local`
2) Run your `npm run dev staging`,  `npm run dev production` or  `npm run dev local` and this will include the extra env file on top of the default `.env` one.

### DIFF
--- a/server/env.js
+++ b/server/env.js
@@ -1,6 +1,9 @@
+import fs from 'fs';
+import path from 'path';
+
 import debug from 'debug';
 import dotenv from 'dotenv';
-import { get, has } from 'lodash';
+import { get, has, last } from 'lodash';
 
 if (!process.env.NODE_ENV) {
   process.env.NODE_ENV = 'development';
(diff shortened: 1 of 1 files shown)


server/env.js
```diff
@@ -16,6 +19,19 @@ if (!process.env.NODE_CONFIG_ENV) {
 }
 
 dotenv.config();
+
+// Load extra env file on demand
+// `npm run dev staging` / `npm run dev production`
+if (process.env.EXTRA_ENV || process.env.OC_ENV === 'development') {
+  const extraEnv = process.env.EXTRA_ENV || last(process.argv);
+  if (['local', 'staging', 'production'].includes(extraEnv)) {
```
Betree, line 27: Having a limited set is helpful to make this safe, though being able to define custom envs could also be nice:

```bash
$ ls
.env.staging
.env.local
.env.local-prod-copy
...

$ npm run dev local-prod-copy
```

Betree (commented):
Great idea, that'd be helpful in frontend too

server/env.js
znarf: @Betree Do you want me to relax? Should not hurt?

Betree: I think so, especially since the file needs to be prefixed with `.env.`. It seems safe

PR #6640, opened 2021-09-28
Add Support to return transactions view timeseries data

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4658

Related Front-End PR: https://github.com/opencollective/opencollective-frontend/pull/6918


server/graphql/v2/object/Host.js
```diff
@@ -437,6 +472,218 @@ export const Host = new GraphQLObjectType({
...
+          }
+          const contributionsCount = await models.Transaction.count({
+            where,
+          });
+          let contributionAmountOverTime;
+          if (args.timeUnit) {
+            const dateFrom = args.dateFrom ? moment(args.dateFrom).toISOString() : undefined;
+            const dateTo = args.dateTo ? moment(args.dateTo).toISOString() : undefined;
+            contributionAmountOverTime = await sequelize.query(
+              `SELECT DATE_TRUNC('${args.timeUnit}', "createdAt") AS "date", sum(amount) as "amount", "currency"
```
Betree, line 524: I've actually taken this shortcut myself with the other timeseries since timeUnit is a safe enum, but we try to never use template strings in queries because that's an open door for SQL injections. We instead use `replacements` or `bind`:

```suggestion
              `SELECT DATE_TRUNC(:timeUnit, "createdAt") AS "date", sum(amount) as "amount", "currency"
```

```es6
replacements: { timeUnit }
```

I'll take care of updating the ones in `host-metrics`.

SudharakaP: Thanks for pointing it out. I've just refactored the two queries here to make use of `replacements`.

server/graphql/v2/object/Host.js
```diff
@@ -437,6 +472,236 @@ export const Host = new GraphQLObjectType({
...
+            timeUnit: args.timeUnit,
+            nodes: resultsToAmountNode(expenseAmountOverTime),
+          };
+
+          const expensesAmountSum = await models.Transaction.sum('amount', { where });
+          const dailyAverageAmount = expensesCount ? Math.abs(expensesAmountSum) / numberOfDays : 0;
+          return {
+            expensesCount,
+            expenseAmountOverTime,
+            invoicesCount: models.Transaction.count({
```
Betree, line 686: I just discovered today that you can instead return functions that will only run if the field is fetched in the query. Let's do that on all these expensive fields :rocket: :rocket: :rocket: 

```suggestion
            invoicesCount: () => models.Transaction.count({
```

SudharakaP: I've made sure that every field of the `contributionStats` and `expenseStats` are functions. So this should be good now I think. 😉

server/graphql/v2/object/Host.js
```diff
@@ -534,9 +548,68 @@ export const Host = new GraphQLObjectType({
...
+            let contributionAmountOverTime;
+            if (args.timeUnit) {
+              const dateFrom = args.dateFrom ? moment(args.dateFrom).toISOString() : undefined;
+              const dateTo = args.dateTo ? moment(args.dateTo).toISOString() : undefined;
+              contributionAmountOverTime = await sequelize.query(
+                `SELECT DATE_TRUNC(:timeUnit, "createdAt") AS "date", sum(amount) as "amount", "currency"
+                 FROM "Transactions"
+                 WHERE kind = 'CONTRIBUTION'
+                   AND "HostCollectiveId" = :hostCollectiveId
+                   AND type = 'CREDIT'
```
Betree, line 562:
```suggestion
                   AND type = 'CREDIT'
                   AND "deletedAt" IS NULL
```

server/graphql/v2/object/Host.js
```diff
@@ -534,9 +548,68 @@ export const Host = new GraphQLObjectType({
...
+
+              const counts = contributionAmountOverTime.reduce((prev, curr) => {
+                const count = prev.get(curr.date) || 0;
+                prev.set(curr.date.toISOString(), curr.amount + count);
+                return prev;
+              }, new Map());
+
+              contributionAmountOverTime = [...counts].map(([date, amount]) => {
+                return { date, amount, currency: host.currency };
+              });
```
Betree, line 598: I don't understand why we would want to have `amountForThisMonth = SUM(previousMonthsAmounts)`. And even if we wanted that, the resulting metric would be false here since we're not fetching the entire history.

server/graphql/v2/object/Host.js
```diff
@@ -590,14 +668,68 @@ export const Host = new GraphQLObjectType({
...
+                {
+                  type: sequelize.QueryTypes.SELECT,
+                  replacements: {
+                    hostCollectiveId: host.id,
+                    timeUnit: args.timeUnit,
+                    collectiveIds,
+                    dateFrom,
+                    dateTo,
+                  },
+                },
```
Betree, line 703: The query seems 99% similar to the other one, only the `kind` changes. Could we extract that somewhere? Maybe a `getTransactionsTimeseries` in `server/lib/queries.js`?

server/graphql/v2/object/Host.js
```diff
@@ -590,14 +668,68 @@ export const Host = new GraphQLObjectType({
...
+
+              const counts = expenseAmountOverTime.reduce((prev, curr) => {
+                const count = prev.get(curr.date) || 0;
+                prev.set(curr.date.toISOString(), curr.amount + count);
+                return prev;
+              }, new Map());
+
+              expenseAmountOverTime = [...counts].map(([date, amount]) => {
+                return { date, amount: Math.abs(amount), currency: host.currency };
+              });
```
Betree, line 718: Same here, I don't get why we use cumulative amounts.

server/graphql/v2/object/Host.js
SudharakaP: Yes, seems this is a mistake on my part. We don't need this sum. I've corrected that.

server/graphql/v2/object/Host.js
SudharakaP: I've created a new function, `getTransactionsTimeseries` in `server/lib/queries.js` and move this query there. 👍🏽

server/graphql/v2/object/Host.js
```diff
@@ -534,9 +549,47 @@ export const Host = new GraphQLObjectType({
             const collectiveIds = collectives.map(collective => collective.id);
             where.CollectiveId = { [Op.in]: collectiveIds };
           }
+
+          const contributionAmountOverTime = async () => {
+            let contributionAmountOverTime;
+            if (args.timeUnit) {
```
znarf, line 555: Can this be removed as args.timeUnit should be non-null?

server/graphql/v2/object/Host.js
```diff
@@ -590,14 +648,48 @@ export const Host = new GraphQLObjectType({
...
+          let collectiveIds;
           if (args.account) {
             const collectives = await fetchAccountsWithReferences(args.account, { throwIfMissing: true });
             const collectiveIds = collectives.map(collective => collective.id);
             where.CollectiveId = { [Op.in]: collectiveIds };
           }
+
+          const expenseAmountOverTime = async () => {
+            let expenseAmountOverTime;
+            if (args.timeUnit) {
```
znarf, line 660: Non null?

server/graphql/v2/object/Host.js
```diff
@@ -590,14 +648,48 @@ export const Host = new GraphQLObjectType({
...
+                dateFrom,
+                dateTo,
+              );
+              expenseAmountOverTime = expenseAmountOverTime.map(expenseAmount =>
+                convertCurrencyAmount(expenseAmount.currency, host.currency, expenseAmount.date, expenseAmount.amount),
+              );
+              expenseAmountOverTime = await Promise.all(expenseAmountOverTime);
+              expenseAmountOverTime = expenseAmountOverTime.map(({ date, amount }) => {
+                return { date, amount: Math.abs(amount), currency: host.currency };
+              });
```
Betree, line 678: We're mapping the result 2 times, could we do that in one go? And since we do almost the exact same thing for contributions, maybe we could think about a smart helper to reduce the code duplication between them?

```es6
const result = await Promise.all(expenseAmountOverTime.map(async expenseAmount => {
  return {
    amount: Math.abs(await convertCurrencyAmount(...)),
    ...
  }
}))
```

server/lib/queries.js
```diff
@@ -1125,6 +1125,38 @@ const getTaxFormsRequiredForAccounts = async (accountIds = [], year) => {
   return getTaxFormsOverTheLimit(results, 'collectiveId');
 };
 
+/**
+ * Returns the contribution or expense amounts over time.
+ */
+const getTransactionsTimeSeries = async (kind, type, hostCollectiveId, timeUnit, collectiveIds, dateFrom, dateTo) => {
+  return sequelize.query(
+    `SELECT DATE_TRUNC(:timeUnit, "createdAt") AS "date", sum(amount) as "amount", "currency"
```
znarf, line 1133: Better look at the amountInHostCurrency and group by hostCurrency.

server/graphql/v2/object/Host.js
SudharakaP: Corrected this one by moving the logic to a helper function. Also note that I've moved the `Math.abs` to the front-end as I felt that it's the duty of the front end to present the data in the chat as they want; we now just send the expense data points as negative values. 😉

server/graphql/v2/object/Host.js
SudharakaP: Yes, thanks for pointing it out.

server/lib/queries.js
SudharakaP: Corrected this one to include the `amountInHostCurrency` and `hostCurrency`.

znarf:
Do we want actually want the amount or the net amount here (after or before considering payment processor fees)?

server/graphql/v2/object/Host.js
```diff
@@ -64,6 +67,46 @@ const getNumberOfDays = (startDate, endDate, host) => {
   return days(since, endDate || undefined);
 };
 
+const convertCurrencyAmount = async (fromCurrency, toCurrency, date, amount) => {
+  const fxRate = await getFxRate(fromCurrency, toCurrency, date);
+  const convertedAmount = Math.round(amount * fxRate);
+  return { date: date, amount: convertedAmount };
+};
```
znarf, line 74: This is a bit reinventing the wheel, `convertToCurrency` in `lib/currency` should be enough?

server/graphql/v2/object/Host.js
```diff
@@ -64,6 +67,46 @@ const getNumberOfDays = (startDate, endDate, host) => {
...
+    dateFrom,
+    dateTo,
+  );
+  const dataPointsInHostCurrency = transactionDatePoints.map(async contributionAmount => {
+    const contributionAmountInHostCurrency = await convertCurrencyAmount(
+      contributionAmount.currency,
+      host.currency,
+      contributionAmount.date,
+      contributionAmount.amount,
+    );
```
znarf, line 100: Maybe we should be confident that  the hostCurrency in the ledger is the host.currency? And no need to convert? Could always throw an Error if that's not the case.

SudharakaP: I've simplified the logic to remove this conversion as I think it's not longer needed as we are now getting the `amountInHostCurrency`. This part I only added when we were using the previous query which had `amount` instead of `amountInHostCurrency`.

SudharakaP:
> Do we want actually want the amount or the net amount here (after or before considering payment processor fees)?

My thinking is here we need the net amount. Since according to current design there's no indication anywhere that we are considering payment processor fees. It looks like a general overview to the user and the tool-tip also says, `Transactions related to contributions and expenses.` which to me suggests that. But I am open to suggestions. 🤔 

cc: @Betree

Betree (approved):
Looks good to me. I'm up for merging, test with live data and iterate if needed

PR #6647, opened 2021-09-29
HostMetrics: Use replacements for timeseries

Betree:
As reported in https://github.com/opencollective/opencollective-api/pull/6640#discussion_r718296524

Cleaner & safer.

### DIFF
--- a/server/lib/host-metrics.js
+++ b/server/lib/host-metrics.js
@@ -60,20 +60,14 @@ function computeDates(startDate, endDate) {
   return { startDate: startDate.toISOString(), endDate: endDate.toISOString() };
 }
 
-const getGroupTimeUnitFragments = (groupTimeUnit, table) => {
-  if (!groupTimeUnit) {
-    return { select: '', groupBy: '', orderBy: '' };
-  } else {
-    return {
-      select: `, DATE_TRUNC('${groupTimeUnit}', ${table}."createdAt") AS "date"`,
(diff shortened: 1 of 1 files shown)


server/lib/host-metrics.js
```diff
@@ -192,16 +186,16 @@ GROUP BY t1."hostCurrency"`,
 
 export async function getHostFeesTimeSeries(host, { startDate, endDate, timeUnit } = {}) {
   const newResults = await sequelize.query(
-    `SELECT SUM(t1."amountInHostCurrency") as "_amount", t1."hostCurrency" as "_currency", DATE_TRUNC('${timeUnit}', t1."createdAt") as "date"
+    `SELECT SUM(t1."amountInHostCurrency") as "_amount", t1."hostCurrency" as "_currency", DATE_TRUNC(:timeUnit, t1."createdAt") as "date"
```
SudharakaP, line 189: @Betree : Sorry I missed providing this review, but in this function we name variables as `t1` but there's no `t2` and such. Maybe it's better to just name the variable to `t` instead? Same applies to `getHostFeeShare` and `getHostFeeShareTimeSeries` I think. 😉  

Except for that everything else looks good to me. 😄

Betree: I think that part hasn't been touched for some time, I'm not sure why they were named that way. We could indeed rename.

PR #6651, opened 2021-09-30
Don't authenticate users if no profile available

Betree:
Though we'll always try to avoid it, it happens that we end up with a user whose `CollectiveId` is null or soft-deleted. In such cases we currently react poorly, allowing the user to log in but failing on the first query / mutation that will do a `remoteUser.getCollective()`.

This change will make authentication fail if we cannot retrieve the collective for user.

### DIFF
--- a/test/server/routes/users.test.js
+++ b/test/server/routes/users.test.js
@@ -93,7 +93,7 @@ describe('server/routes/users', () => {
     });
     it('should fail if expired token is provided', async () => {
       // Given a user and an authentication token
-      const user = await models.User.create({ email: 'test@mctesterson.com' });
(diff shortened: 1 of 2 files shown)


test/server/routes/users.test.js
```diff
@@ -93,7 +93,7 @@ describe('server/routes/users', () => {
     });
     it('should fail if expired token is provided', async () => {
       // Given a user and an authentication token
-      const user = await models.User.create({ email: 'test@mctesterson.com' });
+      const user = await fakeUser({ email: 'test@mctesterson.com' });
```
Betree, line 96: These were created without collective, which triggered the new check

znarf (approved):
Perfect.

PR #6659, opened 2021-10-01
Implement some optional resolvers

Betree:
An alternative, simpler approach to what was proposed in https://github.com/opencollective/opencollective/issues/4586#issuecomment-912416325: if you return a function for one of your GraphQL object fields, it will automatically be resolved by GraphQL. The function doesn't run if the field is not fetched.


znarf (approved):
Looks good!

That is obvious now that I'm looking at it.

PR #6660, opened 2021-10-01
refact(Wise): unscheduleExpenseForPayment to fetch batch version from Wise

kewitz:
Fixes bug where the version number we persisted on our side does not match the latest version of the Batch Group.
```
opencollective-prod-api app/web.1 error: Error: Group version conflict
```


Betree (approved):
I don't have the details about the original error, but that looks good

PR #6683, opened 2021-10-07
Add Collective Filter Support to Host Fees Section

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4657

Front End PR: https://github.com/opencollective/opencollective-frontend/pull/6967

### DIFF
--- a/server/graphql/v2/object/Host.js
+++ b/server/graphql/v2/object/Host.js
@@ -97,6 +97,10 @@ export const Host = new GraphQLObjectType({
       hostMetrics: {
         type: new GraphQLNonNull(HostMetrics),
         args: {
+          account: {
+            type: new GraphQLList(new GraphQLNonNull(AccountReferenceInput)),
+            description: 'A collection of accounts for which the metrics should be returned.',
+          },
           from: {
             type: GraphQLString,
             description: "Inferior date limit in which we're calculating the metrics",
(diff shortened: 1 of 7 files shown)


server/graphql/v2/object/Host.js
```diff
@@ -117,7 +121,15 @@ export const Host = new GraphQLObjectType({
           },
         },
         async resolve(host, args) {
-          const metrics = await host.getHostMetrics(args.dateFrom || args.from, args.dateTo || args.to);
+          let collectiveIds;
+          if (args.account) {
+            const collectives = await fetchAccountsWithReferences(args.account, {
+              throwIfMissing: true,
```
znarf, line 127: Maybe we don't have to throw an error here?

SudharakaP: Thanks, makes sense to not throw by errors here if one collective id/slug is missing. 🤔 I've corrected this. 👍🏽

server/lib/budget.js
```diff
@@ -134,12 +134,18 @@ export async function getTotalNetAmountReceivedAmount(collective, { startDate, e
...
-export async function getTotalMoneyManagedAmount(host, { startDate, endDate, currency, version } = {}) {
+export async function getTotalMoneyManagedAmount(host, { startDate, endDate, collectiveIds, currency, version } = {}) {
   version = version || host.settings?.budget?.version || 'v1';
   currency = currency || host.currency;
 
-  const hostedCollectives = await host.getHostedCollectives();
-  const ids = hostedCollectives.map(c => c.id);
+  let hostedCollectives, ids;
+  if (collectiveIds?.length > 0) {
+    ids = collectiveIds;
```
znarf, line 143: No validation of the ids here. On purpose?

SudharakaP: Since these `id`s are used to get the sum of all transactions at https://github.com/opencollective/opencollective-api/blob/46013b0d7e33baf7640c98a5ab4fb4b515287d68/server/lib/budget.js#L189

If an `id` cannot be found/filtered on then we just have to ignore it in my understanding and just filter on the remaining `id`s that are passed. This is the reason for not having any validation here. But we can put a validation just to print out an info log but not sure if that's really needed. 🤔

SudharakaP:
As discussed with @Betree I am merging this one along with the front-end PR.

PR #6728, opened 2021-10-20
TransactionInvoice: Load correct host

Betree:
Fix https://github.com/opencollective/opencollective/issues/4496


server/graphql/v1/queries.js
```diff
@@ -143,16 +147,32 @@ const queries = {
       const fromCollectiveId = transaction.paymentMethodProviderCollectiveId();
 
       // Load transaction host
-      if (transaction.HostCollectiveId) {
-        transaction.host = await transaction.getHostCollective();
```
Betree, line 147: We introduced https://github.com/opencollective/opencollective/issues/5009 here by not adapting the fields that were depending on it below (`get(transaction.host, 'settings.invoiceTitle'),`)

PR #6729, opened 2021-10-20
PDFReceipts: Block unauthenticated requests

Betree:
Requires https://github.com/opencollective/opencollective-frontend/pull/7028


Betree:
We should now be ready for this, will try to deploy it soon

PR #6766, opened 2021-11-01
Implement merchant id  in for transactions GraphQL

znarf:
@Betree great if you can have a look.

I'm wondering what kind of permission we should put on this field and how to implement it.


Betree (commented):
For the permissions, I would say only return for payment method admin or host admin?

PR #6774, opened 2021-11-02
Expense payee total payout sum

juliekitty:
Related to https://github.com/opencollective/opencollective/issues/4491

Add totalPaidExpense in stats


server/lib/budget.js
```diff
@@ -107,6 +107,30 @@ export function getTotalAmountReceivedAmount(collective, { startDate, endDate, c
...
+  if (startDate) {
+    where.createdAt = where.createdAt || {};
+    where.createdAt[Op.gte] = startDate;
+  }
+  if (endDate) {
+    where.createdAt = where.createdAt || {};
+    where.createdAt[Op.lt] = endDate;
+  }
+  const result = await models.Expense.findOne({
+    attributes: [[sequelize.fn('COALESCE', sequelize.fn('SUM', sequelize.col('amount')), 0), 'amount']],
```
znarf, line 127: Add currency and group by currency.

server/lib/budget.js
```diff
@@ -107,6 +107,30 @@ export function getTotalAmountReceivedAmount(collective, { startDate, endDate, c
...
+    where: where,
+    raw: true,
+  });
```
znarf, line 130: Iterate over the results and do currency conversion to consolidate on one currency.

server/lib/budget.js
```diff
@@ -107,6 +107,30 @@ export function getTotalAmountReceivedAmount(collective, { startDate, endDate, c
...
+  return result.amount;
```
znarf, line 131: Return amount AND currency.

server/graphql/v2/object/AccountStats.js
```diff
@@ -80,6 +82,36 @@ export const AccountStats = new GraphQLObjectType({
...
+          dateTo: {
+            type: GraphQLDateTime,
+            description: 'Calculate total amount received before this date',
+          },
+          dateFrom: {
+            type: GraphQLDateTime,
+            description: 'Calculate total amount received after this date',
+          },
+          currency: {
+            type: GraphQLString,
```
znarf, line 102: Use "Enum" currency instead.

server/graphql/v2/object/AccountStats.js
```diff
@@ -80,6 +82,36 @@ export const AccountStats = new GraphQLObjectType({
...
+            description: 'An optional currency. If not provided, will use the collective currency.',
+          },
+        },
+        async resolve(collective, args) {
+          return await collective.getTotalAmountPaidExpensesAmount({
```
znarf, line 107: `return await` can be just `return`.

znarf: `getTotalAmountPaidExpensesAmount` is a weird name with 2 times "amount" in it. Update with something else?

znarf: I actually see that "getTotalAmountReceivedAmount" is already existing so that's fine ... We should do better later.

server/lib/budget.js
```diff
@@ -107,6 +107,42 @@ export function getTotalAmountReceivedAmount(collective, { startDate, endDate, c
...
+
+  const results = await models.Expense.findAll({
+    attributes: ['currency', [sequelize.fn('COALESCE', sequelize.fn('SUM', sequelize.col('amount')), 0), 'amount']],
+    where: where,
+    group: 'currency',
+    raw: true,
+  });
+
+  let total = 0;
+  for (const result of Object.values(results)) {
```
znarf, line 137: Don't see the point of `Object.values(results)`, `results is an array and we should just use it like this: `results`

server/models/Collective.js
```diff
@@ -2468,6 +2469,14 @@ function defineModel() {
     return getTotalAmountReceivedAmount(this, options).then(result => result.value);
   };
 
+  Collective.prototype.getTotalAmountPaidExpensesAmount = function (options) {
+    return getTotalAmountPaidExpenses(this, options);
+  };
+
+  Collective.prototype.getTotalAmountPaidExpenses = function (options) {
```
znarf, line 2476: getTotalPaidExpenses

server/models/Collective.js
```diff
@@ -2468,6 +2469,14 @@ function defineModel() {
     return getTotalAmountReceivedAmount(this, options).then(result => result.value);
   };
 
+  Collective.prototype.getTotalAmountPaidExpensesAmount = function (options) {
```
znarf, line 2472: getTotalPaidExpensesAmount

server/graphql/v2/object/AccountStats.js
```diff
@@ -80,6 +82,36 @@ export const AccountStats = new GraphQLObjectType({
           return collective.getTotalAmountReceivedAmount({ kind, startDate: args.dateFrom, endDate: args.dateTo });
         },
       },
+      totalPaidExpense: {
```
znarf, line 85: We need some consistency on plural vs singular. If other function say "paidExpenses", we should be plural here. Or the opposite.

server/graphql/v2/object/AccountStats.js
znarf: In the end, whatever the other, `getTotalPaidExpensesAmount` looks clearer.

znarf (changes requested):
Looks generally good! Some changes requested.

znarf (approved):
Good job!

PR #6779, opened 2021-11-02
Allow User to Edit Expense Even if Draft Key was Passed in URL

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4902

Front End PR: https://github.com/opencollective/opencollective-frontend/pull/7081

### DIFF
--- a/server/graphql/v2/input/ExpenseUpdateInput.ts
+++ b/server/graphql/v2/input/ExpenseUpdateInput.ts
@@ -67,9 +67,5 @@ export const ExpenseUpdateInput = new GraphQLInputObjectType({
       type: LocationInput,
       description: 'The address of the payee',
     },
-    status: {
-      type: GraphQLString,
-      description: 'The status of the expense such as DRAFT, PENDING etc',
-    },
   }),
 });
--- a/server/graphql/v2/input/ExpenseUpdateInput.ts
+++ b/server/graphql/v2/input/ExpenseUpdateInput.ts
@@ -67,5 +67,9 @@ export const ExpenseUpdateInput = new GraphQLInputObjectType({
       type: LocationInput,
       description: 'The address of the payee',
     },
+    status: {
+      type: GraphQLString,
+      description: 'The status of the expense such as DRAFT, PENDING etc',
+    },
   }),
 });
--- a/server/graphql/v2/mutation/ExpenseMutations.ts
+++ b/server/graphql/v2/mutation/ExpenseMutations.ts
@@ -135,7 +135,7 @@ const expenseMutations = {
         fromCollective: payeeExists && (await fetchAccountWithReference(expense.payee, { throwIfMissing: true })),
       };
 
-      if (args.draftKey && expense.status === expenseStatus.DRAFT) {
+      if (args.draftKey) {
         // It is a submit on behalf being completed
         const expenseId = getDatabaseIdFromExpenseReference(args.expense);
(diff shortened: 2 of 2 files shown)


SudharakaP:
Correct me if I am wrong, but I don't see the necessity for this condition. The condition is only triggered when a `draftKey` is passed. If a `draftKey` is not passed the user can edit the expense. Ideally I think we should give the user the ability to edit he expense even if the draft key is passed or not as the user might be accessing the (possibly already submitted and pending) expense using the invite link. 🤔

znarf (changes requested):
I don't think we should enter the `if (args.draftKey)` condition if the expense is not a DRAFT anymore.

SudharakaP:
> I don't think we should enter the `if (args.draftKey)` condition if the expense is not a DRAFT anymore.

@znarf : Makes sense and feel that's a better solution. I've corrected the code to add that condition. 🤔

server/graphql/v2/input/ExpenseUpdateInput.ts
```diff
@@ -67,5 +67,9 @@ export const ExpenseUpdateInput = new GraphQLInputObjectType({
       type: LocationInput,
       description: 'The address of the payee',
     },
+    status: {
+      type: GraphQLString,
+      description: 'The status of the expense such as DRAFT, PENDING etc',
+    },
```
znarf, line 73: Why is this needed? The type look wrong but instead of fixing it I'm wondering if this can be removed.

SudharakaP: This is needed because now we are passing the status as an arg for the `ExpenseUpdateInput` object. Specifically at, 

https://github.com/opencollective/opencollective-frontend/blob/db5ddcfcc0fd0bee15af15d73268116562a509b8/components/expenses/ExpenseForm.js#L117

We can probably remove this if we calculate the `existingExpense` outside of the `if` condition; 

https://github.com/opencollective/opencollective-api/blob/b9cdeb9b9dcc438a4d508769cc76a0a4934110e6/server/graphql/v2/mutation/ExpenseMutations.ts#L138-L143

However I didn't do this as passing it from the front-end seems straightforward and doesn't make an unnecessary db call. Do you have a better alternative? 🤔

znarf: Can we simply not pass it? 

@Betree what do you think?

Betree: I don't get why we'd need to pass a `status`, but it definitely looks wrong:
- What does it mean if someone passes `HELLO` as a status? (the current field is not an enum)
- What does it mean if someone passes `PAID` as a status?

`status` is not something we want people to be able to edit, it's an indication of the result from other mutations (`processExpense`). We need to find a better API there.

Betree: Can't we just omit `draftKey` if it's not a draft?

SudharakaP: That makes sense. I've refactored the front-end pr to omit the `draftKey` if it's not a draft. Thanks for the idea; that seems such a simple solution. 😉

SudharakaP:
@znarf : I am closing this PR as this is no longer required; only the front-end PR is required as per the above comment; https://github.com/opencollective/opencollective-api/pull/6779#discussion_r743803180

PR #6783, opened 2021-11-03
Add webhook for Stripe 'issuing_transaction.created' event

Forster-Groux:
Part of opencollective/opencollective#4807

### DIFF
--- a/server/paymentProviders/stripe/index.js
+++ b/server/paymentProviders/stripe/index.js
@@ -210,6 +210,8 @@ export default {
   webhook: request => {
     const requestBody = request.body;
 
+    debug(`Stripe webhook event received : ${request.rawBody}`);
+
     // Stripe sends test events to production as well
     // don't do anything if the event is not livemode
     // NOTE: not using config.env because of ugly tests
--- a/server/paymentProviders/stripe/index.js
+++ b/server/paymentProviders/stripe/index.js
@@ -14,6 +14,7 @@ import models from '../../models';
 
 import alipay from './alipay';
 import creditcard from './creditcard';
+import { processTransaction } from './virtual-cards';
 
 const debug = debugLib('stripe');
 
@@ -206,13 +207,20 @@ export default {
     }
   },
 
-  webhook: requestBody => {
+  webhook: request => {
+    const requestBody = request.body;
+
     // Stripe sends test events to production as well
     // don't do anything if the event is not livemode
     // NOTE: not using config.env because of ugly tests
     if (process.env.OC_ENV === 'production' && !requestBody.livemode) {
       return Promise.resolve();
     }
+
+    if (requestBody.type === 'issuing_transaction.created') {
+      return processTransaction(requestBody.data.object, request.headers['stripe-signature'], request.rawBody);
+    }
+
     /**
      * We check the event on stripe directly to be sure we don't get a fake event from
      * someone else
--- a/server/paymentProviders/stripe/virtual-cards.js
+++ b/server/paymentProviders/stripe/virtual-cards.js
@@ -1,6 +1,13 @@
 import { omit } from 'lodash';
 import Stripe from 'stripe';
 
+import activities from '../../constants/activities';
+import { types as CollectiveTypes } from '../../constants/collectives';
+import ExpenseStatus from '../../constants/expense_status';
+import ExpenseType from '../../constants/expense_type';
+import { TransactionKind } from '../../constants/transaction-kind';
+import { getFxRate } from '../../lib/currency';
+import logger from '../../lib/logger';
 import models from '../../models';
 
 export const assignCardToCollective = async (cardNumber, expireDate, cvv, collectiveId, host, userId) => {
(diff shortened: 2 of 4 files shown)


server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -51,3 +58,173 @@ export const assignCardToCollective = async (cardNumber, expireDate, cvv, collec
 
   return await models.VirtualCard.create(cardData);
 };
+
+export const processTransaction = async (stripeEvent, stripeSignature) => {
+  const stripeTransaction = stripeEvent.data.object;
+
+  const virtualCard = await models.VirtualCard.findOne({
+    where: {
+      id: stripeTransaction.card,
```
znarf, line 67: I don't get how this works. On our side, id is an uuid that we generate ourselves?

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -51,3 +58,173 @@ export const assignCardToCollective = async (cardNumber, expireDate, cvv, collec
...
+
+  let expense;
+  try {
+    const slug = toString(stripeTransaction['merchant_data']['network_id']).toLowerCase();
+    const [vendor] = await models.Collective.findOrCreate({
+      where: { slug },
+      defaults: { name: stripeTransaction['merchant_data']['name'], type: CollectiveTypes.VENDOR },
+    });
+
+    const hostCurrencyFxRate = await getFxRate('USD', host.currency);
```
znarf, line 141: Do we assume that all incoming transactions will be in dollars? Should we already prepare for the case where it's not?

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -51,3 +58,173 @@ export const assignCardToCollective = async (cardNumber, expireDate, cvv, collec
...
+      data: { token: stripeTransaction.id },
+    },
+  });
+  if (existingExpense) {
+    logger.warn(`Virtual Card charge already reconciled, ignoring it: ${stripeTransaction.id}`);
+    return;
+  }
+
+  const UserId = virtualCard.UserId;
+  const isRefund = amount < 0;
```
znarf, line 117: We know that this is how privacy works. But, is it the same with Stripe for Refunds? If it's not, we could be implementing code that is ultimately not needed or need to be different.

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -51,3 +58,173 @@ export const assignCardToCollective = async (cardNumber, expireDate, cvv, collec
...
+    if (existingTransaction) {
+      logger.warn(`Virtual Card refund already reconciled, ignoring it: ${stripeTransaction.id}`);
+      return;
+    }
+  }
+
+  let expense;
+  try {
+    const slug = toString(stripeTransaction['merchant_data']['network_id']).toLowerCase();
+    const [vendor] = await models.Collective.findOrCreate({
```
znarf, line 136: Why would this be returned in an array?

server/paymentProviders/stripe/virtual-cards.js
Forster-Groux: On our side, id is currently set as the stripe card id or the privacy card token

server/paymentProviders/stripe/virtual-cards.js
Forster-Groux: Transaction are in the currency of the card, right now we can only have USD cards. I think we should extend this behavior for other currency but in other PR.

server/paymentProviders/stripe/virtual-cards.js
Forster-Groux: Wow good catch, I assume it was the same but it is not ! I will modify it accordingly.

server/paymentProviders/stripe/virtual-cards.js
Forster-Groux: `findOrCreate` return an instance with the object and a boolean. This is the javascript syntax equivalent of :
```
const sequelizeResult = await models.Collective.findOrCreate(...);
const vendor = sequelizeResult[0];
```
See https://sequelize.org/master/manual/model-querying-finders.html

server/paymentProviders/stripe/virtual-cards.js
znarf: Ok!

server/paymentProviders/stripe/virtual-cards.js
znarf: Are you confident that transaction made in EUR for those cards will come back to us in USD if the card is USD?

Anyway, no problem, this can be handled later

server/paymentProviders/stripe/virtual-cards.js
znarf: Right.

server/paymentProviders/stripe/virtual-cards.js
Forster-Groux: Yes I am confident.

server/paymentProviders/stripe/index.js
```diff
@@ -206,13 +207,20 @@ export default {
     }
   },
 
-  webhook: requestBody => {
+  webhook: request => {
+    const requestBody = request.body;
+
```
znarf, line 212: Would be great to add some optional logging of all received webhook requests, using the debug library.

znarf (commented):
It's still a draft. What's missing to merge this PR?

server/paymentProviders/stripe/index.js
```diff
@@ -206,13 +207,22 @@ export default {
...
+    debug('stripe event triggered', requestBody.id);
```
znarf, line 213: I was more thinking about a FULL log of the event `JSON.stringify(requestBody)`;

```suggestion
    debug(`Stripe Webhook Event received: ${JSON.stringify(requestBody)}`);
```

znarf (approved):

znarf:
Let's try this.

PR #6786, opened 2021-11-04
Add support to return the total money managed for host over time

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4606

Related Front End PR: https://github.com/opencollective/opencollective-frontend/pull/7071


server/lib/host-metrics.js
```diff
@@ -258,6 +258,30 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
...
 }
 
+export async function getTotalMoneyManagedTimeSeries(host, { startDate, endDate, timeUnit } = {}) {
+  const results = await sequelize.query(
+    `SELECT SUM(t1."amountInHostCurrency") as "_amount", t1."hostCurrency" as "_currency", DATE_TRUNC(:timeUnit, t1."createdAt") as "date"
+FROM "Transactions" as t1
+WHERE t1."HostCollectiveId" = :HostCollectiveId
+AND t1."kind" != 'PLATFORM_FEE'
+AND t1."kind" != 'PLATFORM_TIP'
+AND t1."kind" != 'HOST_FEE'
```
Betree, line 268: Why filter these? Does it match the results from https://github.com/opencollective/opencollective-api/blob/9588f9e3b746ec25796a5877810c1b1be82232ed/server/lib/budget.js#L137? 

Can we add a test for this, ideally comparing the results of `getTotalMoneyManagedAmount` and `getTotalMoneyManagedTimeSeries` to make sure they correspond?

server/graphql/v2/object/HostMetricsTimeSeries.ts
```diff
@@ -51,5 +51,14 @@ export const HostMetricsTimeSeries = new GraphQLObjectType({
         return { dateFrom, dateTo, timeUnit, nodes: resultsToAmountWithSettlementNode(results) };
       },
     },
+    totalMoneyManagedProgress: {
```
znarf, line 54: ..Progress is a weird name for the property.

SudharakaP: Change the naming to `totalMoneyManagedTimeSeries`

server/lib/host-metrics.js
SudharakaP: I've added a test and also did some minor adjustments; previously we had a mistake when calculating this stat since we where summing up both credits and debits. 🤯

server/graphql/v2/object/HostMetricsTimeSeries.ts
znarf: Other `TimeSeries` stats don't have a suffix `TimeSeries` so I don't see why this particular stat deserve a suffix.

SudharakaP: @znarf : You are correct, we don't need the suffix; I've just made it `totalMoneymanaged` so that it's consistent.

Betree (changes requested):

server/lib/host-metrics.js
```diff
@@ -258,6 +258,31 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
...
+GROUP BY t1."netAmountInCollectiveCurrency", t1."hostCurrency", DATE_TRUNC(:timeUnit, t1."createdAt")
+ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
+    {
+      replacements: { ...computeDates(startDate, endDate), timeUnit },
+      type: sequelize.QueryTypes.SELECT,
+    },
+  );
+
+  const timeSeries = await convertCurrencyForTimeSeries(results, host.currency);
+  const mergedTimeSeries = [...timeSeries.map(point => ({ ...point, amount: Math.abs(point.amount) }))];
```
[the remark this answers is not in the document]
SudharakaP: Ops, this is a mistake; I've corrected it. 👍🏽

server/lib/host-metrics.js
```diff
@@ -258,6 +258,31 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
...
+
+  return orderBy(mergedTimeSeries, 'date');
```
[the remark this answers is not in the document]
SudharakaP: Thanks for pointing it out. Corrected. 👍🏽

test/server/graphql/v2/query/HostMetricsQuery.test.js
```diff
@@ -99,6 +116,17 @@ describe('server/graphql/v2/query/HostMetricsQuery', () => {
...
+
+      it('correctly calculates totalMoneyManaged', async () => {
+        const dateFrom = new Date('2021-01-01').toISOString();
+        const dateTo = new Date().toISOString();
+        const variables = { slug: host.slug, dateFrom, dateTo };
+        const queryResponse = await graphqlQueryV2(hostMetricsQuery, variables);
+        const hostMetrics = queryResponse.data.host.hostMetrics;
+        const hostMetricsTimeSeriesNodes = queryResponse.data.host.hostMetricsTimeSeries.totalMoneyManaged.nodes;
+        const totalMoneyManaged = sumBy(hostMetricsTimeSeriesNodes, node => node.amount.valueInCents);
+        expect(hostMetrics.totalMoneyManaged.valueInCents).to.equal(totalMoneyManaged);
```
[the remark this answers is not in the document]
SudharakaP: This one I am not exactly sure tbh. 🤔 My understanding is that the `hostMetrics.totalMoneyManaged` gives the cumulative amount of the money managed by the host. For example see,

https://github.com/opencollective/opencollective-api/blob/e0a67eb968f94357da3b1257e6d5c4d5879fd1d5/server/lib/budget.js#L192-L198

where the total money managed calculation is done. This seems to be a sum of all the transactions. This is why I thought that those two should match. But I understand your point where the money managed should be how much money is managed currently by the host (for the given month). But in this case I think we need to probably should change the heading; `Total Money Managed` or indicate somewhere that it really reflects the amount of money managed by the host for the given month? 🤔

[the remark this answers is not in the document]
Betree: Yes, to get the amount managed for a given date you'll need the query to do something like:

```
SUM(credits) - SUM(debits)
WHERE "createdAt" <= "myDate"
```

Your approach was indeed not crazy in that sense, but since you were filtering on `createdAt` transactions created before `startDate` were not included either so we always considered that you start from $0 on `startDate`. @znarf can maybe help you to build this one.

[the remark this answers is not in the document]
znarf: With a simplistic approach, sounds like a somewhat intensive thing to compute total money managed over time (iterating on dates on computing the balance).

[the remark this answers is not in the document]
SudharakaP: @Betree @znarf : For this one, I've adjusted the query and the tests so that we only consider the following transaction kinds; 

https://github.com/opencollective/opencollective-api/blob/902d7b7914e18dc37891a10628f00881e80d0a84/server/lib/host-metrics.js#L269

My thinking is we should not consider things like platform tips and payment processor fee here. Let me know hoe it looks. 

@Betree : So for the tests I've adjusted them such that I always pass the `startDate` as the `host.createDate` so that we have this type of query which makes sense. 

```
SUM(credits) - SUM(debits)
WHERE "createdAt" <= "myDate"
```

However I feel that this might create confusion for users in the front-end. The users have a period filter and ideally to get the stats right we have to ignore the `startDate` and only consider the `endDate`. Do you think we need some other design change on the front-end so that users know exactly that in the money managed section we don't consider the `startDate`? 🤔

server/lib/host-metrics.js
```diff
@@ -258,6 +258,29 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
...
 }
 
+export async function getTotalMoneyManagedTimeSeries(host, { startDate, endDate, timeUnit } = {}) {
+  const hostedCollectives = await host.getHostedCollectives();
+  const ids = hostedCollectives.map(c => c.id);
+
+  const results = await sequelize.query(
+    `SELECT t1."netAmountInCollectiveCurrency" as "_amount", t1."hostCurrency" as "_currency", DATE_TRUNC(:timeUnit, t1."createdAt") as "date"
+FROM "Transactions" as t1
+WHERE t1."CollectiveId" IN (${ids})
```
Betree, line 268: Variables need to be injected using `replacements`

server/lib/host-metrics.js
```diff
@@ -258,6 +258,29 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
   return orderBy(mergedTimeSeries, 'date');
 }
 
+export async function getTotalMoneyManagedTimeSeries(host, { startDate, endDate, timeUnit } = {}) {
+  const hostedCollectives = await host.getHostedCollectives();
+  const ids = hostedCollectives.map(c => c.id);
+
+  const results = await sequelize.query(
+    `SELECT t1."netAmountInCollectiveCurrency" as "_amount", t1."hostCurrency" as "_currency", DATE_TRUNC(:timeUnit, t1."createdAt") as "date"
```
znarf, line 266: Everything should to be calculate in hostCurrency so we should not look at netAmountInCollectiveCurrency, but amountInHostCurrency.

server/lib/host-metrics.js
znarf, line 268: We need to add a condition to be sure that the Transactions are accounted on the Host (`HostCollectiveId`)

server/graphql/v2/object/HostMetricsTimeSeries.ts
```diff
@@ -51,5 +51,14 @@ export const HostMetricsTimeSeries = new GraphQLObjectType({
         return { dateFrom, dateTo, timeUnit, nodes: resultsToAmountWithSettlementNode(results) };
       },
     },
+    totalMoneyManaged: {
+      type: new GraphQLNonNull(TimeSeriesAmount),
+      description: 'History of the total money managed by this host',
+      resolve: async ({ host, dateFrom, dateTo, timeUnit }) => {
```
Betree, line 57: Don't we want the ability to filter by `accounts`?

server/lib/host-metrics.js
```diff
@@ -258,6 +258,29 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
...
+FROM "Transactions" as t1
+WHERE t1."CollectiveId" IN (${ids})
+AND t1."kind" IN ('CONTRIBUTION', 'EXPENSE', 'ADDED_FUNDS', 'BALANCE_TRANSFER')
```
znarf, line 269: I don't see the point of filtering the kinds.

Betree (commented):
We've reviewed the code with @znarf and it's not far from being correct. The main problem is what was mentioned in https://github.com/opencollective/opencollective-api/pull/6786/files#r748098709: if we just filter the transactions with `startDate > :startDate`, the balance at `startDate` will be considered `0`, which will usually be incorrect.

The next steps could be:
- Get the results like what we're doing now (just fix the query with the feedback provided above)
- Compute the balance at `startDate`
- Add this balance to all amounts

server/graphql/v2/object/HostMetricsTimeSeries.ts
SudharakaP: Yes, somehow we've missed this one. I've added support for filter by collective. 👍🏽 

I think we are missing this on the host fee chart as well. I'll open a new issue for that.

server/lib/host-metrics.js
SudharakaP: Done.

SudharakaP: Done. 👍🏽

server/lib/host-metrics.js
```diff
@@ -258,27 +259,43 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
...
 }
 
-export async function getTotalMoneyManagedTimeSeries(host, { startDate, endDate, timeUnit } = {}) {
+export async function getTotalMoneyManagedTimeSeries(
+  host,
+  { startDate, endDate, collectiveIds = null, timeUnit } = {},
+) {
   const hostedCollectives = await host.getHostedCollectives();
-  const ids = hostedCollectives.map(c => c.id);
+  if (!collectiveIds) {
```
znarf, line 267: If you only need `hostedCollectives` when `collectiveIds` is falsy, then only fetch it when that's the case, not before.

server/lib/host-metrics.js
```diff
@@ -258,27 +259,43 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
...
   );
 
+  const balanceAtStartDate = await getTotalMoneyManagedAmount(host, {
+    endDate: startDate,
+    collectiveIds,
+    currency: host.currency,
+  });
   const timeSeries = await convertCurrencyForTimeSeries(results, host.currency);
-  return [...timeSeries.map(point => ({ ...point, amount: point.amount }))];
+  return timeSeries.map(point => ({ ...point, amount: point.amount + balanceAtStartDate.value }));
```
znarf, line 298: My understanding is that the amounts need to be added together:

```
January balance: balanceAtStartDate.value + point.amount
February balance: balanceAtStartDate.value + January amount  + point.amount
March balance: balanceAtStartDate.value + February amount  + point.amount
...
```

Or:

```
January balance: balanceAtStartDate.value + point.amount
February balance: January balance + point.amount
March balance: Februarybalance  + point.amount
...
```

server/lib/host-metrics.js
SudharakaP: Ops, seems I've missed this one. Corrected it.

server/lib/host-metrics.js
SudharakaP: Ah, thanks; I had a feeling there's still something missing here; this is it. I've corrected it. Let me know if you see any further issues.

SudharakaP:
> We've reviewed the code with @znarf and it's not far from being correct. The main problem is what was mentioned in https://github.com/opencollective/opencollective-api/pull/6786/files#r748098709: if we just filter the transactions with `startDate > :startDate`, the balance at `startDate` will be considered `0`, which will usually be incorrect.
> 
> The next steps could be:
> 
>     * Get the results like what we're doing now (just fix the query with the feedback provided above)
> 
>     * Compute the balance at `startDate`
> 
>     * Add this balance to all amounts

@Betree : Thanks. I think these are done now. Let me know if you see anything else. 👍🏽

server/lib/host-metrics.js
SudharakaP: This one I am not sure. If we take into account platform fee, host fee etc wouldn't the total money managed metric be problematic. Also when I am comparing this numbers with `getTotalMoneyManagedAmount`;

https://github.com/opencollective/opencollective-api/blob/e0a67eb968f94357da3b1257e6d5c4d5879fd1d5/server/lib/budget.js#L192-L198

they don't seem to tally with each other if we include the platform fee and stuff. So I was thinking we should exclude those and only consider these kinds. 🤔

server/graphql/v2/object/HostMetricsTimeSeries.ts
SudharakaP: Actually the host fee chart doesn't need the filtering as its specifically for one year. Sorry I spoke too soon. 😄

server/lib/host-metrics.js
znarf: > they don't seem to tally

do you have a test or a concrete example to back this?

> if we include the platform fee and stuff

this is an highly imprecise way of formulating it. what is "platform fee and stuff" exactly?

> So I was thinking we should exclude those and only consider these kinds.

how did you pick up this list, what are you filtering exactly, for which reason?

I see no reason to filter any kinds, everything should be counted.

server/lib/host-metrics.js
```diff
@@ -258,6 +259,49 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
...
+
+  const balanceAtStartDate = await getTotalMoneyManagedAmount(host, {
+    endDate: startDate,
+    collectiveIds,
+    currency: host.currency,
+  });
+  const timeSeries = await convertCurrencyForTimeSeries(results, host.currency);
+  let sum;
+  return timeSeries.map(point => {
+    sum = (sum || 0) + point.amount + balanceAtStartDate.value;
```
znarf, line 300: It's a bit hard to believe that we end up with this when I gave precise guidance on how to compute this metric. https://github.com/opencollective/opencollective-api/pull/6786#discussion_r751090599

It's plain wrong and doesn't make sense if you read it.

SudharakaP: Ops. 🤯 Seems I am summing the `balanceAtStartDate` repeatedly which I indeed wrong. I've corrected this. Let me know if that looks better. Sorry about that.

server/lib/host-metrics.js
SudharakaP: > > they don't seem to tally
> 
> do you have a test or a concrete example to back this?

Yes, I am going by the following two tests; 

https://github.com/opencollective/opencollective-api/blob/50e81954a99eb0c7a0180ee876569b1ae3a84293/test/server/graphql/v2/query/HostMetricsQuery.test.js#L119-L145

> > if we include the platform fee and stuff
> 
> this is an highly imprecise way of formulating it. what is "platform fee and stuff" exactly?

Sorry, what I meant was all the transaction kinds excluding, `CONTRIBUTION, EXPENSE, ADDED_FUNDS, BALANCE_TRANSFER`. These are, `HOST_FEE, HOST_FEE_SHARE, HOST_FEE_SHARE_DEBT, PAYMENT_PROCESSOR_COVER, PAYMENT_PROCESSOR_FEE, PLATFORM_FEE, PLATFORM_TIP, PLATFORM_TIP_DEBT` and `PREPAID_PAYMENT_METHOD`.

> > So I was thinking we should exclude those and only consider these kinds.
> 
> how did you pick up this list, what are you filtering exactly, for which reason?

When I test it with these kinds (such as `PLATFORM_TIP` or `HOST_FEE`) the test which I've given above will not pass which is why I thought that these should be excluded.

> I see no reason to filter any kinds, everything should be counted.

Let me think more on this and try to count everything. Maybe I should be missing something here; or maybe my test is somehow wrong. 🤔 🤔

SudharakaP: Ah, I found the problem here. It seems that I didn't put the host id when considering the collective ID list. I think we need this as we are considering the total money managed by the host in this case. That's why the tests were failing. I've corrected that and removed the kinds as you suggested. Sorry for the confusion. 😢 

https://github.com/opencollective/opencollective-api/blob/7a0082ed943070f417113af2998a44f23e048ee2/server/lib/host-metrics.js#L269

server/lib/host-metrics.js
```diff
@@ -266,6 +267,49 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
...
+  { startDate, endDate, collectiveIds = null, timeUnit } = {},
+) {
+  if (!collectiveIds) {
+    const hostedCollectives = await host.getHostedCollectives();
+    collectiveIds = hostedCollectives.map(c => c.id);
+    collectiveIds.push(host.id);
+  }
+
+  const results = await sequelize.query(
+    `SELECT SUM(t1."amountInHostCurrency") as "_amount", t1."hostCurrency" as "_currency", DATE_TRUNC(:timeUnit, t1."createdAt") as "date"
```
znarf, line 281:
```suggestion
    `SELECT
      SUM(COALESCE("amountInHostCurrency", 0)) +
        SUM(COALESCE("platformFeeInHostCurrency", 0)) +
        SUM(COALESCE("hostFeeInHostCurrency", 0)) +
        SUM(COALESCE("paymentProcessorFeeInHostCurrency", 0)) +
        SUM(COALESCE("taxAmount" * "hostCurrencyFxRate", 0)) as "_amount",
      t1."hostCurrency" as "_currency",
      DATE_TRUNC(:timeUnit, t1."createdAt") as "date"
```

znarf: Please compute as "netAmountInHostCurrency".

SudharakaP: I've done the above change. 

> Please compute as "netAmountInHostCurrency".

Do you mean changing the name as well to `netAmountInHostCurrency` instead of `_amount`? This will have a problem as the `convertCurrencyForTimeSeries` function relies on the `_amount` as the variable name. 🤔 

https://github.com/opencollective/opencollective-api/blob/d3c341a90cc710d99c0a3ab76d3d23a14a2d2f22/server/lib/host-metrics.js#L37

znarf: No, I didn't mean this.

znarf (approved):
Looks good to me. Tested. @Betree for final approval ?

server/lib/host-metrics.js
```diff
@@ -266,6 +267,56 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
   return orderBy(mergedTimeSeries, 'date');
 }
 
+export async function getTotalMoneyManagedTimeSeries(
+  host,
+  { startDate, endDate, collectiveIds = null, timeUnit } = {},
+) {
+  if (!collectiveIds) {
+    const hostedCollectives = await host.getHostedCollectives();
+    collectiveIds = hostedCollectives.map(c => c.id);
```
Betree, line 276: I'm not sure about this part. We're going to load the full list of Collectives (in the case of OSC, 4000+) in RAM with their `longDescription`, `settings` and all the heavy stuff then populate that into the model just to get an `id`. I don't want to block on this because this PR is old and we really need to move it forward, but we should:
1. At least add an `attributes: ['id']` to avoid fetching unnecessary fields 
2. Have a dedicated query that returns just an array of integers
3. Do we really need that? Isn't filtering on `HostCollectiveId` enough?

Betree (approved):
Not blocking so we can move forward, but the feedback shared below should be considered a soon as possible. We already have performances issues on the reports page, we should be careful with the new queries we introduce

server/lib/host-metrics.js
SudharakaP: @Betree : Ah, good point. I think we can remove this one and instead use a template string to remove the collective id filtering from the query when `collecticeIds` are empty. I've made the change. Let me know if that looks good. 😃

Betree: @znarf wdyt about that? Is it ok to not filter on `CollectiveId`?

znarf: If we want to be consistent in our numbers, we currently have to pass the `collectiveIds` to match what we're doing in the totalMoneyManaged metric (`lib/budget` `getTotalMoneyManagedAmount`). In simple cases, it's the same number, but it's better to be on the safe side and have the exact same behavior.

Betree: Alright, then @SudharakaP let's go with one of the other proposals. Just make sure we only load the IDs and not the full models

SudharakaP: @Betree : I've just added a dedicated query. Let me know how it looks now. Thanks. 😃

server/lib/host-metrics.js
```diff
@@ -266,6 +267,59 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
   return orderBy(mergedTimeSeries, 'date');
 }
 
+export async function getTotalMoneyManagedTimeSeries(
+  host,
+  { startDate, endDate, collectiveIds = null, timeUnit } = {},
+) {
+  if (!collectiveIds) {
+    const results = await sequelize.query(`SELECT id FROM "Collectives" WHERE "HostCollectiveId" = :HostCollectiveId`, {
```
Betree, line 275: This will include deleted collectives and the ones not approved yet (`isActive=false`). Let's make it simple (you'll need to update the `getHostedCollectives` function):

```suggestion
    const hostedCollectives = await host.getHostedCollectives({ attributes: ['id'] });
```

Betree: Actually `getHostedCollectives` is pretty heavy as well. Feel free to just update your query above, but ignore deleted and inactive

SudharakaP: I've corrected it. Thanks. 👍🏽

SudharakaP: Ops, I just saw your second message. Let me adjust that. 😄

SudharakaP: Okay I've corrected it to include the `deletedAt` and `isActive` conditions. I realized we also do this same fetching of collectives here; 

https://github.com/opencollective/opencollective-api/blob/d198a0f09788381c3305f50cd4e0e84f2499c4c8/server/lib/budget.js#L178-L189 

And so I've corrected that with the same query as well.

server/lib/budget.js
```diff
@@ -174,12 +174,18 @@ export async function getTotalMoneyManagedAmount(host, { startDate, endDate, col
...
-    hostedCollectives = await host.getHostedCollectives();
-    ids = hostedCollectives.map(c => c.id);
+    const results = await sequelize.query(
+      `SELECT id FROM "Collectives" WHERE "HostCollectiveId" = :HostCollectiveId AND "deletedAt" IS NULL AND "isActive" = TRUE`,
+      {
+        replacements: { HostCollectiveId: host.id },
+        type: sequelize.QueryTypes.SELECT,
+      },
+    );
+    ids = results.map(result => result.id);
```
znarf, line 188: Sorry, I would like this to be refactored and replaced by a `host.getHostedCollectiveIds` helper.

server/lib/host-metrics.js
```diff
@@ -266,6 +267,62 @@ ORDER BY DATE_TRUNC(:timeUnit, t1."createdAt")`,
...
+) {
+  if (!collectiveIds) {
+    const results = await sequelize.query(
+      `SELECT id FROM "Collectives" WHERE "HostCollectiveId" = :HostCollectiveId AND "deletedAt" IS NULL AND "isActive" = TRUE`,
+      {
+        replacements: { HostCollectiveId: host.id },
+        type: sequelize.QueryTypes.SELECT,
+      },
+    );
+    collectiveIds = results.map(result => result.id);
```
znarf, line 282: Refactor and replace by a `host.getHostedCollectiveIds` helper.

server/lib/budget.js
Betree: I've made the move already in https://github.com/opencollective/opencollective-api/pull/6886

PR #6788, opened 2021-11-04
Improve performances of the contactCollective mutation

Betree:
This PR addresses a performance issue reported in https://opencollective.freshdesk.com/a/tickets/42271, by using a standard Redis cache for rate-limiting rather than the `Activity` table; it's currently super slow to filter entries by `UserId`.


znarf (approved):
Great!

PR #6812, opened 2021-11-09
Introduce Locations table

Betree:
Part of https://github.com/opencollective/opencollective/issues/3494

_Work in progress_


Betree:
We can re-use this work in the future if needed but we're not ready to prioritize it

PR #6837, opened 2021-11-15
EventEmail: handle empty location

Betree:
Fix https://opencollective.slack.com/archives/C0RD4S62E/p1634240916037000

### DIFF
--- a/templates/emails/ticket.confirmed.hbs
+++ b/templates/emails/ticket.confirmed.hbs
@@ -18,7 +18,7 @@ Subject: {{ order.quantity }} {{pluralize "ticket" n=order.quantity}} confirmed 
 {{#if event.address}}
   {{#if isOffline }}
     The location is {{ event.locationName }} ({{event.address}}).
-  {{else if event.address}}
+  {{else}}
     Join us at {{event.address}}.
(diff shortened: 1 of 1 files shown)


templates/emails/ticket.confirmed.hbs
```diff
@@ -14,10 +14,13 @@ Subject: {{ order.quantity }} {{pluralize "ticket" n=order.quantity}} confirmed
...
-{{else if event.address}}
-  Join us at {{event.address}}.
+
+{{#if event.address}}
+  {{#if isOffline }}
+    The location is {{ event.locationName }} ({{event.address}}).
+  {{else if event.address}}
+    Join us at {{event.address}}.
+  {{/if}}
 {{/if}}
```
snitin315, line 24: There is already a check for `event.address` above.

```suggestion
{{#if event.address}}
  {{#if isOffline }}
    The location is {{ event.locationName }} ({{event.address}}).
  {{else}}
    Join us at {{event.address}}.
  {{/if}}
{{/if}}
```

Betree: Good catch!

PR #6838, opened 2021-11-15
Replace transaction hashid with uuid

znarf:


server/graphql/v2/input/TransactionReferenceInput.ts
```diff
@@ -18,27 +17,20 @@ const TransactionReferenceInput = new GraphQLInputObjectType({
...
 const fetchTransactionWithReference = async (
   input: Record<string, unknown>,
   { loaders = null, throwIfMissing = false } = {},
 ): Promise<typeof models.Transaction> => {
-  const dbId = getDatabaseIdFromTransactionReference(input);
   let transaction = null;
-  if (dbId) {
-    transaction = await (loaders ? loaders.Transaction.byId.load(dbId) : models.Transaction.findByPk(dbId));
+  if (input.id) {
+    transaction = await models.Transaction.findOne({ where: { uuid: input.id } });
```
Betree, line 29: On the Frontend, this is going to have an impact. People who trigger mutations from an already-loaded page while after we deploy the change my face errors like `Can't refund ... transaction does not exists`. We can either accept that as a small-enough risk (easily solved by reloading the page), or we implement a fallback like:

```es6
transaction = await models.Transaction.findOne({ where: { uuid: input.id } });
if (!transaction) {
  try {
    id = idDecode(input.id, IDENTIFIER_TYPES.TRANSACTION);
    transaction = await models.Transaction.findByPk(input.id});
  } catch {
    // Ignore invalid GQLV2 IDs
  }
}
```

Betree (commented):
Depending on whether we want to implement the thingy mentioned below, we may want to remove `IDENTIFIER_TYPES.TRANSACTION`.

server/graphql/v2/input/TransactionReferenceInput.ts
znarf: I'll take the risk and remove `IDENTIFIER_TYPES.TRANSACTION`

PR #6845, opened 2021-11-16
Add webhook for authorization request from Stripe

Forster-Groux:


server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -52,23 +57,111 @@ export const assignCardToCollective = async (cardNumber, expireDate, cvv, collec
...
+
+  if (!virtualCard) {
+    throw new Error(`Virtual card ${stripeAuthorization.card.id} not found`);
+  }
+
+  const host = virtualCard.host;
+
+  await checkStripeEvent(host, stripeEvent);
+
+  const amount = stripeAuthorization.amount;
```
znarf, line 71: Would be great to handle zero-decimal currencies properly here. https://stripe.com/docs/currencies#zero-decimal. Search in the codebase convertToStripeAmount ZERO_DECIMAL_CURRENCIES

znarf: To make it future proof, some currency conversion could be interesting, converting balance or amount to the other currency.

znarf: Both suggestions can be added as TODO notes.

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -52,23 +57,111 @@ export const assignCardToCollective = async (cardNumber, expireDate, cvv, collec
...
+    });
+
+    await models.ExpenseItem.create({
+      ExpenseId: expense.id,
+      incurredAt,
+      CreatedByUserId: UserId,
+      amount,
+    });
+
+    await models.Transaction.createDoubleEntry({
```
znarf, line 129: I don't want us to create the Transaction at this point, we should be waiting for the `issuing_transaction.created` event.

server/paymentProviders/utils.js
```diff
@@ -35,125 +37,157 @@ export const persistTransaction = async (
...
+        VirtualCardId: virtualCard.id,
+        data: { authorizationId: fromAuthorizationId },
+      },
+    });
+
+    if (processingExpense && processingExpense.status === ExpenseStatus.PROCESSING) {
+      await processingExpense.update({
+        status: ExpenseStatus.PAID,
+        data,
+      });
```
znarf, line 66: By not using the `setPaid` utility, we skip setting some things like setting `HostCollectiveId: collective.HostCollectiveId`

znarf:
I rebased.

PR #6855, opened 2021-11-17
Add custom card name when assigning a virtual card

Forster-Groux:


server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -67,6 +67,7 @@ const virtualCardMutations = {
         cardNumber,
         expireDate,
         cvv,
+        args.virtualCard.name,
```
znarf, line 70: Interesting that it was already available in `VirtualCardInput`

znarf (approved):

server/graphql/v2/mutation/VirtualCardMutations.ts
Forster-Groux: I would have say "weird" ^^

PR #6861, opened 2021-11-18
Refresh card details from stripe hourly

Forster-Groux:


znarf (approved):
Looks good. Is there a need for a Frontend PR to tweak what we're doing with the status there?

Forster-Groux:
> Looks good. Is there a need for a Frontend PR to tweak what we're doing with the status there?

There is a frontend PR to display the state but the two PR are independent and can be merge separately.

PR #6882, opened 2021-11-23
Set custom timeout for mergeAccounts

Betree:


Betree:
Closing this one for now, I've benchmarked the individual queries and found the two bottlenecks. The indexes introduced in https://github.com/opencollective/opencollective-api/pull/6883 should solve the timeout issue.

PR #6886, opened 2021-11-23
getHostedCollectives performance: add attributes (select)

Betree:
This PR is both a performance optimization (by using a single SQL query + `attributes`) and the first step towards https://github.com/opencollective/opencollective/issues/4987 as it adds a filter on `HostCollectiveId` + `isActive` to the query we return.


znarf (approved):
Looks good!

PR #6887, opened 2021-11-23
Add refund support to `host` payment method

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4369

Front end PR: https://github.com/opencollective/opencollective-frontend/pull/7174

### DIFF
--- a/server/paymentProviders/opencollective/host.js
+++ b/server/paymentProviders/opencollective/host.js
@@ -1,8 +1,11 @@
+import Promise from 'bluebird';
+
 import { maxInteger } from '../../constants/math';
 import { TransactionKind } from '../../constants/transaction-kind';
 import { TransactionTypes } from '../../constants/transactions';
 import { getFxRate } from '../../lib/currency';
-import { getHostFee, getHostFeeSharePercent } from '../../lib/payments';
+import { createRefundTransaction, getHostFee, getHostFeeSharePercent } from '../../lib/payments';
+import { formatCurrency } from '../../lib/utils';
 import models from '../../models';
 
 const paymentMethodProvider = {};
@@ -10,6 +13,22 @@ const paymentMethodProvider = {};
 paymentMethodProvider.features = {
   recurring: false,
   waitToCharge: false,
+};
+
+paymentMethodProvider.refundTransaction = async (transaction, user) => {
+  const host = await models.Collective.findByPk(transaction.CollectiveId);
+
+  const balance = await host.getBalanceWithBlockedFunds();
+  if (balance < transaction.amount) {
+    throw new Error(
+      `Not enough funds available (${formatCurrency(
+        balance,
+        host.currency,
+      )} left) to process this refund (${formatCurrency(transaction.amount, transaction.currency)})`,
+    );
+  }
+
+  return await createRefundTransaction(transaction, 0, null, user);
 };
 
 // We don't check balance for "Added Funds"
--- a/server/paymentProviders/opencollective/host.js
+++ b/server/paymentProviders/opencollective/host.js
@@ -37,7 +37,7 @@ paymentMethodProvider.refundTransaction = async (transaction, user) => {
     throw new Error(
       `Not enough funds available (${formatCurrency(
         balance,
-        collective.currency,
+        transaction.currency,
       )} left) to process this refund (${formatCurrency(transaction.amount, transaction.currency)})`,
     );
   }
--- a/server/paymentProviders/opencollective/host.js
+++ b/server/paymentProviders/opencollective/host.js
@@ -20,19 +20,29 @@ paymentMethodProvider.refundTransaction = async (transaction, user) => {
     throw new Error('Cannot refund a transaction from the same collective');
   }
 
-  const host = await models.Collective.findByPk(transaction.CollectiveId);
+  if (transaction.type === TransactionTypes.DEBIT) {
+    transaction = await transaction.getRelatedTransaction({ type: TransactionTypes.CREDIT });
+  }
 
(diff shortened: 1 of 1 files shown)


SudharakaP:
@znarf : One thing to note is that for now I've added this condition, 

https://github.com/opencollective/opencollective-api/blob/a07b639a5af40e27eb744c2c55821427b59a9d4f/server/paymentProviders/opencollective/host.js#L19-L21

so that users cannot refund a transaction from the same host. I was wondering how we should handle this case. So if they say add funds choosing the same host as source then what should refund do? Maybe just remove that transaction? But then the UI becomes a bit out of context as it's not a refund, it's a deletion of the transaction. 🤔 🤔

server/paymentProviders/opencollective/host.js
```diff
@@ -12,6 +15,26 @@ paymentMethodProvider.features = {
...
+  if (balance < transaction.amount) {
+    throw new Error(
+      `Not enough funds available (${formatCurrency(
+        balance,
+        host.currency,
+      )} left) to process this refund (${formatCurrency(transaction.amount, transaction.currency)})`,
+    );
+  }
+
+  return await createRefundTransaction(transaction, 0, null, user);
```
znarf, line 35:
```suggestion
  return createRefundTransaction(transaction, 0, null, user);
```

server/paymentProviders/opencollective/host.js
```diff
@@ -12,6 +15,26 @@ paymentMethodProvider.features = {
   waitToCharge: false,
 };
 
+paymentMethodProvider.refundTransaction = async (transaction, user) => {
+  if (transaction.CollectiveId === transaction.FromCollectiveId) {
+    throw new Error('Cannot refund a transaction from the same collective');
+  }
+
+  const host = await models.Collective.findByPk(transaction.CollectiveId);
```
znarf, line 23: Why would the `host` be matching `transaction.CollectiveId`. That doesn't make sense.

server/paymentProviders/opencollective/host.js
```diff
@@ -12,6 +15,26 @@ paymentMethodProvider.features = {
...
+
+  const balance = await host.getBalanceWithBlockedFunds();
+  if (balance < transaction.amount) {
```
znarf, line 26: How do you ensure there is no currency mismatch?

znarf:
> so that users cannot refund a transaction from the same host. 

You're confusing `FromCollectiveId` and `HostCollectiveId`, `FromCollectiveId` is not the `host` here. It's possible that is the `host` but it can be any profile. 

About the following:

```
 if (transaction.CollectiveId === transaction.FromCollectiveId) { 
   throw new Error('Cannot refund a transaction from the same collective'); 
 } 
```

We don't need to support this in the current PR but that would be great to find a solution, so we can remove the condition.

server/paymentProviders/opencollective/host.js
znarf:
```suggestion
  const collective = await models.Collective.findByPk(transaction.CollectiveId);
```

znarf: There is no guarantee that we are looking at a CREDIT and the Collective is the one where the money was added, look at how the `collective` paymentMethod is handling this.

znarf: Or look at createRefundTransaction,

```
  if (transaction.type === DEBIT) {
    transaction = await transaction.getRelatedTransaction({ type: CREDIT });
  }
```

SudharakaP: Thanks, I've added this condition and checked to see if a `CREDIT` transaction exists. Let me know if I've missed anything> 🤔

server/paymentProviders/opencollective/host.js
SudharakaP: Ops, I've corrected this by passing the `transaction.currency` when obtaining the balance amount.

SudharakaP:
> > so that users cannot refund a transaction from the same host.
> 
> You're confusing `FromCollectiveId` and `HostCollectiveId`, `FromCollectiveId` is not the `host` here. It's possible that is the `host` but it can be any profile.

Ah thanks for pointing it out. What I meant was from the collective to itself. You are right it might not be a host; I was testing with the host so probably the mistake. Sorry for the confusion. 

> About the following:
> 
> ```
>  if (transaction.CollectiveId === transaction.FromCollectiveId) { 
>    throw new Error('Cannot refund a transaction from the same collective'); 
>  } 
> ```
> 
> We don't need to support this in the current PR but that would be great to find a solution, so we can remove the condition.

I've opened a new issue for that; https://github.com/opencollective/opencollective/issues/4992

server/paymentProviders/opencollective/host.js
```diff
@@ -20,19 +20,29 @@ paymentMethodProvider.refundTransaction = async (transaction, user) => {
...
+  const collective = await models.Collective.findByPk(transaction.CollectiveId);
 
-  const balance = await host.getBalanceWithBlockedFunds();
+  const balance = await collective.getBalanceWithBlockedFunds({ currency: transaction.currency });
   if (balance < transaction.amount) {
     throw new Error(
       `Not enough funds available (${formatCurrency(
         balance,
-        host.currency,
+        collective.currency,
```
znarf, line 40: If the balance is expressed  in `transaction.currency` then why format it with `collective.currency` ?

SudharakaP: It's a mistake; corrected it. Thanks. 👍🏽

znarf (approved):
Looks good, not tested.

PR #6889, opened 2021-11-24
Add create stripe virtual card for collective mutation

Forster-Groux:


server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -87,6 +87,70 @@ const virtualCardMutations = {
...
+    type: new GraphQLNonNull(VirtualCard),
+    args: {
+      name: {
+        type: GraphQLString,
+        description: 'Virtual Card name',
+      },
+      monthlyLimit: {
+        type: GraphQLInt,
+        description: 'Virtual Card monthly limit',
+      },
```
znarf, line 101: We generally try to express amounts using the `Amount` and `AmountInput` objects.

Give clarity on wether the value is expressed in cents (x 100) or in plain currency.

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -87,6 +87,70 @@ const virtualCardMutations = {
       return virtualCard;
     },
   },
+  createCardForCollective: {
```
znarf, line 90: `createVirtualCard` sounds better

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -87,6 +87,70 @@ const virtualCardMutations = {
...
+      const virtualCard = await stripe.createCardForCollective(
+        host,
+        collective.id,
+        user.id,
+        args.name,
+        args.monthlyLimit,
+      );
+
+      await models.Activity.create({
+        type: activities.COLLECTIVE_VIRTUAL_CARD_ASSIGNED,
```
znarf, line 141: The event doesn't look exactly right.

znarf: Create a new type?

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -87,6 +87,70 @@ const virtualCardMutations = {
       return virtualCard;
     },
   },
+  createCardForCollective: {
+    description: 'Create new Stripe Virtual Card for existing hosted collective',
```
znarf, line 91: I would make it generic and have a `provider` argument, it can default to Stripe. If it's different, simple exception.

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -87,6 +87,70 @@ const virtualCardMutations = {
...
+        description: 'Account where the virtual card will be associated',
+      },
+      assignee: {
+        type: new GraphQLNonNull(AccountReferenceInput),
+        description: 'Individual account responsible for the card',
+      },
+    },
+    async resolve(_: void, args, req: express.Request): Promise<VirtualCardModel> {
+      if (!req.remoteUser) {
+        throw new Unauthorized('You need to be logged in to assign a virtual card');
```
znarf, line 113: not assigning but creating here

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -39,21 +39,40 @@ export const assignCardToCollective = async (cardNumber, expireDate, cvv, name,
...
-    spendingLimitAmount: matchingCard['spending_controls']['spending_limits'][0]['amount'],
-    spendingLimitInterval: matchingCard['spending_controls']['spending_limits'][0]['interval'].toUpperCase(),
-  };
+  return createCard(matchingCard, name, collectiveId, host.id, userId);
+};
+
+export const createCardForCollective = async (host, collectiveId, userId, name, monthlyLimit) => {
+  const connectedAccount = await host.getAccountForPaymentProvider(providerName);
+
+  const stripe = getStripeClient(host.slug, connectedAccount.token);
```
znarf, line 48: I think we had a more sophisticated for the token, when the host is the platform account (`=== 'opencollective'`)

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -39,21 +39,40 @@ export const assignCardToCollective = async (cardNumber, expireDate, cvv, name,
...
+    cardholder: cardholders.data[0].id,
+    currency: 'usd',
+    type: 'virtual',
+    status: 'active',
+    // eslint-disable-next-line camelcase
+    spending_controls: {
+      // eslint-disable-next-line camelcase
+      spending_limits: [
+        {
+          amount: monthlyLimit * 100,
```
znarf, line 66: What happens if no `monthlyLimit` is passed?

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -87,6 +87,70 @@ const virtualCardMutations = {
...
+    description: 'Create new Stripe Virtual Card for existing hosted collective',
+    type: new GraphQLNonNull(VirtualCard),
+    args: {
+      name: {
+        type: GraphQLString,
+        description: 'Virtual Card name',
+      },
+      monthlyLimit: {
+        type: GraphQLInt,
+        description: 'Virtual Card monthly limit',
```
znarf, line 100: Should we have a default here? And also maybe check the value make sense compared to what we're expecting? Maybe we should consider some maximum.

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -87,6 +87,70 @@ const virtualCardMutations = {
...
+
+      const collective = await fetchAccountWithReference(args.account, { loaders: req.loaders, throwIfMissing: true });
+      const host = await collective.getHostCollective();
+
+      if (!req.remoteUser.isAdminOfCollective(host)) {
+        throw new Unauthorized("You don't have permission to edit this collective");
+      }
+
+      const assignee = await fetchAccountWithReference(args.assignee, {
+        loaders: req.loaders,
```
znarf, line 124: Should we throw an error?

znarf (commented):
Looks generally good.

A good amount of minor feedback to handle.

server/graphql/v2/mutation/VirtualCardMutations.ts
Forster-Groux: I really don't like writing code for maybe future logic.

server/paymentProviders/stripe/virtual-cards.js
Forster-Groux: It had been refactored inside the `getStripeClient`

server/paymentProviders/stripe/virtual-cards.js
Forster-Groux: It is required for now

server/graphql/v2/mutation/VirtualCardMutations.ts
Forster-Groux: Maybe, for now it's just a required attribute. Let me know what rules do you want.

server/graphql/v2/mutation/VirtualCardMutations.ts
Forster-Groux: In which case ?

znarf: If assignee is not found

server/paymentProviders/stripe/virtual-cards.js
znarf: Cool, didn't see that.

server/graphql/v2/mutation/VirtualCardMutations.ts
znarf: I still think it's more elegant and consistent, without requiring much code to be written.

As you want, that's fine.

server/graphql/v2/mutation/VirtualCardMutations.ts
znarf: Maybe a constant like $1000 as a maximum for now.

server/graphql/v2/mutation/VirtualCardMutations.ts
Forster-Groux: I have to disagree. We don't know if it will happen, and if it happen we don't know what we will do at this time. Maybe we will handle the provider as an .env variable, maybe it will be selected and associated with the collective elsewhere.
Problem is when someone will come back to this code in the future he might think the provider attribute is here for a reason and lose a lot of time figuring out why it is there.

znarf: > As you want, that's fine.

server/graphql/v2/mutation/VirtualCardMutations.ts
znarf: Could you do something here? I see that if assigned is not found, we have a crash later on `const user = await assignee.getUser();`

server/graphql/v2/object/Host.js
```diff
@@ -385,6 +385,7 @@ export const Host = new GraphQLObjectType({
             },
             limit: args.limit,
             offset: args.offset,
+            order: [['createdAt', 'DESC']],
```
znarf, line 388: Better to implement it like in `interface/Account.js` virtualCards resolver.

```
      orderBy: {
        type: ChronologicalOrderInput,
        defaultValue: ChronologicalOrderInput.defaultValue,
      },
```

then

```
 order: [[args.orderBy.field, args.orderBy.direction]],
```

server/graphql/v2/mutation/VirtualCardMutations.ts
znarf: `throwIfMissing: true` should be enough?

Forster-Groux: Ok, I also added it to the assign card mutation

server/graphql/v2/object/Host.js
```diff
@@ -355,6 +356,7 @@ export const Host = new GraphQLObjectType({
           limit: { type: new GraphQLNonNull(GraphQLInt), defaultValue: 100 },
           offset: { type: new GraphQLNonNull(GraphQLInt), defaultValue: 0 },
           state: { type: GraphQLString, defaultValue: null },
+          orderBy: { type: OrderByInput, defaultValue: { field: 'createdAt', direction: 'DESC' } },
```
znarf, line 359: Anything wrong with `ChronologicalOrderInput` which was suggested?

Forster-Groux: `ChronologicalOrderInput` is acting weird. There are default value in the definition of the input but when I do `orderBy: {type : ChronologicalOrderInput}`, I have an error saying `cannot read field attribute from undefined`.
In `pendingApplications` query field who is using `ChronologicalOrderInput`, `defaultValue: { field: 'createdAt', direction: 'DESC' }` is specified. If I do this it works but it seems redondant with the definition of `ChronologicalOrderInput`.
Furthermore, do we really want a specific orderBy for date instead of a generic one ?
What do you prefer ?

znarf: In most cases, I see that it's used like the following:

```
    orderBy: {
      type: new GraphQLNonNull(ChronologicalOrderInput),
      defaultValue: CHRONOLOGICAL_ORDER_INPUT_DEFAULT_VALUE,
    },
```

Have you tried that way?

> Furthermore, do we really want a specific orderBy for date instead of a generic one ?

It's an established practice to use this specialized input for such case. Yes, I think it's good to use it there.

znarf (approved):

PR #6896, opened 2021-11-26
 Correct Inconsistencies of Timeseries for Transactions

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4995

Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/7193

### DIFF
--- a/server/graphql/v2/object/Host.js
+++ b/server/graphql/v2/object/Host.js
@@ -68,10 +68,12 @@ const getNumberOfDays = (startDate, endDate, host) => {
 };
 
 const getTimeUnit = numberOfDays => {
-  if (numberOfDays <= 7) {
-    return 'WEEK';
-  } else if (numberOfDays <= 366) {
-    return 'MONTH';
+  if (numberOfDays < 21) {
+    return 'DAY'; // Up to 3 weeks
+  } else if (numberOfDays < 90) {
+    return 'WEEK'; // Up to 3 months
+  } else if (numberOfDays < 365 * 3) {
+    return 'MONTH'; // Up to 3 years
   } else {
     return 'YEAR';
   }
@@ -558,7 +560,7 @@ export const Host = new GraphQLObjectType({
               throwIfMissing: true,
               attributes: ['id'],
             });
-            const collectiveIds = collectives.map(collective => collective.id);
+            collectiveIds = collectives.map(collective => collective.id);
             where.CollectiveId = { [Op.in]: collectiveIds };
           }
 
@@ -652,7 +654,7 @@ export const Host = new GraphQLObjectType({
           let collectiveIds;
           if (args.account) {
             const collectives = await fetchAccountsWithReferences(args.account, { throwIfMissing: true });
-            const collectiveIds = collectives.map(collective => collective.id);
+            collectiveIds = collectives.map(collective => collective.id);
             where.CollectiveId = { [Op.in]: collectiveIds };
           }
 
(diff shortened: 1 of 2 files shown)


server/graphql/v2/object/Host.js
```diff
@@ -646,11 +657,13 @@ export const Host = new GraphQLObjectType({
           const expenseAmountOverTime = async () => {
             const dateFrom = args.dateFrom ? moment(args.dateFrom).toISOString() : undefined;
             const dateTo = args.dateTo ? moment(args.dateTo).toISOString() : undefined;
+            const timeUnit = getTimeUnit(numberOfDays);
```
Betree, line 660: What happens to `args.timeUnit`? Sounds to me like we should remove its `defaultValue`, do something like:

```suggestion
            const timeUnit = args.timeUnit || getTimeUnit(numberOfDays);
```

And clarify all that in the field description.

SudharakaP: I've changed this to make sure the default value is removed and made the changes. Thanks. 🙂

server/graphql/v2/object/Host.js
```diff
@@ -67,6 +67,16 @@ const getNumberOfDays = (startDate, endDate, host) => {
...
 };
 
+const getTimeUnit = numberOfDays => {
+  if (numberOfDays <= 7) {
+    return 'WEEK';
+  } else if (numberOfDays <= 366) {
+    return 'MONTH';
+  } else {
+    return 'YEAR';
+  }
```
Betree, line 77:
```suggestion
  if (numberOfDays <= 21) {
    return 'DAY'; // Up to 3 weeks
  } else if (numberOfDays <= 90) {
    return 'WEEK'; // Up to 3 months
  } else if (numberOfDays <= 365 * 3) {
    return 'MONTH'; // Up to 3 years
  } else {
    return 'YEAR';
  }
```

Betree (commented):
We can always improve them later, but the following values for the time unit would make more sense. They'll prevent having a single column in the chart, which is something we should avoid.

Basically, we want the chart to display meaningful data regardless of the selected time range.

server/graphql/v2/object/Host.js
SudharakaP: @Betree : The requirement to show the weekly amounts seems a bit complicated to wrap my head around. For now I've added,

https://github.com/opencollective/opencollective-api/blob/34c1e2a719caa1f7f56241af27cd674aeaee2390/server/graphql/v2/object/Host.js#L71-L77

I think that's better that what was there hopefully. If no objections, I'll do this weekly amount display in a separate PR? 🤔

Betree: If you do `numberOfDays <= 31` you'll end up with the same issue of having a chart that only contains a single value. Play with it to find something that looks nice, but something like `90` sounds like a decent default.

No problem to add `WEEK` in a separate PR if it complicates things (and actually, we should only add it if there's a value in doing so). Could you shortly explain where the complexity to add this is coming from?

SudharakaP: Actually looking at this today I realized the front-end implementation for this dynamic timeunits can be greatly simplified. I've just did that and now it should look much better and also I've added the week support. 🙂 Sorry about this, I feel I was making a too complicated mess out of this. 🤔

znarf:
This is simply crashing. 

Screenshot 2021-12-06 at 10 58 31

znarf (changes requested):
The page should not be crashing (might be frontend side).

SudharakaP:
> The page should not be crashing (might be frontend side).

Yes this is my mistake on the front end side. I've corrected it.

Betree (approved):
Works for me, we'll see with experience if the default values for `timeUnit` make sense

PR #6915, opened 2021-12-02
Make sure to throw proper error when start and end time is not specified in events

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4865

Front end PR: https://github.com/opencollective/opencollective-frontend/pull/7216

image


SudharakaP:
I'll tackle the problem on a separate issue; https://github.com/opencollective/opencollective/issues/5011 as this is not related to https://github.com/opencollective/opencollective/issues/4865

PR #6917, opened 2021-12-03
Make settlement script look at everything owed 

znarf:
and not just current month.

Fix https://github.com/opencollective/opencollective/issues/4520

### DIFF
--- a/cron/monthly/host-settlement.ts
+++ b/cron/monthly/host-settlement.ts
@@ -74,12 +74,12 @@ export async function run(baseDate: Date | moment.Moment = defaultDate): Promise
   );
 
   for (const host of hosts) {
-    const pendingPlatformTips = await getPendingPlatformTips(host, { startDate, endDate });
-    const pendingHostFeeShare = await getPendingHostFeeShare(host, { startDate, endDate });
-
     if (HOST_ID && host.id !== parseInt(HOST_ID)) {
       continue;
     }
+
+    const pendingPlatformTips = await getPendingPlatformTips(host);
+    const pendingHostFeeShare = await getPendingHostFeeShare(host);
 
     const plan = await host.getPlan();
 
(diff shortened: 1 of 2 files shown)


server/lib/host-metrics.js
```diff
@@ -414,9 +414,9 @@ export async function getPendingHostFeeShare(host, { startDate, endDate, collect
           AND t."kind" = 'HOST_FEE_SHARE_DEBT'
           AND t."deletedAt" IS NULL
           AND ts."deletedAt" IS NULL
-          AND ts."status" IN ('OWED', 'INVOICED')
-          AND t."createdAt" >= :startDate
-          AND t."createdAt" <= :endDate
+          AND ts."status" IN ('OWED')
```
Betree, line 417: Not sure about this one. If it's `INVOICED`, then it's still pending no?

cron/monthly/host-settlement.ts
```diff
@@ -90,11 +90,10 @@ export async function run(baseDate: Date | moment.Moment = defaultDate): Promise
 FROM "Transactions" as t
 INNER JOIN "TransactionSettlements" ts ON ts."TransactionGroup" = t."TransactionGroup" AND t.kind = ts.kind
 WHERE t."CollectiveId" = :CollectiveId
-AND t."createdAt" >= :startDate AND t."createdAt" < :endDate
 AND t."kind" IN ('PLATFORM_TIP_DEBT', 'HOST_FEE_SHARE_DEBT')
 AND t."isDebt" IS TRUE
 AND t."deletedAt" IS NULL
-AND ts."status" != 'SETTLED'`,
+AND ts."status" = 'OWED'`,
```
Betree, line 96: Good change, this one could have resulted in duplicate settlements for the same transaction

server/lib/host-metrics.js
znarf: It depend the definition. In the context of the settlement process, we don't want to Invoice multiple time.

I think it's fine to move to this definition.

znarf: Alternative is to pass an explicit `kind` when we want it.

Betree: I agree for the settlement script. For host metrics however, when I look at my pending host fee share, I expect to see everything that I still haven't paid - invoiced or not.

znarf: You might be right theoretically.

However, as far as I know, we're not surfacing this metric anymore, it's only used in the context of the settlement.

Still, I took the feedback in consideration and refactored, so there is no surprise.

PR #6918, opened 2021-12-03
Send an email after creating a virtual card

Forster-Groux:


znarf (changes requested):
I think the end user doesn't care if the card is created or assigned. 

It's a technical detail that is not relevant for them. I would change the wording and make sure we use the same email template in both cases.

That way, we'll also have only one template to maintain.

server/constants/activities.js
```diff
@@ -31,8 +31,7 @@ export default {
   COLLECTIVE_UPDATE_CREATED: 'collective.update.created',
   COLLECTIVE_UPDATE_PUBLISHED: 'collective.update.published',
   COLLECTIVE_CONTACT: 'collective.contact',
-  COLLECTIVE_VIRTUAL_CARD_ASSIGNED: 'collective.virtualcard.assigned',
-  COLLECTIVE_VIRTUAL_CARD_CREATED: 'collective.virtualcard.created',
```
znarf, line 35: Really small change, we have a practice of keeping past activities that are not used anymore. Because they are still used in the database for past entries.

Instead of removing them, can you move them to the bottom of the list?

znarf: @Forster-Groux Could you do the change so we can merge?

Forster-Groux: I did it yesterday by amending the last commit (there is no new commit). I've just repushed because I forgot the graphQL schema. It should be good now.

znarf: Perfect.

znarf (approved):

PR #6919, opened 2021-12-03
Always return the amounts in host currency for receipts

Betree:
Part of https://github.com/opencollective/opencollective/issues/5015


znarf (approved):
Looks good.

PR #6929, opened 2021-12-06
Correct the default start/end times for host metrics

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4995

This handles the default start time and end time inconsistencies in the API. The problem I saw here is that on some metrics we use things like,

`AND t1."createdAt" >= :startDate AND t1."createdAt" <= :endDate`

I've made things consistent by using two separate template strings for `startDate` and `endDate` at all places.

### DIFF
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -9,6 +9,7 @@ import { memoize } from './cache';
 import { convertToCurrency } from './currency';
 import sequelize, { Op } from './sequelize';
 import { amountsRequireTaxForm } from './tax-forms';
+import { computeDates } from './utils';
 
 const twoHoursInSeconds = 2 * 60 * 60;
 const models = sequelize.models;
@@ -1128,7 +1129,15 @@ const getTaxFormsRequiredForAccounts = async (accountIds = [], year) => {
 /**
  * Returns the contribution or expense amounts over time.
  */
-const getTransactionsTimeSeries = async (kind, type, hostCollectiveId, timeUnit, collectiveIds, dateFrom, dateTo) => {
+const getTransactionsTimeSeries = async (
+  kind,
+  type,
+  hostCollectiveId,
+  timeUnit,
+  collectiveIds,
+  dateFrom = null,
+  dateTo = null,
+) => {
   return sequelize.query(
     `SELECT DATE_TRUNC(:timeUnit, "createdAt") AS "date", sum("amountInHostCurrency") as "amount", "hostCurrency" as "currency"
        FROM "Transactions"
@@ -1137,8 +1146,8 @@ const getTransactionsTimeSeries = async (kind, type, hostCollectiveId, timeUnit,
          AND type = :type
          AND "deletedAt" IS NULL
          ${collectiveIds ? `AND "CollectiveId" IN (:collectiveIds)` : ``}
-         ${dateFrom ? `AND "createdAt" >= :dateFrom` : ``}
-         ${dateTo ? `AND "createdAt" <= :dateTo` : ``}
+         ${dateFrom ? `AND "createdAt" >= :startDate` : ``}
+         ${dateTo ? `AND "createdAt" <= :endDate` : ``}
        GROUP BY DATE_TRUNC(:timeUnit, "createdAt"), "hostCurrency"
        ORDER BY DATE_TRUNC(:timeUnit, "createdAt")
       `,
@@ -1150,8 +1159,7 @@ const getTransactionsTimeSeries = async (kind, type, hostCollectiveId, timeUnit,
         hostCollectiveId,
         timeUnit,
         collectiveIds,
-        dateFrom,
-        dateTo,
+        ...computeDates(dateFrom, dateTo),
       },
     },
   );
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -9,7 +9,7 @@ import { memoize } from './cache';
 import { convertToCurrency } from './currency';
 import sequelize, { Op } from './sequelize';
 import { amountsRequireTaxForm } from './tax-forms';
-import { computeDates } from './utils';
+import { computeDatesAsISOStrings } from './utils';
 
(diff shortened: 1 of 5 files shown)


server/lib/queries.js
```diff
@@ -1137,8 +1146,8 @@ const getTransactionsTimeSeries = async (kind, type, hostCollectiveId, timeUnit,
          AND type = :type
          AND "deletedAt" IS NULL
          ${collectiveIds ? `AND "CollectiveId" IN (:collectiveIds)` : ``}
-         ${dateFrom ? `AND "createdAt" >= :dateFrom` : ``}
-         ${dateTo ? `AND "createdAt" <= :dateTo` : ``}
+         ${dateFrom ? `AND "createdAt" >= :startDate` : ``}
+         ${dateTo ? `AND "createdAt" <= :endDate` : ``}
```
znarf, line 1150: This is confusing to have different names in the same block, dateFrom/startDate. This should be avoided.

znarf (changes requested):
By reading the code, I don't see how this address the problem we're trying to fix.

If and `endDate` is passed and **no** `startDate` is passed, we're still defaulting `startDate` to the first of the month. If `endDate` is before the first of the month, we have an empty metric.

Have you tested? Do I get something wrong?

server/lib/queries.js
SudharakaP: I've changed this to have the same names. 👍🏽

SudharakaP:
> By reading the code, I don't see how this address the problem we're trying to fix.
> 
> If and `endDate` is passed and **no** `startDate` is passed, we're still defaulting `startDate` to the first of the month. If `endDate` is before the first of the month, we have an empty metric.
> 
> Have you tested? Do I get something wrong?

@znarf: Note that in this PR we have the following constructs in `host-metrics.js`;

https://github.com/opencollective/opencollective-api/blob/4c9ed4660c42ee42f59d5768bd26a938c7c1a74d/server/lib/host-metrics.js#L92-L93

This makes sure that if `startDate` is not passed the sql query doesn't use the `startDate` at all. Before this PR this kinda construct was only used on some queries not all queries. This is the reason I thought of making it consistent. Now I think you might be looking at the function, 

https://github.com/opencollective/opencollective-api/blob/4c9ed4660c42ee42f59d5768bd26a938c7c1a74d/server/lib/utils.js#L558-L563

I thought of reusing this function and moved it to `utils.js` as we used it in `queries.js` as well. Maybe not related to this PR necessarily but I think we could ideally return `null` if the start or end date is not passed just to avoid any confusion. 🤔

server/lib/queries.js
```diff
@@ -1137,8 +1146,8 @@ const getTransactionsTimeSeries = async (kind, type, hostCollectiveId, timeUnit,
          AND type = :type
          AND "deletedAt" IS NULL
          ${collectiveIds ? `AND "CollectiveId" IN (:collectiveIds)` : ``}
-         ${dateFrom ? `AND "createdAt" >= :dateFrom` : ``}
-         ${dateTo ? `AND "createdAt" <= :dateTo` : ``}
+         ${startDate ? `AND "createdAt" >= :startDate` : ``}
+         ${endDate ? `AND "createdAt" <= :endDate` : ``}
```
znarf, line 1150: It's still a problematic and confusing pattern that the `startDate` used in the condition is not the same as the `startDate` passed as replacement. 

It's making the code really difficult to follow

znarf: This should be refactored.

znarf: What's the point of computing a default value for `startDate`/`endDate` in `computeDates` if we're not using it?

znarf: Or, if you're not interested in the default values, then just don't use `computeDates`.

SudharakaP: Yes, I've just refactored it so that we remove this default value condition. I've renamed `computeDates` to `computeDatesAsISOStrings` since that's what this function does currently. Let me know if any other refactoring needs to be done.

server/models/Collective.js
```diff
@@ -3045,9 +3045,8 @@ function defineModel() {
     if (!this.isHostAccount || !this.isActive || this.type !== types.ORGANIZATION) {
       return null;
     }
-
-    from = from ? moment(from) : moment().utc().startOf('month');
-    to = to ? moment(to) : moment(from).utc().endOf('month');
+    from = from ? moment(from) : null;
+    to = to ? moment(to) : null;
```
znarf, line 3049: You're not updating the inline documentation above?

znarf: Changing the behavior is gonna obviously break the previous Host Metrics panel (related https://github.com/opencollective/opencollective/issues/4989). What's your strategy around this?

SudharakaP: > You're not updating the inline documentation above?

Ops, missed that. I've updated it now. 🤔

> Changing the behavior is gonna obviously break the previous Host Metrics panel (related [opencollective/opencollective#4989](https://github.com/opencollective/opencollective/issues/4989)). What's your strategy around this?

Thanks for bringing this up. Something I've overlooked; I'll do a PR today to tackle https://github.com/opencollective/opencollective/issues/4989 so we can merge these two PRs together. Hope that sounds good? 🤔

znarf: > What's your strategy around this?

SudharakaP: My thinking was that we show the message like this https://github.com/opencollective/opencollective-frontend/pull/7246 and only show the metrics in one place (which is the new dashboard). Or is there a specific need that we need to show the host metrics in the old host metrics page as well until we completely remove it? 🤔

SudharakaP: @znarf : Now that https://github.com/opencollective/opencollective-frontend/pull/7246 is merged I think we are good to go with this one?

znarf: Ok

SudharakaP:
Merging this one after CI pass if no objections. 👍🏽

PR #6932, opened 2021-12-07
Add Expense.quote resolver and Wise quote caching

kewitz:
Solves https://github.com/opencollective/opencollective/issues/3089.
Solves https://github.com/opencollective/opencollective/issues/4963.

This allows us to display Wise fees before paying for the expense and optimizing the performance of Wise expense payment flow.

### DIFF
--- a/server/paymentProviders/transferwise/index.ts
+++ b/server/paymentProviders/transferwise/index.ts
@@ -104,6 +104,15 @@ async function quoteExpense(
 ): Promise<QuoteV2> {
   await populateProfileId(connectedAccount);
 
+  const isExistingQuoteValid =
+    expense.data?.quote &&
+    expense.data.quote.paymentOption &&
+    moment.utc().subtract(60, 'seconds').isBefore(expense.data.quote.expirationTime);
+  if (isExistingQuoteValid) {
+    logger.debug(`quoteExpense(): reusing existing quote...`);
+    return expense.data.quote;
+  }
+
   const token = await getToken(connectedAccount);
 
   const quoteParams = {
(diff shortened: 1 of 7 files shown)


server/paymentProviders/transferwise/index.ts
```diff
@@ -104,6 +104,15 @@ async function quoteExpense(
 ): Promise<QuoteV2> {
   await populateProfileId(connectedAccount);
 
+  const isExistingQuoteValid =
+    expense.data?.quote &&
+    expense.data.quote.paymentOption &&
+    moment.utc().subtract(60, 'seconds').isBefore(expense.data.quote.expirationTime);
```
znarf, line 110: What's the rationale for those 60 seconds?

What is the expirationTime in practice?

kewitz: I'm considering a 60 seconds buffer for the transaction creation and fund to be complete. Basically, the whole process that is sometimes taking 25 seconds and causing the network error in https://github.com/opencollective/opencollective/issues/4963.

The expiration time is related to how long does the Quote you created to fulfill the transaction will be valid.

znarf: What is a normal value for `expirationTime` in practice?

znarf (approved):
Looks good, please answer the question on expirationTime so I understand better.

server/paymentProviders/transferwise/index.ts
kewitz: ISO date string.

znarf: Rephrasing, I just wanted to ask how far it is in the future.

Claudio060320 (commented):
Duplicate of #

PR #6940, opened 2021-12-08
Add option to enforce "Approver cannot Pay for Expense"

kewitz:
Solves https://github.com/opencollective/opencollective/issues/3328.

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -7,6 +7,7 @@ import { types as collectiveTypes } from '../../constants/collectives';
 import statuses from '../../constants/expense_status';
 import expenseType from '../../constants/expense_type';
 import FEATURE from '../../constants/feature';
+import { hasOptedInForFeature } from '../../lib/allowed-features';
 import { getFxRate } from '../../lib/currency';
 import logger from '../../lib/logger';
 import { floatAmountToCents } from '../../lib/math';
@@ -244,6 +245,14 @@ export const canPayExpense = async (req: express.Request, expense: typeof models
   } else if (!canUseFeature(req.remoteUser, FEATURE.USE_EXPENSES)) {
     return false;
   } else {
+    const collective = expense.collective || (await models.Collective.findByPk(expense.CollectiveId));
+    const host = await collective.getHostCollective();
+    if (hasOptedInForFeature(host, FEATURE.APPROVERS_CANNOT_PAY_EXPENSES)) {
+      const approver = await expense.getApproverUser();
+      if (approver && approver.id === req.remoteUser.id) {
+        return false;
+      }
+    }
     return isHostAdmin(req, expense);
   }
 };
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -245,8 +245,7 @@ export const canPayExpense = async (req: express.Request, expense: typeof models
   } else if (!canUseFeature(req.remoteUser, FEATURE.USE_EXPENSES)) {
     return false;
   } else {
-    const collective = expense.collective || (await models.Collective.findByPk(expense.CollectiveId));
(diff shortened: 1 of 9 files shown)


server/graphql/common/expenses.ts
```diff
@@ -244,6 +245,14 @@ export const canPayExpense = async (req: express.Request, expense: typeof models
   } else if (!canUseFeature(req.remoteUser, FEATURE.USE_EXPENSES)) {
     return false;
   } else {
+    const collective = expense.collective || (await models.Collective.findByPk(expense.CollectiveId));
```
Betree, line 248: We should try to only use loaders in these helpers, as they're massively used in lists. Same for loading the host and expense approver below

```suggestion
    const collective = expense.collective || (await req.loaders.Collective.byId.load(expense.CollectiveId));
```

Maybe we can create a new loader (e.g. `req.loaders.Expense.getHost`) by extracting the code from https://github.com/opencollective/opencollective-api/blob/917c228c5ae0b086cb2d6a9a2c09a93d51e3c247/server/graphql/v2/object/Expense.js#L154-L161?

Betree (commented):
The core feature looks good, I just want to push for using loaders to reduce the performance impact when fetched in lists.

server/graphql/common/expenses.ts
kewitz: I agree, I'll abstract the host loader.

server/models/Expense.js
```diff
@@ -353,6 +354,19 @@ function defineModel() {
...
+  Expense.prototype.getApproverUser = async function () {
+    const approvalActivity = await models.Activity.findOne({
+      where: {
+        type: activities.COLLECTIVE_EXPENSE_APPROVED,
+        ExpenseId: this.id,
+      },
+      order: [['createdAt', 'DESC']],
+    });
+    if (approvalActivity) {
+      return await models.User.findByPk(approvalActivity.UserId);
```
znarf, line 366:
```suggestion
      return models.User.findByPk(approvalActivity.UserId);
```

znarf (approved):
Good for me.

server/graphql/v2/object/Expense.js
```diff
@@ -150,14 +150,7 @@ const Expense = new GraphQLObjectType({
...
         async resolve(expense, _, req) {
-          if (expense.HostCollectiveId) {
-            return req.loaders.Collective.byId.load(expense.HostCollectiveId);
-          } else {
-            const collective = await req.loaders.Collective.byId.load(expense.CollectiveId);
-            if (collective.HostCollectiveId) {
-              return req.loaders.Collective.byId.load(collective.HostCollectiveId);
-            }
-          }
+          return req.loaders.Expense.host.load(expense.id);
```
Betree, line 153: Nice :ok_hand:

server/graphql/common/expenses.ts
```diff
@@ -244,6 +245,13 @@ export const canPayExpense = async (req: express.Request, expense: typeof models
   } else if (!canUseFeature(req.remoteUser, FEATURE.USE_EXPENSES)) {
     return false;
   } else {
+    const host = await req.loaders.Expense.host.load(expense.id);
+    if (hasOptedInForFeature(host, FEATURE.APPROVERS_CANNOT_PAY_EXPENSES)) {
+      const approver = await expense.getApproverUser();
```
Betree, line 250: Not requested, but I'd be tempted to put this one in a loader as well. The `Activities` table is very big, querying it is always a bit risky (even tho it's well indexed on `ExpenseId`).

Betree (approved):

server/graphql/common/expenses.ts
kewitz: I'll punt this one because I'm planning to refactor how we track approvals in https://github.com/opencollective/opencollective/issues/3081.

PR #6945, opened 2021-12-09
Add support form mutation

flickz:
Follows up https://github.com/opencollective/opencollective-frontend/pull/7235

### DIFF
--- a/server/graphql/v2/mutation/SupportMutations.ts
+++ b/server/graphql/v2/mutation/SupportMutations.ts
@@ -29,6 +29,9 @@ const supportMessageMutation = {
       topic: {
         type: new GraphQLNonNull(GraphQLString),
       },
+      link: {
+        type: GraphQLString,
+      },
       message: {
         type: new GraphQLNonNull(GraphQLString),
       },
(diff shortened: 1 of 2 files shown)


server/graphql/v2/mutation/SupportMutations.ts
```diff
@@ -0,0 +1,60 @@
...
+    },
+    resolve: async (_, args): Promise<Record<string, unknown>> => {
+      if (!isValidEmail(args.email)) {
+        throw new ValidationFailed('Provide a valid email');
+      }
+
+      const subject = 'New support message';
+
+      const emailBody = `
+          Name: <strong>${args.name}</strong></br>
```
Betree, line 44: We're probably going with a Frontend solution (I'll let @znarf comment on that) but just for the reference, we should never use template strings like that because of code injection. If you try to pass:

```es6
name = '<script>alert("Hello!")</script>'
```

You'll get a not-so-nice surprise. Of course, most email clients won't interpret the `script` tag, but you get the idea: malicious users would be able to inject what they want in this template.

znarf:
@flickz We would prefer to handle this directly from the Frontend side, similar to https://github.com/opencollective/foundation-website/blob/master/api/contact.js

znarf:
@flickz let's move to frontend handling as discussed in https://github.com/opencollective/opencollective-api/pull/6945#issuecomment-989599612

PR #6946, opened 2021-12-09
Add support to refund added funds from same collective

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4992

refund-self-added-funds

### DIFF
--- a/test/stories/ledger.test.ts.snap
+++ b/test/stories/ledger.test.ts.snap
@@ -175,3 +175,21 @@ exports[`test/stories/ledger Level 3: Host and collective with different currenc
 | PLATFORM_TIP            | CREDIT | 10000000  | JPY      | 90000                | USD          | 0          | OC Inc | Ben    | OC Inc |            | false    |
 | PLATFORM_TIP            | DEBIT  | -10000000 | JPY      | -90000               | USD          | 0          | Ben    | OC Inc | NULL   |            | false    |"
 `;
+
+exports[`test/stories/ledger Level 4: Refund added funds️ Refund added funds with different collectives 1`] = `
+"
+| kind        | type   | amount | paymentFee | To      | From    | Host | Settlement | isRefund |
+| ----------- | ------ | ------ | ---------- | ------- | ------- | ---- | ---------- | -------- |
+| ADDED_FUNDS | CREDIT | 10000  | 0          | Webpack | ESLint  | OSC  |            | true     |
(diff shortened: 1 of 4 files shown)


znarf:
Could you add a ledger snapshot test to make sure it's covered and we don't have a regression in the future?

znarf (changes requested):
I tested and it works.

We should add some test coverage (test/stories/ledger).

We should also follow up on allowing Refunds on Expenses with same from/to.

SudharakaP:
> I'm expecting some test coverage to move this PR forward.

Working on this now, sorry for being a bit late on this.

SudharakaP:
> I tested and it works.
> 
> We should add some test coverage (test/stories/ledger).

@znarf : Added some ledger tests for refunds of added funds between two collectives and also a refund for a fund added from the same collective.

> We should also follow up on allowing Refunds on Expenses with same from/to.

Created a new issue for this at, https://github.com/opencollective/opencollective/issues/5039

znarf:
@Betree what do you think about the test and the update to the setup made by @SudharakaP. Would you have done it differently?

test/stories/ledger.test.ts.snap
```diff
@@ -175,3 +175,18 @@ exports[`test/stories/ledger Level 3: Host and collective with different currenc
...
+| ADDED_FUNDS | CREDIT | 10000  | 0          | ESLint  | Webpack | OSC  |            | false    |
+| ADDED_FUNDS | DEBIT  | -10000 | 0          | Webpack | ESLint  | OSC  |            | false    |"
+`;
+
+exports[`test/stories/ledger Level 4: Refund added funds️ Refund added funds with same collective 1`] = `
+"
+| kind        | type   | amount | paymentFee | To     | From   | Host | Settlement | isRefund |
+| ----------- | ------ | ------ | ---------- | ------ | ------ | ---- | ---------- | -------- |
+| ADDED_FUNDS | CREDIT | 10000  | 0          | ESLint | ESLint | OSC  |            | false    |"
+`;
```
Betree, line 192: Here we're only snapshotting the data before the refund (aka the setup data). Since the goal is to test the refund, it'd be way more useful to snapshot **after** the refund, to make sure we include both the transaction and its refund.

SudharakaP: @Betree : I've added the snapshot after the refund. Let me know if you see any further issues.

Betree: That looks good to me, I'll let @znarf share a final feedback with this new snapshot in place

znarf (approved):

PR #6949, opened 2021-12-09
Handle different currencies for virtual cards

Forster-Groux:


server/graphql/v2/object/VirtualCard.ts
```diff
@@ -110,6 +110,7 @@ export const VirtualCard = new GraphQLObjectType({
         }
       },
     },
+    currency: { type: GraphQLString },
```
znarf, line 113: We have a dedicated Enum type for Currencies `Currency`

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -56,7 +56,7 @@ export const createVirtualCard = async (host, collective, userId, name, monthlyL
 
   const issuingCard = await stripe.issuing.cards.create({
     cardholder: cardholders.data[0].id,
-    currency: 'usd',
+    currency: host.currency.toLowerCase(),
```
znarf, line 59: Make sense.

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -86,9 +86,21 @@ export const processAuthorization = async (stripeAuthorization, stripeEvent) =>
...
+      data: { authorizationId: stripeAuthorization.id },
+    },
+  });
+
+  if (existingExpense) {
+    logger.warn(`Virtual Card authorization already reconciled, ignoring it: ${stripeAuthorization.id}`);
+    return;
+  }
+
+  const currency = stripeAuthorization.currency.toUpperCase();
```
znarf, line 101: Do we have a guarantee that this will always be the currency of the card? If you have something, could you add a link to the documentation of what to expect.

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -86,9 +86,21 @@ export const processAuthorization = async (stripeAuthorization, stripeEvent) =>
...
+  const amount = convertToStripeAmount(currency, Math.abs(stripeAuthorization.amount));
+  const balance = await host.getBalanceWithBlockedFundsAmount({ currency });
```
znarf, line 103: Ouch! I haven't spotted this before, but we don't care about the balance of the host, we care about the balance of the collective. This is really important.

server/paymentProviders/utils.js
```diff
@@ -41,7 +41,8 @@ export const persistTransaction = async (virtualCard, transaction) => {
   const host = virtualCard.host;
   const collective = virtualCard.collective;
   const vendor = await getOrCreateVendor(transaction.vendorProviderId, transaction.vendorName);
-  const hostCurrencyFxRate = await getFxRate('USD', host.currency);
+  const currency = transaction.currency || 'USD';
+  const hostCurrencyFxRate = await getFxRate(currency, host.currency);
   const description = `Virtual Card charge: ${vendor.name}`;
 
   if (transaction.fromAuthorizationId) {
```
znarf, line 48: What does this condition mean actually, could you add a comment?

I know it's not part of the PR, but spotting this while reading the code again.

server/paymentProviders/stripe/virtual-cards.js
znarf: In the end, doesn't matter it seems, as we're handling everything properly after. Could be any currency and the code would handle this right.

znarf (approved):

PR #6958, opened 2021-12-10
Disable two factor authentication for read only mode

SudharakaP:
If someone has enabled two factor authentication and try to connect to the prod db for example with read only mode there's errors since decryption of the two factor authentication doesn't work without the `DB_ENCRYPTION_SECRET_KEY` from prod. 

Thus I suggest we disable the two factor authentication when connecting with readonly mode so that when testing we don't have to disable the two factor authentication for the account. 🤔


znarf:
I don't see a clear relationship between read only mode and 2fa, except that in you anecdotally want to disable the two things at the same time. Problem is better solved by: https://github.com/opencollective/opencollective-api/pull/6992

PR #6967, opened 2021-12-14
Fix depcheck test

Betree:


Betree:
Everything seems back to normal. Closing

PR #6970, opened 2021-12-15
Update editVirtualCard and deleteVirtualCard mutations

Forster-Groux:


server/paymentProviders/privacy/index.ts
```diff
@@ -90,7 +90,7 @@ const resumeCard = async (virtualCard: VirtualCardModel): Promise<VirtualCardMod
   setCardState(virtualCard, 'OPEN');
 
 const deleteCard = async (virtualCard: VirtualCardModel): Promise<void> => {
-  const host = await models.Collective.findByPk(virtualCard.HostCollectiveId);
+  const host = virtualCard.host;
```
znarf, line 93: Not a requested change.

Just to note that I think this is more fragile for the future.

One day, a developer will reuse this and pass a `virtualCard` without the `host` relation, this will fail.

It's a bit difficult generally to keep track of the relations required by a function, so more practicable to not rely on them and request them on demand,

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -203,10 +206,26 @@ const virtualCardMutations = {
...
+
+        if (monthlyLimitInCents > 100000) {
+          throw new BadRequest(`Monthly limit should not exceed 1000 ${virtualCard.currency}`, undefined, {
+            monthlyLimit: `Monthly limit should not exceed 1000 ${virtualCard.currency}`,
+          });
+        }
+
+        updateAttributes['spendingLimitAmount'] = monthlyLimitInCents;
+
+        await stripe.updateVirtualCard(virtualCard, monthlyLimitInCents);
```
znarf, line 225: Not a requested change.

It's a bit strange to me to have `updateVirtualCard` and have `monthlyLimitInCents` as single argument. I would have passed it within an object `await stripe.updateVirtualCard(virtualCard, {monthlyLimit});`

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -203,10 +206,26 @@ const virtualCardMutations = {
...
 
-      return await virtualCard.update({
-        privateData: { ...(virtualCard.get('privateData') as Object), cvv, expireDate },
+      const updateAttributes = {
+        name: args.name,
         UserId,
-      });
+      };
+
+      if (args.monthlyLimit && virtualCard.spendingLimitInterval === 'MONTHLY' && virtualCard.provider === 'STRIPE') {
```
znarf, line 214: Not a requested change.

It's generally a good practice (at least in our codebase) to establish constants for keywords such as 'MONTHLY' and 'STRIPE' .

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -203,10 +206,26 @@ const virtualCardMutations = {
         UserId = user.id;
       }
 
-      return await virtualCard.update({
-        privateData: { ...(virtualCard.get('privateData') as Object), cvv, expireDate },
+      const updateAttributes = {
+        name: args.name,
         UserId,
-      });
+      };
```
znarf, line 212: What if we don't pass an `args.assignee`? Will `UserId` be null? That's not desirable.

Same for `name`.

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -162,28 +162,31 @@ const virtualCardMutations = {
...
+      name: {
+        type: GraphQLNonNull(GraphQLString),
+        description: 'Virtual card name',
       },
       assignee: {
-        type: AccountReferenceInput,
-        description: 'The individual account reference to attach the Virtual Card to if desired',
+        type: new GraphQLNonNull(AccountReferenceInput),
+        description: 'Individual account responsible for the card',
+      },
```
znarf, line 175: In another comment, I was asking what happens if these are null and now realize the `GraphQLNonNull`.

I think those arguments should be optional.

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -162,28 +162,31 @@ const virtualCardMutations = {
     type: new GraphQLNonNull(VirtualCard),
     args: {
       virtualCard: {
-        type: new GraphQLNonNull(VirtualCardUpdateInput),
-        description: 'Virtual Card being edited its new values',
+        type: new GraphQLNonNull(VirtualCardReferenceInput),
+        description: 'Virtual card reference',
+      },
```
znarf, line 167: It's acceptable in Mutations (and maybe a good practice we would like to enforce) to use an `UpdateInput` instead of a `ReferenceInput`.

But that's also ok like this as we don't have a clear good practice established at this point.

znarf (changes requested):
Looks good but a few desired changes around nullability of update arguments.

server/graphql/v2/mutation/VirtualCardMutations.ts
Forster-Groux: I agree if we have a generic update mutation but that not really the case here.

znarf (approved):

PR #6971, opened 2021-12-15
Add processPendingOrder.details mutation argument

kewitz:
Related to https://github.com/opencollective/opencollective/issues/4069

Related Front End PR: https://github.com/opencollective/opencollective-frontend/pull/7268

### DIFF
--- a/server/graphql/v2/input/OrderDetailsInput.ts
+++ b/server/graphql/v2/input/OrderDetailsInput.ts
@@ -1,0 +1,19 @@
+import { GraphQLInputObjectType } from 'graphql';
+
+import { AmountInput } from './AmountInput';
+
+export const OrderDetailsInput = new GraphQLInputObjectType({
+  name: 'OrderDetailsInput',
+  description: 'Input to set amount details received by the host',
+  fields: () => ({
+    totalAmount: {
+      type: AmountInput,
+    },
+    paymentProcessorFeesAmount: {
+      type: AmountInput,
+    },
+    platformTipAmount: {
+      type: AmountInput,
+    },
+  }),
+});
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -1,5 +1,5 @@
 import { GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
-import { isNil, isNull, isUndefined } from 'lodash';
+import { difference, isEmpty, isNil, isNull, isUndefined, keys } from 'lodash';
 
 import activities from '../../../constants/activities';
 import status from '../../../constants/order_status';
@@ -298,13 +298,10 @@ const orderMutations = {
     description: 'A mutation for the host to approve or reject an order',
     args: {
       order: {
-        type: new GraphQLNonNull(OrderReferenceInput),
+        type: new GraphQLNonNull(OrderUpdateInput),
       },
       action: {
         type: new GraphQLNonNull(ProcessOrderAction),
-      },
-      orderUpdate: {
-        type: OrderUpdateInput,
       },
     },
     async resolve(_, args, req) {
@@ -320,8 +317,9 @@ const orderMutations = {
           throw new ValidationFailed(`Only pending/expired orders can be marked as paid, this one is ${order.status}`);
         }
 
-        if (args?.details) {
-          const { totalAmount, paymentProcessorFeesAmount, platformTipAmount } = args.details;
+        const hasUpdates = !isEmpty(difference(keys(args.order), ['id', 'legacyId']));
+        if (hasUpdates) {
+          const { totalAmount, paymentProcessorFeesAmount, platformTipAmount } = args.order;
           const host = await toAccount.getHostCollective();
           if (totalAmount) {
             if (totalAmount?.currency !== order.currency) {
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -19,8 +19,8 @@ import { getDecodedId } from '../identifiers';
 import { fetchAccountWithReference } from '../input/AccountReferenceInput';
 import { AmountInput, getValueInCentsFromAmountInput } from '../input/AmountInput';
 import { OrderCreateInput } from '../input/OrderCreateInput';
-import { OrderDetailsInput } from '../input/OrderDetailsInput';
 import { fetchOrderWithReference, OrderReferenceInput } from '../input/OrderReferenceInput';
+import { OrderUpdateInput } from '../input/OrderUpdateInput';
 import { getLegacyPaymentMethodFromPaymentMethodInput } from '../input/PaymentMethodInput';
 import { fetchPaymentMethodWithReference, PaymentMethodReferenceInput } from '../input/PaymentMethodReferenceInput';
 import { fetchTierWithReference, TierReferenceInput } from '../input/TierReferenceInput';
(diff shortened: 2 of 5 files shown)


test/server/graphql/v2/mutation/OrderMutations.test.ts
```diff
@@ -672,5 +697,79 @@ describe('server/graphql/v2/mutation/OrderMutations', () => {
...
+              paymentProcessorFeesAmount: { valueInCents: 100, currency: 'USD' },
+              platformTipAmount: { valueInCents: 100, currency: 'USD' },
+            },
+          },
+          hostAdminUser,
+        );
+
+        if (result.errors) {
+          console.log(result.errors);
+        }
```
SudharakaP, line 759:
```suggestion
```

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -315,6 +320,42 @@ const orderMutations = {
...
+          }
+          if (platformTipAmount) {
+            if (platformTipAmount?.currency !== order.currency) {
+              throw new ValidationFailed('platformTipAmount currency must match order currency');
+            }
+            if (!order.data) {
+              order.set('data', {});
+            }
+            order.set(
+              'data.platformTip',
```
SudharakaP, line 352: Correct me if I am wrong, but my understanding is that this should be `platformFee` not `platformTip` as that is what we normally store within the `data` object. It's kinda confusing. 🤔 

```suggestion
              'data.platformFee',
```

SudharakaP (changes requested):

server/graphql/v2/input/OrderDetailsInput.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { AmountInput } from './AmountInput';
+
+export const OrderDetailsInput = new GraphQLInputObjectType({
+  name: 'OrderDetailsInput',
+  description: 'Input to set amount details received by the host',
+  fields: () => ({
+    totalAmount: {
+      type: AmountInput,
+    },
```
znarf, line 11: We're not exposing `totalAmount` so far.

In `OrderCreateInput`, we have `amount`

> description: 'The contribution amount for 1 quantity, without platform contribution and taxes',

I think it's best to re-use this definition, so we don't introduce yet a different definition with a different name.

This mean that frontend will have to construct somehow it by doing `amount = amountReceived + paymentProcessorFeesAmount - platformTipAmount`.

And that you will have to reconstruct it `totalAmount = amount + platformTipAmount`

server/graphql/v2/input/OrderDetailsInput.ts
```diff
@@ -0,0 +1,19 @@
+import { GraphQLInputObjectType } from 'graphql';
+
+import { AmountInput } from './AmountInput';
+
+export const OrderDetailsInput = new GraphQLInputObjectType({
+  name: 'OrderDetailsInput',
```
znarf, line 6: `OrderUpdateInput`

znarf: make it generic.

server/graphql/v2/input/OrderDetailsInput.ts
```diff
@@ -0,0 +1,19 @@
...
+  description: 'Input to set amount details received by the host',
```
znarf, line 7: If we make it more generic, description should be amended.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -301,6 +303,9 @@ const orderMutations = {
       action: {
         type: new GraphQLNonNull(ProcessOrderAction),
       },
+      details: {
```
znarf, line 306: I don't like `details` as name, I would propose `order` but it's already taken. Any other idea?

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -315,6 +320,42 @@ const orderMutations = {
           throw new ValidationFailed(`Only pending/expired orders can be marked as paid, this one is ${order.status}`);
         }
 
+        if (args?.details) {
+          const { totalAmount, paymentProcessorFeesAmount, platformTipAmount } = args.details;
+          const host = await toAccount.getHostCollective();
+          if (totalAmount) {
+            if (totalAmount?.currency !== order.currency) {
+              throw new ValidationFailed('totalAmount currency must match expense currency');
```
znarf, line 328: message is wrong

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -315,6 +320,42 @@ const orderMutations = {
...
+            }
+            order.set('totalAmount', totalAmount.valueInCents || floatAmountToCents(totalAmount.value));
+          }
+          if (paymentProcessorFeesAmount) {
+            if (paymentProcessorFeesAmount?.currency !== host.currency) {
+              throw new ValidationFailed('paymentProcessorFeesAmount currency must match host currency');
```
znarf, line 334: I don't think we should assume paymentProcessorFees to be passed as hostCurrency. It simply needs to be consistent with `order.currency` and should be later converted to `hostCurrency` when inserted in the ledger.

server/graphql/v2/mutation/OrderMutations.js
znarf: There is a forward compatibility utility `getPlatformTip` that take care of both form https://github.com/opencollective/opencollective-api/blob/main/server/lib/payments.js#L787

We ultimately want to use only `platformTip` and migrate everything, but whatever that works for now.

server/graphql/v2/mutation/OrderMutations.js
znarf: Looking at the `editAccount` mutation, I think we should add `id` and `legacyId` to `OrderUpdateInput`

That way we can make the `order` args  an `OrderUpdateInput` instead of `OrderReferenceInput` and don't need a separated `details`.

server/graphql/v2/input/OrderDetailsInput.ts
kewitz: True, I think `totalAmount` is coercing the frontend to speak backend abstractions.

Following what was designed, we want:

Screenshot from 2021-12-16 10-51-44

"Amount received by Accountable": Is this the amount received by the host (donation + tip)?
"Payment processor fees": This is the difference between the amount initiated and the amount received.
"Platform tip amount": Intended tip. To deduct from the amount received.

Is that correct?

znarf: I think so.

I added some comments there:
https://github.com/opencollective/opencollective-frontend/pull/7268#discussion_r770545634

server/graphql/v2/mutation/OrderMutations.js
kewitz: @SudharakaP we're moving the platform away from platform fees towards platform tips. The exception is still OSC, but in that case, I suggest we just hide the field in the frontend.

Platform-tip is working.

server/graphql/v2/mutation/OrderMutations.js
kewitz: Yes, I think that makes sense. I'll update this mutation.

kewitz:
Continues in https://github.com/opencollective/opencollective-api/pull/6976

server/graphql/v2/mutation/OrderMutations.js
kewitz: In the user story, the host admin is checking the amount in their bank statement, wouldn't be easier to get all the amounts in the host currency? If not, should the host admin be able to pick which currency their passing?
Perhaps I should just consider that any currency is fine and do the calculations in the API, WDYT?

PR #6976, opened 2021-12-16
Add processPendingOrder.details mutation argument

kewitz:
Related to https://github.com/opencollective/opencollective/issues/4069

Related Front End PR: https://github.com/opencollective/opencollective-frontend/pull/7268

### DIFF
--- a/server/graphql/v2/input/OrderUpdateInput.ts
+++ b/server/graphql/v2/input/OrderUpdateInput.ts
@@ -1,10 +1,18 @@
-import { GraphQLInputObjectType } from 'graphql';
+import { GraphQLInputObjectType, GraphQLInt, GraphQLString } from 'graphql';
 
 import { AmountInput } from './AmountInput';
 
 export const OrderUpdateInput = new GraphQLInputObjectType({
   name: 'OrderUpdateInput',
   fields: () => ({
+    id: {
+      type: GraphQLString,
+      description: 'The public id identifying the order (ie: dgm9bnk8-0437xqry-ejpvzeol-jdayw5re)',
+    },
+    legacyId: {
+      type: GraphQLInt,
+      description: 'The legacy public id identifying the order (ie: 4242)',
+    },
     totalAmount: {
       type: AmountInput,
     },
--- a/server/graphql/v2/input/OrderUpdateInput.ts
+++ b/server/graphql/v2/input/OrderUpdateInput.ts
@@ -2,9 +2,8 @@ import { GraphQLInputObjectType } from 'graphql';
 
 import { AmountInput } from './AmountInput';
 
-export const OrderDetailsInput = new GraphQLInputObjectType({
-  name: 'OrderDetailsInput',
-  description: 'Input to set amount details received by the host',
+export const OrderUpdateInput = new GraphQLInputObjectType({
+  name: 'OrderUpdateInput',
   fields: () => ({
     totalAmount: {
       type: AmountInput,
--- a/server/graphql/v2/input/OrderUpdateInput.ts
+++ b/server/graphql/v2/input/OrderUpdateInput.ts
@@ -1,7 +1,5 @@
 import { GraphQLInputObjectType, GraphQLInt, GraphQLString } from 'graphql';
 
-import { AmountInput } from './AmountInput';
-
 export const OrderUpdateInput = new GraphQLInputObjectType({
   name: 'OrderUpdateInput',
   fields: () => ({
@@ -13,14 +11,17 @@ export const OrderUpdateInput = new GraphQLInputObjectType({
       type: GraphQLInt,
       description: 'The legacy public id identifying the order (ie: 4242)',
     },
-    totalAmount: {
-      type: AmountInput,
+    amount: {
+      type: GraphQLInt,
+      description: 'Amount received by collective, excluding any tips or fees, in order currency in cents',
     },
-    paymentProcessorFeesAmount: {
-      type: AmountInput,
+    paymentProcessorFee: {
+      type: GraphQLInt,
+      description: 'Amount paid in fees for the payment processor in order currency in cents',
     },
-    platformTipAmount: {
-      type: AmountInput,
+    platformTip: {
+      type: GraphQLInt,
+      description: 'Amount intended as tip for the platform in order currency in cents',
     },
   }),
 });
--- a/server/graphql/v2/input/OrderUpdateInput.ts
+++ b/server/graphql/v2/input/OrderUpdateInput.ts
@@ -1,5 +1,7 @@
 import { GraphQLInputObjectType, GraphQLInt, GraphQLString } from 'graphql';
 
+import { AmountInput } from './AmountInput';
+
 export const OrderUpdateInput = new GraphQLInputObjectType({
   name: 'OrderUpdateInput',
   fields: () => ({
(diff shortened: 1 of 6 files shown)


server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -315,6 +317,49 @@ const orderMutations = {
...
+          const paymentProcessorFeesAmount = getAmountInCents(args.order.paymentProcessorFeesAmount);
+          const platformTipAmount = getAmountInCents(args.order.platformTipAmount);
+          if (amount) {
+            if (args.order.amount?.currency !== order.currency) {
+              throw new ValidationFailed('amount currency must match order currency');
+            }
+
+            if (
+              args.order.paymentProcessorFeesAmount?.currency !== order.currency ||
+              args.order.paymentProcessorFeesAmount?.currency !== host.currency
```
znarf, line 335: I don't see why we need to enforce this.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -315,6 +317,49 @@ const orderMutations = {
...
+            ) {
+              throw new ValidationFailed('platformTipAmount currency must match order and host currency');
+            }
+            const totalAmount = amount + paymentProcessorFeesAmount;
+            order.set('totalAmount', totalAmount);
+          }
+          if (paymentProcessorFeesAmount) {
+            if (args.order.paymentProcessorFeesAmount?.currency !== host.currency) {
```
znarf, line 343: I don't see why we need to enforce this.

server/paymentProviders/opencollective/manual.js
```diff
@@ -62,7 +62,10 @@ async function processOrder(order) {
   const platformTipEligible = await isPlatformTipEligible(order, host);
   const platformTip = getPlatformTip(order);
   const platformTipInHostCurrency = Math.round(platformTip * hostCurrencyFxRate);
-  const paymentProcessorFeeInHostCurrency = order.data?.paymentProcessorFeeInHostCurrency || 0;
+
+  const paymentProcessorFee = order.data?.paymentProcessorFee || 0;
+  const paymentProcessorFeeInHostCurrency =
+    order.data?.paymentProcessorFeeInHostCurrency || Math.round(paymentProcessorFee * hostCurrencyFxRate) || 0;
```
znarf, line 68: I forgot about this, I don't remember exactly why we introduced it.

https://github.com/opencollective/opencollective-api/pull/6138

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -315,6 +316,31 @@ const orderMutations = {
           throw new ValidationFailed(`Only pending/expired orders can be marked as paid, this one is ${order.status}`);
         }
 
+        const hasUpdates = !isEmpty(difference(keys(args.order), ['id', 'legacyId']));
+        if (hasUpdates) {
```
znarf, line 320: Not sure about that pattern. We're gonna send the values all the time from now on?

Should we check that it's set and different instead?

Anyway, that should work.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -315,6 +316,31 @@ const orderMutations = {
...
+            }
+            order.set('data.paymentProcessorFee', paymentProcessorFee);
+          }
+          if (platformTip) {
+            if (!order.data) {
+              order.set('data', {});
+            }
+            order.set('data.platformTip', platformTip);
+            // Some parts of the order flow still uses data.platformFee
+            order.set('data.platformFee', platformTip);
```
znarf, line 339: Cool, setting the two is a good way to anticipate a future migration.

server/graphql/v2/input/OrderUpdateInput.ts
```diff
@@ -0,0 +1,27 @@
...
+    id: {
+      type: GraphQLString,
+      description: 'The public id identifying the order (ie: dgm9bnk8-0437xqry-ejpvzeol-jdayw5re)',
+    },
+    legacyId: {
+      type: GraphQLInt,
+      description: 'The legacy public id identifying the order (ie: 4242)',
+    },
+    amount: {
+      type: GraphQLInt,
```
znarf, line 15: No!

znarf: Can we have `AmountInput` back please.

server/graphql/v2/mutation/OrderMutations.js
kewitz: No. This is backward compatible with existing API calls, and should only be used if necessary.

server/graphql/v2/input/OrderUpdateInput.ts
kewitz: We can. But we'll also add validation to ensure the currency matches the order.
Also, keep in mind we only use `AmountInput` in Order, so this is consistent only within this object but inconsistent with the rest of the inputs.

server/graphql/v2/mutation/OrderMutations.js
kewitz: Yes, I think it is time for us to create a proper column to track the platform contribution and clean this tech debt.

server/graphql/v2/input/OrderUpdateInput.ts
znarf: If we're not using `Amount` / `AmountInput` elsewhere, that's a problem.

We created this pattern so we stop having to think if an amount is in cents or in plain value, and to stop wondering in which currency it's expressed.

Any new development should use this pattern.

server/graphql/v2/mutation/OrderMutations.js
znarf: With this code "hasUpdates" is misleading because it will tell you there is an update even if the amount has not changed. You will then write to the database while it's not needed.

It works but it's misleading and not necessarily optimal.

kewitz: Sure, I think it is not explicit enough. Implicitly it makes sense since none of these arguments are property of order and I'm not sure in which case someone would pass these amounts without consciously wanting to update them.

znarf: > I'm not sure in which case someone would pass these amounts without consciously wanting to update them.

While Marking an order as paid without changing the values. That will be basically all the time?

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -315,6 +316,48 @@ const orderMutations = {
...
+        const hasAmounts = !isEmpty(difference(keys(args.order), ['id', 'legacyId']));
+        if (hasAmounts) {
+          const { amount, paymentProcessorFee, platformTip } = args.order;
+          if (amount) {
+            if (amount.currency !== order.currency) {
+              throw new ValidationFailed('Amount currency must match order currency.');
+            }
+            if (platformTip.currency !== order.currency) {
+              throw new ValidationFailed('Platform tip currency must match order currency.');
+            }
```
znarf, line 328: We're checking this line 347 right?

server/graphql/v2/mutation/OrderMutations.js
kewitz: I thought that in such a case, the field would be empty and not pre-set with the existing value.
@SudharakaP can you confirm how exactly this works?

kewitz: I think always passing the value is prone to error. I think we should leave it empty so the user can know if they're updating the value or not.

znarf (approved):
Good for me

PR #6983, opened 2021-12-20
Block CNY currency payout on business hosts

kewitz:
A lagging problem related to Wise internal restrictions that are leaking into our platform. Despite AliPay appearing as the only viable Transfer method in the Bank Account payout form, this is not actually allowed by Wise.

```
Expense error: TransferWise validation error: Selected recipient type CHINESE_ALIPAY is not supported for this payment
```

References:
https://wise.com/help/articles/2955298/guide-to-cny-transfers
https://wise.com/us/blog/new-send-cny-instantly-to-alipay-users


znarf (approved):
Looks good. Do we have a place where we document those exceptions?

kewitz:
@znarf we don't. I feel like we should, even though we're just enforcing Wise's business rules, we abstract them enough to the point it looks like we're the ones responsible for blocking these.

I think we need to tweak this UX a little bit to reduce the burden. Or accept all currencies and automatically set these expenses as "Custom Payout Method", as we do with hosts that are not connected to Wise.

PR #6985, opened 2021-12-21
Add support for refunds of expenses from same collective

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4286

refund-of-self-expense

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -258,7 +258,13 @@ export const refundPaymentProcessorFeeToCollective = async (transaction, refundT
  */
 export async function createRefundTransaction(transaction, refundedPaymentProcessorFee, data, user) {
   /* If the transaction passed isn't the one from the collective
-   * perspective, the opposite transaction is retrieved. */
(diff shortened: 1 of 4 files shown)


server/lib/payments.js
```diff
@@ -259,7 +259,7 @@ export const refundPaymentProcessorFeeToCollective = async (transaction, refundT
 export async function createRefundTransaction(transaction, refundedPaymentProcessorFee, data, user) {
   /* If the transaction passed isn't the one from the collective
    * perspective, the opposite transaction is retrieved. */
-  if (transaction.type === DEBIT) {
+  if (transaction.type === DEBIT && transaction.FromCollectiveId !== transaction.CollectiveId) {
```
znarf, line 262: If you add a condition like this, you better add a generous comment to explain why.

SudharakaP: Modified the existing comment to handle this. Thanks. 👍🏽

znarf (approved):

PR #6991, opened 2021-12-22
Fix payout method permission for expense across hosts

znarf:


server/graphql/common/expenses.ts
```diff
@@ -519,18 +519,21 @@ const getPayoutMethodFromExpenseData = async (expenseData, remoteUser, fromColle
...
+      // No need for extra checks
       if (
-        // Payout Method from Collective
-        pm.CollectiveId !== fromCollective.id &&
-        // Bank Account or PayPal Payout Method from Host
-        !(
-          pm.CollectiveId === fromCollective.HostCollectiveId &&
-          [PayoutMethodTypes.BANK_ACCOUNT, PayoutMethodTypes.PAYPAL].includes(pm.type)
-        )
+        pm.CollectiveId === fromCollective.HostCollectiveId &&
```
Betree, line 528: So conceptually, we'll always allow collective admins to use their host's payout methods - not only for cross-hosts, since this condition will be valid for an expense from Babel to Webpack. I don't see how this could be directly abused, it looks pretty safe.

znarf: It's not really desirable in that case. But it's safe and the Frontend should not expose this option at this stage.

PR #6966, opened 2021-12-14
Fix refunds with feesPayer

Betree:
Fix https://github.com/opencollective/opencollective/issues/5044
Related to https://github.com/opencollective/opencollective/issues/1073

PR #6993, opened 2021-12-22
Fix netAmount & tests for refunds with a platform fee

Betree:
A prequel of https://github.com/opencollective/opencollective-api/pull/6966 to fix some existing inconsistencies with refunds.

Also added more ledger snapshot tests.


server/lib/payments.js
```diff
@@ -194,7 +194,7 @@ export const buildRefundForTransaction = (t, user, data, refundedPaymentProcesso
     // We're handling payment processor fees and host fees in separate transactions
     refund.hostFeeInHostCurrency = 0;
     refund.paymentProcessorFeeInHostCurrency = 0;
-    refund.netAmountInCollectiveCurrency = -netAmount({ ...t, paymentProcessorFeeInHostCurrency: 0 });
+    refund.netAmountInCollectiveCurrency = netAmount(refund);
```
Betree, line 197: This line was wrong: `{ ...t }` would not end up as intended, because values are actually stored in `...t.dataValues`. As a result, we ended up with `NaN` stored in `netAmountInCollectiveCurrency` which resulted in having it automatically re-computed later, but with some inconsistencies.

test/server/graphql/v1/refundTransaction.test.js
```diff
@@ -267,6 +284,12 @@ describe('server/graphql/v1/refundTransaction', () => {
         where: { OrderId: transaction.OrderId },
       });
 
+      // Snapshot
+      await snapshotTransactionsForRefund(allTransactions);
+
+      // Collective balance should go back to 0
+      expect(await collective.getBalance()).to.eq(0);
```
Betree, line 291: Before the fix above, this line would return `-250`: we were not accounting the refunded platform fees correctly

test/server/lib/payments.test.js.snap
```diff
@@ -2,24 +2,24 @@
...
-| HOST_FEE                | CREDIT | true     |        | Host            | Collective      | Host            | 250    | USD      | 0                         | 0          |            | Refund of \\"Host Fee\\"                                   |
-| CONTRIBUTION            | DEBIT  | true     |        | User            | Collective      | Host            | -4500  | USD      | 0                         | 0          |            | Refund of \\"Monthly subscription to Webpack\\"            |
-| CONTRIBUTION            | CREDIT | true     |        | Collective      | User            | NULL            | 4500   | USD      | 0                         | 0          |            | Refund of \\"Monthly subscription to Webpack\\"            |"
+| kind                    | type   | isRefund | isDebt | From            | To              | Host            | amount | currency | platformFee | paymentFee | Settlement | description                                            |
```
Betree, line 5: Huge diff there but it's only because of the new alias `platformFeeInHostCurrency` -> `platformFee`

PR #7000, opened 2021-12-24
createOrder: Require additional info for logged in users

Betree:
Part of https://github.com/opencollective/opencollective/issues/4692

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -204,7 +204,8 @@ const checkAndUpdateProfileInfo = async (order, fromAccount, isGuest) => {
 
   // Update account with new info, unless we're making a guest contribution for an existing account
   // (we don't want to let guests update the profile of an existing account that they may not own)
-  if (!isGuest || fromAccount.data?.isGuest) {
+  const isVerifiedProfile = !fromAccount.data?.isGuest;
+  if (!isGuest || !isVerifiedProfile) {
     if (mustUpdateLocation(fromAccount.location, location)) {
       accountUpdatePayload.data = { ...fromAccount.data, address: location.structured || fromAccount.data?.structured };
       accountUpdatePayload.locationName = location.name || fromAccount.locationName;
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -179,13 +179,13 @@ const mustUpdateNames = (fromAccount, fromAccountInfo) => {
  * Checks that the profile has all requirements for this contribution (name, address, etc) and updates
  * it if necessary. Must be called **after** checking permissions and authentication!
  */
-const checkAndUpdateProfileInfo = async (order, fromAccount, isGuest) => {
+const checkAndUpdateProfileInfo = async (order, fromAccount, isGuest, currency) => {
   const { totalAmount, fromAccountInfo, guestInfo } = order;
   const accountUpdatePayload = {};
   const location = fromAccountInfo?.location || guestInfo?.location || fromAccount.location;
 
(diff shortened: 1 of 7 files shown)


server/graphql/v1/mutations/orders.js
```diff
@@ -158,20 +158,69 @@ const checkGuestContribution = async (order, loaders) => {
...
+ */
+const checkAndUpdateProfileInfo = async (order, fromAccount, isGuest) => {
+  const { totalAmount, fromAccountInfo, guestInfo } = order;
+  const accountUpdatePayload = {};
+  const location = fromAccountInfo?.location || guestInfo?.location || fromAccount.location;
+
+  // Only enforce profile checks for guests at the moment
+  if (isGuest) {
+    // Contributions that are more than $5000 must have an address attached
+    if (totalAmount > 5000e2) {
```
kewitz, line 190: I like the cents notation, but we should also verify the currency. 👍🏼

server/graphql/v1/mutations/orders.js
```diff
@@ -158,20 +158,69 @@ const checkGuestContribution = async (order, loaders) => {
...
+      const name = fromAccountInfo?.name || fromAccountInfo?.legalName || fromAccount.name || fromAccount.legalName;
+      if (!name) {
+        throw new BadRequest('Contributions that are more than $250 must have a name attached');
+      }
+    }
+  }
+
+  // Update account with new info, unless we're making a guest contribution for an existing account
+  // (we don't want to let guests update the profile of an existing account that they may not own)
+  if (!isGuest || fromAccount.data?.isGuest) {
```
kewitz, line 207: This conditional is a bit cryptic and looks broken.
Are we missing an `!` before `fromAccount.data.isGuest`?
Maybe we declare an auxiliary variable like:

```js
const userIsExistingGuest = !!fromAccount.data?.isGuest;
if (!isGuest || userIsExistingGuest) {
```

server/graphql/v1/mutations/orders.js
```diff
@@ -158,20 +158,69 @@ const checkGuestContribution = async (order, loaders) => {
...
+  if (isGuest) {
+    // Contributions that are more than $5000 must have an address attached
+    if (totalAmount > 5000e2) {
+      if (!location.structured && (!location.address || !location.country)) {
+        throw new BadRequest('Contributions that are more than $5000 must have an address attached');
+      }
+    }
+
+    // Contributions that are more than $250 must have a name attached
+    if (totalAmount > 250e2) {
```
kewitz, line 197: Verify currency as well.

kewitz (changes requested):

server/graphql/v1/mutations/orders.js
Betree: Damn that a very good point, we're not considering currency at any point here.

Betree: I opened a followup issue for that [here](https://github.com/opencollective/opencollective/issues/5092), and I'm going to ignore this check if the currency is not USD for now.

server/graphql/v1/mutations/orders.js
Betree: It's not broken, but it's a bit hard to parse indeed. The idea is:
- Either you're not contributing as a guest, and it's fine to update your profile directly (because you're authenticated)
- Otherwise, you can only update the profile if it's not verified yet (because we don't want anyone submitting an order with your email to be able to edit your profile)

Your suggestion looks good, I'm gonna do something in this spirit

```suggestion
  const existingProfileIsGuest = Boolean(fromAccount.data?.isGuest);
  if (!isGuest || existingProfileIsGuest) {
```

PR #7003, opened 2021-12-28
Cascade host fees changes to children

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5065

- Cascade host fee changes to children
- Add a script to update host fees for existing transactions

Usage:

```bash
$ npm run script scripts/ledger/update-transactions.ts -- --account apex --hostFeePercent 15
648 transaction(s) to update with the following infos:
 - Update 324 transactions with 15% host fee

This action is irreversible. Are you sure you want to continue? (Yes/No)
> 
```

# Migration strategy for the Social Change Nest

1. Deploy
2. Run the script
3. Manually update the fee structure and `hostFeePercent` for all concerned accounts (see the slugs list below)
4. Check the budgets to make sure everything is correct

> 'gaf-film-club-discusses-social-reproduction-and-solidarity-economy-53e79758','connecting-with-health-professionals-58e2cba9','calling-ethnic-minority-applied-psychologists-32cff7cb','trust-the-people-online-course-0160501e','global-poetry-showcase','vappaw-future','reclaiming-our-town-centres-7c343779','community-family-christmas-party-c9e5e5cb','cylch-merched-launch-party-e5144e59'

SQL query:
```sql
SELECT c."hostFeePercent", pc."hostFeePercent", pc."data", c."data" 
FROM "Collectives" c 
INNER JOIN "Collectives" pc ON c."ParentCollectiveId" = pc.id
WHERE c."HostCollectiveId" = 98478
AND c."isActive" IS TRUE
AND c."hostFeePercent" != pc."hostFeePercent" 
AND c."deletedAt" IS NULL
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,6 +137,7 @@
     "chai-as-promised": "^7.1.1",
     "chai-jest-snapshot": "^2.0.0",
     "codecov": "^3.8.2",
+    "commander": "^8.3.0",
     "commitizen": "^4.2.4",
     "cz-conventional-changelog": "^3.3.0",
     "eslint": "^8.1.0",
--- a/scripts/ledger/update-transactions.ts
+++ b/scripts/ledger/update-transactions.ts
@@ -1,0 +1,155 @@
+#!/usr/bin/env ./node_modules/.bin/babel-node
+
+/**
+ * This script aims to provide a command-line interface to update transactions in a safe way.
+ */
+
+import '../../server/env';
+
+import { Command } from 'commander'; // eslint-disable-line node/no-unpublished-import
+import { groupBy, pick } from 'lodash';
+
+import models, { Op } from '../../server/models';
+import { confirm } from '../common/helpers';
+
+// Some local caches to optimize the process
+let accountsCache = {};
+
+// Define modifiers as a map of option<>settings
+const MODIFIERS = {
+  hostFeePercent: {
+    description: 'Update transactions with this host fee percent',
+    parseArgs: value => {
+      const result = parseFloat(value);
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -137,6 +137,7 @@
     "chai-as-promised": "^7.1.1",
     "chai-jest-snapshot": "^2.0.0",
     "codecov": "^3.8.2",
+    "commander": "^8.3.0",
```
Betree, line 140: Introduced this one to have an easy way to parse command-line args

znarf: We used to have `argparse` in some scripts. Would be great to consolidate.

Maybe a small RFC to explain the choice?

scripts/ledger/update-transactions.ts
```diff
@@ -0,0 +1,155 @@
+#!/usr/bin/env ./node_modules/.bin/babel-node
+
+/**
+ * This script aims to provide a command-line interface to update transactions in a safe way.
+ */
+
+import '../../server/env';
+
+import { Command } from 'commander'; // eslint-disable-line node/no-unpublished-import
```
Betree, line 9: Let's switch to `argparse`

Betree:
I initially missed that we had `argparse` available in the codebase, I should have used that from the start. However when I tried to migrate my work there I realized that - in comparison - `commander` actually provides a very friendly API with better documentation and proper maintenance. I'll therefore go with the comment in https://github.com/opencollective/opencollective-api/pull/7003#discussion_r777340020 suggesting to open an RFC to consolidate things.

PR #7004, opened 2021-12-29
Allow users to publish update without notifying any one

snitin315:
Refers https://github.com/opencollective/opencollective/issues/4093

Related Frontend PR - https://github.com/opencollective/opencollective-frontend/pull/7297

### DIFF
--- a/server/graphql/v2/object/Update.js
+++ b/server/graphql/v2/object/Update.js
@@ -87,7 +87,15 @@ const Update = new GraphQLObjectType({
           let hostedCollectivesCount = 0;
 
           if (audience === 'NO_ONE') {
-            return null;
+            return {
+              id: `${update.id}-${audience}`,
+              individuals: 0,
+              organizations: 0,
+              collectives: 0,
+              coreContributors: 0,
+              hosted: 0,
+              total: 0,
+            };
           }
 
           if (audience !== 'COLLECTIVE_ADMINS') {
--- a/server/graphql/v2/object/Update.js
+++ b/server/graphql/v2/object/Update.js
@@ -86,6 +86,10 @@ const Update = new GraphQLObjectType({
           let membersStats = {};
           let hostedCollectivesCount = 0;
 
+          if (audience === 'NO_ONE') {
+            return null;
+          }
+
           if (audience !== 'COLLECTIVE_ADMINS') {
             membersStats = await update.getAudienceMembersStats(audience);
           }
--- a/test/server/models/Update.test.js
+++ b/test/server/models/Update.test.js
@@ -160,8 +160,7 @@ describe('server/models/Update', () => {
       });
 
       it('returns an empty array when the audience in NO_ONE', async () => {
-        const emptyCollective = await fakeCollective();
-        const update = await fakeUpdate({ CollectiveId: emptyCollective.id, notificationAudience: 'NO_ONE' });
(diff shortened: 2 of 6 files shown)


server/graphql/v2/object/Update.js
```diff
@@ -86,6 +86,10 @@ const Update = new GraphQLObjectType({
           let membersStats = {};
           let hostedCollectivesCount = 0;
 
+          if (audience === 'NO_ONE') {
+            return null;
```
Betree, line 90: :ice_cream: I'm not a fan of returning `null` here. Returning `null` translates as "I don't want to reply". And since they're no `id` in the response, it means no caching is possible on the frontend side. Here it's ok to reply with the same `UpdateAudienceStats` object, with `0` for every values

Betree (approved):
Minor feedback below, but the code looks great

server/graphql/v2/object/Update.js
snitin315: @Betree Updated 👍🏻

test/server/models/Update.test.js
```diff
@@ -159,6 +159,13 @@ describe('server/models/Update', () => {
         expect(usersToNotify.length).to.eq(0);
       });
 
+      it('returns an empty array when the audience in NO_ONE', async () => {
+        const emptyCollective = await fakeCollective();
+        const update = await fakeUpdate({ CollectiveId: emptyCollective.id, notificationAudience: 'NO_ONE' });
```
Betree, line 164: `emptyCollective` has no member, so this test doesn't serve its purpose. Let's use a collective with some data:

```suggestion
        const update = await fakeUpdate({ CollectiveId: collective.id, notificationAudience: 'NO_ONE' });
```

test/server/models/Update.test.js
```diff
@@ -209,6 +216,13 @@ describe('server/models/Update', () => {
         expect(count).to.eq(0);
       });
 
+      it('returns 0 when the audience is NO_ONE', async () => {
+        const emptyCollective = await fakeCollective();
+        const update = await fakeUpdate({ CollectiveId: emptyCollective.id, notificationAudience: 'NO_ONE' });
```
Betree, line 221:
```suggestion
        const update = await fakeUpdate({ CollectiveId: collective.id, notificationAudience: 'NO_ONE' });
```

test/server/models/Update.test.js
Betree, line 164: `emptyCollective` has no member, so this test doesn't serve its purpose. Let's use a collective with some data:

```suggestion
        const update = await fakeUpdate({ CollectiveId: collective.id, notificationAudience: 'NO_ONE' });
```

test/server/models/Update.test.js
Betree, line 221: `emptyCollective` has no member, so this test doesn't serve its purpose. Let's use a collective with some data:

```suggestion
        const update = await fakeUpdate({ CollectiveId: collective.id, notificationAudience: 'NO_ONE' });
```

Betree (commented):
@snitin315 The tests do not seem to pass after updating to a real collective, could you check?

snitin315:
Yes, looking into it.

snitin315:
I have fixed the Update test. The only test failing now is `notifySubscribers >
         doesn't notify admin of host if unsubscribed` which is not reliable as per https://github.com/opencollective/opencollective/issues/5119

Betree (approved):
Perfect, thank you!

PR #7006, opened 2021-12-30
ThankYouEmail: Include accountants in recipients

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3626

### DIFF
--- a/server/lib/notifications.js
+++ b/server/lib/notifications.js
@@ -5,8 +5,9 @@ import debugLib from 'debug';
 import { get, remove } from 'lodash';
 import sanitizeHtml from 'sanitize-html';
 
-import { activities, channels } from '../constants';
+import { activities, channels, roles } from '../constants';
 import activityType from '../constants/activities';
+import { types as CollectiveType } from '../constants/collectives';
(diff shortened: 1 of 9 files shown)


server/lib/notifications.js
```diff
@@ -208,6 +209,31 @@ export async function notifyAdminsOfCollective(CollectiveId, activity, options =
...
+      `notifyAdminsAndAccountantsOfCollective> can't notify ${activity.type}: no collective found with id ${CollectiveId}`,
+    );
+  }
+
+  let usersToNotify = [];
+  if (collective.type === CollectiveType.USER && !collective.isIncognito) {
+    // Incognito profiles rely on the `Members` entry to know which user it belongs to
+    usersToNotify = [await collective.getUser()];
+  } else {
+    usersToNotify = await collective.getMembersUsers({ role: [roles.ACCOUNTANT, roles.ADMIN] });
```
znarf, line 226: What about Events and Projects? We implicitly rely on the the one from the parent.

And to anticipate, maybe in the future Events and Projects will be allowed to have extra ADMINs too.

LolyPopToyou: We
Was

PR #7012, opened 2022-01-03
Load order in `createHostFeeShareTransactions`

Betree:


server/models/Transaction.js
```diff
@@ -805,7 +805,7 @@ function defineModel() {
   ) => {
     let order;
     if (transaction.OrderId) {
-      order = await Transaction.findByPk(transaction.OrderId);
+      order = await models.Order.findByPk(transaction.OrderId);
```
znarf, line 808: Ouch.

server/lib/payments.js
```diff
@@ -914,11 +914,15 @@ export const getHostFeeSharePercent = async (order, host = null) => {
   const possibleValues = [];
 
   if (order) {
-    if (order.paymentMethod?.service === 'stripe' && order.paymentMethod?.type === 'creditcard') {
-      possibleValues.push(plan?.creditCardHostFeeSharePercent);
+    // Preload payment method
+    if (!order.paymentMethod && order.PaymentMethodId) {
```
znarf, line 918: Maybe simpler like this. But that's a detail. ```suggestion
    if (!order.paymentMethod) {
```

File: server/lib/payments.js [hunk 3]
```diff
@@ -914,11 +914,15 @@ export const getHostFeeSharePercent = async (order, host = null) => {
   const possibleValues = [];
 
   if (order) {
-    if (order.paymentMethod?.service === 'stripe' && order.paymentMethod?.type === 'creditcard') {
-      possibleValues.push(plan?.creditCardHostFeeSharePercent);
+    // Preload payment method
```
znarf, line 917: That's the right approach.

We need to be defensive and can't rely on the property being set.

znarf (approved):

PR #7013, opened 2022-01-03
Reduce Wise fallback cron checker frequency

kewitz:
Because of the increased amount of transactions being paid with Wise, the checker has been calling Wise's API more than 10K a day, mostly for checking if paid expenses didn't bounce back.

Since this is process is redundant to the main webhook event handler, I'm reducing the cron frequency from hourly to daily.

Related to: https://github.com/opencollective/opencollective/issues/5076


Betree (approved):
Cool! Sounds like a good quick fix

PR #7019, opened 2022-01-05
Refactor: Track platform tip in Orders table

kewitz:


znarf (commented):
Looks good.

No migration of past entries? Are we gonna support isFeesOnTop / platformFee forever?

While we're at it, would be interesting to track platform tip eligibility as a column in the Orders table.

Both could be done in follow up issues but let's make sure to track them if we go this way.

kewitz:
@znarf I think we can migrate past orders, no problem. And if we want to track platform tip eligibility I'm also doing it now.

migrations/20220105170000-add-platformTip-to-orders.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('Orders', 'platformTipAmount', { type: Sequelize.INTEGER, defaultValue: 0 });
+    await queryInterface.addColumn('OrderHistories', 'platformTipAmount', { type: Sequelize.INTEGER, defaultValue: 0 });
+    await queryInterface.addColumn('Orders', 'platformTipEligible', { type: Sequelize.BOOLEAN, defaultValue: false });
+    await queryInterface.addColumn('OrderHistories', 'platformTipEligible', {
+      type: Sequelize.BOOLEAN,
+      defaultValue: false,
```
znarf, line 10: I would default to `NULL`, it's problematic to mark it `FALSE` when we actually don't know or if it's unclear.

migrations/20220105170001-migrate-past-platformTip-orders.js
```diff
@@ -0,0 +1,20 @@
...
+module.exports = {
+  up: async queryInterface => {
+    await queryInterface.sequelize.query(
+      `
+        UPDATE
+          "Orders"
+        SET
+          "platformTipAmount" = ("data"->'platformFee')::int,
+          "platformTipEligible" = true
+        WHERE "data"->>'isFeesOnTop' = 'true';
```
znarf, line 12: That's not true, that will be confusing.

server/lib/payments.js
```diff
@@ -832,6 +833,10 @@ export const getHostFee = async (order, host = null) => {
 };
 
 export const isPlatformTipEligible = async (order, host = null) => {
+  if (order.collective?.data?.disablePlatformTips) {
```
znarf, line 836: What is this?

kewitz: This is the edge-case I added for BRL contributions related to Stripe's `aplication_fee` limitation.

znarf: Can we remove this now that we're landing https://github.com/opencollective/opencollective-api/pull/7020?

migrations/20220105170001-migrate-past-platformTip-orders.js
```diff
@@ -0,0 +1,37 @@
...
+
+module.exports = {
+  up: async queryInterface => {
+    await queryInterface.sequelize.query(
+      `
+        UPDATE
+          "Orders"
+        SET
+          "platformTipAmount" = ("data"->'platformFee')::int,
+          "platformTipEligible" = true
```
znarf, line 11: Sure, platformTipEligible is TRUE if a platformTip is set. Would be interesting to populate it when there is no platformTip.

Why not try to do it like suggested in https://github.com/opencollective/opencollective-api/pull/7019#discussion_r780403540 ?

migrations/20220105170001-migrate-past-platformTip-orders.js
```diff
@@ -0,0 +1,37 @@
...
+        WHERE "data"->>'isFeesOnTop' = 'true' AND ("data"->'platformFee')::int > 0;
+      `,
+    );
+
+    await queryInterface.sequelize.query(
+      `
+        UPDATE
+          "Orders" as o
+        SET
+          "platformTipAmount" = (o."data"->'platformFee')::int,
```
znarf, line 21: What if platformFee is not set? OrderId=511771 for example.

migrations/20220105170001-migrate-past-platformTip-orders.js
```diff
@@ -0,0 +1,37 @@
...
+          "platformTipEligible" = true
+        FROM
+          "Transactions" as t
+        WHERE
+          t."OrderId" = o."id"
+          AND t."kind" = 'CONTRIBUTION'
+          AND t."type" = 'CREDIT'
+          AND t."data"->>'platformTipEligible' = 'true'
```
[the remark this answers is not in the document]
kewitz: I don't think I follow what you mean.
Doesn't `platformTipEligible = false` implies that there's no platform tip attached to the order?

migrations/20220105170001-migrate-past-platformTip-orders.js
znarf: Is `(o."data"->'platformFee')::int` coalesced to `null` ?

migrations/20220105170001-migrate-past-platformTip-orders.js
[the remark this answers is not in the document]
znarf: Yes, but it's interesting to know that the Contribution was NOT eligible to platform tips with `platformTipEligible=false`.

migrations/20220105170001-migrate-past-platformTip-orders.js
kewitz: Exactly. This is valid:
```sql
SELECT NULL::int
```

PR #7020, opened 2022-01-06
Fix: Keep debt of Platform Tips and Host Fee Share for currencies that do not support application_fee

kewitz:
Resolves: https://github.com/opencollective/opencollective/issues/3863


znarf (approved):
Looks great.

PR #7025, opened 2022-01-07
Populate credit card expiryDate and migrate existing data

kewitz:
Resolves: https://github.com/opencollective/opencollective/issues/4058


znarf (approved):
Looks good to me. Ben for final review.

Betree (approved):
I haven't tested but the code looks good, it's super cool that we're tackling this inconsistency!

Betree:
@kewitz We missed a little thing with cards expiring during the current month that was difficult to spot through testing, I've documented the issue here: https://github.com/opencollective/opencollective/issues/5116

PR #7026, opened 2022-01-08
Prevent crashing when user denies stripe connection

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5103

stripe-crash-prevent


znarf (approved):
Looks good.

PR #7041, opened 2022-01-11
Submit on Behalf of Organization

kewitz:
Related to https://github.com/opencollective/opencollective/issues/4224.

GraphQL object update to support Organization details when submitting an expense on behalf.


server/graphql/v2/input/ExpenseInviteDraftInput.ts
```diff
@@ -13,6 +13,16 @@ import { ExpenseType } from '../enum/ExpenseType';
 import { LocationInput } from './LocationInput';
 import { PayoutMethodInput } from './PayoutMethodInput';
 
+const ExpenseInviteeOrg = new GraphQLInputObjectType({
+  name: 'ExpenseInviteeOrg',
```
znarf, line 17: We try to avoid short version of terms like "org". "Organization" is better.

PR #7046, opened 2022-01-12
Credit Card expiryDate follow-up

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5116.

### DIFF
--- a/cron/monthly/expiring-cards.js
+++ b/cron/monthly/expiring-cards.js
@@ -26,8 +26,8 @@ const fetchExpiringCreditCards = async () => {
       // Expiry Date is set to be the last second of the valid month.
       // This query will return all the creditcards that will expire by the end of the current month.
       expiryDate: {
-        [Op.gt]: moment.utc().startOf('month'),
-        [Op.lt]: moment.utc().endOf('month'),
+        [Op.gte]: moment.utc().startOf('month'),
+        [Op.lte]: moment.utc().endOf('month'),
       },
     },
     include: [
--- a/cron/monthly/expiring-cards.js
+++ b/cron/monthly/expiring-cards.js
@@ -2,16 +2,15 @@
 import '../../server/env';
 
 import config from 'config';
+import moment from 'moment';
 
 import logger from '../../server/lib/logger';
 import * as libPayments from '../../server/lib/payments';
-import models from '../../server/models';
+import models, { Op } from '../../server/models';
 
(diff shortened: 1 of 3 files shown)


server/graphql/v2/input/PaymentMethodInput.ts
```diff
@@ -84,7 +84,7 @@ export const getLegacyPaymentMethodFromPaymentMethodInput = async (
     };
     if (pm.creditCardInfo.expYear && pm.creditCardInfo.expMonth) {
       const { expYear, expMonth } = pm.creditCardInfo;
-      paymentMethod['expiryDate'] = moment.utc(`${expYear}-${expMonth}`, 'YYYY-MM');
+      paymentMethod['expiryDate'] = moment.utc(`${expYear}-${expMonth}`, 'YYYY-MM').endOf('month');
```
znarf, line 87: What about the one already added since the PR was deployed?

znarf: It's covered by the migration?

cron/monthly/expiring-cards.js
```diff
@@ -20,13 +19,15 @@ if (config.env === 'production' && date !== 7 && date !== 21 && !process.env.OFF
...
     where: {
       type: 'creditcard',
-      data: {
-        expMonth: month,
-        expYear: year,
+      // Expiry Date is set to be the last second of the valid month.
+      // This query will return all the creditcards that will expire by the end of the current month.
+      expiryDate: {
+        [Op.gt]: moment.utc().startOf('month'),
+        [Op.lt]: moment.utc().endOf('month'),
```
znarf, line 30: What are the precise values being used in the comparaison practically?

Could it be that we end up with a condition like expiryDate < `2022-01-31 23:59:59` and expiryDate=2022-01-31 23:59:59. Not returning entries.

kewitz: Moment will return the timestamp to the last millisecond, but I'll update to `gte` and `lte` to be more explicit.

server/graphql/v2/input/PaymentMethodInput.ts
kewitz: Yes, the migration will update existing credit cards.

Betree (approved):
The migration worked well on my local data. The code looks good for the rest

PR #7057, opened 2022-01-14
Fix tip check with fixed amounts

Betree:
It seems that https://github.com/opencollective/opencollective-api/pull/7019 introduced a regression in the scenario of fixed amount tier + platform tip. We were not deducing the tip from the expected amount anymore, which resulted in an error like:
> This tier uses a fixed amount. Order total must be $xx. You set $yy

I randomly discovered this while testing the "move tiers" feature, it seems that no user has been impacted by this.


kewitz (approved):
Nice catch!

PR #7065, opened 2022-01-17
Specify tier when adding funds

Betree:
Requirement for https://github.com/opencollective/opencollective-frontend/pull/7352


SudharakaP (commented):
💯

PR #7074, opened 2022-01-25
Add Support for Parent Updates to be sent to Project/Event Contributors

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/4673

send-emails-to-all

### DIFF
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -333,7 +333,7 @@ export const usersToNotifyForUpdateSQLQuery = `
         --- Include child collective's contributors
         c."ParentCollectiveId" IS NOT NULL
         AND c.id = m."CollectiveId"
-        AND m."role" IN ('BACKER', 'ATTENDEE')
+        AND m."role" IN (:targetRoles)
       )
     )
     GROUP BY mc.id
@@ -395,7 +395,7 @@ export const countMembersToNotifyForUpdateSQLQuery = `
         --- Include child collective's contributors
         collective."ParentCollectiveId" IS NOT NULL
         AND collective.id = m."CollectiveId"
-        AND m."role" IN ('BACKER', 'ATTENDEE')
+        AND m."role" IN (:targetRoles)
       )
     )
     GROUP BY mc.id
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -293,7 +293,7 @@ export const usersToNotifyForUpdateSQLQuery = `
   WITH collective AS (
     SELECT c.*
     FROM "Collectives" c
-    WHERE id = :collectiveId
+    WHERE id = :collectiveId OR "ParentCollectiveId" = :collectiveId
   ), hosted_collectives AS (
     SELECT hc.*
     FROM "Collectives" hc
@@ -310,7 +310,6 @@ export const usersToNotifyForUpdateSQLQuery = `
   ), member_collectives AS (
     SELECT mc.*
     FROM "Members" m
-    FULL OUTER JOIN "Collectives" c ON (c.id = :collectiveId OR c."ParentCollectiveId" = :collectiveId)
     INNER JOIN "Collectives" mc ON m."MemberCollectiveId" = mc.id
     CROSS JOIN collective
     WHERE m."deletedAt" IS NULL
@@ -331,8 +330,8 @@ export const usersToNotifyForUpdateSQLQuery = `
         AND m."role" IN ('ADMIN', 'MEMBER')
       ) OR (
         --- Include child collective's contributors
-        c."ParentCollectiveId" IS NOT NULL
-        AND c.id = m."CollectiveId"
+        collective."ParentCollectiveId" = :collectiveId
+        AND collective.id = m."CollectiveId"
         AND m."role" IN (:targetRoles)
       )
     )
@@ -393,7 +392,7 @@ export const countMembersToNotifyForUpdateSQLQuery = `
         AND m."role" IN ('ADMIN', 'MEMBER')
       ) OR (
         --- Include child collective's contributors
-        collective."ParentCollectiveId" IS NOT NULL
+        collective."ParentCollectiveId" = :collectiveId
         AND collective.id = m."CollectiveId"
         AND m."role" IN (:targetRoles)
       )
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -310,6 +310,7 @@ export const usersToNotifyForUpdateSQLQuery = `
   ), member_collectives AS (
     SELECT mc.*
     FROM "Members" m
+    FULL OUTER JOIN "Collectives" c ON (c.id = :collectiveId OR c."ParentCollectiveId" = :collectiveId)
     INNER JOIN "Collectives" mc ON m."MemberCollectiveId" = mc.id
     CROSS JOIN collective
     WHERE m."deletedAt" IS NULL
@@ -328,6 +329,11 @@ export const usersToNotifyForUpdateSQLQuery = `
         collective."ParentCollectiveId" IS NOT NULL
         AND m."CollectiveId" = collective."ParentCollectiveId"
         AND m."role" IN ('ADMIN', 'MEMBER')
+      ) OR (
+        --- Include child collective's contributors
+        c."ParentCollectiveId" IS NOT NULL
+        AND c.id = m."CollectiveId"
+        AND m."role" IN ('BACKER', 'ATTENDEE')
       )
     )
     GROUP BY mc.id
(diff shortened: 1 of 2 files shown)


SudharakaP:
Here I am sending the email to all backers, attendees of events and followers. Let me know if you see any issue with this one. Thanks.

znarf:
@SudharakaP Looks good. Can you add a test for this?

SudharakaP:
> @SudharakaP Looks good. Can you add a test for this?

Yep, working on it. 👍🏽

SudharakaP:
@znarf: Added the tests to check for private and public update on the parent collective.

znarf:
Can you add a test that check that the email are properly sent to the right addresses?

In some tests, we're using something named `emailSendMessageSpy`.

SudharakaP:
> Can you add a test that check that the email are properly sent to the right addresses?
> 
> In some tests, we're using something named `emailSendMessageSpy`.

This is done. Although it seems that in my thinking the better fit here is to use the `expectAllEmailsFrom` function since we already use this in this test suite. So I've went ahead and used that.

server/lib/queries.js
```diff
@@ -328,6 +329,11 @@ export const usersToNotifyForUpdateSQLQuery = `
         collective."ParentCollectiveId" IS NOT NULL
         AND m."CollectiveId" = collective."ParentCollectiveId"
         AND m."role" IN ('ADMIN', 'MEMBER')
+      ) OR (
+        --- Include child collective's contributors
+        c."ParentCollectiveId" IS NOT NULL
+        AND c.id = m."CollectiveId"
+        AND m."role" IN ('BACKER', 'ATTENDEE')
```
znarf, line 336: Why invent something new instead of using `targetRoles`?

server/lib/queries.js
```diff
@@ -375,16 +381,21 @@ export const countMembersToNotifyForUpdateSQLQuery = `
...
-        -- Inlcude parent collective's core contributors
+        -- Include parent collective's core contributors
         collective."ParentCollectiveId" IS NOT NULL
         AND m."CollectiveId" = collective."ParentCollectiveId"
         AND m."role" IN ('ADMIN', 'MEMBER')
+      ) OR (
+        --- Include child collective's contributors
+        collective."ParentCollectiveId" IS NOT NULL
+        AND collective.id = m."CollectiveId"
+        AND m."role" IN ('BACKER', 'ATTENDEE')
```
znarf, line 398: Same as before regarding targetRoles.

server/lib/queries.js
SudharakaP: Makes sense. I've changed that. 👍🏽

znarf (approved):
Looks good. Great if @Betree can have a quick look too.

test/server/models/Update.test.js
```diff
@@ -195,16 +198,60 @@ describe('server/models/Update', () => {
...
-        expectAllEmailsFrom(getOrganizationAdminUsers());
+        expectAllEmailsFrom(parentCollectiveAdmins, receivedEmails);
+        expectAllEmailsFrom(collectiveAdmins, receivedEmails, receivedEmails);
+        expectAllEmailsFrom(individualBackersUsers, receivedEmails);
+        expectAllEmailsFrom(getOrganizationAdminUsers(), receivedEmails);
         expect(usersToNotify.length).to.eq(expectedPrivateTotal);
       });
+
+      it('Notifies child collective users when parent collective public update is made', async () => {
+        collective.remove;
```
Betree, line 209: What is this doing?

test/server/models/Update.test.js
```diff
@@ -195,16 +198,60 @@ describe('server/models/Update', () => {
...
+        const receivedEmails = usersToNotify.map(u => u.email);
+
+        expectAllEmailsFrom(parentCollectiveAdmins, receivedEmails);
+        expectAllEmailsFrom(parentCollectiveBackers, receivedEmails);
+        expectAllEmailsFrom(parentCollectiveFollowers, receivedEmails);
+        expectAllEmailsFrom(individualBackersUsers, receivedEmails);
+        expectAllEmailsFrom(getOrganizationAdminUsers(), receivedEmails);
+
+        // Number of collective admins who are backers
+        const numberOfCollectiveAdminsAddedToBackers = 2;
```
Betree, line 221: Let's move that to a shared variable next to https://github.com/opencollective/opencollective-api/blob/07deff0b9b6f049caac10883448eab9f57705872/test/server/models/Update.test.js#L142, otherwise it's unclear where this number is coming from.

test/server/models/Update.test.js
```diff
@@ -195,16 +198,60 @@ describe('server/models/Update', () => {
...
+
+        expect(usersToNotify.length).to.eq(
+          parentCollectiveAdmins.length +
+            parentCollectiveBackers.length +
+            parentCollectiveFollowers.length +
+            collectiveFollowers.length +
+            individualBackersUsers.length +
+            countAdminsOfMemberOrganizations() +
+            numberOfCollectiveAdminsAddedToBackers,
```
Betree, line 230: Shouldn't that be:

```suggestion
            countAdminsOfMemberOrganizations() -
            numberOfCollectiveAdminsAddedToBackers,
```

The goal of this "admins being backers" thing is to make sure we don't double count them when they have multiple roles. If I have 2 admins that are also backers, I expect the list to have 2 entries, not 4. So `admins.length + backers.length - numberOfCollectiveAdminsAddedToBackers`.

server/lib/queries.js
```diff
@@ -310,6 +310,7 @@ export const usersToNotifyForUpdateSQLQuery = `
   ), member_collectives AS (
     SELECT mc.*
     FROM "Members" m
+    FULL OUTER JOIN "Collectives" c ON (c.id = :collectiveId OR c."ParentCollectiveId" = :collectiveId)
```
Betree, line 313: Why do we need to re-include the collective (`c.id = :collectiveId`)?

And can't we update this query without a new join + condition simply by updating the `WITH collective AS (` to:

```sql
WITH collective AS (
    SELECT c.*
    FROM "Collectives" c
    WHERE id = :collectiveId OR "ParentCollectiveId" = :collectiveId
```

server/lib/queries.js
```diff
@@ -375,16 +381,21 @@ export const countMembersToNotifyForUpdateSQLQuery = `
...
         m."CollectiveId" = collective.id AND m."role" IN (:targetRoles)
       ) OR (
-        -- Inlcude parent collective's core contributors
+        -- Include parent collective's core contributors
         collective."ParentCollectiveId" IS NOT NULL
         AND m."CollectiveId" = collective."ParentCollectiveId"
         AND m."role" IN ('ADMIN', 'MEMBER')
+      ) OR (
+        --- Include child collective's contributors
+        collective."ParentCollectiveId" IS NOT NULL
```
Betree, line 396: :ice_cream:  Same effect, but I would find it a bit clearer with:

```suggestion
        collective."ParentCollectiveId" = :collectiveId
```

Betree (commented):
Looking good, but there are some details we should look at

Betree:
@SudharakaP It would be nice for you to use `EXPLAIN ANALYZE` to monitor the complexity of the queries while updating them. These are really expensive pieces of code when used with large hosts like OSC. We could end up with timeouts if the complexity increases too much

test/server/models/Update.test.js
SudharakaP: Ops, seems like a mistakenly added part. I've removed it. Thanks.

SudharakaP:
> @SudharakaP It would be nice for you to use `EXPLAIN ANALYZE` to monitor the complexity of the queries while updating them. These are really expensive pieces of code when used with large hosts like OSC. We could end up with timeouts if the complexity increases too much

@Betree : Thanks for mentioning this. I'm looking into this and trying to understand what `EXPLAIN ANALYZE` returns. To me it seems that the complexity has increased a bit at the group key `mc_2.id`; https://explain.depesz.com/s/bv5g#html. Thinking of a way to reduce this. 🤔

Betree:
@SudharakaP I can recommend this tool to look at the results in a nice way: https://explain.dalibo.com. For tests to be relevant, they have to be done on large datasets. I recommend running the query against staging/prod, using OSC as the sending collective.

It's ok to have a small difference, what we want to prevent is complexity skyrocketing because of things like an additional JOIN not using an index. If that's helpful for you, we can plan a call and go through the testing steps together. Let me know :slightly_smiling_face:

SudharakaP:
> @SudharakaP I can recommend this tool to look at the results in a nice way: https://explain.dalibo.com. For tests to be relevant, they have to be done on large datasets. I recommend running the query against staging/prod, using OSC as the sending collective.
> 
> It's ok to have a small difference, what we want to prevent is complexity skyrocketing because of things like an additional JOIN not using an index. If that's helpful for you, we can plan a call and go through the testing steps together. Let me know slightly_smiling_face

Thanks, a call will be better. I've DMed you. 🙂

znarf:
Goals for Q1 Sprint 1 Week 6:
- [ ] Study SQL performance to finalize

Betree (approved):
**For `usersToNotifyForUpdateSQLQuery`**

We went from:
```
Group  (cost=185705.16..186102.09 rows=396933 width=1061) (actual time=890.909..893.413 rows=5132 loops=1)
```

To:
```
Group  (cost=186418.22..186815.16 rows=396933 width=1061) (actual time=900.692..910.269 rows=5194 loops=1)
```

**For `countMembersToNotifyForUpdateSQLQuery`**

We went from:

```
GroupAggregate  (cost=68.87..68.87 rows=1 width=40) (actual time=3.755..3.788 rows=5 loops=1)
```

To:
```
GroupAggregate  (cost=678.16..678.17 rows=2 width=40) (actual time=4.137..4.192 rows=5 loops=1)
```

# Conclusion

- No impact on the first query, complexity is almost the same, and execution time didn't change
- x10 complexity on the second query, but the execution time for what we tested (`opencollective`, `babel`, `opensource`) was stable and consistent.

Ok to merge and deploy :rocket:

PR #7076, opened 2022-01-25
Disable Captcha for non-credit card orders

kewitz:
Resolves: https://github.com/opencollective/opencollective/issues/5151

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -12,6 +12,7 @@ import CAPTCHA_PROVIDERS from '../../../constants/captcha-providers';
 import { types } from '../../../constants/collectives';
 import FEATURE from '../../../constants/feature';
 import status from '../../../constants/order_status';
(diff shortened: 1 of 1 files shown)


server/graphql/v1/mutations/orders.js
```diff
@@ -225,15 +226,16 @@ const checkAndUpdateProfileInfo = async (order, fromAccount, isGuest, currency)
 async function checkCaptcha(order, remoteUser, reqIp) {
   const requestedProvider = order.guestInfo?.captcha?.provider;
   const isCaptchaEnabled = parseToBoolean(config.captcha?.enabled);
+  const isCreditCardOrder = order.paymentMethod.type === PAYMENT_METHOD_TYPE.CREDITCARD;
```
znarf, line 229: Should we be defensive on `order.paymentMethod` being null?

I don't know here but in some cases it's null for Bank Transfers.

kewitz: True. I'll add a conditional operator here.

PR #7077, opened 2022-01-25
Move Collective budget Stats to GQL V2

kewitz:
Fixes: https://github.com/opencollective/opencollective/issues/4449


server/graphql/v2/object/AccountStats.js
```diff
@@ -149,6 +162,20 @@ export const AccountStats = new GraphQLObjectType({
...
+      totalNetAmountReceived: {
+        description: 'Total net amount received',
+        type: new GraphQLNonNull(Amount),
+        async resolve(collective) {
+          const value = await collective.getTotalNetAmountReceived();
+          return { value, currency: collective.currency };
+        },
+      },
+      activeRecurringContributions: {
+        type: GraphQLJSON,
```
Betree, line 174: I missed that part, we should have a proper type for this. `GraphQLJSON` is too broad

PR #7090, opened 2022-01-27
Move from graphql-iso-date to graphql-scalars

znarf:
graphql-iso-date is not actively maintained anymore and has moved to graphql-scalars.


znarf:
@Betree fyi, merging this.

PR #7095, opened 2022-01-28
Relax Condition to Obtain Transaction to Refund

SudharakaP:
A small bug was introduced at, https://github.com/opencollective/opencollective-api/pull/6999. By fixing the condition `type: TransactionTypes.CREDIT` we are assuming that for all expenses a `CREDIT` exists. However when the expense is submitted from a collective to itself  and then when we try to refund it (supported by, https://github.com/opencollective/opencollective-api/pull/6985) this will give an error since for these kind of expenses a `CREDIT` does not exist. 

Related Slack Discussion: https://opencollective.slack.com/archives/CEZUS9WH3/p1643365892438949?thread_ts=1643361163.425029&cid=CEZUS9WH3


Betree (approved):
Lgtm!

PR #7107, opened 2022-01-31
Allow delete expense for SPAM

Betree:


znarf:
Small question, where is it coming from?

In some cases, it's handy to have SPAM expenses not deleted so we can have a look at them through the interface.

Betree:
@znarf It's coming from https://opencollective.slack.com/archives/G46PNRCGP/p1643620342761559. From the collective admin point of view, I think it's expected that SPAM gets removed from their pages.

znarf:
Except the single expense page, in which screen is it appearing?

Betree:
@znarf Only there I think, but the feedback was:
> a facebook open source admin marked an expense as spam, but then realized that his name appears in the trail as having marked the expense as spam and he'd rather not have the spammer see this

znarf:
Ok, if we allow deletion of `rejected` then why not `spam`. But we have to remember that as soon as we delete, we loose the single expense page and the context (change of status, history, comments).

PR #7113, opened 2022-02-01
Make sure only host admins can refund added funds

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5173

### From the Perspective of Collective Admins

image

### From the Perspective of a Host Admin

image

### DIFF
--- a/server/graphql/common/transactions.ts
+++ b/server/graphql/common/transactions.ts
@@ -3,6 +3,7 @@ import moment from 'moment';
 
 import { roles } from '../../constants';
 import orderStatus from '../../constants/order_status';
+import { TransactionKind } from '../../constants/transaction-kind';
 import { TransactionTypes } from '../../constants/transactions';
 import models from '../../models';
 
@@ -121,10 +122,16 @@ export const canRefund = async (
   if (transaction.type !== TransactionTypes.CREDIT || transaction.OrderId === null || transaction.isRefund === true) {
     return false;
   }
+
   const timeLimit = moment().subtract(30, 'd');
   const createdAtMoment = moment(transaction.createdAt);
   const transactionIsOlderThanThirtyDays = createdAtMoment < timeLimit;
-  if (transactionIsOlderThanThirtyDays) {
+  /*
+   * 1) We only allow the transaction to be refunded by Collective admins if it's within 30 days.
+   *
+   * 2) To ensure proper accounting, we only allow added funds to be refunded by Host admins and never by Collective admins.
+   */
+  if (transactionIsOlderThanThirtyDays || transaction.kind === TransactionKind.ADDED_FUNDS) {
     return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin]);
   } else {
     return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin, isPayeeCollectiveAdmin]);
--- a/test/server/graphql/common/transactions.test.js
+++ b/test/server/graphql/common/transactions.test.js
@@ -2,7 +2,8 @@ import { expect } from 'chai';
 import { useFakeTimers } from 'sinon';
 
 import { roles } from '../../../../server/constants';
-import { canDownloadInvoice, canRefund } from '../../../../server/graphql/common/transactions';
+import { TransactionKind } from '../../../../server/constants/transaction-kind';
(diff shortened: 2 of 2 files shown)


server/graphql/common/transactions.ts
```diff
@@ -121,11 +122,21 @@ export const canRefund = async (
...
+
+  const isPayeeRootOrHostAdmin = await remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin]);
+  /*
+   * We do not allow added funds to be refunded by collective admins as it can lead
+   * to issues and confusions. The host admins will have no idea that the refund was initiated so
+   * the money will not be refunded.
+   */
+  if (transaction.kind === TransactionKind.ADDED_FUNDS && !isPayeeRootOrHostAdmin) {
+    return false;
+  }
```
znarf, line 134:
```suggestion
  if (transaction.kind === TransactionKind.ADDED_FUNDS) {
    return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin]);
  }
  
```

server/graphql/common/transactions.ts
```diff
@@ -121,11 +122,21 @@ export const canRefund = async (
...
   const timeLimit = moment().subtract(30, 'd');
   const createdAtMoment = moment(transaction.createdAt);
   const transactionIsOlderThanThirtyDays = createdAtMoment < timeLimit;
   if (transactionIsOlderThanThirtyDays) {
-    return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin]);
+    return isPayeeRootOrHostAdmin;
```
znarf, line 139:
```suggestion
    return remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin]);
```

server/graphql/common/transactions.ts
```diff
@@ -121,11 +122,21 @@ export const canRefund = async (
   if (transaction.type !== TransactionTypes.CREDIT || transaction.OrderId === null || transaction.isRefund === true) {
     return false;
   }
+
+  const isPayeeRootOrHostAdmin = await remoteUserMeetsOneCondition(req, transaction, [isRoot, isPayeeHostAdmin]);
```
znarf, line 126:
```suggestion
```

SudharakaP:
> Proposed code would have worked but it's making things more complicated than they should.
> 
> Made some proposals.

Thanks for the suggestions. I've changed the code accordingly. 👍🏽

znarf (changes requested):
Why not simply do the following?

```transactionIsOlderThanThirtyDays || transaction.kind === TransactionKind.ADDED_FUNDS``` ?

File: server/graphql/common/transactions.ts [hunk 4]
```diff
@@ -121,10 +122,14 @@ export const canRefund = async (
...
     return false;
   }
+
   const timeLimit = moment().subtract(30, 'd');
   const createdAtMoment = moment(transaction.createdAt);
   const transactionIsOlderThanThirtyDays = createdAtMoment < timeLimit;
-  if (transactionIsOlderThanThirtyDays) {
+  /*
+   * To ensure proper accounting, we only allow added funds to be refunded by Host admins and never by Collective admins.
+   */
```
znarf, line 131: The description is not accurate anymore as it doesn't explain `transactionIsOlderThanThirtyDays`.

znarf (changes requested):
Code comment is now outdated.

server/graphql/common/transactions.ts
SudharakaP: I've added another line to emphasize that the `transactionIsOlderThanThirtyDays` condition as well. We didn't have a comment about this part before, so I am not exactly sure if there's a reason for this 30 day limitation; if so let me know and I'll update the comment. 🤔

SudharakaP:
> Why not simply do the following?
> 
> `transactionIsOlderThanThirtyDays || transaction.kind === TransactionKind.ADDED_FUNDS` ?

@znarf : Thanks. This is done. I also noticed that we should add some tests for this and I've done that as well. When adding tests I've noticed we use a wrong function to test the transaction rejections which I've opened a new PR; 
https://github.com/opencollective/opencollective-api/pull/7115

znarf (approved):

PR #7115, opened 2022-02-02
Use `canReject` function instead of `canRefund` in API Tests

SudharakaP:
A small regression caused long time ago may which we haven't noticed. We are using `canRefund` function instead of the `canReject` function in some of our API tests which check for transaction rejections.


znarf:
@SudharakaP If it's a regression, when was it introduced?

SudharakaP:
> @SudharakaP If it's a regression, when was it introduced?

Sorry, my mistake. Seems like it was there since the test was introduced at, https://github.com/opencollective/opencollective-api/pull/4648. So it's not a regression, but rather a mistake in the initial addition of this test I believe.

znarf (approved):

PR #7116, opened 2022-02-02
Base support for multi-currency expenses

Betree:
Related to https://github.com/opencollective/opencollective/issues/5174

WIP


server/graphql/common/expenses.ts
```diff
@@ -1177,31 +1246,36 @@ const checkHasBalanceToPayExpense = async (
...
     }
   } else {
     throw new Error(`Expense fee payer "${expense.feesPayer}" not supported yet`);
   }
 
-  // Ensure the collective has enough funds to pay the expense
-  if (totalAmountToPay > balance) {
+  // Ensure the collective has enough funds to cover the fees for this expense, with an error margin of 20% of the expense amount
+  // to account for fluctuating rates. Example: to pay for a $100 expense in euros, the collective needs to have at least $120.
+  if (balanceInExpenseCurrency < getMinExpectedBalance(totalAmountToPay)) {
```
SudharakaP, line 1268: Just a small suggestion, maybe for the future. 😉 My thinking is that this check might be more useful to do in the expense summary page and the host dashboard expenses section or more simply the pay expense modal itself and then if the check fails display a warning/error box in the modal itself with the error and disable the `Mark as Paid` button. My reasoning is as follows; 

It would be nice for the admin who is going to pay the expense to know about this limitation before they go through the whole flow and click on the `Mark as Paid` button. Also having this kinda long error message on a toast seems a bit weird as the toast is small and on one side of the screen and disappears after a while, my thinking is it's better suited for a more front and center place. 

We don't necessary need to call the API and as we just need an approximate conversion we can always use `cryptonator` which we already use in the front-end; https://github.com/opencollective/opencollective-frontend/blob/97ae7d6941fe0ba187e42070e7cdd49768e40a3c/server/content-security-policy.js#L46.

SudharakaP (commented):
Nicely done. 🚀  I have just one suggestion; more on the design side maybe about the way the 20% limitation is handled.

server/graphql/common/expenses.ts
Betree: The way it's handled here is in line with what we do for same-currency balance checks. But yes, I agree that we could do a better job at surfacing it in the frontend. Having that requires more info though, especially the part from https://github.com/opencollective/opencollective/issues/5175 which is about surfacing the expense amount in both currencies.
I don't want to spend time on this now because this rule is very approximate and subject to change. That's the kind of things we'll need to iterate on for the public release, but we don't need to have a perfect UX for the core team release.

> Also we might not necessarily need to call the API and as we just need an approximate conversion we can always use something like; https://github.com/fawazahmed0/currency-api#readme I think similar to what we do for [crypto conversion on front-end](https://github.com/opencollective/opencollective-frontend/blob/97ae7d6941fe0ba187e42070e7cdd49768e40a3c/server/content-security-policy.js#L46) using Cryptonator.

That's part of what's required for https://github.com/opencollective/opencollective/issues/5175, to provide an estimate based on the currency conversion. We don't need to introduce a new API for that, we already have the ability to fetch from https://data.fixer.io with some caching in place in the database: https://github.com/opencollective/opencollective-api/blob/a543c0ff34c7da7575c2650c7379aaea1b9bbec2/server/lib/currency.ts#L102

PR #7117, opened 2022-02-02
Make hostFeePercent optional in addFunds mutation

znarf:
While working on a project using the API, I realized that it was not intuitive to force people to explicitly pass `hostFeePercent`. It's more intuitive to rely on the default.

Also taking this as an opportunity to introduce some configuration on the host or on the collective if we want custom values: `addedFundsHostFeePercent`.


q805:
可以拉取請求合併嗎？謝謝

Betree (approved):
Looking good :+1:

PR #7119, opened 2022-02-02
Implement base Collective Policy structure and "Expense Author cannot Approve" policy

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5184


server/graphql/common/expenses.ts
```diff
@@ -300,6 +301,16 @@ export const canApprove: ExpensePermissionEvaluator = async (req, expense, optio
     }
     return false;
   } else {
+    const collective = expense.collective || (await models.Collective.findByPk(expense.CollectiveId));
+    if (collective.hasPolicy(POLICIES.EXPENSE_AUTHOR_CANNOT_APPROVE) && req.remoteUser.id === expense.UserId) {
```
Betree, line 305: We must try to stick with loaders as much as we can in these permissions evaluators. This one is called in the expenses list, a loader will prevent calling 15x the DB

```suggestion
    expense.collective = expense.collective || (await req.loaders.Collective.byId.load(expense.CollectiveId));
    if (expense.collective?.hasPolicy(POLICIES.EXPENSE_AUTHOR_CANNOT_APPROVE) && req.remoteUser.id === expense.UserId) {
```

server/models/Collective.js
```diff
@@ -3184,6 +3186,20 @@ function defineModel() {
...
+      }
+    }
+
+    return this.update({ data: { ...this.data, policies: uniq(policies) } });
+  };
+
+  Collective.prototype.hasPolicy = function (policy) {
+    return Boolean(this.data?.policies?.includes(policy));
+  };
+
```
Betree, line 3202: This is great!

Betree (commented):
I love the way you've implemented this generic policy flag at the collective level, that's a simple and solid layer for implementing new things in the future!

server/graphql/common/expenses.ts
kewitz: Perfect! I feel like I always forget about loaders and you're always catching that. 😂

server/models/Collective.js
kewitz: Thanks. I added the minimum amount of methods, this can change a bit in the future.

PR #7126, opened 2022-02-03
Add top level queries in GraphQL v2 for consistency

znarf:
Fix https://github.com/opencollective/opencollective/issues/5188


Betree (commented):
Cool, updating the schemas would be nice to help see the impact of this change

znarf:
@Betree I updated the schema, always a good idea when the API changes. Unfortunately, as I decided to order the queries alphabetically in the code, it looks like a huge change.

PR #7129, opened 2022-02-04
Add the Payment Processor Fee Directly to Expense Amount When Payee is Paying the Fees

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5194

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -202,14 +202,14 @@ export const buildRefundForTransaction = (t, user, data, refundedPaymentProcesso
     } else if (feesPayer === ExpenseFeesPayer.PAYEE) {
       if (refundedPaymentProcessorFee) {
         // If the fee gets refunded, we add it as a positive value on the refund transactions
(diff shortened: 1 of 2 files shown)


SudharakaP:
@Betree : If this is in agreement, I'll add a migration script as well to convert the existing records (and also we need to adjust the PR which uses this; https://github.com/opencollective/opencollective-frontend/pull/7391). I just didn't yet as I wasn't exactly sure if there's a specific reason you added this payment processor fee on the refunds. But let me know. 😉

Betree (changes requested):
The ledger was carefully designed to record things this way. See my feedback on https://github.com/opencollective/opencollective/issues/5194#issuecomment-1031207925 for the why.

Let's discuss directly on the issue if you still think this is not correct.

SudharakaP:
Closing this as per; https://github.com/opencollective/opencollective/issues/5194#issuecomment-1031256779

PR #7132, opened 2022-02-04
Add Expense "Author cannot Approve" Policy Mutations and Resolvers

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5199

### DIFF
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -28,6 +28,7 @@ import {
 } from '../enum';
 import { PaymentMethodService } from '../enum/PaymentMethodService';
 import { PaymentMethodType } from '../enum/PaymentMethodType';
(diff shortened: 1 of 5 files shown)


server/graphql/v2/interface/Account.js
```diff
@@ -574,6 +575,15 @@ const accountFieldsDefinition = () => ({
       };
     },
   },
+  policies: {
+    type: new GraphQLList(Policy),
+    async resolve(account, _, req) {
+      if (req.remoteUser?.isAdminOfCollective(account)) {
+        return account.data?.policies || [];
+      }
+      return null;
```
Betree, line 584: It won't be useful for now, but I think this field should be public. Everyone should be able to list the policies of the collective, as it's not private info and it might concern other people than collective admins in the future

Betree (approved):

PR #7136, opened 2022-02-07
fix: correct `InvalidToken` usage in email confirmation

snitin315:
For https://github.com/opencollective/opencollective-frontend/pull/7450

The usage was incorrect, the second argument should be the `code` string.

https://github.com/opencollective/opencollective-api/blob/7190454f9b123f86193bdcf8e5739b417e75ca93/server/graphql/errors.ts#L49-L53


znarf (approved):
Looks good to me!

@Betree for second approval.

Betree (approved):
Looks good to me

PR #7142, opened 2022-02-08
chore: update `sequelize` to the latest version

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/5215

There was also an unknown bug in `sequelize` v6.16.0, I created a PR to point it out and fix the issue in the `sequelize` repository itself - https://github.com/sequelize/sequelize/pull/14071 https://github.com/sequelize/sequelize/pull/14071#issuecomment-1033487156

The change was accepted and a new patch version `v6.16.1` is now released.

## 

There was another unknown regression, I reported it in https://github.com/sequelize/sequelize/issues/14129

The issue was fixed and a new patch version `v6.16.3` is now released.

### DIFF
--- a/server/models/PaypalPlan.ts
+++ b/server/models/PaypalPlan.ts
@@ -29,7 +29,7 @@ interface PaypalPlanCreateWithProductAttributes extends PaypalPlanCommonCreateAt
   product?: PaypalProductCreateAttributes;
 }
 
-type PaypalPlanCreateAttributes = PaypalPlanCreateWithProductIdAttributes & PaypalPlanCreateWithProductAttributes;
+type PaypalPlanCreateAttributes = PaypalPlanCreateWithProductIdAttributes | PaypalPlanCreateWithProductAttributes;
 
 class PaypalPlan extends Model<PaypalPlanAttributes, PaypalPlanCreateAttributes> implements PaypalPlanAttributes {
   id: string;
--- a/server/models/PaypalPlan.ts
+++ b/server/models/PaypalPlan.ts
@@ -29,7 +29,7 @@ interface PaypalPlanCreateWithProductAttributes extends PaypalPlanCommonCreateAt
   product?: PaypalProductCreateAttributes;
 }
 
-type PaypalPlanCreateAttributes = PaypalPlanCreateWithProductIdAttributes | PaypalPlanCreateWithProductAttributes;
+type PaypalPlanCreateAttributes = PaypalPlanCreateWithProductIdAttributes & PaypalPlanCreateWithProductAttributes;
 
 class PaypalPlan extends Model<PaypalPlanAttributes, PaypalPlanCreateAttributes> implements PaypalPlanAttributes {
   id: string;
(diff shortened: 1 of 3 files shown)


Vic23M (approved):
Fusión de cambios

znarf (approved):
Thank you ! That looks good.

server/models/PaypalPlan.ts
```diff
@@ -29,7 +29,7 @@ interface PaypalPlanCreateWithProductAttributes extends PaypalPlanCommonCreateAt
   product?: PaypalProductCreateAttributes;
 }
 
-type PaypalPlanCreateAttributes = PaypalPlanCreateWithProductIdAttributes | PaypalPlanCreateWithProductAttributes;
+type PaypalPlanCreateAttributes = PaypalPlanCreateWithProductIdAttributes & PaypalPlanCreateWithProductAttributes;
```
znarf, line 32: @Betree does that make sense for you?

Betree: Not really. A union makes sense there because you expect the create attributes to include either a `product` **or** a `ProductId`. I don't see why we'd need an intersection here. @snitin315 could you explain the rationale?

snitin315: @Betree `npm run type:check` is failing with `union`. 

Screenshot 2022-02-16 at 6 53 13 AM

znarf: @snitin315 according to @Betree, we have to make it work with `union`, randomly changing to `intersection` is certainly not acceptable.

znarf (changes requested):
Change required.

server/models/PaypalPlan.ts
snitin315: Looks like a bug in `sequelize` itself. They [introduced a new way of creating a model](https://github.com/sequelize/sequelize/blob/v6/docs/manual/other-topics/typescript.md#usage) with the help of `InferAttributes`, `InferCreationAttributes`, and `CreationOptional` utilities.

That would change the model something like this - 

```js
class PaypalPlan 
    extends Model<InferAttributes<PaypalPlan, { omit: 'product' }>, InferCreationAttributes<PaypalPlan>> 
    implements PaypalPlanAttributes 
{
  id: string;
  ProductId: CreationOptional<string>;
  product?: CreationOptional<PaypalProductCreateAttributes>;
  currency: string;
  interval: string;
  amount: number;
  createdAt: Date;
  updatedAt: Date;
  deletedAt: Date;

  constructor(...args) {
    super(...args);
    restoreSequelizeAttributesOnClass(new.target, this);
  }
}
```

I have reported this regression in https://github.com/sequelize/sequelize/issues/14129, waiting for their response.

znarf:
Thank you for the follow up @snitin315, this looks promising.

snitin315:
@znarf I have updated the dependency with the latest patch release for the blocking issue - https://github.com/sequelize/sequelize/issues/14129

All good now. We can safely merge.

znarf (approved):

PR #7151, opened 2022-02-09
Handle Payment Processor Fee for Expenses in the Same Way for all `feePayer` Types

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5217


znarf:
This is breaking Ledger tests that are precisely intended to cover that behavior.

Introduced in: https://github.com/opencollective/opencollective/issues/5044

PR #7154, opened 2022-02-10
wip: CreateCollective update

znarf:


server/models/MemberInvitation.js
```diff
@@ -193,23 +201,35 @@ function defineModel() {
...
+        include: [{ model: models.Collective, as: 'collective' }],
+      },
+      sequelizeParams,
+    );
 
-    const invitation = await MemberInvitation.create({
-      ...memberParams,
-      CollectiveId: collective.id,
-    });
+    console.log('MemberInvitation.create', memberParams, collective.id);
```
znarf, line 224: Remove

server/graphql/v2/mutation/CreateCollectiveMutation.js
```diff
@@ -112,12 +114,58 @@ async function createCollective(_, args, req) {
...
+
       const collective = await models.Collective.create(collectiveData, { transaction });
+
       // Add authenticated user as an admin
-      await collective.addUserWithRole(user, roles.ADMIN, { CreatedByUserId: user.id }, {}, transaction);
+      if (!args.skipDefaultAdmin) {
+        await collective.addUserWithRole(user, roles.ADMIN, { CreatedByUserId: user.id }, {}, transaction);
+      }
+
+      if (args.inviteMembers && args.inviteMembers.length) {
```
Betree, line 125: I think we should add 2 restrictions there:
- Given that this can be used to know whether an email is registered on Open Collective, there should be a rate limit (to prevent email enumeration)
- `args.inviteMembers` should have a max length (30 admins?)

znarf: In the current state, we're ultimately honoring `maxCoreContributorsPerAccount` because we're using `MemberInvitation.invite`, seems this is 60 admins.

server/graphql/v2/mutation/CreateCollectiveMutation.js
```diff
@@ -32,6 +35,13 @@ async function createCollective(_, args, req) {
     host = await fetchAccountWithReference(args.host, { loaders });
   }
 
+  if (remoteUser) {
+    const rateLimit = new RateLimit(`collective_create_${remoteUser.id}`, 60, ONE_HOUR_IN_SECONDS, true);
+    if (!(await rateLimit.registerCall())) {
+      throw new RateLimitExceeded();
+    }
+  }
+
```
znarf, line 44: @Betree added rate limiting there

Betree: No rate limit when unauthenticated? Maybe we can do something like:

```es6
const rateLimitKey = remoteUser ? `collective_create_${remoteUser.id}` : `collective_create_ip_${req.ip}`
```

PR #7178, opened 2022-02-18
Ability to filter orders by frequency

znarf:
Answer https://github.com/opencollective/opencollective/discussions/5226


Betree (approved):
Great!

PR #7179, opened 2022-02-18
Fix zombie PayPal subscriptions

Betree:
For https://github.com/opencollective/opencollective/issues/5247

### DIFF
--- a/server/lib/subscriptions.ts
+++ b/server/lib/subscriptions.ts
@@ -27,13 +27,16 @@ export const updatePaymentMethodForSubscription = async (
   const newOrderData = { PaymentMethodId: newPaymentMethod.id, status: newStatus };
 
   // Subscription changes
-  let newSubscriptionData;
+  const newSubscriptionData = { isActive: true, deactivatedAt: null };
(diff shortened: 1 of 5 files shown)


server/lib/subscriptions.ts
```diff
@@ -27,13 +27,16 @@ export const updatePaymentMethodForSubscription = async (
...
   const wasManagedExternally = getIsSubscriptionManagedExternally(prevPaymentMethod);
   const isManagedExternally = getIsSubscriptionManagedExternally(newPaymentMethod);
   if (wasManagedExternally !== isManagedExternally) {
-    newSubscriptionData = { isManagedExternally, paypalSubscriptionId: null };
+    newSubscriptionData['isManagedExternally'] = isManagedExternally;
+    newSubscriptionData['paypalSubscriptionId'] = null;
   }
 
+  // Need to cancel previous subscription
+  await order.Subscription.deactivate();
```
Betree, line 39: This is the root fix, we must make sure the previous subscription is deactivated before activating the new one

znarf (approved):
Great!

PR #7183, opened 2022-02-21
createEvent mutation

znarf:
Fix https://github.com/opencollective/opencollective/issues/5190

### DIFF
--- a/server/graphql/v2/mutation/CreateEventMutation.js
+++ b/server/graphql/v2/mutation/CreateEventMutation.js
@@ -1,5 +1,7 @@
 import { GraphQLNonNull } from 'graphql';
+import slugify from 'limax';
 import { pick } from 'lodash';
+import { v4 as uuid } from 'uuid';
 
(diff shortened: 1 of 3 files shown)


server/graphql/v2/mutation/CreateEventMutation.js
```diff
@@ -0,0 +1,72 @@
...
+const createEventMutation = {
+  type: Event,
+  args: {
+    event: {
+      description: 'Information about the Event to create (name, slug, description, tags, settings)',
+      type: new GraphQLNonNull(EventCreateInput),
+    },
+    account: {
+      description: 'Reference to the parent Account creating the Event.',
+      type: AccountReferenceInput,
```
Betree, line 64: Looks like this one is required too

```suggestion
      type: new GraphQLNonNull(AccountReferenceInput),
```

Betree (approved):
Looking good :+1:

PR #7191, opened 2022-02-22
Inherit admins and accountants from Parent Collective

kewitz:
Fixes https://github.com/opencollective/opencollective/issues/4798


server/graphql/v2/interface/HasMembers.js
```diff
@@ -45,6 +48,20 @@ export const HasMembersFields = {
         };
       }
 
+      // Inherith Accountants and Admin from parent collective for Events and Projects
```
znarf, line 51:
```suggestion
      // Inherit Accountants and Admin from parent collective for Events and Projects
```

server/graphql/v2/interface/HasMembers.js
```diff
@@ -45,6 +48,20 @@ export const HasMembersFields = {
...
+      if ([CollectiveTypes.EVENT, CollectiveTypes.PROJECT].includes(collective.type)) {
+        const inheritedRoles = [MemberRoles.ACCOUNTANT, MemberRoles.ADMIN];
```
znarf, line 53: I think `MEMBER` should be inherited too.

server/graphql/v2/interface/HasMembers.js
```diff
@@ -45,6 +48,20 @@ export const HasMembersFields = {
...
 
+      // Inherit Accountants and Admin from parent collective for Events and Projects
+      if ([CollectiveTypes.EVENT, CollectiveTypes.PROJECT].includes(collective.type)) {
+        const inheritedRoles = [MemberRoles.ACCOUNTANT, MemberRoles.ADMIN, MemberRoles.MEMBER];
+        where = {
+          [Op.or]: [
+            where,
+            {
+              CollectiveId: collective.ParentCollectiveId,
+              role: { [Op.in]: args.role ? intersection(args.role, inheritedRoles) : inheritedRoles },
```
Betree, line 59: Won't this crash if I request only `MemberRoles.BACKER`? Sounds like it could generate something like:

```sql
OR role IN ()
```

Which is invalid SQL

kewitz: Sequelize will translate this to `"Member"."role" IN (NULL)` which is valid.

PR #7196, opened 2022-02-23
GQLV1: Implement CollectiveInterface children resolver

kewitz:


znarf (approved):
Looks good to me. Great if @Betree can also approve.

server/graphql/v1/CollectiveInterface.js
```diff
@@ -1343,9 +1353,24 @@ const CollectiveFields = () => {
...
             },
           ],
         });
+
+        if (args.includeChildren) {
+          return flatten(
+            memberships.map(m => {
+              if ([roles.ADMIN, roles.ACCOUNTANT].includes(m.role) && m.collective.children?.length > 0) {
+                const children = m.collective.children;
+                return [m, ...children.map(child => ({ role: m.role, collective: child }))];
```
Betree, line 1366: `MemberType` expects an instance of `models.Member`. In this current version, we're going to miss the `id` (which is problematic for caching), `description`, and all the other fields. Or am I missing something?

kewitz: @Betree you're correct, this is a workaround that I thought was justified by the requested feature but perhaps I was wrong.
To be honest, I think the existing role inheritance between parent and child collective is tech debt.
The Members data abstraction sits at a lower level than the membership inheritance business rule, the correct way to represent this is to update Members records when adding/removing memberships to the parenting collective.

If you and @znarf agree with this analysis I'll start working on that refactor today.

znarf: > the correct way to represent this is to update Members records when adding/removing memberships to the parenting collective.

I would really like to avoid this as consistency will be complicated to achieve on the long term.

server/graphql/v1/CollectiveInterface.js
```diff
@@ -587,6 +587,7 @@ export const CollectiveInterfaceType = new GraphQLInterfaceType({
       id: { type: GraphQLInt },
       createdByUser: { type: UserType },
       parentCollective: { type: CollectiveInterfaceType },
+      children: { type: new GraphQLList(CollectiveInterfaceType) },
```
Betree, line 590:
```suggestion
      children: { type: new GraphQLNonNull(new GraphQLList(CollectiveInterfaceType)) },
```

Betree (approved):
Perfect, I like the simplicity of this

PR #7199, opened 2022-02-24
Share code between transactions/orders root query and account resolvers in GraphQL v2

znarf:
We found that it was counter-intuitive for developer experience to have different parameters on the root level `transactions`/`orders` query and on the `account.transactions`/`account.orders` resolvers.

There is a clear opportunity to share code between those 2 parts, removing duplicate code and making `account.transactions` and `account.orders` more powerful.

### DIFF
--- a/server/graphql/v2/query/collection/OrdersCollectionQuery.ts
+++ b/server/graphql/v2/query/collection/OrdersCollectionQuery.ts
@@ -87,7 +87,10 @@ export const OrdersCollectionArgs = {
     type: GraphQLString,
     description: 'The term to search',
   },
-  tierSlug: { type: GraphQLString },
+  tierSlug: {
+    type: GraphQLString,
+    deprecationReason: '2022-02-25: Should be replaced by a tier reference. Not existing yet.',
+  },
   onlySubscriptions: {
     type: GraphQLBoolean,
     description: `Only returns orders that have a subscription (monthly/yearly). Don't use together with frequency.`,
--- a/server/graphql/v2/query/collection/OrdersCollectionQuery.ts
+++ b/server/graphql/v2/query/collection/OrdersCollectionQuery.ts
@@ -39,7 +39,8 @@ const getJoinCondition = (
 };
 
 export const OrdersCollectionArgs = {
-  ...CollectionArgs,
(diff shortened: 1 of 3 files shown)


znarf:
@Betree the diff is a bit impressive but I can present you the changes when you have time.

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -37,165 +38,197 @@ const getJoinCondition = (
...
+        // Needs to be root or admin of the profile to see incognito orders
+        if (req.remoteUser?.isAdminOfCollective(account) || req.remoteUser?.isRoot()) {
+          const incognitoProfile = await account.getIncognitoProfile();
+          if (incognitoProfile) {
+            accountConditions.push(getJoinCondition(incognitoProfile, 'fromCollective'));
+          }
+        } else {
+          // Is this desirable? Some current tests don't like it.
+          // throw new Error('Only admins and root can fetch incognito orders');
         }
```
Betree, line 130: The alternative of not including them as done here looks ok to me

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -37,165 +38,197 @@ const getJoinCondition = (
...
+  },
+  dateTo: {
+    type: GraphQLDateTime,
+    description: 'Only return orders that were created after this date',
+  },
+  searchTerm: {
+    type: GraphQLString,
+    description: 'The term to search',
+  },
+  tierSlug: { type: GraphQLString },
```
Betree, line 89: Should we have this one deprecated and accept a tier reference instead?

Betree (approved):
@znarf The diff looks very clear to me when hiding whitespace changes.

It's a beautiful refactor, we should take it as a reference for similar cases (e.g. expense being available at the root level + `account.expenses`).

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
znarf: Good idea. I can follow up.

PR #7211, opened 2022-02-25
Record multi-currency expenses in the ledger

Betree:
Require https://github.com/opencollective/opencollective-frontend/pull/7520
Part of https://github.com/opencollective/opencollective/issues/5146#issuecomment-1023321220

This PR includes a refactor of the way we record paid expenses.

**Why**
While implementing support for multi-currency expenses, I realized that the code supposed to check for the payment response (to get the fees and FX rate) was split in multiple places. Some of it was living under the payment provider libraries, and some of it was living under `server/lib/transactions.js` > `createFromPaidExpense`. [Here](https://github.com/opencollective/opencollective-api/blob/a486075c7ebd7a5ac9620e78f51c1765e8fac6fa/server/lib/transactions.js#L93) for example, we were parsing PayPal response, it seems out of the responsibility scope for a function that's supposed to "create transactions from paid expenses".

Multi-currency expenses required some changes in the way we handle these amounts (especially regarding FX rates) and my first implementation quickly became messy, with duplicated code in multiple places.

**How**
Though affecting multiple places, the refactor described below is quite simple in its essence: make `createTransactionsFromPaidExpense` agnostic of the payout method service. Its purpose should be solely about creating transactions from a paid expense, and the responsibility of parsing the payout provider response should be in the code of the payout providers libraries calling it.

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -1353,7 +1353,7 @@ const checkHasBalanceToPayExpense = async (
 
   // Ensure the collective has enough funds to pay the expense, with an error margin of 20% of the expense amount
   // to account for fluctuating rates. Example: to pay for a $100 expense in euros, the collective needs to have at least $120.
-  const getMinExpectedBalance = amountToPay => (isSameCurrency ? amountToPay : amountToPay * 1.2);
+  const getMinExpectedBalance = amountToPay => (isSameCurrency ? amountToPay : Math.round(amountToPay * 1.2));
 
   // Check base balance before fees
   if (balanceInExpenseCurrency < getMinExpectedBalance(expense.amount)) {
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -12,11 +12,12 @@ import POLICIES from '../../constants/policies';
 import { TransactionKind } from '../../constants/transaction-kind';
 import { hasFeature } from '../../lib/allowed-features';
 import { getFxRate } from '../../lib/currency';
+import errors from '../../lib/errors';
 import logger from '../../lib/logger';
 import { floatAmountToCents } from '../../lib/math';
 import * as libPayments from '../../lib/payments';
 import { notifyTeamAboutSpamExpense } from '../../lib/spam';
-import { createFromPaidExpense as createTransactionFromPaidExpense } from '../../lib/transactions';
+import { createTransactionsFromPaidExpense } from '../../lib/transactions';
 import {
   handleTwoFactorAuthenticationPayoutLimit,
   resetRollingPayoutLimitOnFailure,
@@ -1112,30 +1113,6 @@ async function markExpenseAsPaid(expense, remoteUser, isManualPayout = false): P
   return expense;
 }
 
-async function createTransactions(
-  host,
-  expense,
-  fees: {
-    paymentProcessorFeeInHostCurrency?: number;
-    hostFeeInHostCurrency?: number;
-    platformFeeInHostCurrency?: number;
-  } = {},
-  data = {},
-) {
-  debug('marking expense as paid and creating transactions in the ledger', expense.id);
-  return await createTransactionFromPaidExpense(
-    host,
-    null,
-    expense,
-    null,
-    expense.UserId,
-    fees['paymentProcessorFeeInHostCurrency'],
-    fees['hostFeeInHostCurrency'],
-    fees['platformFeeInHostCurrency'],
-    data,
-  );
-}
-
 async function payExpenseWithPayPalAdaptive(remoteUser, expense, host, paymentMethod, toPaypalEmail, fees = {}) {
   debug('payExpenseWithPayPalAdaptive', expense.id);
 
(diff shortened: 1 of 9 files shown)


server/graphql/common/expenses.ts
```diff
@@ -1180,7 +1191,16 @@ async function payExpenseWithPayPalAdaptive(remoteUser, expense, host, paymentMe
...
-  await createTransactions(host, expense, fees, data);
+  if (host.settings?.transferwise?.ignorePaymentProcessorFees) {
+    // TODO: We should not just ignore fees, they should be recorded as a transaction from the host to the collective
+    // See https://github.com/opencollective/opencollective/issues/5113
+    fees.paymentProcessorFeeInHostCurrency = 0;
+  } else if (data?.paymentOption?.fee?.total) {
+    fees.paymentProcessorFeeInHostCurrency.fees = Math.round(data.paymentOption.fee.total * 100);
+  }
+
+  const fxRate = 1; // TODO
```
Betree, line 1202: @kewitz Let's talk about this one, I'm not sure where's the best place to retrieve this info (the quote?)

kewitz: Correct me if I'm wrong, but...

We always store the DEBIT transaction in the same currency as the collective, which can differ from the host in some edge cases.
We want to store the CREDIT transaction in the same currency of the expense, which now can be any* currency the user picks and can also differ from the collective.

For DEBIT:
- We have the actual value that was debited from the Host's Wise account.
- Use getFxRate() between host and collective.

For CREDIT:
- Use have the actual FX rate we have in expense.data.quote because it should be the actual rate used between Host and User.

server/lib/transactions.ts
```diff
@@ -57,140 +56,142 @@ export function getTransactions(collectiveids, startDate = new Date('2015-01-01'
...
-    } else if (transactionData?.paymentOption?.fee?.total) {
-      paymentProcessorFeeInHostCurrency = Math.round(transactionData.paymentOption.fee.total * 100);
-    }
-    // Notice this is the FX rate between Host and Collective, the user is not involved here and that's why TransferWise quote rate is irrelevant here.
-    hostCurrencyFxRate = await getFxRate(expense.currency, host.currency);
-    paymentProcessorFeeInCollectiveCurrency = Math.round((1 / hostCurrencyFxRate) * paymentProcessorFeeInHostCurrency);
-    hostFeeInCollectiveCurrency = Math.round((1 / hostCurrencyFxRate) * hostFeeInHostCurrency);
-    platformFeeInCollectiveCurrency = Math.round((1 / hostCurrencyFxRate) * platformFeeInHostCurrency);
+  // Get the right FX rate for the expense
+  let expenseToHostFxRate;
```
kewitz, line 154: 🍨 I rather use a ternary and declare `expenseToHostFxRate` as a constant.

```suggestion
  const expenseToHostFxRate = expenseToHostFxRateConfig === 'auto'
    ? await getFxRate(expense.currency, host.currency, expense.incurredAt || expense.createdAt)
    : expenseToHostFxRateConfig;
```

server/paymentProviders/paypal/payouts.ts
```diff
@@ -93,7 +94,18 @@ export const checkBatchItemStatus = async (
...
     case 'SUCCESS':
       if (expense.status !== status.PAID) {
-        await createTransactionFromPaidExpense(host, null, expense, null, expense.UserId, 0, 0, 0, item);
+        const fees = {};
+        const fxRate = 1 / (toNumber(item.currency_conversion?.exchange_rate) || 1);
+        if (item.payout_item_fee) {
+          const paymentProcessorFeeInExpenseCurrency = floatAmountToCents(toNumber(item.payout_item_fee.value));
+          fees['paymentProcessorFeeInHostCurrency'] = Math.round(paymentProcessorFeeInExpenseCurrency * fxRate);
+          if (item.payout_item_fee.currency !== expense.currency) {
+            // payout_item_fee is always supposed to be in currency_conversion.to_amount.currency. This is a sanitiy check just in case
```
SudharakaP, line 103:
```suggestion
            // payout_item_fee is always supposed to be in currency_conversion.to_amount.currency. This is a sanity check just in case
```

server/graphql/common/expenses.ts
```diff
@@ -1152,16 +1129,50 @@ async function payExpenseWithPayPalAdaptive(remoteUser, expense, host, paymentMe
...
+
+      case 'ERROR':
+        // Backward compatible error message parsing
+        // eslint-disable-next-line no-case-declarations
+        const errorMessage =
+          executePaymentResponse.payErrorList?.payError?.[0].error?.message ||
+          executePaymentResponse.payErrorList?.[0].error?.message;
+        throw new errors.ServerError(
+          `Error while paying the expense with PayPal: "${errorMessage}". Please contact support@opencollective.com or pay it manually through PayPal.`,
+        );
```
SudharakaP, line 1157: I think we can wrap the case in a block rather than adding this es-lint exception. 🤔 

```suggestion
      case 'ERROR': {
        // Backward compatible error message parsing
        const errorMessage =
          executePaymentResponse.payErrorList?.payError?.[0].error?.message ||
          executePaymentResponse.payErrorList?.[0].error?.message;
        throw new errors.ServerError(
          `Error while paying the expense with PayPal: "${errorMessage}". Please contact support@opencollective.com or pay it manually through PayPal.`,
        );
      }
```

server/lib/transactions.ts
```diff
@@ -57,140 +56,140 @@ export function getTransactions(collectiveids, startDate = new Date('2015-01-01'
...
+ *
+ * TODO: This function should accept an `amount` to automatically represent what was really paid from the payment provider,
+ * in case it differs from expense.amount (e.g. when fees are put on the payee)
+ */
+export async function createTransactionsFromPaidExpense(
+  host,
+  expense,
+  fees: FEES_IN_HOST_CURRENCY = DEFAULT_FEES,
+  /** Set this to a different value if the expense was paid in a currency that differs form the host's */
+  expenseToHostFxRateConfig: number | 'auto',
```
SudharakaP, line 144: Do we need this `auto` ? We could just do,

```suggestion
  expenseToHostFxRateConfig: number = null,
```

And then set, 

```js
// Get the right FX rate for the expense
  const expenseToHostFxRate =
    expenseToHostFxRateConfig || await getFxRate(expense.currency, host.currency, expense.incurredAt || expense.createdAt);
```

This way we can get rid of the duplicating `auto` arg in the function references.

server/graphql/common/expenses.ts
Betree: As mentioned above, this code is just a cut/paste that I don't want to change now to reduce the risk of side effects and make this PR simpler to review. We can clean the code later.

server/lib/transactions.ts
Betree: I want the automatic FX rate to be an explicit feature that developers enable consciously. If we make it optional, anyone implementing a new payout provider could end up calling this function with the simplest prototype, while we usually want to retrieve the FX rate from the payout provider. Only expenses paid manually or using our core system are supposed to have an automatic fx rate.

PR #7221, opened 2022-02-28
Fix Virtual Card details resolver permission check for Projects and Events

kewitz:
Fixes https://sentry.io/organizations/open-collective/issues/3051366523/?project=1736806


znarf (commented):
If you fetch the collective, then I suggest to simplify and use the `isAdminOfCollective` utility.

Or even better? `isAdminOfCollectiveOrHost`

PR #7229, opened 2022-03-02
Fix PaymentMethod.name resolver

kewitz:
Two fixes in one:
- Collective balance payment method name should be public.
- The resolver should consider children collectives now that we support Funds and Events.


znarf (changes requested):
Could you take care of all fields and not just `name`?

PR #7235, opened 2022-03-03
Adapt PaymentMethodId for fixed PayPal orphan subscriptions

Betree:
The transactions created during the first migration had their payment method invalidly set to the credit card instead of PayPal


znarf (approved):
Looks good

PR #7237, opened 2022-03-03
Refactor Wise: throw pre-check errors synchronously

kewitz:
Fixes https://github.com/opencollective/opencollective/issues/5253

I'm moving some Wise `createTransaction` pre-checks outside the try/catch that reports errors asynchronously.

I think there's still value in moving the Expense to error and generating activity with more information for the user, but this pre-check is mainly related to the Host and it makes more sense to deliver the error synchronously as a toast to the host admin paying/scheduling the expense.


Betree (approved):
Looks good to me!

PR #7239, opened 2022-03-04
Do no require address when contributing from the same host

Betree:
Resolve https://opencollective.slack.com/archives/G46PNRCGP/p1646138401161639

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -184,9 +184,10 @@ const checkAndUpdateProfileInfo = async (order, fromAccount, isGuest, currency) 
   const { totalAmount, fromAccountInfo, guestInfo } = order;
   const accountUpdatePayload = {};
   const location = fromAccountInfo?.location || guestInfo?.location || fromAccount.location;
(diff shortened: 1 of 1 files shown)


server/graphql/v1/mutations/orders.js
```diff
@@ -184,9 +184,10 @@ const checkAndUpdateProfileInfo = async (order, fromAccount, isGuest, currency)
   const { totalAmount, fromAccountInfo, guestInfo } = order;
   const accountUpdatePayload = {};
   const location = fromAccountInfo?.location || guestInfo?.location || fromAccount.location;
+  const isContributingFromSameHost = fromAccount.HostCollectiveId === order.collective.HostCollectiveId;
 
   // Only enforce profile checks for guests and USD contributions at the moment
-  if (isGuest && currency === 'USD') {
+  if (isGuest && currency === 'USD' && !isContributingFromSameHost) {
```
Betree, line 190: This is cosmetic as `isGuest` will be false when contributing with an existing profile, but I wanted to make sure we keep track of this condition in the API

PR #7244, opened 2022-03-07
Fix Webhook Issues

kewitz:
These fixes address the recent warning email sent by Stripe related to the high error rate we're having on our API.


kewitz:
Merging this as a hot-fix because Stripe imposed a deadline to fix this.

PR #7251, opened 2022-03-09
Add scripts/wise/fix-missing-transactions script

kewitz:


server/graphql/common/expenses.ts
```diff
@@ -1284,8 +1284,11 @@ export const getExpenseFees = async (
     if (!connectedAccount) {
       throw new Error('Host is not connected to Transferwise');
     }
-    const quote = await paymentProviders.transferwise.getTemporaryQuote(connectedAccount, payoutMethod, expense);
-    const paymentOption = quote.paymentOptions.find(p => p.payIn === 'BALANCE' && p.payOut === quote.payOut);
+    const quote =
+      expense.data.quote ||
+      (await paymentProviders.transferwise.getTemporaryQuote(connectedAccount, payoutMethod, expense));
+    const paymentOption =
+      expense.data.paymentOption || quote.paymentOptions.find(p => p.payIn === 'BALANCE' && p.payOut === quote.payOut);
```
Betree, line 1291: Nice, that way the script will use existing fees from `data` if available.

Betree (approved):

server/graphql/common/expenses.ts
kewitz: Yes, but in retrospect, since we call this before paying for the expense, I'll be more explicit and add a flag to `getExpenseFees` options for when we want to use existing data.

PR #7252, opened 2022-03-09
Block payments for Multi-Currency Expenses if not enabled

Betree:
Because of https://github.com/opencollective/opencollective/issues/5298


znarf (commented):
Looks good.

PR #7253, opened 2022-03-09
Documentation for adding new models

Betree:


znarf:
I would put this in the `docs` folder directly. And maybe keep a dummy Readme to "redirect" for people looking in the `models` folder.

PR #7254, opened 2022-03-09
Improve API for expenses fxRates and better retrieve values

Betree:
Part of https://github.com/opencollective/opencollective/issues/5300

This change introduces a new `exchangeRate` field on Amount to describe the currency conversion that happened (if any). This design was suggested in https://github.com/opencollective/opencollective/issues/5198#issuecomment-1062725900.

**Example query**

```gql
{
  expense(expense: {legacyId: 2434}) {
    amountInExpenseCurrency: amountV2(currencySource: EXPENSE) {
      valueInCents
      currency
      fxRate {
        percentage
        source
        fromCurrency
        toCurrency
        date
        isApproximate
      }
    }
    amountInAccountCurrency: amountV2(currencySource: ACCOUNT) {
      valueInCents
      currency
      fxRate {
        percentage
        source
        fromCurrency
        toCurrency
        date
        isApproximate
      }
    }
    amountInHostCurrency: amountV2(currencySource: HOST) {
      valueInCents
      currency
      fxRate {
        percentage
        source
        fromCurrency
        toCurrency
        date
        isApproximate
      }
    }
  }
}
```

**Example response**

```json
{
  "data": {
    "expense": {
      "amountInExpenseCurrency": {
        "valueInCents": 3300,
        "currency": "EUR",
        "fxRate": null
      },
      "amountInAccountCurrency": {
        "valueInCents": 3630,
        "currency": "USD",
        "fxRate": {
          "percentage": 1.1,
          "source": "OPENCOLLECTIVE",
          "fromCurrency": "EUR",
          "toCurrency": "USD",
          "date": "2022-02-09T09:46:18.772Z",
          "isApproximate": true
        }
      },
      "amountInHostCurrency": {
        "valueInCents": 3630,
        "currency": "USD",
        "fxRate": {
          "percentage": 1.1,
          "source": "OPENCOLLECTIVE",
          "fromCurrency": "EUR",
          "toCurrency": "USD",
          "date": "2022-02-09T09:46:18.772Z",
          "isApproximate": true
        }
      }
    }
  }
}
```


server/graphql/v2/object/CurrencyFxRate.ts
```diff
@@ -0,0 +1,36 @@
...
+import { GraphQLDateTime } from 'graphql-scalars';
+
+import { Currency } from '../enum';
+import { CurrencyFxRateSourceType } from '../enum/CurrencyFxRateSourceType';
+
+const CurrencyFxRate = new GraphQLObjectType({
+  name: 'CurrencyFxRate',
+  description: 'Fields for a currency fx rate',
+  fields: () => ({
+    percentage: {
```
znarf, line 11: Not sure about this name. Alternatives: `value`, `rate`?

server/graphql/v2/object/Amount.js
```diff
@@ -33,5 +35,10 @@ export const Amount = new GraphQLObjectType({
         }
       },
     },
+    fxRate: {
```
znarf, line 38: I find `fxRate` a bit jargon for a public API. It might be ok but we could replace by something like `exchangeRate`.

znarf: Would it make sense for this field to be a list? In case we want to feature multiple rates or multiple currencies?

`exchangeRates: GraphQLList(CurrencyFxRate),`

server/graphql/v2/object/CurrencyFxRate.ts
```diff
@@ -0,0 +1,36 @@
...
+      type: new GraphQLNonNull(GraphQLFloat),
+      description: 'Amount of this item',
+    },
+    source: {
+      type: new GraphQLNonNull(CurrencyFxRateSourceType),
+      description: 'Where does the FX rate comes from',
+    },
+    fromCurrency: {
+      type: new GraphQLNonNull(Currency),
+    },
```
znarf, line 21: `from` is what we use internally, I've seen some API like Fixer using `base`.

znarf: Is it necessary though? I guess this should always be the same as `amount.currency` ?

Betree: It's not necessary here as we indeed have the same info in `Amount`. But the idea was to be able to re-use this object and have something able to completely describe a currency conversion regardless of the context. We'll also need this `fromCurrency` if we want rates to be cached at some point (id would be something like `${fromCurrency}-${toCurrency}-${source}-${date}`.

Regarding the name itself, I personally find the `fromCurrency`/`toCurrency` pair predictable and clearer than `base`. To compare:
- Wise uses `sourceCurrency`/`targetCurrency`
- PayPal uses an object like:

```ts
  currency_conversion: {
    to_amount: { value: number; currency: string };
    from_amount: { value: number; currency: string };
    exchange_rate: string;
  };
```

server/graphql/v2/object/Amount.js
Betree: > I find `fxRate` a bit jargon for a public API. It might be ok but we could replace by something like `exchangeRate`.

I agree that "fxRate" sounds a bit jargon. Today we use it in `transaction.hostCurrencyFxRate`, we should ideally make the fields consistent. If that's ok for you I'll update this one to `exchangeRate` and will open an issue about renaming `hostCurrencyFxRate` to `hostCurrencyExchangeRate`



> Would it make sense for this field to be a list? In case we want to feature multiple rates or multiple currencies?
> 
> `exchangeRates: GraphQLList(CurrencyFxRate),`

This field is supposed to showcase the `fxRate` that was used to generate the amount. I don't think exposing other rates here would make sense.

server/graphql/v2/object/CurrencyFxRate.ts
Betree: `value` sounds good, it's consistent with what we do elsewhere

znarf (approved):
Good for me!

PR #7264, opened 2022-03-10
Fix VirtualCard.requested email template

kewitz:
image


znarf (approved):
Looks good. Already better.

PR #7265, opened 2022-03-10
Update Missing Condition for Active Bank Accounts

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/3193

When working on https://github.com/opencollective/opencollective-frontend/pull/7569 I noticed that when removing a payout method we set `isSaved` to `ture`;

https://github.com/opencollective/opencollective-api/blob/2ec845d0d8ec719b6d7854584fced0bfc8795264/server/graphql/v2/mutation/PayoutMethodMutations.ts#L89

Then when we obtain a existing bank account we do not check for this condition resulting in already removed (unsaved) bank accounts.

### DIFF
--- a/server/graphql/v2/mutation/PayoutMethodMutations.ts
+++ b/server/graphql/v2/mutation/PayoutMethodMutations.ts
@@ -46,6 +46,7 @@ const payoutMethodMutations = {
           where: {
             data: { isManualBankTransfer: true },
             CollectiveId: collective.id,
+            isSaved: true,
           },
         });
         if (existingBankAccount) {


SudharakaP:
@Betree: Also as a related question, do you know why do we set this `isSaved` flag instead of soft deleting it by setting `deletedAt`? 🤔

Betree:
@SudharakaP The reason for having this `isSaved` flag (compared to `deletedAt`) is that we usually don't want to remove a payout method if it has been used for an expense. We need to keep track of how the money went out. If we set `deletedAt`, all `models.PayoutMethod.find(...)` will return null, which can break the code in some places (e.g. to display the payout type on the expense). In this context, the goal of `isSaved` is to have a way to filter out the payout method when **listing** them without affecting other places.

server/graphql/v2/mutation/PayoutMethodMutations.ts
```diff
@@ -46,6 +46,7 @@ const payoutMethodMutations = {
           where: {
             data: { isManualBankTransfer: true },
             CollectiveId: collective.id,
+            isSaved: true,
```
Betree, line 49: @kewitz I think you're the one who implemented this, what do you think about this?

TLDR ; should we fetch and update the removed bank transfer payout method (that's what we were doing) or ignore everything removed and create a new one (@SudharakaP's proposal)?

I feel Suds made a good call here but just want to make sure it won't break other things.

kewitz: Yes, I think this is a valid update.
Just keep in mind that every payout method created with `{ isManualBankTransfer: true }` is by default saved.

SudharakaP: @Betree @kewitz : Thanks for the comments. Are we good to merge this? 🤔

Betree: I don't have the context of this anymore but I had no blocker at the time. Since @kewitz is good with this then I am too

PR #7266, opened 2022-03-11
Fix memberOf query when `includeIncognito` is true

Betree:
Fix https://github.com/opencollective/opencollective/issues/5311
Introduced in https://github.com/opencollective/opencollective-api/pull/7023

Some queries were broken by the update in https://github.com/opencollective/opencollective-api/pull/7023 that invalidly filtered results to get only the incognito profiles.

### DIFF
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -1331,15 +1331,15 @@ const CollectiveFields = () => {
         if (roles && roles.length > 0) {
           where.role = { [Op.in]: roles };
         }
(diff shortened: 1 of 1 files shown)


server/graphql/v1/CollectiveInterface.js
```diff
@@ -1331,15 +1331,15 @@ const CollectiveFields = () => {
         if (roles && roles.length > 0) {
           where.role = { [Op.in]: roles };
         }
-        const collectiveConditions = { deletedAt: null, isIncognito: false };
```
Betree, line 1334: `deletedAt` is already set by default with sequelize

PR #7271, opened 2022-03-11
Follow up on host plan deprecation

znarf:


server/models/Transaction.js
```diff
@@ -812,6 +812,15 @@ function defineModel() {
...
     }
 
+    // Skip if missing or misconfigured
+    const hostFeeShareCollective = await models.Collective.findByPk(HOST_FEE_SHARE_TRANSACTION_PROPERTIES.CollectiveId);
+    const hostFeeShareHostCollective = await models.Collective.findByPk(
+      HOST_FEE_SHARE_TRANSACTION_PROPERTIES.HostCollectiveId,
+    );
+    if (!hostFeeShareCollective || !hostFeeShareHostCollective) {
+      return;
+    }
```
kewitz, line 822: Oh, so it was an edge-case related to CollectiveId.
I think this makes perfect sense, we should later create a new DB seed export to make sure our environments are consistent.

kewitz (approved):

PR #7284, opened 2022-03-15
Add support for tag searching for the `/search` query in GraphQL V1

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/2970

Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/7579

### DIFF
--- a/server/lib/search.js
+++ b/server/lib/search.js
@@ -125,7 +125,8 @@ export const searchCollectivesInDB = async (
         AND (${TS_VECTOR} @@ plainto_tsquery('english', :vectorizedTerm)
         OR ${TS_VECTOR} @@ plainto_tsquery('simple', :vectorizedTerm)
         OR name ILIKE '%' || :term || '%'
-        OR slug ILIKE '%' || :term || '%')`;
+        OR slug ILIKE '%' || :term || '%'
+        OR :term = any(tags))`;
     }
   } else {
     term = '';


znarf:
What do you mean by "/search endpoint" ? I don't know anything like this.

SudharakaP:
> What do you mean by "/search endpoint" ? I don't know anything like this.

Maybe wrong terminology, but what I meant is the logic on the API that we have for searching collectives; https://github.com/opencollective/opencollective-api/blob/533521783a17371eb37d07bcf7a596ba866c617a/server/graphql/v1/queries.js#L901-L904

znarf:
It's then the `search` query in GraphQL API v1.

znarf (approved):

Betree (commented):
@SudharakaP I don't understand this change, tags are already part of the TS vector: https://github.com/opencollective/opencollective-api/blob/5eb8644ce6cd5d0150be2e9d95a4d95d75030640/server/lib/search.js#L79. Was it not working?

The problem with the approach taken here is that the slug is not included in the `ts_rank`, so results may not be sorted accurately.

SudharakaP:
> @SudharakaP I don't understand this change, tags are already part of the TS vector:
> 
> https://github.com/opencollective/opencollective-api/blob/5eb8644ce6cd5d0150be2e9d95a4d95d75030640/server/lib/search.js#L79
> . Was it not working?

@Betree : Thanks for the review. 

It was working, but I also wanted to support tags such as `jhipster-tag` where there is punctuation; which is an obvious use case I thought. With the previous version it seems that if the user types `jhipster-tag` no results appear as `array_to-tsvector` seems to do some normalization. If they search for `jhipstertag` results appear but that isn't intuitive for the user. 

> 
> The problem with the approach taken here is that the slug is not included in the `ts_rank`, so results may not be sorted accurately.

I don't understand this part, we didn't do any changes to the `slug` in this change; slug is already included here so isn't it not part of `ts_rank`? 

https://github.com/opencollective/opencollective-api/blob/5eb8644ce6cd5d0150be2e9d95a4d95d75030640/server/lib/search.js#L77

Betree:
> I don't understand this part, we didn't do any changes to the slug in this change; slug is already included here so isn't it not part of ts_rank?

The `TS_RANK` is done like this:

```sql
ts_rank(${TS_VECTOR}, plainto_tsquery('english', :vectorizedTerm))
```

If it doesn't match in the `WHERE`, it won't match in the `TS_RANK` either because it's the exact same query. Concretly, we're doing the same thing as the query below. If you try and run it, you'll see that it indeed returns a line, but with a TS_RANK = 0:

```sql
SELECT ts_rank(array_to_tsvector(ARRAY['jhipster-tag']), plainto_tsquery('simple', 'jhipster-tag'))
WHERE 'jhipster-tag' = ANY(ARRAY['jhipster-tag'])
```

> It was working, but I also wanted to support tags such as jhipster-tag where there is punctuation; which is an obvious use case I thought. With the previous version it seems that if the user types jhipster-tag no results appear as array_to-tsvector seems to do some normalization. If they search for jhipstertag results appear but that isn't intuitive for the user.

It would have been great to document the issue beforehand or explain that in the PR description.

The following line indeed returns false:

```sql
array_to_tsvector(ARRAY['jhipster-tag']) @@ plainto_tsquery('simple', 'jhipster-tag')
```

I can recommend a different way of building the TS_VECTOR which will work with this example:

```sql
to_tsvector(array_to_string(ARRAY['jhipster-tag'], ' ')) @@ plainto_tsquery('simple', 'jhipster-tag'),
```

SudharakaP:
@Betree : Thanks a lot for the explanation. Makes senses, I've opened a new PR; https://github.com/opencollective/opencollective-api/pull/7310 with the proposed fix. Seems to work perfectly.

PR #7295, opened 2022-03-18
Fix Transaction.amountInHostCurrency for feesPayer == PAYEE

kewitz:
- [x] Fix how we calculate amountInHostCollective on DEBITS
- [x] Write a migration to fix existing records

### DIFF
--- a/server/lib/transactions.ts
+++ b/server/lib/transactions.ts
@@ -189,6 +189,7 @@ export async function createTransactionsFromPaidExpense(
   // If the payee is assuming the fees, we adapt the amounts
   if (expense.feesPayer === 'PAYEE') {
     transaction.amount += processedAmounts.paymentProcessorFee.inCollectiveCurrency;
+    transaction.amountInHostCurrency += processedAmounts.paymentProcessorFee.inHostCurrency;
     transaction.netAmountInCollectiveCurrency += processedAmounts.paymentProcessorFee.inCollectiveCurrency;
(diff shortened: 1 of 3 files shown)


znarf (commented):
We have a transaction validator, could be interesting to use it as it would have catch the issue originally.

```
await Transaction.validate(transaction);
```

kewitz:
> We have a transaction validator, could be interesting to use it as it would have catch the issue originally. To be used in tests:
> 
> ```
> await Transaction.validate(transaction);
> ```

@znarf 
Interestingly enough, we commented the check for `amountInHostCurrency in oppositeTransaction should match`.
This fix is literally rewriting this rule to `amountInHostCurrency should match amount * fxRate`, are we sure about this?

znarf:
The commented test is between the transaction value and the opposite transaction value. 

The normal check is enforced:

```
    Transaction.assertAmountsLooselyEqual(
      Math.round(transaction.amountInHostCurrency / hostCurrencyFxRate),
      transaction.amount,
      'amountInHostCurrency should match amount',
    );
```

znarf:
What about validating Transactions in the test suite like I suggested?

kewitz:
@znarf I don't like that idea, this function is not unit-tested and has some caveats.
Ideally, `Transaction.validate` method would have its own unit tests and these tests would break only when refactoring the method itself.
Also, having redundant check on tests is a good and not a bad thing.

znarf (approved):

PR #7306, opened 2022-03-21
Add support to filter by country in `search` Query

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5332

Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/7606

### DIFF
--- a/server/graphql/v1/queries.js
+++ b/server/graphql/v1/queries.js
@@ -16,6 +16,7 @@ import models, { Op, sequelize } from '../../models';
 import { allowContextPermission, PERMISSION_TYPE } from '../common/context-permissions';
 import { canDownloadInvoice } from '../common/transactions';
 import { Forbidden, NotFound, Unauthorized, ValidationFailed } from '../errors';
+import { CountryISO } from '../v2/enum';
 
 import { ApplicationType } from './Application';
 import {
@@ -931,6 +932,10 @@ const queries = {
         description: 'Whether to skip recent accounts (48h)',
         defaultValue: false,
       },
+      countries: {
+        type: new GraphQLList(CountryISO),
+        description: 'Limit the search to collectives belonging to these countries',
+      },
       limit: {
         type: GraphQLInt,
         description: 'Limit the amount of results. Defaults to 20',
@@ -942,7 +947,7 @@ const queries = {
       },
     },
     async resolve(_, args, req) {
-      const { limit, offset, term, types, isHost, hostCollectiveIds, skipRecentAccounts } = args;
+      const { limit, offset, term, types, isHost, hostCollectiveIds, skipRecentAccounts, countries } = args;
       const cleanTerm = term ? term.trim() : '';
       logger.info(`Search Query: ${cleanTerm}`);
       const listToStr = list => (list ? list.join('_') : '');
@@ -969,6 +974,7 @@ const queries = {
           hostCollectiveIds,
           isHost,
           skipRecentAccounts,
+          countries,
         });
         return generateResults(collectives, total);
       }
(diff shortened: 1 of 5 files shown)


znarf:
What do you mean by `/search` ?

server/lib/search.js
```diff
@@ -86,11 +86,15 @@ export const searchCollectivesInDB = async (
   term,
   offset = 0,
   limit = 100,
-  { types, hostCollectiveIds, isHost, onlyActive, skipRecentAccounts, hasCustomContributionsEnabled } = {},
+  { types, hostCollectiveIds, isHost, onlyActive, skipRecentAccounts, hasCustomContributionsEnabled, countries } = {},
 ) => {
   // Build dynamic conditions based on arguments
   let dynamicConditions = '';
   let isUsingTsVector = false;
+  let countryCodes = 'ALL';
```
znarf, line 94: Our generic practice is to use a `null` value instead of a `'ALL'` string like here.

server/graphql/v1/queries.js
```diff
@@ -931,6 +931,10 @@ const queries = {
         description: 'Whether to skip recent accounts (48h)',
         defaultValue: false,
       },
+      countries: {
+        type: new GraphQLList(GraphQLString),
```
znarf, line 935: Why not use an Enum instead of a generic GraphQLString?

znarf (commented):
Maybe you should migrate to the API v2 query first, so you don't have to modify the GraphQL v1 query that we're about to deprecate.

SudharakaP:
> What do you mean by `/search` ?

I meant the [search query](https://github.com/opencollective/opencollective-api/blob/0e61e31c519117d63acb155a9ac0d9168533557d/server/graphql/v1/queries.js#L904); I've renamed the title. 

 >Maybe you should migrate to the API v2 query first, so you don't have to modify the GraphQL v1 query that we're about to deprecate.

Can we do this as a separate task after this is merged? Because this seems a bit complicated than I expected; we have to check carefully if everything works as expected after migrating the `search` query to GraphQL v2. There are some fields (such as [`company`](https://github.com/opencollective/opencollective-frontend/blob/7e8ed06da80c82f0c9ae4d2563e1dc305f92fb5c/pages/search.js#L393)) that might need to be queried differently. 🤔

server/lib/search.js
SudharakaP: I've changed this to use `null`. Thanks. 👍🏽

server/graphql/v1/queries.js
SudharakaP: Thanks, that makes sense, we already have the `CountryISO` enum, so I've used that. 👍🏽

znarf (approved):
Good for me.

server/graphql/v1/queries.js
Betree: According to https://github.com/opencollective/opencollective/issues/4378, shouldn't that be `country`?

server/graphql/v2/query/collection/AccountsQuery.ts
```diff
@@ -65,6 +65,7 @@ const AccountsQuery = {
         onlyActive: args.isActive ? true : null,
         skipRecentAccounts: args.skipRecentAccounts,
         hasCustomContributionsEnabled: args.hasCustomContributionsEnabled,
+        countries: args.countries,
```
Betree, line 68: I don't see `args.countries` being defined in `args`

Betree: Also, according to https://github.com/opencollective/opencollective/issues/4378, shouldn't that be `country`?

SudharakaP: > I don't see `args.countries` being defined in `args`

Thanks Ben. This is due to migrating to GQL v2 and the front-end query still using graphql v1. Anyways, I've opened a new PR so that the country filter uses GQL v2; https://github.com/opencollective/opencollective-api/pull/7363

> Also, according to [opencollective/opencollective#4378](https://github.com/opencollective/opencollective/issues/4378), shouldn't that be `country`?

Good point, I've missed this. This was corrected in the aforementioned PR as well.

PR #7307, opened 2022-03-22
refactor(virtualcard): rename virtual card `expireDate` to `expiryDate`

francois-travais:
Resolve https://github.com/opencollective/opencollective/issues/4960 with https://github.com/opencollective/opencollective-frontend/pull/7599

I couldn't really test the migration. Best would be to run it against a copy of the production data.

### DIFF
--- a/migrations/20220321220618-rename-virtual-card-epire-date-to-expiry-date.js
+++ b/migrations/20220321220618-rename-virtual-card-epire-date-to-expiry-date.js
@@ -1,0 +1,57 @@
+'use strict';
+
+import { omit } from 'lodash';
+
+import { crypto } from '../server/lib/encryption';
+
+async function renameFieldInVirtualCardsPrivateData(queryInterface, from, to) {
+  const [virtualCards] = await queryInterface.sequelize.query(
+    `SELECT "id", "privateData" FROM "VirtualCards" WHERE "privateData" IS NOT NULL;`,
+  );
+  console.info(`Renaming '${from}' to '${to}' for ${virtualCards.length} VirtualCards...`);
+  try {
+    await queryInterface.sequelize.transaction(async transaction => {
+      for (const virtualCard of virtualCards) {
+        const originalPrivateData = JSON.parse(crypto.decrypt(virtualCard.privateData));
+        if (!(from in originalPrivateData)) {
+          console.debug(`'${from}' not in 'privateData' for VirtualCard with id=${virtualCard.id}`);
+          continue;
+        }
+        const privateData = {
+          ...omit(originalPrivateData, [from]),
+        };
+        privateData[to] = originalPrivateData[from];
+        const encryptedPrivateData = JSON.stringify(crypto.encrypt(privateData));
+        console.debug(`Renaming '${from}' to '${to}' for VirtualCard with id=${virtualCard.id}`);
+        await queryInterface.sequelize.query(
+          `
+            UPDATE "VirtualCards"
+            SET "privateData" = :privateData
+            WHERE "id" = :id;
+          `,
+          {
+            transaction,
+            replacements: {
+              encryptedPrivateData,
+              id: virtualCard.id,
+            },
+          },
+        );
+      }
+    });
+    console.info('Done.');
+  } catch (e) {
+    console.error('Oops, something went wrong and I rolled back the transaction.');
+    console.error(e);
+  }
+}
+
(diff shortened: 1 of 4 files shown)


migrations/20220321220618-rename-virtual-card-epire-date-to-expiry-date.js
```diff
@@ -0,0 +1,93 @@
...
+  async up(queryInterface) {
+    const [virtualCards] = await queryInterface.sequelize.query(
+      `SELECT "id", "privateData" FROM "VirtualCards" WHERE "privateData" IS NOT NULL;`,
+    );
+    console.info(`Renaming 'expireDate' to 'expiryDate' for ${virtualCards.length} VirtualCards...`);
+    try {
+      await queryInterface.sequelize.transaction(async transaction => {
+        for (const virtualCard of virtualCards) {
+          const originalPrivateData = JSON.parse(crypto.decrypt(virtualCard.privateData));
+          if (!('expireDate' in originalPrivateData)) {
```
Betree, line 17: Would have been interesting to pre-filter that with SQL

```sql
SELECT "id", "privateData"
FROM "VirtualCards"
WHERE "privateData" IS NOT NULL
AND "privateData" ? 'expireDate';
```

migrations/20220321220618-rename-virtual-card-epire-date-to-expiry-date.js
```diff
@@ -0,0 +1,93 @@
...
+      });
+      console.info('Done.');
+    } catch (e) {
+      console.error('Oops, something went wrong and I rolled back the transaction.');
+      console.error(e);
+    }
+  },
+
+  async down(queryInterface) {
+    const [virtualCards] = await queryInterface.sequelize.query(
```
Betree, line 52: :ice_cream: Since 90% of the code between up/down is shared, it could be nice to have a dedicated helper like `renameFieldInVirtualCardsPrivateData(oldName, newName)` in this file to simplify the code and prevent copy-paste oversights.

The thing could look like:

```es6
up: () => renameFieldInVirtualCardsPrivateData('expireDate', 'expiryDate');
down: () => renameFieldInVirtualCardsPrivateData('expiryDate', 'expireDate');
```

Betree (commented):
Great things about this PR:
- It runs in a SQL transaction
- We're using raw SQL queries rather than `models`, that's what we want in migrations
- The code looks nice & clean

Thanks for the contrib @francois-travais!

migrations/20220321220618-rename-virtual-card-epire-date-to-expiry-date.js
znarf: I thought about it too but the data is encrypted.

migrations/20220321220618-rename-virtual-card-epire-date-to-expiry-date.js
```diff
@@ -0,0 +1,57 @@
...
+        await queryInterface.sequelize.query(
+          `
+            UPDATE "VirtualCards"
+            SET "privateData" = :privateData
+            WHERE "id" = :id;
+          `,
+          {
+            transaction,
+            replacements: {
+              encryptedPrivateData,
```
Betree, line 35:
```suggestion
            SET "privateData" = :encryptedPrivateData
            WHERE "id" = :id;
          `,
          {
            transaction,
            replacements: {
              encryptedPrivateData,
```

migrations/20220321220618-rename-virtual-card-epire-date-to-expiry-date.js
```diff
@@ -0,0 +1,57 @@
...
+        const originalPrivateData = JSON.parse(crypto.decrypt(virtualCard.privateData));
+        if (!(from in originalPrivateData)) {
+          console.debug(`'${from}' not in 'privateData' for VirtualCard with id=${virtualCard.id}`);
+          continue;
+        }
+        const privateData = {
+          ...omit(originalPrivateData, [from]),
+        };
+        privateData[to] = originalPrivateData[from];
+        const encryptedPrivateData = JSON.stringify(crypto.encrypt(privateData));
```
Betree, line 24: This fails with `RangeError: Invalid array length`.`crypto.encrypt` expects a string

```es6
const encryptedPrivateData = crypto.encrypt(JSON.stringify(privateData));
```

PR #7310, opened 2022-03-22
Correct `ts-rank` calculation when considering tags in search

SudharakaP:
This is with regard to the discussion at, https://github.com/opencollective/opencollective-api/pull/7284

It seems that with the aforementioned PR we are calculating the `ts-rank` incorrectly by changing the were clause thus always resulting in an incorrect sort. The purpose of that change is to support tags such as `jhipster-tag` where there is punctuation; which is an obvious use case. With the previous version it seems that if the user types jhipster-tag no results appear as array_to_tsvector seems to do some normalization. If they search for jhipstertag results appear but that isn't intuitive for the user.

However we need to make changes to the `TS_VECTOR` rather than changing the `WHERE` clause as mismatching these two results in an incorrect sort. Thus I am changing the code to what is suggested here; https://github.com/opencollective/opencollective-api/pull/7284#issuecomment-1074876710


Betree (commented):
Let's update the indexes too. You can copy-paste `migrations/20211123102842-add-search-collectives-index.js` and just replace the index content.

server/lib/search.js
```diff
@@ -76,7 +76,7 @@ export const TS_VECTOR = `
   to_tsvector('english', name)
   || to_tsvector('simple', slug)
   || to_tsvector('english', COALESCE(description, ''))
-  || COALESCE(array_to_tsvector(tags), '')
+  || COALESCE(to_tsvector(array_to_string(tags, ' ')), '')
```
Betree, line 79: These two both seems to fail:

```sql
SELECT array_to_string(ARRAY[])
SELECT array_to_string(NULL)
```

Let's make this a bit more resiliant:

```suggestion
  || COALESCE(to_tsvector(array_to_string(COALESCE(tags::varchar[], ARRAY[]::varchar[]), ' ')), '')
```

SudharakaP: Done. Thanks for pointing it out. 👍🏽

SudharakaP:
> Let's update the indexes too. You can copy-paste `migrations/20211123102842-add-search-collectives-index.js` and just replace the index content.

Thanks. I've updated the migration as well. 👍🏽

Betree (approved):

server/lib/search.js
SudharakaP: @Betree : Actually I realized that this one has a problem. When trying to execute this script we get, 

```
ERROR: functions in index expression must be marked IMMUTABLE
```
Will investigate this more. 🤔

SudharakaP:
@SudharakaP : Sorry for taking some time one this one, I was stuck with the defining of the function through Sequelize. Anyways, now I've seem to figured it out. Take a look and let me know if you see any issues. The issue here was that we had an immutable function `array_to_string` within the index creation which is not valid. So I created a wrapper function `array_to_string_immutable` so that we don't have this issue. 

Reference: https://dba.stackexchange.com/questions/299039/optimize-query-matching-first-n-items-of-an-array

Betree (approved):
Not able to test it for now, but all good for me if it works :+1:

PR #7311, opened 2022-03-23
Add no reply email address to "Expense Paid for" emails

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5351

image

### DIFF
--- a/server/graphql/v2/mutation/HostApplicationMutations.ts
+++ b/server/graphql/v2/mutation/HostApplicationMutations.ts
@@ -218,7 +218,7 @@ const sendPrivateMessage = async (host, collective, message: string): Promise<vo
   const adminUsers = await collective.getAdminUsers();
   await emailLib.send(
     'host.application.contact',
-    NO_REPLY_EMAIL,
+    [],
     {
       host: host.info,
       collective: collective.info,
@@ -226,6 +226,7 @@ const sendPrivateMessage = async (host, collective, message: string): Promise<vo
     },
     {
       bcc: adminUsers.map(u => u.email),
+      from: NO_REPLY_EMAIL,
     },
   );
 };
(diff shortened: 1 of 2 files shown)


server/graphql/v2/mutation/HostApplicationMutations.ts
```diff
@@ -218,14 +218,15 @@ const sendPrivateMessage = async (host, collective, message: string): Promise<vo
   const adminUsers = await collective.getAdminUsers();
   await emailLib.send(
     'host.application.contact',
-    NO_REPLY_EMAIL,
+    [],
```
SudharakaP, line 221: I think this is an bug where we put the `NO_REPLY_EMIL` on the recipient field that doesn't make sense. 🤔

Betree: This was done because `generateEmailFromTemplateAndSend` expects a `recipient` to be set, but passing an empty array indeed seems like a better hack :+1:

Betree (commented):
Other templates have been added to the [issue](https://github.com/opencollective/opencollective/issues/5351) since you opened this PR. Let's either add them here or not close the issue

RichardLitt:
Thanks everyone. I can open a new issue in the future for more of these, too - hopefully these three are it.

SudharakaP:
> Other templates have been added to the [issue](https://github.com/opencollective/opencollective/issues/5351) since you opened this PR. Let's either add them here or not close the issue

@Betree : I've updated the PR with the new ones @RichardLitt mentioned. 👍🏽

RichardLitt:
Thanks, @SudharakaP!

SudharakaP:
With regard to https://github.com/opencollective/opencollective/issues/5351#issuecomment-1125179426 I am closing this. We'll reopen and iterate if needed in the future.

PR #7317, opened 2022-03-24
Add support to query tag frequencies

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5333

Related front-end PR: https://github.com/opencollective/opencollective-frontend/pull/7611

### DIFF
--- a/server/graphql/v1/queries.js
+++ b/server/graphql/v1/queries.js
@@ -936,6 +936,10 @@ const queries = {
         type: new GraphQLList(CountryISO),
         description: 'Limit the search to collectives belonging to these countries',
       },
+      tags: {
+        type: new GraphQLList(GraphQLString),
+        description: 'Limit the search to collectives belonging to these tags',
+      },
       limit: {
         type: GraphQLInt,
         description: 'Limit the amount of results. Defaults to 20',
@@ -947,7 +951,7 @@ const queries = {
       },
     },
     async resolve(_, args, req) {
-      const { limit, offset, term, types, isHost, hostCollectiveIds, skipRecentAccounts, countries } = args;
+      const { limit, offset, term, types, isHost, hostCollectiveIds, skipRecentAccounts, countries, tags } = args;
       const cleanTerm = term ? term.trim() : '';
       logger.info(`Search Query: ${cleanTerm}`);
       const listToStr = list => (list ? list.join('_') : '');
@@ -975,6 +979,7 @@ const queries = {
           isHost,
           skipRecentAccounts,
           countries,
+          tags,
         });
         return generateResults(collectives, total);
       }
--- a/server/graphql/v1/queries.js
+++ b/server/graphql/v1/queries.js
@@ -936,10 +936,6 @@ const queries = {
         type: new GraphQLList(CountryISO),
         description: 'Limit the search to collectives belonging to these countries',
       },
-      tags: {
-        type: new GraphQLList(GraphQLString),
-        description: 'Limit the search to collectives belonging to these tags',
-      },
       limit: {
         type: GraphQLInt,
         description: 'Limit the amount of results. Defaults to 20',
@@ -951,7 +947,7 @@ const queries = {
       },
     },
     async resolve(_, args, req) {
-      const { limit, offset, term, types, isHost, hostCollectiveIds, skipRecentAccounts, countries, tags } = args;
+      const { limit, offset, term, types, isHost, hostCollectiveIds, skipRecentAccounts, countries } = args;
       const cleanTerm = term ? term.trim() : '';
       logger.info(`Search Query: ${cleanTerm}`);
       const listToStr = list => (list ? list.join('_') : '');
@@ -979,7 +975,6 @@ const queries = {
           isHost,
           skipRecentAccounts,
           countries,
-          tags,
         });
         return generateResults(collectives, total);
       }
--- a/server/graphql/v2/collection/TagStatsCollection.js
+++ b/server/graphql/v2/collection/TagStatsCollection.js
@@ -4,7 +4,7 @@ import { Collection, CollectionFields } from '../interface/Collection';
 import { TagStats } from '../object/TagStats';
 
 export const TagStatsCollection = new GraphQLObjectType({
-  name: 'TagCollection',
+  name: 'TagStatsCollection',
   interfaces: [Collection],
   description: 'A collection of "Tags"',
   fields: () => {
--- a/server/graphql/v2/collection/TagStatsCollection.js
+++ b/server/graphql/v2/collection/TagStatsCollection.js
@@ -1,0 +1,18 @@
+import { GraphQLList, GraphQLObjectType } from 'graphql';
+
+import { Collection, CollectionFields } from '../interface/Collection';
+import { TagStats } from '../object/TagStats';
+
+export const TagStatsCollection = new GraphQLObjectType({
+  name: 'TagCollection',
+  interfaces: [Collection],
+  description: 'A collection of "Tags"',
+  fields: () => {
+    return {
+      ...CollectionFields,
+      nodes: {
+        type: new GraphQLList(TagStats),
+      },
+    };
+  },
+});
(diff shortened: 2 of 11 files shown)


server/graphql/v2/query/index.js
```diff
@@ -39,6 +40,7 @@ const query = {
   orders: OrdersCollectionQuery,
   organization: OrganizationQuery,
   project: ProjectQuery,
+  tagFrequency: TagFrequencyQuery,
```
znarf, line 43: - We're usually naming collections `sCollectionQuery`, so here that would be TagFrequenciesCollectionQuery
- But why name it `tagFrequency` if it's a collection of `TagStats`, tagStats would make more sense, so `tagStats: TagStatsCollectionQuery`

SudharakaP: `tagStats` makes sense; I've refactored it. 👍🏽

server/graphql/v2/collection/TagStatsCollection.js
```diff
@@ -0,0 +1,18 @@
+import { GraphQLList, GraphQLObjectType } from 'graphql';
+
+import { Collection, CollectionFields } from '../interface/Collection';
+import { TagStats } from '../object/TagStats';
+
+export const TagStatsCollection = new GraphQLObjectType({
+  name: 'TagCollection',
```
znarf, line 7: Why the mismatch between `TagStatsCollection` and `TagCollection`?

SudharakaP: Sorry, mistake on my part when typing. I've corrected this. 🤔

znarf (changes requested):
Can you update the GraphQL schemas to their latest version?

znarf (commented):
I think it was a mistake in the past to name `TagStats` instead of simply `Tag`. But that out of scope for this PR.

SudharakaP:
> Can you update the GraphQL schemas to their latest version?

This is done. 👍🏽 

> I think it was a mistake in the past to name `TagStats` instead of simply `Tag`. But that out of scope for this PR.

That makes sense. I've created a new issue for this to rename; https://github.com/opencollective/opencollective/issues/5380

znarf (approved):
Good for me. Maybe @Betree for a small second look.

server/graphql/v1/queries.js
```diff
@@ -936,6 +936,10 @@ const queries = {
         type: new GraphQLList(CountryISO),
         description: 'Limit the search to collectives belonging to these countries',
       },
+      tags: {
```
Betree, line 939: According to https://github.com/opencollective/opencollective/issues/4378, shouldn't that be `tag`?

server/graphql/v2/query/collection/TagStatsCollectionQuery.js
```diff
@@ -0,0 +1,18 @@
...
+import { TagStatsCollection } from '../../collection/TagStatsCollection';
+
+const TagStatsCollectionQuery = {
+  type: new GraphQLNonNull(TagStatsCollection),
+  async resolve(_, args) {
+    const tagFrequencies = await rawQueries.getTagFrequencies({
+      ...pick(args, ['limit', 'offset']),
+    });
+
+    return { nodes: tagFrequencies, limit: args.limit, offset: args.offset };
```
Betree, line 14: I don't see any `args` defined here. Or am I missing something?

server/lib/search.js
```diff
@@ -143,6 +152,12 @@ export const searchCollectivesInDB = async (
     dynamicConditions += `AND "countryISO" IN (:countryCodes) `;
   }
 
+  if (tags?.length) {
+    tags.forEach(tag => {
+      dynamicConditions += `AND '${tag}' = ANY("tags") `;
```
Betree, line 157: :no_entry_sign: Never directly replace, that's the door for SQL injections. If you query this with something like:

```gql
search(tag: "xxx' = 'yyy' OR collective.data."anyPrivateField" LIKE '%bruteforce%' AND 'xxx") {
  ...
}
```

You'll have a nice surprise. 

Also, the Postgres containment operator seems like a better fit here:

```sql
# Will returns all collectives whose tags contains the array searched_tags
WHERE collective.tags @> :searched_tags
```

Betree (changes requested):

server/graphql/v1/queries.js
SudharakaP: Thanks for pointing it out. I've migrated this query to v2 so now this is no longer a problem. https://github.com/opencollective/opencollective-api/blob/3a48ed4a8f5df52d2358608faca041363ec5d1af/server/graphql/v2/query/collection/AccountsQuery.ts#L19

server/graphql/v2/query/collection/TagStatsCollectionQuery.js
SudharakaP: Ops, correct I've missed adding this to v2. Anyways, I've migrated this and the [relevant front-end PR](https://github.com/opencollective/opencollective-frontend/pull/7611) to use v2 now so this is fixed. 👍🏽

Betree: I still don't see any `args` being defined here

server/lib/search.js
```diff
@@ -143,6 +157,10 @@ export const searchCollectivesInDB = async (
     dynamicConditions += `AND "countryISO" IN (:countryCodes) `;
   }
 
+  if (tags?.length) {
+    dynamicConditions += `AND "tags" @> (:searchedTags) `;
+  }
+
```
Betree, line 163: With https://github.com/opencollective/opencollective-api/pull/7376 `tags` are now included in the stored TS vector. The search should therefore rely on TS vectors, they're weighted and indexed which makes them both more accurate and performant.

server/graphql/v2/query/collection/TagStatsCollectionQuery.js
SudharakaP: Sorry about this. I was a bit confused on this one. Initially I didn't add an `arg` here since we were tag frequencies based on all the collectives. I've now added support to only return tags based on the `searchTerm`.

server/lib/search.js
Betree: Forget about the previous comment. This approach is less performant, but since we're searching for explicitly passed tags (which we know exist in the DB), full-text search while more performant would be less accurate. We don't want to search for these tags in names and long descriptions.

All good to me!

server/graphql/v2/query/collection/TagStatsCollectionQuery.js
Betree: I still don't see `limit` and `offset` in `args`

SudharakaP: Oh... 🤯 .... I've added the `limit` and `offset` as well here. 👍🏽

server/lib/search.js
```diff
@@ -59,46 +60,37 @@ export const searchCollectivesByEmail = async (email, user, offset = 0, limit =
...
-  return term?.trim().replace(/\s+/g, ' ');
-};
-
-/**
- * Removes special ILIKE characters like `%
- */
-const sanitizeSearchTermForILike = term => {
-  return term.replace(/(_|%|\\)/g, '\\$1');
-};
-
```
Betree, line 84: I don't get why we're moving these. It makes sense to me to have helpers called **search**TermToTsVector, trim**Search**Term, or sanitize**Search**TermForILike (that are purely dedicated to help with search) exported from `lib/search`. If we put all shared code in `lib/utils` we're just going to end up with a gigantic file.

SudharakaP: That makes sense. I thought somehow that since we are using these multiple times they should go to `utils` but now when you point out that these are all related specifically for search, it makes sense.

SudharakaP:
I'll be merging this today if there's no objections as per https://opencollective.slack.com/archives/C035S573ZD2/p1649400119638669?thread_ts=1649368382.283779&cid=C035S573ZD2 all reviews are addressed.

znarf:
Let's merge and move forward please!

PR #7318, opened 2022-03-24
Remove unused thank you emails

Betree:
This email could not possibly be working as it was using the legacy `group` variable


templates/emails/thankyou.kendraio.hbs
```diff
@@ -1,77 +0,0 @@
-{{#if interval }}
-Subject: Thank you for your {{currency donation.amount currency=donation.currency}}/{{interval}} contribution to {{group.name}}
```
Betree, line 2: This :arrow_up: `group` was renamed to `collective` [in 2017](https://github.com/opencollective/opencollective-api/blob/0335e830f09ebbd89fa37d34531ce1e8a0267c67/migrations/archives/201707140000-GroupToCollective.js)

Betree:
@znarf I wonder if we shouldn't go the extra mile and remove custom emails for inactive collectives. E.g https://opencollective.com/laprimaire was a temporary initiative and hasn't seen any activity since 2019, yet we still have special rules for them in the code.

znarf:
@Betree anything that seems inactive should be removed, definitely.

server/lib/email.js
```diff
@@ -293,14 +293,8 @@ const generateEmailFromTemplate = (template, recipient, data = {}, options = {})
       template = 'thankyou.wwcode';
     } else if (['foundation', 'opensource'].includes(hostSlug)) {
       template = `thankyou.${hostSlug}`;
-    } else if (includes(['chsf', 'kendraio', 'brusselstogether', 'sustainoss', 'ispcwa'], slug)) {
```
Betree, line 296: - `chsf` & `ispcwa` templates were already removed
- `brusselstogether` has [no financial activity](https://opencollective.com/brusselstogether/transactions) since its creation
- `sustainoss` template could not be triggered as it's now hosted by OSC, which triggered the other template above (`else if (['foundation', 'opensource'].includes(hostSlug))`)

PR #7332, opened 2022-03-26
fix: update `nextChargeDate` while updating recurring contribution interval

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/5297

- [x] When updating from monthly to yearly: next charge date should be set to 1st January of next year
- [x] When updating from yearly to monthly
  - If the last charge was before the 15th of the month: the next charge date should be set to the 1st of next month
  - If the last charge was after the 15th of the month: the next charge date should be set to the 1st of the month after next month

### DIFF
--- a/server/lib/subscriptions.ts
+++ b/server/lib/subscriptions.ts
@@ -151,6 +151,25 @@ export const updateSubscriptionDetails = async (
     newSubscriptionData['interval'] = tier.interval;
   }
 
+  // Update next charge date
+  if (newOrderData['interval']) {
+    const previousInterval = order.interval;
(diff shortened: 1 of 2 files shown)


snitin315:
Not sure why these `UpdateMutation` tests are failing. Seems unrelated.

Screenshot 2022-03-28 at 7 07 16 AM

Betree (approved):
Looks perfect! We may want to unify the behavior of `getNextChargeDateForUpdateFromExternalToInternal`, but it's out of scope for this PR as all items from the issue have been tackled nicely.

I'll merge and follow up in a separate PR, will tag you for the review in case you want to have a look.

PR #7333, opened 2022-03-28
Fix: scheduleExpenseForPayment throw if PayPal is not connected

kewitz:
Adding a check before allowing a Host to schedule an expense for payment.
This covers the edge case where the Host requests access to the PayPal Payouts feature but does not connect to PayPal.
Keep in mind that we still allow users to submit PayPal expenses because the PayPal Adaptive Payment does not require a connected account.

### DIFF
--- a/server/graphql/common/features.ts
+++ b/server/graphql/common/features.ts
@@ -191,6 +191,15 @@ export const getFeatureStatusResolver =
           ? FEATURE_STATUS.ACTIVE // TODO: This flag is misused, there's a confusion between ACTIVE and AVAILABLE
           : FEATURE_STATUS.DISABLED;
       }
+      case FEATURE.PAYPAL_PAYOUTS: {
+        const hasConnectedAccount = await models.ConnectedAccount.count({
+          where: { service: 'paypal', CollectiveId: collective.id },
+          limit: 1,
+        });
+        return hasFeature(collective, FEATURE.PAYPAL_PAYOUTS) && hasConnectedAccount
+          ? FEATURE_STATUS.ACTIVE
+          : FEATURE_STATUS.DISABLED;
+      }
       default:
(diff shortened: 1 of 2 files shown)


Betree (commented):
To me, the "correct" fix would be to make sure we don't display "Schedule for payment" while the host is not truly connected to payouts. In other words, make sure we keep using adaptive until the host connects the new integration.

It's quite edge case, so I have no blocker if you want to merge this fix as-is; it's still an improvement on what we have today.

znarf:
While API does the job, I think we should try to fix it on the Frontend side.

server/graphql/common/features.ts
```diff
@@ -191,6 +191,15 @@ export const getFeatureStatusResolver =
...
           : FEATURE_STATUS.DISABLED;
       }
+      case FEATURE.PAYPAL_PAYOUTS: {
+        const hasConnectedAccount = await models.ConnectedAccount.count({
+          where: { service: 'paypal', CollectiveId: collective.id },
+          limit: 1,
+        });
+        return hasFeature(collective, FEATURE.PAYPAL_PAYOUTS) && hasConnectedAccount
+          ? FEATURE_STATUS.ACTIVE
+          : FEATURE_STATUS.DISABLED;
```
Betree, line 201: `hasFeature` is already checked in https://github.com/opencollective/opencollective-api/blob/f75c268e52270d11c941222c3f30e790f657fe0f/server/graphql/common/features.ts#L114, I don't think we need to do it again here

```suggestion
        const hasConnectedAccount = await models.ConnectedAccount.count({
          where: { service: 'paypal', CollectiveId: collective.id },
          limit: 1,
        });
        return hasConnectedAccount ? FEATURE_STATUS.ACTIVE : FEATURE_STATUS.DISABLED;
```

PR #7359, opened 2022-03-31
Add Support filtering without Search Term

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5328

Front-End PR: https://github.com/opencollective/opencollective-frontend/pull/7643

### DIFF
--- a/server/graphql/v2/query/collection/AccountsQuery.ts
+++ b/server/graphql/v2/query/collection/AccountsQuery.ts
@@ -58,8 +58,8 @@ const AccountsQuery = {
   async resolve(_: void, args): Promise<CollectionReturnType> {
     const { offset, limit } = args;
 
-    if (args.searchTerm) {
-      const cleanTerm = args.searchTerm.trim();
+    if (args.searchTerm || args.isHost || args.type || args.countries) {
+      const cleanTerm = args.searchTerm?.trim();
 
       const extraParameters = {
         orderBy: args.orderBy || { field: 'RANK', direction: 'DESC' },
--- a/server/graphql/v2/query/collection/AccountsQuery.ts
+++ b/server/graphql/v2/query/collection/AccountsQuery.ts
@@ -57,50 +57,31 @@ const AccountsQuery = {
   },
   async resolve(_: void, args): Promise<CollectionReturnType> {
     const { offset, limit } = args;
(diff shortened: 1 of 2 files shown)


server/graphql/v2/query/collection/AccountsQuery.ts
```diff
@@ -55,8 +55,8 @@ const AccountsQuery = {
   async resolve(_: void, args): Promise<CollectionReturnType> {
     const { offset, limit } = args;
 
-    if (args.searchTerm) {
-      const cleanTerm = args.searchTerm.trim();
+    if (args.searchTerm || args.isHost || args.type || args.countries) {
+      const cleanTerm = args.searchTerm?.trim();
```
znarf, line 59: That's not well thought for me. Can you consider a better way to switch between the `searchCollectivesInDB` behavior and the other one?

I think we should use `searchCollectivesInDB` by default and use the other behavior only when required.

SudharakaP: It seems that we use `models.Collective.findAndCountAll` when we want to filter out `supportedPaymentMethodService`. This is not something we support through `searchCollectivesInDB`. And thus I've changed the code to make sure we use `models.Collective.findAndCountAll` when we pass in `supportedPaymentMethodService` filter and `searchCollectivesInDB` for all other cases. Let me know your thoughts.

Betree: Not a fan of having this condition but it seems like the cleanest workaround for now. Note that `supportedPaymentMethodService` was introduced for Hacktoberfest and I'm not sure if anyone still uses it today. I wouldn't be against deprecation if it simplifies things.

I posted a message on [Slack](https://opencollective.slack.com/archives/C035S573ZD2/p1649834612770829) about discussing the separations of concerns - querying for accounts and searching. But that's for later.

Betree (approved):

server/graphql/v2/query/collection/AccountsQuery.ts
SudharakaP: @Betree : I've done some refactoring at, https://github.com/opencollective/opencollective-api/pull/7411 to remove the `supportedPaymentMethodService` as this is no longer used anywhere in the front-end. Was this used any place other than the front-end in the hacktoberfest. In any case since the Hacktoberfest is over I suggest we remove it anyways. 🤔

PR #7362, opened 2022-04-01
Create missing Expenses table index

kewitz:


Betree (approved):
Oh wow, we had no indexes on these this whole time? :sweat_smile:

kewitz:
> Oh wow, we had no indexes on these this whole time? 😅

Yes, François and I were checking DB performance earlier today and caught this, which should be related to the currently worst performing queries.

Betree:
> > Oh wow, we had no indexes on these this whole time? sweat_smile
> 
> Yes, François and I were checking DB performance earlier today and caught this, which should be related to the currently worst performing queries.

Great catch, that's going to improve performances in many places. I just tried running the migration locally, it works fine!

PR #7376, opened 2022-04-04
Search: store ts vectors in Collectives table

Betree:
Resolve https://github.com/opencollective/opencollective/issues/3825 with an alternative approach

As a bonus, we're now indexing `longDescription`.

# Stats

### Migration time

12.855s against a local prod DB copy

### Rows update time

Since we have an auto-generated column, I wanted to make sure that the row's update times were not going up. Testing updates on 100000 collectives with the following query showed no significant difference:

```sql
BEGIN;
EXPLAIN ANALYZE
UPDATE "Collectives" SET name = 'UPDATED' WHERE id < 100000 RETURNING *;
ROLLBACK;
```

Both results in something like:
> Update on "Collectives"  (cost=2081.10..25556.00 rows=98152 width=2440) (actual time=5.985..1771.365 rows=98944 loops=1)

### Search performance

Average: -98% execution time

Details:
- `open source`:
  - Before: `Limit  (cost=445094.12..445117.05 rows=20 width=1933) (actual time=1089.901..1092.494 rows=20 loops=1)`
  - After: `Limit  (cost=76.23..76.25 rows=5 width=1965) (actual time=39.911..39.917 rows=20 loops=1)`
  - Change: Complexity divided by 5857 (-99.98%), execution time divided by 28 (-96.42%)
- `foundation`: 
  - Before: `Limit  (cost=446345.38..446345.43 rows=20 width=1933) (actual time=1036.212..1036.278 rows=20 loops=1)`
  - After: `Limit  (cost=3624.67..3624.72 rows=20 width=1965) (actual time=4.827..4.832 rows=20 loops=1)`
  - Change: Complexity divided by 123 (-99.19%), execution time divided by 259 (-99.61%)
- `javascript`: 
  - Before: `Limit  (cost=446345.38..446345.43 rows=20 width=1933) (actual time=1074.604..1074.683 rows=20 loops=1)`
  - After: `Limit  (cost=1933.89..1933.94 rows=20 width=1965) (actual time=7.044..7.048 rows=20 loops=1)`
  - Change: Complexity divided by 231 (-99.57%), execution time divided by 153 (-99.35%)

<details>

Helpers:

```es6
 getRate = (from, to) => (to - from) / from * 100
compute = (from, to, timeBefore, timeAfter) => console.log(`Complexity divided by ${(from / to).toFixed(0)} (${getRate(from, to).toFixed(2)}%), execution time divided by ${(timeBefore / timeAfter).toFixed(0)} (${getRate(timeBefore, timeAfter).toFixed(2)}%)`)
```

</details>

### DIFF
--- a/server/lib/search.js
+++ b/server/lib/search.js
@@ -83,18 +83,6 @@ const sanitizeSearchTermForILike = term => {
 };
 
 /**
- * TSVector to search for collectives names/description/slug
- * Updating the value here requires generating a new migration to update the index.
- * See `migrations/20201119100223-update-collectives-search-index.js`
- */
-export const TS_VECTOR = `
-  to_tsvector('english', name)
-  || to_tsvector('simple', slug)
-  || to_tsvector('english', COALESCE(description, ''))
-  || COALESCE(to_tsvector(array_to_string(COALESCE(tags::varchar[], ARRAY[]::varchar[]), ' ')), '')
-`;
-
-/**
  * Search collectives directly in the DB, using a full-text query.
  */
 export const searchCollectivesInDB = async (
@@ -153,10 +141,8 @@ export const searchCollectivesInDB = async (
     } else {
       isUsingTsVector = true;
       dynamicConditions += `
-        AND (${TS_VECTOR} @@ plainto_tsquery('english', :vectorizedTerm)
-        OR ${TS_VECTOR} @@ plainto_tsquery('simple', :vectorizedTerm)
-        OR name ILIKE '%' || :term || '%'
-        OR slug ILIKE '%' || :term || '%')`;
+        AND ("searchTsVector" @@ plainto_tsquery('english', :vectorizedTerm)
+        OR "searchTsVector" @@ plainto_tsquery('simple', :vectorizedTerm))`;
     }
   } else {
     term = '';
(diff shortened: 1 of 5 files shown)


znarf (approved):
Looks amazing.

Betree:
Moving this to staging then production to try it live

server/lib/search.js
```diff
@@ -153,10 +141,8 @@ export const searchCollectivesInDB = async (
     } else {
       isUsingTsVector = true;
       dynamicConditions += `
-        AND (${TS_VECTOR} @@ plainto_tsquery('english', :vectorizedTerm)
-        OR ${TS_VECTOR} @@ plainto_tsquery('simple', :vectorizedTerm)
-        OR name ILIKE '%' || :term || '%'
-        OR slug ILIKE '%' || :term || '%')`;
```
SudharakaP, line 159: Remove these two will make the search term be exact as evidenced here; https://opencollective.slack.com/archives/G46PNRCGP/p1649215867073719. As you can see Alina is searching for `1k proj` and there's no search results whereas the exact term `1k project` gives the correct results. I think we should have these two regardless of the `searchTsVector` since the name and slug could be spelled partway (especially if it's too long). 🤔

Betree: @SudharakaP TS vectors are meant for full-text search, they offer more options and better performances than ILIKEs. There are ways to do partial searches with them, let's not bring that back.

SudharakaP: Proposing fix at, https://github.com/opencollective/opencollective-api/pull/7403

PR #7383, opened 2022-04-06
Update Expense Item amount when processing virtual card charge

znarf:


Betree (commented):
Looks like this was done in https://github.com/opencollective/opencollective-api/blob/39645b2603f30e174451a102dc21f78da7ba1fd9/server/paymentProviders/utils.js#L71

PR #7388, opened 2022-04-07
Fix Wise webhook response time

kewitz:
It turns out the problem wasn't the index but the way Sequelize would build up the query.

Originally, Sequelize would cast the ID number to double-precision, effectively bypassing our index:
```
WHERE ("Transaction"."deletedAt" IS NULL
AND (CAST(("Transaction"."data"#>>'{transfer,id}') AS DOUBLE PRECISION) = 392897660
AND "Transaction"."updatedAt" >= '2022-03-28 00:59:24.757 +00:00')) LIMIT 1;
```

With this update, we get rid of that forced casting and we end up hitting the index when analyzing the query:
```
WHERE ("Transaction"."deletedAt" IS NULL
AND (("Transaction"."data"#>>'{transfer,id}') = '392897660'
AND "Transaction"."updatedAt" >= '2022-03-28 01:07:05.081 +00:00')) LIMIT 1;
```

### DIFF
--- a/server/paymentProviders/transferwise/webhook.ts
+++ b/server/paymentProviders/transferwise/webhook.ts
@@ -1,4 +1,5 @@
 import { Request } from 'express';
+import { toString } from 'lodash';
 import moment from 'moment';
 import { Op } from 'sequelize';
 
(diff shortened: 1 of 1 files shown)


znarf (approved):
Looks good!

Let's create an issue for the more long term solution (dedicated column?)

PR #7397, opened 2022-04-08
Stripe: Add issuing_card.updated event listener

kewitz:
So we can keep cards in sync with Stripe both ways.


znarf:
Can you confirm that we're also handling this case properly in the synchronization script? reconcile-transactions.

znarf (approved):

kewitz:
> Can you confirm that we're also handling this case properly in the synchronization script? reconcile-transactions.

Yes, we're doing in that script as well.

PR #7403, opened 2022-04-11
Add Support to Search Parts of Strings in `/search`

SudharakaP:
We need to give ability for users to search part of text in the `/search` page. For example if there's a collective called with text `jhipster` included in one of it's fields it should appear when a user searches for `jhips` or `jhipste` etc. This PR solves that. 

Related Discussion: https://github.com/opencollective/opencollective-api/pull/7376#discussion_r844297165

### DIFF
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -1161,16 +1161,16 @@ const getTagFrequencies = async args => {
   let searchTermFragment = '';
   let term = args.searchTerm;
   // Cleanup term
+  term = sanitizeSearchTermForILike(trimSearchTerm(term));
   if (term && term.length > 0) {
-    term = sanitizeSearchTermForILike(trimSearchTerm(term));
     if (term[0] === '@') {
       // When the search starts with a `@`, we search by slug only
       term = term.replace(/^@+/, '');
       searchTermFragment = `AND slug ILIKE '%' || :term || '%' `;
     } else {
-      searchTermFragment = `
-        AND ("searchTsVector" @@ plainto_tsquery('english', :vectorizedTerm)
-        OR "searchTsVector" @@ plainto_tsquery('simple', :vectorizedTerm))`;
+      searchTermFragment += `
+        AND ("searchTsVector" @@ to_tsquery('english', :vectorizedTerm':*')
+        OR "searchTsVector" @@ to_tsquery('simple', :vectorizedTerm':*'))`;
     }
   } else {
     term = '';
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -11,7 +11,6 @@ import { PayoutMethodTypes } from '../models/PayoutMethod';
 
 import { memoize } from './cache';
 import { convertToCurrency } from './currency';
-import { sanitizeSearchTermForILike, searchTermToTsVector, trimSearchTerm } from './search';
 import sequelize, { Op } from './sequelize';
 import { amountsRequireTaxForm } from './tax-forms';
 import { computeDatesAsISOStrings } from './utils';
@@ -1154,50 +1153,6 @@ const getTransactionsTimeSeries = async (
   );
 };
 
-/**
- * Returns tags along with their frequency of use.
- */
-const getTagFrequencies = async args => {
-  let searchTermFragment = '';
-  let term = args.searchTerm;
-
-  if (term && term.length > 0) {
-    // Cleanup term
-    term = sanitizeSearchTermForILike(trimSearchTerm(term));
-    if (term[0] === '@') {
-      // When the search starts with a `@`, we search by slug only
-      term = term.replace(/^@+/, '');
-      searchTermFragment = `AND slug ILIKE '%' || :term || '%' `;
-    } else {
-      searchTermFragment += `
-        AND ("searchTsVector" @@ to_tsquery('english', :vectorizedTerm':*')
-        OR "searchTsVector" @@ to_tsquery('simple', :vectorizedTerm':*'))`;
-    }
-  } else {
-    term = '';
-  }
-
-  return sequelize.query(
-    `SELECT UNNEST(tags) AS tag, COUNT(id)
-      FROM "Collectives"
-      WHERE "deletedAt" IS NULL
-      ${searchTermFragment}
-      GROUP BY UNNEST(tags)
-      ORDER BY count DESC
-      LIMIT :limit
-      OFFSET :offset`,
-    {
-      type: sequelize.QueryTypes.SELECT,
-      replacements: {
-        term,
-        vectorizedTerm: searchTermToTsVector(term),
-        limit: args.limit,
-        offset: args.offset,
-      },
-    },
-  );
-};
-
 const serializeCollectivesResult = JSON.stringify;
 
 const unserializeCollectivesResult = string => {
@@ -1231,7 +1186,6 @@ const queries = {
   getMembersOfCollectiveWithRole,
   getMembersWithBalance,
   getMembersWithTotalDonations,
-  getTagFrequencies,
   getTaxFormsRequiredForAccounts,
   getTaxFormsRequiredForExpenses,
   getTopBackers,
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -1160,9 +1160,10 @@ const getTransactionsTimeSeries = async (
 const getTagFrequencies = async args => {
   let searchTermFragment = '';
   let term = args.searchTerm;
-  // Cleanup term
-  term = sanitizeSearchTermForILike(trimSearchTerm(term));
+
   if (term && term.length > 0) {
+    // Cleanup term
+    term = sanitizeSearchTermForILike(trimSearchTerm(term));
     if (term[0] === '@') {
       // When the search starts with a `@`, we search by slug only
       term = term.replace(/^@+/, '');
--- a/server/lib/search.js
+++ b/server/lib/search.js
@@ -155,9 +155,9 @@ export const searchCollectivesInDB = async (
     dynamicConditions += `AND "tags" @> (:searchedTags) `;
   }
 
-  // Cleanup term
-  term = sanitizeSearchTermForILike(trimSearchTerm(term));
   if (term && term.length > 0) {
+    // Cleanup term
+    term = sanitizeSearchTermForILike(trimSearchTerm(term));
     if (term[0] === '@') {
       // When the search starts with a `@`, we search by slug only
       term = term.replace(/^@+/, '');
--- a/server/lib/search.js
+++ b/server/lib/search.js
@@ -60,12 +60,17 @@ export const searchCollectivesByEmail = async (email, user, offset = 0, limit = 
 };
 
 /**
- * Turn a search string into a TS vector using 'OR' operator.
+ * Sanitize and then turn a search string into a TS vector using 'OR' operator.
  *
- * Ex: "open potatoes" => "open|potatoes"
+ * Examples: "open potatoes" => "open|potatoes", "crème brulée => "creme|brulee"
+ *
  */
 export const searchTermToTsVector = term => {
-  return term.replace(/\s+/g, '|');
+  const sanitizedTerm = term
+    .normalize('NFD')
+    .replace(/[\u0300-\u036f]/g, '')
+    .replace(/[^a-zA-Z0-9-\/_ ]/g, '');
+  return sanitizedTerm.trim().replace(/\s+/g, '|');
 };
 
 /**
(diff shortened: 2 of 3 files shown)


Betree (commented):
Tests don't seem to be passing

image


Did you try to estimate the complexity/performances impact of this change?

It looks like a good improvement overall. Note that it won't work for multi-terms search, for example searching for `hel test` won't match the string `hello world` because the tsquery will look be generated as `'hel' | 'test':*`. We can keep that as a further improvement.

server/lib/search.js
```diff
@@ -82,6 +82,13 @@ export const sanitizeSearchTermForILike = term => {
   return term.replace(/(_|%|\\)/g, '\\$1');
 };
 
+/**
+ * Sanitize search term for to_tsquery
+ */
+export const sanitizeSearchTerm = term => {
+  return term.replace(/[^a-zA-Z0-9 ]/g, '');
```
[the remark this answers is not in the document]
SudharakaP: Thanks for the ideas. I've incorporated these and refactored the logic to `searchTermToTsVector`. 👍🏽

SudharakaP:
@Betree : 

> Tests don't seem to be passing

Sorry I thought I fixed this, but somehow seems to have missed. This is due to a missing `trim` which we have to apply after we replace things within `searchTermToTsVector`; 

https://github.com/opencollective/opencollective-api/blob/df649ca50614c9899ccf29680b3eb3471222b26f/server/lib/search.js#L73

As otherwise it can end up with a trailing space character. 

> Did you try to estimate the complexity/performances impact of this change?

I did a `EXPLAIN ANALYSE` on this and here are the results. This was done for three search terms `test`, `jhipster` and `webpack` no `prod`. 

1) For `test` it seems that the execution time fluctuates between, 100 and 800 before and after the changes with most runs around the 100 mark. Some random runs,

**After**

Limit  (cost=72896.45..72896.46 rows=20 width=1770) (actual time=100.994..101.000 rows=20 loops=1)
Limit  (cost=72896.45..72896.46 rows=20 width=1770) (actual time=160.661..160.669 rows=20 loops=1)
Limit  (cost=72896.45..72896.46 rows=20 width=1770) (actual time=797.467..797.475 rows=20 loops=1)

**Before**

Limit  (cost=19197.40..19197.41 rows=20 width=1770) (actual time=83.504..83.510 rows=20 loops=1)
Limit  (cost=19197.40..19197.41 rows=20 width=1770) (actual time=90.014..90.021 rows=20 loops=1)
Limit  (cost=19197.40..19197.41 rows=20 width=1770) (actual time=131.378..131.386 rows=20 loops=1)
Limit  (cost=19197.40..19197.41 rows=20 width=1770) (actual time=517.688..517.697 rows=20 loops=1)
Limit  (cost=19197.40..19197.41 rows=20 width=1770) (actual time=759.036..759.045 rows=20 loops=1)

2) For `webpack` it seems that the execution time seems to be around 60-200 before and after the changes with most runs below 100. Some random runs,

**After**
Limit  (cost=72898.45..72898.46 rows=20 width=1770) (actual time=63.573..63.580 rows=20 loops=1)
Limit  (cost=72898.45..72898.46 rows=20 width=1770) (actual time=62.612..62.618 rows=20 loops=1)
Limit  (cost=72898.45..72898.46 rows=20 width=1770) (actual time=110.816..110.822 rows=20 loops=1)
Limit  (cost=72898.45..72898.46 rows=20 width=1770) (actual time=68.624..68.631 rows=20 loops=1)

**Before**
Limit  (cost=19199.40..19199.41 rows=20 width=1770) (actual time=230.780..230.789 rows=20 loops=1)
Limit  (cost=19199.40..19199.41 rows=20 width=1770) (actual time=62.878..62.885 rows=20 loops=1)
Limit  (cost=19199.40..19199.41 rows=20 width=1770) (actual time=62.449..62.456 rows=20 loops=1)
Limit  (cost=19199.40..19199.41 rows=20 width=1770) (actual time=233.694..233.704 rows=20 loops=1)
Limit  (cost=19199.40..19199.41 rows=20 width=1770) (actual time=62.242..62.249 rows=20 loops=1)

3) For `jhipster` it seems that the execution time seems to be around 60-200 before and after the changes with most runs below 100. Some random runs,

**After**
Limit  (cost=72898.45..72898.46 rows=20 width=1770) (actual time=20.393..20.397 rows=2 loops=1)
Limit  (cost=72898.45..72898.46 rows=20 width=1770) (actual time=20.297..20.301 rows=2 loops=1)
Limit  (cost=72898.45..72898.46 rows=20 width=1770) (actual time=24.298..24.301 rows=2 loops=1)
Limit  (cost=72898.45..72898.46 rows=20 width=1770) (actual time=20.520..20.523 rows=2 loops=1)

**Before**
Limit  (cost=19199.40..19199.41 rows=20 width=1770) (actual time=20.330..20.333 rows=2 loops=1)
Limit  (cost=19199.40..19199.41 rows=20 width=1770) (actual time=20.258..20.261 rows=2 loops=1)
Limit  (cost=19199.40..19199.41 rows=20 width=1770) (actual time=20.140..20.143 rows=2 loops=1)
Limit  (cost=19199.40..19199.41 rows=20 width=1770) (actual time=122.278..122.283 rows=2 loops=1)

So overall I don't see a difference in the performance. 🤔 
 
> It looks like a good improvement overall. Note that it won't work for multi-terms search, for example searching for `hel test` won't match the string `hello world` because the tsquery will look be generated as `'hel' | 'test':*`. We can keep that as a further improvement.

I am wondering if this is really needed at this point. Should we work on this right away or do you think we should wait to see if there's interest by the users for this kinda multi term searches? The more the search gets smarter it might be best to invest on something like Algolia? 🤔

Betree:
> Before
> Limit (cost=19199.40..19199.41 rows=20 width=1770) (actual time=20.330..20.333 rows=2 loops=1)
> 
> After
> Limit (cost=72898.45..72898.46 rows=20 width=1770) (actual time=20.520..20.523 rows=2 loops=1)
>
> So overall I don't see a difference in the performance. thinking

Well, the cost going from `19199` to `72898` is a x 4 impact on complexity :-) Did you run these tests against a DB with some prod data? Differences can be significant with big DBs. I'm also surprised by the numbers in "Before" as I was running queries like `open source` with a complexity close to 2000 in https://github.com/opencollective/opencollective-api/pull/7376.

> > It looks like a good improvement overall. Note that it won't work for multi-terms search, for example searching for hel test won't match the string hello world because the tsquery will look be generated as 'hel' | 'test':*. We can keep that as a further improvement.
> 
> I am wondering if this is really needed at this point. Should we work on this right away or do you think we should wait to see if there's interest by the users for this kinda multi term searches? The more the search gets smarter it might be best to invest on something like Algolia? thinking

As per my comment: `We can keep that as a further improvement.`

Betree:
Doesn't have to be in this PR, but it would be great to have automated tests to benchmark the perfs/complexity of the search and track its evolution easily.

SudharakaP:
> Well, the cost going from `19199` to `72898` is a x 4 impact on complexity :-) Did you run these tests against a DB with some prod data? Differences can be significant with big DBs. I'm also surprised by the numbers in "Before" as I was running queries with a complexity close to 2000 in #7376.

Yes I ran these against prod. But you are right seems x4 complexity, I was looking at the `actual time`. Of course the numbers seem small when running against smaller dbs. For example in the local db it's like, 🤔 

Before for `jhipster` search: Limit  (cost=88.61..88.62 rows=1 width=1412) (actual time=0.283..0.285 rows=12 loops=1)
After for `jhipster` search: Limit  (cost=111.76..111.77 rows=1 width=1412) (actual time=0.273..0.275 rows=12 loops=1)

So are these significant differences only on big DBs something we should investigate? I am a bit confused why it gives a x4 increase in prod but not in local. I was expecting a similar cost increase. 🤔 

> As per my comment: `We can keep that as a further improvement.`

I've created https://github.com/opencollective/opencollective/issues/5449 to track that. 

> Doesn't have to be in this PR, but it would be great to have automated tests to benchmark the perfs/complexity of the search and track its evolution easily.

I've opened a PR for this at, https://github.com/opencollective/opencollective/issues/5450. I am not exactly sure what the benchmark should be. I was thinking maybe we could see if the time to return the results do not exceed 1.5x of the time on previous runs. But let me know your thoughts on this. 🤔

server/lib/search.js
```diff
@@ -151,17 +156,17 @@ export const searchCollectivesInDB = async (
   }
 
   // Cleanup term
+  term = sanitizeSearchTermForILike(trimSearchTerm(term));
   if (term && term.length > 0) {
-    term = sanitizeSearchTermForILike(trimSearchTerm(term));
     if (term[0] === '@') {
```
Betree, line 161: Why are we getting that `term = ` out of the `if (term && term.length > 0) {`? I don't see the goal of saniziting if it's empty, and it will crash if term is nullish.

server/lib/queries.js
```diff
@@ -1161,16 +1161,16 @@ const getTagFrequencies = async args => {
   let searchTermFragment = '';
   let term = args.searchTerm;
   // Cleanup term
+  term = sanitizeSearchTermForILike(trimSearchTerm(term));
   if (term && term.length > 0) {
-    term = sanitizeSearchTermForILike(trimSearchTerm(term));
     if (term[0] === '@') {
```
Betree, line 1166: Same stuff here, I don't get why we're moving that out of the `if (term`

SudharakaP: This is a mistake on my part I think, we should have this within. 🤔 I've corrected that. 👍🏽

server/lib/search.js
SudharakaP: Corrected this one as well. 👍🏽

SudharakaP:
> > So are these significant differences only on big DBs something we should investigate? I am a bit confused why it gives a x4 increase in prod but not in local. I was expecting a similar cost increase.
> 
> The most likely cause in these cases is a n+1 query, aka complexity increases with the number of rows. Let's say that, for each row, we trigger a subquery to fetch related data. If there are 3 rows, then you'll have a total of 4 queries (1 for the main results, and one for each row). So complexity goes from 1 to 4 (x4). Now if you run the same query with 1000 rows, you'll get a total 1001 queries, so the complexity goes from 1 to 1001 (x1001).
> 
> Overall the performance impact of this change seems under control, especially since most of the complexity increases compared to #7376 seem related to #7351. Merging this change then focusing on optimizing the sort by ACTIVITY is probably the most impactful thing to do.

@Betree : Thanks much for the explanation. That makes perfect sense. I've corrected the other parts of this PR. 👍🏽

server/lib/search.js
```diff
@@ -60,12 +60,17 @@ export const searchCollectivesByEmail = async (email, user, offset = 0, limit =
...
+ * Examples: "open potatoes" => "open|potatoes", "crème brulée => "creme|brulee"
  *
- * Ex: "open potatoes" => "open|potatoes"
  */
 export const searchTermToTsVector = term => {
-  return term.replace(/\s+/g, '|');
+  const sanitizedTerm = term
+    .normalize('NFD')
+    .replace(/[\u0300-\u036f]/g, '')
+    .replace(/[^a-zA-Z0-9-\/_ ]/g, '');
```
Betree, line 72: We should add a comment to explain this one. We both know what this is, but the next person who's gonna look at this code will have a hard time understanding this. Plus variable name is not ideal, we're not really sanitizing here, just removing the accents. I can suggest `termWithoutAccents`

server/lib/queries.js
```diff
@@ -1160,17 +1160,18 @@ const getTransactionsTimeSeries = async (
...
     term = sanitizeSearchTermForILike(trimSearchTerm(term));
     if (term[0] === '@') {
       // When the search starts with a `@`, we search by slug only
       term = term.replace(/^@+/, '');
       searchTermFragment = `AND slug ILIKE '%' || :term || '%' `;
     } else {
-      searchTermFragment = `
-        AND ("searchTsVector" @@ plainto_tsquery('english', :vectorizedTerm)
-        OR "searchTsVector" @@ plainto_tsquery('simple', :vectorizedTerm))`;
+      searchTermFragment += `
```
Betree, line 1172: :ice_cream: IMO this `getTagFrequencies` has its place in `server/lib/search.js`. I'm open to other places, but `server/lib/queries.js` is a black hole that we should avoid when possible. Putting all our SQL queries in the same file seems a bit messy.

Regarding the change on the line, going from `=` to `+=` seems unnecessary and forces you to look at the previous code to understand what we're concatening our string with. Unless there's a good reason to change,I would keep it consistent and remove the `+`

Betree (commented):
Some feedback below, feel free to merge when you think this is ready

server/lib/search.js
SudharakaP: The proper term I think is [Diacritic](https://en.wikipedia.org/wiki/Diacritic). So I renamed this to, `termWithoutDiacritics`.

server/lib/queries.js
SudharakaP: Makes sense. I've made the changes.

SudharakaP:
> Some feedback below, feel free to merge when you think this is ready

Thanks for the reviews. Will be merging this when CI passes.

PR #7411, opened 2022-04-13
Remove the older `supportedPaymentMethodService` field and relavant logic

SudharakaP:
Related to the discussion at, https://github.com/opencollective/opencollective-api/pull/7359#discussion_r849163069. We are not using `supportedPaymentMethodService` and we should be default use the `searchCollectivesInDB` when we are doing search. Thus I propose refactoring this part to remove the unnecessary logic.

### DIFF
--- a/server/graphql/schemaV2.graphql
+++ b/server/graphql/schemaV2.graphql
@@ -9563,14 +9563,6 @@ type Query {
     hasCustomContributionsEnabled: Boolean
 
     """
-    Only accounts that support one of these payment services will be returned
-    """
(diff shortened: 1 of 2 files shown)


Betree (commented):
Not the way to go with the public API. First, deprecate then remove after some time.

We don't know whether people depend on this or not.

znarf (changes requested):
Deprecate first.

SudharakaP:
> Deprecate first.

Done at, https://github.com/opencollective/opencollective-api/pull/7440

Betree (approved):
Rebased & solved conflicts

PR #7412, opened 2022-04-15
fix: update editMember mutation

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/5394

### DIFF
--- a/server/graphql/v2/mutation/MemberMutations.js
+++ b/server/graphql/v2/mutation/MemberMutations.js
@@ -148,19 +148,20 @@ const memberMutations = {
       // Edit member
       const editableAttributes = pick(args, ['role', 'description', 'since']);
 
-      const [, members] = await models.Member.update(editableAttributes, {
-        returning: true,
+      const member = await models.Member.findOne({
         where: {
           MemberCollectiveId: memberAccount.id,
           CollectiveId: account.id,
         },
       });
 
-      if (!members.length) {
+      if (!member) {
         throw new ValidationFailed(`Member ${memberAccount.slug} does not exist in Collective ${account.slug}`);
       }
 
-      return members[0];
+      await member.update(editableAttributes);
+
+      return member;
     },
   },
   removeMember: {


server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -148,19 +148,20 @@ const memberMutations = {
       // Edit member
       const editableAttributes = pick(args, ['role', 'description', 'since']);
 
-      const [, members] = await models.Member.update(editableAttributes, {
```
snitin315, line 151: /cc @Betree 
Not sure why, but sometimes `.update` was creating a new instance of the member with a different id as well while updating the original instance too. Which created duplicate entries in the database.

I suppose we need a migration here to delete those duplicated entries on production.

Betree: Is it a bug with sequelize itself? Do you have a reproducible example to easily test this?

We use the `{model}.Update` method in multiple places, so if that's the case the consequences may be bad and we'll need to take a deep look at the rest of the code.

snitin315: Could be a bug in sequealize itself after the recent upgrade. I will try creating a reproducible test case.

Actually there are two ways to update:

>there's an update method for the Model and a separate update method for an Instance (record). Model.update() updates ALL matching records and returns an array see Sequelize documentation. Instance.update() updates the record and returns an instance object.

https://stackoverflow.com/questions/8158244/how-to-update-a-record-using-sequelize-for-node/59744504#59744504

Here we were using `Model.update` and this seems to be the only place where we use this. Everywhere else we have `Instance.update`. Need to check once again.

WikiRik: I do not think the recent update causes this, but we're happy to look into this if you can reproduce this

Betree: > Here we were using `Model.update` and this seems to be the only place where we use this. Everywhere else we have `Instance.update`. Need to check once again.

It's not the only place, see [this one](https://github.com/opencollective/opencollective-api/blob/45cd9922bbd6697ccee900b190b49eb8ccb10ab2/server/graphql/common/expenses.ts#L1110) or [this one](https://github.com/opencollective/opencollective-api/blob/45cd9922bbd6697ccee900b190b49eb8ccb10ab2/server/graphql/v1/mutations/paymentMethods.js#L219) for example. A search with `models\..+\.update` returns 49 results for API (and it may miss some when models are imported directly).

A bug in Sequelize is possible, but we should suspect our own code first. Is the Frontend always passing an `id` when editing? Is it calling the right mutation?

We need a clear understanding and reproduction of the issue before jumping on the fix.

snitin315: Yeah, those many instances could be problematic. Let me try to write a simple test case and double-check everything closely.

snitin315: I am not able to reproduce this issue of multiple cards locally anymore, it could be a cache issue then.
I was not able to create a reproducible example.

snitin315: But the query is actually returning 4 member nodes, all for the same member but each node has a different id. Can it be because of cache?

Screenshot 2022-04-24 at 9 25 51 AM
Screenshot 2022-04-24 at 9 26 28 AM

Betree: Different members IDs mean we've created multiple members entries, it's not a cache issue. It's however difficult to go further without a reproducible example, your local data could be corrupted for other reasons.

Betree: @snitin315 Please let us know if you'd prefer to drop the investigation, this issue is a real problem so we want to iterate as quickly as possible.

snitin315: @Betree yeah sure, dropping it off.

PR #7413, opened 2022-04-15
Fixes and unit tests for budget queries

Betree:
Fix https://github.com/opencollective/opencollective/issues/4420

### DIFF
--- a/server/graphql/loaders/index.js
+++ b/server/graphql/loaders/index.js
@@ -377,7 +377,8 @@ export const loaders = req => {
         SELECT o."TierId" AS "TierId", COALESCE(SUM(s."amount"), 0) AS "total"
         FROM "Orders" o
         INNER JOIN "Subscriptions" s ON o."SubscriptionId" = s.id
-        WHERE "TierId" IN (?)
+        WHERE o."TierId" IN (?)
+        AND o."deletedAt" IS NULL
         AND s."isActive" = TRUE
         AND s."interval" = 'month'
         GROUP BY "TierId";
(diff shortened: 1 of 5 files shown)


server/graphql/loaders/index.js
```diff
@@ -377,7 +377,8 @@ export const loaders = req => {
         SELECT o."TierId" AS "TierId", COALESCE(SUM(s."amount"), 0) AS "total"
         FROM "Orders" o
         INNER JOIN "Subscriptions" s ON o."SubscriptionId" = s.id
-        WHERE "TierId" IN (?)
+        WHERE o."TierId" IN (?)
+        AND o."deletedAt" IS NULL
```
Betree, line 381: This should also optimize the query since the index ignores deleted orders

server/lib/budget.js
```diff
@@ -374,45 +374,42 @@ export async function getYearlyIncome(collective) {
...
-          COALESCE(SUM("netAmountInCollectiveCurrency"*12),0) FROM "activeMonthlySubscriptions")
+        )
         +
-        (SELECT
-          COALESCE(SUM(t."netAmountInCollectiveCurrency"),0) FROM "Transactions" t
-          LEFT JOIN "Orders" d ON t."OrderId" = d.id
-          LEFT JOIN "Subscriptions" s ON d."SubscriptionId" = s.id
-          WHERE t."CollectiveId" = :CollectiveId
-            AND t."RefundTransactionId" IS NULL
-            AND CAST(("t"."data"#>>'{internal}') AS BOOLEAN) IS NOT true
```
znarf, line 399: We agreed that it was ok to try to stop supporting this property. Because it's not actively used since around 1 year. We will re-introduced if needed.

znarf (approved):

PR #7415, opened 2022-04-15
Misc improvements for update credit card page

Betree:
Part of https://github.com/opencollective/opencollective/issues/5440
Require https://github.com/opencollective/opencollective-frontend/pull/7698

### DIFF
--- a/server/graphql/v1/queries.js
+++ b/server/graphql/v1/queries.js
@@ -875,9 +875,17 @@ const queries = {
       id: { type: GraphQLInt },
       code: { type: GraphQLString },
     },
-    resolve(_, args) {
+    async resolve(_, args, req) {
       if (args.id) {
-        return models.PaymentMethod.findByPk(args.id);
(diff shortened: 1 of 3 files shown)


server/graphql/v1/queries.js
```diff
@@ -874,9 +874,17 @@ const queries = {
...
     },
-    resolve(_, args) {
+    async resolve(_, args, req) {
       if (args.id) {
-        return models.PaymentMethod.findByPk(args.id);
+        const paymentMethod = await models.PaymentMethod.findByPk(args.id, {
+          include: [{ model: models.Collective, required: true }],
+        });
+
+        if (!paymentMethod || !req.remoteUser?.isAdminOfCollective(paymentMethod.Collective)) {
```
Betree, line 883: This permission check wasn't really an issue (we already check private fields) but it still allowed to enumerate all payment methods. We only fetch by `id` on the page to update the payment method, so there should be no impact.

znarf (approved):

PR #7425, opened 2022-04-20
Fix forever processing Stripe Virtual Card charges

kewitz:
Fixes https://github.com/opencollective/opencollective/issues/5439


server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -217,6 +217,22 @@ export const processDeclinedAuthorization = async (stripeAuthorization, stripeEv
...
+
+  if (!expense) {
+    logger.error(
+      `Stripe: could not find expense attached to reversed authorization ${stripeAuthorization.id}`,
+      stripeEvent,
+    );
+    return;
+  }
+  await expense.update({ status: ExpenseStatus.CANCELED });
+
```
znarf, line 235: Could you add a comment and explain when this can be happening from the user standpoint?

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -247,6 +263,37 @@ export const processTransaction = async (stripeTransaction, stripeEvent) => {
...
+  if (!virtualCard) {
+    logger.error(`Stripe: could not find virtual card ${stripeAuthorization.card.id}`, stripeEvent);
+    return;
+  }
+
+  if (stripeEvent) {
+    await checkStripeEvent(virtualCard.host, stripeEvent);
+  }
+
+  if (stripeAuthorization.status === 'reversed') {
```
znarf, line 277: Could it be that the expense was already reversed and canceled on our side? We would then do this multiple time.

kewitz: I think it is worth adding a check here and updating only if necessary. Thanks!

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -247,6 +263,37 @@ export const processTransaction = async (stripeTransaction, stripeEvent) => {
...
+
+    if (!expense) {
+      logger.error(
+        `Stripe: could not find expense attached to reversed authorization ${stripeAuthorization.id}`,
+        stripeEvent,
+      );
+      return;
+    }
+
+    await expense.update({ status: ExpenseStatus.CANCELED });
```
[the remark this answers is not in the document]
kewitz: It is. We're not returning anything outside an HTTP 200 status code in the webhook controller.

znarf (approved):
Cool, let's have a try. Let's be vigilant and check it's working as expected once in production.

PR #7426, opened 2022-04-20
Archiving collective also archives its children projects and events fixes 

ralexrdz:
Functionality and test.

First, hopefully, of many contributions

^_^ 

https://i.imgur.com/knmUOIc.png

[fixes this issue](https://github.com/opencollective/opencollective/issues/4774)


server/graphql/v1/mutations/collectives.js
```diff
@@ -463,7 +464,19 @@ export async function archiveCollective(_, args, req) {
   }
 
   // TODO: cascade deactivation to EVENTs and PROJECTs?
+  const subProjects = await collective.getProjects();
```
znarf, line 467: We generally don't use "sub", events and projects are always under a collective.

```suggestion
  const projects = await collective.getProjects();
```

server/graphql/v1/mutations/collectives.js
```diff
@@ -463,7 +464,19 @@ export async function archiveCollective(_, args, req) {
...
+  const subEvents = await collective.getEvents();
```
znarf, line 468:
```suggestion
  const events = await collective.getEvents();
```

server/graphql/v1/mutations/collectives.js
```diff
@@ -463,7 +464,19 @@ export async function archiveCollective(_, args, req) {
   }
 
   // TODO: cascade deactivation to EVENTs and PROJECTs?
```
znarf, line 466: Remove todo?

test/server/graphql/v1/mutation.test.js
```diff
@@ -96,6 +100,15 @@ describe('server/graphql/v1/mutation', () => {
...
   });
 
+  beforeEach('create a project collective', async () => {
+    await models.Collective.create(
+      Object.assign(utils.data('project1'), {
+        CreatedByUserId: user1.id,
+        ParentCollectiveId: collective1.id,
+      }),
+    );
+  });
```
Betree, line 110: We have a simpler way to create test entities nowadays that don't require adding new entries to `test/mocks/data.js`:

```suggestion
  beforeEach('create a project  under collective1', async () => {
    await fakeProject({ ParentCollectiveId: collective1.id });
  });
```

test/server/graphql/v1/mutation.test.js
```diff
@@ -343,6 +356,54 @@ describe('server/graphql/v1/mutation', () => {
...
+        expect(result.errors[0].message).to.equal('You need to be logged in as an Admin.');
+      });
+
+      it('should archive its children projects and events', async () => {
+        await utils.graphqlQuery(archiveCollectiveMutation, { id: collective1.id }, user1);
+
+        const projects = (await collective1.getProjects()).map(p => ({ id: p.id, name: p.name, isActive: p.isActive }));
+        const events = (await collective1.getEvents()).map(p => ({ id: p.id, name: p.name, isActive: p.isActive }));
+
+        expect(projects).to.all.have.property('isActive', false);
```
Betree, line 399: I would make sure these arrays are not empty, in case someones edits the code used to seed the data

```suggestion
        expect(projects.length).to.eq(1);
        expect(events.length).to.eq(1);
        expect(projects).to.all.have.property('isActive', false);
```

test/server/graphql/v1/mutation.test.js
```diff
@@ -13,6 +14,9 @@ import * as utils from '../../../utils';
 let host, user1, user2, user3, collective1, event1, ticket1;
 let sandbox, executeOrderStub, emailSendSpy, emailSendMessageSpy;
 
+chai.use(chaiThings);
```
Betree, line 17: chai plugins are added globally in `test/setup.js` to make them available in all tests.

package.json
```diff
@@ -51,6 +51,7 @@
     "bcrypt": "5.0.1",
     "bluebird": "3.7.2",
     "burner-email-providers": "1.0.67",
+    "chai-things": "^0.2.0",
```
znarf, line 54: Why is this dependency needed?

znarf (changes requested):
Looks generally good! A couple of things to change or fix.

znarf:
According to the test suite, the update to `package-lock.json` was not committed.

test/server/graphql/v1/mutation.test.js
ralexrdz: I realize that I should have added package-lock to the commit. My bad

package.json
ralexrdz: I used it to test for the propery isActive on an array of projects and events

ralexrdz:
I'm adding package-lock

ralexrdz:
just commited package-lock.json. Let me know if anything else is needed

test/server/graphql/v1/mutation.test.js
Betree: Yep, but this comment still stands :wink:

package.json
znarf: Doesn't seem needed as a main dependency in the project if it's only used for tests.

znarf:
> just commited package-lock.json. Let me know if anything else is needed

Thank you. We gave some other feedback, would be great if you could have a look at at that too.

test/server/graphql/v1/mutation.test.js
ralexrdz: just fixed it

server/graphql/v1/mutations/collectives.js
ralexrdz: removed

package.json
ralexrdz: moved to devDependencies

ralexrdz (commented):
Just finished the comments and rebase from original repo

server/graphql/v1/mutations/collectives.js
```diff
@@ -462,8 +463,19 @@ export async function archiveCollective(_, args, req) {
     return updatedCollective;
   }
 
-  // TODO: cascade deactivation to EVENTs and PROJECTs?
+  const projects = await collective.getProjects();
+  const events = await collective.getEvents();
 
+  if (projects && projects.length) {
+    projects.forEach(s => {
+      s.update({ isActive: false, deactivatedAt: Date.now(), approvedAt: null, HostCollectiveId: null });
```
Betree, line 471: Looks like something we would want to `await` for, otherwise errors while archiving the children will be ignored

server/graphql/v1/mutations/collectives.js
```diff
@@ -462,8 +463,19 @@ export async function archiveCollective(_, args, req) {
...
+    });
+  }
+  if (events && events.length) {
+    events.forEach(s => {
+      s.update({ isActive: false, deactivatedAt: Date.now(), approvedAt: null, HostCollectiveId: null });
```
znarf, line 476: We usually avoid short variables, just make it clear what it is.

```suggestion
    events.forEach(event => {
      event.update({ isActive: false, deactivatedAt: Date.now(), approvedAt: null, HostCollectiveId: null });
```

server/graphql/v1/mutations/collectives.js
znarf, line 471:
```suggestion
    projects.forEach(project => {
      project.update({ isActive: false, deactivatedAt: Date.now(), approvedAt: null, HostCollectiveId: null });
```

ralexrdz: you are right. Just added it

server/graphql/v1/mutations/collectives.js
ralexrdz: understoop and agree. Just fixed it

server/graphql/v1/mutations/collectives.js
ralexrdz: understoop and agree. Just fixed it

server/graphql/v1/mutations/collectives.js
```diff
@@ -462,8 +463,19 @@ export async function archiveCollective(_, args, req) {
...
   }
 
-  // TODO: cascade deactivation to EVENTs and PROJECTs?
+  const projects = await collective.getProjects();
+  const events = await collective.getEvents();
 
+  if (projects && projects.length) {
+    projects.forEach(async project => {
+      await project.update({ isActive: false, deactivatedAt: Date.now(), approvedAt: null, HostCollectiveId: null });
+    });
```
Betree, line 472: We're still not awaiting here. Try this:

```es6
const testPromise = new Promise(resolve => setTimeout(() => resolve(console.log('Done!')), 5000));
[42].forEach(async () => await testPromise());
console.log('Called promises');
```

You'll notice that "Called promises" will be displayed before "Done!".

We need to use `Promise.all` or a similar mechanism.

server/lib/queries.js
```diff
@@ -1153,6 +1154,49 @@ const getTransactionsTimeSeries = async (
   );
 };
 
+/**
+ * Returns tags along with their frequency of use.
+ */
+const getTagFrequencies = async args => {
```
Betree, line 1160: I don't understand why we have this change here, merging issue?

Betree (changes requested):

server/graphql/v1/mutations/collectives.js
ralexrdz: You are right. Forgot that I should be using PromiseAll or new ES6 for...of instead. Commiting...

server/lib/queries.js
ralexrdz: cause the rebase I did from main

ralexrdz:
Just fixed the async fors. 

The other changes you were finding strange appear when I was doing git rebase upstream/main

@Betree Maybe I'm not doing the best aproach for new commits.

I have these remotes:
- origin as ralexrdz/opencollective-api fork
- upstream as opencollective/opencollective-api
- 
This is what I do when adding changes from your comments;

- git fetch --all () 
- git rebase upstream/main
- Fixing conflicts
- git pull from origin main
- fixing conflcts 
- pushing to origin main

Any recommendation?

Betree:
@ralexrdz The commits history of this PR indeed doesn't look normal. You can get some recommendations on best practices from https://github.com/opencollective/opencollective-api/blob/main/CONTRIBUTING.md.

The current state being complicated to review, I'd recommend having a fresh start:
- Either reset to `main` then cherry-pick the changes you want to keep
- Or rebase your current PR on `main` in interactive mode and drop all the unrelated commits

ralexrdz:
> @ralexrdz The commits history of this PR indeed doesn't look normal. You can get some recommendations on best practices from https://github.com/opencollective/opencollective-api/blob/main/CONTRIBUTING.md.
> 
> The current state being complicated to review, I'd recommend having a fresh start:
> 
>     * Either reset to `main` then cherry-pick the changes you want to keep

I did the reseting option. Let me know if everything is ok.

Betree:
@ralexrdz The history looks much better, thanks! CI however seems to indicate a problem with the `package-lock.json`

image

ralexrdz:
I'll try a new npm install to create a package.json

Betree (changes requested):
We're not removing the `HOST` membership for children's projects and events. We're not checking their balances either ([like we do here](https://github.com/opencollective/opencollective-api/blob/c1ceac167a794b510cecc5f3debed8a38f1e652b/server/graphql/v1/mutations/collectives.js#L436)).

Rather than calling `project|event`.`update`, I think we should have a recurring function that makes the same check on all accounts regardless of their types.

Edit: we should make sure we don't double-send [the archived email](https://github.com/opencollective/opencollective-api/blob/main/templates/emails//archived.collective.hbs) while implementing this

Betree:
Alternative: we're introducing a `getConsolidatedBalance` that could solve the abovementioned issue in a simpler way

PR #7429, opened 2022-04-20
feat(MemberInvitation.js): resend invitation email

alurban:
When an invitation already exists, resend the invitation email on request from the frontend. This is done to enable a user-facing "Resend Invite" button, see opencollective/opencollective-frontend#7710.

Fixes opencollective/opencollective#5030

### DIFF
--- a/server/models/MemberInvitation.js
+++ b/server/models/MemberInvitation.js
@@ -9,35 +9,6 @@ import sequelize, { DataTypes } from '../lib/sequelize';
 
 export const MEMBER_INVITATION_SUPPORTED_ROLES = [roles.ACCOUNTANT, roles.ADMIN, roles.MEMBER];
 
-const sendInvitationEmail = async (invitation, memberParams, models, sequelizeParams, collective, skipDefaultAdmin) => {
-  // Load users
-  const memberUser = await models.User.findOne({
-    where: { CollectiveId: memberParams.MemberCollectiveId },
-    include: [{ model: models.Collective, as: 'collective' }],
-    ...sequelizeParams,
-  });
-
-  if (!memberUser) {
-    throw new Error('user not found');
-  }
-
-  // Determine collective creator
-  const createdByUser = await models.User.findByPk(memberParams.CreatedByUserId, {
-    include: [{ model: models.Collective, as: 'collective' }],
-    ...sequelizeParams,
-  });
-
-  // Send member invitation
-  await emailLib.send('member.invitation', memberUser.email, {
-    role: MemberRoleLabels[memberParams.role] || memberParams.role.toLowerCase(),
-    invitation: pick(invitation, 'id'),
-    collective: pick(collective, ['slug', 'name']),
-    memberCollective: pick(memberUser.collective, ['slug', 'name']),
-    invitedByUser: pick(createdByUser, ['collective.slug', 'collective.name']),
-    skipDefaultAdmin: skipDefaultAdmin || false,
-  });
-};
-
 function defineModel() {
   const { models } = sequelize;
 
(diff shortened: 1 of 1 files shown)


server/models/MemberInvitation.js
```diff
@@ -9,6 +9,35 @@ import sequelize, { DataTypes } from '../lib/sequelize';
 
 export const MEMBER_INVITATION_SUPPORTED_ROLES = [roles.ACCOUNTANT, roles.ADMIN, roles.MEMBER];
 
+const sendInvitationEmail = async (invitation, memberParams, models, sequelizeParams, collective, skipDefaultAdmin) => {
```
Betree, line 12: Can we have this as a class method of `MemberInvitation`? Calling `invitation.sendEmail(...)` sounds right, it will also simplify the function and make it consistent with the rest of the business logic that mostly lives there (e.g. `accept`, `decline`)

Betree (commented):
Looking good! I think we can slightly improve the way the code is structured, but overall this seems like the right approach.

server/models/MemberInvitation.js
alurban: @Betree that's an excellent idea and I've implemented it in the latest commit. One caveat: `invitation.sendEmail` leads to an error, presumably because it's a circular reference (`sendEmail` hasn't been defined before exporting the class). So in my implementation I use `this.sendEmail`, which passes the new end-to-end test I've added in opencollective/opencollective-frontend#7710.

I've convinced myself that this is functionally equivalent since all the same parameters are passed, but of course I'd appreciate a reviewer's feedback.

server/models/MemberInvitation.js
```diff
@@ -151,6 +151,35 @@ function defineModel() {
 
   // ---- Static methods ----
 
+  MemberInvitation.sendEmail = async (memberParams, sequelizeParams, collective, skipDefaultAdmin) => {
```
Betree, line 154: Since we're using `this` in this method (which is the right thing to do), then it's not a static method. It should be above `// ---- Static methods ----`. I'm also surprised that it works with `MemberInvitation.sendEmail =`, it should be `MemberInvitation.prototype.sendEmail =`

alurban: Interestingly it doesn't work as an instance method with `this`, but it *does* work with your original spec:

```diff
diff --git a/server/models/MemberInvitation.js b/server/models/MemberInvitation.js
index b8744e2c7..82219edee 100644
--- a/server/models/MemberInvitation.js
+++ b/server/models/MemberInvitation.js
@@ -149,9 +149,7 @@ function defineModel() {
     return this.destroy();
   };
 
-  // ---- Static methods ----
-
-  MemberInvitation.sendEmail = async (memberParams, sequelizeParams, collective, skipDefaultAdmin) => {
+  MemberInvitation.prototype.sendEmail = async function (memberParams, sequelizeParams, collective, skipDefaultAdmin) {
     // Load users
     const memberUser = await models.User.findOne({
       where: { CollectiveId: memberParams.MemberCollectiveId },
@@ -180,6 +178,8 @@ function defineModel() {
     });
   };
 
+  // ---- Static methods ----
+
   MemberInvitation.invite = async function (collective, memberParams, { transaction, skipDefaultAdmin } = {}) {
     const sequelizeParams = transaction ? { transaction } : undefined;
 
@@ -214,7 +214,7 @@ function defineModel() {
     });
 
     if (existingInvitation) {
-      await this.sendEmail(memberParams, sequelizeParams, collective, skipDefaultAdmin);
+      await existingInvitation.sendEmail(memberParams, sequelizeParams, collective, skipDefaultAdmin);
       return existingInvitation.update(pick(memberParams, ['role', 'description', 'since']), sequelizeParams);
     }
 
@@ -235,7 +235,7 @@ function defineModel() {
       sequelizeParams,
     );
 
-    await this.sendEmail(memberParams, sequelizeParams, collective, skipDefaultAdmin);
+    await invitation.sendEmail(memberParams, sequelizeParams, collective, skipDefaultAdmin);
 
     return invitation;
   };
```

Do you have a preference?

Betree (commented):
The code looks good to me, I'll test & merge if it works!

alurban:
> The code looks good to me, I'll test & merge if it works!

sounds great, thanks @Betree!

alurban:
@Betree have you had a chance to test?

server/models/MemberInvitation.js
Betree: I've updated the code to rely on `this` rather than `memberParams`

Betree (approved):
Updated the code for `sendEmail` to make sure it takes most its data from `this` since it's a class method (i.e. using `this.MemberCollectiveId` rather than `memberParams.MemberCollectiveId`)

PR #7432, opened 2022-04-21
Add CI job to track performance evolution (Artillery)

Betree:


Betree:
I'm not satisfied with what I implemented here nor in https://github.com/opencollective/opencollective-api/pull/7452.

My conclusion after trying both of these tools:
- They both make it difficult to have multiple scenarios/test suites, kind of like we have multiple files for cypress. The solution that I'd like to see should ideally support out-of-the-box defining multiple files (test/perf/search.js, test/perf/contributors.js, etc) and run them in one go, generating a single report to summarize everything.
- They both make it difficult to display a bundled result (related to the point above) in a Github PR

Still think the need exists, but I hope we'll be able to find a better solution in the future!

PR #7433, opened 2022-04-21
Include per-kind Host Metrics transactions timeseries

Betree:
Part of https://github.com/opencollective/opencollective/issues/5372
Requirement for https://github.com/opencollective/opencollective-frontend/pull/7715

### DIFF
--- a/server/graphql/v2/object/HostMetricsTimeSeries.ts
+++ b/server/graphql/v2/object/HostMetricsTimeSeries.ts
@@ -1,9 +1,13 @@
 import { GraphQLNonNull, GraphQLObjectType } from 'graphql';
 
+import { TransactionKind } from '../../../constants/transaction-kind';
+import { TransactionTypes } from '../../../constants/transactions';
 import * as HostMetricsLib from '../../../lib/host-metrics';
 import { fetchAccountsWithReferences } from '../input/AccountReferenceInput';
+import { getTimeSeriesFields } from '../interface/TimeSeries';
 
 import { TimeSeriesAmount } from './TimeSeriesAmount';
+import { TimeSeriesAmountWithKind } from './TimeSeriesAmountWithKind';
 import { TimeSeriesAmountWithSettlement } from './TimeSeriesAmountWithSettlement';
 
 export const resultsToAmountNode = results => {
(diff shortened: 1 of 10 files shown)


server/graphql/v2/object/HostMetricsTimeSeries.ts
```diff
@@ -58,13 +71,69 @@ export const HostMetricsTimeSeries = new GraphQLObjectType({
...
+      resolve: async ({ host, account, dateFrom, dateTo, timeUnit }) => {
+        let collectiveIds;
+        if (account) {
+          const collectives = await fetchAccountsWithReferences(account, { attributes: ['id'] });
+          collectiveIds = collectives.map(collective => collective.id);
+        }
+
+        const amountDataPoints = await HostMetricsLib.getTransactionsTimeSeries(
+          TransactionKind.EXPENSE,
+          TransactionTypes.DEBIT,
```
SudharakaP, line 122: Except for these two lines, everything else is the same within the `resolve` function in `totalReceived` (starting line 86). Maybe we can put them into a common function to avoid the duplication? 🤔

SudharakaP (approved):

server/graphql/v2/object/HostMetricsTimeSeries.ts
Betree: It's a copy-paste of the resolvers introduced with `expenseAmountOverTime` + `contributionAmountOverTime`, which had pretty good factorization already (`fetchAccountsWithReferences`, `resultsToAmountWithKindNode`, `getTransactionsTimeSeries` are great helpers). This code takes a lot of lines, but mainly because of formatting.

I however agree that some parts could be factorized more, so I've added a small [change](https://github.com/opencollective/opencollective-api/pull/7433/commits/36f5375c7f8ed9adaa9f44dce16714134c43ecfe):
- Move `args.account` resolution up in `hostMetricsTimeSeries` resolver (since we need the same for all timeseries anyway)
- Stop trying to generate defaults for `dateFrom`/`dateTo`, the parent's again taking care of it
- Move all optional args for `getTotalMoneyManagedTimeSeries` to an object

PR #7440, opened 2022-04-22
Deprecate `supportedPaymentMethodService`

SudharakaP:
We need to deprecate `supportedPaymentMethodService` field first before removing it. 

Related https://github.com/opencollective/opencollective-api/pull/7411

### DIFF
--- a/server/graphql/v2/query/collection/AccountsQuery.ts
+++ b/server/graphql/v2/query/collection/AccountsQuery.ts
@@ -40,6 +40,7 @@ const AccountsQuery = {
     supportedPaymentMethodService: {
       type: new GraphQLList(PaymentMethodService),
       description: 'Only accounts that support one of these payment services will be returned',
+      deprecationReason: '2022-04-22: Introduced for Hacktoberfest and not used anymore.',
     },
     skipRecentAccounts: {
       type: GraphQLBoolean,


server/graphql/v2/query/collection/AccountsQuery.ts
```diff
@@ -39,7 +39,7 @@ const AccountsQuery = {
     },
     supportedPaymentMethodService: {
       type: new GraphQLList(PaymentMethodService),
-      description: 'Only accounts that support one of these payment services will be returned',
+      description: '@deprecated: Introduced for Hacktoberfest and not used anymore.',
```
znarf, line 42: What is this syntax for deprecation? I don't see it used anywhere.

znarf (changes requested):
Change requested.

server/graphql/v2/query/collection/AccountsQuery.ts
SudharakaP: Ops, sorry, a mistake on my part; we need to use `deprecationReason` here. 🤯

Betree (approved):

znarf (approved):

MattIPv4:
👋 Just wanted to note that Hacktoberfest still makes queries using this arg

Betree:
@MattIPv4 Thanks for letting us know, we'll make sure to keep this argument available. The error itself was introduced in a separate PR, we're tracking it in https://github.com/opencollective/opencollective/issues/5542. Sorry for the inconvenience!

@SudharakaP could you add a link to this comment in the `deprecationReason` so the next person knows what to look for before removing this `supportedPaymentMethodService` argument?

SudharakaP:
> wave Just wanted to note that Hacktoberfest still makes queries using this arg, and it seems that passing it into a query to the API now results in an error that is breaking Hacktoberfest's logic for pulling OpenCollective projects on the site. :(

@MattIPv4: Thanks for reporting this issue. 

@Betree : I've opened a PR for [opencollective/opencollective#](https://github.com/opencollective/opencollective/issues/5542) and for the change of comment at, https://github.com/opencollective/opencollective-api/pull/7513.

MattIPv4:
Thank you both! 💙

PR #7452, opened 2022-04-25
Add CI job to track performance evolution (K6)

Betree:
Alternative to https://github.com/opencollective/opencollective-api/pull/7432


Betree:
Though more powerful & popular, K6 is too painful to set up with multi tests. Best chance if we want to look back at this will be https://github.com/grafana/k6-template-es6. Closing to go back to artillery

PR #7455, opened 2022-04-25
Return Collectives based on Parent Country in `search` query 

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5447

Related Front-End PR: https://github.com/opencollective/opencollective-frontend/pull/7740

return-collectives-based-on-parent


SudharakaP:
@Betree : Just to let you know that I did an performance check of the modified query with `EXPLAIN ANALYSIS` and it seems there's no difference in the modified query on time or cost. 

Before: Limit  (cost=111.77..111.78 rows=1 width=1412) (actual time=0.215..0.216 rows=1 loops=1)
After: Limit  (cost=120.07..120.07 rows=1 width=1412) (actual time=0.363..0.365 rows=2 loops=1)

server/lib/search.js
```diff
@@ -197,13 +197,18 @@ export const searchCollectivesInDB = async (
...
-    WHERE "deletedAt" IS NULL
-    AND "deactivatedAt" IS NULL
-    AND ("data" ->> 'isGuest')::boolean IS NOT TRUE
-    AND ("data" ->> 'hideFromSearch')::boolean IS NOT TRUE
-    AND name != 'incognito'
-    AND name != 'anonymous'
-    AND "isIncognito" = FALSE ${dynamicConditions}
+    ${
+      countryCodes
+        ? 'FULL OUTER JOIN "Collectives" parentCollective ON c."ParentCollectiveId" = parentCollective.id'
```
Betree, line 202: Seems more suited

```suggestion
        ? 'LEFT JOIN "Collectives" parentCollective ON c."ParentCollectiveId" = parentCollective.id'
```

Betree (approved):

server/lib/search.js
SudharakaP: Thanks. I've corrected this. 👍🏽

Betree (commented):
As per the feedback in the frontend PR, let's not include the host's country

SudharakaP:
> As per the feedback in the frontend PR, let's not include the host's country

I've changed it to not include the host country. 👍🏽

server/lib/search.js
```diff
@@ -92,7 +92,7 @@ const sanitizeSearchTermForILike = term => {
   return term.replace(/(_|%|\\)/g, '\\$1');
 };
 
-const getSearchTermSQLConditions = term => {
+const getSearchTermSQLConditions = (term, collectiveTable) => {
```
SudharakaP, line 95: @Betree : After the rebase I've added a new arg to this function as when we use it within the `searchCollectivesInDB` function we need to refer to to the collective table.

server/lib/search.js
```diff
@@ -151,42 +151,42 @@ export const searchCollectivesInDB = async (
...
+      dynamicConditions += `AND (c."settings"->>'disableCustomContributions')::boolean IS NOT TRUE `;
     } else {
-      dynamicConditions += `AND ("settings"->>'disableCustomContributions')::boolean IS TRUE `;
+      dynamicConditions += `AND (c."settings"->>'disableCustomContributions')::boolean IS TRUE `;
     }
   }
 
   if (countryCodes) {
-    dynamicConditions += `AND "countryISO" IN (:countryCodes) `;
+    dynamicConditions += `AND (c."countryISO" IN (:countryCodes) OR parentCollective."countryISO" `;
```
Betree, line 182: I don't get how this condition works (`OR parentCollective."countryISO"`? looks like we're missing something) plus it looks like there will be a syntax error for missing `)`.

It's super simple to add tests for this function in `test/server/lib/search.test.js`, let's make it right.

Betree (changes requested):

server/lib/search.js
SudharakaP: Ops, I've mistakenly deleted part of the string. I've corrected it and added some tests as well. Thanks.

SudharakaP: Also I've created a new issue; https://github.com/opencollective/opencollective/issues/5530 to keep track of adding tests. I think we could add more tests around tags and type filters but we'll do it separately as this issue is only about country filter.

test/server/lib/search.test.js
```diff
@@ -127,6 +127,31 @@ describe('server/lib/search', () => {
         expect(results2[0].id).to.eq(accountWithDiacritics.id);
       });
 
+      it('when searching with a country filter give correct results', async () => {
+        const collective1Name = 'JHipster Canada';
+        const collective2Name = 'JHipster Sri Lanka';
+        const collective1 = await fakeCollective({ name: collective1Name, countryISO: 'CA' });
+        const collective2 = await fakeCollective({ name: collective2Name, countryISO: 'LK' });
+        const [results1] = await searchCollectivesInDB('JHipster');
```
Betree, line 135: This one is not searching by country, what is it testing?

Betree (approved):
Haven't tested but the code looks correct

test/server/lib/search.test.js
SudharakaP: Ah, I added this to check if the search is working properly even with the countries being present in the collectives. But I think this is rather redundant as we test for the search terms in other tests so I've removed this part and only kept the part where it search by country. 👍🏽

PR #7457, opened 2022-04-26
Move `geoLocationLatLong` to JSONB

Betree:
As per https://github.com/opencollective/opencollective/issues/3171, this removes the dependency on Postgis.

Sequelize was converting all `GEOMETRY(POINT)` to objects like `{ type: 'POINT', coordinates: [0, 0] }` (basically what we get by calling `st_asgeojson`). Storing the JSONB directly seems to have the exact same effect regarding the model usage. Since no query depends on it, it seems that there's nothing more to do.

### DIFF
--- a/migrations/20220426103129-move-coordinates-to-jsonb.js
+++ b/migrations/20220426103129-move-coordinates-to-jsonb.js
@@ -1,0 +1,28 @@
+'use strict';
+
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+      ALTER TABLE "CollectiveHistories"
+      ALTER COLUMN "geoLocationLatLong" TYPE JSONB
+      USING JSONB("geoLocationLatLong")
+    `);
+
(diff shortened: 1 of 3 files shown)


znarf (changes requested):
Looks good! Unfortunately, the migration doesn't seems to work in CI.

Betree:
> Looks good! Unfortunately, the migration doesn't seems to work in CI.

Yep it came in with my last change which `DROP EXTENSION postgis`. I may keep that out of the PR for now

znarf (approved):
Excellent. Next step is to remove the dependency fully in documentation and setup instructions.

Maybe we will have to update our development data dump so it doesn't need postgis?

Betree:
> Excellent. Next step is to remove the dependency fully in documentation and setup instructions.
> 
> Maybe we will have to update our development data dump so it doesn't need postgis?

Exactly. I posted the next steps in https://github.com/opencollective/opencollective/issues/3171#issuecomment-1109726932

PR #7459, opened 2022-04-26
get reason when cancelling recurring contribution

macovedj:
API changes for [this](https://github.com/opencollective/opencollective/issues/1697)


macovedj:
@Betree was thinking about it and realized that the cypress tests would fail if they don't know to point at this branch, and became curious about what that workflow looks like.  I guess you all probably merge your api branches before you merge your front end branches?

Betree:
@macovedj Our CI supports matching branches: if two branches have the same name between the frontend & API, they'll be used together.

templates/emails/subscription.canceled.hbs
```diff
@@ -12,7 +12,10 @@ Subject: Contribution cancelled to {{{collective.name}}}
 
 <p>Your recurring contribution to {{collective.name}} (cced) for {{currency subscription.amount currency=subscription.currency}}/{{subscription.interval}} has been cancelled.</p>
 
-<p>We are sad to see you go 😞. If you have any feedback, we want to hear about it. Just hit 'reply all'.</p>
+{{#if reason}}
+  <p>You provided the following reason for cancelling:</p>
+  <p>{{reason}}</p>
+{{/if}}
```
Betree, line 18: I edited this to:
- Move reason up (it was previously appearing after the closing sentence - `Warmly - xxx`.
- Removed `If you have any feedback, we want to hear about it. Just hit 'reply all'`, this email now uses a no-reply

Betree (approved):

PR #7460, opened 2022-04-26
Fix 2FA Expense payment checker

kewitz:


server/controllers/users.js
```diff
@@ -89,10 +89,13 @@ export const signin = (req, res, next) => {
...
   if (twoFactorAuthenticationEnabled && req.remoteUser.twoFactorAuthToken !== null) {
-    const token = req.remoteUser.jwt({ scope: 'twofactorauth' }, auth.TOKEN_EXPIRATION_SESSION);
+    const token = req.remoteUser.jwt(
+      { scope: 'twofactorauth', sesionId: req.jwtPayload?.sessionId },
+      auth.TOKEN_EXPIRATION_SESSION,
+    );
     res.send({ token });
   } else {
-    const token = req.remoteUser.jwt({}, auth.TOKEN_EXPIRATION_SESSION);
+    const token = req.remoteUser.jwt({ sesionId: req.jwtPayload?.sessionId }, auth.TOKEN_EXPIRATION_SESSION);
```
znarf, line 98:
```suggestion
    const token = req.remoteUser.jwt({ sessionId: req.jwtPayload?.sessionId }, auth.TOKEN_EXPIRATION_SESSION);
```

Betree:
Related: https://github.com/opencollective/opencollective/issues/3781

znarf:
We use Expense payment a lot with the API these days (1kproject) and we also need a solution that works well with so called "API Keys" (Personal Tokens).

kewitz:
> We use Expense payment a lot with the API these days (1kproject) and we also need a solution that works well with so called "API Keys" (Personal Tokens).

@znarf I think requiring 2FA codes from an API key request is against the purpose of creating the API key in the first place.
I suggest we require the 2FA code to create an API key and ignore the 2FA key for actions when using an API key.

WDYT?

znarf:
> I think requiring 2FA codes from an API key request is against the purpose of creating the API key in the first place.

@kewitz I don't understand how you come to that conclusion. I disagree, I think it's a great security feature.

kewitz:
@znarf people usually create an API key to automate their actions. I don't think it is reasonable to expect someone to automate their actions and require them to always be around in order to type a 2FA code.

kewitz:
@znarf just confirmed this implementation is compatible with the 1K Project script.
Notice that since we're not generating tokens for API Key type of calls, there's also no session attributed to these calls.
Meaning that when you use 2FA code in a single API Key call, you'll still be enabling all calls from the same API Key to pay for expenses.

Do you think this is acceptable?

znarf:
> Do you think this is acceptable?

Sure, but I would make it explicit, and not just rely on `sessionId` being empty. https://github.com/opencollective/opencollective-api/pull/7460/files#diff-7c3df5f093276d2cc8207c916aa13f831d4e6bb7dbfc32bab603a64aefe93b03R34

kewitz:
Rebased.

znarf (approved):
That should works.

Secondary review from @Betree required.

Betree (approved):
Looks great and I like that it doesn't introduce too much complexity

kewitz:
@Betree, I was thinking about adding session control in this PR but I'll just create a new issue later on so we can elaborate on it.

Betree:
> @Betree, I was thinking about adding session control in this PR but I'll just create a new issue later on so we can elaborate on it.

@kewitz There's already [an issue](https://github.com/opencollective/opencollective/issues/3781) for that with some raw specs proposal

PR #7465, opened 2022-04-28
Update ExpenseItem amount when persisting Stripe charge

kewitz:
Bug caught by user and submitted through a support ticket:
image


znarf (approved):
Thank you for this!

Do we have some entries in the database that needs to be fixed?

kewitz:
> Thank you for this!
> 
> Do we have some entries in the database that needs to be fixed?

Probably. It is on my to-do list to merge this and fix what's pending. Doing it ASAP.

PR #7469, opened 2022-04-29
Consolidated balance for Collectives, including children balance

znarf:
Fix https://github.com/opencollective/opencollective/issues/5500

### DIFF
--- a/server/lib/budget.js
+++ b/server/lib/budget.js
@@ -40,21 +40,30 @@ export async function getBalanceAmount(collective, { startDate, endDate, currenc
   });
 }
 
-export async function getConsolidatedBalance(collective) {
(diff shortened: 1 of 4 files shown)


server/lib/budget.js
```diff
@@ -40,6 +40,32 @@ export async function getBalanceAmount(collective, { startDate, endDate, currenc
   });
 }
 
+export async function getConsolidatedBalanceAmount(collective, { currency, version }) {
+  version = version || collective.settings?.budget?.version || 'v1';
+  currency = currency || collective.currency;
+
+  const children = await collective.getChildren();
```
Betree, line 47: Cheap optimization

```suggestion
  const children = await collective.getChildren({ attributes: ['id'] });
```

Betree (approved):
Looks good

PR #7477, opened 2022-05-02
Search: use websearch_to_tsquery

Betree:
Implements https://opencollective.slack.com/archives/C035S573ZD2/p1651291317620119

Changes:
- Refactor to use a common helper to build conditions between search and `getTagFrequencies`
- No change for single-term search (e.g. `babel`)
- Move the default search function for multi-term to `websearch_to_tsquery`, which implements a more standard behavior:
  - Removes partial search on multi-terms search. Example: `babel apex` won't search for `babel apex:*` anymore. This was  [perceived](https://github.com/opencollective/opencollective-api/pull/7403#pullrequestreview-939167493) as problematic anyway.
  - Switches the default strategy for multi-terms to "AND": previously searching for `babel apex` would return a query like `babel|apex`. It will now return `babel & apex`
  - Adds support for common operators:
     - `babel AND apex`
     - `babel OR apex`
     - `"exact search"`

### DIFF
--- a/server/lib/search.js
+++ b/server/lib/search.js
@@ -60,20 +60,6 @@ export const searchCollectivesByEmail = async (email, user, offset = 0, limit = 
 };
 
 /**
- * Sanitize and then turn a search string into a TS vector using 'OR' operator.
- *
- * Examples: "open potatoes" => "open|potatoes", "crème brulée => "creme|brulee"
- *
- */
-const searchTermToTsVector = term => {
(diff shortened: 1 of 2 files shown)


SudharakaP (approved):
Thanks, this seems to work well. I've tested it. Only comment I have is that I think we should add this to the documentation; the types of searches we could do such as `OR`, `AND`, "exact match".

PR #7492, opened 2022-05-04
paymentMethod in Transactions GQL query

ralexrdz:
Part of https://github.com/opencollective/opencollective/issues/5517


server/graphql/v2/query/collection/TransactionsCollectionQuery.ts
```diff
@@ -275,6 +280,21 @@ export const TransactionsCollectionResolver = async (args, req: express.Request)
...
+
+    if (args.paymentMethodType.includes(null)) {
+      paymentMethodConditions.push({ PaymentMethodId: null });
+    }
+
+    if (paymentMethodConditions.length) {
+      include.push({ model: models.PaymentMethod });
+      where.push({ [Op.or]: paymentMethodConditions });
+    }
+  }
```
Betree, line 297: Maybe I should add a comment here for clarity. Basically, we'll allowing to filter by `null` to get transactions without payment methods

PR #7494, opened 2022-05-05
Granular freeze

macovedj:
Initial attempt at [this](https://github.com/opencollective/opencollective/issues/5383#issuecomment-1116951033)

### DIFF
--- a/migrations/20220505003315-is-frozen.js
+++ b/migrations/20220505003315-is-frozen.js
@@ -1,0 +1,45 @@
+'use strict';
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    /**
+     * Add altering commands here.
+     *
+     * Example:
+     * await queryInterface.createTable('users', { id: Sequelize.INTEGER });
+     */
+    queryInterface
+      .addColumn(
+        'Collectives', // table name
+        'isFrozen', // new field name
+        {
+          type: Sequelize.BOOLEAN,
+          allowNull: true,
+        },
+      )
+      .then(() =>
+        queryInterface.addColumn('CollectiveHistories', 'isFrozen', {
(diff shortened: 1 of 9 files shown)


migrations/20220505003315-is-frozen.js
```diff
@@ -0,0 +1,45 @@
...
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    /**
+     * Add altering commands here.
+     *
+     * Example:
+     * await queryInterface.createTable('users', { id: Sequelize.INTEGER });
+     */
+    queryInterface
```
Betree, line 11: Need to return the promise for sequelize to properly await. We tend to prefer using `async`/`await`

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -165,7 +165,7 @@ const accountMutations = {
       },
       action: {
         type: new GraphQLNonNull(
-          new GraphQLEnumType({ name: 'AccountFreezeAction', values: { FREEZE: {}, UNFREEZE: {} } }),
+          new GraphQLEnumType({ name: 'AccountFreezeAction', values: { FREEZE: {}, UNFREEZE: {}, FREEZEKEEPCONTRIBUTIONS: {} } }),
```
Betree, line 168: This `FREEZEKEEPCONTRIBUTIONS` is a bit aggressive and won't play well with the rest of the issue that we'll later want to implement: being able to select the features you want to freeze, We should keep FREEZE/UNFREEZE and add a new param for this

Betree (changes requested):

Betree:
Closing as stale. We made some progres on this topic after https://github.com/opencollective/opencollective/issues/3593

PR #7497, opened 2022-05-05
Move activity definition from transactions sum to count

Betree:
As proposed in https://github.com/opencollective/opencollective/issues/5329#issuecomment-1102367434


znarf (approved):
Let's have a try.

PR #7502, opened 2022-05-05
Incomplete Expense Status

kewitz:
Prototype for https://github.com/opencollective/opencollective/issues/5251.

### DIFF
--- a/server/graphql/v2/mutation/ExpenseMutations.ts
+++ b/server/graphql/v2/mutation/ExpenseMutations.ts
@@ -18,6 +18,7 @@ import {
   canVerifyDraftExpense,
   createExpense,
   editExpense,
+  markExpenseAsIncomplete,
   markExpenseAsSpam,
   markExpenseAsUnpaid,
   payExpense,
@@ -291,6 +292,8 @@ const expenseMutations = {
           return approveExpense(req, expense);
         case 'UNAPPROVE':
           return unapproveExpense(req, expense);
+        case 'MARK_AS_INCOMPLETE':
+          return markExpenseAsIncomplete(req, expense);
         case 'REJECT':
           return rejectExpense(req, expense);
         case 'MARK_AS_SPAM':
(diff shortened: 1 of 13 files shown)


templates/emails/collective.expense.incomplete.hbs
```diff
@@ -0,0 +1,23 @@
+Subject: Your expense to {{{collective.name}}} for {{currency expense.amount currency=expense.currency}} requires attention
+
+
+{{> header}}
+
+<center>
+  <p>Your expense, <a href="{{config.host.website}}/{{collective.slug}}/expenses/{{expense.id}}">{{expense.description}}</a>, submitted to {{{collective.name}}} was flagged as incomplete and requires your attention:</p>
+  {{#if message}}
+  <p>{{breaklines message}}</p>
```
Betree, line 9: Sounds simpler, safer (for the sanitization) and more consistent with what we do eslsewhere to do:

```suggestion
  <p style="white-space: pre-line;">{{message}}</p>
```

Betree (approved):
The code architecture looks great & simple. This is a good foundation for the future!

templates/emails/collective.expense.incomplete.hbs
kewitz: That doesn't seem to work everywhere. It doesn't work on maildev, for example.

Betree: @kewitz Maildev had some breaking changes in their latest versions, we rolled that back. Are the other emails not working as well? I'm surprised because we use this `white-space: pre-line` technique in many places

server/graphql/v2/mutation/ExpenseMutations.ts
```diff
@@ -291,6 +296,8 @@ const expenseMutations = {
           return approveExpense(req, expense);
         case 'UNAPPROVE':
           return unapproveExpense(req, expense);
+        case 'MARK_AS_INCOMPLETE':
+          return markExpenseAsIncomplete(req, expense, args.paymentParams?.message);
```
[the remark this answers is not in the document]
Betree: Arguably, `shouldRefundPaymentProcessorFee` is part of payment info: you're providing details about how the refund 
payment is supposed to be processed. It's more difficult to justify why `message` ended up there. 

Renaming `paymentParams` could make sense, but that would be a breaking change - not worth the effort IMO.

templates/emails/collective.expense.incomplete.hbs
Betree: Following up in https://github.com/opencollective/opencollective-api/pull/7531

PR #7508, opened 2022-05-09
Application support in GraphQL v2

znarf:


server/graphql/v2/input/ApplicationReferenceInput.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+export const fetchApplicationWithReference = async input => {
+  let application;
+  if (input.id) {
+    const id = idDecode(input.id, IDENTIFIER_TYPES.APPLICATION);
+    application = await models.Application.findByPk(id);
+  } else if (input.legacyId) {
+    application = await models.Application.findByPk(input.legacyId);
+  } else if (input.clientId) {
+    application = await models.Application.findOne('clientId', input.clientId);
```
znarf, line 38:
```suggestion
    application = await models.Application.findOne({where: { 'clientId': input.clientId });
```

znarf (commented):
Missing update.

server/graphql/v2/mutation/ApplicationMutations.js
```diff
@@ -0,0 +1,62 @@
...
+    },
+  },
+  async resolve(_, args, req) {
+    if (!req.remoteUser) {
+      throw new Unauthorized('You need to be authenticated to delete an application.');
+    }
+
+    const app = await fetchApplicationWithReference(args.application);
+    if (!app) {
+      throw new NotFound(`Application with id ${args.id} not found`);
```
znarf, line 55: Wrong error message if args != id

server/graphql/v2/query/ApplicationQuery.js
```diff
@@ -0,0 +1,19 @@
...
+
+import { ApplicationReferenceInput, fetchApplicationWithReference } from '../input/ApplicationReferenceInput';
+import { Application } from '../object/Application';
+
+const ApplicationQuery = {
+  type: Application,
+  args: {
+    application: {
+      type: new GraphQLNonNull(ApplicationReferenceInput),
+      description: 'Identifiers to retrieve the Order',
```
Betree, line 11:
```suggestion
      description: 'Identifiers to retrieve the Application',
```

server/graphql/v2/input/ApplicationReferenceInput.js
Betree, line 38: Syntax?

server/graphql/v2/mutation/index.js
```diff
@@ -1,5 +1,6 @@
 import accountMutations from './AccountMutations';
 import { addFundsMutation } from './AddFundsMutations';
+import applicationMutations from './ApplicationMutations';
```
Betree, line 3: We're doing `export const` in there so we probably need

```suggestion
import * as applicationMutations from './ApplicationMutations';
```

Betree: But we'll need to rename the exports if we do that

server/graphql/v2/input/ApplicationCreateInput.js
```diff
@@ -0,0 +1,13 @@
...
+
+import URL from '../scalar/URL';
+
+export const ApplicationCreateInput = new GraphQLInputObjectType({
+  name: 'ApplicationCreateInput',
+  description: 'Input type for Application',
+  fields: () => ({
+    name: { type: GraphQLString },
+    description: { type: GraphQLString },
+    callbackUrl: { type: URL },
```
Betree, line 11: Should we add the type here?

znarf: What type?

server/graphql/v2/mutation/ApplicationMutations.js
```diff
@@ -0,0 +1,87 @@
...
+export const updateApplicationMutation = {
+  type: Application,
+  args: {
+    application: {
+      type: new GraphQLNonNull(ApplicationUpdateInput),
+    },
+  },
+  async resolve(_, args, req) {
+    if (!req.remoteUser) {
+      throw new Unauthorized('You need to be authenticated to delete an application.');
```
Betree, line 52:
```suggestion
      throw new Unauthorized('You need to be authenticated to update an application.');
```

server/graphql/v2/mutation/index.js
znarf: I moved back to the approach we had in other files.

server/graphql/v2/input/ApplicationCreateInput.js
Betree: Application type (oauth)

Betree: Even if it has only one value for now, to make it explicit that we're creating an oauth app, not an API key

PR #7511, opened 2022-05-10
Track core member activities (Invite, Create, Remove)

kewitz:


server/graphql/v2/mutation/MemberInvitationMutations.js
```diff
@@ -148,6 +149,21 @@ const memberInvitationMutations = {
 
       if (args.accept) {
         await invitation.accept();
+        if ([MemberRoles.ACCOUNTANT, MemberRoles.ADMIN, MemberRoles.MEMBER].includes(invitation.role)) {
```
Betree, line 152: I think we can put that in `MemberInvitation.accept`, IMO this is part of the business logic and we'll want to record the activity regardless of where we're accepting the invitation from.

server/models/Activity.js
```diff
@@ -51,7 +51,9 @@ function defineModel() {
 
       hooks: {
         afterCreate(activity) {
-          notify(activity); // intentionally no return statement, needs to be async
+          if (activity.data?.notify !== false) {
+            notify(activity); // intentionally no return statement, needs to be async
+          }
```
Betree, line 56: Great addition

Betree (commented):
Great implementation. We should also make sure we cover the member invitations

kewitz:
Rebased.

server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -161,6 +162,20 @@ const memberMutations = {
         throw new ValidationFailed(`Member ${memberAccount.slug} does not exist in Collective ${account.slug}`);
       }
 
+      if ([MemberRoles.ACCOUNTANT, MemberRoles.ADMIN, MemberRoles.MEMBER].includes(this.role)) {
```
Betree, line 165: It seems that we've broken this condition in the latest commits (needs to use `args`, not `this` here). I'm pushing a fix in https://github.com/opencollective/opencollective-api/pull/7598.

PR #7513, opened 2022-05-10
Change deprecation reason for `supportedPaymentMethodService` field

SudharakaP:
Related to https://github.com/opencollective/opencollective-api/pull/7440#issuecomment-1121972791

### DIFF
--- a/server/graphql/v2/query/collection/AccountsQuery.ts
+++ b/server/graphql/v2/query/collection/AccountsQuery.ts
@@ -40,7 +40,8 @@ const AccountsQuery = {
     supportedPaymentMethodService: {
       type: new GraphQLList(PaymentMethodService),
       description: 'Only accounts that support one of these payment services will be returned',
-      deprecationReason: '2022-04-22: Introduced for Hacktoberfest and not used anymore.',
+      deprecationReason:
+        '2022-04-22: Introduced for Hacktoberfest. Reference: https://github.com/opencollective/opencollective-api/pull/7440#issuecomment-1121504508',
     },
     skipRecentAccounts: {
       type: GraphQLBoolean,


server/graphql/v2/query/collection/AccountsQuery.ts
```diff
@@ -52,8 +53,9 @@ const AccountsQuery = {
       description: 'Limit the search to collectives belonging to these countries',
     },
     orderBy: {
-      type: OrderByInput,
+      type: new GraphQLNonNull(OrderByInput),
       description: 'The order of results',
+      defaultValue: { field: 'ACTIVITY', direction: 'DESC' },
```
Betree, line 58: This would change the default sort field from `RANK` to `ACTIVITY`. We don't want to change the default behavior.

Betree (changes requested):
This fails with:
image

Going with a different fix in a separate PR. Can we keep this one only about updating the `deprecationReason`?

server/graphql/v2/query/collection/AccountsQuery.ts
```diff
@@ -52,8 +53,9 @@ const AccountsQuery = {
       description: 'Limit the search to collectives belonging to these countries',
     },
     orderBy: {
-      type: OrderByInput,
+      type: new GraphQLNonNull(OrderByInput),
       description: 'The order of results',
+      defaultValue: { field: 'RANK', direction: 'DESC' },
```
SudharakaP, line 58: @Betree : I've defaulted this to `RANK` for all cases for now as the `ACTIVITY` sort is not efficient when there's no search term. I believe after, https://github.com/opencollective/opencollective/issues/5454 we should change this default field to `ACTIVITY`.

Betree: @SudharakaP This will still crash:

image

As per my comment above, I'm going with a different fix in a separate PR. Can we keep this one only about updating the `deprecationReason`?

SudharakaP: Ah, I see the problem now, I was thinking that we were using the `RANK` in the same way in both branches. Anyways, yes I've update the PR to keep it only for `deprecationReason`.

Betree (approved):

Betree:
Followup in https://github.com/opencollective/opencollective-api/pull/7514

PR #7521, opened 2022-05-10
Host require minimum number of admins for Collectives

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3593


server/graphql/v2/interface/HasMembers.js
```diff
@@ -89,4 +90,26 @@ export const HasMembersFields = {
       return { nodes: result.rows, totalCount: result.count, limit: args.limit, offset: args.offset };
     },
   },
+  memberInvitations: {
```
znarf, line 93: mutualize code with memberInvitations top level query.

znarf (commented):
Check createCollectiveMutation:
- disable automated approval if not enough admins
- OR throw exception

server/graphql/v2/mutation/HostApplicationMutations.ts
```diff
@@ -113,6 +148,24 @@ const HostApplicationMutations = {
         throw new ValidationFailed('This collective application has already been approved');
       }
 
+      // If approving, check if Collective has enough admins
+      const admins = await models.Member.findAll({
+        where: {
+          CollectiveId: account.id,
+          role: MemberRoles.ADMIN,
+        },
+      });
```
znarf, line 157: Should we count MemberInvitations here? I personally think not.

kewitz: We should, at least if we want to be consistent with the specs and designs.

kewitz:
Rebased.

kewitz:
@znarf I had to update our existing DB dump because of E2E tests.
The reason is that the open source create collective flow requires the 'opensource' slug host and our current dump is outdated. The current dump is from the time we had separated collective/host for OSC.

server/graphql/v2/interface/Account.js
```diff
@@ -529,12 +537,13 @@ const accountFieldsDefinition = () => ({
     },
   },
   policies: {
-    type: new GraphQLList(Policy),
+    type: new GraphQLNonNull(GraphQLJSON),
```
Betree, line 540: I would advocate for something typed there. JSON is great for flexibility, but then there's no way for frontends to know what to expect form this field. Also, that looks like a breaking change?

kewitz: My idea is that each policy can have its own structure to store its specific parameters.
I think this could definitely be typed, but I don't think that, without a typed frontend, it will make much of a difference since we'll need to reproduce these structures using React's PropTypes.

This is a breaking change that is being addressed in this same PR and includes a migration to update existing data.

kewitz: Also, I first thought it would be an elegant solution to have policies as a polymorphic value that could be either true/false or, in the case it was true, an object that represented the policy parameters:
```js
{
  policies: {
    SIMPLE_POLICY: true,
    MINIMUM_ADMINS: { numberOfAdmins: 3, freezeCollectives: true },
  }
}
```

Betree: > without a typed frontend, it will make much of a difference since we'll need to reproduce these structures using React's PropTypes.

It does make a difference with ESLint, which detects when we fetch nested fields that are deprecated/removed or have a typo in their names. This is our public API, so I would see this not from the point of view of "What are our frontend needs?" but rather "What would a developer (_internal or external_) expect here?".

> This is a breaking change that is being addressed in this same PR and includes a migration to update existing data.

But what about the frontend, is it not going to be impacted by this?

> Also, I first thought it would be an elegant solution to have policies as a polymorphic value that could be either true/false or, in the case it was true, an object that represented the policy parameters

Typing doesn't prevent us from doing that. In fact, it will help to detect breaking changes if we ever change the type.

If we don't use something typed here, then I don't see the point of using this `policies` field VS retrieving this info directly in `settings.minimumNumberOfAdmins`.

```es6
const HostPolicies = new GraphQLObjectType({
  name: 'HostPolicies',
  description: 'The policies for a given host',
  fields: () => ({
    SIMPLE_POLICY: { type: GraphQLBoolean, description: '...' },
    MINIMUM_ADMINS: {
      type: new GraphQLObjectType({
        name: 'HostPolicyMinimumAdminsConfig',
        fields: () => ({
          numberOfAdmins: { ... },
          freezeCollectives: {}
        })
      })
    }
  })
})
```

kewitz: > But what about the frontend, is it not going to be impacted by this?

Yes, addressed in https://github.com/opencollective/opencollective-frontend/pull/7771

Cool, I agree and I'll type policies.

server/graphql/v2/interface/Account.js
```diff
@@ -185,6 +188,13 @@ const accountFieldsDefinition = () => ({
       },
     },
   },
+  memberInvitations: {
+    description: 'Get pending member invitations for this account',
+    type: new GraphQLList(MemberInvitation),
```
znarf, line 193: What's the thought process to have a simple `GraphQLList` instead of a `*Collection` structure?

server/graphql/v2/mutation/MemberInvitationMutations.js
```diff
@@ -148,6 +150,22 @@ const memberInvitationMutations = {
 
       if (args.accept) {
         await invitation.accept();
+
+        // Restore financial contributions if Collective now has enough admins to comply with host policy
+        const collective = await invitation.getCollective();
+        console.log({ collective });
```
znarf, line 156: console log

server/graphql/v2/mutation/MemberInvitationMutations.js
```diff
@@ -148,6 +150,22 @@ const memberInvitationMutations = {
...
+          const adminCount = await models.Member.count({
+            where: {
+              CollectiveId: collective.id,
+              role: MemberRoles.ADMIN,
+            },
+          });
+          if (host.data?.policies?.[POLICIES.COLLECTIVE_MINIMUM_ADMINS]?.numberOfAdmins <= adminCount) {
+            await collective.enableFeature(FEATURE.RECEIVE_FINANCIAL_CONTRIBUTIONS);
+          }
+        }
```
znarf, line 168: Should this deserve a separated helper?

server/graphql/v2/interface/Account.js
znarf: Ok, to match the behavior of the existing `memberInvitations` top level query.

cron/daily/41-suspend-accounts-without-minimum-amount-of-admins.ts
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  for (const collective of collectives) {
+    const admins = collective.members.length;
+    const requiredAdmins = collective.host.data.policies.COLLECTIVE_MINIMUM_ADMINS.numberOfAdmins;
+    if (admins < requiredAdmins) {
+      logger.info(
+        `Collective ${collective.slug} has ${admins} out of ${requiredAdmins} required admins, suspending donations.`,
+      );
+      await collective.disableFeature(FEATURE.RECEIVE_FINANCIAL_CONTRIBUTIONS);
```
znarf, line 34: Do they receive an alert?

znarf (approved):
Looks good for me. I see many repetitions that I think would be better to consolidate in some helper library (lib/policies.js) ?

Feel a bit wrong to repeat long error prone sequences like the following, and have variations in the way we defend around missing `host`, empty `data`, missing policy.

`collective.host.data.policies.COLLECTIVE_MINIMUM_ADMINS.numberOfAdmins`

server/graphql/v2/mutation/MemberInvitationMutations.js
```diff
@@ -148,6 +150,22 @@ const memberInvitationMutations = {
 
       if (args.accept) {
         await invitation.accept();
+
+        // Restore financial contributions if Collective now has enough admins to comply with host policy
+        const collective = await invitation.getCollective();
+        console.log({ collective });
+        if (collective.data?.features?.[FEATURE.RECEIVE_FINANCIAL_CONTRIBUTIONS] === false) {
+          const host = await collective.getHost();
```
znarf, line 158: Missing host?

cron/daily/41-suspend-accounts-without-minimum-amount-of-admins.ts
znarf: Can we double check how this affects Projects and Events of the Collective?

server/graphql/v2/interface/Account.js
kewitz: @znarf exactly. I also think that, since this list should be pretty reduced we won't need to paginate it.
I'm OK with updating this if we decide every list should be collections.

kewitz:
Rebased.

cron/daily/41-suspend-accounts-without-minimum-amount-of-admins.ts
```diff
@@ -0,0 +1,50 @@
...
+import { roles } from '../../server/constants';
+import FEATURE from '../../server/constants/feature';
+import POLICIES from '../../server/constants/policies';
+import logger from '../../server/lib/logger';
+import { getPolicy } from '../../server/lib/policies';
+import models, { sequelize } from '../../server/models';
+
+const run = async () => {
+  const collectives = await models.Collective.findAll({
+    include: [
```
Betree, line 12: Shouldn't we pre-filter by `type`? Events and projects inherit the admins, so it doesn't seem necessary to check them.

PR #7534, opened 2022-05-12
Expense: INCOMPLETE status leads to APPROVED after editing

kewitz:


Betree:
@kewitz There's an important issue with this change: since collective admins can mark expenses as incomplete too then this can bypass the regular approval:
1. Submit an expense with incomplete info (PENDING)
2. Collective admin marks the expense as incomplete (INCOMPLETE)
3. Edit the expense, it goes to APPROVED even though the collective admin has never approved it

Betree:
If we want to keep this feature (I think we do, it's a nice one) then from my point of view we should only do it if the last status before going to `INCOMPLETE` was APPROVED. With pseudocode:

```es6
else if (status === expenseStatus.INCOMPLETE) {
  const statuses = await models.Activity.findAll({ where: ..., order: [['createdAt', 'ASC']] });
  const lastStatusBeforeIncomplete = findLast(statuses, s => s !== 'INCOMPLETE');
  status = ['APPROVED' || 'ERROR'].includes(lastStatusBeforeIncomplete) ? 'APPROVED' : 'PENDING';
}

```

PR #7538, opened 2022-05-16
Misc fixes and improvements for `processHostApplication`

Betree:
After investigating an issue [reported on Slack](https://opencollective.slack.com/archives/C03AWL58JRM/p1652469349747269) with errors during collective approvals, I improved a few things on the mutation:
- Run approve in an SQL transaction to make sure we don't end up approving half accounts if something goes wrong
- Use a `UPDATE WHERE` query rather than a `Promise.all(..., c => c.update(....))` for simpler code & better performances
- Wait for the approval SQL query to succeed before sending the "Approved" email
- Wait for the reject SQL query to succeed before sending the "Rejected" email


Betree (commented):
_Removed_

PR #7539, opened 2022-05-16
Add tests for GQLV2 Application query & mutations

Betree:
For https://github.com/opencollective/opencollective-api/pull/7508


Betree:
Commit will be cherrypicked and edited directly

PR #7541, opened 2022-05-16
Add support to get crypto currency to usd exchange rate using TGB

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5448

Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/7788


server/lib/currency.ts
```diff
@@ -264,3 +265,20 @@ export function reduceArrayToCurrency(array: AmountWithCurrencyAndDate[], curren
...
+  }
+  try {
+    const { rate: cryptoToUSDRate } = await getCryptoToUSDRate(cryptoCurrency);
+    const USDToCollectiveCurrencyRate = await getFxRate('USD', fiatCurrency);
+    return cryptoToUSDRate * USDToCollectiveCurrencyRate;
+  } catch (error) {
+    logger.error(error);
+    return null;
+  }
+}
```
znarf, line 284: At line 144 of `server/paymentProviders/thegivingblock/index.js` we're fetching a crypto fxRate without the need of this. What is justifying it?

server/graphql/v2/query/CryptoExchangeRateQuery.js
```diff
@@ -0,0 +1,22 @@
...
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'Crypto currency symbol. Example: BTC',
+    },
+    collectiveCurrency: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'Fiat currency symbol. Example: USD',
+    },
+  },
+  async resolve(_, args) {
+    return getCryptoFxRate(args.cryptoCurrency, args.collectiveCurrency);
```
znarf, line 18: Why do we need something specific to crypto? Could we consider something generic instead `exchangeRateQuery `

znarf, line 18: Why do we need something specific to crypto? Could we consider something generic instead `exchangeRateQuery `

server/graphql/v2/query/CryptoExchangeRateQuery.js
```diff
@@ -0,0 +1,22 @@
...
+
+const CryptoExchangeRateQuery = {
+  type: GraphQLString,
+  args: {
+    cryptoCurrency: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'Crypto currency symbol. Example: BTC',
+    },
+    collectiveCurrency: {
+      type: new GraphQLNonNull(GraphQLString),
```
znarf, line 13: Generally, a currency should always be an Enum.

znarf, line 13: Generally, a currency should always be an Enum.

server/lib/currency.ts
```diff
@@ -264,3 +264,13 @@ export function reduceArrayToCurrency(array: AmountWithCurrencyAndDate[], curren
...
+
+/**
+ * Gets exchange rate between given crypto currency and fiat.
+ */
+export async function exchangeRateQuery(cryptoCurrency: string, fiatCurrency: string): Promise<string> {
+  if (!cryptoCurrency || !fiatCurrency) {
+    return null;
+  }
+  return getFxRate(cryptoCurrency, fiatCurrency);
+}
```
znarf, line 276: I don't see what this function is doing.

server/graphql/v2/query/CryptoExchangeRateQuery.js
```diff
@@ -0,0 +1,24 @@
+import { GraphQLNonNull, GraphQLString } from 'graphql';
+
+import { exchangeRateQuery } from '../../../lib/currency';
+import { Currency } from '../enum';
+import { CryptoCurrency } from '../enum/Currency';
+
+const CryptoExchangeRateQuery = {
```
znarf, line 7: Why do we need something specific to crypto? Could we consider something generic instead: exchangeRateQuery

server/graphql/v2/query/CryptoExchangeRateQuery.js
SudharakaP: I've changed both the `collectiveCurrency` as well as the `cryptoCurrency` to enums.

server/lib/currency.ts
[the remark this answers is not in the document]
SudharakaP: @znarf : This is interesting. I spent some time looking into how line 144 of `server/paymentProviders/thegivingblock/index.js` and I think there's a problem there. It uses the `fetchFxRates` function internally. And that in turn uses, `https://data.fixer.io/`. I think this might work for basic crypto like `BTC` but do not support other crypto currencies. If we are going this route, we have to use, https://coinlayer.com/. 

The complete list of what `data.fixer.io` supports is given at by `http://data.fixer.io/api/latest?access_key = API_KEY` endpoint. And this is what I got; 

```
{"AED":3.843287,"AFN":94.698551,"ALL":119.597048,"AMD":478.695814,"ANG":1.886015,"AOA":429.881799,"ARS":123.461607,"AUD":1.504471,"AWG":1.883956,"AZN":1.777998,"BAM":1.945875,"BBD":2.11297,"BDT":91.61939,"BGN":1.955595,"BHD":0.39441,"BIF":2106.306088,"BMD":1.046352,"BND":1.451604,"BOB":7.205001,"BRL":5.2004,"BSD":1.046436,"BTC":3.5896715e-5,"BTN":81.109262,"BWP":12.730946,"BYN":3.531532,"BYR":20508.494452,"BZD":2.109368,"CAD":1.349004,"CDF":2097.410525,"CHF":1.034631,"CLF":0.032412,"CLP":894.34848,"CNY":7.067587,"COP":4257.573912,"CRC":699.822946,"CUC":1.046352,"CUP":27.728322,"CVE":109.291137,"CZK":24.6668,"DJF":185.958022,"DKK":7.441811,"DOP":57.810521,"DZD":152.843753,"EGP":19.108161,"ERN":15.695277,"ETB":53.881779,"EUR":1,"FJD":2.286174,"FKP":0.856262,"GBP":0.847477,"GEL":3.102446,"GGP":0.856262,"GHS":8.004598,"GIP":0.856262,"GMD":56.660158,"GNF":9247.657522,"GTQ":8.026482,"GYD":218.935718,"HKD":8.213076,"HNL":25.582999,"HRK":7.535512,"HTG":116.681814,"HUF":383.791782,"IDR":15423.277227,"ILS":3.520519,"IMP":0.856262,"INR":81.431384,"IQD":1527.673566,"IRR":44260.67923,"ISK":139.32156,"JEP":0.856262,"JMD":161.968803,"JOD":0.741873,"JPY":134.161162,"KES":121.794106,"KGS":83.998504,"KHR":4248.188059,"KMF":492.256339,"KPW":941.716815,"KRW":1334.966689,"KWD":0.320704,"KYD":0.87203,"KZT":453.282298,"LAK":13843.234127,"LBP":1587.315328,"LKR":376.724339,"LRD":160.616623,"LSL":16.793831,"LTL":3.089605,"LVL":0.632928,"LYD":5.029125,"MAD":10.557161,"MDL":19.961774,"MGA":4231.968622,"MKD":61.30021,"MMK":1937.48614,"MNT":3218.338181,"MOP":8.45944,"MRO":373.547398,"MUR":45.251157,"MVR":16.156012,"MWK":852.776506,"MXN":20.984951,"MYR":4.599243,"MZN":66.788515,"NAD":16.794561,"NGN":434.424251,"NIO":37.386489,"NOK":10.360244,"NPR":129.769883,"NZD":1.661931,"OMR":0.40283,"PAB":1.046436,"PEN":3.949986,"PGK":3.677921,"PHP":54.907314,"PKR":206.971386,"PLN":4.647947,"PYG":7168.456934,"QAR":3.80979,"RON":4.947143,"RSD":117.484602,"RUB":68.532323,"RWF":1068.848321,"SAR":3.925027,"SBD":8.500065,"SCR":13.594645,"SDG":475.573167,"SEK":10.525984,"SGD":1.455501,"SHP":1.441242,"SLL":13550.255534,"SOS":606.345977,"SRD":22.055527,"STD":21657.368809,"SVC":9.155982,"SYP":2628.907015,"SZL":16.815016,"THB":36.266426,"TJS":13.070381,"TMT":3.672695,"TND":3.226894,"TOP":2.434913,"TRY":16.684392,"TTD":7.097819,"TWD":31.195717,"TZS":2433.813942,"UAH":30.917224,"UGX":3834.291762,"USD":1.046352,"UYU":43.339855,"UZS":11666.821997,"VEF":223741629001.3448,"VND":24213.626026,"VUV":119.562077,"WST":2.6989,"XAF":652.587114,"XAG":0.048847,"XAU":0.000576,"XCD":2.827818,"XDR":0.781465,"XOF":658.680865,"XPF":118.813571,"YER":261.907391,"ZAR":16.82596,"ZMK":9418.422908,"ZMW":17.81621,"ZWL":336.924839}
```

And trying to make a payment on staging using SUSHI, will print this error on the api; 

image

So I think even there we have to use the same approach of using the currency returned by TGB as I've done in this PR. WDYT? If you agree, I will open a new PR and work on this next. 🤔

server/lib/currency.ts
SudharakaP: Sorry about this, I initially thought that we could simply use `getFxRate` and made a commit like this; but later reverted it back to my original function; please see the above comment; https://github.com/opencollective/opencollective-api/pull/7541#discussion_r876427807

server/lib/currency.ts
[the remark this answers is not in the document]
znarf: The code is designed for failure and it's not a problem:

```
 // update approximate amount in order currency
  let cryptoFxRate;
  try {
    cryptoFxRate = await getFxRate(order.data.thegivingblock.pledgeCurrency, order.currency);
  } catch (e) {
    console.log(e);
  }
  if (cryptoFxRate) {
    const totalAmount = Math.round(order.data.thegivingblock.pledgeAmount * cryptoFxRate * 100);
    await order.update({ totalAmount });
  }
```

What are the crypto currencies we're currently supporting and Fixer not? Should we care?

server/graphql/v2/query/CryptoExchangeRateQuery.js
znarf: It's overly specific, if it's in GraphQL API v2, I want something semantically clean and generic.

server/graphql/v2/query/CryptoExchangeRateQuery.js
znarf: It's overly specific, if it's in GraphQL API v2, I want something semantically clean and generic.

server/lib/currency.ts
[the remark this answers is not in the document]
SudharakaP: Yes, the failure isn't necessarily a problem, since even if the crypto currency exchange rate doesn't exist on fixer that doesn't matter as this just for the interim before the transaction is verified. If the user actually contributes this amount then using we will [update that amount to what's returned from TGB](https://github.com/opencollective/opencollective-api/blob/869a5fb0c9e6144c57be4c219602be3183f2647e/server/paymentProviders/thegivingblock/webhook.js#L37). 

So I don't think we should care about this; however when showing the exchange rate on the we cannot use `fetchFxRates` as that doesn't return the correct amount. So your initial question of, 

> At line 144 of `server/paymentProviders/thegivingblock/index.js` we're fetching a crypto fxRate without the need of this. What is justifying it?

we have to use a method to obtain the exchange rate that returns the correct values for all supported cryptocurrencies of TGB. This is the reason I used the specialized function with a call to the TGB endpoint; `crypto-to-usd-rate`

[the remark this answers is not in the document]
SudharakaP: > What are the crypto currencies we're currently supporting and Fixer not? 

Fixer seems to support `BTC` only. If we want more using their API we should use https://coinlayer.com/

server/graphql/v2/query/CryptoExchangeRateQuery.js
SudharakaP: Ah so you mean we change this query should support both crypto exchange rates and non-crypto exchange rates (user can pass in crypto currency or fiat currency symbols)? Or do you mean we just rename it as `exchangeRateQuery`? The part I didn't quite get is renaming this to `exchnageRateQuery` doesn't tally with what it does. Currently it only returns the exchange rate for crypto currencies. 🤔

znarf: > Ah so you mean we change this query should support both crypto exchange rates and non-crypto exchange rates (user can pass in crypto currency or fiat currency symbols)?

Yes

server/graphql/v2/query/CryptoExchangeRateQuery.js
SudharakaP: I've refactored this to `ExchangeRateQuery`.

server/lib/currency.ts
```diff
@@ -269,3 +270,32 @@ export function reduceArrayToCurrency(array: AmountWithCurrencyAndDate[], curren
...
+    ) {
+      const { rate: cryptoToUSDRate } = await getCryptoToUSDRate(toCurrency);
+      const fromCurrencyUSDRate = await getFxRate(fromCurrency, 'USD');
+      return fromCurrencyUSDRate / cryptoToUSDRate;
+    } else if (SUPPORTED_CRYPTO_CURRENCIES.includes(fromCurrency) && SUPPORTED_CRYPTO_CURRENCIES.includes(toCurrency)) {
+      const cryptoFromCurrencyToUSD = await getCryptoToUSDRate(fromCurrency);
+      const cryptoToCurrencyToUSD = await getCryptoToUSDRate(toCurrency);
+      return cryptoFromCurrencyToUSD / cryptoToCurrencyToUSD;
+    } else {
+      return getFxRate(fromCurrency, toCurrency);
```
znarf, line 295: Let's move the generic case and have it first.

server/lib/currency.ts
```diff
@@ -269,3 +270,32 @@ export function reduceArrayToCurrency(array: AmountWithCurrencyAndDate[], curren
...
   );
 }
+
+/**
+ * Gets exchange rate between given crypto currency and fiat.
+ */
+export async function getExchangeRate(fromCurrency: string, toCurrency: string): Promise<string> {
+  try {
+    if (SUPPORTED_CRYPTO_CURRENCIES.includes(fromCurrency) && !SUPPORTED_CRYPTO_CURRENCIES.includes(toCurrency)) {
+      const { rate: cryptoToUSDRate } = await getCryptoToUSDRate(fromCurrency);
```
znarf, line 280: `getCryptoToUSDRate` should just returns the rate, instead of having to do `{ rate: cryptoToUSDRate }`.

server/graphql/v2/query/ExchangeRateQuery.js
```diff
@@ -0,0 +1,23 @@
+import { GraphQLNonNull, GraphQLString } from 'graphql';
+
+import { getExchangeRate } from '../../../lib/currency';
+import { Currency } from '../enum';
+
+const ExchangeRateQuery = {
+  type: GraphQLString,
```
znarf, line 7: Is String a good choice here?

server/constants/currencies.ts
```diff
@@ -169,6 +169,52 @@ const IBANCurrencies = {
   ZAR: 'Rand',
   ZMW: 'Zambian Kwacha',
   ZWL: 'Zimbabwe Dollar',
+  BTC: 'Bitcoin',
```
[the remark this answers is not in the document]
SudharakaP: I've separated them to two sections. 👍🏽 

https://github.com/opencollective/opencollective-api/blob/c552f260b16976121bc61000bc8018bf0dfbd052/server/constants/currencies.ts#L174

server/lib/currency.ts
SudharakaP: Done. 👍🏽

server/lib/currency.ts
SudharakaP: Done. 👍🏽

server/graphql/v2/query/ExchangeRateQuery.js
SudharakaP: Ops, I think `GraphQLFloat` is a better choice as what we are returning is actually a `float`. I've changed it to that. 👍🏽

server/lib/currency.ts
```diff
@@ -269,3 +270,32 @@ export function reduceArrayToCurrency(array: AmountWithCurrencyAndDate[], curren
...
+      SUPPORTED_CRYPTO_CURRENCIES.includes(toCurrency) &&
+      !SUPPORTED_CRYPTO_CURRENCIES.includes(fromCurrency)
+    ) {
+      const { rate: cryptoToUSDRate } = await getCryptoToUSDRate(toCurrency);
+      const fromCurrencyUSDRate = await getFxRate(fromCurrency, 'USD');
+      return fromCurrencyUSDRate / cryptoToUSDRate;
+    } else if (SUPPORTED_CRYPTO_CURRENCIES.includes(fromCurrency) && SUPPORTED_CRYPTO_CURRENCIES.includes(toCurrency)) {
+      const cryptoFromCurrencyToUSD = await getCryptoToUSDRate(fromCurrency);
+      const cryptoToCurrencyToUSD = await getCryptoToUSDRate(toCurrency);
+      return cryptoFromCurrencyToUSD / cryptoToCurrencyToUSD;
```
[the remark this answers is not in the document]
SudharakaP: For this I was thinking right now we don't need two of the three different cases. We only need the first case where the `fromCurrency` is crypto and the `toCurrency` if fiat. So I've went ahead and removed the two cases. I think when I initially wrote this code I was thinking about the future and making things too complicated. I don't think we need to add code that we don't use. WDYT? 🤔

server/lib/currency.ts
```diff
@@ -269,3 +270,21 @@ export function reduceArrayToCurrency(array: AmountWithCurrencyAndDate[], curren
...
+ * Gets exchange rate between given crypto currency and fiat.
+ */
+export async function getExchangeRate(fromCurrency: string, toCurrency: string): Promise<string> {
+  try {
+    if (!SUPPORTED_CRYPTO_CURRENCIES.includes(fromCurrency) && !SUPPORTED_CRYPTO_CURRENCIES.includes(toCurrency)) {
+      return getFxRate(fromCurrency, toCurrency);
+    } else {
+      const cryptoToUSDRate = await getCryptoToUSDRate(fromCurrency);
+      const USDtoCurrencyRate = await getFxRate('USD', toCurrency);
+      return cryptoToUSDRate * USDtoCurrencyRate;
```
znarf, line 284: I don't get the rationale, please double check.

Make sure of the behavior when `fromCurrency` is fiat and `toCurrency` crypto and add some comments to explain if that's indeed correct.

SudharakaP: My thinking is that we should have this function only compute the crypto exchange rate. So `fromCurrency` is always crypto and `toCurrency` is always fiat. To that end, I've changed the naming of the arguments to `cryptoCurrency` and `fiatCurrency` and also renamed the function name to `getCryptoExchangeRate`. Does this make things clearer? Also I realized we don't need the default `getFxRate` call here as again we only use the function when the `fromCurrency` is a crypto currency.

znarf: That's not true, in the `exchangeRateQuery` we accept any currency.

SudharakaP: Ops sorry, you are correct. I was losing the context of this in my mind and making a mess. Anyways, I've simplified the function a bit. I think this looks better. Let me know. Thanks.

server/lib/currency.ts
```diff
@@ -269,3 +270,30 @@ export function reduceArrayToCurrency(array: AmountWithCurrencyAndDate[], curren
...
+    const isFromCurrencyCrypto = SUPPORTED_CRYPTO_CURRENCIES.includes(fromCurrency);
+    const isToCurrencyCrypto = SUPPORTED_CRYPTO_CURRENCIES.includes(toCurrency);
+    const cryptoFromCurrencyToUSD = isFromCurrencyCrypto ? await getCryptoToUSDRate(fromCurrency) : undefined;
+    const cryptoToCurrencyToUSD = isToCurrencyCrypto ? await getCryptoToUSDRate(toCurrency) : undefined;
+
+    if (!isFromCurrencyCrypto && !isToCurrencyCrypto) {
+      return getFxRate(fromCurrency, toCurrency);
+    } else if (isFromCurrencyCrypto && !isToCurrencyCrypto) {
+      const toCurrencyUSDRate = await getFxRate(toCurrency, 'USD');
+      return cryptoFromCurrencyToUSD / toCurrencyUSDRate;
```
znarf, line 288: Why use division instead of doing the opposite? It's the more natural way to use fx Rates.

znarf, line 288:
```suggestion
      const fxRate = await getFxRate('USD', toCurrency);
      return cryptoFromCurrencyToUSD * fxRate;
```

znarf (changes requested):
After reviewing this code once again, I have no confidence in it working properly.

Please back all cases with tests.

server/lib/currency.ts
```diff
@@ -269,3 +270,30 @@ export function reduceArrayToCurrency(array: AmountWithCurrencyAndDate[], curren
...
 }
+
+/**
+ * Gets exchange rate between given crypto currency and fiat.
+ */
+export async function getExchangeRate(fromCurrency: string, toCurrency: string): Promise<string> {
+  try {
+    const isFromCurrencyCrypto = SUPPORTED_CRYPTO_CURRENCIES.includes(fromCurrency);
+    const isToCurrencyCrypto = SUPPORTED_CRYPTO_CURRENCIES.includes(toCurrency);
+    const cryptoFromCurrencyToUSD = isFromCurrencyCrypto ? await getCryptoToUSDRate(fromCurrency) : undefined;
```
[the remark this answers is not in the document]
SudharakaP: Done. 👍🏽

SudharakaP:
> After reviewing this code once again, I have no confidence in it working properly.
> 
> Please back all cases with tests.

Thanks. I've added some test cases to cover the `getExchangeRate` function. 

https://github.com/opencollective/opencollective-api/blob/1d0912c29be47591e538910d7d0264703b899034/test/server/lib/currency.test.js#L133-L163

PR #7551, opened 2022-05-18
Add support for period sign in search term

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5568

image


Betree (approved):
:+1:

PR #7557, opened 2022-05-19
Prevent Sending Email Notifications for Orders from Collective to Itself

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5494

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -477,7 +477,7 @@ export const sendEmailNotifications = (order, transaction) => {
     order.fromCollective?.HostCollectiveId !== order.collective?.id &&
     // Check if transaction is from a collective to itself. This happens when we add funds
     // choosing the source as itself. In this case do not send an email.
-    order.fromCollective.id !== order.collective.id
+    order.fromCollective?.id !== order.collective?.id
   ) {
     sendOrderConfirmedEmail(order, transaction); // async
   } else if (order.status === status.PENDING && order.paymentMethod?.type === 'crypto') {
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -472,9 +472,12 @@ export const sendEmailNotifications = (order, transaction) => {
     // Check if transaction is from host to one of its hosted collective/fund/project/event
     // fromCollective: host, collective: a collective/fund/project/event
     order.fromCollective?.id !== order.collective?.HostCollectiveId &&
-    // Check is transaction is from a collective/fund/project/event to its host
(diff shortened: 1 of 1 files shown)


server/lib/payments.js
```diff
@@ -472,9 +472,12 @@ export const sendEmailNotifications = (order, transaction) => {
...
     // fromCollective: host, collective: a collective/fund/project/event
     order.fromCollective?.id !== order.collective?.HostCollectiveId &&
-    // Check is transaction is from a collective/fund/project/event to its host
+    // Check if transaction is from a collective/fund/project/event to its host
     // fromCollective: a collective/fund/project/event, collective: host of fromCollective
-    order.fromCollective?.HostCollectiveId !== order.collective?.id
+    order.fromCollective?.HostCollectiveId !== order.collective?.id &&
+    // Check if transaction is from a collective to itself. This happens when we add funds
+    // choosing the source as itself. In this case do not send an email.
+    order.fromCollective.id !== order.collective.id
```
Betree, line 480: In the checks above, we assume that `order.fromCollective` / `order.collective` can be bullish (`?` coalesce operator).  Either:
- It's legit => then let's add it here as well
- Or it's unnecessary => let's create a `debt` followup issue to remove them

server/lib/payments.js
```diff
@@ -472,9 +472,12 @@ export const sendEmailNotifications = (order, transaction) => {
     // Check if transaction is from host to one of its hosted collective/fund/project/event
     // fromCollective: host, collective: a collective/fund/project/event
     order.fromCollective?.id !== order.collective?.HostCollectiveId &&
-    // Check is transaction is from a collective/fund/project/event to its host
+    // Check if transaction is from a collective/fund/project/event to its host
     // fromCollective: a collective/fund/project/event, collective: host of fromCollective
-    order.fromCollective?.HostCollectiveId !== order.collective?.id
+    order.fromCollective?.HostCollectiveId !== order.collective?.id &&
+    // Check if transaction is from a collective to itself. This happens when we add funds
+    // choosing the source as itself. In this case do not send an email.
```
Betree, line 479: Great comment

Betree (commented):
Haven't tested but the code looks pretty straightforward. Let's address the review below then feel free to merge.

server/lib/payments.js
SudharakaP: It seems that these are unnecessary. But for now I've added them just in case. I've created a debt issue and will iterate on it later after some more testing; https://github.com/opencollective/opencollective/issues/5576

PR #7563, opened 2022-05-19
canMarkAsIncomplete only approved and errored expenses

kewitz:
Related to https://github.com/opencollective/opencollective-api/pull/7534#issuecomment-1125725790


Betree (approved):
That works if it doesn't break any other use case. The alternative if we want to support this again in the future is to look at the approval history (see https://github.com/opencollective/opencollective-api/pull/7534#issuecomment-1125729670)

PR #7568, opened 2022-05-20
Improve OAuth authorize flow

Betree:


server/routes.js
```diff
@@ -113,12 +113,25 @@ export default async app => {
...
+  app.post('/oauth/token', noCache, app.oauth.token());
   app.post(
     '/oauth/authorize',
+    noCache,
     app.oauth.authorize({ allowEmptyState: true, authenticateHandler: authorizeAuthenticateHandler }),
   );
-  app.post('/oauth/authenticate', app.oauth.authenticate());
+  app.post('/oauth/authenticate', noCache, app.oauth.authenticate());
+  app.get('/oauth/me', noCache, async (req, res) => {
+    if (req.remoteUser) {
```
Betree, line 124: Shouldn't we reply with a 404 if there's no `remoteUser`?

znarf: Maybe, I'm gonna remove this thing for now, will potentially re-implement it differently.

znarf (approved):

PR #7574, opened 2022-05-23
Handle null collective in `user.isAdminOfCollective`

Betree:
Fix https://sentry.io/organizations/open-collective/issues/3275588026

Consistent with what we do in `isAdminOfCollectiveOrHost` and other permissions helpers.

### DIFF
--- a/server/models/User.js
+++ b/server/models/User.js
@@ -315,7 +315,9 @@ function defineModel() {
 
   // Slightly better API than the former
   User.prototype.isAdminOfCollective = function (collective) {
-    if (collective.type === 'EVENT' || collective.type === 'PROJECT') {
+    if (!collective) {
+      return false;
+    } else if (collective.type === 'EVENT' || collective.type === 'PROJECT') {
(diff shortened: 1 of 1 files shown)


znarf:
Ok for me. In which case would `collective` be null?

znarf (approved):

Betree:
> Ok for me. In which case would `collective` be null?

For this Sentry issue, it's an `expense.collective` (expense #71653) that has been deleted, not sure why & how but I've seen similar cases where we have a non-deleted expense on a deleted collective in the past.

znarf:
@BenJam for expense 71653, this might have been a case of deletion script?

Betree:
@znarf #462373 doesn't have its `data.isBanned` set to true, so it has not been banned using `sql/ban-collectives.sql`. What could be the other places, the bayesian script?

Betree:
@znarf That sounds likely. Would be great to have a look at https://github.com/opencollective/opencollective/issues/2751 and see how we could unify with `ban-collectives` and `merge-accounts` to make sure we delete all related data.

PR #7576, opened 2022-05-23
OAuth: implement OAuthAuthorizations

znarf:


server/graphql/v2/object/Application.js
```diff
@@ -91,6 +91,7 @@ export const Application = new GraphQLObjectType({
         if (userToken) {
           return {
             account: await req.loaders.Collective.byId.load(userToken.user.CollectiveId),
+            createdByAccount: () => req.loaders.Collective.byUserId.load(userToken.UserId),
```
znarf, line 94: I don't understand the point, we already have `account` above, same thing.

Betree: That's right, the confusion is on my side

PR #7577, opened 2022-05-23
Fix Virtual Card pending expense checker to ignore refunds

kewitz:


server/paymentProviders/utils.js
```diff
@@ -127,6 +127,17 @@ export const persistTransaction = async (virtualCard, transaction) => {
...
 
+    const expense = await models.Expense.findOne({
+      where: {
+        CollectiveId: collective.id,
+        FromCollectiveId: vendor.id,
+        VirtualCardId: virtualCard.id,
+        type: ExpenseType.CHARGE,
+        status: ExpenseStatus.PAID,
+        amount,
+      },
```
znarf, line 138: I would order DESC to make sure to pick the latest ?

znarf: And then should we somehow track that it was refunded?

server/paymentProviders/utils.js
```diff
@@ -127,6 +129,28 @@ export const persistTransaction = async (virtualCard, transaction) => {
...
+      order: [['createdAt', 'DESC']],
+    });
+
+    if (expense) {
+      const [originalCreditTransaction] = await expense.getTransactions({
+        where: { type: TransactionTypes.CREDIT, kind: TransactionKind.EXPENSE },
+      });
+      if (originalCreditTransaction && originalCreditTransaction.amount === amount) {
```
znarf, line 148:
```suggestion
      if (originalCreditTransaction?.amount === amount) {
```

arifulhaq651-netizen (commented):
Arif 6

PR #7579, opened 2022-05-24
Update dev docs after GQLV2 URL as default

Betree:


docs/dev.md
```diff
@@ -2,40 +2,46 @@
...
 
-```
+```gql
 query {
-  Collective(slug:"apex") {
-      id
-      slug
-      name
-      description
+  account(slug: "apex") {
```
znarf, line 11:
```suggestion
  collective(slug: "apex") {
```

So we don't have to do ```... on AccountWithContributions```

docs/dev.md
```diff
@@ -2,40 +2,46 @@
...
+    id
+    slug
+    name
+    description
+    ... on AccountWithContributions {
```
znarf, line 16: Remove for simplicity

PR #7595, opened 2022-05-30
Add Support to Redirect to Welcome Page on First Login

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5573

⚠️ Require: https://github.com/opencollective/opencollective-frontend/pull/7834

Front end PR: https://github.com/opencollective/opencollective-frontend/pull/7834

### DIFF
--- a/server/models/User.js
+++ b/server/models/User.js
@@ -210,6 +210,9 @@ function defineModel() {
 
   User.prototype.generateLoginLink = function (redirect = '/', websiteUrl) {
     const lastLoginAt = this.lastLoginAt ? this.lastLoginAt.getTime() : null;
+    if (!lastLoginAt) {
+      redirect = '/welcome-to-oc';
+    }
     const token = this.jwt({ scope: 'login', lastLoginAt });
     // if a different websiteUrl is passed
     // we don't accept that in production to avoid fishing related issues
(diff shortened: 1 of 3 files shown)


Betree (changes requested):
If I understand correctly what we're doing here, we'll be breaking some fundamental flows on the website. Let's take an example:
- I've been asked to submit an expense on https://opencollective.com/babel/expenses/new
- I go there, click on sign up
- After clicking the email link, I expect to land back at where I was, submitting an expense
- Instead, I'm redirected to a page I don't care about that offers me to create a collective

I don't see this mentioned anywhere on https://github.com/opencollective/opencollective/issues/5573. Why are we doing this?

SudharakaP:
> 

@Betree : Here my intention was that if the user logs in the for the first time, then we show the welcome screen. I missed this kinda flow where the user actually signs in but needs to be re-directed back to the page they were in. Thinking further I think that we need to show the welcome screen in certain instances. Specifically if the user signs in **without** a redirect url; then only we should show the welcome screen. Does that sound reasonable? 🤔

SudharakaP:
> It will be far simpler and more coherent to implement such a change in the frontend, so let's do it there.

@Betree : I've added the logic to handle this at, https://github.com/opencollective/opencollective-frontend/pull/7834. 

https://github.com/opencollective/opencollective-frontend/blob/b17b6af9eb239e2119271cefa6059dddecbba3e1/components/SignInOrJoinFreeV2.js#L98-L108

PR #7608, opened 2022-06-02
Fix user object used in sendOrderConfirmedEmail

znarf:
- If the `createdByUser` is no longer admin of the Organization, PDF could not be fetched.
- If the `createdByUser` opted out the `thankyou` email, the email would not be sent to current admins.

```
SELECT *
FROM "Orders" o
LEFT JOIN "Collectives" c ON c."id" = o."FromCollectiveId"
LEFT JOIN "Users" u ON u."id" = o."CreatedByUserId"
WHERE c."type" != 'USER' AND o."status" = 'ACTIVE'
AND NOT EXISTS (
    SELECT * FROM "Members"
    WHERE "deletedAt" IS NULL
    AND "CollectiveId" = c."id"
    AND "role" = 'ADMIN'
    AND "MemberCollectiveId" = u."CollectiveId"
)
```

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -603,10 +603,28 @@ const validatePayment = payment => {
   }
 };
 
+// For legacy purpose, we want to get a single user that we will use for:
+// - authentication with the PDF service
+// - constructing notification/activity objects
+// We can't rely on createdByUser because they have moved out of the Organization, Collective, etc ...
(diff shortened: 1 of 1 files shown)


znarf:
Follow up issue: https://github.com/opencollective/opencollective/issues/5615

PR #7613, opened 2022-06-03
Move Collectives from githubHandle to repositoryUrl

Betree:
API part of https://github.com/opencollective/opencollective/issues/4165

This PR should have no breaking change.


znarf (approved):
Looks good to me. Maybe need another look from someone else.

Betree:
@kewitz or @SudharakaP would love a second pair of eyes on this one if either of you have the time!

znarf:
@Betree feel free to merge if no more feedback is coming

server/models/Collective.js
```diff
@@ -466,23 +467,39 @@ function defineModel() {
...
-            const [, username, , repository] = regexResult;
-            const formattedHandle = repository ? `${username}/${repository}` : username;
-            this.setDataValue('githubHandle', formattedHandle);
+      repositoryUrl: {
+        type: DataTypes.STRING,
+        validate: {
+          notEmpty: true,
+          isUrl: {
+            msg: 'Repository URL must be a valid URL',
+          },
```
SudharakaP, line 496: Don't we have to do any generic length checking on the url here? Just like we do for the githubHandle ? 🤔

Betree: The repository URL can be a Github URL, a Gitlab URL or anything self-hosted. We can't really enforce strict rules here, apart from the default `2083` length limit from `isUrl`

server/lib/github.js
```diff
@@ -200,3 +202,64 @@ export async function checkGithubStars(githubHandle, accessToken) {
...
+    const regexResult = pathname.match(githubPathnameRegex);
+    if (regexResult) {
+      const handle = trim(regexResult[0], '/');
+      if (githubHandleRegex.test(handle)) {
+        return handle;
+      }
+    }
+  } catch {
+    // Ignore invalid URLs
+  }
```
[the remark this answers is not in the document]
Betree: @SudharakaP The [updated frontend](https://github.com/opencollective/opencollective-frontend/pull/7872) already resolves this by changing the field to an URL rather than a Github handle.

But I'll look at it since we want to prevent any regression with this PR. Thanks!

PR #7598, opened 2022-05-31
Fix editing/removing members

Betree:
Fix https://sentry.io/organizations/open-collective/issues/3313012577/?project=5199682&referrer=slack
Fix https://sentry.io/organizations/open-collective/issues/3313034483/?project=5199682&referrer=slack
Followup on https://github.com/opencollective/opencollective-api/pull/7511

PR #7616, opened 2022-06-04
test: add tests for MemberInvitation mutations

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/5605

> **Note**
>
> The following cases (Editining an invitation and declining an invitation) actually don't document changes in an activity.

`it('should edit the role, description and since and document the changes in an activity', () => {});`
` it('can decline the invitation and document the changes in an activity', () => {});`

https://github.com/opencollective/opencollective-api/blob/8916c35847d63b40a167dce4a8d2944d8b4c25aa/server/models/MemberInvitation.js#L162-L165
https://github.com/opencollective/opencollective-api/blob/8916c35847d63b40a167dce4a8d2944d8b4c25aa/server/models/MemberInvitation.js#L228-L231

### DIFF
--- a/server/models/MemberInvitation.js
+++ b/server/models/MemberInvitation.js
@@ -160,7 +160,18 @@ function defineModel() {
   };
 
   MemberInvitation.prototype.decline = async function () {
-    return this.destroy();
+    await this.destroy();
+    const collective = this.collective || (await this.getCollective());
+    const memberCollective = this.memberCollective || (await this.getMemberCollective());
+    await models.Activity.create({
+      type: ActivityTypes.COLLECTIVE_CORE_MEMBER_INVITATION_DECLINED,
+      CollectiveId: this.CollectiveId,
+      data: {
+        notify: false,
+        memberCollective: memberCollective?.activity,
+        collective: collective?.activity,
+      },
+    });
   };
 
   MemberInvitation.prototype.sendEmail = async function (remoteUser, skipDefaultAdmin = false, sequelizeParams = null) {
--- a/server/models/MemberInvitation.js
+++ b/server/models/MemberInvitation.js
@@ -143,7 +143,7 @@ function defineModel() {
       purgeCacheForCollective(collective.slug);
     }
 
-    if ([roles.ACCOUNTANT, roles.ADMIN, roles.MEMBER].includes(this.role)) {
+    if (MEMBER_INVITATION_SUPPORTED_ROLES.includes(this.role)) {
       const member = await models.Collective.findByPk(this.MemberCollectiveId);
       await models.Activity.create({
         type: ActivityTypes.COLLECTIVE_CORE_MEMBER_ADDED,
@@ -241,7 +241,7 @@ function defineModel() {
       invitation = await MemberInvitation.create({ ...memberParams, CollectiveId: collective.id }, sequelizeParams);
       invitation.collective = collective;
 
-      if ([roles.ACCOUNTANT, roles.ADMIN, roles.MEMBER].includes(this.role)) {
+      if (MEMBER_INVITATION_SUPPORTED_ROLES.includes(memberParams.role)) {
         const member = await models.Collective.findByPk(memberParams.MemberCollectiveId);
         await models.Activity.create({
           type: ActivityTypes.COLLECTIVE_CORE_MEMBER_INVITED,
--- a/test/server/graphql/v2/mutation/MemberInvitationMutations.test.ts
+++ b/test/server/graphql/v2/mutation/MemberInvitationMutations.test.ts
@@ -3,6 +3,7 @@ import gqlV2 from 'fake-tag';
 import { describe, it } from 'mocha';
 import { createSandbox } from 'sinon';
 
+import ActivityTypes from '../../../../../server/constants/activities';
 import roles from '../../../../../server/constants/roles';
 import { idEncode, IDENTIFIER_TYPES } from '../../../../../server/graphql/v2/identifiers';
 import emailLib from '../../../../../server/lib/email';
@@ -93,8 +94,17 @@ describe('MemberInvitationMutations', () => {
         user1,
       );
 
+      const activity = await models.Activity.findAll({
+        where: {
+          type: ActivityTypes.COLLECTIVE_CORE_MEMBER_INVITED,
+          CollectiveId: collective1.id,
+        },
+      });
       const createdMemberInvitation = result.data.inviteMember;
+
       expect(result.errors).to.not.exist;
+      expect(activity.length).to.equal(1);
+      expect(activity[0].data.memberCollective.id).to.equal(user2.id);
       expect(createdMemberInvitation.role).to.equal(roles.MEMBER);
       expect(createdMemberInvitation.description).to.equal('new user 2 as MEMBER');
       expect(createdMemberInvitation.since.toISOString()).to.equal(new Date('01 January 2022').toISOString());
@@ -362,6 +372,14 @@ describe('MemberInvitationMutations', () => {
         },
         user3,
       );
+      const activity = await models.Activity.findAll({
+        where: {
+          type: ActivityTypes.COLLECTIVE_CORE_MEMBER_ADDED,
+          CollectiveId: collective1.id,
+        },
+      });
+      expect(activity.length).to.equal(1);
+      expect(activity[0].data.memberCollective.id).to.equal(user3.id);
       expect(result.errors).to.not.exist;
       expect(result.data.replyToMemberInvitation).to.equal(true);
     });
--- a/test/server/graphql/v2/mutation/MemberInvitationMutations.test.ts
+++ b/test/server/graphql/v2/mutation/MemberInvitationMutations.test.ts
@@ -85,7 +85,7 @@ describe('MemberInvitationMutations', () => {
         {
           memberAccount: { id: idEncode(user2.id, IDENTIFIER_TYPES.ACCOUNT) },
           account: { id: idEncode(collective1.id, IDENTIFIER_TYPES.ACCOUNT) },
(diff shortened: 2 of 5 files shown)


server/models/MemberInvitation.js
```diff
@@ -241,7 +241,7 @@ function defineModel() {
       invitation = await MemberInvitation.create({ ...memberParams, CollectiveId: collective.id }, sequelizeParams);
       invitation.collective = collective;
 
-      if ([roles.ACCOUNTANT, roles.ADMIN, roles.MEMBER].includes(this.role)) {
+      if (MEMBER_INVITATION_SUPPORTED_ROLES.includes(memberParams.role)) {
```
snitin315, line 244: 🐞 `this.role` is undefined here, so it wasn't actually creating the `COLLECTIVE_CORE_MEMBER_INVITED` activity.

Betree: Good catch, it's similar to the fix made in https://github.com/opencollective/opencollective-api/pull/7598.

test/server/graphql/v2/mutation/MemberInvitationMutations.test.ts
```diff
@@ -0,0 +1,413 @@
...
+  before(async () => {
+    collective1 = await models.Collective.create(utils.data('collective1'));
+  });
+  before(async () => {
+    collective2 = await models.Collective.create(utils.data('collective2'));
+  });
+  before(() => collective1.addUserWithRole(host, roles.HOST));
+  before(() => collective1.addUserWithRole(user1, roles.ADMIN));
+  before(() => collective2.addUserWithRole(host, roles.HOST));
+  before(() => collective2.addUserWithRole(user2, roles.ADMIN));
```
Betree, line 60: That's very verbose. The latest best practice is to use the helpers from `fake-data`. We can rewrite as something like:

```suggestion
  before(async () => {
    await utils.resetTestDB();
    sandbox = createSandbox();
    sendEmailSpy = sandbox.spy(emailLib, 'sendMessage');
    user = await fakeUser();
    collective = await fakeCollective({ admin: user });
    host = collective.host;
  });
```

I also don't get why we need so many users? If they have different settings, could we make that clearer by using more explicit variable names? Example:

```es6
collectiveAdminUser = await fakeUser();
hostAdminUser = await fakeUser();
```

test/server/graphql/v2/mutation/MemberInvitationMutations.test.ts
```diff
@@ -0,0 +1,413 @@
...
+      const result = await utils.graphqlQueryV2(
+        inviteMemberMutation,
+        {
+          memberAccount: { id: idEncode(user3.id, IDENTIFIER_TYPES.ACCOUNT) },
+          account: { id: idEncode(collective1.id, IDENTIFIER_TYPES.ACCOUNT) },
+          description: 'new user 2 as ADMIN',
+          role: roles.MEMBER,
+          since: new Date('01 January 2022').toISOString(),
+        },
+        user2,
```
Betree, line 128: Here, for example, I think we could make the test clearer with a different variable name
```suggestion
        randomUser,
```

test/server/graphql/v2/mutation/MemberInvitationMutations.test.ts
```diff
@@ -0,0 +1,413 @@
...
+      const result = await utils.graphqlQueryV2(
+        inviteMemberMutation,
+        {
+          memberAccount: { id: idEncode(user2.id, IDENTIFIER_TYPES.ACCOUNT) },
+          account: { id: idEncode(collective1.id, IDENTIFIER_TYPES.ACCOUNT) },
+          description: 'new user 2 as MEMBER',
+          role: roles.MEMBER,
+          since: new Date('01 January 2022').toISOString(),
+        },
+        user1,
```
Betree, line 96:
```suggestion
        collectiveAdminUser,
```

test/server/graphql/v2/mutation/MemberInvitationMutations.test.ts
snitin315: Yeah, they have different settings. I've used better naming now and migrated to the `fakeUser()` utility.

test/server/graphql/v2/mutation/MemberInvitationMutations.test.ts
```diff
@@ -0,0 +1,437 @@
...
+
+  describe('inviteMember', () => {
+    it('should create a new member invitation and its related activity', async () => {
+      sendEmailSpy.resetHistory();
+      const randomUserToInvite = await fakeUser();
+      const result = await utils.graphqlQueryV2(
+        inviteMemberMutation,
+        {
+          memberAccount: { id: idEncode(randomUserToInvite.collective.id, IDENTIFIER_TYPES.ACCOUNT) },
+          account: { id: idEncode(collective.id, IDENTIFIER_TYPES.ACCOUNT) },
```
snitin315, line 58: Here `randomUserToInvite` is created with id `4` and there seems to be a collective with type `ORGANIZATION` also present with `id: 4`. And since these are clashing `idEncode(randomUserToInvite.id, IDENTIFIER_TYPES.ACCOUNT)` is actually resolving to the collective with type `ORGANIZATION` and not the `User` one.

I don't know if this s=is limitation of `fakUser` or if it can be possible in real DB as well. So I have used `randomUserToInvite.collective.id` since `fakeUser` creates this with `type: USER`
https://github.com/opencollective/opencollective-api/blob/aaa0b0d03efe4b143f6967614a67d305862dfab0/test/test-helpers/fake-data.ts#L62-L65

Betree: `randomUserToInvite` is a user (`Users`) table. `randomUserToInvite.collective` is its profile/account (`Collectives` table). They're two different things and thus it would make no sense to pass `memberAccount: user`.

Using `memberAccount: user.collective` is the right thing to do here.

test/server/graphql/v2/mutation/MemberInvitationMutations.test.ts
```diff
@@ -0,0 +1,437 @@
...
+      expect(activity.length).to.equal(1);
+      expect(activity[0].data.memberCollective.id).to.equal(randomUserToInvite.collective.id);
+      expect(createdMemberInvitation.role).to.equal(roles.MEMBER);
+      expect(createdMemberInvitation.description).to.equal('new user 2 as MEMBER');
+      expect(createdMemberInvitation.since.toISOString()).to.equal(new Date('01 January 2022').toISOString());
+      expect(sendEmailSpy.callCount).to.equal(1);
+      expect(sendEmailSpy.args[0][1]).to.equal('Invitation to join webpack test collective on Open Collective');
+    });
+    it('must be authenticated as an admin of the collective', async () => {
+      sendEmailSpy.resetHistory();
```
Betree, line 84: :ice_cream: Since we're doing it in almost every function, a good practice might be to put this `resetHistory` in a `beforeEach`

test/server/graphql/v2/mutation/MemberInvitationMutations.test.ts
```diff
@@ -0,0 +1,437 @@
...
+      );
+      expect(sendEmailSpy.callCount).to.equal(0);
+      expect(result.errors).to.have.length(1);
+      expect(result.errors[0].message).to.equal('Only admins can send an invitation.');
+    });
+    it('can only add with role accountant, admin, or member', async () => {
+      // Able to add ACCOUNTANT Role
+      sendEmailSpy.resetHistory();
+      const randomUserToInviteAsAccountant = await fakeUser();
+      const result1 = await utils.graphqlQueryV2(
```
Betree, line 106: :ice_cream: This test could have been shorter and simpler by doing something like:

```es6
[roles.ACCOUNTANT, roles.ADMIN, roles.MEMBER].forEach(role => {
  // Tests for role here
})
```

Betree (approved):
Looking great! I'm going to push some small modifications (mostly, adding a new activity when declining the invitation) but your implementation works well and definitely matches the bounty criteria. Feel free to submit the expense already if you want to :slightly_smiling_face:

test/server/graphql/v2/mutation/MemberInvitationMutations.test.ts
```diff
@@ -0,0 +1,437 @@
...
+      expect(sendEmailSpy.args[0][1]).to.equal('Invitation to join webpack test collective on Open Collective');
+    });
+    it('must be authenticated as an admin of the collective', async () => {
+      sendEmailSpy.resetHistory();
+      const randomUser = await fakeUser();
+      const anotherRandomUser = await fakeUser();
+      const result = await utils.graphqlQueryV2(
+        inviteMemberMutation,
+        {
+          memberAccount: { id: idEncode(anotherRandomUser.id, IDENTIFIER_TYPES.ACCOUNT) },
```
Betree, line 90: Actually related to https://github.com/opencollective/opencollective-api/pull/7616/files#r893101351, this should be: 

```suggestion
          memberAccount: { id: idEncode(anotherRandomUser.CollectiveId, IDENTIFIER_TYPES.ACCOUNT) },
```

Or

```suggestion
          memberAccount: { id: idEncode(anotherRandomUser.collective.id, IDENTIFIER_TYPES.ACCOUNT) },
```

Same for the tests below. I'm including that in my upcoming commit, let me know if it doesn't make sense to you

snitin315: It makes sense now. I got confused with the two tables `Users` and `Collective` (with type User).

Betree: `Collectives` is a confusing table name that should really be named `Accounts` or `Profiles`. Because of its wide usage, it's difficult to rename, but we'll get there someday!

PR #7622, opened 2022-06-07
CurrencyExchangeRate: Add DB index

Betree:
Spotted while working on a related query, this should greatly improve performance. A simple query such as the following one would take 13s:

```sql
SELECT DISTINCT "from", "to"
FROM "CurrencyExchangeRates"
WHERE "from" = 'USD'
```

Also included a test as I was curious about the results of `getMany`


kewitz (approved):
Thanks for adding tests! LGTM, :shipit:!

znarf (approved):

PR #7623, opened 2022-06-07
VirtualCards: Ensure vendor slug is also based on name

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5583

### DIFF
--- a/server/paymentProviders/utils.js
+++ b/server/paymentProviders/utils.js
@@ -5,6 +5,7 @@ import ExpenseType from '../constants/expense_type';
 import { TransactionKind } from '../constants/transaction-kind';
 import { TransactionTypes } from '../constants/transactions';
 import { getFxRate } from '../lib/currency';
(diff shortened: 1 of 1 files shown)


server/paymentProviders/utils.js
```diff
@@ -235,15 +236,20 @@ export const persistTransaction = async (virtualCard, transaction) => {
 
 export const getOrCreateVendor = async (vendorProviderId, vendorName) => {
   const slug = vendorProviderId.toString().toLowerCase();
+  const uniqueSlug = crypto.hash(`${slug}${vendorName}`);
```
znarf, line 239: Could we try to have slugs that are a bit more human friendly such as:

`slugify(vendorName)-whateverHash`

kewitz: We can. But keep in mind that we do not render or expose vendor slugs anywhere.

PR #7626, opened 2022-06-08
test: add tests for member mutations

ersachin3112:
Fix https://github.com/opencollective/opencollective/issues/5604


test/server/graphql/v2/mutation/MemberMutations.test.ts
```diff
@@ -0,0 +1,268 @@
...
+    });
+    it('can only update role to accountant, admin or member', async () => {
+      [roles.ADMIN, roles.MEMBER, roles.ACCOUNTANT, roles.HOST].forEach(async role => {
+        const result = await utils.graphqlQueryV2(
+          editMemberMutation,
+          {
+            memberAccount: { id: idEncode(collectiveMemberUser.CollectiveId, IDENTIFIER_TYPES.ACCOUNT) },
+            account: { id: idEncode(collective.id, IDENTIFIER_TYPES.ACCOUNT) },
+            description: `role changed to ${role}`,
+            role: roles.ADMIN,
```
Betree, line 112: We intend to test whether we can change the role to `role` here

```suggestion
            role,
```

test/server/graphql/v2/mutation/MemberMutations.test.ts
```diff
@@ -0,0 +1,268 @@
...
+          account: { id: idEncode(collective.id, IDENTIFIER_TYPES.ACCOUNT) },
+          description: 'invite user to delete invitation test case',
+          role: roles.MEMBER,
+          isInvitation: true,
+        },
+        collectiveAdminUser,
+      );
+
+      expect(result.errors).to.not.exist;
+      expect(result.data.removeMember).to.equal(true);
```
Betree, line 188: As per the test title, we need to make sure that the invitation gets deleted here. Possible solution:

```suggestion
      expect(result.data.removeMember).to.equal(true);
      await invitation.reload({ paranoid: false });
      expect(invitation.deletedAt).to.not.be.null;
```

Betree (commented):
Looks great! Some problems to fix below, but we're close.

test/server/graphql/v2/mutation/MemberMutations.test.ts
ersachin3112: @Betree resolved.

Betree (approved):
Looking great, merging!

PR #7631, opened 2022-06-09
Change search by slug logic to "startsWith" instead of "contains"

SudharakaP:
This changes the search by slug (which we do by prefixing the search term by `@` sign) to `startWith` the given string instead of returning everything that `contains` the given search string. 

Related to comment; https://github.com/opencollective/opencollective/issues/5434#issuecomment-1098832658
 
image

Previously we had,

image


Betree (approved):
Works for me if the tests pass

PR #7638, opened 2022-06-13
Email Subscriptions: Type Models and Libraries

kewitz:


server/models/Activity.ts
```diff
@@ -1,10 +1,28 @@
+import { CreationOptional, InferAttributes, InferCreationAttributes } from 'sequelize';
+
+import ActivityTypes from '../constants/activities';
 import notify from '../lib/notifications';
-import sequelize, { DataTypes } from '../lib/sequelize';
+import sequelize, { DataTypes, Model } from '../lib/sequelize';
+
+export class Activity extends Model<InferAttributes<Activity>, InferCreationAttributes<Activity>> {
```
Betree, line 7: Great that you went with the `InferAttributes` approach directly!

Betree (approved):
Nice cleanup!

PR #7650, opened 2022-06-15
InviteMembers: use transaction and add test

Betree:
Followup on https://github.com/opencollective/opencollective-api/pull/7511

Create collective form was broken when inviting new admins because we were not passing the SQL transaction down.

I also made a few changes:
- Add test for `applyToHost`
- Add test for `createCollective`
- `processInviteMembersInput`: make the function signature clearer by:
  - making `collective` a positional param (since it's required)
  - making `inviteMemberInputs` a positional param that matched the GraphQL `MemberInvitationInput`
  - Moving everything else to `options`

image

image

### DIFF
--- a/server/models/MemberInvitation.js
+++ b/server/models/MemberInvitation.js
@@ -253,16 +253,19 @@ function defineModel() {
       invitation.collective = collective;
 
       if (MEMBER_INVITATION_SUPPORTED_ROLES.includes(memberParams.role)) {
-        const member = await models.Collective.findByPk(memberParams.MemberCollectiveId);
-        await models.Activity.create({
-          type: ActivityTypes.COLLECTIVE_CORE_MEMBER_INVITED,
-          CollectiveId: collective.id,
-          data: {
-            notify: false,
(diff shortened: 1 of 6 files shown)


server/models/MemberInvitation.js
```diff
@@ -253,16 +253,19 @@ function defineModel() {
...
       if (MEMBER_INVITATION_SUPPORTED_ROLES.includes(memberParams.role)) {
-        const member = await models.Collective.findByPk(memberParams.MemberCollectiveId);
-        await models.Activity.create({
-          type: ActivityTypes.COLLECTIVE_CORE_MEMBER_INVITED,
-          CollectiveId: collective.id,
-          data: {
-            notify: false,
-            memberCollective: member.activity,
-            collective: collective.activity,
+        const memberCollective = await models.Collective.findByPk(memberParams.MemberCollectiveId, sequelizeParams);
```
Betree, line 256: Passing `sequelizeParams` to this one and the `Activity.create` is the root fix

PR #7657, opened 2022-06-16
Fix SCHEDULE_FOR_PAYMENT 2FA bug

kewitz:


kewitz:
@znarf yes because it is the same mutation, we're only using a different `ACTION` parameter.
I see this as a patch, ultimately I would like to have these scheduled PayPal payouts batches being manually paid as we do with Wise. And in that case, we would move this 2FA verification to the "Pay Batch" action and not here where we're only scheduling.

znarf (approved):
Good for me!

PR #7675, opened 2022-06-23
TransactionBalances materialized view

znarf:
Issue: https://github.com/opencollective/opencollective/issues/3875


server/graphql/v2/interface/Transaction.js
```diff
@@ -548,5 +552,14 @@ export const TransactionFields = () => {
         }
       },
     },
+    balance: {
```
znarf, line 555: should we name it `balanceInHostCurrency` ?

PR #7683, opened 2022-06-25
feat: add updated_at and published_at in DateTimeField enum

danielsum:
Resolve [#5690](https://github.com/opencollective/opencollective/issues/5690) 

# Description

This is a very small fix to order updates by publishedAt instead of createdAt (and in the meantime I added also UpdatedAt in case of).

This need to be merge, tested and synchronized with a pull request in the Frontend repository : [https://github.com/opencollective/opencollective-frontend/pull/7944](https://github.com/opencollective/opencollective-frontend/pull/7944)

# Screenshots

Not applicable


server/graphql/v2/enum/DateTimeField.js
```diff
@@ -8,5 +8,13 @@ export const DateTimeField = new GraphQLEnumType({
       value: 'createdAt',
       description: 'The creation time of a resource',
     },
+    UPDATED_AT: {
+      value: 'updatedAt',
+      description: 'The updated time of a resource',
+    },
+    PUBLISHED_AT: {
+      value: 'publishedAt',
+      description: 'The published time of a resource (if available)',
```
znarf, line 17:
```suggestion
      description: 'The publication time of a resource (if available)',
```

server/graphql/v2/enum/DateTimeField.js
```diff
@@ -8,5 +8,13 @@ export const DateTimeField = new GraphQLEnumType({
       value: 'createdAt',
       description: 'The creation time of a resource',
     },
+    UPDATED_AT: {
+      value: 'updatedAt',
+      description: 'The updated time of a resource',
```
znarf, line 13:
```suggestion
      description: 'The update time of a resource',
```

znarf:
@danielsum Welcome! I'm not sure what's wrong with the tests.

I added you to the "Contributors" team so you can trigger CI  properly.

danielsum:
Ok I can have a look this evening but it seems related to other testing parts : https://github.com/opencollective/opencollective-api/runs/7069827142?check_suite_focus=true

I think all the tokens related to payment should be mocked during testing or at least we can trigger the stripe-cli or ngrok inside the the testing script to mock a webhook calls. Other solution will be a script to refresh the tokens needed but I think relying on that my break at the Payment Provider Side and slow down the testing.

znarf:
@danielsum could you rebase and push to our repository?

I don't see why your change would trigger such error in CI.

danielsum:
Ok I think I understand the issue, I suppose that you have some .env that I didn't have. So maybe the best will be to simply commit and push from the original repository (Open Collective). I've rebase everything from the main and push but typically the error looks like some missing .env informations for me.

It's basically just one file to update : DateTimeField.js so I think it will be easier to merge request from the initial remote origin. Have a nice day.

server/graphql/v2/enum/DateTimeField.js
```diff
@@ -8,5 +8,13 @@ export const DateTimeField = new GraphQLEnumType({
       value: 'createdAt',
       description: 'The creation time of a resource',
     },
+    UPDATED_AT: {
+      value: 'updatedAt',
+      description: 'The update time of a resource',
+    },
+    PUBLISHED_AT: {
+      value: 'publishedAt',
```
Betree, line 16: The approach looks good, my only concern is that we'll get an ugly Sequelize error whenever we try to fetch by `PUBLISHED_AT` on a model that doesn't support it (e.g. `expenses` query). Ideally, I'd love us to be clear on the supported types rather than throwing errors. Something like:

```es6
orderBy: {
  type: new ChronologicalOrderInput({ fields: ['createdAt', 'publishedAt'] }),
},
```

danielsum: Ok @Betree , I have created a new ChronologicalWithPublishedAtOrderInput type with a new DateTimeWithPublishedAtField enum. It will be applied only on the Account.updates to prevent any error  => will it be clear enough for you ? I can adapt the name of the input or enum to follow your own convention. Just let me know and I will rename the new input and enum file 👍

Betree: Personal preference, I would use names to make it explicit that these types are for `Updates`:
- `UpdateChronologicalOrderInput` (input)
- `UpdateDateTimeField` (enum) 

As you mentioned during the call, a shared interface could be a nice to have!

Betree:
Closing in favor of https://github.com/opencollective/opencollective-api/pull/7818

PR #7693, opened 2022-06-29
Archive children

znarf:
Follow up from: https://github.com/opencollective/opencollective-api/pull/7426


Betree (approved):
Looks good! Would have been great to have that in an SQL transaction or as a unique `UPDATE` query to make sure we can't end up with partially archived networks if one of them fails. But that can be treated as a follow-up enhancement.

Betree (commented):
@hdiniz Is implementing a different solution in https://github.com/opencollective/opencollective-api/pull/8002/files#diff-f35d6394d656027361220f1c2baea5ee1fa2ff0e027ac8e6d9b3fb6d2541b86dR516-R517: by re-using `changeHost`, the recursive will be mostly be being taken care of.

We should then set `deactivatedAt` on all children, but that can be done with one really simple SQL query:

```es6
await models.Collective.update(
  { deactivatedAt: parent.deactivatedAt }, 
  { where: { ParentCollectiveId: parent.id, deactivatedAt: null }
})
```

The tests from this PR are nice though, we should make sure we cherry-pick them.

PR #7701, opened 2022-06-30
Correct undefined errors when returning `Application` fields

SudharakaP:
If the user is not logged in there's a possibility of erroring out in the API when we try to obtain the `Application` fields. This happens for example in the sign-in pages when we try to obtain the application account details. 

Related to https://github.com/opencollective/opencollective-frontend/pull/7966


znarf (changes requested):
I don't understand what you are trying to achieve there.

Why are you not making `clientId` and `redirectUri` public ?

SudharakaP:
> I don't understand what you are trying to achieve there.
> 
> Why are you not making `clientId` and `redirectUri` public ?

At first I thought all we need in the front-end for the purpose of showing the OAuth sign-in flow is to just not error out when returning these fields from the API (returning `undefined` is okay) as these returned values (`clientID` and `redirectURI`) are only used within the admin pages not within the sign-in flow. And hence the null checks, I thought was a minimal change rather than making the fields unnecessarily public. 

However thinking further I think making the public makes more sense given we these two fields are not private information. 🤔 I've went ahead and made them public.

SudharakaP:
> > as these returned values (clientID and redirectURI) are only used within the admin pages not within the sign-in flow.
> 
> Why would we fetch them in `applicationQuery` if they're not used by the Sign In flow ? That does not make sense.

@znarf: Note that these are passed and used in the `ApplicationApproveScreen.js`. Here the user is already logged in. 

https://github.com/opencollective/opencollective-frontend/blob/08e18be9ef2275aadaeed8be7c91e540f6f8fb38/pages/oauth/authorize.js#L80-L81

https://github.com/opencollective/opencollective-frontend/blob/08e18be9ef2275aadaeed8be7c91e540f6f8fb38/components/oauth/ApplicationApproveScreen.js#L38-L45

`authorize.js` renders the `SignInOrJoinFree` or the `ApplicationApproveScreen` conditionally. So the sign-in flow is not always triggered.

znarf (changes requested):
Tests are not passing, I should not have to tell this.

znarf:
I now understand but the way it was formulated was impossible to figure initially "as these returned values (clientID and redirectURI) are only used within **the admin pages**".

SudharakaP:
> Tests are not passing, I should not have to tell this.

Sorry, I will make sure to double check this in the future for all PRs. 

> I now understand but the way it was formulated was impossible to figure initially "as these returned values (clientID and redirectURI) are only used within **the admin pages**".

I need to find a way to better explain these stuff. Sorry for the confusion.

znarf:
Making `req.remoteUser` optional was actually a good thing, too bad you removed this change.

PR #7712, opened 2022-07-05
Implement OAuth scope

znarf:


server/graphql/common/conversations.ts
```diff
@@ -14,21 +15,29 @@ interface CreateConversationParams {
...
 }
 
+export const checkRemoteUserCanUseConversations = req => {
+  if (!req.remoteUser) {
+    throw new Unauthorized('You need to be logged in to manage conversations');
+  }
+  if (!canUseFeature(req.remoteUser, FEATURE.CONVERSATIONS)) {
+    throw new FeatureNotAllowedForUser();
+  }
+  if (req.userToken && !req.userToken.getScope().includes('conversations')) {
```
Betree, line 25: :ice_cream: 

```suggestion
  if (req.userToken && !req.userToken.hasScope('conversations')) {
```

server/graphql/v2/interface/Account.js
```diff
@@ -780,10 +803,11 @@ export const AccountFields = {
...
       },
     },
-    description: 'The list of payment methods that this collective can use to pay for Orders',
+    description:
+      'The list of payment methods that this collective can use to pay for Orders. Admin only. Scope: "orders".',
     async resolve(collective, args, req) {
-      if (!req.remoteUser?.isAdminOfCollective(collective)) {
-        return [];
+      if (!req.remoteUser?.isAdminOfCollective(collective) || !checkScope(req, 'paymentMethods')) {
+        return null;
```
Betree, line 810: Returning `null` while `type` says `GraphQLNonNull` will generate an ugly error. It sounds like the right thing to do though, so I'd remove the `GraphQLNonNull`.

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -80,6 +81,12 @@ const accountMutations = {
           throw new Forbidden();
         }
 
+        if (!req.remoteUser.isAdminOfCollective(account)) {
+          checkRemoteUserCanUseHost(req);
+        } else {
+          checkRemoteUserCanUseAccount(req);
+        }
```
Betree, line 88: I don't really get this condition, why check `canUseHost` if admin and `canUseAccount` if not an admin?

server/lib/oauth/model.ts
```diff
@@ -179,8 +182,18 @@ const model: OauthModel = {
   // -- Scope --
 
   async verifyScope(token: Token, scope: string | string[]): Promise<boolean> {
+    debug('model.verifyScope', token, scope);
+
     return true; // Scope verification is not implemented yet, but it's required by the library
   },
+
+  /*
+  async validateScope(user: typeof models.User, client: Client, scope: string | string[]): Promise<string | string[]> {
```
Betree, line 191: Maybe add a comment to explain why we're not implementing it

server/models/OAuthAuthorizationCode.ts
```diff
@@ -73,6 +74,28 @@ OAuthAuthorizationCode.init(
...
+          'paymentMethods',
+          'host',
+          'root',
+          'conversations',
+          'updates',
+          'webhooks',
+          'applications',
+          'connectedAccounts',
+        ),
+      ),
```
Betree, line 96: :ice_cream: 

```suggestion
      type: DataTypes.ARRAY(DataTypes.ENUM(Object.values(OAuthScopes))),
```

server/models/UserToken.ts
```diff
@@ -71,6 +80,28 @@ UserToken.init(
...
+          'paymentMethods',
+          'host',
+          'root',
+          'conversations',
+          'updates',
+          'webhooks',
+          'applications',
+          'connectedAccounts',
+        ),
+      ),
```
Betree, line 102: :ice_cream: Doing that in all models will reduce the number of places to update when editing the scopes, reducing the risk of forgetting one


```suggestion
      type: DataTypes.ARRAY(DataTypes.ENUM(Object.values(OAuthScopes))),
```

server/graphql/v2/mutation/AccountMutations.ts
Betree: That makes sense, I did not check the code above. It's actually clear enough.

server/graphql/v2/enum/OAuthScope.js
```diff
@@ -0,0 +1,52 @@
...
+export const OAuthScope = new GraphQLEnumType({
+  name: 'OAuthScope',
+  description: 'All supported OAuth scopes',
+  values: {
+    [oAuthScopes.email]: {
+      description: 'email: Access your email address.',
+    },
+    [oAuthScopes.incognito]: {
+      description: 'incognito: Access your incognito account.',
+    },
```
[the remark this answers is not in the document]
znarf: Yes, that can be interesting for the documentation effort.

PR #7713, opened 2022-07-05
Refact: Notifications library and notify functions

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5711

Sorry about the diff, I just broke down the whole notifications lib.


server/models/Notification.ts
```diff
@@ -182,6 +184,40 @@ export class Notification extends Model<InferAttributes<Notification>, InferCrea
...
 
+  static async getUnsubscribers(_where: {
+    type?: ActivityClasses | ActivityTypes;
+    CollectiveId?: number | number[];
+    UserId?: number | number[];
+  }) {
+    debug('getUnsubscribers', { _where });
+    const getUsers = notifications => notifications.map(notification => notification.User);
+
+    const include = [{ model: models.User }] as any;
```
Betree, line 195: With:

```es6
import { Includeable } from 'sequelize';
```

```suggestion
    const include = [{ model: models.User }] as Includeable[];
```

Also curious about why we need to cast the `where` below to `any`, there's probably a more elegant solution.

Not a blocker if this is time-consuming.

server/models/Notification.ts
```diff
@@ -182,6 +184,40 @@ export class Notification extends Model<InferAttributes<Notification>, InferCrea
...
+    where.type = flatten([where.type, `${where.type}.for.host`, ...classes]);
+
+    const unsubs = await Notification.findAll({
+      where,
+      include,
+    }).then(getUsers);
+
+    const subs = collective
+      ? await Notification.findAll({
+          where: { ...where, active: true, CollectiveId: [collective.id, collective.ParentCollectiveId] },
```
Betree, line 213: I don't really get this, aren't unsubscribed users represented by `active: false` - in which case this function will never return them?

test/server/lib/recurring-contributions.test.js
```diff
@@ -442,7 +388,7 @@ describe('server/lib/recurring-contributions', () => {
         expect(entry.status).to.equal('failure');
 
         // And then the nextChargeDate is ajusted for two days later
-        expect(order.Subscription.nextChargeDate.getTime()).to.equal(new Date('2018-01-30 0:0').getTime());
+        expect(order.Subscription.nextChargeDate.getTime()).to.equal(moment().startOf('day').add(2, 'days').valueOf());
```
Betree, line 391: I don't understand why we need to update this test, seems unrelated?

Betree (commented):
Great to see all this code migrated to Typescript! That's definitely a place where it adds value.

cron/monthly/collective-report.js
```diff
@@ -210,11 +211,12 @@ const processCollective = async collective => {
...
-        type: 'collective.monthlyreport',
-        data: emailData,
-      };
-      return notifyAdminsOfCollective(collective.id, activity, options);
+      const activity = await models.Activity.create({
+        type: ActivityTypes.COLLECTIVE_MONTHLY_REPORT,
+        CollectiveId: collective.id,
+        data: { ...emailData, notify: false },
+      });
+      return notify.collective(activity, options);
```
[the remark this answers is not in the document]
kewitz: I think consistency and logging. But I agree that, since this is a cron work, it is probably not necessary.

server/models/Notification.ts
kewitz: True. The idea is being able to support exceptions at the collective level, for example:
Unsubscribe from every Transaction notification but subscribe to Transaction notifications of a specific Collective.

test/server/lib/recurring-contributions.test.js
kewitz: I had to remove the `fakeTimer` because it was getting in conflict with `waitForCondition`.

server/models/Notification.ts
```diff
@@ -182,6 +184,47 @@ export class Notification extends Model<InferAttributes<Notification>, InferCrea
...
+    const include = [{ model: models.User }];
+    const where = { active: false };
+    if (_where.UserId) {
+      where['UserId'] = _where.UserId;
+    }
+
+    const collective = _where.CollectiveId && (await models.Collective.findByPk(_where.CollectiveId));
+    if (collective) {
+      where['CollectiveId'] = compact([collective.id, collective.ParentCollectiveId, collective.HostCollectiveId]);
+      // If user is informed, also fetch global settings
```
Betree, line 204: I don't really understand what `informed` means in this context? Overall this entire function is pretty difficult for me to understand, it would really benefit from adding some comments to explain what we're doing at each stage.

kewitz: Sorry about the language. The idea is that if UserId is passed as an argument, we should also consider global notification settings: basically, the setting of an UserId unsubscribe to a whole Activity Type without any CollectiveId passed to it.

kewitz:
Since this is a big refactor, I'll be waiting until Wednesday to merge this since tomorrow I'll probably be off most of the day.

kewitz:
Rebased and fixed conflicts.

PR #7717, opened 2022-07-06
Make the `graphql-inspector` version fixed

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5724

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -374,14 +374,14 @@ jobs:
         uses: actions/checkout@main
 
       - name: GraphQL Inspector (V1)
-        uses: kamilkisiela/graphql-inspector@master
+        uses: kamilkisiela/graphql-inspector@v3.2.0
         with:
           name: 'GraphQL Inspector - Schema v1'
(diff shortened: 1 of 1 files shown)


znarf (approved):
Great, let's do this for now.

znarf (changes requested):
Hold on. It seems master is working fine now.

SudharakaP:
> Hold on. It seems master is working fine now.

Yes that's because after my report they fixed it just some hours back and pushed a new version (v3.2.0). 😉 

https://github.com/kamilkisiela/graphql-inspector/issues/2128#issuecomment-1176314788

I've adjusted this PR to use that new version (v3.2.0).

znarf:
@SudharakaP I'm not sure if we need to fix the version. Will we have the automated dependency updates for it?

@Betree what do you think?

Betree:
> @SudharakaP I'm not sure if we need to fix the version. Will we have the automated dependency updates for it?
> 
> @Betree what do you think?

Dependabot supports updating Github actions. I'm not sure about renovate?

znarf:
Good for me if updates are automatically managed.

PR #7720, opened 2022-07-07
Add details to the "transaction created" webhook event

Betree:
Also added the ability to fetch by transaction group in the `transactions` query


znarf (approved):
Looks good to me.

PR #7735, opened 2022-07-11
Add required gQL Notifications mutators and resolvers

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5712


server/graphql/v2/query/NotificationsQuery.ts
```diff
@@ -0,0 +1,50 @@
...
+      throw new Forbidden('You need to be logged in to query for notifications');
+    }
+
+    const userCollective = await fetchAccountWithReference(args.individual, { throwIfMissing: true });
+    const user = await userCollective.getUser();
+    if (!user) {
+      throw new Error('Individual must be an User account');
+    } else if (user.id !== req.remoteUser.id) {
+      throw new Forbidden('You can only query for your own Notifications');
+    }
```
Betree, line 34: :ice_cream: I would omit the `individual` argument altogether if this is only for currently logged-in user. Or eventually put it on `loggedInUser.notifications` rather than a root query. But that's only a personal preference.

Betree (commented):
I would really like us to review the term "Notification" before merging.

server/graphql/v2/query/index.js
```diff
@@ -38,6 +39,7 @@ const query = {
   hosts: HostsCollectionQuery,
   individual: IndividualQuery,
   memberInvitations: MemberInvitationsQuery,
+  notifications: NotificationsQuery,
```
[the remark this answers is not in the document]
kewitz: We'll rename it to 'ActivitySubscriptions'.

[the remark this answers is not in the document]
Betree: @kewitz Sorry for the late reply, I don't think we should rename our models, I just want us to take this as an opportunity to introduce a better name, kind of what we need by introducing `Accounts` instead of `Collectives`. `ActivitySubscriptions` looks perfect to me.

server/graphql/v2/query/NotificationsQuery.ts
znarf: Root query doesn't seem really useful indeed. Having a property on `individual` should be good.

server/graphql/v2/mutation/NotificationMutations.ts
```diff
@@ -0,0 +1,44 @@
...
+      type: { type: new GraphQLNonNull(NotificationType) },
+      account: {
+        type: AccountReferenceInput,
+        description: 'Scope account which this notification preference is applied to',
+      },
+    },
+    async resolve(_: void, args, req: express.Request) {
+      if (!req.remoteUser) {
+        throw new Unauthorized('You need to be logged in to toggle email notification subscription');
+      }
```
znarf, line 25:
```suggestion
      checkRemoteUserCanUseAccount(req);
```

kewitz:
Closed in favor of https://github.com/opencollective/opencollective-api/pull/7774.
I've implemented most of the reviewed comments.

PR #7737, opened 2022-07-11
Send an email notification when an application is connected through OAuth

hdiniz:
Resolves https://github.com/opencollective/opencollective/issues/5614

# Email for new oauth application authorized
Screenshot from 2022-07-11 14-29-31

# Email for new oauth application authorized with scopes
Screenshot from 2022-07-11 14-28-59

There is a refactor of the notifications flow on https://github.com/opencollective/opencollective-api/pull/7713 close to being merged, when that happens I will rebase this change.

### DIFF
--- a/server/constants/activities.ts
+++ b/server/constants/activities.ts
@@ -61,6 +61,7 @@ enum ActivityTypes {
   WEBHOOK_PAYPAL_RECEIVED = 'webhook.paypal.received',
   COLLECTIVE_MONTHLY = 'collective.monthly',
   ORDER_CANCELED_ARCHIVED_COLLECTIVE = 'order.canceled.archived.collective',
+  OAUTH_APPLICATION_AUTHORIZED = 'oauth.application.authorized',
   ORDER_PROCESSING = 'order.processing',
   ORDER_PROCESSING_CRYPTO = 'order.processing.crypto',
   ORDER_PENDING_CONTRIBUTION_NEW = 'order.new.pendingFinancialContribution',
@@ -97,6 +98,7 @@ export const TransactionalActivities = [
   ActivityTypes.ORDER_PROCESSING_CRYPTO,
   ActivityTypes.PAYMENT_CREDITCARD_EXPIRING,
   ActivityTypes.TAXFORM_REQUEST,
+  ActivityTypes.OAUTH_APPLICATION_AUTHORIZED,
 ];
 
 export enum ActivityClasses {
--- a/templates/emails/oauth.application.authorized.hbs
+++ b/templates/emails/oauth.application.authorized.hbs
@@ -1,0 +1,25 @@
+Subject: Open Collective: A third-party OAuth application has been added to your account
+
+{{> header}}
+
+<br />
+Hi,
+<br />
+<br />
+A third-party OAuth application ({{application.name}}) was recently authorized to access your account.
+
+{{#if scope.length}}
+<br />
+<br />
+Scopes:
+<br />
+<ul>
+    {{#each scope}}
+    <li>{{this}}</li>
+    {{/each}}
+</ul>
+{{/if}}
+<br />
+<br />
+
+{{> footer}}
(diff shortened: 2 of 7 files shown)


server/models/OAuthAuthorizationCode.ts
```diff
@@ -84,6 +85,20 @@ OAuthAuthorizationCode.init(
...
+        models.Activity.create({
+          type: activities.OAUTH_APPLICATION_AUTHORIZED,
+          UserId: oauthAuthorizationCode.UserId,
+          data: {
+            application: application.publicInfo,
+            scope: oauthAuthorizationCode.scope || [],
+          },
+        });
+      },
+    },
```
kewitz, line 101: This is a clever solution, and technically this should work.
At Open Collective we tend to avoid unexplicit side-effects since they rely too much on context and can make debugging difficult. It is also good to write explicit (and sometimes dumb) code because that helps people outside our core-team to contribute.

So, for consistency, I would rather see this activity being created in `lib/oauth/model.ts` inside the `saveAuthorizationCode` function.
If you're up for another challenge, it would also be nice to update that function unit test to check if the activity was created.

server/lib/notifications.ts
```diff
@@ -375,6 +375,10 @@ const populateCommentActivity = async activity => {
 async function notifyByEmail(activity: Activity) {
   debug('notifyByEmail', activity.type);
   switch (activity.type) {
+    case ActivityTypes.OAUTH_APPLICATION_AUTHORIZED:
+      notifyUserId(activity.UserId, activity);
+      break;
```
kewitz, line 380: 💯

server/constants/activities.ts
```diff
@@ -91,6 +92,7 @@ export const TransactionalActivities = [
   ActivityTypes.ORDER_PROCESSING_CRYPTO,
   ActivityTypes.PAYMENT_CREDITCARD_EXPIRING,
   ActivityTypes.TAXFORM_REQUEST,
+  ActivityTypes.OAUTH_APPLICATION_AUTHORIZED,
```
kewitz, line 95: Nice one! 🎯

templates/emails/oauth.application.authorized.hbs
```diff
@@ -0,0 +1,27 @@
...
+    <br />
+    <br />
+    A third-party OAuth application ({{application.name}}) was recently authorized to access your account.
+    
+    {{#if scope.length}}
+    <br />
+    <br />
+    Scopes:
+    <br />
+    <ul>
```
kewitz, line 17: By HTML standards, flow elements shouldn't be used inside the paragraph element.
It is a good idea to move the list element to the root level of this file.

kewitz (changes requested):
Thanks for the contribution! Overall I think you nailed the solution, but I have some contextual information I would like to share in this review.

ps. Getting rid of the `afterCreate` hook in favor of creating the Activity at the library level is the only one I have a strong opinion about and would like to have it changed before I can approve.

server/models/OAuthAuthorizationCode.ts
hdiniz: Thanks @kewitz! I made this change.

templates/emails/oauth.application.authorized.hbs
hdiniz: Thanks for catching this, I updated it.

test/server/lib/oauth/model.test.ts
```diff
@@ -182,6 +183,7 @@ describe('server/lib/oauth/model', () => {
         authorizationCode: randStr(),
         expiresAt: new Date(),
         redirectUri: 'https://test.com',
+        scope: ['email'],
```
znarf, line 186: Why have you added `scope` here?

hdiniz: To validate below that the activity is logged correctly with the application and scopes authorized by the user

kewitz (approved):

Betree:
Moved to our main repo: https://github.com/opencollective/opencollective-api/pull/7804

PR #7740, opened 2022-07-11
Add Slight Adjustments to Styles in Custom Email Template Body

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/2682

⚠️ Requires: https://github.com/opencollective/opencollective-frontend/pull/8019

I noticed that there were some slight discrepancy in the styles of the custom email template string. What we have currently,

image

In this PR I've changed this to the following which I think better represents what's given in the [design preview](https://www.figma.com/file/ZQBMWhnGGtRWeIZknFW1eP/%5BOC-Design%5D-Production-Ready-%E2%9C%85?node-id=17991%3A193953),

image

### DIFF
--- a/templates/emails/thankyou.hbs
+++ b/templates/emails/thankyou.hbs
@@ -30,7 +30,8 @@ Subject: Thank you for your contribution to {{collective.name}}
 {{#if customMessage}}
 <p>A message from {{collective.name}}:</p>
 
-<div style="color: #494B4D; font-size: 14px; line-height: 18px; padding-left: 10px; border-left: 5px solid #F1F2F3">{{{customMessage}}}</div>
+<img width="40" height="33" src="{{config.host.website}}/static/images/email/quotation-custom-email.png" />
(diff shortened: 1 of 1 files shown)


templates/emails/thankyou.hbs
```diff
@@ -29,7 +29,7 @@ Subject: Thank you for your contribution to {{collective.name}}
 
 <p>A message from {{collective.name}}:</p>
 
-<p>{{{customMessage}}}</p>
+<div style="color: #494B4D; font-size: 14px; line-height: 18px; padding-left: 10px; border-left: 5px solid #F1F2F3">{{{customMessage}}}</div>
```
znarf, line 32: In a recent review you told me there was no need to put a condtion before displaying `customMessage` because it would be invisible anyway. This is not the case anymore with this change.

znarf (changes requested):

templates/emails/thankyou.hbs
SudharakaP: ops, sorry I seem to have missed this condition. 😢 I've added this.

SudharakaP:
@znarf: I've also update the PR with adding the quotation mark as per the design;

image

znarf (approved):

PR #7743, opened 2022-07-12
Account permissions resolver

Betree:
For https://github.com/opencollective/opencollective/issues/5658#issuecomment-1181678094

Similarily to the `expense.permissions` and `order.permissions` fields, this PR adds an `account.permissions` field to check remove user's permissions on this account.

It uses the latest pattern implemented by @kewitz where we return an object like `{ allowed, reason }` to optionally add more details when features are not accessible.


server/graphql/v2/interface/Account.js
```diff
@@ -854,6 +855,11 @@ export const AccountFields = {
     },
   },
   webhooks: accountWebhooks,
+  permissions: {
+    type: new GraphQLNonNull(AccountPermissions),
+    description: 'Logged-in user permissions on an account',
+    resolve: collective => collective, // Individual resolvers in `AccountPermissions`
```
kewitz, line 861: Just so you know, lodash has an `identity` function you can import.

kewitz (approved):

PR #7755, opened 2022-07-15
Add support to show multi currency expense amounts in emails

SudharakaP:
https://github.com/opencollective/opencollective/issues/5272

Screenshot 2022-07-15 at 10-58-33 MailDev ( 11)

And if the expense use the same currency as host,

image


Betree (approved):
Haven't tested but the code looks good. I'd recommend testing with a collective that has a different currency than its host, both with a multi-currency expense and a currency in the collective currency if not done already.

PR #7761, opened 2022-07-18
Cleanup GraphQL loaders

Betree:
- Remove unused 
- Add `cacheKeyFn` when necessary

### DIFF
--- a/server/graphql/loaders/index.js
+++ b/server/graphql/loaders/index.js
@@ -5,9 +5,7 @@ import moment from 'moment';
 
 import orderStatus from '../../constants/order_status';
 import { TransactionTypes } from '../../constants/transactions';
-import { getListOfAccessibleMembers } from '../../lib/auth';
 import { getBalances, getBalancesWithBlockedFunds } from '../../lib/budget';
-import { reportErrorToSentry } from '../../lib/sentry';
 import models, { Op, sequelize } from '../../models';
 
 import collectiveLoaders from './collective';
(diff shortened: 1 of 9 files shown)


server/graphql/loaders/index.js
```diff
@@ -239,43 +216,6 @@ export const loaders = req => {
     ),
   };
 
-  // @deprecated Getting orgs emails by `CreatedByUserId` is unreliable. See https://github.com/opencollective/opencollective/issues/3415
-  context.loaders.getOrgDetailsByCollectiveId = new DataLoader(OrgCollectiveIds =>
```
Betree, line 243: Removing this one, as the comment says it was unreliable. We now have `Account.emails` on GQLV2 for this purpose

server/graphql/loaders/index.js
```diff
@@ -591,21 +525,6 @@ export const loaders = req => {
...
-              OrderId: { [Op.in]: OrderIds },
-              ...options.where,
-            },
-            order: [['createdAt', 'DESC']],
-          }).then(results => sortResults(OrderIds, results, 'OrderId', []));
-        },
-        cache,
-        options,
-        'transactions',
-      ),
```
Betree, line 608: There's already a `Transaction.byOrderId` above, and this one was not working since each call was instantiating a new `DataLoader`, defeating the purpose of using one.

server/graphql/v1/CollectiveInterface.js
```diff
@@ -1997,14 +1997,8 @@ export const OrganizationCollectiveType = new GraphQLObjectType({
...
       email: {
         type: GraphQLString,
-        async resolve(orgCollective, args, req) {
-          if (!req.remoteUser) {
-            return null;
-          }
-          return (
-            orgCollective && req.loaders.getOrgDetailsByCollectiveId.load(orgCollective.id).then(user => user.email)
-          );
-        },
```
Betree, line 2007: These changes on `email` are the only thing breaking. I've opened a discussion in the only place that depends on it and will rollback if it's problematic: https://github.com/opencollective/opencollective-rest/pull/451

server/graphql/v1/types.js
```diff
@@ -1693,16 +1693,11 @@ export const OrderType = new GraphQLObjectType({
...
-          const query = {
-            where: {},
-            limit: args.limit || 10,
-            offset: args.offset || 0,
-          };
-          if (args.type) {
-            query.where.type = args.type;
-          }
-          return req.loaders.Transaction.findByOrderId(query).load(order.id);
+        async resolve(order, args, req) {
```
Betree, line 1696: Aligned this with GQLV2. As mentioned, the previous data loader was not working. We're only using this on the confirm order page, so performance shouldn't be an issue.

PR #7764, opened 2022-07-19
Fix: helloworks test fails if aws key is not set

hdiniz:
The import `s3` is undefined if the aws key is not set in the awsS3 module, causing these test cases to fail:

e.g: https://github.com/opencollective/opencollective-api/runs/7399682937?check_suite_focus=true#step:9:1291

With this change, we stub the lib function uploadToS3 instead which works regardless of the environment configuration.

### DIFF
--- a/test/server/controllers/helloworks.test.js
+++ b/test/server/controllers/helloworks.test.js
@@ -1,10 +1,10 @@
 import { expect } from 'chai';
 import config from 'config';
 import { get } from 'lodash';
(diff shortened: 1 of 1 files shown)


znarf (approved):
Looks good to me. @Betree for another review.

kewitz (approved):

Betree (approved):
Perfect!

WikiRik:
I believe this fixes https://github.com/opencollective/opencollective/issues/5258

PR #7765, opened 2022-07-19
Add custom email message for event tickets

SudharakaP:
We have missed adding the custom email block for ticket confirmation emails. This PR corrects it. 

Resolve https://github.com/opencollective/opencollective/issues/5760

image


znarf (approved):

server/lib/payments.js
```diff
@@ -613,6 +613,7 @@ const sendOrderConfirmedEmail = async (order, transaction) => {
   const user = await order.getUserForActivity();
   const host = await collective.getHostCollective();
   const parentCollective = await collective.getParentCollective();
+  const customMessage = collective.settings?.customEmailMessage || parentCollective?.settings?.customEmailMessage;
```
Betree, line 616: I'm personally not a fan of this variable name (`collective.settings.customEmailMessage`), it doesn't tell you which email this is for and you'd be tempted to think it's actually intended to add a message to all your communications.

Given that we may be adding other custom messages in the future, I'd recommend something like:
```es6
settings = {
  customEmailMessages: {
    thankYou: '<...>'
  }
}
```

SudharakaP: This is handled at, https://github.com/opencollective/opencollective/issues/5780

PR #7768, opened 2022-07-20
Add Accountants as recipients of the Collective Report

Betree:
Realized in https://github.com/opencollective/opencollective/issues/5755 that we never added accountants as recipients of this email. 

@kewitz tagging you on the review to make sure I'm using this in the right way :wink:

### DIFF
--- a/cron/monthly/collective-report.js
+++ b/cron/monthly/collective-report.js
@@ -6,6 +6,7 @@ import config from 'config';
 import { omit, pick } from 'lodash';
 import moment from 'moment';
 
+import { roles } from '../../server/constants';
 import ActivityTypes from '../../server/constants/activities';
 import { TransactionKind } from '../../server/constants/transaction-kind';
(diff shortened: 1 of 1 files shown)


cron/monthly/collective-report.js
```diff
@@ -142,7 +143,7 @@ const processCollective = async collective => {
   ];
 
   let emailData = {};
-  const options = { attachments: [] };
+  const options = { attachments: [], roles: [roles.ADMIN, roles.ACCOUNTANT] };
```
kewitz, line 146: The property is in the singular because it matches the query interface of the model `role` property.
```suggestion
  const options = { attachments: [], role: [roles.ADMIN, roles.ACCOUNTANT] };
```

Shouldn't TS catch this?

Betree: Good catch, updating it now. TS wouldn't catch this because we're in a JS file

PR #7769, opened 2022-07-20
Fix check for collective host in add funds permission

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5658

Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/8035


server/graphql/v2/object/AccountPermissions.ts
```diff
@@ -18,7 +18,7 @@ const AccountPermissions = new GraphQLObjectType({
       type: new GraphQLNonNull(Permission),
       description: 'Whether the current user can mark this order as expired',
       resolve(collective, _, req: express.Request): PermissionFields {
-        return { allowed: checkScope(req, 'host') && req.remoteUser.isHostAdmin(collective) };
```
SudharakaP, line 21: Such a function `isHostAdimin` seems not to exist. Therefor I am suggesting this alternative approach. 

cc: @Betree

server/graphql/v2/object/AccountPermissions.ts
```diff
@@ -18,7 +18,7 @@ const AccountPermissions = new GraphQLObjectType({
...
+        return { allowed: (checkScope(req, 'host') && req.remoteUser?.isAdmin(collective?.HostCollectiveId)) || false };
```
Betree, line 21: Collective always exists in this context

```suggestion
        return { allowed: (checkScope(req, 'host') && req.remoteUser?.isAdmin(collective.HostCollectiveId)) || false };
```

PR #7771, opened 2022-07-20
Allow host to turn on or off expense types for collectives

danielsum:
Part of [https://github.com/opencollective/opencollective/issues/5525](https://github.com/opencollective/opencollective/issues/5525)

# Description

Here are the PR related to this issue : https://github.com/opencollective/opencollective/issues/5525

As discussed, the options to override the expenseTypes setting at the collective level is on stand by but I have already handled that at the Api level. I have also added a graphql test to test wrong expense type post.

# Screenshots

dad_-_New_expense_-_Open_Collective


server/graphql/common/expenses.ts
```diff
@@ -865,12 +867,56 @@ export async function createExpense(
...
+        throw new FeatureNotSupportedForCollective('Expense of type receipt has been disabled by the collective');
+      } else if (
+          !isBoolean(collective.settings?.expenseTypes?.hasReceipt)
+          && collective.host.settings?.expenseTypes?.hasReceipt === false
+      ) {
+        throw new FeatureNotSupportedForCollective('Expense of type receipt has been disabled by the host');
+      }
+      break
+  }
+
```
Betree, line 919: From what I understand, we only cover the `createExpense` mutation here. As mentioned during the call, we want to cover the `editExpense` one as well **when changing the type of the expense**. In other words, editing an expense that has an invalid type shouldn't block, but changing the type of an expense from `RECEIPT` to `GRANT` when grants are forbidden shouldn't be allowed.

Can we:
1. Move all this code to a `checkExpenseType` helper
2. Re-use it in `editExpense`
3. Bonus: Refactor the code to reduce the duplication

test/server/graphql/v2/mutation/ExpenseMutations.test.js
```diff
@@ -224,6 +224,19 @@ describe('server/graphql/v2/mutation/ExpenseMutations', () => {
...
+      const user = await fakeUser();
+      const collective = await fakeCollective({ settings: { expenseTypes: { hasGrant: false } } });
+      const payee = await fakeCollective({ type: 'ORGANIZATION', admin: user.collective, address: null });
+      const expenseData = { ...getValidExpenseData(), payee: { legacyId: payee.id } };
+      const result = await graphqlQueryV2(
+        createExpenseMutation,
+        { expense: { ...expenseData, type: 'GRANT' }, account: { legacyId: collective.id } },
+        user,
+      );
+      expect(result.errors).to.exist;
```
Betree, line 237: Would be great to check the error type or message here, to make sure it's failing for the right reason

Betree (commented):
The code looks clean, good job! Linting issues to fix (see CI). Also added some comments below on things we need to iterate on.

@danielsum The `Resolve` in the PR description is wrong, can it be updated?

Betree:
Resolved the two items above, added tests, and completed the feature by also looking for the flag on the parent collective. We'll need to implement this one in the frontend too.

danielsum:
Hi @Betree , yes for the testing I didn't have the time to finish the testing in the afternoon as in the morning I had this overheat cpu issue and have to join some friends in the afternoon. But let me know if you want me to finish the PR 👍

Betree (approved):
@danielsum It's all good! The PR is ready; I'll look at the frontend part sometime this week.

Betree:
Introduced a new `supportedExpenseTypes` field to have a generic way to check this & centralize the logic in the API

kewitz (approved):
Looking good! :shipit:

PR #7773, opened 2022-07-21
Fix check for minimum number of admins when removing

Betree:


server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -229,7 +229,7 @@ const memberMutations = {
           });
 
           const policy = getPolicy(host, POLICIES.COLLECTIVE_MINIMUM_ADMINS);
-          if (policy?.numberOfAdmins <= adminCount) {
+          if (policy?.numberOfAdmins && adminCount <= policy.numberOfAdmins) {
```
Betree, line 232: The check was reversed here (`minNumberOfAdmins <= adminCount`). I also added a `policy?.numberOfAdmins` before; in my experience it's safer to separate the nullish condition check because:

```es6
$ undefined < 5
false
$ null < 5
true
```

znarf (approved):

PR #7774, opened 2022-07-21
Email Notifications: Objects, resolvers and mutations

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5712


server/graphql/v2/interface/Account.js
```diff
@@ -595,6 +597,26 @@ const accountFieldsDefinition = () => ({
...
+  activitySubscriptions: {
+    type: new GraphQLNonNull(new GraphQLList(ActivitySubscriptions)),
+    description: 'List of activities that the logged-in user is subscribed for this collective',
+    args: {
+      channel: {
+        type: ActivityChannel,
+      },
+    },
+    async resolve(collective, args, req) {
+      if (!req.remoteUser) {
```
Betree, line 609: scope check

server/graphql/v2/mutation/ActivitySubscriptionsMutations.ts
```diff
@@ -0,0 +1,45 @@
...
+      type: { type: new GraphQLNonNull(ActivityAndClassesType) },
+      account: {
+        type: AccountReferenceInput,
+        description: 'Scope account which this notification preference is applied to',
+      },
+    },
+    async resolve(_: void, args, req: express.Request) {
+      checkRemoteUserCanUseAccount(req);
+
+      const collective =
```
Betree, line 25: Admin permission check?

server/graphql/v2/object/ActivitySubscriptions.ts
```diff
@@ -0,0 +1,59 @@
...
+import { GraphQLDateTime } from 'graphql-scalars';
+
+import { ActivityChannel } from '../enum/ActivityChannel';
+import { getIdEncodeResolver, IDENTIFIER_TYPES } from '../identifiers';
+import { Account } from '../interface/Account';
+
+import { Individual } from './Individual';
+
+export const ActivitySubscriptions = new GraphQLObjectType({
+  name: 'ActivitySubscriptions',
```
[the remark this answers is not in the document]
kewitz: Nice catch, I'll fix this object name.

server/graphql/v2/mutation/ActivitySubscriptionsMutations.ts
kewitz: Not necessary since on email notifications the only affected is the user receiving the email. In this type of activity, the collective is only about the scope of notifications.

server/graphql/v2/interface/Account.js
Betree: Also I'd be tempted to return `null` if no `remoteUser` or doesn't have the right scope, to differentiate with cases where the list is really empty. No blocker.

server/graphql/v2/mutation/ActivitySubscriptionsMutations.ts
```diff
@@ -0,0 +1,43 @@
...
+        args.account && (await fetchAccountWithReference(args.account, { loaders: req.loaders, throwIfMissing: true }));
+
+      const where = { UserId: req.remoteUser.id, type: args.type };
+      if (collective) {
+        where['CollectiveId'] = collective.id;
+      }
```
Betree, line 32: I don't get how/where this `where` is being used?

kewitz: Nice catch, it was legacy from the toggle implementation.

PR #7776, opened 2022-07-22
Remove default bcc

znarf:


server/lib/email.ts
```diff
@@ -125,8 +125,6 @@ const sendMessage = (
   html: string,
   options: SendMessageOptions = {},
 ) => {
-  options.bcc = options.bcc || 'emailbcc@opencollective.com';
```
Betree, line 128: @znarf Should we keep it for staging? I still find it useful to have a place where I can get all staging emails.

znarf: I'm not relying on it personally. We don't send much emails on staging?

Why not look at Mailgun logs instead?

Betree: *Note: this was discussed in a call*

Betree:
Related: I just noticed staging emails are sent from info@opencollective.com, which mixes them with the prod emails in mailgun's logs. @znarf should we change that?

Betree (approved):

PR #7778, opened 2022-07-22
Restrict which accounts can be "from"/"origin" in Added Funds

znarf:
We were surprised this morning to have a Collective as origin of the funds and that created problems because the admin charged Host Fees for the transaction. Trying to be a bit more restrictive here.

Looked at the data and decided to allow for added funds in internal to Collectives, as I discovered this was commonly used today.


Betree (approved):
Looks good, maybe we should couple that with `isTrustedHost` in the future

PR #7779, opened 2022-07-22
Restrict valid cardholders in Stripe issuing

znarf:
As seen this morning, trying to be a bit more selective when fetching cardholders.


Betree:
Merging, I'll follow up if there's a problem

PR #7780, opened 2022-07-22
Add sanitization for custom email settings string

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5776 

image

https://user-images.githubusercontent.com/12435965/180504782-092a0df0-a00a-41fc-a953-658bbc4431da.mp4

### DIFF
--- a/server/graphql/v2/mutation/AccountMutations.ts
+++ b/server/graphql/v2/mutation/AccountMutations.ts
@@ -17,6 +17,7 @@ import plans from '../../../constants/plans';
 import cache from '../../../lib/cache';
 import * as collectivelib from '../../../lib/collectivelib';
 import { crypto } from '../../../lib/encryption';
+import { buildSanitizerOptions, sanitizeHTML } from '../../../lib/sanitize-html';
 import { verifyTwoFactorAuthenticatorCode } from '../../../lib/two-factor-authentication';
 import models, { sequelize } from '../../../models';
 import { checkRemoteUserCanUseAccount, checkRemoteUserCanUseHost } from '../../common/scope-check';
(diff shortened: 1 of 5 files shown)


server/lib/collectivelib.ts
```diff
@@ -148,11 +149,10 @@ export function validateSettings(settings: any): string | boolean {
...
   /*
-   * Validate customEmailMessage length. We are using 511 characters to account
-   * for the "div" tags that enclose the content <div>..<div/>. These divs are auto
-   * added by the RichTextEditor.js
+   * Validate customEmailMessage length.
    */
-  if (settings.customEmailMessage && settings.customEmailMessage.length > 511) {
+  if (settings.customEmailMessage && stripHTML(settings.customEmailMessage).length > 500) {
+    console.log(stripHTML(settings.customEmailMessage));
     return 'Message should be less than 500 characters';
```
Betree, line 156: Would be great to be more explicit on this one.

```suggestion
    return 'Custom "Thank you" email message should be less than 500 characters';
```

server/lib/search.js
```diff
@@ -178,7 +178,7 @@ export const searchCollectivesInDB = async (
   }
 
   if (tags?.length) {
-    searchedTags = `{${tags.map(tag => `"${tag}"`).join(',')}}`;
+    searchedTags = `{${tags.map(tag => `"${tag.toLowerCase()}"`).join(',')}}`;
```
Betree, line 181: I don't get why we're changing this here?

Betree (commented):
Looks like the right approach but there are a few things to resolve still

server/lib/search.js
SudharakaP: Sorry about this one, I was trying this out for a different issue and somehow accidentally checked in.

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -99,7 +115,8 @@ const accountMutations = {
         }
 
         const settings = account.settings ? cloneDeep(account.settings) : {};
-        set(settings, args.key, args.value);
+        const value = sanitizeSettingsValue(args.key, args.value);
```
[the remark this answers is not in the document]
SudharakaP: Ah, makes sense. I've moved it to the Collective mode. 👍🏽

Betree (commented):
Tests are failing because `settings` can be null.

I've added an additional test to make sure the feature works.

SudharakaP:
> Tests are failing because `settings` can be null.
> 
> I've added an additional test to make sure the feature works.

Thanks for the review and the additional check. I've fixed the problem. 👍🏽

Betree (approved):

PR #7795, opened 2022-08-01
Feat/add move expenses root action

hdiniz:
Resolves https://github.com/opencollective/opencollective/issues/3981

Adds a mutation to move expenses to a given destination account.


Betree (commented):
Thanks for your contribution @hdiniz, it looks perfect to me. I'll add a small iteration in the next few days to support an edge case:
> Expenses that have attached transactions are a difficult case to handle. For now, we want to throw an error if any expense in the list has transactions attached.

Betree (approved):

PR #7808, opened 2022-08-05
Make the passed tag lower case before searching

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5702

### DIFF
--- a/server/lib/search.js
+++ b/server/lib/search.js
@@ -177,11 +177,6 @@ export const searchCollectivesInDB = async (
     countryCodes = `${countries.join(',')}`;
   }
 
-  if (tags?.length) {
-    searchedTags = tags.map(tag => tag.toLowerCase());
-    dynamicConditions += `AND c."tags" @> Array[:searchedTags]::varchar[] `;
-  }
-
   if (hostCollectiveIds && hostCollectiveIds.length > 0) {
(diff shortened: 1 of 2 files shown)


server/lib/search.js
```diff
@@ -177,10 +177,6 @@ export const searchCollectivesInDB = async (
     countryCodes = `${countries.join(',')}`;
   }
 
-  if (tags?.length) {
-    searchedTags = `{${tags.map(tag => `"${tag}"`).join(',')}}`;
-  }
-
```
Betree, line 183: Moved this one down, it added unnecessary complexity to have two `if (tags?.length)` checks

server/lib/search.js
```diff
@@ -218,7 +214,8 @@ export const searchCollectivesInDB = async (
   }
 
   if (tags?.length) {
-    dynamicConditions += `AND c."tags" @> (:searchedTags) `;
+    searchedTags = tags.map(tag => tag.toLowerCase());
+    dynamicConditions += `AND c."tags" @> Array[:searchedTags]::varchar[] `;
```
Betree, line 218: Fixed the condition here to get rid of `searchedTags = `{${tags.map(tag => `"${tag}"`).join(',')}}`;` which was breaking the query when searching for something with `"`.

PR #7809, opened 2022-08-06
fix(editCollective): allow resetting location of an event

snitin315:
Refers https://github.com/opencollective/opencollective/issues/5490

Related Frontend PR - https://github.com/opencollective/opencollective-frontend/pull/8072

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -308,7 +308,19 @@ export function editCollective(_, args, req) {
   }
 
   // Set location values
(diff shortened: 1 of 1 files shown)


Betree (changes requested):
Good stuff, I wasn't aware that our API did not support this.

We should differentiate between `null` and `undefined`:
- calling `editCollective(name": $name, location: null)` should reset the location
- calling `editCollective(name": $name)` should **not** reset the location

Betree (approved):
Looking good!

PR #7810, opened 2022-08-07
Add lastUsedAt field in OAuthAuthorization

la-costa:
Implements the back end functionality for show last usage date for connected application at front end.

### DIFF
--- a/server/graphql/v2/object/OAuthAuthorization.js
+++ b/server/graphql/v2/object/OAuthAuthorization.js
@@ -1,6 +1,7 @@
 import { GraphQLList, GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
 import { GraphQLDateTime } from 'graphql-scalars';
 
+import models from '../../../models';
 import { OAuthScope } from '../enum/OAuthScope';
 import { getIdEncodeResolver, IDENTIFIER_TYPES } from '../identifiers';
 import { Application } from '../object/Application';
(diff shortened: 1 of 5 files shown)


server/graphql/v2/object/OAuthAuthorization.js
```diff
@@ -38,6 +39,19 @@ export const OAuthAuthorization = new GraphQLObjectType({
       type: new GraphQLNonNull(GraphQLDateTime),
       description: 'The time of expiration',
     },
+    lastUsedAt: {
+      type: GraphQLDateTime,
+      description: 'The last time of token was used',
+      async resolve(authorization) {
+        const activity = await models.Activity.findOne({
```
Betree, line 46: We should add a permission check here; only an admin of the token should be able to know its last usage date

server/graphql/v2/object/OAuthAuthorization.js
```diff
@@ -38,6 +39,19 @@ export const OAuthAuthorization = new GraphQLObjectType({
...
+          where: {
```
Betree, line 47: Small perf improvements: add a `SELECT` on the query to get only the field we're interested in

```suggestion
          attributes: ['createdAt'],
          where: {
```

Betree (changes requested):
Looking good!

To get rid of the ESLint error, we should run `npm run graphql:update` while the API is running.

server/graphql/v2/object/OAuthAuthorization.js
la-costa: I added the permission check, can you check if the implementation is correct?

I also already updated GraphQL

test/server/graphql/v2/query/ApplicationQuery.test.ts
```diff
@@ -16,6 +16,7 @@ const applicationQuery = gqlV2/* GraphQL */ `
       redirectUri
       clientId
       clientSecret
+      lastUsedAt
```
Betree, line 19: `lastUsedAt` lives in `OAuthAuthorization`, not `Application` (this is the reason why tests are failing)

la-costa: Thanks for the note, I already fixed it.

However, one of the tests is still failing and the reason is that the test data is not correctly instantiated. I tried to create the `OAuthAuthorization`, but I was not successful. Can you help me at this point?

Betree: Thanks, I'll take it from here

Betree:
Issue: https://github.com/opencollective/opencollective/issues/5650

PR #7814, opened 2022-08-08
Remove `setupModel` calls

Betree:
While looking at a model, I realized Typescript was missing some errors; it somehow failed to infer the type from inside the function, probably because the argument passed to `setupModel` (the model itself) was not typed.

I figured that, since we have `.git-blame-ignore-revs` now, we could get rid of them and ignore the commit in the history.

This is best reviewed with "Ignore whitespace changes": https://github.com/opencollective/opencollective-api/pull/7814/files?diff=split&w=1

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,3 +1,5 @@
 # .git-blame-ignore-revs
 # style: eslint --fix and prettier
 bd115f4eed74484ab4cee076711c027eab7b1a05
+# Remove `setupModel` calls
+2630abeb6c5b540c6d13b7c8afcc74a57ec58fcf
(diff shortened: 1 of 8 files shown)


Betree:
@znarf I'll wait for your feedback before merging this one

znarf:
Ok, but we should. not squash and merge right ?

I also saw that it's possible to add comments in the `.git-blame-ignore-revs`, that could be useful?

Betree:
> Ok, but we should. not squash and merge right ?
Indeed

> I also saw that it's possible to add comments in the `.git-blame-ignore-revs`, that could be useful?
Adding one now

znarf (approved):

Betree:
Rebased

PR #7815, opened 2022-08-08
Add Support to Send Notification to Admins when Virtual Card is Used

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5696

~⚠️ Not ready to be merged. See https://github.com/opencollective/opencollective-api/pull/7815#issuecomment-1208464028~


SudharakaP:
@kewitz : This PR is not ready to be merged. We need to test this by paying with a virtual card in test mode. I don't have the setup necessary for this test, and when you are back we can discuss. 

cc: @Betree

cron/hourly/reconcile-transactions.js
```diff
@@ -48,6 +50,32 @@ async function reconcileConnectedAccount(connectedAccount) {
           'approvals',
         );
 
+        const collectiveId = card.collectiveId;
+        const collective = models.Collective.findOne({
```
Betree, line 54:
```suggestion
        const collective = await models.Collective.findOne({
```

cron/hourly/reconcile-transactions.js
```diff
@@ -48,6 +50,32 @@ async function reconcileConnectedAccount(connectedAccount) {
...
+          where: {
+            id: collectiveId,
+          },
+        });
+        const user = models.User.findOne({ where: { id: user.id } });
```
Betree, line 59:
```suggestion
        const user = await models.User.findOne({ where: { id: user.id } });
```

cron/hourly/reconcile-transactions.js
```diff
@@ -48,6 +50,32 @@ async function reconcileConnectedAccount(connectedAccount) {
...
+
+        const adminUsers = await collective.getAdminUsers();
+
+        transactions.forEach(transaction => {
+          const expense = models.Expense.findOne({ where: { id: transaction.ExpenseId } });
```
Betree, line 64:
```suggestion
          const expense = await models.Expense.findOne({ where: { id: transaction.ExpenseId } });
```

templates/emails/virtualcard.purchase.hbs
```diff
@@ -0,0 +1,26 @@
...
+
+<center>
+  <h1>
+    A card attached to {{collective.name}} was charged {{currency expense.amount currency=expense.currency
+  precision=2}}.
+  </h1>
+</center>
+<p>
+A card attached to {{collective.name}} was charged {{currency expense.amount currency=expense.currency
+  precision=2}}. The responsible admin {{user.firstName}} {{user.lastName}} will be asked to submit additional
```
Betree, line 13: firstName/lastName is deprecated in favor or `name` on the `Collective` model. It's actually not part of the `User` model anymore I believe?

Betree (changes requested):
The code looks like it won't work (missing async/await) in its current version.

Can we add a test in `test/cron/hourly/reconcile-transactions.test.ts`, mocking the Stripe API when needed? It looks like this is a sensitive script that is going to stay, would be great to cover its behavior and that will allow us to test your changes even without access to the Stripe API.

cron/hourly/reconcile-transactions.js
```diff
@@ -48,6 +50,32 @@ async function reconcileConnectedAccount(connectedAccount) {
           'approvals',
         );
 
+        const collectiveId = card.collectiveId;
```
Betree, line 53: Isn't this one uppercase?

```suggestion
        const collectiveId = card.CollectiveId;
```

cron/hourly/reconcile-transactions.js
```diff
@@ -48,6 +50,32 @@ async function reconcileConnectedAccount(connectedAccount) {
...
+        const collective = models.Collective.findOne({
+          where: {
+            id: collectiveId,
+          },
+        });
```
Betree, line 58: :ice_cream: 

```suggestion
        const collective = await models.Collective.findByPk(collectiveId);
```

or 

```suggestion
        const collective = await card.getCollective();
```

cron/hourly/reconcile-transactions.js
Betree: Or simply

```suggestion
        const user = await models.User.findByPk(user.id);
```

Betree (changes requested):

templates/emails/virtualcard.purchase.hbs
SudharakaP: That's correct. I've refactored this part to use the `Collective` model.

server/lib/stripe.js
```diff
@@ -11,6 +11,12 @@ stripe.setMaxNetworkRetries(2);
 
 export default stripe;
 
+export const StripeCustomToken = token => {
+  const stripe = Stripe(token);
+  stripe.setMaxNetworkRetries(2);
+  return stripe;
+};
+
```
Betree, line 19: I don't understand why we need this?

cron/hourly/reconcile-transactions.js
```diff
@@ -1,21 +1,44 @@
...
+  const collectiveId = card.dataValues.CollectiveId;
+  const collective = await models.Collective.findByPk(collectiveId);
+  const user = await models.User.findByPk(card.dataValues.UserId);
+  const responsibleAdmin = await models.Collective.findByPk(user.CollectiveId);
+
+  const adminUsers = await collective.getAdminUsers();
+
+  for (const transaction of transactions) {
+    const expense = await models.Expense.findByPk(transaction.ExpenseId);
+    emailLib.send(
```
Betree, line 30: Maybe await?

Betree (commented):
Cool, the test looks great! Let's wait for @kewitz or @znarf to look/test this one

server/lib/stripe.js
SudharakaP: Note that, 

https://github.com/opencollective/opencollective-api/blob/c87250acf2ed5a082565c44e3f9a24dce662eda9/cron/hourly/reconcile-transactions.js#L87

We create a new `Stripe` object here regardless of whether we use `config.stripe.secret` or `connectedAccount.token`. This creates a problem for the tests as I don't think if we use this approach it will work with things like, `sandbox.stub(stripe.issuing.transactions, 'list')...` etc. Since there's no way to attach to the new `Stripe` object that's been used in the `reconcile-transactions.js`. Therefore I made a small change there so as to reuse the `Stripe` object created within `stripe.js` when the `host.slug === 'opencollective'`. This way in our tests we could use `slug='opencollective'`,

 https://github.com/opencollective/opencollective-api/blob/6cc538908e2d5f654757ceaccac5d4c251b99407/test/cron/hourly/reconcile-transactions.test.js#L128

and the tests would simply be working fine. 

Also I felt that the creation of `Stripe` objects should be moved to the `stripe` lib.

SudharakaP:
> The code looks like it won't work (missing async/await) in its current version.
> 
> Can we add a test in `test/cron/hourly/reconcile-transactions.test.ts`, mocking the Stripe API when needed? It looks like this is a sensitive script that is going to stay, would be great to cover its behavior and that will allow us to test your changes even without access to the Stripe API.

@Betree : I've added the tests for both `Stripe` as well as `Privacy`. Let me know if there's anything missing. 😃

SudharakaP:
@kewitz : I've changed this implementation as per our discussion. Now we send the email when an authorization is made on the stripe account rather than on the reconciliation. 

cc: @Betree

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -121,6 +120,26 @@ const setCardStatus = async (virtualCard, status = 'canceled' | 'active' | 'inac
...
+    activities.VIRTUAL_CARD_PURCHASE,
+    adminUsers.map(u => u.dataValues.email),
+    {
+      responsibleAdmin: responsibleAdmin.dataValues,
+      collective: collective.dataValues,
+      amount,
+      currency,
+    },
+  );
+};
```
kewitz, line 141: I think this looks good. But for consistency, I would add a `VIRTUAL_CARD_PURCHASE` activity handler in `lib/notifications/email.js`.

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -201,6 +220,8 @@ export const processAuthorization = async (stripeAuthorization, stripeEvent) =>
       CreatedByUserId: UserId,
       amount,
     });
+
+    await sendPurchaseNotifyEmail(virtualCard, amount, currency);
```
kewitz, line 224: If you go with the Activity solution, you would just need to create a VIRTUAL_CARD_PURCHASE activity here with the necessary data.

server/paymentProviders/stripe/virtual-cards.js
SudharakaP: Done. 👍🏽

server/paymentProviders/stripe/virtual-cards.js
SudharakaP: @kewitz : Done. Let me know if all looks good now. 👍🏽

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -120,24 +119,22 @@ const setCardStatus = async (virtualCard, status = 'canceled' | 'active' | 'inac
...
-  await emailLib.send(
-    activities.VIRTUAL_CARD_PURCHASE,
-    adminUsers.map(u => u.dataValues.email),
-    {
+  await models.Activity.create({
+    type: activities.VIRTUAL_CARD_PURCHASE,
+    CollectiveId,
+    data: {
       responsibleAdmin: responsibleAdmin.dataValues,
       collective: collective.dataValues,
```
kewitz, line 133:
```js
  responsibleAdmin: responsibleAdmin.activity,
  collective: collective.activity,
```

kewitz (commented):
Nice, looks good.
Just make sense to use the existing activity structure and not the internal `dataValues`.

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -121,6 +119,24 @@ const setCardStatus = async (virtualCard, status = 'canceled' | 'active' | 'inac
   return data;
 };
 
+const recordInActivityLog = async (virtualCard, amount, currency) => {
+  const CollectiveId = virtualCard.collective.id;
+  const collective = await models.Collective.findByPk(CollectiveId);
+  const user = await models.User.findByPk(virtualCard.UserId);
+  const responsibleAdmin = await models.Collective.findByPk(user.CollectiveId);
```
kewitz, line 126: I think we can extract user and collective from virtual card here.
```suggestion
  const collective = virtualCard.collective;
  const user = virtualCard.user;
  const responsibleAdmin = await models.Collective.findByPk(user.CollectiveId);
```

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -121,6 +119,24 @@ const setCardStatus = async (virtualCard, status = 'canceled' | 'active' | 'inac
...
+
+  await models.Activity.create({
+    type: activities.VIRTUAL_CARD_PURCHASE,
+    CollectiveId,
```
kewitz, line 130: I think we can also store `UserId` here perhaps?

server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -121,6 +119,24 @@ const setCardStatus = async (virtualCard, status = 'canceled' | 'active' | 'inac
   return data;
 };
 
+const recordInActivityLog = async (virtualCard, amount, currency) => {
```
kewitz, line 122: Not sure if necessary to abstract this function if it is being used in a single place.

kewitz (approved):
Looks good! Left a few tips if you want to work on but not necessary.

kewitz (approved):

kewitz:
Make sure the tests are passing before merging it.

PR #7830, opened 2022-08-12
fix: remove collective name from expense email

nosro:
As per https://github.com/opencollective/opencollective/issues/5207 the expense email comment notification goes to both the admin as well as the expense payee, so a singular greeting name doesn't work.

As a quick fix, remove the greeting name to make the email greeting generic for both recipients.

### DIFF
--- a/templates/emails/collective.comment.created.hbs
+++ b/templates/emails/collective.comment.created.hbs
@@ -10,9 +10,9 @@ Subject: {{{collective.name}}}: New comment
 
 {{> header}}
 
-{{#if expense }}
+{{#if expense}}
   <p style="padding: 1em;">
-    Hi {{collective.name}},
+    Hi,
     <br /><br />
     <a href="{{config.host.website}}/{{fromCollective.slug}}">{{fromCollective.name}}</a> just posted
     a new comment on an Expense, <a href="{{config.host.website}}{{path}}">{{expense.description}}</a>:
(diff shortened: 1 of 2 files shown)


nosro:
I'm unclear if one or two templates are involved: `collective.comment.created.hbs` and/or `expense.comment.created.hbs`. I will look further to spot the contexts these are used. Maybe a different greeting name can be set in each.

I can see that there are separate notifications for three recipients potentially: admin of the collective, admin of the host, and author of the expense: https://github.com/nosro/opencollective-api/blob/main/server/lib/notifications/email.ts#L304-L326

A better fix would be to have a separate email template fragment for each recipient to allow for different greeting names, or even better - a recipient name set in the template data to make a single template do the job.

nosro:
Regarding separate individualized greetings, I see the data payload for the email template is defined here:
https://github.com/opencollective/opencollective-api/blob/main/server/models/Comment.js#L161-L176

I don't see names provided in the data payload there. I do see the `CollectiveId` and `UserId` but I imagine these are not human readable index ids, not suitable for a template. The data model passed here would have to be updated with names.

Betree (approved):
> a recipient name set in the template data to make a single template do the job.

Sounds like a good long-term solution! In one of the [templates](https://github.com/opencollective/opencollective-api/blob/3ca523dd42e29ff9aa44f1df0bb5b9159088e00e/templates/emails/taxform.request.hbs#L5) we use `recipientName` as a variable name. Could be great to have a `greeting` partial like:

```hbs
{{#if recipientName}}
<p>Hi {{recipientName}},</p>
{{else}}
<p>Hi,</p>
{{/if}}
```

That could be used like:

```hbs
{{> greeting}}

My email content

{{> footer}}
```

I'll open a follow-up issue to suggest that. For now, your fix will do just great :+1: 

Thank you!

PR #7839, opened 2022-08-15
Add "amount" to expensesTags

znarf:
It's built on top of the existing `expensesTags` properties.

Unfortunately it's not using the "collection" pattern.

### DIFF
--- a/server/graphql/v2/interface/Account.js
+++ b/server/graphql/v2/interface/Account.js
@@ -2,7 +2,6 @@ import { GraphQLBoolean, GraphQLInt, GraphQLInterfaceType, GraphQLList, GraphQLN
 import { GraphQLDateTime } from 'graphql-scalars';
 import { GraphQLJSON } from 'graphql-type-json';
 import { assign, get, invert, isEmpty, pick } from 'lodash';
-import moment from 'moment';
 
 import { types as CollectiveTypes } from '../../../constants/collectives';
 import FEATURE from '../../../constants/feature';
@@ -28,7 +27,6 @@ import { AccountType, AccountTypeToModelMapping, ImageFormat, MemberRole } from 
 import { ActivityChannel } from '../enum/ActivityChannel';
 import { PaymentMethodService } from '../enum/PaymentMethodService';
 import { PaymentMethodType } from '../enum/PaymentMethodType';
-import { TimeUnit } from '../enum/TimeUnit';
 import { idEncode } from '../identifiers';
 import { AccountReferenceInput, fetchAccountWithReference } from '../input/AccountReferenceInput';
 import { ChronologicalOrderInput } from '../input/ChronologicalOrderInput';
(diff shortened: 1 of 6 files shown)


server/models/Expense.js
```diff
@@ -480,15 +481,26 @@ function defineModel() {
...
-      SELECT UNNEST(tags) AS id, UNNEST(tags) AS tag, COUNT(id)
-      FROM "Expenses"
-      WHERE "CollectiveId" = $collectiveId
-      AND "deletedAt" IS NULL
-      GROUP BY UNNEST(tags)
-      ORDER BY count DESC
+      SELECT TRIM(UNNEST(COALESCE(e."tags", '{"${noTag}"}'))) AS id,
+      TRIM(UNNEST(COALESCE(e."tags", '{"${noTag}"}'))) AS tag,
+      COUNT(e."id") as "count",
+      ABS(SUM(t."amountInHostCurrency")) as "amount",
```
kewitz, line 491: We should convert this to collective currency.

server/models/Expense.js
```diff
@@ -498,6 +510,44 @@ function defineModel() {
...
+    timeUnit,
+    { dateFrom = null, dateTo = null } = {},
+  ) {
+    const noTag = 'no tag';
+    return sequelize.query(
+      `
+      SELECT DATE_TRUNC($timeUnit, t."createdAt") AS "date",
+      TRIM(UNNEST(COALESCE(e."tags", '{"${noTag}"}'))) AS label,
+      COUNT(e."id") as "count",
+      ABS(SUM(t."amountInHostCurrency")) as "amount",
```
kewitz, line 524: We should convert this to collective currency.

server/models/Expense.js
```diff
@@ -480,15 +481,26 @@ function defineModel() {
     return legacyPayoutMethod === 'paypal' ? PayoutMethodTypes.PAYPAL : PayoutMethodTypes.OTHER;
   };
 
-  Expense.getMostPopularExpenseTagsForCollective = async function (collectiveId, limit = 100) {
+  Expense.getCollectiveExpensesTags = async function (collectiveId, limit = 100) {
```
kewitz, line 484: I would keep the original `getMostPopularExpenseTagsForCollective` since it is relatively simpler to run.

server/graphql/v2/interface/Account.js
```diff
@@ -292,6 +322,25 @@ const accountFieldsDefinition = () => ({
       limit: { type: new GraphQLNonNull(GraphQLInt), defaultValue: 30 },
     },
   },
+  expensesTagsTimeSeries: {
```
kewitz, line 325: Move to `AccountStats`.

server/models/Expense.js
znarf: Now looking at collective `currency`

server/models/Expense.js
znarf: Now looking at collective `currency`

server/models/Expense.js
znarf: Done

server/graphql/v2/interface/Account.js
znarf: Done

znarf:
Moved to https://github.com/opencollective/opencollective-api/pull/7853

PR #7841, opened 2022-08-15
Add support for returning virtual cards for users

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5695

Related Front-end PR:


server/graphql/v2/interface/Account.js
```diff
@@ -463,11 +463,19 @@ const accountFieldsDefinition = () => ({
         merchantId = (await fetchAccountWithReference(args.merchantAccount, { throwIfMissing: true })).id;
       }
 
+      let where;
+      if (account.type === CollectiveTypes.USER) {
+        const user = await models.User.findOne({
+          where: { CollectiveId: account.id },
+        });
+        where = { UserId: user.id };
```
Betree, line 471: `UserId` is meant to store the user who created the card. It seems like a hack to re-use `account` here, and I'm not sure how it will play if we ever roll out virtual cards for other hosts (whose type can be `USER`).

I'd be in favor of an `assignee` (Individual) argument to achieve this.

server/graphql/v2/object/VirtualCard.ts
```diff
@@ -48,7 +48,10 @@ export const VirtualCard = new GraphQLObjectType({
       type: GraphQLString,
       async resolve(virtualCard, _, req) {
         const collective = await req.loaders.Collective.byId.load(virtualCard.CollectiveId);
-        if (req.remoteUser?.isAdminOfCollectiveOrHost(collective) && checkScope(req, 'virtualCards')) {
+        if (
+          (req.remoteUser?.isAdminOfCollectiveOrHost(collective) || req.remoteUser?.id === virtualCard.UserId) &&
+          checkScope(req, 'virtualCards')
```
Betree, line 53: Maybe add a small helper like `canSeeVirtualCardPrivateInfo` to reduce code duplication and prevent mistakes?

Betree (changes requested):

server/graphql/v2/object/VirtualCard.ts
SudharakaP: Thanks. I've added the helper. 👍🏽

server/graphql/v2/interface/Account.js
SudharakaP: > no_entry_sign `UserId` is meant to store the user who created the card. 

I am not sure what you mean by this. To me it seems that the `UserId` within the `VirtualCard` table stores the user that the virtual card is assigned. In fact we use `UserId` to computer the `assignee` field here; 

https://github.com/opencollective/opencollective-api/blob/9255bd06ac4faae4740af36254595d4e7324dfd2/server/graphql/v2/object/VirtualCard.ts#L35-L50

> I'd be in favor of an `assignee` (Individual) argument to achieve this.

I've added a `assignee` argument and modified the front-end as well to pass that. Let me know if you see any issues. 🤔

Betree: It's the assignee indeed. My feedback was really about the workaround using Account being too hacky.

server/graphql/v2/interface/Account.js
```diff
@@ -463,11 +464,20 @@ const accountFieldsDefinition = () => ({
...
+      let where;
+      if (!isEmpty(args.assignee)) {
+        const CollectiveId = (await fetchAccountWithReference(args.assignee, { throwIfMissing: true })).id;
+        const user = await models.User.findOne({
+          where: { CollectiveId },
+        });
+        where = { UserId: user.id };
+      } else {
+        where = { CollectiveId: account.id };
+      }
```
Betree, line 476:
```suggestion
      const where = { CollectiveId: account.id };
      if (args.assignee) {
        const assignee = await fetchAccountWithReference(args.assignee, { throwIfMissing: true });
        const user = await assignee.getUser();
        if (!user) {
          throw new BadRequest('Could not find the assigned user');
        }
        where.UserId = user.id;
```

server/graphql/v2/interface/Account.js
Betree: Edit: I misread the original issue and did not understand that we wanted to add this to the user settings.

The new API looks wrong too, it doesn't make sense to do:

```gql
collective(slug: "babel") {
  virtualCards(assignee: "betree) { ... }
}
```

and end up with virtual cards for Webpack assigned to betree in the list.

Betree: Suggestion:

```gql
account(slug: "betree) {
  virtualCards(relation: ASSIGNED) { ... }
}
```

With `relation` set to `OWNED` by default.

Not sure whether `relation` is the best term, feel free to dig more.

server/graphql/v2/object/VirtualCard.ts
```diff
@@ -7,6 +7,10 @@ import { Currency } from '../enum/Currency';
 import { Account } from '../interface/Account';
 import { Individual } from '../object/Individual';
 
+const canSeeVirtualCardPrivateInfo = (req, collective, virtualCard) =>
+  (req.remoteUser?.isAdminOfCollectiveOrHost(collective) || req.remoteUser?.id === virtualCard.UserId) &&
```
znarf, line 11: Why do we need this new condition? `req.remoteUser?.id === virtualCard.UserId`. 

Why would the user not be ADMIN of the Collective?

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -330,7 +330,7 @@ const virtualCardMutations = {
         throw new NotFound('Could not find Virtual Card');
       }
 
-      if (!req.remoteUser.isAdmin(virtualCard.HostCollectiveId)) {
+      if (!req.remoteUser.isAdmin(virtualCard.HostCollectiveId) && req.remoteUser.id !== virtualCard.UserId) {
```
znarf, line 333: Why would the user not be ADMIN of the Collective?

server/graphql/v2/mutation/VirtualCardMutations.ts
```diff
@@ -330,7 +330,7 @@ const virtualCardMutations = {
...
         throw new Unauthorized("You don't have permission to edit this Virtual Card");
```
znarf, line 334:
```suggestion
        throw new Unauthorized("You don't have permission to pause this Virtual Card");
```

server/graphql/v2/interface/Account.js
znarf:
```graphql
collective(slug: "babel") { virtualCards(assignee: "betree) { ... } }
````

I'm fine with this API if we make sure that we only fetch cards assigned to "betree" in the "babel" collective. And not call cards assigned to "betree".

znarf:
Why can't a Collective admin not freeze the cards assigned to other members? He's admin right?

server/graphql/v2/interface/Account.js
Betree: If we don't add a virtual cards page in users settings (which I think is a good call) then we indeed don't need this `relation` thing and can directly go with an `assignee` arguments on `virtualCards`.

SudharakaP: Refactored the code to remove the virtual cards page in user's settings. So these changes have been reverted.

server/graphql/v2/object/VirtualCard.ts
SudharakaP: This is no longer needed as I've removed the virtual card section from user settings.

SudharakaP:
> Why can't a Collective admin not freeze the cards assigned to other members? He's admin right?

Now it's possible to do that via the collective settings -> virtual cards page. 👍🏽

znarf (approved):
Looks good to me!

PR #7850, opened 2022-08-18
OAuth scope follow up

znarf:
Fix: https://github.com/opencollective/opencollective/issues/5863
Fix: https://github.com/opencollective/opencollective/issues/5835
Also prepare: https://github.com/opencollective/opencollective/issues/5829

- Add missing "incognito" scope
- Remove "paymentMethods", "payoutMethods" scopes
- Support `null` scope
- Add new "activities" scope


server/graphql/v2/mutation/AddFundsMutations.ts
```diff
@@ -12,7 +12,7 @@ import { Order } from '../object/Order';
 
 export const addFundsMutation = {
   type: new GraphQLNonNull(Order),
-  description: 'Add funds to the given account. Scope: "account".',
+  description: 'Add funds to the given account. Scope: "host".',
```
znarf, line 15: Just a small error in the description.

server/models/UserToken.ts
```diff
@@ -26,12 +26,8 @@ class UserToken extends Model<InferAttributes<UserToken>, InferCreationAttribute
   public declare user?: NonAttribute<typeof models.User>;
   public declare client?: NonAttribute<typeof models.Application>;
 
-  getScope() {
-    return this.scope;
-  }
-
   hasScope(scope) {
-    return this.getScope().includes(scope);
+    return this.scope && this.scope.includes(scope);
```
znarf, line 30: Simpler and support `null` scope.

migrations/20220818111146-oauth-scope-follow-up.js
```diff
@@ -0,0 +1,79 @@
...
+const updateEnums = async (queryInterface, table, column, enumName, values) => {
+  // See https://blog.yo1.dog/updating-enum-values-in-postgresql-the-safe-and-easy-way/
+  await executeQuery(queryInterface, `ALTER TYPE "${enumName}" RENAME TO "${enumName}_old"`);
+  await executeQuery(queryInterface, `CREATE TYPE "${enumName}" AS ENUM(${formatEnums(values)})`);
+  await executeQuery(
+    queryInterface,
+    `ALTER TABLE "${table}" ALTER COLUMN ${column} TYPE "${enumName}" ARRAY USING ${column}::text::"${enumName}"[]`,
+  );
+  await executeQuery(queryInterface, `DROP TYPE "${enumName}_old"`);
+};
```
Betree, line 55: Nice! Could be interesting to move it to `migrations/lib/helpers.ts` to easily re-use it in the future

server/models/UserToken.ts
Betree: :ice_cream: We can make it safer if we specify the return type

```
hasScope(scope): boolean {
```

```suggestion
    return Boolean(this.scope && this.scope.includes(scope);)
```

Betree (approved):
Looking good :+1:

PR #7861, opened 2022-08-19
Wise: Automatically create new batch when a single expense is unscheduled

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5883

Also fix a bug that keeps outdated batch group information on our database and improves how we verify existing batch group status.


mrarif6585:
gh pr checkout 740

PR #7864, opened 2022-08-20
Check if fromCollective can use feature when creating order

kewitz:


server/lib/user-permissions.ts
```diff
@@ -9,14 +9,14 @@ import models from '../models';
  * Returns true if the given user can use the passed feature. Will always return false
  * if user is not set.
  */
-export const canUseFeature = (user: typeof models.User, feature: FEATURE): boolean => {
+export const canUseFeature = (account: typeof models.User | typeof models.Collective, feature: FEATURE): boolean => {
```
Betree, line 12: We should rename, `hasFeature` is the way to check if a feature is enabled for an account. The confusion with this one will quickly emerge if they share the same prototype. I can suggest:
- `isNotBLockedFeature`
- `isNotFrozenFeature`

kewitz: But the data structure is the same, isn't it? My interpretation is that perhaps feature access is a bit more abstract than we first thought, and it also applies to the user model.
Should I just go on and use another function to check for a feature in collectives?

Betree: It is the same data structure indeed, and I think you made a good move changing this prototype to accept either a user or an account. I'm just concerned by the name of this function, we now have two things: `canUseFeature` / `hasFeature` with very similar names and different purposes. But maybe these names are already clear enough and I'm over-anticipating. Let's wait to see if other devs get confused by this.

PR #7876, opened 2022-08-24
enhancement(transactions): Store PayoutMethodId on Transaction

vaughanj10:
Closes https://github.com/opencollective/opencollective/issues/5505

This PR adds a `PayoutMethodId` field to `Transaction` so that we keep track of the correct PayoutMethod over time.


znarf (approved):

znarf:
Follow up, migration of all previous Transactions.

PR #7878, opened 2022-08-25
Add an activity for `collective.edited`

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5864

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -1,6 +1,6 @@
 import config from 'config';
 import slugify from 'limax';
-import { get, omit, truncate } from 'lodash';
+import { cloneDeep, get, isEqual, omit, pick, truncate } from 'lodash';
 import sanitize from 'sanitize-html';
 import { v4 as uuid } from 'uuid';
 
@@ -16,6 +16,7 @@ import { canUseFeature } from '../../../lib/user-permissions';
 import { defaultHostCollective } from '../../../lib/utils';
 import models from '../../../models';
 import { FeatureNotAllowedForUser, NotFound, RateLimitExceeded, Unauthorized, ValidationFailed } from '../../errors';
+import { CollectiveInputType } from '../inputTypes';
 
 const DEFAULT_COLLECTIVE_SETTINGS = {
   features: { conversations: true },
@@ -291,6 +292,18 @@ export async function createCollectiveFromGithub(_, args, req) {
   return collective;
 }
 
+function getCollectiveDataDiff(originalCollective, modifiedCollective) {
+  const collectiveInputTypeFields = Object.keys(CollectiveInputType.getFields());
+  const originalCollectiveData = pick(originalCollective, collectiveInputTypeFields);
+  const modifiedCollectiveData = pick(modifiedCollective, collectiveInputTypeFields);
+  const differenceKeys = Object.keys(originalCollectiveData).filter(
+    k => !isEqual(originalCollectiveData[k], modifiedCollectiveData[k]),
+  );
+  const previousData = pick(originalCollectiveData, differenceKeys);
+  const newData = pick(modifiedCollectiveData, differenceKeys);
+  return { previousData, newData };
+}
+
 export function editCollective(_, args, req) {
   if (!req.remoteUser) {
     throw new Unauthorized('You need to be logged in to edit a collective');
@@ -341,7 +354,7 @@ export function editCollective(_, args, req) {
     newCollectiveData.countryISO = location.country;
   }
 
-  let collective, parentCollective;
+  let originalCollective, collective, parentCollective;
 
   return req.loaders.Collective.byId
     .load(args.collective.id)
@@ -349,6 +362,7 @@ export function editCollective(_, args, req) {
       if (!c) {
         throw new Error(`Collective with id ${args.collective.id} not found`);
       }
+      originalCollective = cloneDeep(c);
       collective = c;
     })
     .then(() => {
@@ -433,6 +447,17 @@ export function editCollective(_, args, req) {
     .then(() => {
       // Ask cloudflare to refresh the cache for this collective's page
       purgeCacheForCollective(collective.slug);
+      const data = getCollectiveDataDiff(originalCollective, collective);
+      // Create the activity which will store the data diff
+      models.Activity.create({
+        type: activities.COLLECTIVE_EDITED,
+        UserId: req.remoteUser.id,
+        UserTokenId: req.userToken?.id,
+        CollectiveId: collective.id,
+        FromCollectiveId: collective.id,
+        HostCollectiveId: collective.approvedAt ? collective.HostCollectiveId : null,
+        data,
+      });
       return collective;
     });
 }
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -444,12 +444,12 @@ export function editCollective(_, args, req) {
         });
       }
     })
-    .then(() => {
+    .then(async () => {
       // Ask cloudflare to refresh the cache for this collective's page
       purgeCacheForCollective(collective.slug);
       const data = getCollectiveDataDiff(originalCollective, collective);
       // Create the activity which will store the data diff
-      models.Activity.create({
+      await models.Activity.create({
         type: activities.COLLECTIVE_EDITED,
         UserId: req.remoteUser.id,
         UserTokenId: req.userToken?.id,
--- a/server/graphql/v2/mutation/AccountMutations.ts
+++ b/server/graphql/v2/mutation/AccountMutations.ts
@@ -99,6 +99,11 @@ const accountMutations = {
           }
         }
 
+        const settings = account.settings ? cloneDeep(account.settings) : {};
+        set(settings, args.key, args.value);
+
+        const previousData = { settings: { [args.key]: account.data?.[args.key] } };
+        const updatedAccount = account.update({ settings }, { transaction });
         await models.Activity.create(
           {
             type: activities.COLLECTIVE_EDITED,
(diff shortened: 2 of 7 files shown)


server/graphql/v1/mutations/collectives.js
```diff
@@ -433,6 +447,17 @@ export function editCollective(_, args, req) {
     .then(() => {
       // Ask cloudflare to refresh the cache for this collective's page
       purgeCacheForCollective(collective.slug);
+      const data = getCollectiveDataDiff(originalCollective, collective);
+      // Create the activity which will store the data diff
+      models.Activity.create({
```
Betree, line 452: Can we await for this one?

server/graphql/v2/object/Activity.ts
```diff
@@ -114,6 +114,12 @@ export const Activity = new GraphQLObjectType({
           toPick.push('movedFromCollective');
         } else if (activity.type === ACTIVITY.COLLECTIVE_MEMBER_CREATED) {
           toPick.push('member.role');
+        } else if (activity.type === ACTIVITY.COLLECTIVE_EDITED) {
+          const collective = await req.loaders.Collective.byId.load(activity.CollectiveId);
+          if (req.remoteUser?.isAdmin(collective.id)) {
```
Betree, line 119:
```suggestion
          if (req.remoteUser?.isAdminOfCollectiveOrHost(collective)) {
```

Betree (commented):
The overall approach looks good :+1: Two things we need to iterate on:
- [ ] Mutations on GQLV2 needs to be covered too
- [ ] We need some unit tests to make sure this is working as expected

SudharakaP:
> The overall approach looks good +1 Two things we need to iterate on:
> 
>     * [x]  Mutations on GQLV2 needs to be covered too
> 
>     * [x]  We need some unit tests to make sure this is working as expected

@Betree : I've covered the GQLV2 mutations as well as added some tests for both v1 and v2. Let me know. 🙂

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -98,6 +99,16 @@ const accountMutations = {
...
         }
 
+        models.Activity.create({
+          type: activities.COLLECTIVE_EDITED,
+          UserId: req.remoteUser.id,
+          UserTokenId: req.userToken?.id,
+          CollectiveId: account.id,
+          FromCollectiveId: account.id,
+          HostCollectiveId: account.approvedAt ? account.HostCollectiveId : null,
+          data: { previousData: { [args.key]: account.data?.[args.key] }, newData: { [args.key]: args.value } },
```
Betree, line 109: `account.data` - are we not editing `settings` here? Also, I would make sure to preserve the existing structure:

```suggestion
          data: { previousData: { settings: previousSettings }, newData: { settings: newSettings } },
```

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -98,6 +99,16 @@ const accountMutations = {
           }
         }
 
+        models.Activity.create({
```
Betree, line 102: Let's `await` for them
```suggestion
        await models.Activity.create({
```

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -260,6 +286,13 @@ const accountMutations = {
         return crypto.hash(code);
       });
 
+      models.Activity.create({
```
Betree, line 289:
```suggestion
      await models.Activity.create({
```

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -303,6 +336,13 @@ const accountMutations = {
         throw new Unauthorized('Two-factor authentication code failed. Please try again');
       }
 
+      models.Activity.create({
```
Betree, line 339:
```suggestion
      await models.Activity.create({
```

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -382,8 +432,18 @@ const accountMutations = {
...
-          case 'currency':
+          case 'currency': {
+            models.Activity.create({
+              type: activities.COLLECTIVE_EDITED,
+              UserId: req.remoteUser.id,
+              UserTokenId: req.userToken?.id,
+              CollectiveId: account.id,
+              FromCollectiveId: account.id,
+              HostCollectiveId: account.approvedAt ? account.HostCollectiveId : null,
+              data: { oldCurrency: account.currency, newCurrency: args.account[key] },
```
Betree, line 443:
```suggestion
              data: { oldData: { currency: account.currency }, newData: { currency: newCurrency } },
```

Betree (commented):
Good job on adding the 2FA activities :+1: We should try to unify how we store the diff in the activity data, to make it easier to use in the frontend. You approach with `previousData`/`newData` looks good to me, I think we should use it everywhere

server/graphql/v2/mutation/AccountMutations.ts
SudharakaP: @Betree : The reason I didn't add these awaits is that in my local when I added `await` for `editAccountSetting` I noticed that there's a certain lag that's happening in the front-end side. Is there a specific reason why we need to add `await` here. As far as I see we don't care for an immediate update of the log. 

[awiat problem.webm](https://user-images.githubusercontent.com/12435965/189729826-06423403-3a59-452e-97ea-d02357ad6caf.webm)

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -382,8 +435,18 @@ const accountMutations = {
...
+          case 'currency': {
+            models.Activity.create({
+              type: activities.COLLECTIVE_EDITED,
+              UserId: req.remoteUser.id,
+              UserTokenId: req.userToken?.id,
+              CollectiveId: account.id,
+              FromCollectiveId: account.id,
+              HostCollectiveId: account.approvedAt ? account.HostCollectiveId : null,
+              data: { previousData: { currency: account.currency }, newData: { currency: args.account[key] } },
+            });
```
Betree, line 447: Can we do that after `setCurrency`, and only if the currency changes? `setCurrency` can fail, for example if there are transactions attached to the account

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -417,6 +480,16 @@ const accountMutations = {
...
+      models.Activity.create({
+        type: activities.COLLECTIVE_EDITED,
+        UserId: req.remoteUser.id,
+        UserTokenId: req.userToken?.id,
+        CollectiveId: account.id,
+        FromCollectiveId: account.id,
+        HostCollectiveId: account.approvedAt ? account.HostCollectiveId : null,
+        data: { previousData: account.data?.policies, newData: args.policies },
+      });
+
```
Betree, line 492: Can we do that after `setPolicies`? `setPolicies` can fail

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -455,6 +528,16 @@ const accountMutations = {
...
+      models.Activity.create({
+        type: activities.COLLECTIVE_DELETED,
+        UserId: req.remoteUser.id,
+        UserTokenId: req.userToken?.id,
+        CollectiveId: account.id,
+        FromCollectiveId: account.id,
+        HostCollectiveId: account.approvedAt ? account.HostCollectiveId : null,
+        data: { name: account.name, slug: account.slug },
+      });
+
```
Betree, line 540: Can we do that after `deleteCollective`? `deleteCollective` can fail

Betree (commented):
Nice tests :+1:

server/graphql/v2/mutation/AccountMutations.ts
SudharakaP: Done. 👍🏽

server/graphql/v2/mutation/AccountMutations.ts
SudharakaP: Done. 👍🏽

server/graphql/v2/mutation/AccountMutations.ts
SudharakaP: Done. 👍🏽

server/graphql/v2/mutation/AccountMutations.ts
SudharakaP: This only effects, `editAccountSetting` and `editAccountFeeStructure` where we have a `sequelize.transaction`. So every change within needs to pass the `transaction` object which I mistakenly didn't pass. I've corrected this and now everything works perfectly with `await`. 👍🏽

Betree:
Tests are not passing

Betree:
Rebased

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -98,6 +99,22 @@ const accountMutations = {
...
+            CollectiveId: account.id,
+            FromCollectiveId: account.id,
+            HostCollectiveId: account.approvedAt ? account.HostCollectiveId : null,
+            data: {
+              previousData: { settings: { [args.key]: account.data?.[args.key] } },
+              newData: { settings: { [args.key]: args.value } },
+            },
+          },
+          { transaction },
+        );
```
Betree, line 116: `account.update` can fail because of the model validations. The activity should be created only if it succeeds.

SudharakaP: Thanks. I've corrected that as well. 👍🏽

server/lib/collectivelib.ts
```diff
@@ -413,5 +415,13 @@ export async function deleteCollective(collective) {
     await user.destroy();
   }
 
+  await models.Activity.create({
+    type: activities.COLLECTIVE_DELETED,
+    CollectiveId: collective.id,
+    FromCollectiveId: collective.id,
+    HostCollectiveId: collective.approvedAt ? collective.HostCollectiveId : null,
+    data: { name: collective.name, slug: collective.slug },
```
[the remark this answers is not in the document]
SudharakaP: Done. 👍🏽

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -99,6 +99,11 @@ const accountMutations = {
           }
         }
 
+        const settings = account.settings ? cloneDeep(account.settings) : {};
+        set(settings, args.key, args.value);
+
+        const previousData = { settings: { [args.key]: account.data?.[args.key] } };
+        const updatedAccount = account.update({ settings }, { transaction });
```
Betree, line 106:
```suggestion
        const updatedAccount = await account.update({ settings }, { transaction });
```

Betree (approved):

PR #7880, opened 2022-08-25
Enhancement on Activity types

Betree:
Require https://github.com/opencollective/opencollective-frontend/pull/8131

This is a breaking change, but since the feature is not released publicly that should be fine.

- Rename `activityType` arg to `type`, make list elements non-nullable
- Update type filter to support passing activities directly (it was previously crashing the query).
- Ignore `COLLECTIVE_TRANSACTION_CREATED` activity, see https://github.com/opencollective/opencollective/issues/5903

### DIFF
--- a/server/graphql/schemaV2.graphql
+++ b/server/graphql/schemaV2.graphql
@@ -8466,9 +8466,9 @@ type Query {
     dateTo: DateTime = null
 
     """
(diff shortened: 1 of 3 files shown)


SudharakaP (approved):
💯 Tested and works well.

PR #7884, opened 2022-08-26
Improve code & complete activity when creating comments

Betree:
Part of https://github.com/opencollective/opencollective/issues/5887

- Add missing `HostCollectiveId` to the activity, make the call synchronous 
- Deprecate `ConversationId` in favor of `conversation` (reference)
- Improve overall code now that we don't need to support GQLV1

### DIFF
--- a/server/graphql/v2/input/ConversationReferenceInput.ts
+++ b/server/graphql/v2/input/ConversationReferenceInput.ts
@@ -1,0 +1,48 @@
+import { GraphQLInputObjectType, GraphQLInt, GraphQLString } from 'graphql';
+
+import models from '../../../models';
+import { NotFound } from '../../errors';
+import { idDecode, IDENTIFIER_TYPES } from '../identifiers';
+
+/**
+ * An input for referencing Conversations
+ */
+export const ConversationReferenceInput = new GraphQLInputObjectType({
+  name: 'ConversationReferenceInput',
+  fields: () => ({
+    id: {
+      type: GraphQLString,
+      description: 'The public id identifying the update',
+    },
+    legacyId: {
+      type: GraphQLInt,
+    },
+  }),
+});
+
+export const getConversationDatabaseIdFromUpdateReference = input => {
+  if (input['id']) {
+    return idDecode(input['id'], IDENTIFIER_TYPES.UPDATE);
+  } else if (input['legacyId']) {
+    return input['legacyId'];
+  } else {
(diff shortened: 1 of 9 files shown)


server/graphql/v2/input/ConversationReferenceInput.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * An input for referencing Conversations
+ */
+export const ConversationReferenceInput = new GraphQLInputObjectType({
+  name: 'ConversationReferenceInput',
+  fields: () => ({
+    id: {
+      type: GraphQLString,
+      description: 'The public id identifying the update',
```
SudharakaP, line 15:
```suggestion
      description: 'The public id identifying the conversation',
```

server/graphql/v2/input/ConversationReferenceInput.ts
```diff
@@ -0,0 +1,48 @@
...
+    return idDecode(input['id'], IDENTIFIER_TYPES.UPDATE);
+  } else if (input['legacyId']) {
+    return input['legacyId'];
+  } else {
+    return null;
+  }
+};
+
+/**
+ * Retrieve an expense from an `ExpenseReferenceInput`
```
SudharakaP, line 34:
```suggestion
 * Retrieve an conversation from an `ConversationReferenceInput`
```

server/graphql/v2/input/ConversationReferenceInput.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+export const fetchConversationWithReference = async (input, { loaders = null, throwIfMissing = false } = {}) => {
```
SudharakaP, line 36: This function is not used anywhere? 🤔

SudharakaP (approved):

server/graphql/v2/input/ConversationReferenceInput.ts
znarf: a conversation

znarf:
Looks good!

server/graphql/v2/input/ConversationReferenceInput.ts
Betree: Not yet, but it's there to match the pattern we have in all reference helpers: `fetch*WithReference`. Will be useful when we update the `conversation` query.

PR #7890, opened 2022-08-26
WIP: Create transactions when Wise transfer is paid refactor

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5435


kewitz:
@znarf I think it will not be necessary to update reports.

The reason why this doesn't affect reports is that this doesn't essentially change how we deal with expenses paid with Wise.
We're just delaying accounting the expense to the point where we have a confirmation the transaction was received. "Blocked fund" is not a concept that we take into consideration in our reports, it is just a check we have on our business logic.

PR #7897, opened 2022-08-30
Skip loader when dates are passed to the balance resolver

znarf:
Fix: https://github.com/opencollective/opencollective/issues/5920


Betree (approved):
Makes sense!

PR #7901, opened 2022-08-31
Add Activity Account Filter and Remove Attribute Filter

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5917

Related to: https://github.com/opencollective/opencollective/issues/5918

⚠️ Require: https://github.com/opencollective/opencollective-frontend/pull/8151


server/graphql/v2/query/collection/ActivitiesCollectionQuery.ts
```diff
@@ -18,12 +18,18 @@ const ActivitiesCollectionArgs = {
   limit: { ...CollectionArgs.limit, defaultValue: 100 },
   offset: CollectionArgs.offset,
   account: {
-    type: new GraphQLNonNull(AccountReferenceInput),
-    description: 'The account associated with the Activity',
+    type: new GraphQLList(new GraphQLNonNull(AccountReferenceInput)),
```
Betree, line 21: We still want people to pass at least one account to filter by

```suggestion
    type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(AccountReferenceInput))),
```

server/graphql/v2/query/collection/ActivitiesCollectionQuery.ts
```diff
@@ -47,36 +53,34 @@ const ActivitiesCollectionQuery = {
   args: ActivitiesCollectionArgs,
   async resolve(_: void, args, req): Promise<CollectionReturnType> {
     const { offset, limit } = args;
-    const account = await fetchAccountWithReference(args.account, { throwIfMissing: true });
+    let accounts;
+    if (args.account) {
+      accounts = await fetchAccountsWithReferences(args.account, { throwIfMissing: true });
+    }
```
Betree, line 59:
```suggestion
    if (!args.account.length) {
      throw '...';
    }
    
    const accounts = await fetchAccountsWithReferences(args.account, { throwIfMissing: true });
```

server/graphql/v2/query/collection/ActivitiesCollectionQuery.ts
```diff
@@ -47,36 +53,34 @@ const ActivitiesCollectionQuery = {
...
+      if (isRootUser || req.remoteUser.isAdminOfCollective(account)) {
+        accountOrConditions.push({ CollectiveId: account.id });
+        if (args.includeChildrenAccounts) {
+          const childIds = await account.getChildren().then(children => children.map(child => child.id));
+          accountOrConditions.push(...childIds.map(id => ({ CollectiveId: id })));
+        }
+        if (args.includeHostedAccounts && account.isHostAccount) {
+          const hostedAccounts = await account.getHostedCollectives({ attributes: ['id'] });
+          accountOrConditions.push(...hostedAccounts.map(hostedAccount => ({ CollectiveId: hostedAccount.id })));
+        }
```
Betree, line 76: If you're fetching 30 accounts here, you'll basically trigger 60 additional SQL queries. This is a massive cost that we can easily avoid: 

```suggestion
        accountOrConditions.push({ CollectiveId: account.id });
        if (args.includeChildrenAccounts) {
          accountOrConditions.push({ ParentCollectiveId: account.id });
        }
        if (args.includeHostedAccounts && account.isHostAccount) {
          accountOrConditions.push({ HostCollectiveId: account.id, isActive: true });
        }
```

test/server/graphql/v2/collection/ActivitiesCollectionQuery.test.ts
```diff
@@ -114,45 +110,6 @@ describe('server/graphql/v2/collection/ActivitiesCollection', () => {
...
-    it('Hosted', async () => {
-      const variables = { account: { legacyId: collective.id }, attribution: 'HOSTED_ACCOUNTS' };
-      const result = await graphqlQueryV2(activitiesCollectionQuery, variables, admin);
-      result.errors && console.error(result.errors);
-      expect(result.errors).to.not.exist;
-      expect(result.data.activities.totalCount).to.eq(1);
-      expect(result.data.activities.nodes[0].host.slug).to.eq(collective.slug);
-    });
-  });
-
```
Betree, line 155: Maybe we can update that to test our new `includeChildrenAccounts` and `includeHostedAccounts` rather than just removing the tests?

Betree (changes requested):

server/graphql/v2/query/collection/ActivitiesCollectionQuery.ts
SudharakaP: @Betree : For the `if (args.includeHostedAccounts && account.isHostAccount) {` part this makes sense. But do you think we could leave the `if (args.includeChildrenAccounts) {` as is? Note that `Activity` doesn't include the `ParentCollectiveId`.

Betree: You can join on `CollectiveId` them

test/server/graphql/v2/collection/ActivitiesCollectionQuery.test.ts
SudharakaP: Done. 👍🏽

server/graphql/v2/query/collection/ActivitiesCollectionQuery.ts
```diff
@@ -47,36 +53,38 @@ const ActivitiesCollectionQuery = {
...
-    const getDistinctFromCondition = value => ({ [Op.or]: [{ [Op.is]: null }, { [Op.ne]: value }] }); // Ideally, we should use `IS DISTINCT FROM` (https://wiki.postgresql.org/wiki/Is_distinct_from), but it's not supported by Sequelize yet: https://github.com/sequelize/sequelize/issues/12612
+    const where = { [Op.or]: accountOrConditions };
+    const include = [];
+    for (const account of accounts) {
+      if (isRootUser || req.remoteUser.isAdminOfCollective(account)) {
+        accountOrConditions.push({ CollectiveId: account.id }, { FromCollectiveId: account.id });
+        if (args.includeChildrenAccounts) {
+          accountOrConditions.push({ '$Collective.ParentCollectiveId$': account.id });
+          include.push({
+            model: models.Collective,
```
Betree, line 74: Cheap optimization

```suggestion
            model: models.Collective,
            attributes: [],
```

server/models/index.js
```diff
@@ -82,6 +82,8 @@ export function setupModels() {
   /**
    * Relationships
    */
+  // Activity
+  m.Activity.hasMany(m.Collective, { foreignKey: 'ParentCollectiveId', as: 'collective' });
```
Betree, line 86: This doesn't look necessary, but even if it is the `as` is wrong: you want `activity.getCollective` to return the collective, not the parent.

Betree (approved):
Looking good :+1:

server/models/index.js
SudharakaP: Ops, I was experimenting with this one and forgot to remove it. So yes this isn't necessary. 👍🏽

server/graphql/v2/query/collection/ActivitiesCollectionQuery.ts
```diff
@@ -67,15 +67,18 @@ const ActivitiesCollectionQuery = {
...
+
+        // Include all activities related to the account's children
         if (args.includeChildrenAccounts) {
           accountOrConditions.push({ '$Collective.ParentCollectiveId$': account.id });
-          include.push({
-            model: models.Collective,
-            attributes: [],
-            required: true,
-          });
+          if (include.length === 0) {
```
SudharakaP, line 84: @Betree : Don't quite understand why we need this condition? I think `include` is always empty.

Betree: @SudharakaP No, because we're in a loop. The previous version of the code pushed a new `include` for each account, to 10 accounts resulted in 10 includes

SudharakaP: Ops, I didn't see the loop. 🤯

PR #7908, opened 2022-09-02
Do not send unsubscribe link in transactional emails

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5876


Betree (approved):
Looks good! We should also try to improve the error message in the frontend, Sentry logs show quite a few people are clicking on them from old messages and it currently shows an empty message.

PR #7909, opened 2022-09-02
Fix: Ready to Pay filter to consider expense.currency

kewitz:


server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -57,12 +65,25 @@ const updateFilterConditionsForReadyToPay = async (where, include, host): Promis
     // AND ((CollectiveId = 1 AND amount < 5000) OR (CollectiveId = 2 AND amount < 3000))
     const collectiveIds = uniq(expensesWithoutPendingTaxForm.map(e => e.CollectiveId));
     const balances = await getBalancesWithBlockedFunds(collectiveIds); // TODO: move to new balance calculation v2 when possible
-    where[Op.and].push({
-      [Op.or]: Object.values(balances).map(({ CollectiveId, value }) => ({
-        CollectiveId,
-        amount: { [Op.lte]: value },
-      })),
-    });
+    const fxMap = await loadFxRatesMap(
```
znarf, line 68:
```suggestion
    const fxRates = await loadFxRatesMap(
```

znarf: Or

```suggestion
    const fxRatesMap = await loadFxRatesMap(
```

znarf (approved):
Looks a bit complicated but I don't have a better solution.

Good for me!

PR #7911, opened 2022-09-02
includeChildren parameter in AccountStats

znarf:
And refactor

### DIFF
--- a/server/lib/budget.js
+++ b/server/lib/budget.js
@@ -8,7 +8,7 @@ import { getFxRate } from './currency';
 const { CREDIT, DEBIT } = TransactionTypes;
 const { PROCESSING, SCHEDULED_FOR_PAYMENT } = expenseStatus;
 
-async function computeSumFromResults(results, currency) {
+async function sumTransactionsInCurrency(results, currency) {
   let total = 0;
 
   for (const result of Object.values(results)) {
@@ -67,7 +67,7 @@ export async function getBalanceAmount(
   });
 
   // Sum and convert to final currency
-  const value = await computeSumFromResults(results, currency);
+  const value = await sumTransactionsInCurrency(results, currency);
 
   return { value, currency };
 }
@@ -144,7 +144,7 @@ export async function getTotalAmountReceivedAmount(
   });
 
   // Sum and convert to final currency
-  const value = await computeSumFromResults(results, currency);
+  const value = await sumTransactionsInCurrency(results, currency);
 
   return { value, currency };
 }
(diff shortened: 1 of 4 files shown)


znarf:
@kewitz can you look into this?

server/lib/budget.js
```diff
@@ -8,19 +8,45 @@ import { getFxRate } from './currency';
 const { CREDIT, DEBIT } = TransactionTypes;
 const { PROCESSING, SCHEDULED_FOR_PAYMENT } = expenseStatus;
 
+async function computeSumFromResults(results, currency) {
```
kewitz, line 11: Maybe name something related to currency conversion: `sumTransactionsInCurrency`.

kewitz (approved):
LGTM. It would be nice to improve the `lib/budget` test suite, but this can be tackled later.

server/models/Expense.js
```diff
@@ -531,28 +538,25 @@ function defineModel() {
       INNER JOIN "Transactions" t
         ON t."ExpenseId" = e."id"
       INNER JOIN "Collectives" c
-        ON (
-          c."id" = $collectiveId
-          ${includeChildren ? `OR c."ParentCollectiveId" = $collectiveId` : ``}
-        )
-        AND c."deletedAt" IS NULL
+        ON c."id" = t."CollectiveId" AND c."deletedAt" IS NULL
```
Betree, line 541: I don't get the reason for this change; unless it was buggy, it seems less verbose & more performant to me to have a simple SQL `id = :id OR ParentCollectiveId = :id` than having 2 queries to get children then filter.

server/lib/budget.js
```diff
@@ -276,6 +280,9 @@ async function sumCollectivesTransactions(
 
   if (ids) {
     where.CollectiveId = ids;
+    if (excludeCrossCollectiveTransactions) {
+      where.FromCollectiveId = { [Op.notIn]: ids };
+    }
```
Betree, line 285: Nice :+1:

Betree (commented):
Would be nice to link an issue for context. I also agree that we should have a plan to add some tests given that small changes in this part of the code can easily introduce regressions.

server/models/Expense.js
kewitz: It's personal preference, the idea is to simplify the query by not mixing filtering logic with the join operation.
TBH, I'm not sure how this will affect performance, but I can analyze it further if necessary.

Betree: Filtering in `JOIN` is fine IMO, unless I'm missing something I don't think it's an anti-pattern. We can move the condition to `WHERE` if it's clearer.

Altogether I don't think this refactor adds any clarity; I find it more difficult to read a 7 lines JS filter (that gets duplicated in another function below) than a single-line SQL filter. I don't think there's a massive performance difference, but it's almost certain that these 2 queries will be slower than a single (supposedly well-indexed) one. Also, "No need to touch it if it works" :smile:

PR #7923, opened 2022-09-10
Create greeting partial

nosro:
Addresses: https://github.com/opencollective/opencollective/issues/5872

This will allow more flexible greetings in email templates. A recipient name is used, if provided.

If a collective - wide notification is sent that collective name will be used rather than individual names.

Caveat: I have not tested this locally. I'm not quite at the point where I am able to do that. So there is some small risk, I think. If you are able to verify this change first, that would be good.

### DIFF
--- a/server/lib/notifications/email.ts
+++ b/server/lib/notifications/email.ts
@@ -45,6 +45,8 @@ export const notify = {
   ) {
     const userId = options?.user?.id || options?.userId || activity.UserId;
     const user = options?.user || (await models.User.findByPk(userId, { include: [models.Collective] }));
+
+    // TODO We're not using the `unsubscribed` option here, we should
     const unsubscribed = await models.Notification.getUnsubscribers({
       type: activity.type,
       UserId: user.id,
--- a/server/lib/notifications/email.ts
+++ b/server/lib/notifications/email.ts
@@ -50,6 +50,11 @@ export const notify = {
       CollectiveId: options?.collective?.id || activity.CollectiveId,
     });
 
+    if (!activity.data.recipientName) {
+      const userCollective = await models.Collective.findByPk(user.CollectiveId);
+      activity.data.recipientName = userCollective.name;
+    }
+
     const isTransactional = TransactionalActivities.includes(activity.type);
     if (unsubscribed.length === 0) {
       debug('notifying.user', user.id, user && user.email, activity.type);
@@ -116,6 +121,8 @@ export const notify = {
           CollectiveId: collective.ParentCollectiveId ? [collective.ParentCollectiveId, collective.id] : collective.id,
           role,
         });
+
+    activity.data.recipientName = activity.data.fromCollective ? activity.data.fromCollective : collective.name;
 
     return notify.users(users, activity, { ...options, collective });
   },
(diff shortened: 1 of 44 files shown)


nosro:
If this looks like the way to go, I can update all the email templates with a greeting as part of this PR, or a follow-up.

nosro:
I can't decipher the automatic check GraphQL Inspector, "Error: Resource not accessible by integration".

Perhaps that is due to a remote PR from the fork.

nosro:
I think I've made an error in getting the user name. It looks like that might reside in the "userCollective". I'll make a change to get that.

Betree (commented):
Just a note to say that the code looks very good so far. Let me know if you need any help!

templates/partials/greeting.hbs
```diff
@@ -0,0 +1,5 @@
+{{#if recipientName ~}}
```
Betree, line 1: I didn't know handlebars had this `~` feature to ignore [whitespaces](https://handlebarsjs.com/guide/expressions.html#whitespace-control), this is awesome!

templates/partials/greeting.hbs
```diff
@@ -0,0 +1,5 @@
...
+<p>Hi {{recipientName}},</p>
+{{else ~}}
+<p>Hi,</p>
+{{/if ~}}
```
Betree, line 5: Seeing this in context now, I realize it will add unnecessary markup in many places and we'll end up with things like `<h2><p>name</p></h2>`. I wonder if we shouldn't be unopinionated here and just go with:

```suggestion
{{#if recipientName ~}}
Hi {{recipientName}},
{{else ~}}
Hi,
{{/if ~}}
```

@nosro what do you think?

server/lib/notifications/email.ts
```diff
@@ -48,6 +48,11 @@ export const notify = {
       CollectiveId: options?.collective?.id || activity.CollectiveId,
     });
 
+    if (!activity.data.recipientName) {
+      const userCollective = await models.Collective.findByPk(user.CollectiveId);
+      activity.data.recipientName = userCollective.name;
```
Betree, line 53: To add support for legal name:

```suggestion
      activity.data.recipientName = userCollective.name || userCollective.legalName;
```

Betree (commented):
I'm sorry for the late review on this, I went into vacation then forgot to followup on the PR when I came back! @nosro thanks for iterating, it looks really good and overall I have no blocking feedback. Feel free to let me know if you don't have the time to iterate on the comments below, I can take over if that's the case.

Something I forgot to spec in https://github.com/opencollective/opencollective/issues/5872: it would be great to have a fallback on `legalName` when available. It's a private field where people can set their "real" account name, usually used to generate documents like invoices.

I'll add a suggestion below to add this, but it doesn't need to be part of this PR.

Betree:
Hi @nosro, need any help on that one? Happy to take it from here if you don't have the time :slightly_smiling_face:

Betree:
Moved to https://github.com/opencollective/opencollective-api/pull/8204; I'll take it from there. thanks for your contribution @nosro!

PR #7930, opened 2022-09-12
Ignore Parent to Children contributions in contributionsAmount and contributionsAmountTimeSeries

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5950


znarf (changes requested):
Looks good, but we also need to handle transactions from the children to the parent.

PR #7933, opened 2022-09-12
Do Not Send Attachment Notification if Receipt is Already Attached

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5951

### DIFF
--- a/templates/emails/virtualcard.purchase.hbs
+++ b/templates/emails/virtualcard.purchase.hbs
@@ -11,7 +11,7 @@ Subject: Virtual Card Purchase
 <p>
 A card attached to {{collective.name}} was charged {{currency amount currency=currency
   precision=2}}. The responsible admin {{responsibleAdmin.name}} will be asked to submit additional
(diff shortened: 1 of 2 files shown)


templates/emails/virtualcard.purchase.hbs
```diff
@@ -11,7 +11,7 @@ Subject: Virtual Card Purchase
 <p>
 A card attached to {{collective.name}} was charged {{currency amount currency=currency
   precision=2}}. The responsible admin {{responsibleAdmin.name}} will be asked to submit additional
-information. For accounting reasons, the Fiscal Host will need a description and receipt. The card may be suspended if receipts are not provided.
+information. For accounting reasons, the Fiscal Host will need a description and receipt. Please <a href="{{config.host.website}}/{{collective.slug}}/expenses/{{expense.id}}?edit=1">upload a receipt</a> if you already have one. The card may be suspended if receipts are not provided.
```
znarf, line 14: Would be great at some point to have the right URL for projects and events, instead of relying on the frontend redirect.

znarf (approved):
Looks good to me.

PR #7934, opened 2022-09-13
Add Collective Name to the Missing Receipt Email

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5949


templates/emails/collective.expense.missing.receipt.hbs
```diff
@@ -11,7 +11,7 @@ Subject: Action Required: Submit receipt for card expense
 <p>
 {{#if user}}
 A card attached to {{fromCollective.name}} was charged {{currency expense.amount currency=expense.currency}} and we need the responsible admin to input additional
-information.
+information. This expense was submitted to {{collective.name}}.
```
znarf, line 14: Why do we only have this under the `#if user` condition?

Betree, line 14: Would be great for this to be a link, and same for `fromCollective`

Betree (approved):

templates/emails/collective.expense.missing.receipt.hbs
znarf: Looks like this is a remain that was made useless after this refactor https://github.com/opencollective/opencollective-api/pull/6874  and can be removed if we're not using it anymore.

RichardLitt:
Ah, shoot. I edited and submitted a PR for https://github.com/opencollective/opencollective-api/pull/7936 before I saw this edit. I added data to make compiling easier, and changed the language a bit - maybe we could take some edits from that, @SudharakaP?

templates/emails/collective.expense.missing.receipt.hbs
SudharakaP: Done. 👍🏽

SudharakaP:
> Ah, shoot. I edited and submitted a PR for #7936 before I saw this edit. I added data to make compiling easier, and changed the language a bit - maybe we could take some edits from that, @SudharakaP?

I've incorporated changes from that PR here as well and removed the unnecessary condition as pointed out at, https://github.com/opencollective/opencollective-api/pull/7934#discussion_r969219392. The new template looks like,

image

templates/emails/collective.expense.missing.receipt.hbs
SudharakaP: Removed the else part. 👍🏽

RichardLitt:
The only thing I wonder is if there is a way we could mark the title and collective name with quotes. I think "Team dinner card expense" is an awkward phrase to parse. Maybe: `Please submit the receipt for the "Team Dinner" card expense` would be more natural?

SudharakaP:
> The only thing I wonder is if there is a way we could mark the title and collective name with quotes. I think "Team dinner card expense" is an awkward phrase to parse. Maybe: `Please submit the receipt for the "Team Dinner" card expense` would be more natural?

Done; so it looks;

image

RichardLitt:
Thanks. Looks good to me.

Betree (approved):

templates/emails/collective.expense.missing.receipt.hbs
```diff
@@ -5,16 +5,12 @@ Subject: Action Required: Submit receipt for card expense
...
   </h1>
 </center>
 <p>
-{{#if user}}
-A card attached to {{fromCollective.name}} was charged {{currency expense.amount currency=expense.currency}} and we need the responsible admin to input additional
-information.
-{{else}}
-You need to provide additional info about the card charge of {{currency expense.amount currency=expense.currency}} for {{fromCollective.name}}.
-{{/if}}
+A card attached to <a href="{{config.host.website}}/{{collective.slug}}">{{collective.name}}</a> was charged {{currency expense.amount currency=expense.currency}} for {{expense.description}} and we need the responsible admin to input additional information.
```
znarf, line 12: What if it's a project or an event ? The slug will not be enough.

znarf: Why have we added expense.description ?

znarf: Expense description is NOT looking like "Team Dinner" in production so it's super misleading to take this as an example.

The current expense description looks like this: "Virtual Card charge: Zoom". Quite different ...

Betree: The case for events/projects URLs is difficult to handle in emails, we usually don't have access to the parents in there; I voluntarily skipped it too in https://github.com/opencollective/opencollective-api/pull/7937. The redirect should do the job, we could iterate in the future by auto-loading parents from collectives before generating templates.

SudharakaP: @Betree : I propose a fix here, https://github.com/opencollective/opencollective-api/pull/7949 using the `getUrlPath` function. Let me know if that makes things better? 🤔 

@znarf : I've replaced the description with `fromCollective.name`.

PR #7932, opened 2022-09-12
Add graphqlV2 tier mutations

hdiniz:
Adds mutations for creating and editing tiers on the graphqlv2 endpoint.

PR #7940, opened 2022-09-13
Remove getDecodedId helper

Betree:
This helper was introduced in https://github.com/opencollective/opencollective-api/pull/2869, but its purpose is unclear:
- We want all IDs to be generated for a specific type (comment, expense, etc) so having a generic function doesn't make sense
- It's unclear what is expected from this helper, and ID V2 ? V1 ?
- It was confusing for some developers who used that instead of `idDecode` (see https://github.com/opencollective/opencollective-api/pull/7932#discussion_r969194402)

### DIFF
--- a/server/graphql/v2/identifiers.js
+++ b/server/graphql/v2/identifiers.js
@@ -86,12 +86,3 @@ export const getIdEncodeResolver =
   (type, idField = 'id') =>
   entity =>
     idEncode(entity[idField], type);
-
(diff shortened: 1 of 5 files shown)


znarf (approved):

znarf:
Seems it was introduced there https://github.com/opencollective/opencollective-api/pull/2869, doesn't make sense to me. Good change!

PR #7942, opened 2022-09-13
Automatically set transferwise.ott when connecting to Wise

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/5941


Betree (approved):
Good & simple fix :ok_hand:

PR #7944, opened 2022-09-14
Return invited expense user emails

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5873

Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/8186

### DIFF
--- a/server/graphql/loaders/collective.ts
+++ b/server/graphql/loaders/collective.ts
@@ -82,6 +82,7 @@ export default {
       }
 
       let administratedMembers = [];
+      let invitedExpenses = [];
 
       // Aggregates all the profiles linked to users
       const uniqueCollectiveIds = uniq(collectiveIds.filter(Boolean));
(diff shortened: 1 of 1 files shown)


server/graphql/loaders/collective.ts
```diff
@@ -113,15 +114,25 @@ export default {
             },
           },
         });
+
+        invitedExpenseUsers = await models.Expense.findAll({
+          attributes: ['FromCollectiveId'],
+          group: ['FromCollectiveId'],
+          raw: true,
+          mapToModel: false,
+          where: { [Op.or]: [{ FromCollectiveId: otherAccountsCollectiveIds }, { status: 'UNVERIFIED' }] },
```
Betree, line 123: :no_entry_sign: The condition here will be translated to the SQL:

```sql
SELECT "FromCollectiveId"
FROM "Expenses"
WHERE "FromCollectiveId" IN (otherAccountsCollectiveIds)
OR status = 'UNVERIFIED'
```

Then we're returning `canSeePrivateInfo` as `invitedExpenseUserIds.has(collectiveId)`, therefore leaking publicly all the emails of users who submitted an expense.

Betree (changes requested):

server/graphql/loaders/collective.ts
Betree: The variable name, `invitedExpenseUsers`, is also confusing - this function returns a list of expenses with `FromCollectiveId` set. `invitedExpenseUserIds` contains a list of collective IDs, not user IDs.

SudharakaP: Sorry, the condition should have an `Op.and` not a `Op.or`. 🤯 

As for the variable names I've renamed them as, `invitedExpenses` and `invitedExpensePayeeIds`.

Betree: So this new condition translates to:

```sql
SELECT "FromCollectiveId"
FROM "Expenses"
WHERE "FromCollectiveId" IN (otherAccountsCollectiveIds)
AND status = 'UNVERIFIED'
```

Then we're returning `canSeePrivateInfo` as `invitedExpenseUserIds.has(collectiveId)`.

If I translate that to plain English, it gives something like:
> Allow remote user to see private information for all requested accounts who have submitted unverified expenses

We need to be really careful here.

SudharakaP: Oh I see the problem here; we need a join to return only those accounts which the signed in user has admin privileges. I've added this to the SQL. Let me know if everything looks good.

Betree (changes requested):
There is no PR description or code comments about why we're doing this. At this point I have no clue about why:
> the collective admin should have been allowed to re-send the email.

Requires us to change `canSeePrivateInfo`. If doing so is really needed:
1. Please explain why
2. Make sure it's carefully tested. This function is what protects our users' private data, we shouldn't mess with it without making sure that we extend its test coverage in https://github.com/opencollective/opencollective-api/blob/2fbbcdae3552bbe64600b147ccf9193e60309956/test/server/graphql/loaders/collective.test.ts. I'm talking about both security and performances - this thing is used in many places.

Betree (changes requested):
Closing as this is touching critical parts of our permissions model without providing enough documentation about why we're doing it this way.

PR #7945, opened 2022-09-14
Use float for Amount's valueInCents

Betree:
An attempt at fixing https://github.com/opencollective/opencollective/issues/5452

I tried with the [GraphQLBigInt](https://www.the-guild.dev/graphql/scalars/docs/scalars/big-int) scalar, but it requires an additional dependency (https://github.com/ardatan/json-bigint-patch) which seems to be causing breaking changes as numbers are then parsed as big ints:

> AssertionError: expected 0n to equal +0

Returning as `float` should be safe a backward-compatible.

### DIFF
--- a/server/graphql/schemaV2.graphql
+++ b/server/graphql/schemaV2.graphql
@@ -4669,7 +4669,7 @@ A financial amount.
 type Amount {
   value: Float
   currency: Currency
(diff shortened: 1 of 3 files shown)


znarf (approved):
Good for me !

PR #7949, opened 2022-09-15
Replace expense description with `fromCollective.name`

SudharakaP:
Related to https://github.com/opencollective/opencollective-api/pull/7934/files#r971968569

### DIFF
--- a/scripts/compile-email.js
+++ b/scripts/compile-email.js
@@ -71,7 +71,6 @@ data['collective.expense.missing.receipt'] = {
   },
   collective: {
     slug: 'wwcodeaustin',
-    urlPath: '/wwcodeaustin',
     name: 'Women Who Code Austin',
   },
   fromCollective: {
--- a/scripts/compile-email.js
+++ b/scripts/compile-email.js
@@ -71,6 +71,7 @@ data['collective.expense.missing.receipt'] = {
   },
   collective: {
     slug: 'wwcodeaustin',
+    urlPath: '/wwcodeaustin',
     name: 'Women Who Code Austin',
   },
   fromCollective: {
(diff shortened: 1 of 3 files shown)


server/models/Expense.js
```diff
@@ -358,6 +358,7 @@ function defineModel() {
         ]),
         host: get(host, 'minimal'),
         collective: { ...this.collective.minimal, isActive: this.collective.isActive },
+        path: this.collective.getUrlPath(),
```
Betree, line 361: I personally think we should address this globally rather than patching emails individually. If we do want to patch them individually, then we need to figure out a better variable name: I don't think anyone will guess that `data.path` will be set to the collective URL when looking at an `expense.missing.receipt` activity.

SudharakaP: @Betree : I've renamed it to `urlPath` to make it explicit and also added with within the `collective` object. This way I think it's clearer. 

As for a global solution, I've opened a new issue and will work on it; https://github.com/opencollective/opencollective/issues/5979

scripts/compile-email.js
```diff
@@ -66,12 +66,12 @@ data['collective.expense.paid'] = {
 data['collective.expense.missing.receipt'] = {
   host: { id: 1, name: 'WWCode', slug: 'wwcode' },
   expense: {
-    description: 'Team dinner',
```
Betree, line 69: Why remove this?

Betree (commented):
Again, I'm personally not convinced about the quick hack for the parent in URL here. If we want to solve this issue (which I don't think is a priority given we have a nicely working redirect) then I'd rather solve it globally. 

No blocker for me overall.

scripts/compile-email.js
SudharakaP: @Betree : Because we are not using the `description` anymore in the email template.

SudharakaP:
> Again, I'm personally not convinced about the quick hack for the parent in URL here. If we want to solve this issue (which I don't think is a priority given we have a nicely working redirect) then I'd rather solve it globally.
> 
> No blocker for me overall.

@Betree : Makes sense. I've removed this quick hack for now as we have and issue to handle this later. So this PR only remove the `description`. Will merge if there's no objections. 👍🏽

PR #7965, opened 2022-09-17
Correct some typos in virtual card suspended email

SudharakaP:
Currently the virtual card suspended email look like this. 

image

Note the typos. With this change it looks like,

image


znarf (approved):

Betree (approved):
:+1:

PR #7972, opened 2022-09-22
Add mutation to un-host a collective by host admins

hdiniz:
Relates to https://github.com/opencollective/opencollective/issues/3330

The removeHost mutation is available both by root actions and by host admins. When successful, an activy is logged and an email email notification is sent to the collective admins with an optional message.

The collective balance must be zero for this mutation to succeed.

image

### DIFF
--- a/templates/emails/collective.unhosted.hbs
+++ b/templates/emails/collective.unhosted.hbs
@@ -1,0 +1,21 @@
+Subject: Important: {{{collective.name}}} has been un-hosted by {{{host.name}}}
+
+{{> header}}
+
+<p style="font-size: 17px; padding: 1em;">
+  Hi {{collective.name}},
+  <br /><br />
+  Your fiscal host, <a href="{{config.host.website}}/{{host.slug}}">{{{host.name}}}</a>, has un-hosted your Collective.
+  <br/><br/>
+  If you have additional questions about this, please <a href="{{config.host.website}}/{{host.slug}}/contact">contact</a> your fiscal host.
+  {{#if message}}
+    <br /><br />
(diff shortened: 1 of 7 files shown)


Betree (commented):
The code looks perfect and the tests are on point! 

I just thought about a rule that we should enforce now that we're going to let host admins do this by themselves: we don't want them to unhost events & projects without unhosting the parent. I suggest adding something like this in the `removeHost` mutation:

```es6
if (account.ParentCollectiveId) {
  throw new ValidationFailed(`Cannot unhost projects/events with a parent. Please unhost the parent instead.`);
}
```

(feel free to figure a better error message)

Betree:
Note about my previous comment: this is already enforced in the frontend, since we only show the "More actions" button in the host admin for collectives & funds: https://github.com/opencollective/opencollective-frontend/blob/45898f71823ff05132d35e6d2f7f4540e50f42be/components/host-dashboard/HostAdminCollectiveCard.js#L127

templates/emails/collective.unhosted.hbs
```diff
@@ -0,0 +1,21 @@
+Subject: Important: {{{collective.name}}} has been un-hosted by {{{host.name}}}
+
+{{> header}}
+
+<p style="font-size: 17px; padding: 1em;">
+  Hi {{collective.name}},
+  <br /><br />
+  Your fiscal host, <a href="{{config.host.website}}/{{host.slug}}">{{{host.name}}}</a>, has un-hosted your Collective.
```
Betree, line 8: :ice_cream: Fyi, we now have a partial for that:

```suggestion
  Your fiscal host, {{> linkCollective collective=host}}, has un-hosted your Collective.
```

Betree (approved):

PR #7976, opened 2022-09-22
Make sure to use `expiryDate` instead of `endDate` for payment method expiry check

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5988

⚠️ Require: https://github.com/opencollective/opencollective-frontend/pull/8221


Betree (commented):
Looks safe to me, all places using this have a `try ... catch` for this case except `getSourcePaymentMethodFromCreateArgs`, but it sounds like the right behavior to block the creation of gift cards with an expired credit card. I'll let @znarf give a second review.

znarf (approved):
Good for me.

PR #7979, opened 2022-09-23
Use less arbitrary rule in checkHasBalanceToPayExpense

kewitz:
I've been receiving some errors related to how we calculate how much balance a collective must have to cover an expense in a different currency.
We've been using a 20% rule across the board, which is not fair when paying to relatively stable currencies.

My new proposal considers the standard deviation (2*σ should cover 95% of rates) of the exchange rate between currencies to calculate a safe margin for the expense amount.

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -1626,7 +1626,7 @@ const generateInsufficientBalanceErrorMessage = ({
   }
 
   if (!isSameCurrency) {
-    msg += `. For expenses submitted in a different currency than the collective, an error margin of 2σ to the latest rate is applied. The maximum amount that can be paid is ${formatCurrency(
+    msg += `. For expenses submitted in a different currency than the collective, an error margin is applied to accommodate for fluctuations. The maximum amount that can be paid is ${formatCurrency(
       Math.round(balance * rate),
       expenseCurrency,
     )}`;
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -9,6 +9,7 @@ import {
   isBoolean,
   isEqual,
   isNil,
+  isNumber,
   keyBy,
   mapValues,
   omitBy,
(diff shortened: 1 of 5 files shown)


server/graphql/common/expenses.ts
```diff
@@ -1612,21 +1612,23 @@ const generateInsufficientBalanceErrorMessage = ({
...
   if (fees) {
-    msg += `, Estimated ${feesName} fees: ${formatCurrency(fees, currency)}`;
+    msg += `, Estimated ${feesName} fees: ${formatCurrency(fees, expenseCurrency)}`;
   }
 
   if (!isSameCurrency) {
-    msg += `. For expenses submitted in a different currency than the collective, an error margin of 20% is applied. The maximum amount that can be paid is ${formatCurrency(
-      Math.round(balance / 1.2),
-      currency,
+    msg += `. For expenses submitted in a different currency than the collective, an error margin of 2σ to the latest rate is applied. The maximum amount that can be paid is ${formatCurrency(
```
Betree, line 1629: I'm not sure all users will be comfortable with the concept of standard deviation. I'd argue for simplification:

```suggestion
    msg += `. For expenses submitted in a different currency than the collective, an error margin is applied to accommodate for fluctuations. The maximum amount that can be paid is ${formatCurrency(
```

server/graphql/common/expenses.ts
```diff
@@ -1643,21 +1645,28 @@ export const checkHasBalanceToPayExpense = async (
...
 
-  // Ensure the collective has enough funds to pay the expense, with an error margin of 20% of the expense amount
-  // to account for fluctuating rates. Example: to pay for a $100 expense in euros, the collective needs to have at least $120.
-  const getMinExpectedBalance = amountToPay => (isSameCurrency ? amountToPay : Math.round(amountToPay * 1.2));
+  // Ensure the collective has enough funds to pay the expense, with an error margin of 2σ (standard deviations) the exchange rate of past 5 days
+  // to account for fluctuating rates. (2σ counts for 95% of rates of the given period)
+  const getMinExpectedBalance = amountToPayInExpenseCurrency =>
+    isSameCurrency
+      ? amountToPayInExpenseCurrency
+      : Math.round(amountToPayInExpenseCurrency / (exchangeStats.latestRate - exchangeStats.stddev * 2));
```
Betree, line 1658: `getMinExpectedBalance` can return null if we don't have any cached information between from/to currencies. I would move all this to a separate function to isolate complexity, with an eventual fallback to the legacy rule of 20% if we're not able to compute anything better,

Betree (commented):
Nice improvement, 20% was indeed too aggressive for most cases.

server/graphql/common/expenses.ts
kewitz: @Betree, I'm not abstracting this to a global scope because I don't think this will be very useful elsewhere.
I converted this to an assertion function and included the error message format in it to simplify it further.

PR #7981, opened 2022-09-23
Add support for root action flags

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5374

Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/8222

### DIFF
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -1177,7 +1177,7 @@ const CollectiveFields = () => {
     isTrustedHost: {
       type: new GraphQLNonNull(GraphQLBoolean),
       description: 'Returns whether this host is trusted or not',
-      resolve: collective => get(collective, 'data.isTrustedHost', false),
+      resolve: collective => Boolean(get(collective, 'data.isTrustedHost')),
     },
     canApply: {
       description: 'Returns whether this host accepts applications for new collectives',
@@ -1219,20 +1219,6 @@ const CollectiveFields = () => {
       type: GraphQLBoolean,
       resolve(collective) {
         return Boolean(collective.deactivatedAt && !collective.isActive);
-      },
-    },
-    isDeleted: {
-      type: GraphQLBoolean,
-      description: 'Returns whether this account is deleted',
-      resolve(collective) {
-        return Boolean(collective.deletedAt);
-      },
-    },
-    isBanned: {
-      type: GraphQLBoolean,
-      description: 'Returns whether this account is banned',
-      resolve(collective) {
-        return Boolean(collective.data?.isBanned);
       },
     },
     isApproved: {
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -1174,6 +1174,11 @@ const CollectiveFields = () => {
         return collective.isHost();
       },
     },
+    isTrustedHost: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      description: 'Returns whether this host is trusted or not',
+      resolve: collective => get(collective, 'data.isTrustedHost', false),
+    },
     canApply: {
       description: 'Returns whether this host accepts applications for new collectives',
       type: GraphQLBoolean,
@@ -1214,6 +1219,20 @@ const CollectiveFields = () => {
       type: GraphQLBoolean,
       resolve(collective) {
         return Boolean(collective.deactivatedAt && !collective.isActive);
+      },
+    },
+    isDeleted: {
+      type: GraphQLBoolean,
+      description: 'Returns whether this account is deleted',
+      resolve(collective) {
+        return Boolean(collective.deletedAt);
+      },
+    },
+    isBanned: {
+      type: GraphQLBoolean,
+      description: 'Returns whether this account is banned',
+      resolve(collective) {
+        return Boolean(collective.data?.isBanned);
       },
     },
     isApproved: {
--- a/server/graphql/v1/queries.js
+++ b/server/graphql/v1/queries.js
@@ -932,13 +932,9 @@ const queries = {
         type: GraphQLBoolean,
         description: 'Filter on whether account is a host',
       },
-      includeDeleted: {
-        type: GraphQLBoolean,
-        description: 'Included collectives which are soft deleted',
-      },
       includeArchived: {
         type: GraphQLBoolean,
-        description: 'Included collectives which are soft archived',
+        description: 'Included collectives which are archived',
       },
       onlyActive: {
         type: GraphQLBoolean,
@@ -976,7 +972,6 @@ const queries = {
         parentCollectiveIds,
         skipRecentAccounts,
         skipGuests,
-        includeDeleted,
         includeArchived,
       } = args;
       const cleanTerm = term ? term.trim() : '';
@@ -1010,7 +1005,6 @@ const queries = {
           isHost,
           skipRecentAccounts,
           skipGuests,
-          includeDeleted,
           includeArchived,
         });
         return generateResults(collectives, total);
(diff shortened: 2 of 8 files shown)


server/graphql/v1/CollectiveInterface.js
```diff
@@ -1174,6 +1174,11 @@ const CollectiveFields = () => {
         return collective.isHost();
       },
     },
+    isTrustedHost: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      description: 'Returns whether this host is trusted or not',
+      resolve: collective => get(collective, 'data.isTrustedHost', false),
```
Betree, line 1180: :ice_cream: `data` is unstructured, if someone sets `data.isTrustedHost` to `null` we'll have a crash here. It's safer to do:

```suggestion
      resolve: collective => Boolean(get(collective, 'data.isTrustedHost')),
```

server/graphql/v2/mutation/RootMutations.ts
```diff
@@ -49,6 +49,61 @@ const BanAccountResponse = new GraphQLObjectType({
...
+        type: GraphQLBoolean,
+        description: 'Specify whether the account is a trusted host',
+      },
+    },
+    async resolve(_: void, args, req: express.Request): Promise<Record<string, unknown>> {
+      checkRemoteUserCanRoot(req);
+      const account = await fetchAccountWithReference(args.account, { throwIfMissing: true, paranoid: false });
+
+      if (args.isDeleted && !account.deletedAt) {
+        await account.destroy();
```
Betree, line 82: :no_entry_sign: We can't just delete the account like that; what if it has attached transactions, hosted collectives, etc? Also, the associated data needs to be deleted too. The `deleteCollective` mutation is supposed to take care of these aspects, let's find a way to re-use the code here.

I suspect this will make the implementation of "restore" more difficult, so I suggest opening an issue to discuss this if you feel that it's too big of a scope.

server/graphql/v2/mutation/RootMutations.ts
```diff
@@ -49,6 +49,61 @@ const BanAccountResponse = new GraphQLObjectType({
...
+        // We have freed the slug, so we have to restore it to its original value
+        const timestampPart = account.slug.match(/-\d+$/)[0];
+        await account.update({ slug: account.slug.replace(timestampPart, '') });
+      }
+
+      if (args.isArchived && !account.deactivatedAt) {
+        await account.update({ deactivatedAt: Date.now() });
+      } else if (!args.isArchived && account.deactivatedAt) {
+        await account.update({ deactivatedAt: null });
+      }
```
Betree, line 95: :no_entry_sign: That means that, if I call this mutation on an archived account:

```gql
editAccountFlags(account, { isTrustedHost: false }) { ... }
```

it will unarchive the account, since `args.isArchived` will be falsy (`!undefined === true`). Let's make this one and the others around a bit more flexible:

```suggestion
      } else if (args.isArchived === false && account.deactivatedAt) {
        await account.update({ deactivatedAt: null });
      }
```

Betree, line 95: :no_entry_sign: Same as delete, we have a process to (un)archive accounts that take care of stuff like making sure we empty the balance first. Let's not bypass that.

server/graphql/v2/query/collection/AccountsCollectionQuery.ts
```diff
@@ -29,6 +29,14 @@ const AccountsCollectionQuery = {
...
       description: 'Only return Fiscal Hosts accounts if true',
     },
+    includeDeleted: {
+      type: GraphQLBoolean,
+      description: 'Included collectives which are soft deleted',
+    },
+    includeArchived: {
+      type: GraphQLBoolean,
+      description: 'Included collectives which are soft archived',
+    },
```
Betree, line 39: :no_entry_sign: These params (at least `includeDeleted`) should be root only. We don't want random users to look at deleted entries in our DB.

Betree (changes requested):
See the blocking feedback below. Also, let's make sure we add tests in `test/server/graphql/v2/mutation/RootMutations.test.ts` when we implement critical actions like that.

server/graphql/v1/queries.js
```diff
@@ -932,6 +932,14 @@ const queries = {
         type: GraphQLBoolean,
         description: 'Filter on whether account is a host',
       },
+      includeDeleted: {
+        type: GraphQLBoolean,
+        description: 'Included collectives which are soft deleted',
+      },
+      includeArchived: {
+        type: GraphQLBoolean,
+        description: 'Included collectives which are soft archived',
```
znarf, line 941: soft archived?

znarf:
Should we put deleting out of scope for now? It seems it's making the issue over complex.

Betree:
> Should we put deleting/deleted out of scope for now? It seems it's making the issue over complex.

I agree. And I'd like to better understand the use case for it: @SudharakaP explained to me that Shannon was currently using the "Search & Destroy" tool to delete accounts, but I don't get why support needs that; people can delete their accounts themselves from the interface.

server/graphql/v1/CollectiveInterface.js
SudharakaP: Done. 👍🏽

server/graphql/v2/mutation/RootMutations.ts
SudharakaP: Yes I've corrected this and the other condition related to trusted host to check explicitly for `isArchived` and `isTrustedHost`

SudharakaP: Thanks, I've used the exiting functions so that this is taken care of. 👍🏽

SudharakaP:
> Should we put deleting/deleted out of scope for now? It seems it's making the issue over complex.

> I agree. And I'd like to better understand the use case for it: @SudharakaP explained to me that Shannon was currently using the "Search & Destroy" tool to delete accounts, but I don't get why support needs that; people can delete their accounts themselves from the interface.

Yeah I've removed the deleting and banning for now from this PR. Since if we are supporting this we need to also need to think about restoring or un-banning which I think is too complex for this PR. 

@Betree : As for why `Search & Destroy` was used, for example this support ticket; https://opencollective.freshdesk.com/a/tickets/119533 the user has lost access to one of their accounts and needs to delete this duplicate account. So they cannot do it themselves. I am wondering how we should handle cases like these?

Betree:
> So they cannot do it themselves. I am wondering how we should handle cases like these?

@SudharakaP For now, they should escalate to engineering. Not only because "Ban accounts" is not the right tool, but also because there are some security aspects that we need to consider when someone who's not able to login asks us to delete the account (how can we be sure they own it?).

If we get many requests like that, then maybe we want to introduce a new dedicated action.

server/graphql/v2/mutation/RootMutations.ts
```diff
@@ -49,6 +50,39 @@ const BanAccountResponse = new GraphQLObjectType({
...
+
+      if (args.isArchived && !account.deactivatedAt) {
+        await archiveCollective(_, account, req);
+      } else if (args.isArchived === false && account.deactivatedAt) {
+        await unarchiveCollective(_, account, req);
+      }
+
+      if (Boolean(args.isTrustedHost) !== Boolean(account.data?.isTrustedHost)) {
+        await account.update({ data: { ...account.data, isTrustedHost: args.isTrustedHost } });
+      }
```
Betree, line 82: Same issue as https://github.com/opencollective/opencollective-api/pull/7981#discussion_r980865197, we don't want hosts to be marked as "Untrusted" if I call the mutation like this:

```es6
editAccountFlags(account: 'foundation', isArchived: false)
```

So:
```suggestion
      if (!isNil(args.isTrustedHost) && Boolean(args.isTrustedHost) !== Boolean(account.data?.isTrustedHost)) {
        await account.update({ data: { ...account.data, isTrustedHost: args.isTrustedHost } });
      }
```

Betree (changes requested):
Small blocking feedback below, the rest looks good!

PR #7982, opened 2022-09-23
Remove duplicate timestamp when freeing the slug

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5996


znarf (approved):
Looks good

PR #7984, opened 2022-09-26
Make sure to free slug after banning

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5987


Betree (commented):
We can add a test in `test/sql/ban-collectives.sql.test.js` to prevent future regression

SudharakaP:
> We can add a test in `test/sql/ban-collectives.sql.test.js` to prevent future regression

@Betree : Thanks. yes, I've added a test here. 👍🏽

Betree (approved):
Nice :100:

PR #8000, opened 2022-09-28
Introduce simplified activity indexes

Betree:
Should resolve https://github.com/opencollective/opencollective/issues/5959 (results with a local prod copy are promising)

The `collective.transaction.created` activity makes up to 80% of all activities created; yet we're not surfacing it in the activity log because it creates too much noise. Indexes that ignore this activity are drastically smaller and allow for quickly pre-filtered queries in the activity log.

### DIFF
--- a/migrations/20220928111043-add-simplified-activity-indexes.js
+++ b/migrations/20220928111043-add-simplified-activity-indexes.js
@@ -1,0 +1,31 @@
+'use strict';
+
+module.exports = {
+  async up(queryInterface) {
+    // FromCollectiveId
+    await queryInterface.sequelize.query(`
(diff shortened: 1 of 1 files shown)


hdiniz (approved):
Using pagination that returns a total count can be problematic in large tables, as it requires postgres to scan the entire result set to get the total count even if a page will only contains a few items.

If the index change is not enough here, we might want to look into using a cursor based pagination in the future.

Betree:
> Using pagination that returns a total count can be problematic in large tables, as it requires postgres to scan the entire result set to get the total count even if a page will only contains a few items.
> 
> If the index change is not enough here, we might want to look into using a cursor based pagination in the future.

@hdiniz True. I'm trying to stay as close as possible to our usual way of doing things, but this is not a usual table and we'll probably need to iterate with more advanced technics at one point or the other.

PR #8002, opened 2022-09-28
Handle virtual card deletion and subscription cancelation on host change

hdiniz:
Related https://github.com/opencollective/opencollective/issues/5993
Related https://github.com/opencollective/opencollective/issues/5971


server/models/Collective.js
```diff
@@ -2409,6 +2409,28 @@ function defineModel() {
...
+        )
+      `,
+      {
+        type: QueryTypes.UPDATE,
+        replacements: [this.id],
+      },
+    );
+
+    const virtualCards = await models.VirtualCard.findAll({ where: { CollectiveId: this.id } });
+    await Promise.all(virtualCards.map(virtualCard => virtualCard.destroy()));
```
Betree, line 2432: :ice_cream: Personal pref because many collectives don't have many virtual cards attached, but an alternative is:

```suggestion
    await models.VirtualCard.destroy({ where: { CollectiveId: this.id } });
```

server/models/Collective.js
```diff
@@ -2409,6 +2409,28 @@ function defineModel() {
...
+            "Subscriptions"."isManagedExternally" AND
+            "Subscriptions"."isActive" AND
+            "Orders"."CollectiveId" = ?
+        )
+      `,
+      {
+        type: QueryTypes.UPDATE,
+        replacements: [this.id],
+      },
+    );
```
Betree, line 2429: :ice_cream: I would isolate this query somewhere in a `cancelNonTransferableSubscriptions` function, with a comment like: `Cancels all subscription that cannot be transferred when changing hosts (i.e. PayPal)`.

As an alternative if we don't do this, we should at least add a comment to explain why we're canceling only `isManagedExternally`; that might not be clear for someone not familiar with this system.

server/models/Order.js
```diff
@@ -446,6 +447,10 @@ function defineModel() {
     });
   };
 
+  Order.cancelActiveOrdersByTierId = function (tierId) {
+    return Order.update({ status: OrderStatuses.CANCELLED }, { where: { TierId: tierId } });
+  };
+
```
Betree, line 453: See the feedback in https://github.com/opencollective/opencollective-api/pull/8004#discussion_r983179952

cron/hourly/70-cancel-subscriptions-for-cancelled-orders.ts
```diff
@@ -28,24 +29,38 @@ const getHostFromOrder = async order => {
   return models.Collective.findByPk(transactions[0].HostCollectiveId);
 };
 
+const getOrderCancelationReason = (collective, order, orderHost) => {
+  if (!order.Tier) {
+    return `Order tier deleted`;
```
Betree, line 34: :no_entry_sign: That's not a safe assumption: when you make a "Custom contribution" (e.g. https://opencollective.com/babel/donate) orders are created without a tier (`TierId = null`). We need to find a better way of doing this, one could be:

```suggestion
  if (order.TierId && !order.Tier) {
    return `Order tier deleted`;
```

Note that we have an open discussion about changing this in https://github.com/opencollective/opencollective/issues/3055, but with so many things relying on this pattern it's unlikely to change anytime soon.

cron/hourly/70-cancel-subscriptions-for-cancelled-orders.ts
```diff
@@ -28,24 +29,38 @@ const getHostFromOrder = async order => {
...
+  } else if (collective.deactivatedAt) {
+    return `@${collective.slug} archived their account`;
+  } else if (!collective.HostCollectiveId) {
+    return `@${collective.slug} was un-hosted`;
+  } else if (collective.HostCollectiveId !== orderHost.id) {
+    return `@${collective.slug} changed host`;
+  } else {
+    return `Order cancelled`;
+  }
```
Betree, line 43: This `getOrderCancelationReason` function is really cool, I'm glad that we're able to extract the reason with such a simple code.

server/graphql/v1/mutations/collectives.js
```diff
@@ -514,8 +514,9 @@ export async function archiveCollective(_, args, req) {
   }
 
   // TODO: cascade deactivation to EVENTs and PROJECTs?
+  await collective.changeHost(null);
 
-  return collective.update({ isActive: false, deactivatedAt: Date.now(), approvedAt: null, HostCollectiveId: null });
+  return collective.update({ isActive: false, deactivatedAt: Date.now(), approvedAt: null });
```
Betree, line 519: :ice_cream: Good call to re-use `changeHost` here. We can probably simplify further given that `changeHost` resets `isActive`/`approvedAt`:

```suggestion
  return collective.update({ deactivatedAt: Date.now() });
```

cron/hourly/70-cancel-subscriptions-for-cancelled-orders.ts
```diff
@@ -79,7 +97,7 @@ export async function run() {
             subscription: order.Subscription,
             collective: order.collective.minimal,
             fromCollective: order.fromCollective.minimal,
-            reasonCode: 'ARCHIVED_ACCOUNT',
+            reasonCode: 'ORDER_CANCELLED',
```
Betree, line 100: Would be great to keep a rich value for this one, we want to use it to track statistics on **why** orders get cancelled. An option can be, from `getOrderCancelationReason`, to return a tuple like `[reasonCode, resonMessage]`

cron/hourly/70-cancel-subscriptions-for-cancelled-orders.ts
```diff
@@ -28,24 +29,38 @@ const getHostFromOrder = async order => {
...
+};
+
 /**
  * When archiving collectives. we need to make sure subscriptions managed externally are properly cancelled
  * by calling the right service method (PayPal). We do this asynchronously to properly deal with rate limiting and
  * performance constraints.
  */
 export async function run() {
   const orphanOrders = await models.Order.findAll({
+    where: { status: OrderStatuses.CANCELLED },
```
Betree, line 53: Are we ready for this? It looks like some places where we're supposed to cancel the orders (such as `archiveCollective`) are not updating the order statuses yet.

Betree (changes requested):

server/models/Collective.js
hdiniz: This should actually have been `virtualCard.delete()`, thanks for pointing this out.
The delete method in the VirtualCard model deletes the card with the provider before deleting from the database

cron/hourly/70-cancel-subscriptions-for-cancelled-orders.ts
```diff
@@ -28,24 +29,38 @@ const getHostFromOrder = async order => {
...
-        required: true,
-        where: { isManagedExternally: true, isActive: true },
+        model: models.Tier,
       },
       {
-        association: 'collective',
+        model: models.Subscription,
         required: true,
-        where: { deactivatedAt: { [Op.not]: null }, isActive: false },
+        where: { isActive: true },
```
Betree, line 61: The following query returns 5 results in prod:

```sql
select * from "Orders" o
INNER JOIN "Subscriptions" s on o."SubscriptionId" = s.id
where o.status = 'CANCELLED'
AND s."isActive"
```

They're old, so I'm assuming it's the leftover of a past bug. I'm noting their IDs just in case and manually marking the subscriptions as inactive:

> 
> 5607
> 525029
> 505772
> 174267
> 181094

server/graphql/v1/mutations/collectives.js
```diff
@@ -514,8 +514,9 @@ export async function archiveCollective(_, args, req) {
   }
 
   // TODO: cascade deactivation to EVENTs and PROJECTs?
+  await collective.changeHost(null);
```
Betree, line 517: `changeHost` is taking care of the recursive with events & projects, which is a really nice addition. An interesting follow-up could be to set `deactivatedAt` on all children and get rid of this TODO.

server/models/Order.js
```diff
@@ -446,6 +447,48 @@ function defineModel() {
...
+          WHERE
+            "Orders".status NOT IN ('PAID', 'CANCELLED', 'REJECTED', 'EXPIRED') AND
+            "Subscriptions"."isManagedExternally" AND
+            "Subscriptions"."isActive" AND
+            "Orders"."CollectiveId" = ?
+        )
+      `,
+      {
+        type: QueryTypes.UPDATE,
+        replacements: [collectiveId],
```
Betree, line 487: :ice_cream: We usually prefer to use named parameters as they're easier to deal with than positional, but no required change here.

Betree (approved):
Looks good to me! Some suggestions & follow-up issues below

cron/hourly/70-cancel-subscriptions-for-cancelled-orders.ts
```diff
@@ -28,24 +29,38 @@ const getHostFromOrder = async order => {
   return models.Collective.findByPk(transactions[0].HostCollectiveId);
 };
 
+const getOrderCancelationReason = (collective, order, orderHost) => {
+  if (order.TierId && !order.Tier) {
+    return ['DELETED_TIER', `Order tier deleted`];
```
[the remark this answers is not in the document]
Betree: Followup issue: https://github.com/opencollective/opencollective/issues/6026

server/graphql/v1/mutations/collectives.js
Betree: Followup in https://github.com/opencollective/opencollective-api/pull/8027

PR #8004, opened 2022-09-28
Add option to stop recurring contributions on tier deletion

hdiniz:
Related https://github.com/opencollective/opencollective/issues/5947
Required https://github.com/opencollective/opencollective-api/pull/8002

### DIFF
--- a/server/graphql/v2/mutation/TierMutations.js
+++ b/server/graphql/v2/mutation/TierMutations.js
@@ -1,4 +1,4 @@
-import { GraphQLNonNull } from 'graphql';
+import { GraphQLBoolean, GraphQLNonNull } from 'graphql';
 
 import models from '../../../models';
 import { checkRemoteUserCanUseAccount } from '../../common/scope-check';
@@ -93,6 +93,10 @@ const tierMutations = {
       tier: {
         type: new GraphQLNonNull(TierReferenceInput),
       },
+      stopRecurringContributions: {
+        type: new GraphQLNonNull(GraphQLBoolean),
+        defaultValue: false,
+      },
     },
     async resolve(_, args, req) {
       checkRemoteUserCanUseAccount(req);
@@ -101,6 +105,10 @@ const tierMutations = {
       const collective = await req.loaders.Collective.byId.load(tier.CollectiveId);
       if (!req.remoteUser.isAdminOfCollective(collective)) {
         throw new Unauthorized();
+      }
+
+      if (args.stopRecurringContributions) {
+        await models.Order.cancelActiveOrdersByTierId(tier.id);
       }
 
       await tier.destroy();
--- a/server/models/Order.js
+++ b/server/models/Order.js
@@ -4,11 +4,12 @@ import { get } from 'lodash';
 import Temporal from 'sequelize-temporal';
 
 import { roles } from '../constants';
+import OrderStatuses from '../constants/order_status';
 import status from '../constants/order_status';
 import TierType from '../constants/tiers';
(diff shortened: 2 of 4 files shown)


server/graphql/v2/mutation/TierMutations.js
```diff
@@ -93,6 +93,9 @@ const tierMutations = {
       tier: {
         type: new GraphQLNonNull(TierReferenceInput),
       },
+      stopRecurringContributions: {
+        type: GraphQLBoolean,
+      },
```
Betree, line 98: :ice_cream: Personal preference, to remove any ambiguity

```suggestion
      stopRecurringContributions: {
        type: new GraphQLNonNull(GraphQLBoolean),
        defaultValue: false,
      },
```

_@hdiniz Note on this feedback:_ Whenever I use :ice_cream: / "Personal preference", you don't have to iterate on the feedback; do so only if you agree and feel like it's a good fit for the PR.

server/models/Order.js
```diff
@@ -446,6 +447,10 @@ function defineModel() {
     });
   };
 
+  Order.cancelActiveOrdersByTierId = function (tierId) {
+    return Order.update({ status: OrderStatuses.CANCELLED }, { where: { TierId: tierId } });
```
Betree, line 451: **[:no_entry_sign: **Blocker**]** We should better filter the orders that we want to update, otherwise we're going to lose a piece of important information for some of them (e.g. "Expired", "Paid", etc).

```suggestion
    return Order.update(
      { status: OrderStatuses.CANCELLED },
      {
        where: {
          TierId: tierId,
          SubscriptionId: { [Op.not]: null },
          status: { [Op.not]: [OrderStatuses.PAID, OrderStatuses.CANCELLED, OrderStatuses.REJECTED, OrderStatuses.EXPIRED] }
        }
    });
```

Betree (changes requested):
The core feature looks good but - unless I missed something - I think we need to update `cron/hourly/70-cancel-archived-accounts-subscriptions.ts` or find an alternative way of doing this ; otherwise we'll end up with zombie subscriptions.

server/models/Order.js
hdiniz: Ops, missed that. I thought about filtering by OrderStatuses.ACTIVE only:

`where: {TierId: tierId, SubscriptionId: { [Op.not]: null }, status: OrderStatuses.ACTIVE }`

wdyt?

Betree: I wondered about this too when I posted my feedback; we do want to cover at least the `NEW` status on top of `ACTIVE`. When PayPal subscriptions orders are created, the flow is basically
1. Create the order with the "New" status
2. Wait for the PayPal webhook "payment captured" event
3. Record transactions & mark the order as active 

This "payment captured" event might take a few minutes to arrive, and though it's an edge case I think it's the right call to include these in the list of orders to cancel because, on PayPal side, they're already active at this stage.

Because of the way PayPal works, `REQUIRE_CLIENT_CONFIRMATION` is not supposed to happen but I think it's good if we have it too: an order requiring a new 3DSecure confirmation is still kind of active.

As for the other statuses:
- `PLEDGED` is not supposed to happen, but I don't think it hurts to have it
- `PENDING` is not supposed to happen either, and https://github.com/opencollective/opencollective/issues/2734 will probably be implemented with Stripe.
- I suppose `ERROR` could happen on existing recurring contributions if the contributor doesn't have enough balance on their PayPal account and then later add a credit card to it

That's why I suggested having an opt-out filter, but feel free to use an opt-in approach if it makes more sense.

Betree (approved):
Looking good!

PR #8006, opened 2022-09-28
CC admins when approving collectives

RichardLitt:
When I am approving or rejecting a collective to join my fiscal host, I would like to get an email in Freshdesk from the message which I send using the platform. I believe this enables that functionality.

I am unsure about the replyTo being the no-reply email address, and whether or not HostCollectiveID is the right email. Ideally, this would be my email, as the host admin making the call. Is that possible?


server/lib/notifications/email.ts
```diff
@@ -413,15 +414,21 @@ export const notifyByEmail = async (activity: Activity) => {
         }
         break;
       }
-      await notify.collective(activity, { collectiveId: activity.data.collective.id });
+      const HostCollectiveId = await collective.getHostCollectiveId();
```
RichardLitt, line 417: Unsure why this is scoped, and where activity.data.collective is validated.

RichardLitt:
Actually, unsure if this goes to Freshdesk at all - this might just go to the email of the Collective Host Admin?

server/lib/notifications/email.ts
```diff
@@ -413,15 +414,21 @@ export const notifyByEmail = async (activity: Activity) => {
...
+      await notify.collective(activity, {
+        collectiveId: activity.data.collective.id,
+        cc: (HostCollectiveId) ? HostCollectiveId : 'no-reply@opencollective.com',
```
Betree, line 420: This part is the core problem and the reason why there's no `cc` currently: we need a way for host admins to define a custom email address for the host account. See https://github.com/opencollective/opencollective/issues/2263

RichardLitt:
Note: Ben is actively working on a better version of this.

RichardLitt:
Thank you, @Betree! 

I really appreciate this merge.

Betree:
@RichardLitt I thought I merged & deployed already, sorry for the delay :grimacing:

RichardLitt:
No worries. :) 

Is there a way for me to test this? I'm not seeing emails in Freshdesk having used the forms.

Betree:
@RichardLitt The form itself is not CCed to the email address in order to prevent flooding support. But the replies will come to the emails we've set (since they're used as the `Reply to` in the emails). They should include the original reply since -by default- replying includes a copy or it. We can iterate if we're not satisfied with the results.

PR #8010, opened 2022-09-29
Host admins to see hosted accounts activity

Betree:
Without this check, the feature does not work when selecting specific hosted collectives from the Host Activity log.


Betree:
Ended up merging a similar fix in https://github.com/opencollective/opencollective-api/pull/8009

PR #8011, opened 2022-09-29
fix: virtual card refunds don't have an ExpenseId

znarf:


Betree:
Resolves https://sentry.io/organizations/open-collective/issues/3631170111/?project=5199682&referrer=slack I guess

PR #8016, opened 2022-09-29
Add private instructions on event ICS file

Betree:
Example file: https://drive.google.com/file/d/1EkEYP9As8leWa-o9RJR6yD7n6_YsqD1j/view?usp=sharing

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -894,6 +894,9 @@ function defineModel() {
         if (this.longDescription) {
           description += `\n\n${this.longDescription}`;
         }
+        if (this.data?.privateInstructions) {
+          description += `\n\n${this.data.privateInstructions}`;
+        }
         let location = this.location.name || '';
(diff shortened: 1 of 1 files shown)


RichardLitt:
This looks good to me from just looking at the code.

RichardLitt (commented):
Perhaps there should also be a note saying that "this is private instructions". Something like: 

```
          description += `\n\n<b>Private instructions:</b>\n\n${this.data.privateInstructions}`;
```

Betree:
> \n\n<b>Private instructions:</b>\n\n$

Done, though I removed the HTML part as I was getting inconsistent results between Calendar clients (even when using the `htmlContent` prop from https://www.npmjs.com/package/ics) that I was not ready to investigate deeper.

RichardLitt (approved):

RichardLitt:
Thank you, @Betree!

PR #8027, opened 2022-10-05
Completely archive children when archiving parent

Betree:
Followup on https://github.com/opencollective/opencollective-api/pull/8002#discussion_r985920152
Cherry-pick the tests from, and closes https://github.com/opencollective/opencollective-api/pull/7693

@hdiniz implemented a new solution for archiving in https://github.com/opencollective/opencollective-api/pull/8002/files#diff-f35d6394d656027361220f1c2baea5ee1fa2ff0e027ac8e6d9b3fb6d2541b86dR516-R517: by re-using `changeHost`, the recursive of marking all children as inactive is handled automatically. All we have to do it to update the `deactivatedAt` flag.

I've also set a special `children.data.archivedFromParent` flag in case we want to use it for un-archiving in the future.

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -14,7 +14,7 @@ import * as github from '../../../lib/github';
 import RateLimit, { ONE_HOUR_IN_SECONDS } from '../../../lib/rate-limit';
 import { canUseFeature } from '../../../lib/user-permissions';
 import { defaultHostCollective } from '../../../lib/utils';
-import models from '../../../models';
+import models, { Op, sequelize } from '../../../models';
 import { FeatureNotAllowedForUser, NotFound, RateLimitExceeded, Unauthorized, ValidationFailed } from '../../errors';
 import { CollectiveInputType } from '../inputTypes';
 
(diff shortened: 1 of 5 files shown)


server/graphql/v1/mutations/collectives.js
```diff
@@ -513,10 +514,26 @@ export async function archiveCollective(_, args, req) {
     return updatedCollective;
   }
 
-  // TODO: cascade deactivation to EVENTs and PROJECTs?
+  // `changeHost` will recursively check children and unhost them
```
znarf, line 517: So this is gonna fail if a children has a balance. Today the message is not clear and are confused why they still have a balance. https://opencollective.slack.com/archives/CJ2ERRZQ8/p1664980298922799

znarf (approved):
Looks good to me.

server/graphql/v1/mutations/collectives.js
Betree: @znarg I introduced a commit to address that: https://github.com/opencollective/opencollective-api/pull/8027/commits/48661938924261e6420fc73d4109140a1f980040

server/lib/queries.js
```diff
@@ -220,6 +220,7 @@ const getTopBackers = (since, until, tags, limit) => {
     LEFT JOIN "Collectives" collective ON collective.id = t."CollectiveId"
     WHERE
       t.type='CREDIT'
+      AND t."deletedAt" IS NULL
```
Betree, line 223: Had to do this small change to make the tests less sensitive in https://github.com/opencollective/opencollective-api/pull/8027/files#diff-ed0b486411042fde03fcb5a88b2fe2159ae7cf484b0d755f4a88c0a666cafaa2. It seemed desirable to ignore deleted transactions anyway.

PR #8029, opened 2022-10-05
feat(stripe-dispute-orders): Create webhooks for new and closed disputes.

vaughanj10:
Related Issue: https://github.com/opencollective/opencollective/issues/5983


config/test.json
```diff
@@ -1,4 +1,5 @@
 {
+  "port": "3061",
```
vaughanj10, line 2: Modified this so that I could run my dev server AND run the test suite at the same time, otherwise there is a port collision.

server/paymentProviders/stripe/creditcard.js
```diff
@@ -308,11 +313,124 @@ export const setupCreditCard = async (paymentMethod, { user, collective } = {})
...
+        type: 'DEBIT',
+        HostCollectiveId: fiscalHost.id,
+        CollectiveId: fiscalHost.id,
+        FromCollectiveId: fiscalHost.id,
+        OrderId: creditTransaction.OrderId,
+        amount: toNegative(hostCurrencyAmount),
+        netAmountInCollectiveCurrency: toNegative(hostCurrencyAmount),
+        amountInHostCurrency: Math.round(toNegative(amount) * hostCurrencyFxRate),
+        currency: currency,
+        hostCurrency: currency,
```
kewitz, line 398:
```suggestion
        hostCurrency: fiscalHost.currency,
```

kewitz (commented):
Looking good. I just caught some misconceptions when creating the Dispute Fee transaction.

server/graphql/v2/object/Host.js
```diff
@@ -727,6 +728,28 @@ export const Host = new GraphQLObjectType({
...
+            WHERE t."HostCollectiveId" = :hostCollectiveId AND o.status = 'DISPUTED'
+            GROUP BY o.id;
+            `,
+            {
+              type: QueryTypes.SELECT,
+              replacements: {
+                hostCollectiveId: collective.id,
+              },
+            },
+          );
```
kewitz, line 748: Not mandatory, but for simple queries, I think a sequelize query would be more consistent here.

server/graphql/v2/object/Host.js
```diff
@@ -727,6 +728,28 @@ export const Host = new GraphQLObjectType({
...
+
+          return !isEmpty(results);
```
kewitz, line 750: If you're checking for emptiness, I suggest using a count query.

kewitz (commented):
Looking good. Just need to review failing tests.

config/test.json
kewitz, line 2: @znarf what do you think about an inconsistent API port for the sake of being able to run tests while also running a dev server?

znarf: Sounds good I think? I have seen similar attempts at this in other places. Actually for cron jobs search for ```process.env.PORT = 3066;```

server/paymentProviders/stripe/creditcard.js
```diff
@@ -308,11 +313,124 @@ export const setupCreditCard = async (paymentMethod, { user, collective } = {})
...
+      const amount = feeDetails.amount;
+      const fiscalHost = await models.Collective.findByPk(creditTransaction.HostCollectiveId);
+      const hostCurrencyFxRate = await getFxRate(currency, fiscalHost.currency);
+      const hostCurrencyAmount = Math.round(toNegative(amount) * hostCurrencyFxRate);
+
+      await models.Transaction.create({
+        type: 'DEBIT',
+        HostCollectiveId: fiscalHost.id,
+        CollectiveId: fiscalHost.id,
+        FromCollectiveId: fiscalHost.id,
```
znarf, line 392: @kewitz are you good with this? it's ok with our current semantics, but having `CollectiveId = FromCollectiveId` is not something we think is the best for the future.

server/paymentProviders/stripe/creditcard.js
```diff
@@ -308,11 +313,124 @@ export const setupCreditCard = async (paymentMethod, { user, collective } = {})
...
+        amountInHostCurrency: hostCurrencyAmount,
+        currency: currency,
+        hostCurrency: fiscalHost.currency,
+        description: 'Dispute Fee paid to Stripe',
+        paymentProcessorFeeInHostCurrency: 0,
+        hostFeeInHostCurrency: 0,
+        platformFeeInHostCurrency: 0,
+        hostCurrencyFxRate,
+        kind: TransactionKind.PAYMENT_PROCESSOR_DISPUTE_FEE,
+        data: event.data,
```
znarf, line 405: What about `TransactionGroup`, should be the same than refund?

server/paymentProviders/stripe/creditcard.js
```diff
@@ -308,11 +313,124 @@ export const setupCreditCard = async (paymentMethod, { user, collective } = {})
...
+
+      // Create transaction for dispute fee debiting the fiscal host
+      const feeDetails = disputeTransaction.fee_details.find(feeDetails => feeDetails.description === 'Dispute fee');
+      const currency = feeDetails.currency.toUpperCase();
+      const amount = feeDetails.amount;
+      const fiscalHost = await models.Collective.findByPk(creditTransaction.HostCollectiveId);
+      const hostCurrencyFxRate = await getFxRate(currency, fiscalHost.currency);
+      const hostCurrencyAmount = Math.round(toNegative(amount) * hostCurrencyFxRate);
+
+      await models.Transaction.create({
```
kewitz, line 388: @vaughanj10, lets double check if OCF and OSC want to disclose this type of transaction in their ledger.

server/graphql/v2/object/Host.js
hdiniz: You can also limit the return to 1 record.

`... LIMIT 1`

server/paymentProviders/stripe/creditcard.js
```diff
@@ -308,11 +313,124 @@ export const setupCreditCard = async (paymentMethod, { user, collective } = {})
   return paymentMethod;
 };
 
+// Charge has a new Stripe dispute
+const createDispute = async event => {
+  const stripeChargeId = event.data.object.charge;
+  const transactions = await models.Transaction.findAll({
+    where: { data: { charge: { id: stripeChargeId } } },
```
hdiniz, line 320: I think we should create an index either for the json column `data` or specifically for this `charge.id` field,
the transaction is our biggest table and I think this select will cause a slow table scan.

```
select id from "Transactions" t where t.data->'charge'->>'id' = 'ch_3L4DagBYycQg1OMf1KxilamS'

Gather  (cost=1000.00..181144.54 rows=16100 width=395)                                           
  Workers Planned: 1                                                                             
  ->  Parallel Seq Scan on "Transactions" t  (cost=0.00..178534.54 rows=9471 width=395)          
        Filter: (((data -> 'charge'::text) ->> 'id'::text) = 'ch_3L4DagBYycQg1OMf1KxilamS'::text)

2 row(s) fetched - 46.702s
```

server/paymentProviders/stripe/creditcard.js
vaughanj10: Checked in with folks on Slack. Going to shift the language on this to something like "Stripe Review Fee"

config/test.json
vaughanj10: I'm not really sure what having a consistent API port for this gives us. Seems like running tests + running the dev server concurrently should be a priority for a productive dev setup.

server/graphql/v2/object/Host.js
vaughanj10: Changed to `LIMIT 1`, good call!

server/graphql/v2/object/Host.js
vaughanj10: Agreed, I couldn't quite figure out how to make this call using a sequelize query so I just wrote it out in SQL.

server/paymentProviders/stripe/creditcard.js
kewitz: This is consistent behavior with what we do on added funds, so I'm fine with that.

server/paymentProviders/stripe/creditcard.js
kewitz: Agreed. We should get the TransactionGroup from the transaction created on line 374.

server/paymentProviders/stripe/index.js
```diff
@@ -255,19 +255,20 @@ export default {
...
-        /* This will cause stripe to send us email alerts, saying
-         * that our stuff is broken. But that should never happen
-         * since they discontinued the support. */
-        throw new errors.BadRequest('Stripe-Bitcoin not supported anymore :(');
+        // Charge has a Stripe dispute
+      } else if (event.type === 'charge.dispute.created') {
+        return creditcard.createDispute(event);
+        // Charge dispute has been closed on Stripe (with status of: won/lost/closed)
+      } else if (event.type === 'charge.dispute.closed') {
+        return creditcard.closeDispute(event);
```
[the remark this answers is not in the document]
vaughanj10: Thank you!

migrations/20221012165909-add-stripe-charge-id-index.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY stripe_charge_id 
+      ON "Transactions" (((data -> 'charge' ->> 'id')::text)) 
+      WHERE (data -> 'charge' ->> 'id') IS NOT NULL`);
```
kewitz, line 8: Have you tested this index? I think I had some trouble in the past related to the JSONB selection operator needing to match the operator used in the query.
I'm not 100% confident here but I implemented something similar this week that I can ensure would work:
```suggestion
      CREATE INDEX CONCURRENTLY IF NOT EXISTS "transactions__stripe_charge_id" ON "Transactions" USING BTREE ((data#>>'{charge,id}') DESC) WHERE data#>>'{charge,id}' IS NOT NULL;
```

vaughanj10: Here is the same query after applying the index:
```
opencollective_prod_snapshot> explain analyze select id from "Transactions" t where t.data->'charge'->>'id' = 'ch_3L4DagBYycQg1OMf1KxilamS';
+-------------------------------------------------------------------------------------------------------------------------------+
| QUERY PLAN                                                                                                                    |
|-------------------------------------------------------------------------------------------------------------------------------|
| Bitmap Heap Scan on "Transactions" t  (cost=387.08..48602.90 rows=16342 width=4) (actual time=0.069..0.074 rows=2 loops=1)    |
|   Recheck Cond: (((data -> 'charge'::text) ->> 'id'::text) = 'ch_3L4DagBYycQg1OMf1KxilamS'::text)                             |
|   Heap Blocks: exact=2                                                                                                        |
|   ->  Bitmap Index Scan on stripe_charge_id  (cost=0.00..382.99 rows=16342 width=0) (actual time=0.055..0.056 rows=2 loops=1) |
|         Index Cond: (((data -> 'charge'::text) ->> 'id'::text) = 'ch_3L4DagBYycQg1OMf1KxilamS'::text)                         |
| Planning Time: 0.230 ms                                                                                                       |
| Execution Time: 0.139 ms                                                                                                      |
+-------------------------------------------------------------------------------------------------------------------------------+
EXPLAIN
Time: 0.018s
```

kewitz: You'll need to analyze the query that is generated by:
```js
const transactions = await models.Transaction.findAll({
    where: { data: { charge: { id: stripeChargeId } } }
});
```

vaughanj10: Ah ok, I can see that the query generated from the `findAll` is using the other query operator `"Transaction"."data"#>>'{charge,id}'` and when I ran that query against the prod dump it was about 10 seconds. I applied the index you posted above and it improved dramatically again. 

```
opencollective_prod_snapshot> SELECT id  FROM "Transactions" WHERE "data"#>>'{charge,id}' = 'ch_3L4DagBYycQg1OMf1KxilamS';
+---------+
| id      |
|---------|
| 2495478 |
| 2495479 |
+---------+
SELECT 2
Time: 9.967s (9 seconds), executed in: 9.959s (9 seconds)
opencollective_prod_snapshot> CREATE INDEX CONCURRENTLY IF NOT EXISTS "transactions__stripe_charge_id" ON "Transactions" USING BTREE ((data#>>'{cha
 rge,id}') DESC) WHERE data#>>'{charge,id}' IS NOT NULL;
CREATE INDEX
Time: 31.947s (31 seconds), executed in: 31.947s (31 seconds)
opencollective_prod_snapshot> SELECT id  FROM "Transactions" WHERE "data"#>>'{charge,id}' = 'ch_3L4DagBYycQg1OMf1KxilamS';
+---------+
| id      |
|---------|
| 2495478 |
| 2495479 |
+---------+
SELECT 2
Time: 0.021s
```

vaughanj10: Would it be worth it to create both indices? Just in case some tries to use this query operator `(data -> 'charge'::text) ->> 'id'::text`?

vaughanj10:
```
opencollective_prod_snapshot> explain analyze SELECT id  FROM "Transactions" WHERE "data"#>>'{charge,id}' = 'ch_3L4DagBYycQg1OMf1KxilamS';
+---------------------------------------------------------------------------------------------------------------------------------------------+
| QUERY PLAN                                                                                                                                  |
|---------------------------------------------------------------------------------------------------------------------------------------------|
| Bitmap Heap Scan on "Transactions"  (cost=387.08..48562.05 rows=16342 width=4) (actual time=0.060..0.066 rows=2 loops=1)                    |
|   Recheck Cond: ((data #>> '{charge,id}'::text[]) = 'ch_3L4DagBYycQg1OMf1KxilamS'::text)                                                    |
|   Heap Blocks: exact=2                                                                                                                      |
|   ->  Bitmap Index Scan on transactions__stripe_charge_id  (cost=0.00..382.99 rows=16342 width=0) (actual time=0.048..0.048 rows=2 loops=1) |
|         Index Cond: ((data #>> '{charge,id}'::text[]) = 'ch_3L4DagBYycQg1OMf1KxilamS'::text)                                                |
| Planning Time: 0.218 ms                                                                                                                     |
| Execution Time: 0.144 ms                                                                                                                    |
+---------------------------------------------------------------------------------------------------------------------------------------------+
EXPLAIN
Time: 0.016s
```

vaughanj10: Resolved the final question off-thread: just going to create an index that matches the operator style used by Sequelize querying.

kewitz (approved):

server/paymentProviders/stripe/creditcard.js
```diff
@@ -308,11 +313,124 @@ export const setupCreditCard = async (paymentMethod, { user, collective } = {})
...
+    );
+
+    // A lost dispute means it was decided as fraudulent
+    if (disputeStatus === 'lost') {
+      // Create refund transaction for the fraudulent charge
+      await createRefundTransaction(creditTransaction, 0, {
+        ...creditTransaction.data,
+        dispute: event,
+        refundTransactionId: creditTransaction.id,
+      });
```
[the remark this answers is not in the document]
vaughanj10: Added a ledger snapshot for disputes!

PR #8036, opened 2022-10-07
Policy to enforce 2FA on all admins 

Betree:
Part of https://github.com/opencollective/opencollective/issues/5289
For https://github.com/opencollective/opencollective-frontend/pull/8264

# What's implemented

I've added the check on multiple mutations but did not want to do something too disruptive. That's why I've set `neverAskForToken` to true for most actions, which will only ensure peeps have 2FA enabled if their orgs require it. We should followup on a better API in https://github.com/opencollective/opencollective/issues/6085.

For now, 2FA will be required with sessions for:
- Create connected account
- Invite member
- Edit member invitation
- Remove member
- Edit member
- Delete tier
- Request virtual card
- Create/Update webhook
- Apply to host
- Create application
- Create payout method
- All root mutations
- Create/Assigning virtual cards

2FA input will be required each time for:
- Delete connected account
- Update email
- Editing account 2FA policy

# To do for this PR
- [x] Introduce the policy
- [x] Protect 2FA policies / settings with 2FA
- [x] Protect all important host queries & mutations
- [x] Make sure 2FA settings cannot be edited from another (unprotected) place (i.e. GQLV1)
  - => The policy itself cannot be edited as it lives inside of `data`. The rolling limit can be, we'll followup in https://github.com/opencollective/opencollective/issues/6044
- [x] Make sure we don't force you to input 2FA code for unimportant stuff (already asked at the logging stage)
- [ ] Implement missing 2FA (see `// check 2fa` comments)
- [x] Make sure we always check policies properly when dealing with accounts with parents
- [x] What if trying to disable the 2FA policy on children accounts?


server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -77,6 +78,11 @@ const accountMutations = {
           throwIfMissing: true,
         });
 
+        // Enforce 2FA if trying to change 2FA rolling limit settings while it's already enabled
+        if (args.key.split('.')[0] === 'payoutsTwoFactorAuth' && account.settings?.payoutsTwoFactorAuth?.enabled) {
+          await TwoFactorAuthLib.validateRequest(req, { alwaysAskForToken: true });
```
hdiniz, line 83: @Betree 

```suggestion
          await TwoFactorAuthLib.validateRequest(req, { alwaysAskForToken: true, requireTwoFactorAuthEnabled: true });
```

`alwaysAskForToken` will require a token if the user has 2FA enabled, otherwise the check will pass
`requireTwoFactorAuthEnabled` will make sure that the user also has 2FA enabled.

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -434,8 +440,15 @@ const accountMutations = {
         throw new Unauthorized();
       }
 
+      // Merge submitted policies with existing ones
       const previousPolicies = account.data?.policies;
       const newPolicies = omitBy({ ...previousPolicies, ...args.policies }, isNull);
+
+      // Enforce 2FA when trying to disable `REQUIRE_2FA_FOR_ADMINS`
+      if (previousPolicies?.REQUIRE_2FA_FOR_ADMINS && !newPolicies.REQUIRE_2FA_FOR_ADMINS) {
+        await TwoFactorAuthLib.validateRequest(req, { alwaysAskForToken: true });
```
hdiniz, line 449: same here with requireTwoFactorAuthEnabled

server/graphql/common/update.ts
```diff
@@ -22,6 +23,8 @@ export async function createUpdate(_, args, req) {
     throw new Forbidden('Only root users can create changelog updates.');
   }
 
+  await twoFactorAuthLib.enforceForAccountAdmins(req, collective, { neverAskForToken: true });
```
hdiniz, line 26: The use of `neverAskForToken` is not clear imo. The flag is used because we only want to check if the user has used 2fa to login if the account requires it, I suggest we do this differently.

We could use the twoFactorAuthLib on the login with 2fa flow and there set a session duration that equals the JWT duration. By default the 2fa enforcing will be using the login session. If we need a different session time or rule we can then explicitly set it in the appropriate mutations.

Betree: I agree that it's not ideal, but I think the solution you proposed would be better if coupled with a concept of "protection level" (soft, hard, etc) as proposed in https://github.com/opencollective/opencollective/issues/6085; which will further complexify this PR that is already quite dense.

As a clarification, I think we can rename `neverAskForToken` to `onlyAskOnLogin`; it might not be the best API but it will make it clearer for other devs reading the code while keeping the door open for further refactoring & clarification when we'll have better specs.

hdiniz (approved):

PR #8038, opened 2022-10-07
Add base for two factor authentication

hdiniz:
This is a proposal for adding a two factor authentication lib to plug into our graphql mutations.

It works as follows:

1. When a request requires 2FA, a GraphQL error with extension code `2FA_REQUIRED` will be returned.
2. The frontend will treat that error and prompt the user for a 2FA code.
3. The frontend will submit the same GraphQL request again unchanged, but adding the header `X-Two-Factor-Authentication` with format:
   
   `X-Two-Factor-Authentication: <two-factor-type> <two-factor-code>`
   e.g. in case of TOTP (Timed one time password):
   `X-Two-Factor-Authentication: totp 123456`

   Passing this value as a header instead of part of the graphql schema more closely follows its purpose of authentication and does not require us to change the schema of every mutation we want to protect.

4. Based on the `two-factor-type`, different providers can be used to validate the code.
5. The lib will store a session by default where the 2FA code will not be required again for some time, there are options to always require it and to modify the session duration and session id so that different operations can have different policies.

This PR refactors the existing 2FA validation on expenses payout to test the library and adds to the updateEmail and createApplication mutation as examples.


server/lib/two-factor-authentication/lib.ts
```diff
@@ -0,0 +1,168 @@
...
+
+function getTwoFactorAuthTokenFromRequest(req: Request): Token {
+  const header = req.get(TwoFactorAuthenticationHeader);
+  if (!header) {
+    return null;
+  }
+
+  const parts = header.split(' ');
+  const type = parts[0] as TwoFactorMethod;
+  const code = parts[1];
```
Betree, line 52: I haven't actually tested (would be great to have some unit tests to cover the actual errors for this one) but I think we can be a bit more strict and do some pre-validation here in case people try to fool with the `x-two-factor-authentication` header. Some examples:

> { x-two-factor-authentication: ' ' } => { type: '', code: '' }
> { x-two-factor-authentication: 'totp' } => { type: 'totp', code: undefined }

```suggestion
  const parts = header.split(' ');
  const type = parts[0] as TwoFactorMethod;
  const code = parts[1];
  if (!type || !code) {
    throw new Error(`Malformed 2FA token header`);
  }
```

server/lib/two-factor-authentication/lib.ts
```diff
@@ -0,0 +1,168 @@
...
+function getSessionKey(req: Request, options: ValidateRequestOptions) {
+  if (typeof options.sessionKey === 'function') {
+    return options.sessionKey();
+  } else if (typeof options.sessionKey === 'string') {
+    return options.sessionKey;
+  } else {
+    const userId = req?.remoteUser?.id;
+    const sessionId = req.jwtPayload?.sessionId;
+    return `${userId}_2fa_${sessionId}`;
+  }
```
Betree, line 83: :ice_cream: I suggest adding a prefix to all cache keys, mostly to facilitate the exploration in Redis GUI tools but that could also be used in the future to easily invalidate all 2FA for a given user.

```suggestion
  const userId = req.remoteUser.id;
  if (typeof options.sessionKey === 'function') {
    return `2fa:${userId}:${options.sessionKey()}`;
  } else if (typeof options.sessionKey === 'string') {
    return `2fa:${userId}:${options.sessionKey}`;
  } else {
    const sessionId = req.jwtPayload.sessionId;
    return `2fa:${userId}:default_${sessionId}`;
  }
```

server/lib/two-factor-authentication/lib.ts
```diff
@@ -0,0 +1,168 @@
...
+  return cache.set(sessionKey, {}, options.sessionDuration);
+}
+
+async function validateRequest(
+  req: Request,
+  options: ValidateRequestOptions = DefaultValidateRequestOptions,
+): Promise<void> {
+  options = { ...DefaultValidateRequestOptions, ...options };
+
+  const remoteUser = req?.remoteUser;
```
Betree, line 115: Unless I'm missing something, I don't think we want to allow a null request here as it would crash `userHasTwoFactorAuthEnabled` anyway

```suggestion
  if (!req.remoteUser) {
    throw new Unauthorized();
  }
  
  const remoteUser = req.remoteUser;
```

server/graphql/v2/mutation/ExpenseMutations.ts
```diff
@@ -291,6 +292,10 @@ const expenseMutations = {
     async resolve(_: void, args, req: express.Request): Promise<typeof Expense> {
       checkRemoteUserCanUseExpenses(req);
 
+      if (args?.paymentParams?.twoFactorAuthenticatorCode) {
+        req.headers[TwoFactorAuthenticationHeader] = `totp ${args.paymentParams.twoFactorAuthenticatorCode}`;
+      }
+
```
Betree, line 298: That's a good hack for now but I think we should open a followup issue about deprecating this `args?.paymentParams?.twoFactorAuthenticatorCode` in favor of the new header you've implemented.

server/graphql/v2/mutation/ApplicationMutations.js
```diff
@@ -20,6 +21,7 @@ const createApplication = {
   },
   async resolve(_, args, req) {
     checkRemoteUserCanUseApplications(req);
+    await twoFactorAuthLib.validateRequest(req);
```
Betree, line 24: I love the simplicity of this :heart_eyes:

server/graphql/v1/mutations.js
```diff
@@ -193,8 +194,9 @@ const mutations = {
         description: 'The new email address for user',
       },
     },
-    resolve: (_, { email }, { remoteUser }) => {
-      return updateUserEmail(remoteUser, email);
+    resolve: async (_, { email }, req) => {
+      await twoFactorAuthLib.validateRequest(req);
```
Betree, line 198: The check for `req.remoteUser` in `validateRequest` I suggested above seems important to have a clean error message for this one

server/graphql/common/expenses.ts
```diff
@@ -1781,13 +1815,12 @@ export async function payExpense(req: express.Request, args: Record<string, unkn
     const useTwoFactorAuthentication =
       isTwoFactorAuthenticationRequiredForPayoutMethod && !forceManual && hostHasPayoutTwoFactorAuthenticationEnabled;
 
+    const expensePaidAmountKey = `${req.remoteUser.id}_2fa_payment_limit`;
+    let currentPaidExpenseAmount;
```
Betree, line 1819: :ice_cream: I find these variable names a bit misleading, I initially thought it was about the amount of the current expense. Suggestions:
- `alreadyPaidExpensesAmountKey`, `totalPaidExpensesAmountKey`
- `alreadyPaidExpensesAmount`, `totalPaidExpensesAmount`

server/graphql/common/expenses.ts
```diff
@@ -611,6 +609,46 @@ export const markExpenseAsSpam = async (
   return updatedExpense;
 };
 
+const ROLLING_LIMIT_CACHE_VALIDITY = 3600; // 1h in secs for cache to expire
+
+async function validateExpensePayout2FALimit(req, expense, expensePaidAmountKey) {
+  const host = await expense.collective.getHostCollective();
```
Betree, line 615: We can take the host as a parameter and save a SQL query, since all places calling this function already loaded it to check `const hostHasPayoutTwoFactorAuthenticationEnabled = get(host, 'settings.payoutsTwoFactorAuth.enabled', false);`

server/lib/two-factor-authentication/lib.ts
```diff
@@ -0,0 +1,168 @@
...
+      supportedMethods: await twoFactorMethodsSupportedByUser(req.remoteUser),
+    });
+  }
+
+  await validateToken(remoteUser, token);
+
+  return storeTwoFactorSession(req, options);
+}
+
+async function twoFactorMethodsSupportedByUser(remoteUser: typeof User) {
```
Betree, line 144: Sounds like this one doesn't need to be `async`

```suggestion
function twoFactorMethodsSupportedByUser(remoteUser: typeof User): TwoFactorMethod[] {
```

server/lib/two-factor-authentication/lib.ts
```diff
@@ -0,0 +1,168 @@
...
+  if (!options.alwaysAskForToken) {
+    if (await hasValidTwoFactorSession(req, options)) {
+      return;
+    }
+  }
+
+  const token = getTwoFactorAuthTokenFromRequest(req);
+  if (!token) {
+    throw new ApolloError('Two-factor authentication required', '2FA_REQUIRED', {
+      supportedMethods: await twoFactorMethodsSupportedByUser(req.remoteUser),
```
Betree, line 135:
```suggestion
      supportedMethods: twoFactorMethodsSupportedByUser(req.remoteUser),
```

server/graphql/v2/mutation/ExpenseMutations.ts
hdiniz: I will open an issue to deprecate and remove this from the frontend as well.

server/lib/two-factor-authentication/lib.ts
hdiniz: Thanks for catching this, enforcing the userID in the key prevents someone passing a key that is valid to all users.

Betree (approved):

PR #8059, opened 2022-10-13
Improvements on `isCollectiveDeletable`

Betree:
The field was pretty expensive to compute, almost 2s in some cases:
image

Since it was fetched on the `LoggedInUser` query (which is going to change with https://github.com/opencollective/opencollective-frontend/pull/8287) the field was intensively used.

I approached this change in two steps:
1. Started by implementing unit tests to make sure this PR wouldn't introduce regressions
2. Optimized the code that was marked as expensive in Sentry performance (see details below)

### DIFF
--- a/server/lib/collectivelib.ts
+++ b/server/lib/collectivelib.ts
@@ -8,13 +8,13 @@ import activities from '../constants/activities';
 import { types as CollectiveTypes } from '../constants/collectives';
 import { MODERATION_CATEGORIES } from '../constants/moderation-categories';
 import { VAT_OPTIONS } from '../constants/vat';
-import models, { Op } from '../models';
+import models, { Op, sequelize } from '../models';
 
 import logger from './logger';
 import { stripHTML } from './sanitize-html';
 import { md5 } from './utils';
 
-const { EVENT, PROJECT, USER } = CollectiveTypes;
+const { USER } = CollectiveTypes;
 
 type AvatarUrlOpts = {
   height?: boolean;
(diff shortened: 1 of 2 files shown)


server/lib/collectivelib.ts
```diff
@@ -290,63 +290,76 @@ export async function isCollectiveDeletable(collective) {
...
+          ON m."CollectiveId" = other_members."CollectiveId"
+          AND other_members.role = 'ADMIN'
+          AND other_members."MemberCollectiveId" != m."MemberCollectiveId"
+          AND other_members."deletedAt" IS NULL
+        WHERE m.role = 'ADMIN'
+        AND m."MemberCollectiveId" = :CollectiveId
+        AND m."deletedAt" IS NULL
+        AND c."deletedAt" IS NULL
+        AND other_members.id IS NULL
+      ) AS "isLastAdminOfAnyCollective"
```
Betree, line 310: The previous look had a n+1 issue. For an account with 10 memberships, it was making 1 (fetch memberships ) + 10 (fetch admins of membership collective) = 11 queries. This turns it into only one, taking advantage of the `EXISTS` keyword optimization.

Explain analyze:

```
Result  (cost=402.84..402.84 rows=1 width=1) (actual time=0.215..0.217 rows=1 loops=1)
  InitPlan 1 (returns $3)
    ->  Nested Loop  (cost=0.25..402.84 rows=1 width=0) (actual time=0.214..0.215 rows=1 loops=1)
          ->  Nested Loop Left Join  (cost=0.17..398.75 rows=1 width=4) (actual time=0.204..0.205 rows=1 loops=1)
                Filter: (other_members.id IS NULL)
                Rows Removed by Filter: 32
"                ->  Index Scan using ""MemberCollectiveId-CollectiveId-role"" on ""Members"" m  (cost=0.08..49.23 rows=27 width=8) (actual time=0.018..0.052 rows=5 loops=1)"
"                      Index Cond: ((""MemberCollectiveId"" = 4469) AND ((role)::text = 'ADMIN'::text))"
"                      Filter: (""deletedAt"" IS NULL)"
                      Rows Removed by Filter: 14
"                ->  Index Scan using ""CollectiveId-role"" on ""Members"" other_members  (cost=0.08..12.93 rows=5 width=12) (actual time=0.016..0.029 rows=6 loops=5)"
"                      Index Cond: ((""CollectiveId"" = m.""CollectiveId"") AND ((role)::text = 'ADMIN'::text))"
"                      Filter: ((""deletedAt"" IS NULL) AND (""MemberCollectiveId"" <> m.""MemberCollectiveId""))"
                      Rows Removed by Filter: 6
"          ->  Index Scan using ""Groups_pkey"" on ""Collectives"" c  (cost=0.08..4.09 rows=1 width=4) (actual time=0.008..0.008 rows=1 loops=1)"
"                Index Cond: (id = m.""CollectiveId"")"
"                Filter: (""deletedAt"" IS NULL)"
Planning Time: 0.995 ms
Execution Time: 0.264 ms
```

server/lib/collectivelib.ts
```diff
@@ -290,63 +290,76 @@ export async function isCollectiveDeletable(collective) {
...
+    ) AS "hasUndeletableData"
+  `,
+    {
+      plain: true,
+      replacements: { CollectiveId: collective.id, UserId: user?.id },
+    },
+  );
 
-  return true;
+  return !hasUndeletableData;
```
Betree, line 362: This query brings a few optimizations compared to the previous:
- We were fetching all the data, regardless of the previous results. i.e. we were counting all expenses even if we detected transactions before. By relying on SQL's ORs, we're making sure we stop querying data as soon as we find something that can prevent the deletion.
- The query order matters: `Transactions` are heavier to fetch because it's a bigger table. By fetching them at the end, we're making sure that we first look at lighter tables.

Explain analyze with OSC (look at the "never executed"):
```
Result  (cost=24.18..24.18 rows=1 width=1) (actual time=0.015..0.016 rows=1 loops=1)
  InitPlan 1 (returns $0)
"    ->  Index Scan using ""ParentCollectiveId"" on ""Collectives""  (cost=0.08..11.28 rows=5 width=0) (actual time=0.013..0.013 rows=1 loops=1)"
"          Index Cond: (""ParentCollectiveId"" = 11004)"
"          Filter: (""deletedAt"" IS NULL)"
  InitPlan 2 (returns $1)
"    ->  Bitmap Heap Scan on ""Expenses""  (cost=7.09..1000.20 rows=609 width=0) (never executed)"
"          Recheck Cond: (((""CollectiveId"" = 11004) AND (""deletedAt"" IS NULL)) OR ((""FromCollectiveId"" = 11004) AND (""deletedAt"" IS NULL)))"
          ->  BitmapOr  (cost=7.09..7.09 rows=609 width=0) (never executed)
                ->  Bitmap Index Scan on expenses_collective_id  (cost=0.00..4.96 rows=602 width=0) (never executed)
"                      Index Cond: (""CollectiveId"" = 11004)"
                ->  Bitmap Index Scan on expenses_from_collective_id  (cost=0.00..2.07 rows=7 width=0) (never executed)
"                      Index Cond: (""FromCollectiveId"" = 11004)"
  InitPlan 3 (returns $2)
"    ->  Bitmap Heap Scan on ""Orders""  (cost=6.81..812.36 rows=132 width=0) (never executed)"
"          Recheck Cond: (((""CollectiveId"" = 11004) AND (""deletedAt"" IS NULL)) OR ((""FromCollectiveId"" = 11004) AND (""deletedAt"" IS NULL)))"
          Filter: ((status)::text = ANY ('{PAID,ACTIVE,CANCELLED}'::text[]))
          ->  BitmapOr  (cost=6.81..6.81 rows=416 width=0) (never executed)
                ->  Bitmap Index Scan on orders_collective_id  (cost=0.00..4.70 rows=412 width=0) (never executed)
"                      Index Cond: (""CollectiveId"" = 11004)"
                ->  Bitmap Index Scan on orders_from_collective_id  (cost=0.00..2.09 rows=4 width=0) (never executed)
"                      Index Cond: (""FromCollectiveId"" = 11004)"
  InitPlan 4 (returns $3)
"    ->  Seq Scan on ""Transactions""  (cost=0.00..183390.16 rows=830198 width=0) (never executed)"
"          Filter: ((""deletedAt"" IS NULL) AND ((""CollectiveId"" = 11004) OR (""FromCollectiveId"" = 11004)))"
Planning Time: 0.440 ms
Execution Time: 0.060 ms
```

hdiniz (approved):
lgtm

PR #8060, opened 2022-10-13
Make sure we check for `ExpenseItems` for attachments as well

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5951


znarf (approved):
Good for me, let's try this approach.

PR #8066, opened 2022-10-14
test(graphql): add tests for sendMessage mutation 

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/6051

- [x]  it('sends the message by email') => {}
- [x]   it('cannot inject code in the email (XSS)') => {}
- [x]   it('returns an error if not authenticated') => {}
- [x]   it('returns an error if collective cannot be contacted') => {}
- [x]   it('returns an error if the feature is blocked for user') => {}
- [x]   it('returns an error if the message is invalid') => {}


test/server/graphql/v2/mutation/AccountMutations.test.ts
```diff
@@ -446,4 +449,135 @@ describe('server/graphql/v2/mutation/AccountMutations', () => {
...
+      expect(result.errors).to.not.exist;
+      expect(result.data.sendMessage.success).to.equal(true);
+
+      await waitForCondition(() => sendEmailSpy.callCount === 1);
+      expect(sendEmailSpy.callCount).to.equal(1);
+      expect(sendEmailSpy.args[0][1]).to.equal(
+        `New message from ${randomUser.collective.name} on Open Collective: Testing`,
+      );
+      expect(sendEmailSpy.args[0][2]).to.include(message);
+      expect(sendEmailSpy.args[0][2]).to.not.include(code);
```
Betree, line 526: Here we don't really guarantee that the injection did not happen. What if the code was simply turned into:

```html
<script>console.log('XSS')</script>
```

(simple quotes)

Would be great to test the exact message:

```suggestion
      expect(sendEmailSpy.args[0][2]).to.eq(`...`);
```

Betree (commented):
Great! Only one small feedback on the way we test XSS, the rest looks all good to merge

test/server/graphql/v2/mutation/AccountMutations.test.ts
```diff
@@ -446,4 +449,138 @@ describe('server/graphql/v2/mutation/AccountMutations', () => {
...
+      );
+
+      expect(result.errors).to.not.exist;
+      expect(result.data.sendMessage.success).to.equal(true);
+
+      await waitForCondition(() => sendEmailSpy.callCount === 1);
+      expect(sendEmailSpy.callCount).to.equal(1);
+      expect(sendEmailSpy.args[0][1]).to.equal(`New message from XSS Collective on Open Collective: Testing`);
+      expect(sendEmailSpy.args[0][2]).to.include(message);
+      expect(sendEmailSpy.args[0][2]).to.not.include('<script>');
```
Betree, line 529: If, somehow, the attack turns out creating a code like:

```
<code attr="1">test</code>
```

This test won't detect it. Can we use the full text here, i.e.:

```suggestion
      expect(sendEmailSpy.args[0][2]).to.eq(fullText);
```

snitin315: @Betree here full text is actually the whole HTML, that's why I didn't add that check.

We may also want to improve existing XSS test cases then:

https://github.com/opencollective/opencollective-api/blob/beed05187b621d24526107362377414716c71429/test/server/lib/notifications.test.ts#L185

<details>
<summary>  sendEmailSpy.args[0][2]   </summary>

```html
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
       <html xmlns="http://www.w3.org/1999/xhtml">
       <head>
         <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
         <title></title>
             <td></td>
             <td>
       
       <p style="color: #494B4D; font-family: 'Helvetica Neue'; line-height: 18px; font-size: 17px; padding: 1em
       Hi Test Collective 127ae549,
         <br><br>
       <a href="http://localhost:3000/user-24640536" style="text-decoration: none; color: #297EFF;">User Namec778c9a9</a> just sent you a message on Open Collective. Simply reply to this email to reply to the sender.
       </p>
       
       <table style="width: 100%;">
       </table><!-- 100% width table -->
       <!-- OpenCollective.com -->
       </body>
       
      </html>
```
</details>

snitin315: @Betree Done ✅

test/server/graphql/v2/mutation/AccountMutations.test.ts
```diff
@@ -446,4 +449,170 @@ describe('server/graphql/v2/mutation/AccountMutations', () => {
...
+        '[http://localhost:3000/static/images/email/logo-email-footer@2x.png]\n' +
+        '[http://localhost:3000]\n' +
+        '\n' +
+        'We can do great things together\n' +
+        '\n' +
+        'You can also follow us on Twitter [https://twitter.com/OpenCollect] or come chat\n' +
+        'with us on our public Slack channel [https://slack.opencollective.com].\n' +
+        '\n' +
+        'Made with ❤️ from all over the world\n' +
+        '[https://docs.opencollective.com/help/about/team]';
```
Betree, line 557: :ice_cream: Not a requested change: Would have been more readable (and probably easier for you) to just copy-paste the expectation in a template string:

```es6
const expectedMessage = `Hi Test Collective,

XSS Collective [http://localhost:3000/xss-collective] just sent you a message on
...
`
```

test/server/graphql/v2/mutation/AccountMutations.test.ts
```diff
@@ -446,4 +449,170 @@ describe('server/graphql/v2/mutation/AccountMutations', () => {
...
+
+      expect(sendEmailSpy.callCount).to.equal(1);
+      expect(sendEmailSpy.args[0][1]).to.equal(`New message from XSS Collective on Open Collective: Testing`);
+      expect(sendEmailSpy.args[0][3].text).to.equal(expectedMessage);
```
Betree, line 561: :no_entry_sign: Looking at `text` is not the proper way to test this: this property is generated from https://github.com/opencollective/opencollective-api/blob/a2018297a70ef372addc44290b5662562409f7af/scripts/compile-email.js#L581 by stripping all HTML content. In other works, it will never contain HTML content.

We should rather look at `html`:
```suggestion
      expect(sendEmailSpy.args[0][3].html).to.equal(expectedMessage);
```

snitin315: Updated to check the HTML content.

Betree (approved):

PR #8068, opened 2022-10-14
feat(OSC onboarding): Automatically approve collectives with enough invited admins

gustavlrsn:
Resolves https://github.com/opencollective/opencollective/issues/6052

# Description
This should automatically approve collectives if you have invited enough admins and pass Github validation, but freeze receiving contributions if there is not enough admins that have accepted their invite.

The freezing will only happen if you have checked the `Freeze collectives...` policy:

image


gustavlrsn:
On top of the GitHub validation we also currently automatically approve collectives if one of these criteria are met
- If there's a parent collective already approved by the host and user is admin of it
- If user is admin of the host

Should these be frozen from contributions if they do not have enough admins also? That is what is currently implemented in this PR at least.

gustavlrsn:
Noted that while on first load after applying with a collective that should be approved but frozen, the contributions boxes do not appear. However, on reload they do appear, but the user is blocked from contributing when clicking on them.

Also noted that while the collective does have the feature `RECEIVE_FINANCIAL_CONTRIBUTIONS` as `DISABLED`. 

`isFrozen` is `false`. Should this be `true`?

There is no message to tell the user that contributions are disabled until enough admins join. `isFrozen: true` would display this message, but would probably be too cryptic for the user to understand what is needed.
image

gustavlrsn:
Added a notification in https://github.com/opencollective/opencollective-frontend/pull/8308

Also adding the `policies` field to v1 graphql (which the collective page uses) so that we can get the information needed to display the notification.

kewitz (approved):

PR #8076, opened 2022-10-16
test: add unit tests for OAuth scopes

snitin315:
Resolve https://github.com/opencollective/opencollective/issues/5737

 
* [x]  For the lib: `test/server/graphql/common/scope-check.test.ts`. You can manually set the `userToken` on a `req` object created with `makeRequest` to test things in this file.
     * [x]  `checkScope`
         * [x]  Returns true if not using OAuth (aka. if there's no req.userToken)
         * [x]  Returns true if user token has scope
         * [x]  Returns false if doesn't have the scope
     * [x]  `enforceScope`
         * Same as the previous one, but throw if not allowed. Error message must be checked.
     * [x]  Check all `checkRemoteUserCanUse*` for 3 things:
         * Execute without errors if the scope is allowed by the user token
         * Execute without errors if not using OAuth (aka. if there's no req.userToken)
         * Throws when not authenticated
         * Throws if the scope is not available on the token
     * [x]  For GraphQL:
       * 1. Add a `oAuthGraphqlQueryV2` helper in `test/utils.js` like: `oAuthGraphqlQueryV2(query, variables, authorization)`
       * 2. In the future, this helper should be used for all queries/mutations enforcing OAuth. As a proof of concept, the helper should be added for the (["Add funds" mutation test](https://github.com/opencollective/opencollective-api/blob/9bfe08e3abc33575cba57fc737334c68d722700e/test/server/graphql/v2/mutation/AddFundsMutations.test.js#L58)), with something like:
        ```js
         it('verifies the scope', async () => {
           const result = await oAuthGraphqlQueryV2(query, variables, await fakeUserToken({ scope: [] });
           expect(result.errors[0]).to.eq(...);
         }
        ```

### DIFF
--- a/test/server/graphql/v2/mutation/AddFundsMutations.test.js
+++ b/test/server/graphql/v2/mutation/AddFundsMutations.test.js
@@ -2,8 +2,8 @@ import { expect } from 'chai';
 import gqlV2 from 'fake-tag';
 
 import { roles } from '../../../../../server/constants';
-import { fakeCollective, fakeProject, fakeTier, fakeUser } from '../../../../test-helpers/fake-data';
-import { graphqlQueryV2 } from '../../../../utils';
+import { fakeCollective, fakeProject, fakeTier, fakeUser, fakeUserToken } from '../../../../test-helpers/fake-data';
+import { graphqlQueryV2, oAuthGraphqlQueryV2 } from '../../../../utils';
 import * as utils from '../../../../utils';
 
 const addFundsMutation = gqlV2/* GraphQL */ `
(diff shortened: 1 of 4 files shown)


test/utils.js
```diff
@@ -226,6 +227,24 @@ export async function graphqlQueryV2(query, variables, remoteUser = null, jwtPay
...
+ */
+export async function oAuthGraphqlQueryV2(
+  query,
+  variables,
+  authorization,
+  remoteUser = null,
+  jwtPayload = null,
+  headers = {},
+) {
+  return graphqlQuery(query, variables, remoteUser, schemaV2, jwtPayload, headers, authorization);
```
Betree, line 245: Seems like this can be simplified

```suggestion
  userToken = null,
  jwtPayload = null,
  headers = {},
) {
  return graphqlQuery(query, variables, userToken.user, schemaV2, jwtPayload, headers, userToken);
```

test/server/graphql/v2/mutation/AddFundsMutations.test.js
```diff
@@ -118,6 +136,26 @@ describe('server/graphql/v2/mutation/AddFundsMutations', () => {
...
+        addFundsMutation,
+        {
+          account: { legacyId: collective.id },
+          fromAccount: { legacyId: randomUser.CollectiveId },
+          amount: { value: 20, currency: 'USD', valueInCents: 2000 },
+          description: 'add funds as admin',
+          hostFeePercent: 6,
+        },
+        userToken,
+        hostAdmin,
```
Betree, line 151: It's wrong to have a user token not match the user we're using it for; in fact, that can't happen in the context of real queries as resolved by `_authenticateUserByJwt`.

Betree (approved):
:+1:

PR #8078, opened 2022-10-17
AddFunds: fix check for parent ID

Betree:
Followup on https://github.com/opencollective/opencollective-api/pull/7778


znarf (approved):
Thank you!

PR #8079, opened 2022-10-17
Allow change virtual card name and assignee by collective admin

hdiniz:
Related https://github.com/opencollective/opencollective/issues/5788


znarf (approved):
Looks good!

PR #8081, opened 2022-10-18
Add support for memo field in add funds modal

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/6049

Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/8301


server/graphql/v2/object/Order.js
```diff
@@ -221,6 +221,20 @@ export const Order = new GraphQLObjectType({
           }
         },
       },
+      memo: {
+        type: GraphQLString,
+        description:
+          'Memo field which adds additional details about the order. For example in added funds this can be a note to mark what method (cheque, money order) the funds were received.',
+        async resolve(order, _, { remoteUser }) {
+          const collective = await models.Collective.findOne({ where: { id: order.CollectiveId } });
```
znarf, line 229: I would even do something like:

```
order.collective = order.collective || (await loaders.Collective.byId.load(order.CollectiveId));
```

znarf: Why fetch if it's already available in the order object? And still be defensive so we don't have to think if it's available or not.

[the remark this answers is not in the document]
SudharakaP: Thanks, makes more sense. 👍🏽

SudharakaP: Thanks. This makes more sense. I've refactored this part. 👍🏽

hdiniz (approved):
lgtm

Betree (approved):

PR #8086, opened 2022-10-18
Change Virtual Card Purchase Email Link to Edit the Expense

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6046

image

### DIFF
--- a/scripts/compile-email.js
+++ b/scripts/compile-email.js
@@ -81,6 +81,17 @@ data['collective.expense.missing.receipt'] = {
     viewLatestExpenses: 'https://opencollective.com/wwcodeaustin/expenses',
   },
 };
+data['virtualcard.purchase'] = {
+  amount: 1250,
+  currency: 'USD',
+  collective: {
+    slug: 'wwcodeaustin',
+    name: 'Women Who Code Austin',
+  },
+  expense: {
+    id: 25229,
+  },
+};
 data['user.card.claimed'] = {
   currency: 'USD',
   initialBalance: 10000,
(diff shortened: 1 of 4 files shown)


SudharakaP:
@alanna : Does the new wording change here looks good? Let me know if you like to make any further changes here. 😄

alanna:
looks good to me!

znarf (commented):
Can you share a screenshot of the rendering? There may be mistakes in the template.

znarf:
We can tackle in a follow up issue, but it's technically not true that the card was charged, at this stage it's just an authorization. Maybe it will be charged, maybe it will be cancelled. The amount may change.

Also, people don't have necessarily a receipt at this point so we should not be too aggressive trying to collect the receipt. This is why we have the other email when the transaction is confirmed.

alanna:
It's fine if they don't upload the receipt at this step, but if they do have it, we should make it easy for them, as they are in the mindset of completing the process and it's fresh in their mind.

znarf:
I liked in the previous copy that we were saying "upload a receipt **if you already have one**"

To be clear, my feedback is not blocking here.

alanna:
Oh yeah that extra phrase sounds fine

SudharakaP:
> Can you share a screenshot of the rendering? There may be mistakes in the template.

I've added a screenshot. And also added a template test to `compile-email.js` for this. 

> I liked in the previous copy that we were saying "upload a receipt **if you already have one**"
> 
> (To be clear, my feedback is not blocking here.)

That makes sense; I've re-added the "upload a receipt **if you already have one**" phrase back.

SudharakaP:
@znarf @alanna : Could you take a look once more as I've changed the template and verify everything looks good to merge?

PR #8089, opened 2022-10-19
Update dev environment stripe connected account token

hdiniz:
In the dev local environment, update stripe token for the connected account of Open Source Collective (id=9805,slug=opensource).


znarf (approved):
Good for me!

PR #8090, opened 2022-10-19
Fetch Credit Card info from Stripe and get rid of AliPay

kewitz:
I'm also typing `lib/stripe` so we can get access to type validation from their API.


server/paymentProviders/stripe/index.js
```diff
@@ -261,8 +257,6 @@ export default {
       }
       if (event.type === 'invoice.payment_succeeded') {
         return creditcard.webhook(requestBody, event);
-      } else if (event.type === 'charge.refund.updated') {
-        return alipay.webhook(requestBody, event);
       } else if (event.type === 'source.chargeable') {
```
znarf, line 260: While we're there, we can remove this.

kewitz: This was already removed in #8029. I'll rebase this after we merge that.

PR #8094, opened 2022-10-20
feat(Virtual cards): Add remaining limit to virtual cards

gustavlrsn:
Resolves https://github.com/opencollective/opencollective/issues/4556

## Description

Adds `remainingLimit` to the VirtualCard type in the API. 

It is calculated based on `spendingLimitAmount` minus expenses associated with the card, starting from the interval period start date.

Interval period start date is based on these definitions from the [Stripe docs](https://stripe.com/docs/api/issuing/cards/object#issuing_card_object-spending_controls-spending_limits-interval):

> daily: Limit applies to a day, starting at midnight UTC.
> 
> weekly: Limit applies to a week, starting on Sunday at midnight UTC.

I'm assuming this means the start of the week starting on Monday, i.e not including transactions on Sunday. `startOf("isoWeek")` in moment.js

> monthly: Limit applies to a month, starting on the 1st.
> 
> yearly: Limit applies to a year, starting on January 1st.


server/graphql/v2/object/VirtualCard.ts
```diff
@@ -103,6 +105,49 @@ export const VirtualCard = new GraphQLObjectType({
...
+            case 'PER_AUTHORIZATION':
+              return spendingLimitAmount;
+            default:
+              return null;
+          }
+
+          const sumExpensesInPeriod = await models.Expense.sum('amount', {
+            where: {
+              VirtualCardId: virtualCard.id,
+              ...(startOfInterval && { incurredAt: { [Op.gte]: startOfInterval } }),
```
znarf, line 143:
```suggestion
              ...(startOfInterval && { incurredAt: { [Op.gte]: startOfInterval } }),
              kind: [PROCESSING, PAID],
```

znarf: We have some charges marked as CANCELED and we should not count them.

znarf (commented):
Looking good today if we operate with the same currency.

May be more complicated in the future with multi-currencies.

server/graphql/v2/object/VirtualCard.ts
```diff
@@ -103,6 +106,50 @@ export const VirtualCard = new GraphQLObjectType({
...
+              break;
+            case 'PER_AUTHORIZATION':
+              return spendingLimitAmount;
+            default:
+              return null;
+          }
+
+          const sumExpensesInPeriod = await models.Expense.sum('amount', {
+            where: {
+              VirtualCardId: virtualCard.id,
```
hdiniz, line 143: We should probably add an index on this column to be performant on the production data.

gustavlrsn: Added an index to Expenses

hdiniz (approved):

PR #8095, opened 2022-10-20
Define enum for virtual card limit intervals

hdiniz:
Relates https://github.com/opencollective/opencollective/issues/5992


server/constants/virtual-cards.ts
```diff
@@ -0,0 +1,8 @@
+export enum VirtualCardLimitIntervals {
+  PER_AUTHORIZATION = 'per_authorization',
+  DAILY = 'daily',
+  WEEKLY = 'weekly',
+  MONTHLY = 'monthly',
+  YEARLY = 'yearly',
+  ALL_TIME = 'all_time',
+}
```
gustavlrsn, line 8: Would it make sense to have these be the same on both sides, i.e. evaluate to the capitalized version?

Would make it easier to use it in a place like this, if I'm assuming correctly we would still save it capitalized in the DB (to not have to migrate any data)?
https://github.com/opencollective/opencollective-api/pull/8094/files#diff-8ff10f8d838faa6fa003e4b73d3a8b70d2f8e571a2a388be7afe241a035c0b71R119-R136

hdiniz: You're right I think, we just have to keep in mind to lowercase it when sending to stripe.

PR #8099, opened 2022-10-20
Support all virtual card limit intervals

hdiniz:
Related https://github.com/opencollective/opencollective/issues/5992
Requires https://github.com/opencollective/opencollective-api/pull/8079

Adds support for all stripe virtual cards limit intervals.


znarf:
Looks good! 

Maybe the limits should be in the config instead of being constants.

Also, we should put whatever maximum make sense for Fiscal Host admins, so they stop editing the cards directly in Stripe when they hit the maximum.

server/graphql/v2/interface/Account.js
```diff
@@ -613,13 +612,8 @@ const accountFieldsDefinition = () => ({
...
       'Policies for the account. To see non-public policies you need to be admin and have the scope: "account".',
-    async resolve(account, _, req) {
-      const policies = account.data?.policies || {};
-      if (req.remoteUser?.isAdminOfCollective(account) && checkScope(req, 'account')) {
-        return policies;
-      }
-
-      return pick(policies, PUBLIC_POLICIES);
+    async resolve(account) {
+      return account;
```
gustavlrsn, line 616: We need to do a similar change in the v1 api because the same `Policies` v2 object is used there

hdiniz:
> Looks good!
> 
> Maybe the limits should be in the config instead of being constants.
> 
> Also, we should put whatever maximum make sense for Fiscal Host admins, so they stop editing the cards directly in Stripe when they hit the maximum.

I am adding this as a host policy that hosts can define, with a default set by our configuration, wdyt?

server/graphql/v2/object/Policies.ts
```diff
@@ -18,6 +27,28 @@ export const Policies = new GraphQLObjectType({
...
+      name: POLICIES.MAXIMUM_VIRTUAL_CARD_LIMIT_AMOUNT_FOR_INTERVAL,
+      type: new GraphQLObjectType({
+        name: POLICIES.MAXIMUM_VIRTUAL_CARD_LIMIT_AMOUNT_FOR_INTERVAL,
+        fields: () => ({
+          [VirtualCardLimitIntervals.ALL_TIME]: { type: GraphQLInt },
+          [VirtualCardLimitIntervals.DAILY]: { type: GraphQLInt },
+          [VirtualCardLimitIntervals.MONTHLY]: { type: GraphQLInt },
+          [VirtualCardLimitIntervals.PER_AUTHORIZATION]: { type: GraphQLInt },
+          [VirtualCardLimitIntervals.WEEKLY]: { type: GraphQLInt },
+          [VirtualCardLimitIntervals.YEARLY]: { type: GraphQLInt },
```
Betree, line 44: We can keep storing them as integers in the database, but out of consistency the GraphQL API should return them as amounts:

```suggestion
          [VirtualCardLimitIntervals.ALL_TIME]: { type: Amount },
          [VirtualCardLimitIntervals.DAILY]: { type: Amount },
          [VirtualCardLimitIntervals.MONTHLY]: { type: Amount },
          [VirtualCardLimitIntervals.PER_AUTHORIZATION]: { type: Amount },
          [VirtualCardLimitIntervals.WEEKLY]: { type: Amount },
          [VirtualCardLimitIntervals.YEARLY]: { type: Amount },
```

Betree (commented):
Nice one! Only one small feedback on the `Amount` type, the rest looks all good to merge

znarf (approved):
Looks good to me!

PR #8100, opened 2022-10-20
Add support for force booking single tickets for events

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5086

Front-end: https://github.com/opencollective/opencollective-frontend/pull/8312


server/graphql/v2/mutation/TierMutations.js
```diff
@@ -40,14 +40,22 @@ const tierMutations = {
...
-      });
+      };
+      let data;
+      if (args.tier.singleTicket !== undefined) {
+        const tier = await models.Tier.findOne({ where: { id: tierId } });
+        data = { ...tier.data, singleTicket: args.tier.singleTicket };
+        tierUpdateData.data = data;
+      }
+
+      return await tier.update(tierUpdateData);
```
Betree, line 58: Why reload the tier here? Can't we just do

```suggestion

      if (args.tier.singleTicket !== undefined) {
        tierUpdateData.data = { ...tier.data, singleTicket: args.tier.singleTicket };
      }

      return tier.update(tierUpdateData);
```

server/models/Collective.js
```diff
@@ -2619,14 +2619,25 @@ function defineModel() {
...
-            return models.Tier.update(tier, { where: { id: tier.id, CollectiveId: this.id } });
+            models.Tier.findOne({ where: { id: tier.id, CollectiveId: this.id } }).then(t => {
+              if (tier.invoiceTemplate) {
+                tier.data = { ...t.data, invoiceTemplate: tier.invoiceTemplate };
+              }
+              if (tier.singleTicket !== undefined) {
+                tier.data = { ...t.data, singleTicket: tier.singleTicket };
+              }
+              return models.Tier.update(tier, { where: { id: tier.id, CollectiveId: this.id } });
+            });
```
Betree, line 2631: Here also, why reload the tier?

SudharakaP: Apparently no as `tier.data` is always undefined here. I've missed this in a previous PR as well, for the `invoiceTemplate` condition. 🤔

server/graphql/v2/mutation/TierMutations.js
SudharakaP: `tier.data` is always undefined here which is why I obtained it using the modal. 🤔

Betree: Would be great to identify the root cause. If we can't fix it easily, then at least add a comment to explain while we're doing that; otherwise, the next engineer looking at it will rightfully refactor to simplify not knowing about the side-effect.

Note that @vaughanj10 [spotted](https://opencollective.slack.com/archives/C0471HKNW7J/p1665696797472689) some strange sequelize behavior with sequelize between `update` and `JSONB`  recently, it could be related.

server/models/Collective.js
SudharakaP: @Betree : And the reason this is always undefined is that we deliberately return an empty object as per, 

https://github.com/opencollective/opencollective-api/blob/b82593b50de6e78f05b7331137ebf4c7dec441b7/server/graphql/v1/CollectiveInterface.js#L1140-L1146. 

So I think here the reloading is justified. 🤔

server/graphql/v2/mutation/TierMutations.js
SudharakaP: @Betree : Sorry, I was testing this again and I think I was mixing this up with the other code below (https://github.com/opencollective/opencollective-api/pull/8100#discussion_r1013417226) and thought this also needs the `tier.data` reload. But here it seems that it works find with the simplification. I've changed the code to what you've suggested. One thing I noted though is that there's some issues with the new tier page. For example, `TierUpdateInput` was missing the `invoiceTemplate` field so I added it,

https://github.com/opencollective/opencollective-api/blob/9205842941eed704bb84a0535e121f16a2bcc3a9/server/graphql/v2/input/TierUpdateInput.ts#L59-L61

server/models/Collective.js
Betree: We already load the tier above with `this.getTiers().then(oldTiers =>`, it's just a matter of re-arranging the code and passing the right variable. It will make the code clearer and more performant.

SudharakaP: This part is no longer needed as with the new Tier page we are not using this at all. I've removed these changes.

Betree (approved):

PR #8104, opened 2022-10-21
Implement emailChanged / emailWaitingForValidationChanged

znarf:
Follow Up on https://github.com/opencollective/opencollective-api/pull/8096

### DIFF
--- a/server/models/User.js
+++ b/server/models/User.js
@@ -28,6 +28,7 @@ function defineModel() {
         unique: true,
         set(val) {
           if (val && val.toLowerCase) {
+            this.emailChanged = true;
             this.setDataValue('email', val.toLowerCase());
           }
         },
@@ -41,7 +42,7 @@ function defineModel() {
           },
           isBurnerEmail: function (val) {
             if (
-              this.changed('email') &&
+              this.emailChanged &&
               isEmailBurner(val.toLowerCase()) &&
               !emailLib.isAuthorizedEmailDomain(val.toLowerCase())
             ) {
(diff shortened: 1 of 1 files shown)


server/models/User.js
```diff
@@ -41,7 +42,7 @@ function defineModel() {
           },
           isBurnerEmail: function (val) {
             if (
-              this.changed('email') &&
+              this.emailChanged &&
```
Betree, line 45: @znarf I'm curious, why add our own flag rather than use `this.changed('email')`?

server/models/User.js
```diff
@@ -56,13 +57,19 @@ function defineModel() {
       emailWaitingForValidation: {
         type: DataTypes.STRING,
         unique: true,
+        set(val) {
+          if (val && val.toLowerCase) {
```
Betree, line 61: Fyi: https://github.com/opencollective/opencollective-api/pull/8185

server/models/User.js
```diff
@@ -28,6 +28,7 @@ function defineModel() {
         unique: true,
         set(val) {
           if (val && val.toLowerCase) {
+            this.emailChanged = true;
```
Betree, line 31: This won't play well if we decide to move the model to typescript

server/models/User.js
znarf: @Betree It was not working. And I could not find any documentation about it.

znarf: Thank you for the link to the documentation. Not sure why I missed it.

I clearly remember it was not working and this is why I had to do this follow up. Maybe its because we tried to use it in a validator hook and that was an unsupported  edge case ?

PR #8105, opened 2022-10-21
sumTransactions wip

znarf:


Betree (commented):
I love the simplicity and flexibility that this brings. 

I think `server/lib/budget.js` would greatly benefit from using Typescript.

PR #8111, opened 2022-10-25
Ignore deleted orders in queries

Betree:
Part of https://github.com/opencollective/opencollective/issues/6036
Resolve https://github.com/opencollective/opencollective/issues/6072

This PR addresses some performance issues related to queries where we don't filter deleted orders. Because indexes are built with `where deletedAt IS NULL`, we need this filter to happen to take advantage of them.

### DIFF
--- a/server/graphql/v2/object/Host.js
+++ b/server/graphql/v2/object/Host.js
@@ -698,7 +698,10 @@ export const Host = new GraphQLObjectType({
             `
             SELECT o.id FROM "Orders" o
             JOIN "Transactions" t ON t."OrderId" = o.id
-            WHERE t."HostCollectiveId" = :hostCollectiveId AND o.status = 'DISPUTED'
+            WHERE t."HostCollectiveId" = :hostCollectiveId
+            AND o.status = 'DISPUTED'
+            AND o."deletedAt" IS NULL
+            AND t."deletedAt" IS NULL
             LIMIT 1;
             `,
             {
--- a/server/lib/queries.js
+++ b/server/lib/queries.js
@@ -102,7 +102,10 @@ const getTotalAnnualBudgetForHost = HostCollectiveId => {
       `
   WITH
     "collectiveids" AS (
-      SELECT id FROM "Collectives" WHERE "HostCollectiveId"=:HostCollectiveId AND "isActive"=true
+      SELECT id FROM "Collectives"
+      WHERE "HostCollectiveId" = :HostCollectiveId
+      AND "isActive"=true
+      AND "deletedAt" IS NULL
     ),
     "monthlyOrdersWithAmountInHostCurrency" AS (
       SELECT o.id, MAX(o."CollectiveId") as "CollectiveId", MAX(t.currency) AS currency, MAX(t."amountInHostCurrency") as "amountInHostCurrency"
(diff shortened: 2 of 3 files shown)


server/graphql/v2/object/Host.js
```diff
@@ -698,7 +698,10 @@ export const Host = new GraphQLObjectType({
             `
             SELECT o.id FROM "Orders" o
             JOIN "Transactions" t ON t."OrderId" = o.id
-            WHERE t."HostCollectiveId" = :hostCollectiveId AND o.status = 'DISPUTED'
+            WHERE t."HostCollectiveId" = :hostCollectiveId
```
Betree, line 701: This one was [taking](https://sentry.io/organizations/open-collective/performance/oc-api:1237505d7dbe48048173596c16222b7c/?project=5199682&query=transaction.duration%3A%3C15m+http.method%3APOST&statsPeriod=24h&transaction=GraphQL%3A+ProcessExpense#span-8429a39a841ceb52) almost 800ms in prod.

Cost without deleted:
> Limit  (cost=0.09..45615.94 rows=1 width=4) (actual time=566.355..566.357 rows=0 loops=1)

With deleted:
> Limit  (cost=0.17..64.95 rows=1 width=4) (actual time=0.035..0.036 rows=0 loops=1)

server/lib/queries.js
```diff
@@ -102,7 +102,10 @@ const getTotalAnnualBudgetForHost = HostCollectiveId => {
       `
   WITH
     "collectiveids" AS (
-      SELECT id FROM "Collectives" WHERE "HostCollectiveId"=:HostCollectiveId AND "isActive"=true
+      SELECT id FROM "Collectives"
```
Betree, line 105: This is for https://github.com/opencollective/opencollective/issues/6072

server/graphql/v2/object/Host.js
Betree: Also, the sequelize version of this would have been much simpler (sequelize automatically filters deleted entries):

```es6
return Boolean(
  await models.Order.findOne({
    where: { status: 'DISPUTED' },
    include: [{ model: models.Transaction, where: { HostCollectiveId: host.id } }],
  }),
);
```

server/lib/security/fraud.ts
```diff
@@ -29,6 +29,7 @@ export const getUserStats = async (user: typeof models.User, interval?: string):
     `
     ${BASE_STATS_QUERY} 
     WHERE o."CreatedByUserId" = :userId
+    AND o."deletedAt" IS NULL
```
Betree, line 32: This one was also [adding](https://sentry.io/organizations/open-collective/performance/oc-api:f50babda231f4d3d9fa695b0c46b8e17/?project=5199682&query=transaction.duration%3A%3C15m+http.method%3APOST&statsPeriod=24h&transaction=GraphQL%3A+CreateOrder#span-a79f42a4df93e7af) around 800ms to the `createOrder`.

vaughanj10 (approved):
Looks good!

server/graphql/v2/object/Host.js
kewitz: Yep, would rather have it as Sequelize JS here.

PR #8113, opened 2022-10-26
Deprecate 2fa form input for payouts

hdiniz:
Deprecate unused 2FA form fields in favor of new 2FA flow.


server/graphql/v2/mutation/ExpenseMutations.ts
```diff
@@ -276,13 +275,9 @@ const expenseMutations = {
               type: GraphQLBoolean,
               description: 'Bypass automatic integrations (ie. PayPal, Transferwise) to process the expense manually',
             },
-            twoFactorAuthenticatorCode: {
```
SudharakaP, line 279: @hdiniz @Betree : This is used in `opencollective-tools` and impacts there; note,

https://github.com/opencollective/opencollective-tools/blob/2601ee0c7e3ab5cba8bc50016840a30dd3de2707/1kproject/pay-expenses.js#L41-L47

PR #8115, opened 2022-10-26
Reset expense.data when manually marking expense as paid

kewitz:
Solves: https://github.com/opencollective/opencollective/issues/6082


SudharakaP (approved):
💯

PR #8118, opened 2022-10-27
feat(stripe-in-review-orders): Create stripe webhooks for In Review Orders

vaughanj10:
Closes: https://github.com/opencollective/opencollective/issues/5982


server/paymentProviders/stripe/creditcard.js
```diff
@@ -486,13 +486,155 @@ async function closeDispute(event) {
...
+        paymentIntentTransaction,
+        0,
+        {
+          ...paymentIntentTransaction.data,
+          review: event,
+          refundTransactionId: paymentIntentTransaction.id,
+        },
+        null,
+        transactionGroup,
+      );
```
kewitz, line 598: As pointed out in the demo, we need to make sure what exactly is being refunded and who's paying for the payment processor fees.

kewitz (approved):

PR #8119, opened 2022-10-28
Permissions library

Betree:
First experiment around https://github.com/opencollective/opencollective/issues/6085


Betree:
Was an interesting experiment but we'll treat this as a real project

PR #8121, opened 2022-10-28
Improve Sock Puppet IP detection

kewitz:
Improvements:
- [x] Make sure that root-generated tokens do not affect lastLoginAt
- [x] Fix sock-puppet security check message
- [x] Set different sock-puppet security check levels based on the type of user correlation


Betree (approved):
Looking good!

PR #8124, opened 2022-10-31
GraphQL V2: handle invalid IDs

Betree:
Will resolve all issues in https://sentry.io/organizations/open-collective/issues/?project=5199682&query=is%3Aunresolved+column+%22NaN%22&referrer=issue-list&statsPeriod=14d and provide a better error message than:

> GraphQLError: column "nan" does not exist


znarf (approved):
Looks good!

PR #8126, opened 2022-10-31
Better link activities with collective

Betree:
Part of https://github.com/opencollective/opencollective/issues/5887

Todo: matching frontend PR to surface the account in relevant activities.

### DIFF
--- a/migrations/20221031094617-enhance-activities-links-collective-created.js
+++ b/migrations/20221031094617-enhance-activities-links-collective-created.js
@@ -2,6 +2,7 @@
 
 /** @type {import('sequelize-cli').Migration} */
 module.exports = {
+  // Splitting the migration in three steps to reduce the cost of each query
   async up(queryInterface) {
     console.time('Linking for COLLECTIVE_CREATED activities');
     await queryInterface.sequelize.query(`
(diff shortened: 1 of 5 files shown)


migrations/20221031094617-enhance-activities-links-collective-created.js
```diff
@@ -0,0 +1,42 @@
...
+      AND a.data -> 'collective' ->> 'id' IS NOT NULL
+    `);
+    console.timeEnd('Linking for COLLECTIVE_REJECTED activities');
+  },
+
+  async down() {
+    /**
+     * No rollback
+     */
+  },
```
SudharakaP, line 41: @Betree : I have been thinking, it's nice to add a lot here so that the user knows when they execute for example `npm run db:migrate:undo` that there's no rollback. Otherwise in the console it looks like it was rolled back. Helps later so that looking at the console log one can know which migrations rolled back and which ones did not. 🤔

Betree: @SudharakaP I like the idea! Something like this I suppose:

```suggestion
  async down() {
    console.log('Not rolling back the changes');
  },
```

For this one I'm gonna add a real rollback query, but I'll adapt this practice on new PRs from now on

SudharakaP: Yes, that's what I meant. 👍🏽

PR #8134, opened 2022-11-01
Expense Security Check: Follow-up and bug fixes

kewitz:
@vaughanj10 I'm suspending the credit card after a dispute is lost.
I also found a bug in credit card fraud checking that is being fixed here.


vaughanj10 (approved):
Looks good!

PR #8140, opened 2022-11-02
spike(collective-balances): Fast accurate balances for Collectives

vaughanj10:


server/paymentProviders/stripe/virtual-cards.js
```diff
@@ -157,10 +158,10 @@ export const processAuthorization = async (stripeAuthorization, stripeEvent) =>
   const currency = stripeAuthorization.pending_request.currency.toUpperCase();
   const amount = convertToStripeAmount(currency, Math.abs(stripeAuthorization.pending_request.amount));
   const collective = virtualCard.collective;
-  const balance = await collective.getBalanceWithBlockedFundsAmount({ currency });
+  const balance = await getCurrentBalance(collective.id, host.id, currency);
```
znarf, line 161: Could we keep `getBalanceWithBlockedFunds` as a name for the function for now?

It's generally more obvious so you don't mistake it with the normal "balance".

znarf:
The fundamentals are great!

I think we want to implement this in a more generic way and not just the virtual card case, can we add the fast balance in:
- lib/budget getBalances
- lib/budget getBalancesWithBlockedFunds

Those functions are used by the GraphQL loader and should be able to fetch the balances from multiple collectives at the same time. Ultimately, we'll need to turn getCurrentBalance/getCurrentBalanceWithBlockedFunds into getCurrentBalances/getCurrentBalancesWithBlockedFunds (plural) that accept a list of collective ids as input.

As the CurrentBalance look at "hostCurrency", it's our "v2" computation of the balance, it's great. But we need an opt out if Collectives are still on the v1 balance (`collective.settings.budget.version === 'v1'`).

znarf:
How can we ensure the view is not gonna hammer the database when we have a peak in transactions inserts (when we process recurring contributions) ?

znarf:
Following up in https://github.com/opencollective/opencollective-api/pull/8140

PR #8141, opened 2022-11-02
Add support to cancel recurring contributions when contributor is archived

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/6094

Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/8353

### DIFF
--- a/server/graphql/v1/mutations/collectives.js
+++ b/server/graphql/v1/mutations/collectives.js
@@ -528,6 +528,9 @@ export async function archiveCollective(_, args, req) {
     },
   );
 
+  // Cancel all subscriptions which the collective is contributing
+  await models.Order.cancelActiveOrdersByCollective(collective.id);
+
   // Mark main account as archived
   return collective.update({ deactivatedAt });
 }
--- a/server/lib/recurring-contributions.js
+++ b/server/lib/recurring-contributions.js
@@ -113,11 +113,6 @@ export async function processOrderWithSubscription(order, options) {
       csvEntry.error = 'The collective has been archived';
       collectiveIsArchived = true;
       cancelSubscription(order);
-    } else if (order.fromCollective.deactivatedAt) {
-      // This means the payee has been archived and the subscription should be cancelled.
-      orderProcessedStatus = 'failure';
-      csvEntry.error = 'The payee has been archived';
-      cancelSubscription(order);
     } else {
       try {
         transaction = await paymentsLib.processOrder(order);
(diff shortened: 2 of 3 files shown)


znarf:
I don't understand why we are using the Payee word here.

Isn't payee someone being paid, and usually used in the context of Expenses?

Betree:
> I don't understand why we are using the Payee word here.
> 
> Isn't payee someone being paid, and usually used in the context of Expenses?

I agree, we rarely use the term in the context of financial contributions. The only related case I am aware of is in `server/graphql/common/transactions.ts`. For financial contributions, the payee would be the one receiving the money so; the collective.

Payer could work, but we probably want to simply say "contributor" here.

Betree (changes requested):
This will only cancel Stripe, not PayPal. you can look at https://github.com/opencollective/opencollective-api/pull/8004 for an example of how it was implemented for a different use case.

SudharakaP:
> This will only cancel Stripe, not PayPal. you can look at #8004 for an example of how it was implemented for a different use case.

Thanks. I've corrected this by changing the implementation. Only sad thing with this one is that it immediately cancels the subscription when the user archives. It would be nice if it only cancels in the next iteration. 🤔 

> I don't understand why we are using the Payee word here.
> 
> Isn't payee someone being paid, and usually used in the context of Expenses?

Agreed. I've removed this work and changed everything to "contributor"

Betree:
> Only sad thing with this one is that it immediately cancels the subscription when the user archives. It would be nice if it only cancels in the next iteration. thinking

You mean on the next occurrence of the recurring contribution? I don't really agree, if I click on "Archive my account", get a modal saying "Your recurring contribution is going to be cancelled" and still click "Confirm", I expect my recurring contribution to be immediately cancelled - not in 1 year (in the case of a yearly contribution)

Betree:
Assigning @hdiniz for the final review, he has a good understanding of the context around this issue

hdiniz (approved):

SudharakaP:
> > Only sad thing with this one is that it immediately cancels the subscription when the user archives. It would be nice if it only cancels in the next iteration. thinking
> 
> You mean on the next occurrence of the recurring contribution? I don't really agree, if I click on "Archive my account", get a modal saying "Your recurring contribution is going to be cancelled" and still click "Confirm", I expect my recurring contribution to be immediately cancelled - not in 1 year (in the case of a yearly contribution)

Yeah that makes sense. 😄

PR #8142, opened 2022-11-03
Improvements on Expense info getter

Betree:
- `Expense.type` was set to TransactionTypes.DEBIT, moved it to the real type
- Removed `public` as it wasn't used


SudharakaP (approved):
💯

PR #8147, opened 2022-11-03
Persist and expose ExpenseAttachedFile name

kewitz:


Betree (approved):
Looks perfect, that will be a great improvement until we implement a better strategy for handling images

PR #8155, opened 2022-11-05
Add Support for Fund Received Date (`AddFundModal.js` and `ConfirmContributionModal.js`)

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5693

Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/8365


server/graphql/common/orders.js
```diff
@@ -75,6 +75,10 @@ export async function addFunds(order, remoteUser) {
     orderData.data.memo = order.memo;
   }
 
+  if (!isNil(order.fundReceivedDate)) {
+    orderData.data.fundReceivedDate = order.fundReceivedDate;
+  }
+
```
Betree, line 81: Not a fan of putting this one in `data`. It seems like something we're going to set for all `Orders` and that we may want to filter/order on in the future. So a new column makes sense.

Could we re-use `order.processedAt` for that? I think the current definition of the column matches what we want.

server/graphql/v2/input/OrderUpdateInput.ts
```diff
@@ -29,5 +30,9 @@ export const OrderUpdateInput = new GraphQLInputObjectType({
       type: GraphQLFloat,
       description: 'Host fee percent to be applied to the order',
     },
+    fundReceivedDate: {
```
znarf, line 33: This doesn't follow the pattern we have for dates in the API.

Can we try to get new fields in the API consistent when we introduce them?

SudharakaP: Thanks. This is now renamed to `processedAt`.

server/graphql/common/orders.js
SudharakaP: Makes sense. I've re-used `order.processedAt`. There was a few changes that needed to be done in the [`payments.js`](https://github.com/opencollective/opencollective-api/blob/759432476ab64df434596b25dcc7a768e2d50090/server/lib/payments.js#L548) as otherwise the `processedAt` date will be overridden.

server/lib/payments.js
```diff
@@ -555,7 +555,7 @@ export const executeOrder = async (user, order, options = {}) => {
   if (!order) {
     return Promise.reject(new Error('No order provided'));
   }
-  if (order.processedAt) {
+  if (order.processedAt && !options.isAddedFund) {
```
Betree, line 558: Maybe add a small comment there to explain this check?

Betree (approved):
Looks good to me. Let's maybe double-check with @znarf that `processedAt` is the right field to store this. Looking at the code, I think it makes sense.

server/lib/payments.js
SudharakaP: I've added a small comment. Let me know if that looks good.

PR #8159, opened 2022-11-08
graphql: add host parameter on accounts query

znarf:


server/graphql/v2/query/collection/AccountsCollectionQuery.ts
```diff
@@ -120,10 +125,16 @@ const AccountsCollectionQuery = {
     } else {
       const cleanTerm = args.searchTerm?.trim();
 
+      let hostCollectiveIds;
+      if (args.host) {
+        const hosts = await fetchAccountsWithReferences(args.host);
+        hostCollectiveIds = hosts.map(host => host.id);
+      }
+
```
Betree, line 133: Cheap optimizations:

```suggestion
      const hostCollectiveIds = await fetchAccountsIdsWithReference(args.host);
```

Betree (approved):

PR #8162, opened 2022-11-09
Make budget version default to v2 

znarf:
- freeze collective with issues on v1 (migration)
- count in `hostCurrency` vs `currency` (v2)

Fix: https://github.com/opencollective/opencollective/issues/5858

### DIFF
--- a/test/server/models/Collective.test.js
+++ b/test/server/models/Collective.test.js
@@ -98,6 +98,7 @@ describe('server/models/Collective', () => {
       createdAt: new Date('2016-06-15'),
       amount: 25000,
       amountInHostCurrency: 25000,
+      paymentProcessorFeeInHostCurrency: -2500,
       netAmountInCollectiveCurrency: 22500,
       currency: 'USD',
       type: 'CREDIT',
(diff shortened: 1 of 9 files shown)


test/server/models/Collective.test.js
```diff
@@ -480,16 +483,30 @@ describe('server/models/Collective', () => {
...
+      let sum = 0;
+      transactions.map(
+        t =>
+          (sum +=
+            (t.amountInHostCurrency || 0) +
+            (t.hostFeeInHostCurrency || 0) +
+            (t.platformFeeInHostCurrency || 0) +
+            (t.paymentProcessorFeeInHostCurrency || 0)),
+      );
+      expect(balance).to.equal(sum);
```
Betree, line 504: I often have a preference for hardcoding values in that kind of situation, otherwise there's always a risk of reproducing in the tests the same error that's in the code. But this one is fairly simple, I don't have a strong opinion.

Betree (approved):
Looking good! I like how explicit this transition strategy is in the code

PR #8166, opened 2022-11-10
Add time series stats to account collection

gustavlrsn:
This is adding a `AccountCollectionStats` field to `AccountCollection` under `stats`, so that you can query for time series information for the particular set of accounts that you have filtered out:

```graphql
query SearchAccounts($dateFrom: DateTime, $dateTo: DateTime) {
  accounts(type: [COLLECTIVE], tag: ["civic tech"], limit: 100) {
    totalCount
    nodes {
      id
      slug
    }
    stats(
      dateFrom: $dateFrom
      dateTo: $dateTo
      timeUnit: YEAR
      includeChildren: true
    ) {
      timeUnit
      contributionsCountTimeSeries {
        nodes {
          date
          count
        }
      }
      totalReceivedTimeSeries {
        nodes {
          date
          amount {
            value
            currency
          }
        }
      }
    }
  }
}
```

Current implementation is that it will only query stats for the collectives that are within the limit.

Since account collection is not limited to a single host I added a new query to get transaction time series for any set of CollectiveIds. Also adding a query for getting transactions **count** time series data.

Did some reorganization to be able to put the new queries in `lib/queries.js`, commented below.

### DIFF
--- a/server/graphql/v2/collection/AccountCollection.js
+++ b/server/graphql/v2/collection/AccountCollection.js
@@ -21,14 +21,10 @@ const AccountCollection = new GraphQLObjectType({
       stats: {
         type: new GraphQLNonNull(AccountCollectionStats),
         args: {
-          ...TimeSeriesArgs,
           includeChildren: { type: GraphQLBoolean, defaultValue: false },
         },
+        description: 'Stats for the returned results (i.e. accounts within the limit)',
         async resolve(collection, args) {
-          const dateFrom = args.dateFrom ? moment(args.dateFrom) : null;
-          const dateTo = args.dateTo ? moment(args.dateTo) : null;
-          const timeUnit = args.timeUnit || getTimeUnit(getNumberOfDays(dateFrom, dateTo, {}) || 1);
-
           const collectiveIds = collection.nodes.map(c => c.id);
 
           if (args.includeChildren) {
@@ -38,7 +34,7 @@ const AccountCollection = new GraphQLObjectType({
             });
             collectiveIds.push(...childCollectives.map(c => c.id));
           }
-          return { timeUnit, dateFrom, dateTo, collectiveIds };
+          return { collectiveIds };
         },
       },
     };
--- a/server/graphql/v2/collection/AccountCollection.js
+++ b/server/graphql/v2/collection/AccountCollection.js
@@ -1,7 +1,5 @@
-import { GraphQLBoolean, GraphQLList, GraphQLNonNull, GraphQLObjectType } from 'graphql';
-import { Op } from 'sequelize';
+import { GraphQLList, GraphQLNonNull, GraphQLObjectType } from 'graphql';
 
-import models from '../../../models';
 import { Account } from '../interface/Account';
 import { Collection, CollectionFields } from '../interface/Collection';
 import { AccountCollectionStats } from '../object/AccountCollectionStats';
@@ -18,21 +16,9 @@ const AccountCollection = new GraphQLObjectType({
       },
       stats: {
         type: new GraphQLNonNull(AccountCollectionStats),
-        args: {
-          includeChildren: { type: GraphQLBoolean, defaultValue: false },
-        },
         description: 'Stats for the returned results (i.e. accounts within the limit)',
-        async resolve(collection, args) {
-          const collectiveIds = collection.nodes.map(c => c.id);
-
-          if (args.includeChildren) {
-            const childCollectives = await models.Collective.findAll({
-              attributes: ['id'],
-              where: { ParentCollectiveId: { [Op.in]: collectiveIds } },
-            });
-            collectiveIds.push(...childCollectives.map(c => c.id));
-          }
-          return { collectiveIds };
+        resolve(collection) {
+          return collection;
         },
       },
     };
--- a/server/graphql/v2/collection/AccountCollection.js
+++ b/server/graphql/v2/collection/AccountCollection.js
@@ -1,6 +1,8 @@
 import { GraphQLBoolean, GraphQLList, GraphQLNonNull, GraphQLObjectType } from 'graphql';
 import moment from 'moment';
+import { Op } from 'sequelize';
 
+import models from '../../../models';
 import { Account } from '../interface/Account';
 import { Collection, CollectionFields } from '../interface/Collection';
 import { AccountCollectionStats } from '../object/AccountCollectionStats';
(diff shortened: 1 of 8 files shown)


server/graphql/v2/object/Host.js
```diff
@@ -626,15 +626,13 @@ export const Host = new GraphQLObjectType({
...
-              TransactionTypes.DEBIT,
-              host.id,
-              timeUnit,
+            const amountDataPoints = await HostMetricsLib.getTransactionsTimeSeries(host.id, timeUnit, {
+              type: TransactionTypes.DEBIT,
+              kind: TransactionKind.EXPENSE,
               collectiveIds,
               dateFrom,
               dateTo,
-            );
```
gustavlrsn, line 637: Switched this to use a query from `host-metrics.js`, where I added a new version to get the time series not grouped by `kind`, that this deprecated query did not do.

gustavlrsn: Did this to clear up the namespace in `lib/queries.js` to add the new `getTransactionsTimeSeries` query not relying on hostId

server/graphql/v2/collection/AccountCollection.js
```diff
@@ -13,6 +16,29 @@ const AccountCollection = new GraphQLObjectType({
...
+          const dateTo = args.dateTo ? moment(args.dateTo) : null;
+          const timeUnit = args.timeUnit || getTimeUnit(getNumberOfDays(dateFrom, dateTo, {}) || 1);
+
+          const collectiveIds = collection.nodes.map(c => c.id);
+
+          if (args.includeChildren) {
+            const childCollectiveIds = await Promise.all(
+              collection.nodes.map(async c => await c.getChildren({ attributes: ['id'] })),
+            );
+            collectiveIds.push(...childCollectiveIds.flat().map(c => c.id));
```
Betree, line 36: Fetching 100 accounts with `includeChildren` will trigger 100 SQL queries just to fetch the list of children IDs. Can we improve that?

Betree (commented):
The API itself looks good to me, we need to be aware that the stats will always be impacted by pagination; the stats returned will only be for the current page

server/graphql/v2/collection/AccountCollection.js
```diff
@@ -13,6 +18,29 @@ const AccountCollection = new GraphQLObjectType({
       nodes: {
         type: new GraphQLList(Account),
       },
+      stats: {
```
znarf, line 21: Add a description and make clear that stats are computed on the returned results, not everything.

server/graphql/v2/object/AccountCollectionStats.ts
```diff
@@ -0,0 +1,41 @@
...
+import { resultsToAmountNode } from '../object/HostMetricsTimeSeries';
+import { TimeSeriesAmount } from '../object/TimeSeriesAmount';
+import { TimeSeriesCount } from '../object/TimeSeriesCount';
+
+export const AccountCollectionStats = new GraphQLObjectType({
+  name: 'AccountCollectionStats',
+  description: 'Account collection stats',
+  fields: () => ({
+    ...getTimeSeriesFields(),
+    contributionsCountTimeSeries: {
```
znarf, line 16: transactionsCountTimeSeries

server/graphql/v2/object/AccountCollectionStats.ts
```diff
@@ -0,0 +1,41 @@
...
+      type: new GraphQLNonNull(TimeSeriesCount),
```
znarf, line 17: Add `kind` and `type` parameters.

server/graphql/v2/object/AccountCollectionStats.ts
```diff
@@ -0,0 +1,41 @@
...
+      description: 'Time series of the number of contributions to these accounts',
+      resolve: async ({ collectiveIds, dateFrom, dateTo, timeUnit }) => {
+        const kind = [TransactionKind.CONTRIBUTION];
```
znarf, line 20: Read from parameter.

server/graphql/v2/object/AccountCollectionStats.ts
```diff
@@ -0,0 +1,41 @@
...
+        const transactionParams = { type: TransactionTypes.CREDIT, kind, dateFrom, dateTo, collectiveIds };
```
znarf, line 21: Use type parameter.

server/graphql/v2/object/AccountCollectionStats.ts
```diff
@@ -0,0 +1,41 @@
...
+        const countNodes = collectiveIds.length
+          ? await queries.getTransactionsCountTimeSeries(timeUnit, transactionParams)
+          : [];
+        return { dateFrom, dateTo, timeUnit, nodes: countNodes };
+      },
+    },
+    totalReceivedTimeSeries: {
```
znarf, line 28: Same here `totalTransactionsAmountTimeSeries`

gustavlrsn: renamed it to `transactionsAmountTimeSeries`, I think it fits better with `transactionsCountTimeSeries`?

server/graphql/v2/object/AccountCollectionStats.ts
```diff
@@ -1,36 +1,63 @@
...
+
+        const transactionParams = { type: args.type, kind: args.kind, dateFrom, dateTo, collectiveIds };
         const countNodes = collectiveIds.length
           ? await queries.getTransactionsCountTimeSeries(timeUnit, transactionParams)
           : [];
         return { dateFrom, dateTo, timeUnit, nodes: countNodes };
       },
     },
-    totalReceivedTimeSeries: {
+    transactionsTimeSeries: {
```
znarf:
```suggestion
    transactionsAmountTimeSeries: {
```

gustavlrsn:
Updated API, from simplifying to only one time series, returning amount with count together:
```graphql
query SearchAccounts(
  $dateFrom: DateTime
  $dateTo: DateTime
  $kind: [TransactionKind]
  $type: TransactionType
) {
  accounts {
    totalCount
    stats {
      transactionsTimeSeries(
        dateFrom: $dateFrom
        dateTo: $dateTo
        kind: $kind
        type: $type
        includeChildren: true
      ) {
        timeUnit
        nodes {
          date
          count
          amount {
            value
            valueInCents
            currency
          }
        }
      }
    }
    nodes {
      id
      name
      slug
    }
  }
}
```

znarf (approved):

PR #8173, opened 2022-11-14
Ability to fetch submitted expenses for user

Betree:
- Introduce a `createdByUserAccount` argument on `expenses`. As proposed by @kewitz, we could further refactor this to have a direct `individual.submittedExpenses` field in the future. Not doing this now as it can be done later as a non-breaking change.
- Add a new `CREATED_BY_ACCOUNT` option for  `currencySource`. This allows us to fetch all amounts to display in the user's currency.

### DIFF
--- a/server/graphql/v2/object/Expense.js
+++ b/server/graphql/v2/object/Expense.js
@@ -98,6 +98,13 @@ const Expense = new GraphQLObjectType({
           } else if (args.currencySource === 'HOST') {
             const host = await loadHostForExpense(expense, req);
             currency = host?.currency;
+          } else if (args.currencySource === 'CREATED_BY_ACCOUNT') {
+            expense.User = expense.User || (await req.loaders.User.byId.load(expense.UserId));
+            if (expense.User) {
+              expense.User.collective =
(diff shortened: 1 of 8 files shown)


server/graphql/v2/object/Expense.js
```diff
@@ -98,6 +98,13 @@ const Expense = new GraphQLObjectType({
           } else if (args.currencySource === 'HOST') {
             const host = await loadHostForExpense(expense, req);
             currency = host?.currency;
+          } else if (args.currencySource === 'CREATED_BY_ACCOUNT') {
+            expense.User = expense.User || (await req.loaders.User.byId.load(expense.UserId));
+            if (expense.User) {
+              expense.User.collective =
+                expense.User.collective || (await req.loaders.Collective.byId.load(expense.User.CollectiveId));
+              currency = expense.User?.collective?.currency || 'USD';
+            }
```
kewitz, line 107: 👍🏼

kewitz (approved):
Cool! Nice work adding the support for calculating the total in the User currency!

PR #8188, opened 2022-11-18
Make TagStats amount nullable

Betree:
Resolve https://sentry.io/organizations/open-collective/issues/3730198773

The field was introduced in https://github.com/opencollective/opencollective-api/pull/7853 as non-nullable, but we don't set any `amount` in many places where we return a `TagStats`.


kewitz (commented):
Nice catch. 🎯

PR #8191, opened 2022-11-18
Deprecate confirmOrder V1

Betree:
Require https://github.com/opencollective/opencollective-frontend/pull/8393

### DIFF
--- a/server/graphql/v1/mutations.js
+++ b/server/graphql/v1/mutations.js
@@ -297,7 +297,7 @@ const mutations = {
   },
   confirmOrder: {
     type: OrderType,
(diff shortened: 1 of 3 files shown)


server/graphql/v1/mutations.js
```diff
@@ -297,6 +297,7 @@ const mutations = {
   },
   confirmOrder: {
     type: OrderType,
+    deprecationReason: '2022-11-18: This endpoint has been moved to GQLV2',
```
znarf, line 300:
```suggestion
    deprecationReason: '2022-11-18: This mutation has been moved to GQLV2',
```

Betree:
@znarf I copy-pasted the message from `createOrder` without paying much attention, I'll update it as well

PR #8198, opened 2022-11-21
Change permissions for MEMBER role on `createOder`

Betree:
Removes the historical ability for `MEMBER` role to create orders using the payment method of the shared account. This was not documented and, according to the query below, never used.

```sql
select * from "Orders" o
INNER JOIN "Users" u ON o."CreatedByUserId" = u.id
INNER JOIN "Members" m ON m."MemberCollectiveId" = u."CollectiveId" AND m."CollectiveId" = o."FromCollectiveId" AND m.role = 'MEMBER'
LEFT JOIN "Members" admin ON m."MemberCollectiveId" = u."CollectiveId" AND m."CollectiveId" = o."FromCollectiveId" AND admin.role = 'ADMIN'
WHERE admin.id IS NULL
ORDER BY o."createdAt" DESC
```

### DIFF
--- a/server/models/PaymentMethod.js
+++ b/server/models/PaymentMethod.js
@@ -408,7 +408,7 @@ PaymentMethod.prototype.getBalanceForUser = async function (user) {
 
   // Independently of the balance of the external source, the owner of the payment method
   // may have set up a monthlyLimitPerMember or an initialBalance
-  if (!this.initialBalance && (!this.monthlyLimitPerMember || (user && user.isAdmin(this.CollectiveId)))) {
+  if (!this.initialBalance && !this.monthlyLimitPerMember) {
     return { amount: balanceAmount, currency: this.currency };
   }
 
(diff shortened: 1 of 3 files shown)


server/models/PaymentMethod.js
```diff
@@ -409,7 +409,7 @@ function defineModel() {
 
     // Independently of the balance of the external source, the owner of the payment method
     // may have set up a monthlyLimitPerMember or an initialBalance
-    if (!this.initialBalance && (!this.monthlyLimitPerMember || (user && user.isAdmin(this.CollectiveId)))) {
+    if (!this.initialBalance && !this.monthlyLimitPerMember) {
```
Betree, line 412: I changed this behavior because, after changing the role in `test/server/graphql/v1/createOrder.test.js` to make sure the `createOrder` was tested with an `ADMIN`, the test failed as the expected balance went to "MAX_INTEGER".

It seems that, in this portion of the code, the `monthlyLimitPerMember` was only enforced for `MEMBER`, not `ADMIN`. While this was certainly the expected behavior in the initial implementation, it does not match the current one.

PR #8199, opened 2022-11-21
Stop relying on pm.creditCardInfo, 

znarf:
Only token should be looked at. This is simpler for the implementer and more solid.

Make brand, country, expMonth, expYear optional. Should actually be deprecated?

### DIFF
--- a/server/graphql/v2/input/PaymentMethodInput.ts
+++ b/server/graphql/v2/input/PaymentMethodInput.ts
@@ -83,25 +83,26 @@ export const getLegacyPaymentMethodFromPaymentMethodInput = async (
     const paymentMethod = {
       service: PAYMENT_METHOD_SERVICE.STRIPE,
       type: PAYMENT_METHOD_TYPE.CREDITCARD,
-      name: pm.name,
+      name: token.card.last4,
       save: pm.isSavedForLater,
       token: pm.creditCardInfo.token,
       data: {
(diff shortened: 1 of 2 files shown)


kewitz (approved):
Nice, refactor. I double-checked with existing data and noticed an always null "name" property.

server/graphql/v2/input/PaymentMethodInput.ts
```diff
@@ -83,25 +83,25 @@ export const getLegacyPaymentMethodFromPaymentMethodInput = async (
...
       service: PAYMENT_METHOD_SERVICE.STRIPE,
       type: PAYMENT_METHOD_TYPE.CREDITCARD,
-      name: pm.name,
+      name: token.card.last4,
       save: pm.isSavedForLater,
       token: pm.creditCardInfo.token,
       data: {
-        ...pick(token.card, ['brand', 'country', 'fullName', 'funding', 'zip', 'fingerprint']),
-        name: token.card.name,
+        ...pick(pm.creditCardInfo, ['fullName', 'zip']), // Not returned by Stripe
```
znarf, line 90: We can also remove fullName, not used since 2017.

znarf:
@kewitz Where exactly have you noticed the `name` property null?

kewitz:
@znarf, the `data.name` should hold the cardholder name, but since we used the simplified card component that only collects the card number, expiry date and cvc, we're not saving anything when we generate the token for the card.

You can check all recent `paymentMethod.data.name`, they're all empty. We can change this in the future if we change the component in the frontend.

PR #8206, opened 2022-11-22
Add stripe customer payment method

hdiniz:
Creates payment intents using a single customerId per stripe account and stores attached payment methods for reuse by same customerId

### DIFF
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -19,7 +19,7 @@ import {
 import { roles } from '../../../constants';
 import activities from '../../../constants/activities';
 import status from '../../../constants/order_status';
-import { PAYMENT_METHOD_SERVICE } from '../../../constants/paymentMethods';
+import { PAYMENT_METHOD_SERVICE, PAYMENT_METHOD_TYPE } from '../../../constants/paymentMethods';
 import { purgeAllCachesForAccount } from '../../../lib/cache';
 import logger from '../../../lib/logger';
 import stripe, { convertToStripeAmount } from '../../../lib/stripe';
@@ -108,6 +108,9 @@ const orderMutations = {
       const collective = await loadAccount(order.toAccount);
       const expectedCurrency = (tier && tier.currency) || collective.currency;
       const paymentMethod = await getLegacyPaymentMethodFromPaymentMethodInput(order.paymentMethod);
+      if (order.paymentMethod?.paymentIntentId) {
+        paymentMethod.paymentIntentId = order.paymentMethod?.paymentIntentId;
+      }
 
       // Ensure amounts are provided with the right currency
       ['platformTipAmount', 'amount', 'tax.amount'].forEach(field => {
@@ -653,6 +656,13 @@ const orderMutations = {
     async resolve(_, args, req) {
       const paymentIntentInput = args.paymentIntent;
 
+      const toAccount = await fetchAccountWithReference(paymentIntentInput.toAccount, { throwIfMissing: true });
+      const hostStripeAccount = await toAccount.getHostStripeAccount();
+      const host = await toAccount.getHostCollective();
+
+      const isPlatformHost = hostStripeAccount.username === config.stripe.accountId;
+
+      let stripeCustomerId;
       let fromAccount;
       if (req.remoteUser) {
         fromAccount =
@@ -662,17 +672,49 @@ const orderMutations = {
         if (!req.remoteUser.isAdminOfCollective(fromAccount)) {
           throw new Unauthorized();
         }
-      }
 
-      const toAccount = await fetchAccountWithReference(paymentIntentInput.toAccount, { throwIfMissing: true });
-      const hostStripeAccount = await toAccount.getHostStripeAccount();
-      const host = await toAccount.getHostCollective();
+        let pm = await models.PaymentMethod.findOne({
+          where: {
+            CollectiveId: fromAccount.id,
+            service: PAYMENT_METHOD_SERVICE.STRIPE,
+            type: PAYMENT_METHOD_TYPE.PAYMENT_INTENT,
+            data: {
+              stripeAccount: hostStripeAccount.username,
+            },
+          },
+        });
+
+        const customer = await stripe.customers.create(
+          {
+            email: req.remoteUser.email,
+            description: `${config.host.website}/${fromAccount.slug}`,
+          },
+          !isPlatformHost
+            ? {
+                stripeAccount: hostStripeAccount.username,
+              }
+            : undefined,
+        );
+
+        if (!pm) {
+          pm = await models.PaymentMethod.create({
+            customerId: customer.id,
+            CreatedByUserId: req.remoteUser.id,
+            CollectiveId: fromAccount.id,
+            service: PAYMENT_METHOD_SERVICE.STRIPE,
+            type: PAYMENT_METHOD_TYPE.PAYMENT_INTENT,
+            data: {
+              stripeAccount: hostStripeAccount.username,
+            },
+          });
+        }
+
+        stripeCustomerId = pm.customerId;
+      }
 
       const totalOrderAmount = getValueInCentsFromAmountInput(paymentIntentInput.amount);
 
       const currency = paymentIntentInput.currency;
-
-      const isPlatformHost = hostStripeAccount.username === config.stripe.accountId;
 
       const paymentMethodTypes =
         host.settings.stripe?.payment_method_types ||
@@ -685,6 +727,9 @@ const orderMutations = {
       try {
         const paymentIntent = await stripe.paymentIntents.create(
           {
+            customer: stripeCustomerId,
+            // eslint-disable-next-line camelcase
+            setup_future_usage: 'off_session',
             description: `Contribution to ${toAccount.name}`,
             amount: convertToStripeAmount(currency, totalOrderAmount),
             currency: paymentIntentInput.amount.currency.toLowerCase(),
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -684,19 +684,19 @@ const orderMutations = {
           },
         });
 
-        const customer = await stripe.customers.create(
-          {
-            email: req.remoteUser.email,
-            description: `${config.host.website}/${fromAccount.slug}`,
-          },
-          !isPlatformHost
-            ? {
-                stripeAccount: hostStripeAccount.username,
-              }
-            : undefined,
-        );
-
         if (!pm) {
+          const customer = await stripe.customers.create(
+            {
+              email: req.remoteUser.email,
+              description: `${config.host.website}/${fromAccount.slug}`,
+            },
+            !isPlatformHost
+              ? {
+                  stripeAccount: hostStripeAccount.username,
+                }
+              : undefined,
+          );
+
           pm = await models.PaymentMethod.create({
             customerId: customer.id,
             CreatedByUserId: req.remoteUser.id,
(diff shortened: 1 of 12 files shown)


server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -662,18 +672,50 @@ const orderMutations = {
...
+          {
+            email: req.remoteUser.email,
+            description: `${config.host.website}/${fromAccount.slug}`,
+          },
+          !isPlatformHost
+            ? {
+                stripeAccount: hostStripeAccount.username,
+              }
+            : undefined,
+        );
```
kewitz, line 697: Will this return the existing customer if the email was already created, or are we creating one disposable customer per intent?

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -662,18 +672,50 @@ const orderMutations = {
...
+
+        if (!pm) {
+          pm = await models.PaymentMethod.create({
+            customerId: customer.id,
+            CreatedByUserId: req.remoteUser.id,
+            CollectiveId: fromAccount.id,
+            service: PAYMENT_METHOD_SERVICE.STRIPE,
+            type: PAYMENT_METHOD_TYPE.PAYMENT_INTENT,
```
kewitz, line 705: If we're going to set SEPA and ACH payment method types, I think we should deprecate the generic `PAYMENT_INTENT`.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -685,6 +727,9 @@ const orderMutations = {
       try {
         const paymentIntent = await stripe.paymentIntents.create(
           {
+            customer: stripeCustomerId,
+            // eslint-disable-next-line camelcase
+            setup_future_usage: 'off_session',
```
kewitz, line 732: 💯

server/paymentProviders/stripe/index.js
```diff
@@ -53,6 +53,10 @@ export default {
     default: creditcard,
     creditcard,
     paymentintent,
+    // eslint-disable-next-line camelcase
+    us_bank_account: paymentintent,
+    // eslint-disable-next-line camelcase
+    sepa_debit: paymentintent,
```
kewitz, line 59: I like that this is explicit about having different types using the same handler, but at the same time, I'm not sure this is not a manifestation that we don't need to store payment methods at the SEPA/ACH granularity.

server/paymentProviders/stripe/webhook.ts
```diff
@@ -424,6 +425,75 @@ export const reviewClosed = async (event: Stripe.Response<Stripe.Event>) => {
...
+      {
+        name: formatPaymentMethodName(paymentMethod),
+        customerId: stripeCustomerId,
+        CollectiveId: stripeCustomer.CollectiveId,
+        service: PAYMENT_METHOD_SERVICE.STRIPE,
+        type: paymentMethod.type,
+        confirmedAt: new Date(),
+        saved: true,
+        data: {
+          paymentMethodId: paymentMethod.id,
```
kewitz, line 488:
```suggestion
          stripePaymentMethodId: stripePaymentMethod.id,
```

server/graphql/v2/mutation/OrderMutations.js
hdiniz: Neither. This stripe call will always create a new user. But we are only creating a user once per stripe host account, if the user was created before we have it stored in the payment method.

server/graphql/v2/mutation/OrderMutations.js
hdiniz: For payments done through the stripe payment element we dont know what payment method was attached to it until after confirmation. In this scenario we still need some generic payment method, at least until we received the webhook. this is also saving the stripe customer id for the host account. wdyt?

server/paymentProviders/stripe/index.js
hdiniz: Agreed. Everything in this provider will be using the same stripe interface, having separate implementations per method would be strange at least. 

I hope we can merge it all at some point and delegate to this provider anything with a PaymentMethod.service == STRIPE. We would still want to store the PaymentMethod.type though, as that is relevant to know if the payment method supports off_session charges, to display to customers saved methods, etc.

server/graphql/v2/mutation/OrderMutations.js
kewitz: In that case, we should update this existing payment method when the information is attached.

server/graphql/v2/mutation/OrderMutations.js
kewitz: Not really. `stripe.customers.create` is being called outside the conditional `if (!pm)`.

hdiniz: Ops, fixed that!

server/graphql/v2/mutation/OrderMutations.js
hdiniz: Done, we attach a proper payment method when the payment intent is processed.

server/graphql/v2/mutation/OrderMutations.js
hdiniz: I refactored to store the stripe customer id as a connected account instead of a dummy payment method.

server/paymentProviders/stripe/payment-intent.ts
```diff
@@ -5,12 +5,18 @@ import type Stripe from 'stripe';
...
-import { APPLICATION_FEE_INCOMPATIBLE_CURRENCIES, refundTransaction, refundTransactionOnlyInDatabase } from './common';
+import {
+  APPLICATION_FEE_INCOMPATIBLE_CURRENCIES,
+  createChargeTransactions,
+  refundTransaction,
+  refundTransactionOnlyInDatabase,
+} from './common';
 
-const processOrder = async (order: typeof models.Order): Promise<void> => {
+async function processNewOrder(order: typeof models.Order) {
```
kewitz, line 19: Interesting, why hoisting?
Important to guarantee this returns undefined. Otherwise, it will break the expected behavior in:
https://github.com/opencollective/opencollective-api/blob/main/server/lib/payments.js#L577-L578

server/paymentProviders/stripe/payment-intent.ts
```diff
@@ -41,11 +47,94 @@ const processOrder = async (order: typeof models.Order): Promise<void> => {
...
+      logger.error('Unknown error with Stripe Payment Intent.');
+      logger.error(paymentIntent);
+      reportMessageToSentry('Unknown error with Stripe Payment Intent', { extra: { paymentIntent } });
+      throw new Error('Something went wrong with the payment, please contact support@opencollective.com.');
+    }
+
+    // Recently, Stripe updated their library and removed the 'charges' property in favor of 'latest_charge',
+    // but this is something that only makes sense in the LatestApiVersion, and that's not the one we're using.
+    const charge = (paymentIntent as any).charges.data[0] as Stripe.Charge;
+    return createChargeTransactions(charge, { order });
```
kewitz, line 117: I don't like this very much:

1. I don't think this will ever be the case since this charge is async.
2. We shouldn't duplicate logic, this is already being handled by the paymentIntentSucceeded webhook.

server/paymentProviders/stripe/payment-intent.ts
```diff
@@ -41,11 +47,94 @@ const processOrder = async (order: typeof models.Order): Promise<void> => {
...
+  } catch (e) {
+    const sanitizedError = pick(e, ['code', 'message', 'requestId', 'statusCode']);
+    const errorMessage = `Error processing Stripe Payment Intent: ${e.message}`;
+    logger.error(errorMessage, sanitizedError);
+    // Hard-delete the order so we can re-use the existing Payment Intent in the next attempt.
+    throw new Error(errorMessage);
+  }
+}
+
+const processOrder = async (order: typeof models.Order): Promise<void> => {
```
kewitz, line 127: The return type is misleading since `processRecurringOrder` can return a transaction.
We need to make sure this doesn't return anything, the existing logic is counting on this processor not creating any transactions:
https://github.com/opencollective/opencollective-api/blob/main/server/lib/payments.js#L577-L578

server/paymentProviders/stripe/webhook.ts
```diff
@@ -47,11 +53,15 @@ const paymentIntentSucceeded = async (event: Stripe.Response<Stripe.Event>) => {
   const transaction = await createChargeTransactions(charge, { order });
 
   await order.update({
-    status: OrderStatuses.PAID,
+    status: !order.SubscriptionId ? OrderStatuses.PAID : undefined,
```
kewitz, line 56: Shouldn't we set this back to `ACTIVE`?

server/paymentProviders/stripe/webhook.ts
```diff
@@ -47,11 +53,15 @@ const paymentIntentSucceeded = async (event: Stripe.Response<Stripe.Event>) => {
...
     processedAt: new Date(),
     data: { ...order.data, paymentIntent },
   });
 
+  if (order.interval && !order.SubscriptionId) {
+    await createSubscription(order);
+  }
```
kewitz, line 63: This should be in the `processRecurringOrder` logic.

kewitz (commented):
I don't think we should bring the subscription complexity into this PR.

server/paymentProviders/stripe/payment-intent.ts
hdiniz: What do you mean by hoisting?

server/paymentProviders/stripe/webhook.ts
hdiniz: The processRecurringOrder is only called as part of the recurring-contributions cron job, it creates and confirms a payment intent off session. 

Therefore, setting up the subscription should come before it, doing it on this webhook ensures us that the first payment was successful and there is a reasonable expectation for the next charges to go through off session.

hdiniz:
> I don't think we should bring the subscription complexity into this PR.

let me know if you think we should split it. IMO feels like there is enough overlap that thinking about those cases together brings some clarity to the whole.

server/paymentProviders/stripe/payment-intent.ts
kewitz: Declaring `processOrder` as a function and not as constant arrow function.

kewitz:
@hdiniz I don't have much context on the subscriptions automation, if there's any bug, we're now automating it.

For example, the simple detail about returning the transaction in `payment-intent.ts:processOrder` we picked in this review would already have caused us to double-count credit the donation to the collective.

If we merge with subscriptions, then we also need to write an extensive test suite for it. Otherwise, we may have something similar that will take a full charge cycle for us to find out. This is why I recommend merging the reusable Payment Methods and leaving the subscription for a separate PR.

server/paymentProviders/stripe/payment-intent.ts
hdiniz: whats the difference?

hdiniz:
> @hdiniz I don't have much context on the subscriptions automation, if there's any bug, we're now automating it.
> 
> For example, the simple detail about returning the transaction in `payment-intent.ts:processOrder` we picked in this review would already have caused us to double-count credit the donation to the collective.
> 
> If we merge with subscriptions, then we also need to write an extensive test suite for it. Otherwise, we may have something similar that will take a full charge cycle for us to find out. This is why I recommend merging the reusable Payment Methods and leaving the subscription for a separate PR.

Agree on splitting it to review more carefully, but note that that would not happen as the flow you mentioned is not called for recurring transactions

server/paymentProviders/stripe/payment-intent.ts
kewitz: LOL, that's what I'm questioning... 😂
I just wanted to know what was the rationale behind changing how we declare this function.
I'm fine with both.

kewitz (approved):

server/paymentProviders/stripe/payment-intent.ts
hdiniz: In this context, I think the difference for me its just a style choice. 😛  Thanks for pointing hoisting though, I didnt know that behavior in javascript!

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6165

PR #8212, opened 2022-11-24
OAuth: skip authorization email if active userToken

znarf:


server/lib/oauth/model.ts
```diff
@@ -159,16 +159,22 @@ const model: OauthModel = {
...
-      CollectiveId: user.CollectiveId,
-      data: {
-        application: application.publicInfo,
-        collective: collective.minimal,
-        scope,
-      },
+    // Only send the email if there is no active user token right now
+    const userToken = await models.UserToken.findOne({
+      where: { ApplicationId: application.id, UserId: user.id },
     });
```
Betree, line 165: A `model.exists` would be nice for those cases (https://github.com/sequelize/sequelize/issues/10187)

Betree (approved):
Looks good. We'll probably want to send another kind of email when the permissions change

znarf:
@Betree asking for new scopes definitely deserve another email.

znarf:
@Betree asking for extra review as I implemented the scope check.

Betree (approved):
:+1:

PR #8214, opened 2022-11-24
Add tagSearchOperator to Accounts collection query

gustavlrsn:
Fix for https://github.com/opencollective/opencollective/issues/6170

Adding a `tagSearchOperator` field to the accounts collection query, that defaults to `AND` which is the current behavior.

```graphql
accounts(tags: ["climate justice", "climate change"], tagSearchOperator: OR)
```

### DIFF
--- a/server/graphql/v2/query/collection/AccountsCollectionQuery.ts
+++ b/server/graphql/v2/query/collection/AccountsCollectionQuery.ts
@@ -82,7 +82,7 @@ const AccountsCollectionQuery = {
       // Bind arguments
       if (args.tag?.length) {
         if (args.tagSearchOperator === 'OR') {
-          where['tags'] = { [Op.in]: args.tag };
+          where['tags'] = { [Op.overlap]: args.tag };
         } else {
           where['tags'] = { [Op.contains]: args.tag };
         }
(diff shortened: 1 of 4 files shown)


znarf (approved):
Looks good!

server/graphql/v2/query/collection/AccountsCollectionQuery.ts
```diff
@@ -21,6 +22,11 @@ const AccountsCollectionQuery = {
       type: new GraphQLList(GraphQLString),
       description: 'Only accounts that match these tags',
     },
+    tagSearchOperator: {
+      type: TagSearchOperator,
```
Betree, line 26: `null` wouldn't be a valid value, and we have a default. We can safely say:

```suggestion
      type: new GraphQLNonNull(TagSearchOperator),
```

server/graphql/v2/query/collection/AccountsCollectionQuery.ts
```diff
@@ -75,7 +81,11 @@ const AccountsCollectionQuery = {
 
       // Bind arguments
       if (args.tag?.length) {
-        where['tags'] = { [Op.contains]: args.tag };
+        if (args.tagSearchOperator === 'OR') {
+          where['tags'] = { [Op.in]: args.tag };
```
Betree, line 85: I'm surprised, `Op.in` is a way to check that a value is included in an array, but as far as I'm aware it's not meant to do an overlap between two arrays:

```sql
SELECT 'test' IN ('test', 'other') # true
```

```sql
SELECT ('test', 'test2') IN ('test', 'other') # ERROR
```

Unless I'm missing something with how Sequelize interprets this, I think we want to do something like:

```suggestion
          where['tags'] = { [Op.overlap]: args.tag };
```

gustavlrsn: Oooh, good catch! I guess I only ran the other query while testing this.

PR #8217, opened 2022-11-25
Fix PayPal adaptive processor fees

Betree:


server/graphql/common/expenses.ts
```diff
@@ -1475,7 +1476,23 @@ async function payExpenseWithPayPalAdaptive(remoteUser, expense, host, paymentMe
...
 
     // Warning senderFees can be null
-    const senderFees = createPaymentResponse.defaultFundingPlan.senderFees;
+    let senderFees = createPaymentResponse.defaultFundingPlan.senderFees;
+    if (!senderFees) {
+      // PayPal stopped providing senderFees in the response, we need to compute it ourselves
+      // We don't have to check for feesPayer here because it is not supported for PayPal adaptive
+      const { fundingAmount } = createPaymentResponse.defaultFundingPlan;
+      const amountPaidByTheHost = floatAmountToCents(parseFloat(fundingAmount.amount));
+      const amountReceivedByPayee = expense.amount;
```
znarf, line 1485: No multi-currency issues here?

znarf: Because adaptive is now disabled for multi-currency cases?

znarf (approved):
Good for me!

scripts/paypal/fix-adaptive-processor-fee.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  for (const debitTransaction of transactions) {
+    const host = await debitTransaction.getHostCollective();
+    const collective = await debitTransaction.getCollective();
+    const expense = await debitTransaction.getExpense();
+    const { defaultFundingPlan } = debitTransaction.data.createPaymentResponse;
+
+    // First approach was to fetch the transaction details from PayPal, but upon further verification
+    // all amounts (paymentInfo.receiver.amount) are equal to the gross expense amount, so we can't use that
```
Betree, line 56: typo here

```suggestion
    // all amounts (paymentInfo.receiver.amount) are equal to the gross expense amount, so we can use that
```

server/graphql/common/expenses.ts
Betree: Exactly, we can take this shortcut since multi-currency is disabled for adaptive. The check below (`if (fundingAmount.code !== expense.currency)`) with a Sentry notification is a way to be notified if, somehow, this rule ever gets broken.

Betree:
Script ran successfully, deploying the fix to prod

PR #8225, opened 2022-11-28
test: Refactor test DB setup to use migrations

hdiniz:
Extracted from https://github.com/opencollective/opencollective-api/pull/8218

Running migrations in the test environment allow us replicate the production DB in tests using the schema and database objects created outside sequelize (e.g views, indexes).

- Use `npm run db:restore:test` to setup your test environment. This restores the archived db and executes pending migrations.
- Use `NODE_ENV=test npm run db:migrate` and similar commands if you need to manage migrations during tests. (e.g testing a new migration before submitting).

When pulling new code, remember to `npm run db:restore:test` or `NODE_ENV=test npm run db:migrate` to keep your local test DB up to date.

- In your tests, the method `resetTestDB` will truncate the test database tables, restoring it to a clean state without recreating the schema.


znarf (approved):
Good for me!

znarf:
@Betree can you have a look?

test/utils.js
```diff
@@ -45,78 +45,17 @@ export const data = path => {
...
 export const resetTestDB = async () => {
-  await sequelize.sync({ force: true }).catch(e => {
-    console.error("test/utils.js> Sequelize Error: Couldn't recreate the schema", e);
-    process.exit(1);
-  });
-  // That could be an alternative but this doesn't work
-  // await sequelize.truncate({ force: true, cascade: true });
-};
+  const resetFn = async () => {
+    await sequelize.truncate({ cascade: true, force: true, restartIdentity: true });
```
Betree, line 49: Cool trick

test/server/models/models-histories.test.js
```diff
@@ -22,7 +22,7 @@ const getHistoryTableName = tableName => {
 
 const tablesWithHistory = ['Collectives', 'Comments', 'Expenses', 'Subscriptions', 'Updates'];
 
-describe('server/models/models-histories', () => {
```
Betree, line 25: @hdiniz Why skip this one?

Betree (commented):
Awesome work, that's a much more solid foundation for the future.

Small question below about a skipped test, the rest looks great!

test/server/models/models-histories.test.js
```diff
@@ -22,7 +22,7 @@ const getHistoryTableName = tableName => {
...
+describe.skip('server/models/models-histories', () => {
```
Betree, line 25: @hdiniz Why skip this one?

hdiniz: I missed/forgot this one, checking now the issue with this was that the history tables dont have the same schema as actual tables, they were not migrated together.

For example, in prod:

`Collectives` has a column `searchTsVector`
`CollectiveHistories` does not have `searchTsVector`

There are other errors like this.

So this test will fail, the only reason it was passing is because we were creating the schema in tests each time based on sequelize definitions.

I think we can skip this for now and create an issue to fix these cases where the history tables are not matching. wdyt?

znarf: @hdiniz could you create a follow up issue?

PR #8232, opened 2022-11-29
Enable recurring contributions for ACH/SEPA

hdiniz:
Resolves https://github.com/opencollective/opencollective/issues/6196

Enables recurring contributions for payment intent based payment methods (currently only enabled for ACH and SEPA bank debits).

### DIFF
--- a/server/paymentProviders/stripe/payment-intent.ts
+++ b/server/paymentProviders/stripe/payment-intent.ts
@@ -105,7 +105,6 @@ async function processRecurringOrder(order: typeof models.Order) {
       return;
     } else if (paymentIntent.status !== 'succeeded') {
       logger.error('Unknown error with Stripe Payment Intent.');
-      logger.error(paymentIntent);
       reportMessageToSentry('Unknown error with Stripe Payment Intent', { extra: { paymentIntent } });
       throw new Error('Something went wrong with the payment, please contact support@opencollective.com.');
     }
--- a/server/paymentProviders/stripe/payment-intent.ts
+++ b/server/paymentProviders/stripe/payment-intent.ts
@@ -5,12 +5,21 @@ import type Stripe from 'stripe';
 import OrderStatuses from '../../constants/order_status';
 import logger from '../../lib/logger';
 import { getApplicationFee } from '../../lib/payments';
+import { reportMessageToSentry } from '../../lib/sentry';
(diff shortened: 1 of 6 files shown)


server/paymentProviders/stripe/payment-intent.ts
```diff
@@ -46,11 +55,75 @@ const processOrder = async (order: typeof models.Order): Promise<void> => {
...
+    chargeRetryCount: order.Subscription.chargeRetryCount,
+  };
+
+  try {
+    let paymentIntent = await stripe.paymentIntents.create(paymentIntentParams, {
+      stripeAccount: hostStripeAccount.username,
+    });
+
+    order.update({ data: { ...order.data, paymentIntent: { id: paymentIntent.id, status: paymentIntent.status } } });
+    await order.save();
```
kewitz, line 97: Do we really need to update order here right before confirming?

Also, `order.update` returns a Promise of the update query, you don't need to call save after it.
```suggestion
    await order.update({ data: { ...order.data, paymentIntent: { id: paymentIntent.id, status: paymentIntent.status } } });
```

server/paymentProviders/stripe/payment-intent.ts
```diff
@@ -46,11 +55,75 @@ const processOrder = async (order: typeof models.Order): Promise<void> => {
     await order.destroy({ force: true });
     throw new Error(errorMessage);
   }
-};
+}
+
+async function processRecurringOrder(order: typeof models.Order) {
```
kewitz, line 60: It would be nice to write a test suite for this function.

server/paymentProviders/stripe/payment-intent.ts
```diff
@@ -46,11 +55,73 @@ const processOrder = async (order: typeof models.Order): Promise<void> => {
...
```
kewitz, line 60: I feel like there's a huge overlap between the two process functions.
Perhaps we should consolidate both logic in a single `processOrder` function that has conditional calls depending on if it is a subscription or not.

Not something we need to tackle right now, just a tech debt really.

kewitz: Since this is a generic `PaymentIntent` function, perhaps we should also check if the payment method we're using in this order supports or not off-cycle charge.

kewitz (approved):
Looks good. We need to make sure to test this in production.

PR #8243, opened 2022-12-01
Make sure accountants also see the memo field

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6205

image


znarf (approved):
Should work!

PR #8245, opened 2022-12-01
Convert User model to Typescript

Betree:
Require https://github.com/opencollective/opencollective-api/pull/8241

This PR could impact some bugs as we were not passing the `transaction` to sequelize correctly in different places (which Typescript detected).


server/graphql/common/user.ts
```diff
@@ -36,9 +37,9 @@ export const createUser = (
     location: Record<string, unknown>;
   },
   { organizationData, sendSignInLink, throwIfExists, redirect, websiteUrl, creationRequest }: CreateUserOptions,
-): Promise<{ user: typeof models.User; organization?: typeof models.Collective }> => {
+): Promise<{ user: User; organization?: typeof models.Collective }> => {
   return sequelize.transaction(async transaction => {
-    let user = await models.User.findOne({ where: { email: userData.email.toLowerCase() } }, { transaction });
```
hdiniz, line 41: :heart: typescript

hdiniz (approved):

server/graphql/common/user.ts
Betree: These small errors will be easily prevented!

PR #8248, opened 2022-12-01
Refact: TransferWise token handler

kewitz:
My main objective here is to reduce the amount of `invalid_token` errors we receive on our side through automatically retrying with a renewed token.

### DIFF
--- a/server/lib/transferwise.ts
+++ b/server/lib/transferwise.ts
@@ -136,7 +136,7 @@ export const requestDataAndThrowParsedError = async (
     return getData(response);
   } catch (e: any) {
     const signatureFailed = e?.response?.headers?.['x-2fa-approval-result'] === 'REJECTED';
-    const hadSignature = e?.response?.headers?.['X-Signature'];
+    const hadSignature = options?.headers?.['X-Signature'];
     if (signatureFailed && !hadSignature) {
       const ott = e.response.headers['x-2fa-approval'];
       const signature = signString(ott);
--- a/server/lib/transferwise.ts
+++ b/server/lib/transferwise.ts
@@ -4,13 +4,15 @@ import fs from 'fs';
 import path from 'path';
 import url from 'url';
 
-import Axios, { AxiosError } from 'axios';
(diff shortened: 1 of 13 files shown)


server/lib/transferwise.ts
```diff
@@ -90,55 +90,104 @@ const parseError = (
...
-  fn: Function,
+export async function getToken(connectedAccount: ConnectedAccount, refresh = false): Promise<string> {
+  // OAuth token, require us to refresh every 12 hours
+  const tokenCreation = moment.utc(connectedAccount.data.created_at);
+  const diff = moment.duration(moment.utc().diff(tokenCreation)).asSeconds();
+  const isOutdated = diff > <number>connectedAccount.data.expires_in - 60;
+  if (refresh || isOutdated) {
+    const newToken = await getOrRefreshToken({ refreshToken: connectedAccount.refreshToken });
+    if (!newToken) {
+      throw new Error('There was an error refreshing the Transferwise token');
```
hdiniz, line 101: Would this require reconnecting the wise account? If so, we could send an email on this error

server/lib/transferwise.ts
```diff
@@ -90,55 +90,104 @@ const parseError = (
...
+    )}`,
+  );
+
+  try {
+    const pRequest = data ? fn(url, data, options) : fn(url, options);
+    const response = await pRequest;
+    return getData(response);
+  } catch (e: any) {
+    const signatureFailed = e?.response?.headers?.['x-2fa-approval-result'] === 'REJECTED';
+    const hadSignature = e?.response?.headers?.['X-Signature'];
```
hdiniz, line 152:
```suggestion
    const hadSignature = options?.headers?.['X-Signature'];
```

looking at what we sent in the request to decide to recurse or not.

server/lib/transferwise.ts
kewitz: That's a very good idea.

kewitz: I'll start by tracking the activity to make sure that it wouldn't spam host admins.

PR #8255, opened 2022-12-04
Convert Comment model to typescript

Betree:
Part of https://github.com/opencollective/opencollective/issues/5893


server/models/Comment.ts
```diff
@@ -0,0 +1,198 @@
...
+      },
+      afterUpdate: async (comment, options) => {
+        if (comment.ConversationId) {
+          const transaction = options.transaction;
+          const conversation = await models.Conversation.findOne({ where: { RootCommentId: comment.id }, transaction });
+          if (conversation) {
+            await conversation.update({ summary: comment.html }, { transaction });
+          }
+        }
+      },
```
Betree, line 191: This is a real change compared to what we did before the migration: we're taking advantage of hooks rather than overriding the `destroy` method from sequelize (see https://github.com/opencollective/opencollective-api/pull/8255/files#diff-02b678ec46cee1ba66b4178ee4e435d7c0fd333d8245ba808c98e9cf86d19b37L150)

hdiniz (approved):

server/graphql/common/scope-check.ts
```diff
@@ -3,7 +3,7 @@ import express from 'express';
 import FEATURE from '../../constants/feature';
```
marineengineer175, line 3: I'm

server/graphql/common/scope-check.ts
```diff
@@ -3,7 +3,7 @@ import express from 'express';
...
 import OAuthScopes from '../../constants/oauth-scopes';
 import { canUseFeature } from '../../lib/user-permissions';
-import models from '../../models';
```
marineengineer175, line 6: [7](url)

marineengineer175 (commented):
Good

PR #8257, opened 2022-12-05
Add combined time series on account

gustavlrsn:
This PR is adding a combined time series on `account.stats`, so that we can get all data that is interesting for the Horizons dashboard in one database query per collective. 

Beyond allowing for big improvements in performance, it's also adding the currently not existing ability of getting number of contributors in a timespan or timeseries, required for the dashboard.

```graphql
query {
  accounts {
    nodes {
      id
      stats {
        combinedTimeSeries(dateFrom: DateTime, dateTo: DateTime, timeUnit: TimeUnit) {
          nodes {
            date
            contributions
            contributors
            totalNetRaised {
              value
              currency
            }
            totalSpent {
              value
              currency
            }
          }
        }
      }
    }
  }
}
```

### TODO
- [ ] discuss/improve API design
- [ ] review what transactions should be included to calculate totalNetRaised and totalSpent
- [x] fix currency conversion if there can be two host currencies on one collective

### DIFF
--- a/server/graphql/v2/object/AccountStats.js
+++ b/server/graphql/v2/object/AccountStats.js
@@ -1,7 +1,7 @@
 import { GraphQLBoolean, GraphQLInt, GraphQLList, GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
 import { GraphQLDateTime } from 'graphql-scalars';
 import { GraphQLJSON } from 'graphql-type-json';
-import { get, has, isNil } from 'lodash';
+import { get, has, initial, isNil } from 'lodash';
 import moment from 'moment';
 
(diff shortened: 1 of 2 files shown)


Betree:
@gustavlrsn Can you provide more info on how this `combinedTimeSeries` is going to be fetched? A link to the code would be ideal, but if not implemented yet, my main question is wether this new field is fetched in a list or if it's something fetched in a modal that's only displayed on click?

gustavlrsn:
@Betree closing this one in favor of https://github.com/opencollective/opencollective-api/pull/8261, which is building on the already existing patterns, and adding dataloaders to avoid `n+1` queries, since this data will be fetched in a list

PR #8258, opened 2022-12-05
Throw generic error when order fails fraud protection

kewitz:
- [x] Throw discrete, identifiable errors
- [x] Review and tweak parameters


server/graphql/v1/mutations/orders.js
```diff
@@ -246,7 +246,9 @@ export async function createOrder(order, req) {
   }
 
   await checkOrdersLimit(order, reqIp, reqMask);
-  await orderFraudProtection(req, order);
+  await orderFraudProtection(req, order).catch(() => {
+    throw new ValidationFailed('Could not process Order', 'FP');
+  });
```
znarf, line 251: Is it expected to be displayed like this on the frontend?

znarf: Not sure about the second parameter?

kewitz: Any suggestions on what we should display in the frontend?

znarf: What do we want to see here:
- For fraudsters trying to abuse the platform?
- For legit users in a false positive situation?

How can we have a message that make sense in both situations?

kewitz: IMO We should be straightforward so fraudsters can cease using the platform and legit users can contact us as soon as possible.

My doubt here is, do you think being straightforward to fraudsters is a good strategy?

znarf: No, because you don't want to give them hints on the specifics (credit card or ip based limit, what is exactly the limit).

Something close to "'Something went wrong with the payment, please contact support@opencollective.com.'" would be good. But not this one, so we can clearly recognize fraud protection was triggered.

We'll still have occasional fraudsters contacting us on support and we'll have a way to recognize them, so we can avoid answering and engaging.

PR #8266, opened 2022-12-07
Add flag for root users

Betree:
Mitigation for https://github.com/opencollective/opencollective/issues/4628
After discussing the topic in https://opencollective.slack.com/archives/GRN3WA3BM/p1670410019547039
Frontend PR: https://github.com/opencollective/opencollective-frontend/pull/8438

This will also limit the risk of privilege escalation issues in case our member invitation system ever get compromised.


server/models/User.ts
```diff
@@ -182,10 +182,9 @@ class User extends Model<InferAttributes<User>, InferCreationAttributes<User>> {
     }
   };
 
-  isRoot = function () {
-    const result = this.hasRole([roles.ADMIN], 1) || this.hasRole([roles.ADMIN], 8686);
-    debug('isRoot?', result);
-    return result;
+  isRoot = function (): boolean {
+    const hasAdminRole = this.hasRole([roles.ADMIN], 1) || this.hasRole([roles.ADMIN], 8686);
```
Betree, line 186: I'm tempted to also move these `1` and `8686` IDs in `config.accountIds.root` to not mess up between dev and prod configs.

server/models/User.ts
```diff
@@ -182,10 +182,12 @@ class User extends Model<InferAttributes<User>, InferCreationAttributes<User>> {
...
-  isRoot = function () {
-    const result = this.hasRole([roles.ADMIN], 1) || this.hasRole([roles.ADMIN], 8686);
-    debug('isRoot?', result);
-    return result;
+  isAdminOfOpenCollectiveInc = function (): boolean {
+    return this.hasRole([roles.ADMIN], 1) || this.hasRole([roles.ADMIN], 8686);
+  };
+
+  isRoot = function (): boolean {
+    return Boolean(this.isAdminOfOpenCollectiveInc() && this.data?.isRoot);
```
kewitz, line 190: I like the double check.

kewitz (approved):

PR #8277, opened 2022-12-09
Handle payment attach event for stripe cards

hdiniz:
Stripe cards are attached directly when the user chooses to save a card during the contribution flow, or when the user setups a card in the settings.


server/paymentProviders/stripe/webhook.ts
```diff
@@ -507,6 +507,11 @@ function formatPaymentMethodName(paymentMethod: Stripe.PaymentMethod) {
 async function paymentMethodAttached(event: Stripe.Response<Stripe.Event>) {
   const stripePaymentMethod = event.data.object as Stripe.PaymentMethod;
 
+  if (stripePaymentMethod.type === 'card') {
```
kewitz, line 510: The safe bet is ignoring everything that is not expected.

kewitz (approved):

PR #8281, opened 2022-12-09
Remove mailing list feature

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5874


Betree:
@znarf Ready for review, I think we're ready to finally remove this old feature

znarf (approved):
Looks good! Like it!

PR #8284, opened 2022-12-10
Fix refunded amount for Expenses with payment processor fee

znarf:
Fix https://github.com/opencollective/opencollective/issues/6230

### DIFF
--- a/migrations/20221212085251-fix-expense-refunds-with-payment-processor-cover.js
+++ b/migrations/20221212085251-fix-expense-refunds-with-payment-processor-cover.js
@@ -1,0 +1,56 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`WITH "ExpenseRefundsCreditsTransactionsUpdate" AS (
+      SELECT tr.id as "id",
+        - (t."amount" ) as "amount",
+        - (t."amountInHostCurrency") as "amountInHostCurrency",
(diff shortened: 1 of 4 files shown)


migrations/20221212085251-fix-expense-refunds-with-payment-processor-cover.js
```diff
@@ -0,0 +1,55 @@
...
+        "amountInHostCurrency" = erdtu."amountInHostCurrency",
+        "netAmountInCollectiveCurrency" = erdtu."netAmountInCollectiveCurrency"
+      FROM "ExpenseRefundsDebitsTransactionsUpdate" erdtu
+      WHERE "Transactions"."id" = erdtu."id"`);
+  },
+
+  async down() {
+    // No rollback
+  },
+};
```
kewitz, line 55: The migration seems correct, I only noticed we have a few transactions with an unusually high payment processor fee, for example, 1744368, 3431872, 2906770.
What is this about?

kewitz (approved):
Nice catch. 🎯 
A close review of the test makes it obvious but at the time I was fixing another issue that lead us to this place.
This test was probably written based on the output of the fixed function.

migrations/20221212085251-fix-expense-refunds-with-payment-processor-cover.js
znarf: I guess people messed up when "Marking as paid" and entered the expense amount instead of the payment processor fees.

Screenshot 2022-12-12 at 15 39 56

PR #8286, opened 2022-12-12
Remove separateHostFees configuration

znarf:
was introduced as a feature flag but the feature is now core functionality for a while


znarf:
Fyi @Betree

Betree (commented):
Great!

PR #8294, opened 2022-12-13
Add open telemetry to development environment

hdiniz:
OpenTelemetry gives us visibility on how the application is executing queries and might gives us insight on performance improvements.

This is only enabling it on development environment where we can test new features against snapshots, for example. In production, this same type of data is available on Sentry performance.

It requires running a container application as collector and visualization is available on http://localhost:16686.

image


package.json
```diff
@@ -44,6 +44,18 @@
...
+    "@opentelemetry/auto-instrumentations-node": "0.35.0",
+    "@opentelemetry/core": "1.8.0",
+    "@opentelemetry/exporter-trace-otlp-proto": "0.34.0",
+    "@opentelemetry/instrumentation": "0.34.0",
+    "@opentelemetry/instrumentation-express": "0.32.0",
+    "@opentelemetry/instrumentation-graphql": "0.33.0",
+    "@opentelemetry/instrumentation-http": "0.34.0",
+    "@opentelemetry/resources": "1.8.0",
+    "@opentelemetry/sdk-trace-base": "1.8.0",
+    "@opentelemetry/sdk-trace-node": "1.8.0",
```
znarf, line 57: Could this be devDependencies ?

hdiniz: A lot of these were not needed as direct dependencies lol. I moved the ones actually needed to dev dependencies.

Betree:
Looks like a great tool to have locally! 

It's unclear how related some of the changes are, would be great to add a line to explain it.

znarf:
Nice change for the dependencies.

Betree (approved):
Tested and approved! This will be super useful when trying to optimize performance locally.

PR #8297, opened 2022-12-13
Add Updates collection query

gustavlrsn:
Adding a `updates` root query, that returns a collection with updates, with optional arguments to filter on host and account tags:

```graphql
query {
  updates(
    host: { slug: "foundation" }
    tag: ["climate", "climate change"]
    limit: 50
  ) {
    totalCount
    limit
    nodes {
      id
      title
      createdAt
    }
  }
}
```

### DIFF
--- a/server/graphql/v2/query/collection/UpdatesCollectionQuery.ts
+++ b/server/graphql/v2/query/collection/UpdatesCollectionQuery.ts
@@ -1,0 +1,47 @@
+import { GraphQLList, GraphQLNonNull, GraphQLString } from 'graphql';
+
+import models, { Op } from '../../../../models';
+import { UpdatesCollection } from '../../collection/UpdatesCollection';
+import { AccountReferenceInput, fetchAccountsIdsWithReference } from '../../input/AccountReferenceInput';
+import { CollectionArgs, CollectionReturnType } from '../../interface/Collection';
+
+const UpdatesCollectionQuery = {
+  type: new GraphQLNonNull(UpdatesCollection),
+  args: {
+    ...CollectionArgs,
+    tag: {
+      type: new GraphQLList(GraphQLString),
+      description: 'Only from accounts that have one of these tags',
+    },
+    host: {
+      type: new GraphQLList(AccountReferenceInput),
+      description: 'Host for the account for which to get updates',
+    },
+  },
+  async resolve(_: void, args): Promise<CollectionReturnType> {
+    const { offset, limit } = args;
+    let include;
+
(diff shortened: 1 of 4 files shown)


server/graphql/v2/query/collection/UpdatesCollectionQuery.ts
```diff
@@ -0,0 +1,47 @@
...
+      if (args.host) {
+        const hostCollectiveIds = await fetchAccountsIdsWithReference(args.host);
+        include.where = { ...include.where, HostCollectiveId: hostCollectiveIds };
+      }
+      if (args.tag) {
+        include.where = { ...include.where, tags: { [Op.overlap]: args.tag } };
+      }
+    }
+
+    const order = [['createdAt', 'DESC']];
```
znarf, line 41: Maybe `publishedAt` is a better one.

server/graphql/v2/query/collection/UpdatesCollectionQuery.ts
```diff
@@ -0,0 +1,47 @@
...
+      };
+      if (args.host) {
+        const hostCollectiveIds = await fetchAccountsIdsWithReference(args.host);
+        include.where = { ...include.where, HostCollectiveId: hostCollectiveIds };
+      }
+      if (args.tag) {
+        include.where = { ...include.where, tags: { [Op.overlap]: args.tag } };
+      }
+    }
+
```
znarf, line 40: Any condition that we need to add all the time? Making sure we're not exposing private or unpublished updates. How are we doing this in the collective updates resolver?

server/graphql/v2/query/collection/UpdatesCollectionQuery.ts
gustavlrsn: Ah, yes, fixed.

server/graphql/v2/query/collection/UpdatesCollectionQuery.ts
gustavlrsn: @znarf Ah, thanks for catching this. This would have been bad. Added a condition to only get published updates!

Also added an `isPrivate: false` condition, even though the privacy applies to the html content only and is handled by the update resolver, it seems reasonable to not include private updates in this query. 

The collective updates resolver has an `onlyPublishedUpdates` (default: false) argument so that you as an admin can see the unpublished ones. But I'm thinking we don't need that in this query right now.

znarf (approved):
Looks good!

PR #8299, opened 2022-12-13
Faster amountReceived and amountSpent calculations

znarf:
Attempt at porting what we've done for fast balances to compute `amountReceived` and `amountSpent`.

TODO:
- [x] helper (equivalent to getCurrentFastBalances)
- [x] loader (equivalent to context.loaders.Collective.currentCollectiveBalance)
- [x] inclusion in getSumCollectivesAmountSpent 
- [x] inclusion in getSumCollectivesAmountReceived

For later:
- [ ] can this also support getContributionsAndContributorsCount ? certainly

### DIFF
--- a/server/lib/budget.js
+++ b/server/lib/budget.js
@@ -95,7 +95,7 @@ export async function getBalances(
 ) {
   const fastResults =
     fastBalance === true && version === DEFAULT_BUDGET_VERSION && !endDate && !includeChildren
-      ? await getCurrentFastBalances(collectiveIds, { loaders, withBlockedFunds })
+      ? await getCurrentCollectiveBalances(collectiveIds, { loaders, withBlockedFunds })
       : {};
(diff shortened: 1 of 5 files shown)


server/lib/budget.js
```diff
@@ -204,7 +226,7 @@ export function getSumCollectivesAmountSpent(
     : 'amountInHostCurrency';
   const transactionType = 'DEBIT_WITHOUT_HOST_FEE';
 
-  return sumCollectivesTransactions(collectiveIds, {
+  const results = await sumCollectivesTransactions(collectiveIds, {
```
hdiniz, line 229: Can we run the fallback query only for the missing collective ids?
```suggestion
  const results = await sumCollectivesTransactions(missingCollectiveIds, {
```

znarf: Good catch!

PR #8307, opened 2022-12-15
Make sure to use correct paypal `captureId`

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6234

### DIFF
--- a/server/paymentProviders/paypal/subscription.ts
+++ b/server/paymentProviders/paypal/subscription.ts
@@ -15,7 +15,7 @@ import User from '../../models/User';
 import { PaymentProviderService } from '../types';
 
 import { paypalRequest } from './api';
-import {getCaptureIdFromPaypalTransaction, refundPaypalCapture} from './payment';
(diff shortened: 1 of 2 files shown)


server/paymentProviders/paypal/subscription.ts
```diff
@@ -344,8 +344,7 @@ const PayPalSubscription: PaymentProviderService = {
   },
 
   async refundTransaction(transaction, user, reason) {
-    // Subscription's transactions are always recorded with paypalSale
-    const captureId = transaction.data.paypalSale?.id;
+    const captureId = getCaptureIdFromPaypalTransaction(transaction);
```
SudharakaP, line 347: @znarf: Here I am trying to re-use our existing function that takes care of the `captureId` as it could be seen in the issue, https://github.com/opencollective/opencollective/issues/6234 the `captureId` isn't necessarily in `paypalSale` object.

znarf:
I will let @Betree review this.

Betree (approved):
Good fix :+1:

PR #8311, opened 2022-12-16
Expense: Mark as Paid with custom FX rate

kewitz:


kewitz:
Closed in favor of https://github.com/opencollective/opencollective-api/pull/8339

PR #8315, opened 2022-12-19
Add personal token 

flickz:
Resolves https://github.com/opencollective/opencollective/issues/5736

- Add personal token model
- Add data migration file to migrate application with API_KEY into PersonalToken table
- Add Personal token graphQL query and mutations.


server/graphql/v2/interface/Account.js
```diff
@@ -370,6 +371,29 @@ const accountFieldsDefinition = () => ({
       return { nodes: result.rows, totalCount: result.count, limit, offset };
     },
   },
+  personalTokens: {
```
znarf, line 374: By definition personalTokens are only on `Individual`s not all `Account`s.

migrations/20221211114041-create-personal-tokens-table.js
```diff
@@ -0,0 +1,84 @@
...
+      },
+      updatedAt: {
+        type: Sequelize.DATE,
+        defaultValue: Sequelize.NOW,
+      },
+      deletedAt: {
+        type: Sequelize.DATE,
+      },
+    });
+    await queryInterface.addIndex('PersonalTokens', ['token'], { unique: true });
```
znarf, line 78: Nice!

server/graphql/v2/query/index.js
```diff
@@ -51,6 +52,7 @@ const query = {
   update: UpdateQuery,
   updates: UpdatesCollectionQuery,
   paypalPlan: PaypalPlanQuery,
+  personalToken: PersonalTokenQuery,
```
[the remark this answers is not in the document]
flickz: It is used to query a single token on the token settings page.

server/graphql/v2/mutation/PersonalTokenMutations.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    const createParams = {
+      ...pick(args.personalToken, ['name', 'scope', 'expiresAt']),
+      CollectiveId: collective.id,
+      UserId: req.remoteUser.id,
+      token: models.PersonalToken.generateToken(),
+    };
+
+    return await models.PersonalToken.create(createParams);
```
Betree, line 52: :ice_cream: 

```suggestion
    return models.PersonalToken.create(createParams);
```

server/graphql/v2/mutation/PersonalTokenMutations.ts
```diff
@@ -0,0 +1,111 @@
...
+    });
+
+    if (!personalToken) {
+      throw new NotFound(`Personal token not found`);
+    } else if (!req.remoteUser.isAdminOfCollective(personalToken.collective)) {
+      throw new Forbidden('Authenticated user is not the token owner.');
+    }
+
+    const updateParams = pick(args.personalToken, ['name', 'scope', 'expiresAt']);
+    return await personalToken.update(updateParams);
```
Betree, line 77: :ice_cream: 

```suggestion
    return personalToken.update(updateParams);
```

server/graphql/v2/input/PersonalTokenCreateInput.ts
```diff
@@ -0,0 +1,19 @@
...
+  name: 'PersonalTokenCreateInput',
+  description: 'Input type for PersonalToken',
+  fields: () => ({
+    name: { type: GraphQLString },
+    scope: { type: new GraphQLList(OAuthScope) },
+    expiresAt: { type: GraphQLString },
+    account: {
+      type: AccountReferenceInput,
+      description: 'The account to use as the owner of the application. Defaults to currently logged in user.',
+    },
```
Betree, line 17: Given that personal tokens can only be created on personal (USER) accounts - at the difference of API keys - I think we should remove this argument altogether and consider that personal tokens are always referring to the currently logged-in user. @znarf wdyt?

server/graphql/v2/object/PersonalToken.ts
```diff
@@ -0,0 +1,50 @@
+import { GraphQLList, GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
+
+import { DateString } from '../../v1/types';
```
Betree, line 3: :no_entry_sign: Never this one on GQLV2. We can instead use:

```suggestion
import { GraphQLDateTime } from 'graphql-scalars';
```

server/models/PersonalToken.ts
```diff
@@ -0,0 +1,92 @@
+import { randomBytes } from 'crypto';
+
+import { CreationOptional, InferAttributes, InferCreationAttributes, NonAttribute } from 'sequelize';
+
+import oAuthScopes from '../constants/oauth-scopes';
+import sequelize, { DataTypes, Model } from '../lib/sequelize';
+
+import models from '.';
+
+class PersonalToken extends Model<InferAttributes<PersonalToken>, InferCreationAttributes<PersonalToken>> {
```
Betree, line 10: Great job for introducing this one with Typescript directly :100:

server/models/PersonalToken.ts
```diff
@@ -0,0 +1,92 @@
...
+  public declare readonly id: CreationOptional<number>;
+  public declare token: string;
+  public declare expiresAt: Date;
+  public declare createdAt: CreationOptional<Date>;
+  public declare updatedAt: CreationOptional<Date>;
+  public declare deletedAt: CreationOptional<Date>;
+  public declare data: Record<string, unknown>;
+  public declare CollectiveId: number;
+  public declare UserId: number;
```
Betree, line 19: :ice_cream: No need to iterate in this PR, but since `User` is now converted to TS we can do:

```suggestion
  public declare UserId: ForeignKey<User['id']>;
```

server/models/PersonalToken.ts
```diff
@@ -0,0 +1,92 @@
...
+  public declare scope: string[];
```
Betree, line 20: :ice_cream: Could have been interesting to type this of based on `oAuthScopes`

server/models/index.js
```diff
@@ -0,0 +1,399 @@
+import sequelize, { Op } from '../lib/sequelize';
```
Betree, line 1: :no_entry_sign: It looks like this file is the result of a merge conflict, only `server/models/index.ts` should be there

Betree (commented):
Looking great! Test coverage is on point. Some feedback below, and one question for @znarf.

Feel free to ignore all the :ice_cream:, we can ever let them go or I'll iterate before merging.

server/graphql/v2/input/PersonalTokenCreateInput.ts
znarf: Agree, that can make things simpler.

But for API consistency, maybe it's better to pass the Account. Let's say that a root admin wants to create a token for someone else, maybe the API design should allow this.

Not saying this particularly desirable here, but wondering if it's a good practice to guess an Account from the loggedInUser at any time.

Betree: > Not saying this particularly desirable here, but wondering if it's a good practice to guess an Account from the loggedInUser at any time.

I think it's ok for things that are scoped for the logged-in user not to require you to pass any account. In all cases, we're already doing it here since we default to the logged-in user if nothing was passed for `account`.

Betree (approved):

PR #8317, opened 2022-12-19
Personal token based authentication

znarf:


server/lib/hyperwatch.js
```diff
@@ -80,10 +80,6 @@ const load = async app => {
           }
         }
 
-        if (req.clientApp) {
-          log = log.setIn(['opencollective', 'application', 'id'], req.clientApp.id);
-        }
```
Betree, line 85: Not sure why we're removing that?

Betree (approved):

server/lib/hyperwatch.js
Betree: Maybe

```es6
          log = log.setIn(['opencollective', 'personalToken', 'id'], req.personalToken.id);
```

Betree (approved):
Tested locally, works like a charm

PR #8319, opened 2022-12-20
Add loader to yearly budget

hdiniz:
Fix: https://github.com/opencollective/opencollective/issues/3182

### DIFF
--- a/migrations/20221220120527-add-transaction-subscription-indexes.js
+++ b/migrations/20221220120527-add-transaction-subscription-indexes.js
@@ -1,0 +1,14 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.addIndex('Transactions', ['CollectiveId', 'createdAt', 'type']);
+    await queryInterface.addIndex('Subscriptions', ['isActive']);
+  },
+
+  async down(queryInterface) {
+    await queryInterface.removeIndex('Transactions', ['CollectiveId', 'createdAt', 'type']);
+    await queryInterface.removeIndex('Subscriptions', ['isActive']);
+  },
+};
--- a/migrations/20221220120527-add-transaction-subscription-indexes.js
+++ b/migrations/20221220120527-add-transaction-subscription-indexes.js
@@ -4,11 +4,9 @@
 module.exports = {
   async up(queryInterface) {
     await queryInterface.addIndex('Transactions', ['CollectiveId', 'createdAt', 'type']);
-    await queryInterface.addIndex('Subscriptions', ['isActive']);
   },
 
   async down(queryInterface) {
     await queryInterface.removeIndex('Transactions', ['CollectiveId', 'createdAt', 'type']);
-    await queryInterface.removeIndex('Subscriptions', ['isActive']);
   },
 };
--- a/server/lib/budget.js
+++ b/server/lib/budget.js
@@ -691,110 +691,111 @@ export async function sumCollectivesTransactions(
   return totals;
 }
 
-export async function getYearlyIncome(collectiveIds) {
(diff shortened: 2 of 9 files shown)


znarf:
Related: https://github.com/opencollective/opencollective/issues/6195

server/graphql/v1/TransactionInterface.js
```diff
@@ -236,85 +236,64 @@ const TransactionFields = () => {
...
-          }
+
+        const fromCollective = await req.loaders.Collective.byId.load(transaction.FromCollectiveId);
+        if (fromCollective.HostCollectiveId) {
+          return req.loaders.Collective.byId.load(fromCollective.HostCollectiveId);
+        }
+
+        const collective = await req.loaders.Collective.byId.load(transaction.CollectiveId);
+        if (collective.HostCollectiveId) {
+          return req.loaders.Collective.byId.load(fromCollective.HostCollectiveId);
```
znarf, line 251: Looks unrelated, different PR?

server/lib/budget.js
```diff
@@ -695,59 +695,109 @@ export async function sumCollectivesTransactions(
...
+              SUM(coalesce(t."hostFeeInHostCurrency", 0)) +
+              SUM(coalesce(t."paymentProcessorFeeInHostCurrency", 0)) +
+              SUM(coalesce(t."taxAmount" * t."hostCurrencyFxRate", 0)) "amount"
           FROM "Transactions" t
-          LEFT JOIN "Orders" d ON t."OrderId" = d.id
-          LEFT JOIN "Subscriptions" s ON d."SubscriptionId" = s.id
-          WHERE t."CollectiveId" = :CollectiveId
-          AND t."RefundTransactionId" IS NULL
+          LEFT JOIN "Orders" o ON t."OrderId" = o.id
+          LEFT JOIN "Subscriptions" s ON s.id = o."SubscriptionId"
```
znarf, line 732: If we want to fetch one time contributions, no need to join Subscriptions.

If we do, it's different than what is implied by "last_twelve_months_one_time_contributions".

hdiniz: I think i picked a bad alias for that query. The original one does join with subscriptions, it sums up past recurring contributions that are not active anymore.

migrations/20221220120527-add-transaction-subscription-indexes.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.addIndex('Transactions', ['CollectiveId', 'createdAt', 'type']);
```
Betree, line 6: Should we remove the existing `CollectiveId-type` index in favor of this one?

znarf: @hdiniz ?

hdiniz: `CollectiveId-Type` is one of the most used indexes on the transactions table, I would look at that on a separate issue to see the query impact. 

This new index might not be compatible with `"CollectiveId-type" btree ("CollectiveId", type)` as it will be ordered by the createdAt before type.

```
select indexrelname, idx_scan, pg_size_pretty(pg_relation_size(indexrelid)) as "Index Size" from pg_stat_all_indexes where schemaname = 'public' and relname = 'Transactions' order by idx_scan desc;
                   indexrelname                   | idx_scan  | Index Size
--------------------------------------------------+-----------+------------
 CollectiveId-FromCollectiveId-type               | 751364426 | 61 MB
 transactions__using_gift_card_from_collective_id | 654061807 | 42 MB
 CollectiveId-type                                | 471101054 | 50 MB
```

znarf: Let's merge like this then.

PR #8320, opened 2022-12-20
Use loader when `net` is true for total amount spent

gustavlrsn:
Allow using the loader for total amount spent when getting total net amount spent, which is what we want to use for the Discover dashboard.

### DIFF
--- a/server/lib/budget.js
+++ b/server/lib/budget.js
@@ -171,7 +171,7 @@ export async function getTotalAmountSpentAmount(
     includeGiftCards,
   };
 
-  if (loaders && version === DEFAULT_BUDGET_VERSION && !net) {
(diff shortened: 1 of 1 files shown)


server/lib/budget.js
```diff
@@ -171,7 +171,7 @@ export async function getTotalAmountSpentAmount(
     includeGiftCards,
   };
 
-  if (loaders && version === DEFAULT_BUDGET_VERSION && !net) {
+  if (loaders && version === DEFAULT_BUDGET_VERSION) {
```
znarf, line 174: Good catch. Thank you!

znarf (approved):
Works for me. Thank you for the catch.

PR #8321, opened 2022-12-20
Add account type argument to Updates collection query

gustavlrsn:
For the Discover dashboard, we need the ability to only get updates from a certain type of accounts (collectives & funds). 

This PR adds an `accountType` argument to the Updates collection query. It also renames the `tag` argument to `accountTag`, to be more clear that they relate to the account and not the update itself, without deprecating the old field since the query was merged only 5 days ago in https://github.com/opencollective/opencollective-api/pull/8297

```graphql
query {
  updates(
    host: [{ slug: "foundation" }]
    limit: 50
    accountTag: ["climate"]
    accountType: [COLLECTIVE, FUND]
  ) {
    totalCount
    nodes {
      title
    }
  }
}
```

### DIFF
--- a/server/graphql/v2/query/collection/UpdatesCollectionQuery.ts
+++ b/server/graphql/v2/query/collection/UpdatesCollectionQuery.ts
@@ -2,6 +2,7 @@ import { GraphQLList, GraphQLNonNull, GraphQLString } from 'graphql';
 
 import models, { Op } from '../../../../models';
 import { UpdatesCollection } from '../../collection/UpdatesCollection';
+import { AccountType, AccountTypeToModelMapping } from '../../enum';
 import { AccountReferenceInput, fetchAccountsIdsWithReference } from '../../input/AccountReferenceInput';
 import { CollectionArgs, CollectionReturnType } from '../../interface/Collection';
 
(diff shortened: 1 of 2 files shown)


server/graphql/v2/query/collection/UpdatesCollectionQuery.ts
```diff
@@ -11,7 +12,12 @@ const UpdatesCollectionQuery = {
     ...CollectionArgs,
     tag: {
       type: new GraphQLList(GraphQLString),
-      description: 'Only from accounts that have one of these tags',
+      description: 'Only return updates from accounts that have one of these tags',
+    },
+    type: {
```
gustavlrsn, line 17: Considering if this rather should be called `accountType`. And the `tag` argument called `accountTag`. Brevity vs clarity?

znarf: Maybe. Because from the context, we are on searching for updates and we may be implying that we're filtering by update type and update tag, which are plausible things.

znarf: I would be good with that change.

gustavlrsn: Renamed these arguments (without deprecating since this new query has only been live for a few days).

znarf (approved):

PR #8327, opened 2022-12-21
Use loaders in transaction collections

hdiniz:


server/graphql/v1/TransactionInterface.js
```diff
@@ -236,85 +236,64 @@ const TransactionFields = () => {
     },
     host: {
       type: UserCollectiveType,
-      async resolve(transaction) {
-        if (transaction && transaction.getHostCollective) {
```
Betree, line 240: This check was not the most effective :upside_down_face:

server/graphql/v1/TransactionInterface.js
```diff
@@ -236,85 +236,64 @@ const TransactionFields = () => {
     },
     host: {
       type: UserCollectiveType,
```
Betree, line 238: Not from this PR, but taking note that this doesn't seem to be the best type. We probably won't fix it since it's GraphQL V1.

server/graphql/v1/TransactionInterface.js
```diff
@@ -236,85 +236,64 @@ const TransactionFields = () => {
...
-          return transaction.getCreatedByUser();
+        if (!transaction.CreatedByUserId) {
+          return;
         }
-        return null;
+
+        const [collective, fromCollective] = await Promise.all([
+          req.loaders.Collective.byId.load(transaction.CollectiveId),
+          req.loaders.Collective.byId.load(transaction.FromCollectiveId),
+        ]);
```
Betree, line 265: :ice_cream: 

```suggestion
        const [collective, fromCollective] = await req.loaders.Collective.byId.loadMany([
          transaction.CollectiveId,
          transaction.FromCollectiveId,
        ]);
```

server/graphql/v1/TransactionInterface.js
```diff
@@ -236,85 +236,64 @@ const TransactionFields = () => {
...
+
+        return req.loaders.User.byId.load(transaction.CreatedByUserId);
       },
     },
     fromCollective: {
       type: CollectiveInterfaceType,
-      resolve(transaction) {
-        // If it's a sequelize model transaction, it means it has the method getFromCollective
-        // otherwise we check whether transaction has 'fromCollective.id', if not we return null
-        if (transaction && transaction.getFromCollective) {
```
Betree, line 283: Now that I see all these *transaction.methodFromTransactionModel*, I wonder if we weren't somehow returning something else (like a raw object) in one of the queries.

Edit: looks like that's the case, as explained here: https://github.com/opencollective/opencollective-api/pull/1578/files#r240981003. But we got rid of this in https://github.com/opencollective/opencollective-api/pull/2384, so these checks are definitely not needed anymore!

server/graphql/v1/types.js
```diff
@@ -1600,8 +1600,11 @@ export const OrderType = new GraphQLObjectType({
       createdByUser: {
         type: UserType,
         async resolve(order, args, req) {
-          const collective = await req.loaders.Collective.byId.load(order.CollectiveId);
-          const fromCollective = await req.loaders.Collective.byId.load(order.FromCollectiveId);
+          const [collective, fromCollective] = await Promise.all([
+            req.loaders.Collective.byId.load(order.CollectiveId),
+            req.loaders.Collective.byId.load(order.FromCollectiveId),
+          ]);
```
Betree, line 1606: :ice_cream: 
```suggestion

          const [collective, fromCollective] = await req.loaders.Collective.byId.loadMany([
            order.CollectiveId,
            order.FromCollectiveId,
          ]);
```

Betree (approved):
All good for me! Some code style suggestions below, but everything looks good. This will make a huge diff for performances!

PR #8328, opened 2022-12-21
Add sans-serif font fallbacks to emails

Betree:
Part of https://github.com/opencollective/opencollective/issues/5812


znarf (approved):
Good change.

PR #8338, opened 2022-12-23
Add support to get the host currency fx rate

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/6240

Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/8457

### DIFF
--- a/server/graphql/v2/object/Order.js
+++ b/server/graphql/v2/object/Order.js
@@ -1,22 +1,31 @@
-import { GraphQLBoolean, GraphQLInt, GraphQLList, GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
+import {
+  GraphQLBoolean,
+  GraphQLFloat,
+  GraphQLInt,
+  GraphQLList,
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLString,
+} from 'graphql';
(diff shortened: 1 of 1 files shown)


server/graphql/v2/object/Order.js
```diff
@@ -235,6 +244,20 @@ export const Order = new GraphQLObjectType({
...
       },
+      hostCurrencyFxRate: {
+        type: GraphQLFloat,
+        description: 'The FX rate used to convert the amount to the host currency',
+        async resolve(order, _, { loaders }) {
+          const collective = order.collective || (await loaders.Collective.byId.load(order.CollectiveId));
+          const hostCollectiveId = collective?.HostCollectiveId;
+          if (hostCollectiveId) {
+            const host = await loaders.Collective.byId.load(hostCollectiveId);
+            return await getFxRate(order.currency, host.currency);
```
znarf, line 255: Seems we now have a loader for fx rates, search for `loaders.CurrencyExchangeRate`.

SudharakaP: Thanks. I've changed this to use `loaders.CurrencyExchangeRate`.

Betree:
Closing as we've dropped the frontend PR: https://github.com/opencollective/opencollective-frontend/pull/8457. We may re-open later depending on how we want to fix this.

PR #8339, opened 2022-12-23
PayExpenseModal: Host admin to inform amounts in host currency when manually paying an Expense

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6238
Relates to https://github.com/opencollective/opencollective/issues/5601#issuecomment-1359186212

### DIFF
--- a/server/lib/transactions.ts
+++ b/server/lib/transactions.ts
@@ -226,23 +226,30 @@ export async function createTransactionsForManuallyPaidExpense(
   assert(paymentProcessorFeeInHostCurrency >= 0, 'Payment processor fee must be positive');
   assert(totalAmountPaidInHostCurrency > 0, 'Total amount paid must be positive');
 
+  // Values are already adjusted to negative DEBIT values
   const isCoveredByPayee = expense.feesPayer === 'PAYEE';
-  const amount = toNegative(totalAmountPaidInHostCurrency - paymentProcessorFeeInHostCurrency);
+  const grossAmount = toNegative(totalAmountPaidInHostCurrency - paymentProcessorFeeInHostCurrency);
   const netAmountInCollectiveCurrency = toNegative(totalAmountPaidInHostCurrency);
   const amounts = {
-    amount,
-    amountInHostCurrency: amount,
+    amount: grossAmount,
+    amountInHostCurrency: grossAmount,
     paymentProcessorFeeInHostCurrency: toNegative(paymentProcessorFeeInHostCurrency),
     netAmountInCollectiveCurrency,
     hostCurrencyFxRate: 1,
   };
 
+  if (isCoveredByPayee) {
+    set(transactionData, 'feesPayer', 'PAYEE');
+    // Not necessary to adjust amounts since the host admin already passes the net amount as the base argument
+  }
+
+  // Adjust values if currency from host is different from the currency of the collective.
   if (host.currency !== expense.collective.currency) {
     assert(
       expense.currency === expense.collective.currency,
       'Expense currency must be the same as collective currency',
     );
-    amounts.hostCurrencyFxRate = expense.amount / amount;
+    amounts.hostCurrencyFxRate = expense.amount / grossAmount;
     amounts.amount = amounts.amount * amounts.hostCurrencyFxRate;
     amounts.netAmountInCollectiveCurrency = amounts.netAmountInCollectiveCurrency * amounts.hostCurrencyFxRate;
   }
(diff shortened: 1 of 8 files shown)


server/lib/transactions.ts
```diff
@@ -213,6 +215,84 @@ export async function createTransactionsFromPaidExpense(
...
+
+  if (host.currency !== expense.collective.currency) {
+    assert(
+      expense.currency === expense.collective.currency,
+      'Expense currency must be the same as collective currency',
+    );
+    amounts.hostCurrencyFxRate = isCoveredByPayee
+      ? // If the payee is covering the fees, the expense amount should be proportional to the total amount paid in host currency
+        expense.amount / totalAmountPaidInHostCurrency
+      : // If the host is covering the fees, the expense amount be proportional to total amount paid minus the fees
```
Betree, line 248: I realize I made a confusing comment when we discussed this on Slack:
- When paying an expense, unless `feesPayer=PAYEE`, the fee is paid by the collective. 
   _Example: A collective with $200 balance, expense is $80, fees are $20 => The collective balance will be $100 after paying._
- When refunding an expense (aka. "Mark as unpaid") the host is indeed covering the fees (via a `PAYMENT_PROCESSOR_COVER` transaction). The intent is to make sure the full amount is refunded to the collective.
   _Example: Host has a $1000 balance. Collective with $200 balance, expense is $80, fees are $20 => The collective balance will be $100 after paying. The collective balance will be back to $200 if the expense is refunded, but the host's will be $1000 - $20 = $980_

```suggestion
      : // If the collective is covering the fees, the expense amount should be proportional to total amount paid minus the fees
```

server/lib/transactions.ts
```diff
@@ -213,6 +215,84 @@ export async function createTransactionsFromPaidExpense(
...
+  paymentProcessorFeeInHostCurrency,
+  totalAmountPaidInHostCurrency,
+  /** Will be stored in transaction.data */
+  transactionData: Record<string, unknown> = {},
+) {
+  assert(paymentProcessorFeeInHostCurrency >= 0, 'Payment processor fee must be positive');
+  assert(totalAmountPaidInHostCurrency >= 0, 'Total amount paid must be positive');
+
+  const isCoveredByPayee = expense.feesPayer === 'PAYEE';
+  const amount = toNegative(totalAmountPaidInHostCurrency - paymentProcessorFeeInHostCurrency);
```
Betree, line 230: As per the comment above (fees are paid by the collective) I believe this should be:

```suggestion
  const amount = toNegative(totalAmountPaidInHostCurrency + paymentProcessorFeeInHostCurrency);
```

server/lib/transactions.ts
Betree, line 248: Note to myself for a second review: aren't these two reversed? I would expect the fees to be ignored in the DEBIT's amount if feesPayer is the PAYEE, and accounted if feesPayer=COLLECTIVE.

Betree (commented):
The code looks much cleaner, this `createTransactionsForManuallyPaidExpense` refactor is a great move. I believe some amounts calculations need to be iterated on (see the comments below) and I'd very much like to make sure we use the same base amounts for the tests to make it as easy as possible to review what changed exactly. 

I'll have a second review before merging, we can also have a call to paid on this if you prefer!

server/lib/transactions.ts
kewitz: Since from now on the host admin is informing the total amount paid (this means the net amount that left their bank account, including fees), here we're just separating the real amount related to the expense from the fees.

The logic of who's paying for the fee comes later on.

server/lib/transactions.ts
kewitz: You're correct. But on top of that, there's just a misconception here.
This is not about the logic of who's paying for the fees, I was just a bit confused about what value to use when calculating the FX rate.
Turns out, the correct here is to always consider the `expense.amount / amount`.

server/lib/transactions.ts
Betree: > The logic of who's paying for the fee comes later on.

Where? From what I see, we're jumping straight from this to `Transaction.createDoubleEntry` without checking the `feesPayer`. Don't we need something like https://github.com/opencollective/opencollective-api/blob/3863342448edba0748ea95bd6cf8964c2092c005/server/lib/transactions.ts#L206 to record fees on the right account?

Betree (commented):
The question around `feesPayer` is the only thing still catching my attention; the rest looks good!

server/lib/transactions.ts
kewitz: I'm sorry @Betree, you're correct. I'll implement this conditional and re-submit for review.

PR #8341, opened 2022-12-23
Update Credit Card Upon Detection of Duplicate Credit Card

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6189

[payment-update.webm](https://user-images.githubusercontent.com/12435965/209355321-0258a656-5c81-41fc-9c5c-e4f8db27bf0e.webm)

### DIFF
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -5,6 +5,7 @@ import config from 'config';
 import debugLib from 'debug';
 import * as hcaptcha from 'hcaptcha';
 import { get, isEmpty, isEqual, isNil, omit, pick, set } from 'lodash';
+import { Op } from 'sequelize';
 
 import activities from '../../../constants/activities';
 import CAPTCHA_PROVIDERS from '../../../constants/captcha-providers';
@@ -568,7 +569,18 @@ export async function createOrder(order, req) {
         if (get(order, 'paymentMethod.service') === 'stripe') {
           // For Stripe `save` will be manually set to `true`, in `processOrder` if the order succeed
           order.paymentMethod.saved = null;
+        } else {
+          order.paymentMethod.saved = Boolean(orderCreated.data.savePaymentMethod);
+        }
 
+        if (isGuest) {
+          set(order.paymentMethod, 'data.isGuest', true);
+        }
+
+        await orderCreated.setPaymentMethod(order.paymentMethod);
+
+        // Update all existing orders with the new payment method
+        if (get(order, 'paymentMethod.service') === 'stripe') {
           // Check if the payment method is already saved
           const paymentMethod = await models.PaymentMethod.findOne({
             where: {
@@ -581,17 +593,18 @@ export async function createOrder(order, req) {
             },
           });
           if (paymentMethod && !order.paymentMethod?.uuid && order.paymentMethod?.save) {
+            await models.Order.update(
+              { PaymentMethodId: orderCreated.paymentMethod.id },
+              {
+                where: {
+                  PaymentMethodId: paymentMethod.id,
+                  status: { [Op.notIn]: ['PAID', 'CANCELLED', 'EXPIRED', 'DISPUTED', 'REFUNDED'] },
+                },
+              },
+            );
             await paymentMethod.destroy();
           }
-        } else {
-          order.paymentMethod.saved = Boolean(orderCreated.data.savePaymentMethod);
         }
-
-        if (isGuest) {
-          set(order.paymentMethod, 'data.isGuest', true);
-        }
-
-        await orderCreated.setPaymentMethod(order.paymentMethod);
       }
       // also adds the user as a BACKER of collective
       await libPayments.executeOrder(remoteUser, orderCreated);
--- a/server/graphql/v1/mutations/orders.js
+++ b/server/graphql/v1/mutations/orders.js
@@ -5,7 +5,6 @@ import config from 'config';
 import debugLib from 'debug';
 import * as hcaptcha from 'hcaptcha';
 import { get, isEmpty, isEqual, isNil, omit, pick, set } from 'lodash';
-import { Op } from 'sequelize';
 
 import activities from '../../../constants/activities';
 import CAPTCHA_PROVIDERS from '../../../constants/captcha-providers';
@@ -28,7 +27,7 @@ import { reportErrorToSentry } from '../../../lib/sentry';
 import twoFactorAuthLib from '../../../lib/two-factor-authentication';
 import { canUseFeature } from '../../../lib/user-permissions';
 import { formatCurrency, parseToBoolean } from '../../../lib/utils';
-import models, { sequelize } from '../../../models';
+import models, { Op, sequelize } from '../../../models';
 import {
   BadRequest,
   FeatureNotAllowedForUser,
@@ -37,7 +36,6 @@ import {
   Unauthorized,
   ValidationFailed,
 } from '../../errors';
-import { deletePaymentMethod, updateExistingOrdersToNewPaymentMethod } from '../../v2/mutation/PaymentMethodMutations';
 
 const debug = debugLib('orders');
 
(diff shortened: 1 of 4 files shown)


znarf:
Passing review and asking @Betree to jump in.

server/graphql/v1/mutations/orders.js
```diff
@@ -551,6 +551,21 @@ export async function createOrder(order, req) {
...
+              service: 'stripe',
+              saved: true,
+              data: {
+                fingerprint: order.paymentMethod.data.fingerprint,
+              },
+            },
+          });
+          if (paymentMethod && !order.paymentMethod?.uuid && order.paymentMethod?.save) {
+            await paymentMethod.destroy();
+          }
```
Betree, line 568: What will happen with orders that hold a reference to the old payment method (i.e. other active subscriptions) if we delete it?

SudharakaP: @Betree : Great point; something I've overlooked. I've just modified the code so that the existing orders (which are still active) are updated with the new payment method. Let me know how it looks; especially the order status I considered I am wondering if I got it precisely.

server/graphql/v2/mutation/PaymentMethodMutations.js
```diff
@@ -78,8 +80,35 @@ const addCreditCard = {
...
+    await models.Order.update(
+      { PaymentMethodId: pm.id },
+      {
+        where: {
+          PaymentMethodId: oldPaymentMethod.id,
+          status: { [Op.notIn]: ['PAID', 'CANCELLED', 'EXPIRED', 'DISPUTED', 'REFUNDED'] },
+        },
+      },
+    );
+
```
Betree, line 111: The order here is not correct:
- Delete old PM
- Create new PM
- Update orders

Creating the new PM can fail, and we would end up in a broken state where the old PM is deleted but the orders are not updated with a new one. Or concurrency issues: recurring contributions are being processed, I'm updating my payment method and the order still holds a reference to the deleted PM.

**Solution:** use a sequelize transaction.

Betree: And why not re-use the existing PM if it already exists, rather than deleting it? Like:


```es6
let paymentMethod = await findExistingCreditCard(fingerprint);
if (!paymentMethod || !checkPermissions(paymentMethod)) {
  paymentMethod = await models.PaymentMethod.create(newPaymentMethodData); 
}

**Review by Betree — CHANGES_REQUESTED** [review]
Blocking feedback about using a SQL transactions.

We also need to cover the `updateOrder` mutation.

There's also too much duplication in this code and it's not really clear while we're doing it while looking at the code. Let's introduce some shared helpers.

File: server/graphql/v2/mutation/PaymentMethodMutations.js (hunk 2 above)
Comment (SudharakaP) [reply server/graphql/v2/mutation/PaymentMethodMutations.js]: @Betree : Thanks. I've added the sequelize transactions, and added the shared helpers. 

> And why not re-use the existing PM if it already exists, rather than deleting it? Like:
> 
> ```js
> let paymentMethod = await findExistingCreditCard(fingerprint);
> if (!paymentMethod || !checkPermissions(paymentMethod)) {
>   paymentMethod = await models.PaymentMethod.create(newPaymentMethodData); 
> }
> ```

This one isn't possible I think since if we try to use the existing PM to update the card then Stripe will give an error such as, 

>The token tok_1MhSiXDjPFcHOcTmpnbDMKmh has already been used. Each token can only be used once to create a source.

because the token that is returned by Stripe should be different given we changed the card details.

File: server/graphql/v1/mutations/orders.js [hunk 3]
```diff
@@ -231,6 +231,45 @@ const hasPaymentMethod = order => {
...
+      where: {
+        CollectiveId: collective.id,
+        service: 'stripe',
+        saved: true,
+        data: {
+          fingerprint: fingerprint,
+        },
+      },
+    },
+    { transaction },
```
Betree, line 247: As per https://sequelize.org/api/v6/class/src/model.js~model#static-method-findOne, the transaction is expected in 1st object. Typescript would have detected this.

server/graphql/v1/mutations/orders.js
```diff
@@ -231,6 +231,45 @@ const hasPaymentMethod = order => {
...
+    // Update all existing orders with the new payment method
+    await models.Order.update(
+      { PaymentMethodId: paymentMethod.id },
+      {
+        where: {
+          PaymentMethodId: oldPaymentMethod.id,
+          status: { [Op.notIn]: ['PAID', 'CANCELLED', 'EXPIRED', 'DISPUTED', 'REFUNDED'] },
+        },
+      },
+      { transaction },
```
Betree, line 268: As per https://sequelize.org/api/v6/class/src/model.js~model#instance-method-update, the transaction is expected in 2nd object. Typescript would have detected this.

server/graphql/v1/mutations/orders.js
```diff
@@ -231,6 +231,45 @@ const hasPaymentMethod = order => {
...
+        data: {
+          fingerprint: fingerprint,
+        },
+      },
+    },
+    { transaction },
+  );
+
+  if (oldPaymentMethod) {
+    await oldPaymentMethod.destroy();
```
Betree, line 251: No transaction?

Betree (changes requested):
SQL transaction is not set properly.

Can we add a test that mocks `PaymentMethod.create` to make it fail, and makes sure the old payment method is not deleted and the order is not updated if that's the case?

server/graphql/v1/mutations/orders.js
SudharakaP: Done. 👍🏽

server/graphql/v1/mutations/orders.js
SudharakaP: Corrected. 👍🏽

SudharakaP:
> SQL transaction is not set properly.
> 
> Can we add a test that mocks `PaymentMethod.create` to make it fail, and makes sure the old payment method is not deleted and the order is not updated if that's the case?

@Betree : I've added some tests here to make sure that we are properly accounting for the updates. Let me know if there's any further suggestions. 😸

test/server/graphql/v2/mutation/PaymentMethodMutations.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      await utils.resetTestDB();
+    });
+
+    it('correctly creates a credit card payment method', async () => {
+      result.errors && console.error(result.errors);
+      expect(result.errors).to.not.exist;
+      expect(result.data.addCreditCard.paymentMethod.name).is.equal('1990');
+      expect(result.data.addCreditCard.paymentMethod.data.brand).is.equal('Visa');
+    });
+
```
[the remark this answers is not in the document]
SudharakaP: @Betree : Yes, I've added another test to check this case. 👍🏽

SudharakaP:
@Betree : Realized I made a small mistake in one of the tests. I've corrected that as well.

server/graphql/v1/mutations/orders.js
```diff
@@ -574,6 +611,20 @@ export async function createOrder(order, req) {
...
+        if (get(order, 'paymentMethod.service') === 'stripe' && order.paymentMethod.data?.fingerprint) {
+          // Check if the payment method is already saved
+          await sequelize.transaction(async transaction => {
+            const oldPaymentMethod = await deletePaymentMethod(
+              fromCollective,
+              order.paymentMethod.data.fingerprint,
+              transaction,
+            );
+
+            await updateExistingOrdersToNewPaymentMethod(orderCreated.paymentMethod, oldPaymentMethod, transaction);
```
Betree, line 625: :ice_cream: I would find it clearer to have the condition here. You're currently forced to go down into `deletePaymentMethod` + `updateExistingOrdersToNewPaymentMethod` to understand what's happening here.

```suggestion
            if (oldPaymentMethod) {
              await updateExistingOrdersToNewPaymentMethod(orderCreated.paymentMethod, oldPaymentMethod, transaction);
            }
```

server/graphql/v1/mutations/orders.js
```diff
@@ -231,6 +231,43 @@ const hasPaymentMethod = order => {
   }
 };
 
+export const deletePaymentMethod = async (collective, fingerprint, transaction) => {
+  // Check if the credit card is already saved
+  const oldPaymentMethod = await models.PaymentMethod.findOne({
+    where: {
+      CollectiveId: collective.id,
+      service: 'stripe',
+      saved: true,
```
Betree, line 240: Just a note that, if someone has a saved credit card and fills out the number for the same card while unchecking "Save for future payments" it will remove the already-saved card from payment methods.

That's arguably a good approach.

Betree (approved):
Tested, seems to be working great. Maybe ask for a last review from @znarf, but there's no blocker on my side.

PR #8344, opened 2022-12-23
Fix issue in sequelize/sql query construction

znarf:


gustavlrsn (approved):

znarf:
Thank you @gustavlrsn

PR #8350, opened 2022-12-27
Fix Stripe PaymentIntent handling

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5773

Should also resolve most of the occurrences for https://sentry.io/organizations/open-collective/issues/3058463744, all those with the title `GraphQLError: Internal Payment error (invalid PaymentIntent)`.

### DIFF
--- a/server/paymentProviders/stripe/creditcard.ts
+++ b/server/paymentProviders/stripe/creditcard.ts
@@ -82,15 +82,16 @@ const createChargeAndTransactions = async (
     });
   }
 
-  paymentIntent = await stripe.paymentIntents.confirm(paymentIntent.id, {
-    stripeAccount: hostStripeAccount.username,
-  });
+  paymentIntent = await stripe.paymentIntents.confirm(
+    paymentIntent.id,
+    { payment_method: stripePaymentMethod.id },
+    { stripeAccount: hostStripeAccount.username },
+  );
 
   /* eslint-enable camelcase */
 
   if (paymentIntent.next_action) {
-    order.data.paymentIntent = { id: paymentIntent.id, status: paymentIntent.status };
-    await order.update({ data: order.data });
+    await order.update({ data: { ...order.data, paymentIntent } }); // Store the payment intent to make sure it will be re-used after the 3D secure confirmation
     const paymentIntentError = new Error('Payment Intent require action');
     paymentIntentError['stripeAccount'] = hostStripeAccount.username;
     paymentIntentError['stripeResponse'] = { paymentIntent };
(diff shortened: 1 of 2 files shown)


server/paymentProviders/stripe/creditcard.ts
```diff
@@ -82,15 +82,16 @@ const createChargeAndTransactions = async (
...
   }
 
-  paymentIntent = await stripe.paymentIntents.confirm(paymentIntent.id, {
-    stripeAccount: hostStripeAccount.username,
-  });
+  paymentIntent = await stripe.paymentIntents.confirm(
+    paymentIntent.id,
+    { payment_method: stripePaymentMethod.id },
+    { stripeAccount: hostStripeAccount.username },
+  );
```
Betree, line 89: While working on this issue, I stumbled upon the fact that trying to re-use a payment intent after a failed verification led to this error (that has [occurrences](https://sentry.io/organizations/open-collective/issues/3058463744) in production):

> GraphQLError: Internal Payment error (invalid PaymentIntent)

The root message from Stripe was:

> You cannot confirm this PaymentIntent because it's missing a payment method. To confirm the PaymentIntent with cus_xxxxxxxxx, specify a payment method attached to this customer along with the customer ID.

Simply passing the `payment_method` here solved the issue and resulted in a successful payment, even after failing the 1st 3Dsecure verification.

server/paymentProviders/stripe/creditcard.ts
```diff
@@ -82,15 +82,16 @@ const createChargeAndTransactions = async (
...
 
   /* eslint-enable camelcase */
 
   if (paymentIntent.next_action) {
-    order.data.paymentIntent = { id: paymentIntent.id, status: paymentIntent.status };
-    await order.update({ data: order.data });
+    await order.update({ data: { ...order.data, paymentIntent } }); // Store the payment intent to make sure it will be re-used after the 3D secure confirmation
```
Betree, line 94: Here the order was not updated successfully. I believe it's an issue with sequelize doing a shallow comparison (rather than a deep comparison) on `data`; thus trying to update without copying the object doesn't have any effect.

A small snippet if you want to verify that assertion for yourself:

```es6
setTimeout(() => {
  models.Order.findByPk(6442).then(o => {
    o.data.test = 1;
    console.log('(1) Has changes?', o.changed());

    o.data = { ...o.data, test: 2 };
    console.log('(2) Has changes?', o.changed());
  });
}, 3000);
```

Result:

> (1) Has changes? false
> (2) Has changes? [ 'data' ]

Betree:
@znarf I'll wait for your feedback on this one when you come back since it's a tricky part of the code you have some experience with. 

Also cc @SudharakaP since we discussed this very precise error yesterday; it will surely resolve those tickets you've seen  where the person got an `Internal Payment error (invalid PaymentIntent)`.

znarf (approved):
Looks good to me.

SudharakaP (approved):
@Betree : Thanks a lot for fixing this one. 👍🏽

kewitz (approved):

PR #8368, opened 2023-01-04
Move issuing webhook events to use each account webhook secret

hdiniz:


znarf:
Looks good to me but it's better to wait for a review from @kewitz.

server/paymentProviders/stripe/webhook.ts
```diff
@@ -564,42 +566,58 @@ async function paymentMethodAttached(event: Stripe.Event) {
...
   try {
-    return stripe.webhooks.constructEvent(request.rawBody, request.headers['stripe-signature'], webhookSigningSecret);
-  } catch (e) {
+    event = stripeClient.webhooks.constructEvent(
+      request.rawBody,
+      request.headers['stripe-signature'],
+      webhookSigningSecret,
+    );
+  } catch {
     throw new Error('Source of event not recognized');
```
kewitz, line 603: It is fine to throw here because the person receiving this is probably trying to forge events and won't try it again.

server/paymentProviders/stripe/webhook.ts
```diff
@@ -614,22 +632,21 @@ export const webhook = async (request: Request<unknown, Stripe.Event>) => {
     return Promise.resolve();
   }
 
-  event = await validateWebhook(request);
+  if (event.type.startsWith('issuing')) {
+    return handleIssuingWebhooks(request);
```
kewitz, line 636: We may receive unexpected events in `handleIssuingWebhooks` that we may fail to parse due to an undefined virtualCardId.

server/paymentProviders/stripe/webhook.ts
```diff
@@ -614,22 +632,21 @@ export const webhook = async (request: Request<unknown, Stripe.Event>) => {
...
+  }
+
+  try {
+    event = stripe.webhooks.constructEvent(
+      request.rawBody,
+      request.headers['stripe-signature'],
+      config.stripe.webhookSigningSecret,
+    );
+  } catch (e) {
+    throw new Error('Source of event not recognized');
```
kewitz, line 646: Yes, we throw it here. Probably someone trying to forge events.

server/paymentProviders/stripe/webhook.ts
```diff
@@ -564,42 +566,58 @@ async function paymentMethodAttached(event: Stripe.Event) {
...
+    const transaction = <Stripe.Issuing.Transaction>event.data.object;
+    virtualCardId = get(transaction, 'transaction.card.id', transaction?.card);
+  } else if (event.type.startsWith('issuing_card')) {
+    virtualCardId = (<Stripe.Issuing.Card>event.data.object).id;
   }
 
-  if (!webhookSigningSecret) {
-    throw new Error('Stripe Webhook Signin Secret not set for Host');
+  if (!virtualCardId) {
+    throw new Error('virtual card id not set in webhook event');
```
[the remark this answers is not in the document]
kewitz: 🔴 This needs to be fixed. Just throw if we're failing to parse the virtualCardId from an expected event, otherwise just return.

kewitz (approved):

PR #8376, opened 2023-01-06
VirtualCards: 30 days grace for missing receipts

kewitz:
This behavior was inconsistent with the email we send and the policy on our website.


znarf:
Good stuff!

PR #8377, opened 2023-01-06
Cancel Pending Host Applications when Host is Deactivated

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6278

[cancel-hosting-applications.webm](https://user-images.githubusercontent.com/12435965/214723391-4dd6dff0-b8d6-4d3a-8121-b849c23ab158.webm)

### DIFF
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1268,7 +1268,7 @@ Collective.prototype.deactivateAsHost = async function () {
     { where: { HostCollectiveId: this.id } },
   );
 
-  await models.Member.update({ deletedAt: new Date() }, { where: { MemberCollectiveId: this.id, role: 'HOST' } });
+  await models.Member.destroy({ where: { MemberCollectiveId: this.id, role: 'HOST' } });
 
   await models.Collective.update(
     { HostCollectiveId: null },
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -1261,6 +1261,16 @@ Collective.prototype.deactivateAsHost = async function () {
     );
   }
 
+  const pendingHostApplications = await models.Collective.count({
+    where: { HostCollectiveId: this.id, type: types.COLLECTIVE, isActive: false },
+  });
+
+  if (pendingHostApplications >= 1) {
+    throw new Error(
+      `You can't deactivate hosting while still having ${pendingHostApplications} pending applications. Please contact support: support@opencollective.com.`,
+    );
+  }
+
   // TODO unsubscribe from OpenCollective tier plan.
 
   await this.deactivateBudget();
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -87,6 +87,7 @@ import userlib from '../lib/userlib';
 import { capitalize, formatCurrency, getDomain, md5 } from '../lib/utils';
 
 import CustomDataTypes from './DataTypes';
+import { HostApplicationStatus } from './HostApplication';
 import Order from './Order';
 import { PayoutMethodTypes } from './PayoutMethod';
 import { SocialLinkType } from './SocialLink';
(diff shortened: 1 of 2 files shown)


Betree (commented):
Should we rather resolve the root cause for https://github.com/opencollective/opencollective/issues/6278? It seems legit to me to disable a host with pending applications, they should just be cancelled

server/models/Collective.js
```diff
@@ -1207,6 +1208,19 @@ Collective.prototype.deactivateAsHost = async function () {
     );
   }
 
+  // Make sure we clean up all pending applications
+  await models.HostApplication.update(
+    { status: HostApplicationStatus.EXPIRED },
+    { where: { HostCollectiveId: this.id } },
+  );
+
+  await models.Member.update({ deletedAt: new Date() }, { where: { MemberCollectiveId: this.id, role: 'HOST' } });
```
Betree, line 1217: Do we really need that one? `HOST` member are not supposed to be created until the application gets approved

Betree:
The overall approach looks good!

server/models/Collective.js
SudharakaP: @Betree : I've double checked this and it seems that we do add an entry to the `Member` table upon applying for the host; 

https://github.com/opencollective/opencollective-api/blob/b41f19b6509d12289772f39b9c1b1b3bf6f64058/server/models/Collective.js#L2343-L2348

SudharakaP:
@Betree : Anything to do here? Okay to merge?

server/models/Collective.js
```diff
@@ -1258,6 +1259,19 @@ Collective.prototype.deactivateAsHost = async function () {
...
+
+  await models.Collective.update(
+    { HostCollectiveId: null },
+    { hooks: false, where: { HostCollectiveId: this.id, isActive: false } },
+  );
+
```
Betree, line 1274: Not needed, we do this already in deactivateBudget: https://github.com/opencollective/opencollective-api/blob/dc25778c2ae332f27891f632d7fd09fc0ba9f7c1/server/models/Collective.js#L1410

```suggestion
```

Betree (commented):
Can we add a small test in https://github.com/opencollective/opencollective-api/blob/dc25778c2ae332f27891f632d7fd09fc0ba9f7c1/test/server/models/Collective.test.js#L363, just to make sure we cover these new changes?

server/models/Collective.js
Betree: Interesting :+1:

SudharakaP:
> Can we add a small test in
> 
> https://github.com/opencollective/opencollective-api/blob/dc25778c2ae332f27891f632d7fd09fc0ba9f7c1/test/server/models/Collective.test.js#L363
> , just to make sure we cover these new changes?

@Betree : I've added a test for checking this case. 👍🏽

server/models/Collective.js
```diff
@@ -1261,6 +1262,19 @@ Collective.prototype.deactivateAsHost = async function () {
     );
   }
 
+  // Make sure we clean up all pending applications
+  await models.HostApplication.update(
+    { status: HostApplicationStatus.EXPIRED },
+    { where: { HostCollectiveId: this.id } },
+  );
+
+  await models.Member.update({ deletedAt: new Date() }, { where: { MemberCollectiveId: this.id, role: 'HOST' } });
```
Betree, line 1271: :ice_cream: Not a requested change, just a cleaner option IMO

```suggestion
  await models.Member.destroy({ where: { MemberCollectiveId: this.id, role: 'HOST' } });
```

test/server/models/Collective.test.js
```diff
@@ -369,6 +369,38 @@ describe('server/models/Collective', () => {
...
+    it('auto cancel pending host applications when host is deactivated', async () => {
+      const temporaryHost = await models.Collective.create({
+        name: 'Temporary Host',
+        slug: 'temporaryHost',
+        hostFeePercent: 0,
+        type: 'ORGANIZATION',
+        currency: 'EUR',
+        CreatedByUserId: user1.id,
+        isHostAccount: true,
+      });
```
Betree, line 381: :ice_cream: Not a requested change: You can use the `fake-data` library for cleaner/simpler tests. Ex. here:

```suggestion
      const temporaryHost = await fakeHot({ admin: user1 });
```

Betree (approved):
Lgtm

test/server/models/Collective.test.js
SudharakaP: Thanks. Done. 👍🏽

PR #8378, opened 2023-01-06
Wise: Fix Host/Expense currency inconsistency

kewitz:
Makes our transactions more consistent with the real value being accounted for on Wise.
We also enforce that expenses should always be paid from the waller matching the Host currency.

### DIFF
--- a/server/paymentProviders/transferwise/index.ts
+++ b/server/paymentProviders/transferwise/index.ts
@@ -102,7 +102,7 @@ async function quoteExpense(
 
   expense.collective = expense.collective || (await models.Collective.findByPk(expense.CollectiveId));
   expense.host = expense.host || (await expense.collective.getHostCollective());
-  const hasMultiCurrency = expense.currency !== expense.host.currency;
+  const hasMultiCurrency = expense.currency !== expense.collective.currency;
   const targetCurrency = payoutMethod.unfilteredData.currency as string;
   const quoteParams = {
     profileId: connectedAccount.data.id,
--- a/server/paymentProviders/transferwise/index.ts
+++ b/server/paymentProviders/transferwise/index.ts
@@ -1,3 +1,4 @@
+import assert from 'assert';
 import crypto from 'crypto';
 
 import { isMemberOfTheEuropeanUnion } from '@opencollective/taxes';
@@ -11,6 +12,7 @@ import activities from '../../constants/activities';
 import status from '../../constants/expense_status';
 import { TransferwiseError } from '../../graphql/errors';
 import cache from '../../lib/cache';
+import { getFxRate } from '../../lib/currency';
 import logger from '../../lib/logger';
 import { reportErrorToSentry } from '../../lib/sentry';
 import * as transferwise from '../../lib/transferwise';
@@ -59,10 +61,13 @@ async function getTemporaryQuote(
   payoutMethod: PayoutMethod,
   expense: Expense,
 ): Promise<QuoteV2> {
+  expense.collective = expense.collective || (await models.Collective.findByPk(expense.CollectiveId));
+  expense.host = expense.host || (await expense.collective.getHostCollective());
+  const rate = await getFxRate(expense.currency, expense.host.currency);
   return await transferwise.getTemporaryQuote(connectedAccount, {
-    sourceCurrency: expense.currency,
+    sourceCurrency: expense.host.currency,
     targetCurrency: <string>payoutMethod.unfilteredData.currency,
-    sourceAmount: expense.amount / 100,
+    sourceAmount: (expense.amount * rate) / 100,
   });
 }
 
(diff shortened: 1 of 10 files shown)


server/paymentProviders/transferwise/index.ts
```diff
@@ -59,10 +61,13 @@ async function getTemporaryQuote(
...
 ): Promise<QuoteV2> {
+  expense.collective = expense.collective || (await models.Collective.findByPk(expense.CollectiveId));
+  expense.host = expense.host || (await expense.collective.getHostCollective());
+  const rate = await getFxRate(expense.currency, expense.host.currency);
   return await transferwise.getTemporaryQuote(connectedAccount, {
-    sourceCurrency: expense.currency,
+    sourceCurrency: expense.host.currency,
     targetCurrency: <string>payoutMethod.unfilteredData.currency,
-    sourceAmount: expense.amount / 100,
+    sourceAmount: (expense.amount * rate) / 100,
```
Betree, line 70: To make what we're doing here more explicit and prevent any floating-point error.

```suggestion
    sourceAmount: centsAmountToFloat(expense.amount * rate),
```

An example of what we're trying to avoid here:
> \> (3*100.1)/100
> 3.0029999999999997

server/paymentProviders/transferwise/index.ts
```diff
@@ -96,31 +101,45 @@ async function quoteExpense(
...
+    if (targetCurrency !== expense.host.currency) {
+      const [exchangeRate] = await transferwise.getExchangeRates(
+        connectedAccount,
+        expense.host.currency,
+        targetCurrency,
+      );
+      assert(exchangeRate, `No exchange rate found for ${expense.host.currency} -> ${targetCurrency}`);
+      rate = exchangeRate.rate;
+    }
+    quoteParams['targetAmount'] = (expense.amount / 100) * rate;
```
Betree, line 142: Same here

```suggestion
    quoteParams['targetAmount'] = centsAmountToFloat(expense.amount * rate);
```

server/paymentProviders/transferwise/index.ts
```diff
@@ -96,31 +101,45 @@ async function quoteExpense(
...
+    assert(
+      expense.currency === payoutMethod.unfilteredData.currency,
+      'For multi-currency expenses, the payout currency must be the same as the expense currency',
+    );
     quoteParams['targetCurrency'] = expense.currency;
     quoteParams['targetAmount'] = expense.amount / 100;
+  } else if (expense.feesPayer === 'PAYEE') {
+    assert(
+      expense.host.currency === expense.currency,
+      'For expenses coverede by the payee, the host currency must be the same as the expense currency',
```
Betree, line 128:
```suggestion
      'For expenses covered by the payee, the host currency must be the same as the expense currency',
```

server/paymentProviders/transferwise/index.ts
```diff
@@ -96,31 +101,45 @@ async function quoteExpense(
...
-  // Adapt for multi-currency
   if (hasMultiCurrency) {
+    assert(
+      expense.collective.currency === expense.host.currency,
+      'For multi-currency expenses, the host currency must be the same as the collective currency',
+    );
+    assert(
+      expense.currency === payoutMethod.unfilteredData.currency,
+      'For multi-currency expenses, the payout currency must be the same as the expense currency',
+    );
```
Betree, line 122: :+1:

server/paymentProviders/transferwise/index.ts
Betree: I'm afraid "expenses covered by the payee" is not necessarily clear, I can suggest:

```suggestion
      'For expenses where fees are covered by the payee, the host currency must be the same as the expense currency',
```

server/paymentProviders/transferwise/index.ts
```diff
@@ -96,31 +101,45 @@ async function quoteExpense(
...
     quoteParams['targetCurrency'] = expense.currency;
     quoteParams['targetAmount'] = expense.amount / 100;
+  } else if (expense.feesPayer === 'PAYEE') {
+    assert(
```
Betree, line 126: I would add a comment there just to make things clear for peeps who don't have the context

```suggestion
    // Using "else if" because customizing the fee payer is not allowed for multi-currency expenses. See `getCanCustomizeFeesPayer`.
    assert(
```

server/paymentProviders/transferwise/index.ts
Betree, line 142: Can the rate fluctuate between the `getExchangeRates` and the actual expense quote?

server/paymentProviders/transferwise/webhook.ts
```diff
@@ -63,18 +64,19 @@ export async function handleTransferStateChange(event: TransferStateChangeEvent)
...
     }
 
-    // Get FX rate
-    const wiseFxRateInfo = getWiseFxRateInfoFromExpenseData(expense, expense.currency, expense.host?.currency);
-    if (!wiseFxRateInfo) {
-      logger.warn(`Could not retrieve the FX rate from Wise for expense #${expense.id}. Falling back to 'auto' mode.`);
-    }
+    const hostAmount =
+      (expense.data?.transfer as Transfer)?.sourceValue ||
+      (expense.data?.quot as QuoteV2)?.sourceAmount -
```
Betree, line 69: Typo?

```suggestion
      (expense.data?.quote as QuoteV2)?.sourceAmount -
```

Betree (commented):
Great work :rocket: 

Some questions & remarks below, but otherwise no blocker on my side

server/paymentProviders/transferwise/index.ts
kewitz: That's a good idea. This `else-if` is definitely a logic simplification.

server/paymentProviders/transferwise/index.ts
kewitz: I think it can. But since we're just using this to calculate how much we should send in order to cover the fees I think this is fine.

PR #8403, opened 2023-01-12
Increase default 2FA session duration to 4h

Betree:
I'm afraid that requiring ops/support teams to re-enter their 2FA every hour can be time-consuming, and the pain could push them to use less secure behaviors (like having their 2FA validated on their local machine rather than a 2nd device).

We could have more frequent requests with [Yubikeys](https://github.com/opencollective/opencollective/issues/5294) in place as it's faster to complete the process, but for now, 4h sounds like a safe enough value to me.

### DIFF
--- a/server/lib/two-factor-authentication/lib.ts
+++ b/server/lib/two-factor-authentication/lib.ts
@@ -11,7 +11,7 @@ import { hasPolicy } from '../policies';
 
 import totp from './totp';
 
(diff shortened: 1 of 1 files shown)


Betree:
Note that for critical operations like deleting the accounts, we'll still ask for 2FA regardless of pre-existing sessions (thanks to the `alwaysAskForToken` parameter).

hdiniz (approved):

PR #8412, opened 2023-01-16
Fix(OSC application): Allow underscores in GitHub handles

gustavlrsn:
Fix for https://github.com/opencollective/opencollective/issues/6280.

The function to parse GitHub URLs into GitHub handles was not expecting underscores. So `https://github.com/brendan8c/FFMPEG_BAT` became `brendan8c/FFMPEG`.

My modification of the RegEx pattern needs a review, it seems to work as expected but I'm not an expert in RegEx syntax.


Betree (commented):
Good catch! Let's add a test in `test/server/lib/github.test.ts` to prevent future regressions

Betree (approved):
Nice fix :+1:

gustavlrsn:
@Betree Thank you!

PR #8417, opened 2023-01-16
Allow Projects and Events to have their own Admins

kewitz:
- [x] Expose inherited member property to the frontend
- [x] Delete existing members from children collectives


znarf (approved):
Looks good!

PR #8422, opened 2023-01-17
feat(Autocomplete tags): Add CollectiveTagStats materialized view and extend tagStats query to use it

gustavlrsn:
API for https://github.com/opencollective/opencollective/issues/5523

Adding a `CollectiveTagStats` materialized view with Collective tag counts, and extending the current `tagStats` query to use it when there is no `searchTerm` provided (which is an existing argument to return tags for collectives that match that search term).

The materialized view also has a HostCollectiveId field by which you can return popular tags for that specific host. When this field is `NULL`, the tag counts are across all collectives (not just ones that have a `NULL` HostCollectiveId).

The PR is also adding an `tagSearchTerm` argument to the `tagStats` query, so that it can be used in an autocomplete tag input field to find matches to what you are currently typing in, as well as a `host` argument.


gustavlrsn:
> We might want another similar materialized view grouping the tag counts by host, to be able to show popular tags specific to that host in the application process.
> * Maybe there a way to combine those into one view, grouped by host, with the caveat that the records grouped by a `null` host field includes data from hosts, to easily get the standard view.

Here's a query/view would achieve this. Perhaps something to include from the beginning to avoid another migration? Maybe together with an index on HostCollectiveId.
```
SELECT tag, count, null AS "HostCollectiveId"
FROM (SELECT DISTINCT unnest(tags) AS tag, COUNT(*) AS count FROM "Collectives" GROUP BY tag) AS AllTags

UNION

SELECT tag, count, "HostCollectiveId"
FROM (SELECT DISTINCT unnest(tags) AS tag, COUNT(*) AS count, "HostCollectiveId" FROM "Collectives" WHERE "HostCollectiveId" IS NOT NULL GROUP BY tag, "HostCollectiveId") AS HostTags

ORDER BY count DESC
```

gustavlrsn:
On second thought I think it makes sense to wait a bit with adding host tag frequencies to this view until we are 100% sure we will use it. This PR together with an autocomplete tag input in the frontend will be a nice first step.

migrations/20230113201828-create-tag-stats-materialized-view.ts
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+    CREATE MATERIALIZED VIEW IF NOT EXISTS "CollectiveTagStats" AS
+    SELECT DISTINCT unnest(tags) AS tag, COUNT(*) AS count FROM "Collectives" GROUP BY tag ORDER BY count DESC
```
Betree, line 8: Let's maybe put the `FROM`, `GROUP BY` and `ORDER BY` to new lines to make that easier to read

Betree, line 8: I'm afraid of the pollution that would come from non-legit accounts. We should at least add a `WHERE "deletedAt" IS NULL`. I would personally go a bit further than that:
- I don't think we care about tags on user profiles
- If we do care about them, we should at least filter `isIncognito IS NOT TRUE AND (c."data" ->> 'isGuest')::boolean IS NOT TRUE `

To prevent SPAM, we may also want to pick some of the conditions implemented for search, such as `skipRecentAccounts` (`AND (COALESCE((c."data"#>>'{spamReport,score}')::float, 0) <= 0.2 OR c."createdAt" < (NOW() - interval '2 day'))`).

Also `AND ((data ->> 'hideFromSearch'::text)::boolean) IS NOT TRUE`

Betree (commented):
> Maybe there a way to combine those into one view, grouped by host, with the caveat that the records grouped by a null host field includes data from hosts, to easily get the standard view.

I agree with you that we don't have to resolve this now.

_Edit: Posted too fast - great work!_

migrations/20230113201828-create-tag-stats-materialized-view.ts
gustavlrsn: @Betree great feedback! 

Regarding user profile tags, I agree that user tags might not be too interesting. However:
  - We have the edit tags field on User collectives in the settings, although probably not very well used since we don't show tags on User profile pages, so maybe a "suboptimal experience" there could be ok by either:
    - skipping autocompletion on that field
    - just autocomplete on Collective tags
- the current query is including User tags, but I don't think it is really too interesting there either since the query is currently only used on the search page which does not include User collectives. So we might want to filter out User collectives in that query too to keep them consistent, if we filter out User tags in this view.

Hmm. I'm happy to go either way, what do you think?

Betree: No strong opinion on my side; I would say go with the solution that seems to be the simplest for you

gustavlrsn: Fixed!

gustavlrsn: @Betree Ok, I'll keep the inclusion of User tags as is for now, will probably require less changes overall. Just pushed the fixes you suggested

gustavlrsn:
Backtracking my backtrack and added the tag counts per host to the materialized view in this PR, since I'll be using that in the frontend to show popular tags for the host in the application form.

Betree (approved):
Perfect!

PR #8429, opened 2023-01-18
Match or create payment method on payment intent success

hdiniz:

This is a base for handling card payment made using the Stripe Payment Element.

When a payment is made using the element, we get the payment method data on the intent succeeded event.

### DIFF
--- a/server/paymentProviders/stripe/webhook.ts
+++ b/server/paymentProviders/stripe/webhook.ts
@@ -31,7 +31,73 @@ import * as virtualcard from './virtual-cards';
 
 const debug = debugLib('stripe');
 
+async function createOrUpdateOrderStripePaymentMethod(
+  order: typeof models.Order,
+  stripeAccount: string,
+  paymentIntent: Stripe.PaymentIntent,
+): typeof models.PaymentMethod {
+  const stripePaymentMethodId =
+    typeof paymentIntent.payment_method === 'string' ? paymentIntent.payment_method : paymentIntent.payment_method?.id;
+
+  const orderPaymentMethod = await models.PaymentMethod.findByPk(order.PaymentMethodId);
+  // order paymentMethod already saved.
+  if (orderPaymentMethod?.data?.stripePaymentMethodId === stripePaymentMethodId) {
+    return orderPaymentMethod;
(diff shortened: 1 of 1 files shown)


server/paymentProviders/stripe/webhook.ts
```diff
@@ -31,7 +31,73 @@ import * as virtualcard from './virtual-cards';
...
+  // order payment method already exists, update order with it.
+  if (matchingPaymentMethod) {
+    await order.update({
+      PaymentMethodId: matchingPaymentMethod.id,
+    });
+    return matchingPaymentMethod;
+  }
+
+  const stripePaymentMethod = await stripe.paymentMethods.retrieve(
+    typeof paymentIntent.payment_method === 'string' ? paymentIntent.payment_method : paymentIntent.payment_method?.id,
```
kewitz, line 67:
```suggestion
   stripePaymentMethodId,
```

server/paymentProviders/stripe/webhook.ts
```diff
@@ -500,12 +568,31 @@ function formatPaymentMethodName(paymentMethod: Stripe.PaymentMethod) {
...
 
+function mapStripePaymentMethodExtraData(pm: Stripe.PaymentMethod): object {
+  if (pm.type === 'card') {
+    return {
+      brand: pm.card.brand,
+      country: pm.card.country,
+      expYear: pm.card.exp_year,
+      expMonth: pm.card.exp_month,
+      funding: pm.card.funding,
+      fingerprint: pm.card.fingerprint,
```
kewitz, line 588: I think we `fingerprint` only exists when creating a new payment method with a token. We need to take care we're not overwriting the original token with an undefined or null value.

kewitz (approved):

server/paymentProviders/stripe/webhook.ts
hdiniz: This is one card created from the PaymentElement component. Not sure if it exists for all methods, but this one is only for cards.

https://dashboard.stripe.com/test/payments/pi_3MS3rLDjPFcHOcTm3QN0oedH

image

PR #8437, opened 2023-01-19
Password implementation

znarf:
Implement: https://github.com/opencollective/opencollective/issues/6337

Frontend: https://github.com/opencollective/opencollective-frontend/pull/8573


server/graphql/v2/mutation/IndividualMutations.ts
```diff
@@ -0,0 +1,89 @@
...
+    args: {
+      password: {
+        type: new GraphQLNonNull(GraphQLString),
+        description: 'The password to set.',
+      },
+      currentPassword: {
+        type: GraphQLString,
+        description: 'The current password (if any) to confirm password change.',
+      },
+    },
```
hdiniz, line 54: This mutation should also be rate limited.

hdiniz:
The sensitive fields have to be redacted from logs and possible error event logging.

server/graphql/v2/mutation/IndividualMutations.ts
znarf: What would we be preventing by rate limiting this?

znarf:
> The sensitive fields have to be redacted from logs and possible error event logging.

@hdiniz What's your recommendation to achieve this?

server/graphql/v2/mutation/IndividualMutations.ts
hdiniz: If an attacker gains access to a valid token to use with this mutation (password reset, or a session token), they would be able to brute force a password change, right?

hdiniz:
> > The sensitive fields have to be redacted from logs and possible error event logging.
> 
> @hdiniz What's your recommendation to achieve this?

These are places where I think we can prevent accidental logging of the fields:

- Use beforeSend and beforeSendTransaction hooks on the Sentry SDK init 
 https://docs.sentry.io/platforms/node/data-management/sensitive-data/#scrubbing-data
https://github.com/opencollective/opencollective-api/blob/main/server/lib/sentry.ts#L31

- Review places where we are logging graphql variables or request params. For Sentry, the hook above might cover everything.
https://github.com/opencollective/opencollective-api/blob/main/server/lib/sentry.ts#L256

- Sentry has an option to scrub sensitive fields when they are ingested, we can enable this and specify field names. This might be a catch all for Sentry, but we would still be sending the data out. 
https://docs.sentry.io/product/data-management-settings/scrubbing/server-side-scrubbing/

- Hyperwatch logs
https://github.com/opencollective/opencollective-api/blob/main/server/lib/hyperwatch.js#L77

server/graphql/v1/types.js
```diff
@@ -260,6 +260,15 @@ export const UserType = new GraphQLObjectType({
           }
         },
       },
+      hasPassword: {
+        type: GraphQLBoolean,
+        description: 'Has the account a password set? For authenticated user: scope: "account".',
```
Betree, line 265: Since it's v1

```suggestion
        description: 'Has the account a password set? For authenticated user',
```

server/graphql/v2/mutation/IndividualMutations.ts
```diff
@@ -0,0 +1,98 @@
...
+    async resolve(_: void, args, req: express.Request): Promise<Record<string, unknown>> {
+      checkRemoteUserCanUseAccount(req);
+
+      const rateLimit = new RateLimit(`individual_set_password_${req.remoteUser.id}`, 10, ONE_HOUR_IN_SECONDS, true);
+      if (!(await rateLimit.registerCall())) {
+        throw new RateLimitExceeded();
+      }
+
+      // Enforce 2FA
+      const account = req.remoteUser.getCollective();
```
Betree, line 65: Interesting that it doesn't crash `enforceForAccountAdmins`. We should probably add a nil check in the function to prevent that. I also assume that typescript did not detect the error because we haven't typed `Collective` yet.

```suggestion
      const account = await req.remoteUser.getCollective();
```

server/graphql/v2/mutation/IndividualMutations.ts
```diff
@@ -0,0 +1,98 @@
...
+      await TwoFactorAuthLib.enforceForAccountAdmins(req, account, { alwaysAskForToken: true });
```
Betree, line 66: :+1: Nice to always ask

server/graphql/v2/mutation/IndividualMutations.ts
```diff
@@ -0,0 +1,98 @@
...
+          throw new Unauthorized('Submit current password to change password.');
+        }
+        const validPassword = await bcrypt.compare(args.currentPassword, req.remoteUser.passwordHash);
+        if (!validPassword) {
+          return new Unauthorized('Invalid current password while attempting to change password.');
+        }
+      }
+
+      // If we're there, it's a success, we can reset the rate limit count
+      await rateLimit.reset();
```
Betree, line 80: :+1:

server/graphql/v2/mutation/IndividualMutations.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      // TODO: extract saltRounds in some configuration or shared library
+      const passwordHash = await bcrypt.hash(args.password, /* saltRounds */ 10);
```
Betree, line 83: I agree with should move this to config

server/graphql/v2/mutation/IndividualMutations.ts
```diff
@@ -0,0 +1,98 @@
...
+      const user = await req.remoteUser.update({ passwordHash });
+
+      await models.Activity.create({
+        type: activities.USER_PASSWORD_SET,
+        UserId: user.id,
+        FromCollectiveId: user.CollectiveId,
+        CollectiveId: user.CollectiveId,
```
Betree, line 90: In case it's done through a 3rd party app

```suggestion
        CollectiveId: user.CollectiveId,
        UserTokenId: req.userToken?.id,
```

Betree (commented):
Looking great!

Just one blocking feedback with a missing `await` below

PR #8443, opened 2023-01-20
Implement UploadedFile model

Betree:
:warning: Require https://github.com/opencollective/opencollective-frontend/pull/8574 to be merged & deployed first
Resolve https://github.com/opencollective/opencollective/issues/6345

**Breaking changes**
- The `/images` REST endpoint now requires a `kind` to upload the file (hence the requirement on https://github.com/opencollective/opencollective-frontend/pull/8574)
- The `/images` REST endpoint is now authenticated

**Differences from the specs**
- Renamed `UploadedFiles.type` in the original proposal to `UploadedFiles.kind` as there was otherwise some confusion  with `UploadedFiles.fileType`
- Removed the `CollectiveId` column since we've defined that linking entities (expense items, updates, etc) was not on the menu for this first draft. Introducing such a column is a risk of adding more confusion/complexity and wouldn't solve anything. We're instead relying on the URL to get the data from the table, which serves the core purpose it's supposed to: provide some metadata about the uploaded files.
- I added the ability to override the existing filename, to match an old design where it was possible to customize it after the upload

**Questions**
- Should the blur hash generation happen client-side?


server/controllers/images.ts
```diff
@@ -0,0 +1,141 @@
...
+    return next(new errors.ServerError('S3 service object not initialized'));
+  }
+
+  /**
+   * We will replace the name to avoid collisions
+   */
+  const fileName = getFilename(file, req.body.fileName);
+  const uploadParams = {
+    Bucket: config.aws.s3.bucket,
+    Key: `${kebabCase(kind)}/${uuid()}/${fileName || uuid()}`,
```
hdiniz, line 125: My suggestion is to add the kind to metadata and file name to the content disposition. And to have a single uuid to refer to this file in storage. 

```suggestion
    Key: `uploaded-files/${uuid()}`,
```

server/models/UploadedFile.ts
```diff
@@ -0,0 +1,160 @@
...
+    },
+    url: {
+      type: DataTypes.STRING,
+      allowNull: false,
+      unique: true,
+      validate: {
+        notNull: true,
+        notEmpty: true,
+        isUrl: true,
+        isValidURL(url: string): void {
```
hdiniz, line 120: I used minio to run this PR locally as a replacement for s3, would be nice to allow localhost URLs here in development
http://localhost:9000/opencollective-dvl/dd47d3c0-9c42-11ed-a0b2-b390eb1764c5

server/controllers/images.ts
```diff
@@ -0,0 +1,141 @@
...
+    Body: file.buffer,
+    ACL: 'public-read',
+    ContentLength: file.size,
+    ContentType: file.mimetype,
```
hdiniz, line 129: ContentDisposition for setting the download filename when accessing the file directly through s3.
Metadata to help identify the source of the file


```suggestion
    ContentType: file.mimetype,
    ContentDisposition: `attachment; filename="${fileName}"`,
    Metadata: {
      CreatedByUserId: `${req.remoteUser.id}`,
      FileKind: kind,
    },
```

server/controllers/images.ts
Betree: Storing under the right kind is meant to better organize files when browsing from S3's UIs. What are the concerns with the kind and filename being visible in the path?

server/controllers/images.ts
Betree: +1 for the `Metadata` field, that's useful.

Not so convinced by the `Content-Disposition` one, especially since we'll have to sanitize the `"` in the filename to make it work. I tend to prefer the `/:uuid/:filename` approach for this reason.

server/controllers/images.ts
hdiniz: My reasoning was that the file name could contain PII or inapropriate user generated content by itself.
But having the url means having access to the file, so I dont know if thats would be an issue.

server/controllers/images.ts
hdiniz: If the file is saved with the original file name, I think there is no need for the content-disposition to be set.

server/controllers/images.ts
```diff
@@ -0,0 +1,145 @@
+import path from 'path';
+
+import { ManagedUpload } from 'aws-sdk/clients/s3';
+import { encode } from 'blurhash';
+import config from 'config';
+import { kebabCase } from 'lodash';
+import sharp from 'sharp';
+import { v1 as uuid } from 'uuid';
```
hdiniz, line 8: future v1 values can be predicted based on known past values, so this could be a v4?

server/controllers/images.ts
```diff
@@ -0,0 +1,145 @@
...
+      user: req.remoteUser,
+      transactionName: 'recordUploadedFile',
+      extra: { kind, fileName, fileType: file.mimetype },
+    });
+  }
+};
+
+const getFilename = (file, fileNameFromArgs) => {
+  const expectedExtension = SUPPORTED_FILE_EXTENSIONS[file.mimetype];
+  const rawFileName = fileNameFromArgs || file.originalname || uuid();
```
hdiniz, line 64: Do you think we should limit the filename size input? s3 object keys are limited to 1024 bytes.

Betree: Let's do that

hdiniz (approved):

PR #8444, opened 2023-01-20
Make Sure `arg` Currencies are Checked in Order Mutations

SudharakaP:
The condition in question seems to not check for anything if it only considers the `order[field]`. I believe this is an error. We should check `arg.order[field]` with `order.currency` instead here.


Betree (approved):
Good catch :+1: Another one where typescript would have helped

PR #8449, opened 2023-01-23
Update payment method before transaction is created

hdiniz:
Fix to update the payment method data before transactions are created, when the payment type is `paymentintent` it is updated to a concrete type.

### DIFF
--- a/server/paymentProviders/stripe/webhook.ts
+++ b/server/paymentProviders/stripe/webhook.ts
@@ -113,6 +113,8 @@ export const paymentIntentSucceeded = async (event: Stripe.Event) => {
     return;
   }
 
+  await createOrUpdateOrderStripePaymentMethod(order, stripeAccount, paymentIntent);
+
   // Recently, Stripe updated their library and removed the 'charges' property in favor of 'latest_charge',
   // but this is something that only makes sense in the LatestApiVersion, and that's not the one we're using.
   const charge = (paymentIntent as any).charges.data[0] as Stripe.Charge;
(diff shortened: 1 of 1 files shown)


server/paymentProviders/stripe/webhook.ts
```diff
@@ -83,7 +82,14 @@ async function createOrUpdateOrderStripePaymentMethod(
...
     },
-  });
+  };
+
+  if (orderPaymentMethod.type === PAYMENT_METHOD_TYPE.PAYMENT_INTENT) {
+    return await orderPaymentMethod.update(paymentMethodFields);
+  }
+
+  // new payment method
+  const pm = await models.PaymentMethod.create(paymentMethodFields);
```
kewitz, line 92: Shouldn't this be an alternate condition to an existing `orderPaymentMethod`? It is not very clear why are we juggling two different payment methods inside this function. 🤔

hdiniz: What do you mean an alternate condition? Do you think this could be refactored to be more clear? Looking at it again, I think we could do without the `if (orderPaymentMethod.type === PAYMENT_METHOD_TYPE.PAYMENT_INTENT)` branch and just create the payment intent as it was before.

The idea in this function is that when a payment is created using the paymentintent payment method type, we only know the concrete payment method (creditcard, etc) when the  success webhook is called for it.

When that happens we have these conditions:

1. The payment method is one that we already store, we update the order to it.
2. The payment method does not exist yet, we create it and update the order to it.

Both cases leave a `paymentintent` type payment method hanging around unattached.

kewitz: Makes sense. In theory, will the user have one paymentintent hanging for each host it donated to?
Should we delete these paymentintents after the success webhook is called?

kewitz (approved):

PR #8450, opened 2023-01-23
Include tags in webhook order data

pwolanin:
Follow-up to the last change in https://github.com/opencollective/opencollective-frontend/pull/8577

Now that tags are getting passed through from the query string and saved to the order, include them in the order webhook data.

### DIFF
--- a/server/lib/webhooks.js
+++ b/server/lib/webhooks.js
@@ -47,6 +47,7 @@ const getOrderInfo = order => {
       'totalAmount',
       'currency',
       'description',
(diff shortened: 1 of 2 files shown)


Betree (approved):
Kudos for including a test! :raised_hands: 

Looks good to me, I'll wait for the tests and will merge & deploy tomorrow if everything's green.

pwolanin:
Looks like I had a silly unterminated string.  Will push that fix.

PR #8452, opened 2023-01-23
Add method to redact sensitive fields from sentry logging

hdiniz:
Here is an example transaction in sentry if the field `variables.paymentIntent.fromAccount.legacyId` was redacted:

https://sentry.io/organizations/open-collective/performance/oc-api:b02a61a3047c4a4dbb17ba25c0dbb895/?breakdown=none&project=5199682&query=http.method%3APOST&showTransactions=p95&sort=-timestamp&statsPeriod=24h&transaction=GraphQL%3A+CreatePaymentIntent

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,6 +55,7 @@
         "express-limiter": "1.6.1",
         "express-session": "1.17.3",
         "express-ws": "5.0.2",
+        "fast-redact": "3.1.2",
         "fs-extra": "11.1.0",
         "get-urls": "10.0.1",
         "graphql": "16.6.0",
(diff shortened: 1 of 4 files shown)


server/lib/hyperwatch.js
```diff
@@ -74,7 +74,8 @@ const load = async app => {
         log = log.deleteIn(['request', 'headers', 'cookie']);
 
         if (req.body && req.body.query && req.body.variables) {
-          log = log.set('graphql', req.body);
+          const bodyStr = redactSensitiveFields(req.body);
```
znarf, line 77: Could we have `redactSensitiveFields` return an object and not a string?

hdiniz: Yes, do you think is preferred? I was on the fend for this as the Sentry method takes a serialized string and the hyperwatch an object.
This library used to redact always return the serialized version.

Betree (commented):
We should also aim to redact headers (e.g. `Authorization`)

Betree:
I'm cherry-picking some of these commits in https://github.com/opencollective/opencollective-api/pull/8823

PR #8453, opened 2023-01-24
Fix amount checks in `createOrder`

Betree:
We missed updating this code when we introduced flexible amounts for PRODUCT/SERVICE tiers and let some orders go without being properly checked, resulting in issues like https://github.com/opencollective/opencollective/issues/6365 not being detected.

### DIFF
--- a/server/graphql/v1/mutations.js
+++ b/server/graphql/v1/mutations.js
@@ -1,4 +1,5 @@
 import { GraphQLBoolean, GraphQLInt, GraphQLList, GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
+import { isNil } from 'lodash';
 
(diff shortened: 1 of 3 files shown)


server/graphql/v1/mutations.js
```diff
@@ -270,7 +270,11 @@ const mutations = {
       },
     },
     async resolve(_, args, req) {
-      const { order } = await createOrder(args.order, req);
+      // Pass the gross amount for a single item
+      const amount =
+        (args.order.totalAmount - (args.order.taxAmount || 0) - (args.order.platformFee || 0)) /
+        (args.order.quantity || 1);
```
Betree, line 276: Had to do this because we still use the V1 mutation in many tests

znarf (approved):

PR #8455, opened 2023-01-24
Handle Stripe mandate.updated webhook

hdiniz:

Handle a payment method mandate update, some payment methods (such as Bacs) require mandates to confirm a new payment.

https://stripe.com/docs/payments/payment-methods/bacs-debit#mandates

### DIFF
--- a/migrations/20230130124611-add-payment-method-stripe-payment-id-unique-index.js
+++ b/migrations/20230130124611-add-payment-method-stripe-payment-id-unique-index.js
@@ -4,8 +4,8 @@
 module.exports = {
   async up(queryInterface) {
     await queryInterface.sequelize.query(`
-      CREATE UNIQUE INDEX "PaymentMethods__stripePaymentMethodId" ON "PaymentMethods"(("data"->>'stripePaymentMethodId'))
-      WHERE "data"->>'stripePaymentMethodId' IS NOT NULL
+      CREATE UNIQUE INDEX "PaymentMethods__stripePaymentMethodId" ON "PaymentMethods"(("data"#>>'{stripePaymentMethodId}'))
+      WHERE "data"#>>'{stripePaymentMethodId}' IS NOT NULL AND "deletedAt" IS NULL
     `);
   },
 
--- a/migrations/20230130124611-add-payment-method-stripe-payment-id-unique-index.js
+++ b/migrations/20230130124611-add-payment-method-stripe-payment-id-unique-index.js
@@ -1,0 +1,15 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+      CREATE UNIQUE INDEX "PaymentMethods__stripePaymentMethodId" ON "PaymentMethods"(("data"->>'stripePaymentMethodId'))
+      WHERE "data"->>'stripePaymentMethodId' IS NOT NULL
+    `);
+  },
+
+  async down(queryInterface) {
+    await queryInterface.removeIndex('PaymentMethods', 'PaymentMethods__stripePaymentMethodId');
+  },
+};
--- a/server/paymentProviders/stripe/webhook.ts
+++ b/server/paymentProviders/stripe/webhook.ts
@@ -679,7 +679,7 @@ async function paymentMethodAttached(event: Stripe.Event) {
       await pm.update({
         customerId: stripeCustomerId,
         CollectiveId: stripeCustomerAccount.CollectiveId,
-      })
+      });
       return;
     }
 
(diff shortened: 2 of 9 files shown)


server/paymentProviders/stripe/webhook.ts
```diff
@@ -92,6 +92,35 @@ async function createOrUpdateOrderStripePaymentMethod(
...
+    where: {
+      data: {
+        stripePaymentMethodId,
+      },
+    },
+  });
+
+  if (!paymentMethod) {
+    return;
+  }
```
kewitz, line 111: 👍🏼

server/paymentProviders/stripe/webhook.ts
```diff
@@ -92,6 +92,35 @@ async function createOrUpdateOrderStripePaymentMethod(
...
+
+  await paymentMethod.update({
+    saved: stripeMandate.type === 'multi_use' && stripeMandate.status !== 'inactive',
+    data: {
+      ...paymentMethod.data,
+      stripeMandate,
+    },
+  });
+
```
kewitz, line 120: Wouldn't be a good idea to cancel any subscriptions if the mandate goes inactive?

kewitz (approved):

server/paymentProviders/stripe/webhook.ts
hdiniz: Stripe mentions that the [Bacs mandate](https://stripe.com/docs/payments/payment-methods/bacs-debit#mandates) can be reactivated, I think we can check this at time of subscription charge to give it time for a reactivation / change of payment method.

server/paymentProviders/stripe/webhook.ts
```diff
@@ -92,6 +92,55 @@ async function createOrUpdateOrderStripePaymentMethod(
...
+      type: stripePaymentMethod.type,
+      confirmedAt: new Date(),
+      saved: stripeMandate.type === 'multi_use' && stripeMandate.status !== 'inactive',
+      data: {
+        stripePaymentMethodId: stripePaymentMethod.id,
+        stripeAccount,
+        ...mapStripePaymentMethodExtraData(stripePaymentMethod),
+        stripeMandate,
+      },
+    });
```
hdiniz, line 128: @kewitz I found while testing this that the `mandate.updated` event is received before a payment method is attached to a customer. So at this time it might be that the payment method does not exist on our side and so we lose the information about the mandate.

I have changed here to create a payment method when we first see it on the `mandate.updated` event, and when we receive the `payment_method.attached` we save the payment method to the customer.

kewitz: We seem to have a race condition between `mandate.update` and `payment_method.attached` since Stripe doesn't guarantee the order of events sent, nor can we guarantee the speed at which we return these functions.

I suggest wrapping both handlers in transactions. We can also use the method `findOrCreate` to simplify the first conditional of these handlers.

migrations/20230130124611-add-payment-method-stripe-payment-id-unique-index.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+      CREATE UNIQUE INDEX "PaymentMethods__stripePaymentMethodId" ON "PaymentMethods"(("data"->>'stripePaymentMethodId'))
```
kewitz, line 7: I'm not 100% sure, but I had some trouble in the past using the `->>` operator.
Sequelize always defaults to `#>>` and for some reason, it seems that the operator will impact the indexing.

migrations/20230130124611-add-payment-method-stripe-payment-id-unique-index.js
```diff
@@ -0,0 +1,15 @@
...
+      WHERE "data"->>'stripePaymentMethodId' IS NOT NULL
```
kewitz, line 8:
```suggestion
      WHERE "data"->>'stripePaymentMethodId' IS NOT NULL AND "deletedAt" IS NULL
```

migrations/20230130124611-add-payment-method-stripe-payment-id-unique-index.js
kewitz, line 7:
```suggestion
      CREATE UNIQUE INDEX "PaymentMethods__stripePaymentMethodId" ON "PaymentMethods"(("data"#>>'{stripePaymentMethodId}'))
```

I'm not 100% sure, but I had some trouble in the past using the `->>` operator.
Sequelize always defaults to `#>>` and for some reason, it seems that the operator will impact the indexing.

It would be nice to verify this behavior locally using `EXPLAIN ANALYZE`.

hdiniz: you're right!

PR #8456, opened 2023-01-24
Children Members Followup

kewitz:


server/graphql/v2/interface/Account.js
```diff
@@ -214,6 +214,10 @@ const accountFieldsDefinition = () => ({
         type: new GraphQLList(AccountType),
         description: 'Type of accounts (BOT/COLLECTIVE/EVENT/ORGANIZATION/INDIVIDUAL)',
       },
+      includeInheritedMembers: {
```
znarf, line 217: `includeInherited` ?

server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -17,9 +17,10 @@ import { AccountReferenceInput, fetchAccountWithReference } from '../input/Accou
 import { Member } from '../object/Member';
 
 const isLastAdmin = async (account, memberAccount) => {
+  const CollectiveId = account.ParentCollectiveId ? [account.ParentCollectiveId, account.id] : account.id;
```
znarf, line 20: Add a comment to explain why this is useful to do this.

PR #8458, opened 2023-01-25
Add configuration and docs for running s3 replacement

hdiniz:
For testing file uploads and other integrations with s3.


Betree (approved):
Cool! I was personally testing against the live buckets (we have some for testing in shared credentials), but that can be useful; especially for tests.

PR #8461, opened 2023-01-26
Fix email action links

Betree:
See https://developers.google.com/gmail/markup/actions/declaring-actions#microdata

itemprop="url" is not the right type for what we're trying to achieve here:
image


Betree:
Tested with a live email, it seems that this is not enough to make it work - we need to register with Google as well. Opened a follow-up issue here: https://github.com/opencollective/opencollective/issues/6382

PR #8462, opened 2023-01-26
Use a common tag sanitizing function and add script to sanitize tags

gustavlrsn:
Fix for https://github.com/opencollective/opencollective/issues/6326 and https://github.com/opencollective/opencollective/issues/6325

This PR:
- merges the two tag sanitizing functions `sanitizeTags` (used for all other tags except Collectives) and `cleanTags` (used for Collectives) into one function.
  - combines all existing sanitizations into one
  - it is also adding a tag transform step that currently converts `opensource` and `open-source` into `open source`
- adds the `validateTags` validator (that checks tag length and number of tags) to all tags fields on all models
- adds a script `db:sanitize-collective-tags` to clean up all Collective tags using the new `sanitizeTags` function


server/lib/tags.ts
```diff
@@ -24,11 +34,28 @@ export const validateTags = (tags: string[] | null, { maxNbTags = 30, maxTagLeng
...
+    return null;
+  } else if (typeof tags === 'string') {
+    tags = [tags];
+  }
+
+  const sanitizedTags = tags
+    .filter(Boolean) // Remove null values
+    .map(t => tagTransforms[t] || t) // Transform common formatting variations of popular tags
+    .map(t => t.trim()) // Trim tags
+    .map(t => t.toLowerCase()) // Lowercase
```
gustavlrsn, line 51: Move this before transforms

gustavlrsn: or rather tagTransform to the end.

migrations/20230126103108-sanitize-collective-tags.js
```diff
@@ -0,0 +1,28 @@
...
+      `SELECT id, tags FROM "Collectives" WHERE tags IS NOT NULL`,
+      { type: queryInterface.sequelize.QueryTypes.SELECT },
+    );
+
+    for (const collective of collectives) {
+      const sanitizedTags = sanitizeTags(collective.tags);
+      // Check if sanitized tags are different from the current tags
+      if (JSON.stringify(sanitizedTags) !== JSON.stringify(collective.tags)) {
+        await queryInterface.sequelize.query(
+          `UPDATE "Collectives" SET tags = ${sanitizedTags ? `'{${sanitizedTags.join(',')}}'` : 'NULL'} WHERE id = :id`,
```
gustavlrsn, line 16: Use replacement for the tags, to avoid SQL injections

migrations/20230126103108-sanitize-collective-tags.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+import { sanitizeTags } from '../server/lib/tags';
+
+module.exports = {
+  up: async queryInterface => {
+    const collectives = await queryInterface.sequelize.query(
+      `SELECT id, tags FROM "Collectives" WHERE tags IS NOT NULL`,
```
gustavlrsn, line 7: Exclude deleted Collectives

migrations/20230126103108-sanitize-collective-tags.js
gustavlrsn: Migration removed, using script with models call instead, not subjected to SQL injection risk.

hdiniz (approved):

server/models/Collective.js
```diff
@@ -553,8 +554,11 @@ const Collective = sequelize.define(
 
     tags: {
       type: DataTypes.ARRAY(DataTypes.STRING),
+      validate: {
+        validateTags,
+      },
```
Betree, line 559: As I remember it, validators run on each model update - not just when the field changed. So as mentioned when we reviewed this together during the call, we need a strategy for existing collectives that don't pass this validation.

test/server/lib/tags.test.ts
```diff
@@ -0,0 +1,19 @@
...
+describe('server/lib/tags', () => {
+  it('sanitize tags', () => {
+    const tags = [
+      'OpenCollective',
+      '#OpenCollective',
+      'o#penCollective',
+      'OpenCollective#',
+      '##OpenCollective',
+      ' test ',
+      'double  space',
```
Betree, line 14: Some interesting cases to add

```suggestion
      '    ', // Empty should be removed
      'OpEn-SoUrCe, // Should be unified under "open source"
```

Betree (commented):
This `sanitizeTags` helper is looking great. The only thing preventing the release for me right now is the `validate` function that is going to break some profiles - as per the queries I sent you during the call:

```sql
SELECT array_length(tags, 1) AS nb_tags, COUNT(*) AS nb_collectives
FROM "Collectives"
WHERE "deletedAt" IS NULL
AND tags IS NOT NULL
GROUP BY nb_tags
ORDER BY nb_tags
```

```sql
SELECT array_length(tags, 1) AS nb_tags, COUNT(*) AS nb_collectives, (COUNT(*)::float / (SELECT COUNT(*) FROM "Collectives" WHERE tags IS NOT NULL)) * 100 || '%' as percentage
FROM "Collectives"
WHERE "deletedAt" IS NULL
AND tags IS NOT NULL
GROUP BY nb_tags
ORDER BY nb_tags
```

You can remove the `validate` function as a quick solution. New values will still get sanitized.

server/models/Collective.js
gustavlrsn: My thinking was that when/if they try to update something on the Collective model, they would be prompted to fix it essentially. But I was then assuming the user would be trying to edit something on the Collective info settings page (and then could easily fix their tags)

But maybe the Collective model is updated as a side effect of doing other unrelated things, and having the tag validation error out then wouldn't be great.

test/server/lib/tags.test.ts
gustavlrsn: Added these!

server/models/Collective.js
```diff
@@ -553,8 +554,11 @@ const Collective = sequelize.define(
...
       set(tags) {
-        this.setDataValue('tags', cleanTags(tags));
+        this.setDataValue('tags', sanitizeTags(tags));
```
gustavlrsn, line 561:
```suggestion
        validateTags(tags);
        this.setDataValue('tags', sanitizeTags(tags));
```

gustavlrsn: @Betree Maybe this is a good middle way?

Technically this would only run when setting the tags array. However, it would run when saving the Edit Collective info page, since that query is currently setting all values regardless whether they change or not.

gustavlrsn:
@Betree added a step in `sanitizeTags` to split up tags on commas. Should be ready for review again.

gustavlrsn:
Fixing the tests..

server/models/Collective.js
```diff
@@ -563,7 +564,9 @@ const Collective = sequelize.define(
     tags: {
       type: DataTypes.ARRAY(DataTypes.STRING),
       set(tags) {
-        this.setDataValue('tags', cleanTags(tags));
+        const sanitizedTags = sanitizeTags(tags);
+        validateTags(sanitizedTags);
+        this.setDataValue('tags', sanitizedTags);
```
Betree, line 569: I suppose this is only triggered when creating or editing the field, which prevents validating it if the collective doesn't touch the field?

Betree (approved):
Looking good!

Betree:
The linting issue is unrelated, see https://github.com/opencollective/opencollective-api/pull/8521

server/models/Collective.js
gustavlrsn: @Betree Yes, only triggered when creating or editing the field.

However, worth knowing is that the Collective Info settings page does save all fields in the form, so it would be triggered there regardless of whether you changed the tags or not.

PR #8464, opened 2023-01-27
Create stripe customer on recurring contribution

hdiniz:
When a payment intent is created for a guest, a customer is not attached to it created, by the time the intent is confirmed, we need to create a customer if the payment method is meant to be reused.

### DIFF
--- a/server/paymentProviders/stripe/payment-intent.ts
+++ b/server/paymentProviders/stripe/payment-intent.ts
@@ -2,6 +2,7 @@ import config from 'config';
 import { pick, toUpper } from 'lodash';
 import type Stripe from 'stripe';
 
+import { Service } from '../../constants/connected_account';
 import OrderStatuses from '../../constants/order_status';
 import logger from '../../lib/logger';
 import { getApplicationFee } from '../../lib/payments';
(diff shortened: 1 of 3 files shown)


server/paymentProviders/stripe/payment-intent.ts
```diff
@@ -32,16 +33,47 @@ async function processNewOrder(order: typeof models.Order) {
...
 
-  if (order.data?.savePaymentMethod) {
+  if (isSavePaymentMethod) {
     // eslint-disable-next-line camelcase
     paymentIntentParams.setup_future_usage = 'off_session';
   }
 
+  let stripeCustomerAccount = await order.fromCollective.getCustomerStripeAccount(hostStripeAccount.username);
+  if (isSavePaymentMethod && !stripeCustomerAccount) {
+    const isPlatformHost = hostStripeAccount.username === config.stripe.accountId;
```
kewitz, line 51:
```suggestion
```

server/paymentProviders/stripe/webhook.ts
```diff
@@ -147,7 +147,6 @@ export const paymentIntentProcessing = async (event: Stripe.Event) => {
   await sequelize.transaction(async transaction => {
     const order = await models.Order.findOne({
       where: {
-        status: [OrderStatuses.NEW, OrderStatuses.PROCESSING],
```
kewitz, line 150: What other status are we expecting here? I think we should keep the list of expected status.

kewitz (changes requested):
I caught a duplicated `isPlatformHost` you can remove and I have a comment on the expected order status we're processing.

TBH, I'm not convinced we should offer to save payment methods for guest contributors.

server/paymentProviders/stripe/webhook.ts
hdiniz: It could be ERROR or ACTIVE if its a processing recurring order.

server/paymentProviders/stripe/payment-intent.ts
```diff
@@ -32,21 +32,51 @@ async function processNewOrder(order: typeof models.Order) {
...
+    paymentIntentParams.customer = customer.id;
+  } else if (stripeCustomerAccount) {
+    paymentIntentParams.customer = stripeCustomerAccount.username;
+  }
+
   try {
     const paymentIntent = await stripe.paymentIntents.update(order.data.paymentIntent.id, paymentIntentParams, {
       stripeAccount: hostStripeAccount.username,
     });
-    await order.update({ status: OrderStatuses.PROCESSING, data: { ...order.data, paymentIntent } });
```
hdiniz, line 49: @kewitz  Removing this status update here to keep the order as NEW, as the payment intent was not yet confirmed (might not be confirmed) and can be misleading to show this order as processing.

hdiniz:
> I caught a duplicated `isPlatformHost` you can remove and I have a comment on the expected order status we're processing.
> 
> TBH, I'm not convinced we should offer to save payment methods for guest contributors.

This is not saving the payment method on our side for guest, its only being saved and attached to a stripe customer so we can reuse the payment method on a recurring contribution.

This was not an issue for credit card payments before because we always cloned a card to a new customer.

kewitz (approved):

PR #8467, opened 2023-01-27
Add bacs debit payment method

hdiniz:
Requires https://github.com/opencollective/opencollective-api/pull/8455
Related https://github.com/opencollective/opencollective/issues/6341


kewitz (approved):
Nice and consistent implementation.
Sorry for asking this now, but why do we need a custom BACS `processOrder`? Isn't this generic enough to use the existing `paymentIntent` implementation?

hdiniz:
> Nice and consistent implementation. Sorry for asking this now, but why do we need a custom BACS `processOrder`? Isn't this generic enough to use the existing `paymentIntent` implementation?

The difference is mostly on how recurring/saved payment methods are used. Bacs requires a mandate to be attached to the intent and it is confirmed on the backend side.

Looking at bancontact next, it seems we will need a separate processOrder method to handle recurring/saved methods as well.

PR #8470, opened 2023-01-27
Add Support to Move Added Funds

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/5651

[move-added-funds.webm](https://user-images.githubusercontent.com/12435965/215042108-9025c518-5ab3-45af-8a09-134aa4f9bb04.webm)

image

### DIFF
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -624,7 +624,7 @@ const orderMutations = {
           );
 
           [, updatedDebits] = await models.Transaction.update(
-            { FromCollectiveId: fromAccount.id },
+            { CollectiveId: fromAccount.id },
             {
               transaction: dbTransaction,
               returning: ['id'],
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -542,7 +542,7 @@ const orderMutations = {
       );
       const ordersIds = orders.map(order => order.id);
       const addedFundOrders = orders.filter(order => isAddedFund(order));
-      const isCollective = fromAccount?.type === types.COLLECTIVE;
+      const isUser = fromAccount?.type === types.USER;
       const addedFundPaymentMethod = fromAccount
         ? await models.PaymentMethod.findOne({
             where: {
@@ -635,7 +635,7 @@ const orderMutations = {
           );
 
           // Update paymentMethodId in transactions for Added Funds
-          if (addedFundOrders.length > 0 && isCollective) {
+          if (addedFundOrders.length > 0 && !isUser) {
             await models.Transaction.update(
               { PaymentMethodId: addedFundPaymentMethod.id },
               {
@@ -684,7 +684,7 @@ const orderMutations = {
           where: { id: ordersIds },
         });
 
-        if (addedFundOrders.length > 0 && isCollective && addedFundPaymentMethod) {
+        if (addedFundOrders.length > 0 && !isUser && addedFundPaymentMethod) {
           await models.Order.update(
             { PaymentMethodId: addedFundPaymentMethod.id },
             {
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -19,7 +19,7 @@ import {
 
 import { roles } from '../../../constants';
 import activities from '../../../constants/activities';
-import { types as CollectiveType, types } from '../../../constants/collectives';
+import { types } from '../../../constants/collectives';
 import { Service } from '../../../constants/connected_account';
 import OrderStatuses from '../../../constants/order_status';
 import { PAYMENT_METHOD_SERVICE, PAYMENT_METHOD_TYPE } from '../../../constants/paymentMethods';
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -543,7 +543,7 @@ const orderMutations = {
 
       for (const order of orders) {
         if (fromAccount) {
-          if (isAddedFund) {
+          if (isAddedFund(order)) {
             if (get(order, 'fromCollective.HostCollectiveId', null) !== get(fromAccount, 'HostCollectiveId', null)) {
               throw new ValidationFailed(
                 `Moving Added Funds when the current source Account has a different Fiscal Host than the new source Account is not supported.`,
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -543,10 +543,12 @@ const orderMutations = {
 
       for (const order of orders) {
         if (fromAccount) {
-          if (isAddedFund && (order.fromCollective.HostCollectiveId || fromAccount.HostCollectiveId)) {
-            throw new ValidationFailed(
-              `Currently we only support moving added funds where both the payee and payer is of type USER`,
-            );
+          if (isAddedFund) {
+            if (get(order, 'fromCollective.HostCollectiveId', null) !== get(fromAccount, 'HostCollectiveId', null)) {
+              throw new ValidationFailed(
+                `Moving Added Funds when the current source Account has a different Fiscal Host than the new source Account is not supported.`,
+              );
+            }
           }
         }
 
--- a/server/graphql/v2/mutation/OrderMutations.js
+++ b/server/graphql/v2/mutation/OrderMutations.js
@@ -19,9 +19,10 @@ import {
 
 import { roles } from '../../../constants';
 import activities from '../../../constants/activities';
+import { types as CollectiveType, types } from '../../../constants/collectives';
 import { Service } from '../../../constants/connected_account';
 import OrderStatuses from '../../../constants/order_status';
-import { PAYMENT_METHOD_SERVICE } from '../../../constants/paymentMethods';
+import { PAYMENT_METHOD_SERVICE, PAYMENT_METHOD_TYPE } from '../../../constants/paymentMethods';
 import { purgeAllCachesForAccount } from '../../../lib/cache';
 import logger from '../../../lib/logger';
 import stripe, { convertToStripeAmount } from '../../../lib/stripe';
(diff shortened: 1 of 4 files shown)


server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -494,15 +495,28 @@ const orderMutations = {
...
+      let isAddedFunds;
+      let paymentMethod;
+
       for (const order of orders) {
-        const isUpdatingPaymentMethod = Boolean(fromAccount);
+        if (
+          fromAccount.HostCollectiveId !== order.collective.HostCollectiveId &&
+          fromAccount.type !== CollectiveType.USER
+        ) {
+          throw new ValidationFailed(`Currently we cannot move Added Funds authoring account to a different host`);
```
Betree, line 513: :ice_cream: Suggested simplification

```suggestion
          throw new ValidationFailed(`Added Funds cannot be moved to a different host`);
```

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -494,15 +495,28 @@ const orderMutations = {
...
       const paymentMethodIds = uniq(orders.map(order => order.PaymentMethodId).filter(Boolean));
       const ordersIds = orders.map(order => order.id);
+      let isAddedFunds;
+      let paymentMethod;
+
       for (const order of orders) {
-        const isUpdatingPaymentMethod = Boolean(fromAccount);
+        if (
+          fromAccount.HostCollectiveId !== order.collective.HostCollectiveId &&
+          fromAccount.type !== CollectiveType.USER
```
Betree, line 511: Why ignore `USER`s here?

Betree (changes requested):
As per https://github.com/opencollective/opencollective/issues/5651#issuecomment-1380523851:
- Where are we updating the Transactions `PaymentMethodId`?
- Where are we updating the Orders `PaymentMethodId`?

server/graphql/v2/mutation/OrderMutations.js
SudharakaP: Done. 👍🏽

server/graphql/v2/mutation/OrderMutations.js
SudharakaP: @Betree : If the user selects any `USER` this condition; `fromAccount.HostCollectiveId !== order.collective.HostCollectiveId ` will always be true and we'll throw the error. Let me know what you think.

SudharakaP:
> As per [opencollective/opencollective#5651 (comment)](https://github.com/opencollective/opencollective/issues/5651#issuecomment-1380523851):
> 
>     * Where are we updating the Transactions `PaymentMethodId`?
> 
>     * Where are we updating the Orders `PaymentMethodId`?

@Betree : Sorry, I did a refactor and somehow skipped this part during the process. I've added it back and now it should be good. 👍🏽

Betree:
Tests seem broken by the latest changes

Betree (commented):
Can we add a new test to cover moving added funds in `test/server/graphql/v2/mutation/OrderMutations.test.ts`?

SudharakaP:
> Tests seem broken by the latest changes

> Can we add a new test to cover moving added funds in `test/server/graphql/v2/mutation/OrderMutations.test.ts`?

@Betree : Seems there were some problems with the implementation when selecting multiple orders at once. I've corrected all and also added two tests. Let me know if all looks good. 👍🏽

server/graphql/v2/mutation/OrderMutations.js
Betree: Some hosts have the type "USER", and therefore have their `HostCollectiveId` set to themselves

Betree: Suggested alternative:

```es6
if (fromAccount) {
  if (fromAccount.HostCollectiveId && fromAccount.HostCollectiveId !== order.collective.HostCollectiveId) {
    throw new ValidationFailed(`Orders cannot be moved to a different host`);
  }
}
```

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -571,6 +602,19 @@ const orderMutations = {
               },
             },
           );
+
+          // Update paymentMethodId in transactions for Added Funds
+          if (addedFundOrders.length > 0 && isCollective) {
```
Betree, line 607: Why only for type=COLLECTIVE?

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -571,6 +602,19 @@ const orderMutations = {
...
+            await models.Transaction.update(
+              { PaymentMethodId: addedFundPaymentMethod.id },
+              {
+                transaction: dbTransaction,
+                where: {
+                  [Op.or]: addedFundOrders.map(order => ({ OrderId: order.id })),
+                },
+              },
+            );
+          }
```
Betree, line 617: We already have some `// Update transactions` above, why not just add `PaymentMethodId` there?

server/graphql/v2/mutation/OrderMutations.js
SudharakaP: If it's a `USER` we will have the `opencollective/host` payment method type. We shouldn't change that. Only for collectives I thought we should change the payment method type (to `opencollective/collective` type).

server/graphql/v2/mutation/OrderMutations.js
SudharakaP: This seems a bit complicated or maybe I am missing something simple. Note that the were clause is different here as we need to only consider added fund orders. Plus the fields to update is also different. I thought trying to do this in the above transactions makes it too complicated with some conditionals. 🤔

server/graphql/v2/mutation/OrderMutations.js
Betree: What if it's a fund, an organization, an event, a project?

server/graphql/v2/mutation/OrderMutations.js
SudharakaP: Ah, great point. I totally missed that hosts can be of type `USER`. 🤔 I've corrected as per your suggestion.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -517,17 +518,44 @@ const orderMutations = {
...
+      const isUser = fromAccount?.type === types.USER;
+      const addedFundPaymentMethod = fromAccount
+        ? await models.PaymentMethod.findOne({
+            where: {
+              CollectiveId: fromAccount.id,
+              service: 'opencollective',
+              type: 'collective',
+              deletedAt: null,
+            },
+          })
```
znarf, line 546: What's that?

Certainly not looking like a payment for added funds. 

It's collective to collective.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -517,17 +518,44 @@ const orderMutations = {
...
+        : null;
+
       for (const order of orders) {
+        if (fromAccount) {
+          if (fromAccount.HostCollectiveId && fromAccount.HostCollectiveId !== order.collective.HostCollectiveId) {
+            throw new ValidationFailed(`Added Funds cannot be moved to a different host`);
+          }
+        }
```
znarf, line 554: What's that? We're comparing the host of the sender to the host of the Collective?

That doesn't make much sense to me.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -517,17 +518,44 @@ const orderMutations = {
...
+
         const isUpdatingPaymentMethod = Boolean(fromAccount);
 
-        if (isUpdatingPaymentMethod) {
+        if (isUpdatingPaymentMethod && !isAddedFund(order)) {
```
znarf, line 558: I don't see what justify the condition here.

znarf (changes requested):
Some changes I could see here looks dangerously wrong.

server/graphql/v2/mutation/OrderMutations.js
znarf: The payment method for Added Funds is the one from the collective receiving funds.

I don't see what justify updating the payment method in this case.

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -594,6 +622,19 @@ const orderMutations = {
               },
             },
           );
+
+          // Update paymentMethodId in transactions for Added Funds
+          if (addedFundOrders.length > 0 && !isUser) {
```
znarf, line 627: I don't see why we would change the behavior for Individuals.

znarf:
I made several comments that could be confusing. Sorry, please ignore them for now, I marked all as "outdated". I'll be looking later at the final version.

SudharakaP:
> I made several comments that could be confusing. Sorry, please ignore them for now, I marked all as "outdated". I'll be looking later at the final version.

@znarf : Thanks. I think one of the thing that I tried to address in this PR is those cases where we have a typical Added fund where the `fromCollective` is of type `USER` but the request is to move this added fund to a collective. So we need to make it such that `fromCollective` is of type `COLLECTIVE` (in other words it becomes `opencollective/host` to `opencollective/collective`). I think this is the reason that there was added complexity because I tried to make sure both of these are handled. However reading your comments I've removed that part of the logic here and simply checked for `hostCollectiveId` as you suggested so that we now **only** move from type `USER` to another `USER` type. If we try we show this,

image

However I do think we need to support the aforementioned case as well; maybe in a separate PR if no objections. 😺

znarf:
It's generally not a good idea to have Collectives as origin of Added Funds. Who's asking for this? Why do we have to support it?

> Only support payee and payer individuals.

It's not common to have an individual as origin of Added Funds (payer) and it should not be possible to have individual as recipient of funds (payee). What is the use case, who's asking for this?

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -518,17 +518,34 @@ const orderMutations = {
...
+      );
       const ordersIds = orders.map(order => order.id);
+
       for (const order of orders) {
+        if (fromAccount) {
+          if (isAddedFund && (order.fromCollective.HostCollectiveId || fromAccount.HostCollectiveId)) {
+            throw new ValidationFailed(
+              `Currently we only support moving added funds where both the payee and payer is of type USER`,
+            );
+          }
```
znarf, line 543:
```suggestion
          if (isAddedFund) {
            if (order.fromCollective.HostCollectiveId !== fromAccount.HostCollectiveId) {
              throw new ValidationFailed(
                `Moving Added Funds when the current source Account has a different Fiscal Host than the new source Account is not supported.`,
              );
            }
          }
```

SudharakaP:
> It's generally not a good idea to have Collectives as origin of Added Funds. Who's asking for this? Why do we have to support it?

I thought or remember this was once done manually. That's what I thought it was necessary. I'll try to find this. 🤔 

> It should not be possible to have an individual as recipient of funds (payee). What is the use case, who's asking for this?

Sorry I meant only payee individual (not payee and payer).

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -518,17 +518,34 @@ const orderMutations = {
...
+          .filter(order => !isAddedFund(order))
+          .map(order => order.PaymentMethodId)
+          .filter(Boolean),
+      );
       const ordersIds = orders.map(order => order.id);
+
       for (const order of orders) {
+        if (fromAccount) {
+          if (isAddedFund && (order.fromCollective.HostCollectiveId || fromAccount.HostCollectiveId)) {
+            throw new ValidationFailed(
```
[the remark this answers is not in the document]
SudharakaP: I've simplified the error message to just say "Only moving added funds are supported".

server/graphql/v2/mutation/OrderMutations.js
SudharakaP: I've fixed this 👍🏽

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -518,17 +518,36 @@ const orderMutations = {
...
+        orders
+          .filter(order => !isAddedFund(order))
+          .map(order => order.PaymentMethodId)
+          .filter(Boolean),
+      );
       const ordersIds = orders.map(order => order.id);
+
       for (const order of orders) {
+        if (fromAccount) {
+          if (isAddedFund) {
```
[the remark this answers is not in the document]
SudharakaP: Ops seems there is a problem here actually. With the default value being different. I made a small correction so that both `order.fromCollective.HostCollectiveId` and `fromAccount.HostCollectiveId` has the same default value. Now it's all good. Sorry about that.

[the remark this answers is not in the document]
znarf: My comment was about `isAddedFund`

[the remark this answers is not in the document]
SudharakaP: Yes I've tested and it works. I am a bit confused here; you've suggested above https://github.com/opencollective/opencollective-api/pull/8470#discussion_r1119855919 this refactoring.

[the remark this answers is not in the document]
znarf: My proposal was based on the existing code. I could not check that the existing code was correct, this is your PR and your task.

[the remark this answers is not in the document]
znarf: `isAddedFund` is a function and not a variable, using it alone like this doesn't make any sense, it's always truthy.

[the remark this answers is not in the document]
SudharakaP: Oh... That makes sense, I was testing different added funds so they all work. But I think I didn't check for other contributions. I've corrected this now. Thanks. 🙏🏽

znarf (approved):
Time to ship this I guess.

PR #8484, opened 2023-01-30
Consider ParentCollective policies in hasPolicy and getPolicy

kewitz:
Implements Policy inheritance at the lib level.

### DIFF
--- a/server/lib/policies.ts
+++ b/server/lib/policies.ts
@@ -2,8 +2,18 @@ import { get, isUndefined } from 'lodash';
 
 import POLICIES, { DEFAULT_POLICIES, Policies } from '../constants/policies';
 
-export const hasPolicy = (collective, policy: POLICIES): boolean =>
-  !isUndefined(get(collective, ['data', 'policies', policy]));
+export const hasPolicy = async (collective, policy: POLICIES): Promise<boolean> => {
+  let account = collective;
+  if (collective?.ParentCollectiveId) {
+    account = await collective.getParentCollective();
+  }
+  return !isUndefined(get(account, ['data', 'policies', policy]));
+};
 
-export const getPolicy = <T extends POLICIES>(collective, policy: T): Policies[T] =>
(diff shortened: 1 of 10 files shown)


server/lib/policies.ts
```diff
@@ -2,8 +2,18 @@ import { get, isUndefined } from 'lodash';
 
 import POLICIES, { DEFAULT_POLICIES, Policies } from '../constants/policies';
 
-export const hasPolicy = (collective, policy: POLICIES): boolean =>
-  !isUndefined(get(collective, ['data', 'policies', policy]));
+export const hasPolicy = async (collective, policy: POLICIES): Promise<boolean> => {
+  let account = collective;
+  if (collective?.ParentCollectiveId) {
+    account = await collective.getParentCollective();
+  }
```
Betree, line 9: Since it's meant to be fetched in many places, multiple times per request, it would be great to have some caching to prevent unnecessary SQL requests. Maybe we can implement that directly in `Collective.prototype.getParentCollective`:

```diff
--- a/server/models/Collective.js
+++ b/server/models/Collective.js
@@ -915,14 +915,17 @@ Collective.prototype.updateSocialLinks = async function (socialLinks) {
   });
 };
 
-Collective.prototype.getParentCollective = function (options) {
+Collective.prototype.getParentCollective = async function (options) {
   if (!this.ParentCollectiveId) {
-    return Promise.resolve(null);
+    return null;
   }
-  if (this.parentCollective) {
-    return Promise.resolve(this.parentCollective);
+
+  if (options) {
+    return models.Collective.findByPk(this.ParentCollectiveId, options);
+  } else {
+    this.parentCollective = this.parentCollective || (await models.Collective.findByPk(this.ParentCollectiveId));
+    return this.parentCollective;
   }
-  return models.Collective.findByPk(this.ParentCollectiveId, options);
 };

```

Betree (approved):
Haven't tested but the code looks good to me!

PR #8488, opened 2023-01-31
Fix order for single tickets as guest

Betree:
Follow-up on https://github.com/opencollective/opencollective-api/pull/8100
Fix https://sentry.io/organizations/open-collective/issues/3898960355

Guest's orders for "single tickets" were crashing with `WHERE parameter "FromCollectiveId" has invalid "undefined" value`.

Also added a condition for the status: we should let people make another order for a ticket if the payment failed.


SudharakaP:
Thanks for fixing this. 👍🏽

PR #8489, opened 2023-01-31
PayPal unified & indexed charge ID

Betree:
As part of https://github.com/opencollective/opencollective/issues/6342, a recurring challenge with PayPal transactions is that they can be stored with different structures depending on what PayPal returns (a charge, a sale or a transaction).

This PRs keeps the existing behavior but introduces a new `data.paypalCaptureId` field that will be indexed, a performance improvement that is much needed for the reconciliation job.

### DIFF
--- a/migrations/20230131115800-index-paypal-transaction-id.js
+++ b/migrations/20230131115800-index-paypal-transaction-id.js
@@ -1,0 +1,43 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    // Copy all IDs to the new standard field (took 40s against a local prod dump)
+    await queryInterface.sequelize.query(`
+      UPDATE "Transactions"
+      SET "data" = JSONB_SET(
+          "data",
+          '{paypalCaptureId}',
+          TO_JSON(
+            COALESCE(
+              "data" -> 'capture' ->> 'id',
+              "data" -> 'paypalSale' ->> 'id',
+              "data" -> 'paypalTransaction' ->> 'id'
+            )::text
+          )::jsonb
+        )
+      WHERE "data" ->> 'paypalCaptureId' IS NULL
+      AND (
+        "data" -> 'capture' ->> 'id' IS NOT NULL
+        OR "data" -> 'paypalSale' ->> 'id' IS NOT NULL
+        OR "data" -> 'paypalTransaction' ->> 'id' IS NOT NULL
+      )
+    `);
+
+    // Create index
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "transactions__data_paypal_capture_id"
+      ON "Transactions"
+      USING BTREE (("data"->>'paypalCaptureId') ASC)
+      WHERE "data"->>'paypalCaptureId' IS NOT NULL
+      AND "deletedAt" IS NULL
+    `);
+  },
+
+  async down(queryInterface) {
+    await queryInterface.sequelize.query(`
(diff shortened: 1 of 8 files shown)


migrations/20230131115800-index-paypal-transaction-id.js
```diff
@@ -0,0 +1,43 @@
...
+        OR "data" -> 'paypalSale' ->> 'id' IS NOT NULL
+        OR "data" -> 'paypalTransaction' ->> 'id' IS NOT NULL
+      )
+    `);
+
+    // Create index
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "transactions__data_paypal_capture_id"
+      ON "Transactions"
+      USING BTREE (("data"->>'paypalCaptureId') ASC)
```
kewitz, line 32:
```suggestion
      USING BTREE (("data"#>>'{paypalCaptureId}') ASC)
```

I've been using this form to match the same operator Sequelize uses in its query because I had an index failing in the past.

scripts/paypal/payment-reconciliator.ts
```diff
@@ -460,7 +461,7 @@ const findOrphanSubscriptions = async (_, commander) => {
       }
 
       // Make sure all transactions exist in the ledger
-      for (const paypalTransaction of <Record<string, unknown>[]>response['transactions']) {
+      for (const paypalTransaction of <PaypalTransaction[]>response['transactions']) {
```
kewitz, line 464: I would rather type `fetchPaypalTransactionsForSubscription` result.
To make sure it is backward compatible and easy, you can type that function by combining a generic with a property type:
```ts
const fetchPaypalTransactionsForSubscription = async (host, subscriptionId): Promise<Record<string, unknown> & { transactions?: PaypalTransaction[] }> =>
```

server/paymentProviders/paypal/payment.ts
```diff
@@ -80,33 +81,48 @@ const recordTransaction = async (
...
-  capture,
+  capture: PaypalCapture,
   { data = undefined, createdAt = undefined } = {},
 ): Promise<typeof models.Transaction> => {
   const currency = capture.amount.currency_code;
   const amount = paypalAmountToCents(capture.amount.value);
   const fee = paypalAmountToCents(get(capture, 'seller_receivable_breakdown.paypal_fee.value', '0.0'));
-  return recordTransaction(order, amount, currency, fee, { data: { ...data, capture }, createdAt });
+  return recordTransaction(order, amount, currency, fee, {
+    data: { ...data, capture, paypalCaptureId: capture.id },
```
kewitz, line 123: Are we sure these IDs are not going to collide? Should we add a prefix based on the type of transaction?

migrations/20230131115800-index-paypal-transaction-id.js
Betree: Interesting, I've used it in many places and never faced any issues so far. Would be great to document the exact problem.

I'll commit the change to be safe

server/paymentProviders/paypal/payment.ts
Betree: They're not supposed to since they all represent the same thing. In any case, it's not a unique index, so it won't crash if that ever happens.

scripts/paypal/payment-reconciliator.ts
Betree: Good call

migrations/20230131115800-index-paypal-transaction-id.js
kewitz: I'll make a proper test for this. Sorry about causing fear.

PR #8492, opened 2023-02-01
Make sure 2fa can be reset using root actions

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4714

Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/8598

### DIFF
--- a/server/graphql/v1/CollectiveInterface.js
+++ b/server/graphql/v1/CollectiveInterface.js
@@ -1188,6 +1188,22 @@ const CollectiveFields = () => {
       description: 'Returns whether this host is trusted or not',
       resolve: collective => Boolean(get(collective, 'data.isTrustedHost')),
     },
+    isTwoFactorAuthEnabled: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      description: 'Returns whether this user has two factor authentication enabled',
+      async resolve(collective) {
+        if (collective.type === types.USER) {
+          const user = await models.User.findOne({
+            attributes: ['id', 'twoFactorAuthToken'],
+            where: { CollectiveId: collective.id },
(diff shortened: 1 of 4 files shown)


server/graphql/v1/CollectiveInterface.js
```diff
@@ -1188,6 +1188,22 @@ const CollectiveFields = () => {
...
+            attributes: ['id', 'twoFactorAuthToken'],
+            where: { CollectiveId: collective.id },
+          });
+          if (user.twoFactorAuthToken) {
+            return true;
+          }
+        }
+        return false;
+      },
+    },
```
Betree, line 1206: Let's not expose whether users have 2FA enabled to the entire world. Needs to be root or admin of the profile.

Betree (changes requested):

server/graphql/v1/CollectiveInterface.js
SudharakaP: Ah, missed this. 👍🏽 Corrected.

server/graphql/v2/mutation/RootMutations.ts
```diff
@@ -67,6 +67,10 @@ export default {
         type: GraphQLBoolean,
         description: 'Specify whether the account is a trusted host',
       },
+      isTwoFactorAuthEnabled: {
+        type: GraphQLBoolean,
+        description: 'Specify whether 2FA should be enabled',
```
Betree, line 72: In practice, we can't enable 2FA through this arg

```suggestion
        description: 'Set this to false to disable 2FA. Other values have no effect.',
```

Betree (approved):

PR #8493, opened 2023-02-01
Handle ticket order succeeded when using payment intent method

hdiniz:
Resolves https://github.com/opencollective/opencollective/issues/6384

When a card payment is made through the stripe payment element, the order is confirmation happens when we receive the payment intent succeeded webhook.

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -633,13 +633,17 @@ const validatePayment = payment => {
 
 const sendOrderConfirmedEmail = async (order, transaction) => {
   const attachments = [];
-  const { collective, tier, interval, fromCollective, paymentMethod } = order;
+  const { collective, interval, fromCollective, paymentMethod } = order;
   const user = await order.getUserForActivity();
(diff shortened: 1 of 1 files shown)


Betree (approved):
Lgtm

server/paymentProviders/stripe/webhook.ts
```diff
@@ -160,7 +160,7 @@ export const paymentIntentSucceeded = async (event: Stripe.Event) => {
   const paymentIntent = event.data.object as Stripe.PaymentIntent;
   const order = await models.Order.findOne({
     where: {
-      status: OrderStatuses.PROCESSING,
+      status: [OrderStatuses.NEW, OrderStatuses.PROCESSING],
```
Betree, line 163: This code is currently changing in https://github.com/opencollective/opencollective-api/pull/8497

pwolanin: Looks like that was merged, so this needs to be rebased/adjusted?

PR #8497, opened 2023-02-02
Listen for new paymentIntent based credit card donations

kewitz:


server/paymentProviders/stripe/webhook.ts
```diff
@@ -160,7 +160,9 @@ export const paymentIntentSucceeded = async (event: Stripe.Event) => {
   const paymentIntent = event.data.object as Stripe.PaymentIntent;
   const order = await models.Order.findOne({
     where: {
-      status: OrderStatuses.PROCESSING,
+      // Since new CreditCard donations are not synchronously validated, we also need to consider those orders that
+      // go straight from NEW to PAID with payment_intent.succeeded.
+      status: [OrderStatuses.NEW, OrderStatuses.PROCESSING, OrderStatuses.ERROR, OrderStatuses.ACTIVE],
```
znarf, line 165: Why do we even filter on limited list of status? Should we consider not having a filter, or exclude instead?

kewitz: I think it is a good practice to be explicit about what we're waiting for. If we receive an event for an order that is already marked as PAID, that is a problem that we would need to know.

znarf: We're accepting `ACTIVE` here, which is not far to be like `PAID`.

kewitz: True. I'll drop it since we're now clearing data.paymentIntent after processing.

server/paymentProviders/stripe/webhook.ts
```diff
@@ -160,7 +160,6 @@ export const paymentIntentSucceeded = async (event: Stripe.Event) => {
   const paymentIntent = event.data.object as Stripe.PaymentIntent;
   const order = await models.Order.findOne({
     where: {
-      status: OrderStatuses.PROCESSING,
```
Betree, line 163: Check https://github.com/opencollective/opencollective-api/pull/8493/files#diff-6bfae61d9f05877d4f96985e5c6158ed8c842db0969271aba8dc76b0909f1d48

PR #8498, opened 2023-02-02
Scripts to synchronize UploadedFiles

Betree:
For https://github.com/opencollective/opencollective-api/pull/8476
Fix https://github.com/opencollective/opencollective/issues/6389
Resolve https://github.com/opencollective/opencollective/issues/6348

### DIFF
--- a/scripts/uploaded-files/complete-files-info.ts
+++ b/scripts/uploaded-files/complete-files-info.ts
@@ -1,0 +1,83 @@
+#!/usr/bin/env ./node_modules/.bin/babel-node
+
+/**
+ * This script can be use to complete missing information from S3 on the "UploadedFile" table.
+ */
+
+import '../../server/env';
+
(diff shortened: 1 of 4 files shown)


scripts/uploaded-files/complete-files-info.ts
```diff
@@ -0,0 +1,83 @@
...
+        where: { id: file.id, url: file.url },
+        validate: false, // Allow any fileType
+      },
+    );
+  }
+};
+
+const main = async () => {
+  // Please avoid running this script without a `where` condition to avoid updating all files
+  const filesToUpdate = await UploadedFile.findAll({
```
hdiniz, line 57: As per comment, should this be updating all uploaded files?

Betree: The intent is for developers to complete this query based on what they need to update

PR #8433, opened 2023-01-19
PayPal reconciliation CRON job

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6342

To do as follow-up:
- Record refunded transactions
- Reconcile PayPal checkout orders
- Reconcile "Pass payment" (payouts)

PR #8501, opened 2023-02-02
PayPal: Authorize instead of capture 

Betree:
Part of https://github.com/opencollective/opencollective-api/pull/8433
Goes with https://github.com/opencollective/opencollective-frontend/pull/8601

Before this PR, we used to capture the payment straight from the contribution flow. 

This led to a situation where, in rare cases, a PayPal request failing later in the process would end up with the order not being created while the payment had been triggered.

The process is now:
1. (Frontend) Authorize the payment on PayPal
2. (Frontend) Call the API with the PayPal OrderId
3. (API) Get the PayPal Order details
4. (API) Confirm the authorization on PayPal order
5. (API) Trigger the capture on authorization
6. (API) Get capture details
7. (API) Record transition

Direct consequences of this PR:
- Payments with PayPal will be slower (now making four queries instead of 2)
- Successful payment cannot end up without an order anymore

### DIFF
--- a/server/paymentProviders/paypal/payment.ts
+++ b/server/paymentProviders/paypal/payment.ts
@@ -1,5 +1,6 @@
 import { get, isUndefined, pickBy, truncate } from 'lodash';
 
+import FEATURE from '../../constants/feature';
 import * as constants from '../../constants/transactions';
 import { getFxRate } from '../../lib/currency';
 import { floatAmountToCents } from '../../lib/math';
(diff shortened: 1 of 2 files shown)


Betree:
> we should maybe have a cronjob to capture authorized orders or maybe mark the Order as ERROR if it fa

To mark it as error for sure, the authorization expires after 3 days. I'll create a followup issue.

PR #8509, opened 2023-02-03
Handle fileSize null on FileInfo object

hdiniz:
Resolve https://sentry.io/organizations/open-collective/issues/3915438717/?project=5199682&referrer=slack

### DIFF
--- a/server/graphql/v2/interface/FileInfo.ts
+++ b/server/graphql/v2/interface/FileInfo.ts
@@ -26,7 +26,7 @@ export const fileInfoFields = {
   size: {
     type: new GraphQLNonNull(GraphQLInt),
     description: 'Size of the file in bytes',
-    resolve: file => file.fileSize,
+    resolve: file => file.fileSize || 0,
   },
 };
 


SudharakaP (approved):

server/graphql/v2/interface/FileInfo.ts
```diff
@@ -26,7 +26,7 @@ export const fileInfoFields = {
   size: {
     type: new GraphQLNonNull(GraphQLInt),
     description: 'Size of the file in bytes',
-    resolve: file => file.fileSize,
+    resolve: file => file.fileSize || 0,
```
SudharakaP, line 29: Do we check this on the front-end (and not show the file size) or just show 0kb on the front-end after this fix?

Betree: It will show 0. Not ideal, but it's an acceptable fix for now. We'll iterate with https://github.com/opencollective/opencollective-api/pull/8498 to fetch all file sizes from S3, ultimately they'll be all non-null

PR #8518, opened 2023-02-06
Add bancontact payment method type handling

hdiniz:

Bancontact is a bank redirect method.
For one time charges the payment is handled by the Stripe Payment Element flow.
For recurring charges, a SEPA Debit is generated on the first successful charge and reused in following payments by this provider.

### DIFF
--- a/server/paymentProviders/stripe/bancontact.ts
+++ b/server/paymentProviders/stripe/bancontact.ts
@@ -1,0 +1,92 @@
+import config from 'config';
+import { pick, toUpper } from 'lodash';
+import type Stripe from 'stripe';
+
+import OrderStatuses from '../../constants/order_status';
+import logger from '../../lib/logger';
+import { getApplicationFee } from '../../lib/payments';
+import { reportMessageToSentry } from '../../lib/sentry';
+import stripe, { convertToStripeAmount } from '../../lib/stripe';
+import models from '../../models';
+
+import { APPLICATION_FEE_INCOMPATIBLE_CURRENCIES, refundTransaction, refundTransactionOnlyInDatabase } from './common';
+
+const processOrder = async (order: typeof models.Order): Promise<void> => {
+  const generatedSepaDebit = order.paymentMethod?.data?.generated_sepa_debit;
+  if (!generatedSepaDebit) {
+    throw new Error(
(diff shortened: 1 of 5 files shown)


server/paymentProviders/stripe/common.ts
```diff
@@ -369,3 +372,110 @@ export async function getOrCloneCardPaymentMethod(
...
+    originOrder?: typeof models.Order;
+    extraData?: object;
+  },
+  createOptions?: CreateOptions,
+): Promise<typeof PaymentMethod> {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const paymentIntentCharges = (originPaymentIntent as any)?.charges?.data;
+  const paymentIntentCharge = <Stripe.Charge>(
+    (paymentIntentCharges && paymentIntentCharges.length >= 1 && paymentIntentCharges[0])
+  );
```
kewitz, line 452: I think this could be further simplified.
```suggestion
  const paymentIntentCharge: Stripe.Charge = (originPaymentIntent as any)?.charges?.data?.paymentIntentCharges?.[0];
```

Also, I believe that there are no guarantees this method is always called with a payment intent that has a charge. You will need to add a conditional if `paymentMethodChargeDetails` is required.

server/paymentProviders/stripe/bancontact.ts
```diff
@@ -0,0 +1,92 @@
...
+  const hostStripeAccount = await order.collective.getHostStripeAccount();
+  const host = await order.collective.getHostCollective();
+  const isPlatformRevenueDirectlyCollected = APPLICATION_FEE_INCOMPATIBLE_CURRENCIES.includes(toUpper(host.currency))
+    ? false
+    : host?.settings?.isPlatformRevenueDirectlyCollected ?? true;
+  const applicationFee = await getApplicationFee(order, host);
+
+  const paymentIntentParams: Stripe.PaymentIntentCreateParams = {
+    customer: order.paymentMethod.customerId,
+    currency: order.currency,
```
kewitz, line 31: Should we assert currency is `EUR`?

kewitz (approved):
LGTM. Please make sure this works in a clean environment before merging.

PR #8524, opened 2023-02-07
Trash all S3 files that are not in the Database

Betree:
Require https://github.com/opencollective/opencollective-api/pull/8498
Resolve https://github.com/opencollective/opencollective/issues/6349
Partially addresses https://github.com/opencollective/opencollective/issues/6350

### DIFF
--- a/scripts/uploaded-files/trash-legacy-files-in-s3.ts
+++ b/scripts/uploaded-files/trash-legacy-files-in-s3.ts
@@ -1,0 +1,132 @@
+#!/usr/bin/env ./node_modules/.bin/babel-node
+
+/**
+ * This script assumes that all files still being used have been recorded to the `UploadedFile` table (using `scripts/uploaded-files/record-existing-files.ts`).
+ * It will move all files that are in S3 but not in the database to a trash folder by prepending `trash/` to the key.
+ */
+
+import '../../server/env';
+
+import fs from 'fs';
+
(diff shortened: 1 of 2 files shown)


Betree:
@hdiniz This is ready for review, I just ran it on a couple of files to make sure it was working in production but waiting for your review to confirm that this is the right strategy.

scripts/uploaded-files/trash-legacy-files-in-s3.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  const nonTrashedObjects = objects.filter(o => !o.Key.startsWith(S3_TRASH_PREFIX));
+  logger.info(`${message} (${nonTrashedObjects.length} non-trashed files)`);
+  return objects;
+};
+
+const main = async options => {
+  const concurrency = parseInt(options['concurrency']) || 3;
+  const allObjects = await getNonTrashedFilesFromS3();
```
hdiniz, line 46: Around how many objects are we handling here? Does it work fine as an input to the sql query below?

Betree: ~240.000 objects. It does work fine, with the query below, mainly because we're taking advantage of the index by doing a strict equal comparison.

hdiniz (approved):

Betree:
The script ran successfully!

PR #8527, opened 2023-02-08
Stripe payment element e2e

hdiniz:


.github/workflows/e2e.yml
```diff
@@ -53,6 +55,11 @@ jobs:
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
 
     steps:
+      - name: Add stripe apt sources
+        run: |
+          curl -s https://packages.stripe.dev/api/security/keypair/stripe-cli-gpg/public | gpg --dearmor | sudo tee /usr/share/keyrings/stripe.gpg
+          echo "deb [signed-by=/usr/share/keyrings/stripe.gpg] https://packages.stripe.dev/stripe-cli-debian-local stable main" | sudo tee -a /etc/apt/sources.list.d/stripe.list
+
```
kewitz, line 62: Do we need to install this both on API and Frontend?

kewitz (changes requested):
1. This effectively turns our E2E test into integration tests.
2. Install Stripe CLI only on one repo if possible.

hdiniz:
[About the tests and integration with Stripe.
](https://github.com/opencollective/opencollective-frontend/pull/8616#issuecomment-1423377823)

.github/workflows/e2e.yml
hdiniz: Yes, they both have a dependency on the CLI to run the test suite.

kewitz (approved):

PR #8529, opened 2023-02-09
Make sure the `amount` is not a required input for tier creation

SudharakaP:
Fix creation of tickets as per discussion at https://opencollective.slack.com/archives/GFH4N961L/p1675891486724949?thread_ts=1675101689.930789&cid=GFH4N961L. This fixes some issues with creating tickets. In particular we don't need `amount` input to be mandatory when creating tickets. 

[tier-ticket-creation.webm](https://user-images.githubusercontent.com/12435965/217690981-df61581c-eafe-4538-a835-5c43565ea630.webm)

### DIFF
--- a/server/graphql/v2/input/TierCreateInput.ts
+++ b/server/graphql/v2/input/TierCreateInput.ts
@@ -17,7 +17,7 @@ export const TierCreateInput = new GraphQLInputObjectType({
   name: 'TierCreateInput',
   fields: () => ({
     amount: {
-      type: new GraphQLNonNull(AmountInput),
+      type: AmountInput,
     },
     name: {
       type: GraphQLNonEmptyString,
--- a/server/graphql/v2/input/TierCreateInput.ts
+++ b/server/graphql/v2/input/TierCreateInput.ts
@@ -8,10 +8,28 @@ import {
 } from 'graphql';
 import { GraphQLNonEmptyString } from 'graphql-scalars';
 
-import { TierAmountType, TierType } from '../enum';
-import { TierFrequency } from '../enum/TierFrequency';
+import { TierType } from '../../../models/Tier';
+import { TierAmountType, TierType as GraphQLTierType } from '../enum';
+import { TierFrequency, TierFrequencyKey } from '../enum/TierFrequency';
 
-import { AmountInput } from './AmountInput';
+import { AmountInput, AmountInputType } from './AmountInput';
+
+export type TierCreateInputFields = {
+  amount?: AmountInputType;
+  name?: string;
+  description?: string;
+  button?: string;
+  goal?: AmountInputType;
+  type: TierType;
+  amountType: 'FLEXIBLE' | 'FIXED';
+  frequency: TierFrequencyKey;
+  presets?: number[];
+  maxQuantity?: number;
+  minimumAmount?: AmountInputType;
+  useStandalonePage?: boolean;
+  invoiceTemplate?: string;
+  singleTicket?: boolean;
+};
 
 export const TierCreateInput = new GraphQLInputObjectType({
   name: 'TierCreateInput',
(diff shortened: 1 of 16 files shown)


server/graphql/v2/mutation/TierMutations.js
```diff
@@ -130,10 +132,12 @@ const tierMutations = {
         tierCreateData.data.invoiceTemplate = args.tier.invoiceTemplate;
       }
 
+      const tier = await models.Tier.create(tierCreateData);
+
```
SudharakaP, line 136: @Betree : The reason I did this change was that I noticed that otherwise when the tier creation or tier edit modal is saved you need to do a page refresh sometimes to notice the changes. This is because previously we were doing the update after the cache is cleared.

server/graphql/v2/mutation/TierMutations.js
```diff
@@ -73,11 +73,13 @@ const tierMutations = {
...
       }
 
+      const updatedTier = await tier.update(tierUpdateData);
+
       // Purge cache
       purgeCacheForCollective(collective.slug);
       purgeCacheForPage(`/${collective.slug}/contribute/${tier.slug}-${tier.id}`);
 
-      return await tier.update(tierUpdateData);
+      return updatedTier;
```
SudharakaP, line 82: The reason I did this change was that I noticed that otherwise when the tier creation or tier edit modal is saved you need to do a page refresh sometimes to notice the changes. This is because previously we were doing the update after the cache is cleared.

Betree (commented):
Lot of duplicate logic between `createTier` and `editTier` (not coming from you). I'll use your PR to refactor this and add some validations.

server/graphql/v2/input/TierCreateInput.ts
```diff
@@ -8,19 +8,37 @@ import {
...
 export const TierCreateInput = new GraphQLInputObjectType({
   name: 'TierCreateInput',
   fields: () => ({
     amount: {
-      type: new GraphQLNonNull(AmountInput),
+      type: AmountInput,
     },
     name: {
-      type: GraphQLNonEmptyString,
+      type: new GraphQLNonNull(GraphQLNonEmptyString),
```
Betree, line 41: Related: https://github.com/opencollective/opencollective-frontend/pull/8622

server/graphql/v2/input/TierUpdateInput.ts
```diff
@@ -36,13 +39,13 @@ export const TierUpdateInput = new GraphQLInputObjectType({
...
-      type: new GraphQLNonNull(TierType),
+      type: TierType,
     },
     amountType: {
-      type: new GraphQLNonNull(TierAmountType),
+      type: TierAmountType,
     },
     frequency: {
-      type: new GraphQLNonNull(TierFrequency),
+      type: TierFrequency,
```
Betree, line 48: Before this PR, partial updates were supported with some fields but not others. This change makes them officially supported, it should be valid to call a mutation like:

```es6
editTier(tier: { id: 'xxx', name: 'New name')
```

to update just one attribute.

server/graphql/v2/mutation/TierMutations.ts
```diff
@@ -152,6 +147,10 @@ const tierMutations = {
       checkRemoteUserCanUseAccount(req);
 
       const tier = await fetchTierWithReference(args.tier, { throwIfMissing: true });
+      if (tier === 'custom') {
+        throw new Error('Cannot delete custom tier. Set settings.disableCustomContributions to true instead.');
+      }
+
```
Betree, line 153: Had to do this because TS (rightfully) complained that `req.loaders.Collective.byId.load('custom')` was not allowed.

server/models/Tier.ts
```diff
@@ -465,7 +465,12 @@ Tier.init(
         }
       },
       validateFlexibleAmount() {
-        if (this.amountType === 'FLEXIBLE' && this.presets && !this.presets.includes(this.amount)) {
+        if (
+          !isNil(this.amount) &&
+          this.amountType === 'FLEXIBLE' &&
+          this.presets &&
+          !this.presets.includes(this.amount)
+        ) {
```
Betree, line 473: @SudharakaP  :+1:

PR #8530, opened 2023-02-09
Fix manually paid Expenses on host and collective with different currencies

kewitz:
- [x] Write a test to reproduce the error.
- [x] Fix the bug.
- [x] Manually fix affected transactions


Betree (approved):
Lgtm!

PR #8542, opened 2023-02-13
Fast Balance: transactions should be ordered by date and not by id

znarf:
We had a problem reported, related to inconsistency in how we order Transactions.

To accommodate how we use the ledger currently, we should assume transactions are ordered by date and not by id.

To be looked at, how does these changes affect speed and indexes?

`CurrentCollectiveTransactionStats` seems also impacted by a similar problem.

### DIFF
--- a/migrations/20230213080003-fast-balance-update.js
+++ b/migrations/20230213080003-fast-balance-update.js
@@ -1,0 +1,132 @@
+'use strict';
+
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`DROP VIEW IF EXISTS "CurrentCollectiveBalance"`);
+
+    await queryInterface.sequelize.query(`DROP MATERIALIZED VIEW IF EXISTS "CollectiveBalanceCheckpoint"`);
+
+    await queryInterface.sequelize.query(`
+      CREATE MATERIALIZED VIEW IF NOT EXISTS "CollectiveBalanceCheckpoint" AS (
+        WITH "LatestTransactions" AS (
+          SELECT "CollectiveId", MAX("createdAt") AS "createdAt" FROM "TransactionBalances" GROUP BY "CollectiveId"
+        )
+        SELECT *
+        FROM "TransactionBalances"
+        WHERE id IN (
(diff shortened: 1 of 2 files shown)


migrations/20230213080003-fast-balance-update.js
```diff
@@ -0,0 +1,132 @@
...
+        LEFT JOIN LATERAL (
+          SELECT
+            SUM(t."amountInHostCurrency") +
+              SUM(coalesce(t."platformFeeInHostCurrency", 0)) +
+              SUM(coalesce(t."hostFeeInHostCurrency", 0)) +
+              SUM(coalesce(t."paymentProcessorFeeInHostCurrency", 0)) +
+              SUM(coalesce(t."taxAmount" * t."hostCurrencyFxRate", 0)) "netAmountInHostCurrency"
+          FROM "Transactions" t
+          WHERE t."CollectiveId" = cbc."CollectiveId"
+            AND t."createdAt" > cbc."createdAt"
```
hdiniz, line 46: This is using the index `transactions__collective_id_created_at_type` :+1:

migrations/20230213080003-fast-balance-update.js
```diff
@@ -0,0 +1,132 @@
...
+          INNER JOIN "LatestTransactions" lt ON lt."CollectiveId" = tb."CollectiveId" AND lt."createdAt" = tb."createdAt"
+          GROUP BY tb."CollectiveId"
+        )
+       )
+    `);
+
+    await queryInterface.sequelize.query(`
+      CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS "collective_balance_checkpoint__collective_id"
+      ON "CollectiveBalanceCheckpoint"("CollectiveId")
+    `);
```
hdiniz, line 27: This index could also be added to `CollectiveTransactionStats(id)` on the other migration.

hdiniz (approved):

migrations/20230213080003-fast-balance-update.js
znarf: Good catch, we used to have it, then lost it in a former migration, had to add it again.

We were about to loose it again.

PR #8550, opened 2023-02-14
Add euro host for e2e tests

hdiniz:
Host configuration to run e2e tests against.


hdiniz:
> I would rather update our dev-data dump than merging a migration for this.

You mean applying this to the development environment also? Right now its only on the tests

kewitz:
@hdiniz yes, we use the same dump for both dev and tests. I'm suggesting things would be simpler if we just updated the dump instead of creating a migration that only runs on the test environment.

hdiniz:
@kewitz I have updated the dev dump to include:
- a EUR currency host, this PR
- A US currency host from a previously merged PR

Betree:
The initial approach @hdiniz took to write a migration was the right one: dev dumps are meant to be regenerated, not incrementally updated. After exporting a dump for Postgres 14, the E2E tests broke because the host payment method was missing in the test data. Having that in a migration prevents the issue, and ensures that our dumps are always generated with the same seed data.

On top of that, migrations make easily searchable documentation, which helps to investigate issues like this one. I'll bring it back in https://github.com/opencollective/opencollective-api/pull/10528.

PR #8551, opened 2023-02-14
Make sure we translate stripe payment method types into ours

kewitz:
Fixes https://open-collective.sentry.io/issues/3796028632/?project=5199682

### DIFF
--- a/server/paymentProviders/stripe/webhook.ts
+++ b/server/paymentProviders/stripe/webhook.ts
@@ -31,6 +31,22 @@ import * as virtualcard from './virtual-cards';
 
 const debug = debugLib('stripe');
 
+const coercePaymentMethodType = (paymentMethodType: Stripe.PaymentMethod.Type): PAYMENT_METHOD_TYPE => {
+  switch (paymentMethodType) {
+    case 'card':
+      return PAYMENT_METHOD_TYPE.CREDITCARD;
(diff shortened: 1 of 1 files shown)


server/paymentProviders/stripe/webhook.ts
```diff
@@ -31,6 +31,22 @@ import * as virtualcard from './virtual-cards';
...
+    case 'sepa_debit':
+      return PAYMENT_METHOD_TYPE.SEPA_DEBIT;
+    case 'bacs_debit':
+      return PAYMENT_METHOD_TYPE.BACS_DEBIT;
+    case 'us_bank_account':
+      return PAYMENT_METHOD_TYPE.US_BANK_ACCOUNT;
+    case 'alipay':
+      return PAYMENT_METHOD_TYPE.ALIPAY;
+    default:
+      return null;
```
hdiniz, line 47: What would be the consequence of setting the type to null?
With the exception of card/creditcard, we are using the same names as Stripe and might continue to do so for new types we implement right?

znarf: Maybe it's sensible to have that as a default.

```
default:
   return paymentMethodType
```

znarf: And add a warning if we're missing the matching constant on our side?

PR #8553, opened 2023-02-15
Resize Uploaded file image before encoding with blurhash

hdiniz:
Resolves https://opencollective.freshdesk.com/a/tickets/167442

Resizes the image before encoding it with blurhash.


server/models/UploadedFile.ts
```diff
@@ -154,8 +155,12 @@ class UploadedFile extends Model<InferAttributes<UploadedFile>, InferCreationAtt
...
       const image = sharp(file.buffer);
       const { width, height } = await image.metadata();
-      const pixels = await image.raw().ensureAlpha().toBuffer({ resolveWithObject: true });
-      const blurHash = encode(pixels.data, width, height, 4, 4);
+      const { data, info } = await image
+        .raw()
+        .ensureAlpha()
+        .resize({ fit: sharp.fit.contain, width: 200 })
+        .toBuffer({ resolveWithObject: true });
+      const blurHash = encode(data, info.width, info.height, 4, 4);
```
znarf, line 163: As extra, to make the code more resilient, can we do a try/catch and have an empty blurHash in case something is wrong when computing it?

znarf (approved):

PR #8558, opened 2023-02-16
Pending Contributions: Order Page and editing Pending Contributions

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6323

### DIFF
--- a/server/graphql/v2/input/OrderCreateInput.ts
+++ b/server/graphql/v2/input/OrderCreateInput.ts
@@ -161,9 +161,17 @@ export const PendingOrderCreateInput = new GraphQLInputObjectType({
       type: GraphQLString,
       description: 'Public order description',
     },
-    customData: {
-      type: GraphQLJSON,
-      description: 'Custom data related to the Pending Order',
+    memo: {
+      type: GraphQLString,
+      description: 'Private memo for the host',
+    },
+    ponumber: {
+      type: GraphQLString,
+      description: 'External identifier for the order',
+    },
+    paymentMethod: {
+      type: GraphQLString,
+      description: 'Payment method expected for this order',
     },
     expectedAt: {
       type: DateString,
@@ -216,9 +224,17 @@ export const PendingOrderEditInput = new GraphQLInputObjectType({
       type: GraphQLString,
       description: 'Public order description',
     },
-    customData: {
-      type: GraphQLJSON,
-      description: 'Custom data related to the Pending Order',
+    memo: {
+      type: GraphQLString,
+      description: 'Private memo for the host',
+    },
+    ponumber: {
+      type: GraphQLString,
+      description: 'External identifier for the order',
+    },
+    paymentMethod: {
+      type: GraphQLString,
+      description: 'Payment method expected for this order',
     },
     expectedAt: {
       type: DateString,
--- a/server/graphql/v2/input/OrderCreateInput.ts
+++ b/server/graphql/v2/input/OrderCreateInput.ts
@@ -141,10 +141,6 @@ export const PendingOrderCreateInput = new GraphQLInputObjectType({
       type: new GraphQLNonNull(AccountReferenceInput),
       description: 'The collective you want to contribute to',
     },
-    childAccount: {
-      type: AccountReferenceInput,
-      description: 'The event/project you want to contribute to',
-    },
     platformTipAmount: {
       type: AmountInput,
       description: 'Platform tip attached to this order',
--- a/server/graphql/v2/input/OrderCreateInput.ts
+++ b/server/graphql/v2/input/OrderCreateInput.ts
@@ -139,7 +139,66 @@ export const PendingOrderCreateInput = new GraphQLInputObjectType({
     },
     toAccount: {
       type: new GraphQLNonNull(AccountReferenceInput),
-      description: 'The collective/event/project you want to contribute to',
+      description: 'The collective you want to contribute to',
+    },
+    childAccount: {
+      type: AccountReferenceInput,
+      description: 'The event/project you want to contribute to',
+    },
+    platformTipAmount: {
+      type: AmountInput,
+      description: 'Platform tip attached to this order',
+    },
+    taxes: {
+      type: new GraphQLList(OrderTaxInput),
+      description: 'Use this field to set the taxes associated to this order',
+    },
+    tier: {
+      type: TierReferenceInput,
+      description: 'The tier you are contributing to',
+    },
+    description: {
+      type: GraphQLString,
+      description: 'Public order description',
+    },
+    customData: {
+      type: GraphQLJSON,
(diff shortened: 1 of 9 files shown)


server/graphql/v2/input/OrderCreateInput.ts
```diff
@@ -139,7 +139,66 @@ export const PendingOrderCreateInput = new GraphQLInputObjectType({
     },
     toAccount: {
       type: new GraphQLNonNull(AccountReferenceInput),
-      description: 'The collective/event/project you want to contribute to',
+      description: 'The collective you want to contribute to',
+    },
+    childAccount: {
+      type: AccountReferenceInput,
+      description: 'The event/project you want to contribute to',
+    },
```
Betree, line 147: I don't get why we need to pass both `toAccount` and `childAccount`, the parent can always be retrieved from `child.getParentCollective()` right?

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -828,6 +832,60 @@ const orderMutations = {
         },
       });
 
+      return order;
+    },
+  },
+  editPendingOrder: {
+    type: new GraphQLNonNull(Order),
+    description: 'To submit a new order. Scope: "orders".',
```
Betree, line 840:
```suggestion
    description: 'To edit a pending order. Scope: "orders".',
```

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -828,6 +832,60 @@ const orderMutations = {
...
+      if (!(await canEdit(req, order))) {
+        throw new ValidationFailed(`Only pending orders can be edited, this one is ${order.status}`);
+      }
+
+      const fromAccount = await fetchAccountWithReference(args.order.fromAccount);
+
+      await order.update({
+        CreatedByUserId: req.remoteUser.id,
+        FromCollectiveId: fromAccount?.id || undefined,
+        totalAmount: args.order.amount.valueInCents,
```
Betree, line 870: We need to keep supporting passing the amount via `amount.value` since the API exposes it

```suggestion
        totalAmount: getValueInCentsFromAmountInput(args.order.amount),
```

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -828,6 +832,60 @@ const orderMutations = {
...
+      }
+
+      if (!(await canEdit(req, order))) {
+        throw new ValidationFailed(`Only pending orders can be edited, this one is ${order.status}`);
+      }
+
+      const fromAccount = await fetchAccountWithReference(args.order.fromAccount);
+
+      await order.update({
+        CreatedByUserId: req.remoteUser.id,
```
Betree, line 868: Why update `CreatedByUserId` here?

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -180,7 +180,15 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
   const searchTermConditions = buildSearchConditions(args.searchTerm, {
     idFields: ['id'],
     slugFields: ['$fromCollective.slug$', '$collective.slug$'],
-    textFields: ['$fromCollective.name$', '$collective.name$', 'description'],
+    textFields: [
+      '$fromCollective.name$',
+      '$collective.name$',
+      'description',
+      'data.ponumber',
+      'data.memo',
```
Betree, line 188: :no_entry_sign: The `memo` is for admins only; we can't allow searching it for everyone.

server/graphql/v2/object/Order.js
```diff
@@ -149,6 +150,13 @@ export const Order = new GraphQLObjectType({
           }
         },
       },
+      hostFeePercent: {
+        type: GraphQLInt,
+        description: 'Host fee percent attached to the Order.',
+        async resolve(order) {
+          return await getHostFeePercent(order);
+        },
+      },
```
Betree, line 159: Maybe a good opportunity to introduce https://github.com/opencollective/opencollective/issues/5606?

server/graphql/v2/object/Order.js
```diff
@@ -235,13 +243,41 @@ export const Order = new GraphQLObjectType({
...
       processedAt: {
         type: GraphQLDateTime,
         description: 'Date the funds were received.',
         async resolve(order) {
           return order?.processedAt;
         },
       },
+      pendingContributionData: {
+        type: GraphQLJSON,
+        description: 'Date the funds should arrive.',
```
Betree, line 268:
```suggestion
        description: 'Data about the pending contirbution',
```

server/graphql/v2/object/Order.js
```diff
@@ -235,13 +243,41 @@ export const Order = new GraphQLObjectType({
...
+      },
       processedAt: {
         type: GraphQLDateTime,
         description: 'Date the funds were received.',
         async resolve(order) {
           return order?.processedAt;
         },
       },
+      pendingContributionData: {
+        type: GraphQLJSON,
```
Betree, line 267: Do we want to type that?

server/graphql/v2/input/OrderCreateInput.ts
kewitz: We don't have to pass both. In fact, you can see that we're using `childAccount` as the `toAccount` if it is passed.
The thing is, since in the frontend it is optional to pick an event/project for the collective, it is harder to manage the state of both values as being `toAccount` in the frontend.

My cheap solution was to just send both as independent values and decided which one to use in the backend.

server/graphql/v2/mutation/TagMutations.ts
```diff
@@ -0,0 +1,70 @@
...
+      expense: {
+        type: ExpenseReferenceInput,
+      },
+    },
+    resolve: async (_: void, args, req: Request) => {
+      if (!req.remoteUser) {
+        throw new Unauthorized();
+      }
+
+      if (args.order) {
```
Betree, line 44: To have a clear pattern in case they provide multiple entities

```suggestion
      if ([args.order, args.expense].filter(Boolean).length !== 1) {
        throw new Error('A single order or expense must be provided');
      }

      if (args.order) {
```

server/graphql/v2/input/OrderCreateInput.ts
Betree: Maybe we can keep storing two values in the frontend, then handle that during the `submit`?

```es6
submit({ variables: { toAccount: childAccount || toAccount })
```

I understand the workaround, but I don't think introducing this complexity in the API is the right thing to do.

kewitz: You're right. I think I ended up overcomplicating all of this. It seems pretty obvious now.

PR #8564, opened 2023-02-17
Add Support to Convert User Account to Org

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/5771

Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/8647

### DIFF
--- a/server/graphql/schemaV2.graphql
+++ b/server/graphql/schemaV2.graphql
@@ -4292,7 +4292,7 @@ type AccountPermissions {
   id: String!
 
   """
-  Whether the current has teh permission to add funds to the account
+  Whether the current user can mark this order as expired
   """
   addFunds: Permission!
 }
--- a/server/graphql/schemaV2.graphql
+++ b/server/graphql/schemaV2.graphql
@@ -13754,6 +13754,16 @@ type Mutation {
   ): Account!
 
   """
+  [Root only] Edits account type from User to Organization
+  """
+  editAccountType(
+    """
+    Account to change the type for
+    """
+    account: AccountReferenceInput!
+  ): Account!
+
+  """
   [Root only] Clears the cache for a given account
   """
   clearCacheForAccount(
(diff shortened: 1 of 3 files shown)


Betree (commented):
I'll let @znarf provide a more comprehensive review as he's more experienced with the process. Some concerns I have:
- This needs unit tests
- What if someone tries to convert a guest account?
- I'm wondering about the `isActive` column value, need to check the DB data
- Note that location is private for users but public for organizations; this should be clearly stated by support in appropriate cases

znarf:
> This needs unit tests

Todo

> What if someone tries to convert a guest account?

This should be prevented until we have such case and sure we want to support it.

SudharakaP:
@Betree : 

> I'll let @znarf provide a more comprehensive review as he's more experienced with the process. Some concerns I have:
> 
>     * This needs unit tests

Added. 👍🏽 

>     * What if someone tries to convert a guest account?

For now I've prevented that. 👍🏽 
 
>     * I'm wondering about the `isActive` column value, need to check the DB data

It seems that when we are creating a new user we set `isActive=false`.

https://github.com/opencollective/opencollective-api/blob/1698e37291ae2f0e21b1ab4a58958059c414bf12/server/models/User.ts#L414

In production we only have 3 users who has `isActive=true`. Not sure exactly how these users have this value set to true, but wondering how this is relevant to this situation? 🤔  
 
>     * Note that location is private for users but public for organizations; this should be clearly stated by support in appropriate cases

I've added a note in the front-end so that the person who's doing this is reminded of this and could notify the user.

image

znarf (approved):
Works for me. Maybe Ben for a final look but not mandatory.

SudharakaP:
> Works for me. Maybe Ben for a final look but not mandatory.

@Betree : You wanna look at this one more, or are we good to merge? 😸

Betree (commented):
No blocker for me

PR #8565, opened 2023-02-17
Always create tier with collective currency

hdiniz:
Enforce tier currency to match collective currency on creation, test that edit tier does not update tier currency.

### DIFF
--- a/migrations/20230217131657-fix-tier-currencies-different-from-collective.js
+++ b/migrations/20230217131657-fix-tier-currencies-different-from-collective.js
@@ -1,0 +1,21 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    queryInterface.sequelize.query(`
+      UPDATE "Tiers"
+      SET currency = "affectedTier"."collectiveCurrency"
+      FROM (
+        select t.id as "TierId", c.currency as "collectiveCurrency"
+        from "Tiers" t inner join "Collectives" c on c.id = t."CollectiveId" 
+        where t.currency <> c.currency and t."deletedAt" is null and t."createdAt" >= '2023-01-01'
+      ) AS "affectedTier"
+      WHERE "Tiers".id = "affectedTier"."TierId";
+    `);
+  },
+
+  async down() {
+    // no rollback
+  },
+};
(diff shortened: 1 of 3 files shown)


server/graphql/v2/mutation/TierMutations.ts
```diff
@@ -123,7 +123,11 @@ const tierMutations = {
       await twoFactorAuthLib.enforceForAccountAdmins(req, account, { onlyAskOnLogin: true });
 
       // Create tier
-      const tier = await TierModel.create({ ...transformTierInputToAttributes(args.tier), CollectiveId: account.id });
+      const tier = await TierModel.create({
+        ...transformTierInputToAttributes(args.tier),
+        CollectiveId: account.id,
+        currency: account.currency,
```
znarf, line 129: At some point in the future, it's possible we'll let Tier be defined in a different currency.

It's already supported by the API.

https://github.com/opencollective/opencollective/issues/3891

znarf: Just a FYI, we don't have to make it work here now.

znarf:
We need a migration, maybe not trying to fix all occurrences as they are some legit use cases for this in the DB, but only the recently created one at least?

migrations/20230217131657-fix-tier-currencies-different-from-collective.js
```diff
@@ -0,0 +1,21 @@
...
+      UPDATE "Tiers"
+      SET currency = "affectedTier"."collectiveCurrency"
+      FROM (
+        select t.id as "TierId", c.currency as "collectiveCurrency"
+        from "Tiers" t inner join "Collectives" c on c.id = t."CollectiveId" 
+        where t.currency <> c.currency and t."deletedAt" is null and t."createdAt" >= '2023-01-01'
+      ) AS "affectedTier"
+      WHERE "Tiers".id = "affectedTier"."TierId";
+    `);
+  },
```
hdiniz, line 16: @znarf 

```
select t."createdAt", t.id, t.name, t.amount, t."amountType", t."minimumAmount", t.presets, t.type, t.currency as "tierCurrency", c.currency as "collectiveCurrency", c.slug from "Tiers" t inner join "Collectives" c on c.id = t."CollectiveId" where t.currency <> c.currency and t."deletedAt" is null and t."createdAt" >= '2023-01-01';
...
(45 rows)

BEGIN
UPDATE "Tiers"
      SET currency = "affectedTier"."collectiveCurrency"
      FROM (
        select t.id as "TierId", c.currency as "collectiveCurrency"
        from "Tiers" t inner join "Collectives" c on c.id = t."CollectiveId"
        where t.currency <> c.currency and t."deletedAt" is null and t."createdAt" >= '2023-01-01'
      ) AS "affectedTier"
      WHERE "Tiers".id = "affectedTier"."TierId";
UPDATE 45
ROLLBACK
```

Betree (approved):
Will resolve https://github.com/opencollective/opencollective/issues/6428. Thanks for taking care of this!

PR #8589, opened 2023-02-23
Mark unused files as deleted in the DB

Betree:
Combined with https://github.com/opencollective/opencollective-api/pull/8524, this will resolve https://github.com/opencollective/opencollective/issues/6350, resolving the last item in https://github.com/opencollective/opencollective/issues/6350.

I'm still trying to figure out the best query, which is challenging regarding performance.

The main  thing I'm trying to prevent is an attack where:
- Bob copies the S3 link from Alice's update to his own update
- Bob deleted his update
- Files gets deleted from Bob's update, and therefore from Alice's update too

### DIFF
--- a/scripts/uploaded-files/record-existing-files.ts
+++ b/scripts/uploaded-files/record-existing-files.ts
@@ -8,77 +8,17 @@
 import '../../server/env';
 
 import { Command } from 'commander';
+import { pickBy } from 'lodash';
 import PQueue from 'p-queue';
 
 import { getFileInfoFromS3 } from '../../server/lib/awsS3';
 import logger from '../../server/lib/logger';
+import { FileFieldsDefinition } from '../../server/lib/uploaded-files';
 import { parseToBoolean } from '../../server/lib/utils';
 import models, { sequelize } from '../../server/models';
 import UploadedFile, { FileKind, SUPPORTED_FILE_TYPES } from '../../server/models/UploadedFile';
 
 const DRY_RUN = process.env.DRY_RUN ? parseToBoolean(process.env.DRY_RUN) : true;
-
-const SIMPLE_FIELDS: Array<{
-  model;
-  urlField: string;
-  UserIdField?: string;
-  kind: FileKind;
-}> = [
-  {
-    model: models.Collective,
-    urlField: 'image',
-    kind: 'ACCOUNT_AVATAR',
-  },
-  {
-    model: models.Collective,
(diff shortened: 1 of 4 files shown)


Betree:
If we're happy with these 2 scripts, we should ultimately move them to CRON jobs.

scripts/uploaded-files/record-existing-files.ts
```diff
@@ -8,78 +8,18 @@
...
 import { getFileInfoFromS3 } from '../../server/lib/awsS3';
 import logger from '../../server/lib/logger';
+import { FileFieldsDefinition } from '../../server/lib/uploaded-files';
 import { parseToBoolean } from '../../server/lib/utils';
 import models, { sequelize } from '../../server/models';
 import UploadedFile, { FileKind, SUPPORTED_FILE_TYPES } from '../../server/models/UploadedFile';
 
 const DRY_RUN = process.env.DRY_RUN ? parseToBoolean(process.env.DRY_RUN) : true;
 
-const SIMPLE_FIELDS: Array<{
```
Betree, line 21: Editing this file to extract the fields definitions to re-use them in different places

scripts/uploaded-files/record-existing-files.ts
```diff
@@ -122,30 +62,56 @@ const recordFile = async (
...
+        return `"${name}"`;
+      } else if (index === splitName.length - 1) {
+        return `->>'${name}'`;
+      } else {
+        return `->'${name}'`;
+      }
+    })
+    .join('');
+};
+
```
hdiniz, line 86: I guess we dont need to worry too much about escaping here, since its input is hard coded from the fields constants right?

The `column#>{path}` operator is useful for json path access, simplifying it from [sequelize](https://github.com/sequelize/sequelize/blob/f6f8ba759271c8eff4acdcabcbb44a2d82b00419/packages/core/src/dialects/postgres/query-generator.js#L982):

```
const prepareFieldNameForSQL = (fieldName: string): string => {
  const [column, ...path] = _.toPath(fieldName);

  return `${column}#>>{${path.join(',')}}`;
};
```

hdiniz: On second thought, for future reference maybe it deserves a comment about not escaping input to avoid copying it to another context.

Betree: Good call

PR #8596, opened 2023-02-24
Convert Update model to Typescript

Betree:
Part of https://github.com/opencollective/opencollective/issues/5893


server/models/Update.ts
```diff
@@ -0,0 +1,538 @@
...
+  };
+
+  // Unpublish update
+  unpublish = async function (remoteUser) {
+    return this.update({ LastEditedByUserId: remoteUser.id, publishedAt: null });
+  };
+
+  delete = async function (remoteUser) {
+    await models.Comment.destroy({ where: { UpdateId: this.id } });
+    await models.Update.update(
```
hdiniz, line 145: Should this be `this.update` instead?

server/models/Update.ts
```diff
@@ -0,0 +1,538 @@
...
+      { deletedAt: new Date(), LastEditedByUserId: remoteUser.id },
+      { where: { id: this.id } },
+    );
+
+    return this;
+  };
+
+  // Returns the User model of the User that created this Update
+  static getUser = function () {
```
hdiniz, line 154: This should not be a static method

server/models/Update.ts
Betree: No, it's a `Model.update(attributes, { where: ... })`. We could write it as:

```es6
update = await models.find(...);
await update.destroy();
```

But there wouldn't be much added value, and I don't want to dig into this kind of changes here, I want to keep the diff minimal to prevent introducing side-effects

server/models/Update.ts
Betree: Work in progress :-)

hdiniz:
LGTM.

Its a shame that github does not display the side by side diff correctly in the PR, it works only when looking at the commit.

hdiniz (approved):

PR #8603, opened 2023-02-27
Fix Pending Contribution email notifications

kewitz:
Fixes a small bug where the collective name is not displayed in the email.
Also includes a button linking to the order page.


Betree (approved):
The fix works great!

PR #8608, opened 2023-02-28
Add Support for `COLLECTIVE_MEMBER_CREATED` activity to Send Webhook Notifications

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6482

image

### DIFF
--- a/server/lib/activities.js
+++ b/server/lib/activities.js
@@ -96,6 +96,16 @@ const doFormatMessage = (activity, format) => {
     update = linkify(format, activity.data.url, 'update');
   }
 
+  // get member data
+  let member;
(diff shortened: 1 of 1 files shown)


Betree (approved):
I didn't test but the code looks good!

josesimoes:
Thank you for the quick fix. However I have a concern: this hook used to generate a more meaningful message which included the contribution amount. This is something that's pertinent to the announcement of a new contributor joining the open collective.
If there is no interface to customize the message, I suggest this is added to the default message.

For reference this is how it used to look like:

image

Which later evolved to:

image

Any chance you can improve this? Thanks!

PR #8609, opened 2023-02-28
Add support for update create/publish webhook events

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6483

image

### DIFF
--- a/server/lib/activities.js
+++ b/server/lib/activities.js
@@ -90,6 +90,12 @@ const doFormatMessage = (activity, format) => {
     );
   }
 
+  // get update data
+  let update;
+  if (activity.data.update) {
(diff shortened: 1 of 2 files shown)


server/models/Update.js
```diff
@@ -265,6 +267,8 @@ const Update = sequelize.define(
           // TODO(InconsistentActivities): Should have HostCollectiveId
           data: {
             update: instance.activity,
+            collective,
+            fromCollective,
```
Betree, line 271:
```suggestion
            collective: collective.activity,
            fromCollective: fromCollective.activity,
```

Betree (commented):
LGTM, small improvement to make in the way we record the activity below

PR #8610, opened 2023-02-28
Fix two OrderPage edge-conditions

kewitz:


server/graphql/v2/object/Order.js
```diff
@@ -184,7 +192,7 @@ export const Order = new GraphQLObjectType({
         },
       },
       hostFeePercent: {
-        type: GraphQLInt,
+        type: GraphQLFloat,
```
znarf, line 195: Could you go extra mile and fix all occurrences of `hostFeePercent` to `GraphQLFloat`

PR #8613, opened 2023-02-28
Add collectiveContextOnly arg to order.transactions resolver

kewitz:


server/graphql/v2/object/Order.js
```diff
@@ -155,8 +155,20 @@ export const Order = new GraphQLObjectType({
       transactions: {
         description: 'Transactions for this order ordered by createdAt ASC',
         type: new GraphQLNonNull(new GraphQLList(Transaction)),
-        resolve(order, _, req) {
-          return req.loaders.Transaction.byOrderId.load(order.id);
+        args: {
+          collectiveContextOnly: {
```
znarf, line 159: That variable name looks cryptic. Anything else we could use?

server/graphql/v2/object/Order.js
```diff
@@ -155,8 +155,20 @@ export const Order = new GraphQLObjectType({
...
+            type: GraphQLBoolean,
+            description:
+              'Only return transactions that are related to the collective context of the order. (Transaction.CollectiveId = Order.CollectiveId)',
```
znarf, line 162: Description doesn't look good for API developers. They don't need to know about Model implementation details.

PR #8626, opened 2023-03-02
Migrate location data to new Locations table

gustavlrsn:
Resolve 
- https://github.com/opencollective/opencollective/issues/3494

Requires 
- https://github.com/opencollective/opencollective-frontend/pull/8719
   -  This is required since the frontend currently expects to have a location object on Collectives/events in many places, even if it is empty. This PR will instead return a `null` location if it is not set.


package.json
```diff
@@ -44,6 +44,7 @@
     "@sentry/node": "7.41.0",
     "@sentry/tracing": "7.41.0",
     "@sentry/types": "7.41.0",
+    "@shopify/address": "4.0.6",
```
gustavlrsn, line 47: 1.8kB minified (is using a request to Shopify API to get formatting data), same library as used in frontend. Used to create a formatted address string that is correct for the country from structured data.

server/graphql/common/expenses.ts
```diff
@@ -1073,18 +1075,24 @@ export async function createExpense(remoteUser: User | null, expenseData: Expens
...
+    (expenseData.payeeLocation?.address || expenseData.payeeLocation?.structured) &&
+    (!existingLocation.address || !existingLocation.structured)
   ) {
-    // Let's take the opportunity to update collective's location
-    await fromCollective.update({
-      address: expenseData.payeeLocation.address,
-      countryISO: expenseData.payeeLocation.country,
-      data: { ...fromCollective.data, address: expenseData.payeeLocation.structured },
-    });
+    await fromCollective.setLocation(expenseData.payeeLocation);
```
gustavlrsn, line 1085: Perhaps we should only update Collective locations from payee location for `USER`s, since other types have a public location (in gql v2, v1 slightly more restrictive), might be unexpected to indirectly expose the `payeeLocation` that is private by default.

server/graphql/common/expenses.ts
```diff
@@ -1301,11 +1309,12 @@ export async function editExpense(req: express.Request, expenseData: ExpenseData
...
-    await fromCollective.update({
-      address: expenseData.payeeLocation.address,
-      countryISO: expenseData.payeeLocation.country,
-    });
+  const existingLocation = await fromCollective.getLocation();
+  if (
+    (expenseData.payeeLocation?.address || expenseData.payeeLocation?.structured) &&
+    (!existingLocation.address || !existingLocation.structured)
+  ) {
+    await fromCollective.setLocation(expenseData.payeeLocation);
```
gustavlrsn, line 1317: Same thing here, only update for USERs?

server/graphql/v2/interface/Account.js
```diff
@@ -393,6 +393,29 @@ const accountFieldsDefinition = () => ({
...
+          return null;
+        }
+        const mainProfile = await req.loaders.Collective.mainProfileFromIncognito.load(collective.id);
+        if (mainProfile) {
+          return mainProfile.getLocation();
+        }
+      }
+
+      return collective.getLocation();
+    },
```
gustavlrsn, line 418: This is moved from `../graphql/v2/object/Individual.js` to easier be able to see permissions for the `location` field in one place.

Worth noting is that there currently is a disrepancy between GQL v1 and v2 for what Collective types they expose location, v1 exposes for `COLLECTIVE`, `EVENT`, `ORGANIZATION`, and v2 for all types except `USERS`.

server/graphql/v2/interface/Account.js
```diff
@@ -393,6 +393,29 @@ const accountFieldsDefinition = () => ({
   location: {
     type: Location,
     description: 'The address associated to this account. This field is always public for collectives and events.',
+    async resolve(collective, _, req) {
+      const canSeeLocation =
+        collective.type !== CollectiveTypes.USER ||
+        (req.remoteUser?.isAdmin(collective.id) && checkScope(req, 'account')) ||
+        (await collective.isHost());
```
gustavlrsn, line 400: These permissions are moved and adapted from Individual.js. The `isHost()` call seemed unintuitive to me, but looks like it makes sure to expose location for USER's that are also hosts, will check if this actually happens..

Worth noting is that there currently is a discrepancy between GQL v1 and v2 for what Collective types they expose location, v1 exposes for `COLLECTIVE`, `EVENT`, `ORGANIZATION`, and v2 for all types except `USERS`.

server/models/index.ts
```diff
@@ -330,6 +332,10 @@ models.Tier.hasMany(models.Order);
 // Legal documents
 models.LegalDocument.belongsTo(models.Collective, { foreignKey: 'CollectiveId', as: 'collective' });
 
+// Location
+models.Location.belongsTo(models.Collective, { foreignKey: 'CollectiveId', as: 'collective' });
+models.Collective.hasOne(models.Location, { foreignKey: 'CollectiveId', as: 'location' });
```
Betree, line 337: Is the intent to move that to `hasMany` in the future?

migrations/20230307121010-add-locations-table.js
```diff
@@ -0,0 +1,78 @@
...
+        allowNull: true,
+      },
+      structured: {
+        type: DataTypes.JSONB,
+        allowNull: true,
+      },
+      geoLocationLatLong: {
+        type: DataTypes.JSONB,
+        allowNull: true,
+      },
```
Betree, line 32: I haven't benchmarked, but intuitively I would say that querying this as JSONB will be much slower than querying two (lat: FLOAT, long: FLOAT) columns or a Postgres `Point` (as suggested in https://github.com/opencollective/opencollective/issues/3494)

gustavlrsn: Makes sense. Got stuck on some Typescript issues I were not able to resolve quickly (extending DataTypes to include `POINT`) but might consider it a future enhancement, to be able to focus on some other projects this cycle, added this in a separate PR here: https://github.com/opencollective/opencollective-api/pull/8704.

Betree (commented):
Nice job, that's not an easy refactor.

A few concerns below that I think need some attention; let me know if you prefer to pair to look at them.

server/lib/format-address.ts
```diff
@@ -0,0 +1,52 @@
...
+        province: zone,
+        zip: postalCode,
+        ...(includeCountry && {
+          country,
+        }),
+      },
+      formattingCountry,
+    );
+  } catch (error) {
+    console.log('Error formatting address using @shopify/address:', error.message);
```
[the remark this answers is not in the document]
gustavlrsn: Added `reportErrorToSentry(error)`

server/graphql/v2/object/Organization.js
```diff
@@ -27,17 +27,18 @@ export const Organization = new GraphQLObjectType({
             - Hosts can see the address of organizations submitting expenses to their collectives.
         `,
         async resolve(organization, _, req) {
+          const location = await organization.getLocation();
```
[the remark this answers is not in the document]
gustavlrsn: Nice, good call, added the loader.

server/graphql/v2/interface/Account.js
gustavlrsn: > I wonder why we moved the location resolver to this file rather than keeping it in Individual.location; it's now unclear where one should look at for the resolver code. Like, if I'm returning an organization, will this be used or Organization.location's resolver?

It was moved since all Account types now need a call to `getLocation()` or rather the newly added loader.

Good point about ambiguous resolving for `Organization.location`, I overlooked also removing that one, which I've now done, so all logic/permissions regarding location resolving is now in the Account interface.

Simply removing the `Organization.location` resolver results in a small change for permissions. Only the country was exposed in v2 if you were not a host. However, v1 exposed the full location for all Organizations, so we're not exposing any previously private data if we choose to let location be public for all Account types except for Users in v2 (which seems like the simpler/better approach)

server/lib/location.ts
```diff
@@ -0,0 +1,9 @@
+import { isEqual } from 'lodash';
+
+import { Location } from '../types/Location';
+
+export const mustUpdateLocation = (existingLocation: Location, newLocation: Location) => {
+  const fields = ['country', 'name', 'address', 'lat', 'long', 'structured'];
+  const hasUpdatedField = field => !isEqual(newLocation?.[field], existingLocation?.[field]);
```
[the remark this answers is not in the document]
gustavlrsn: Good catch, fixed this

server/lib/format-address.ts
```diff
@@ -0,0 +1,52 @@
...
+   * are present in the input type, but not available in the @shopify/address formatter.
+   *
+   * All except Antartica (AQ) are US territories and use the US address format.
+   * The US format is provided as a fallback which does not rely on the Shopify API,
+   * since this can also fail to respond to the getCountry request
+   */
+  try {
+    // Locale is only affecting language, not formatting
+    const addressFormatter = new AddressFormatter(locale);
+    const formattingCountry = await addressFormatter.getCountry(country);
```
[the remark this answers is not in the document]
gustavlrsn: That sounds like a much better solution, opened a follow up issue here: https://github.com/opencollective/opencollective/issues/6617

server/graphql/loaders/location.ts
```diff
@@ -0,0 +1,29 @@
...
+          FROM        "Locations" l
+          WHERE       l."CollectiveId" in (:collectiveIds)
+          AND         l."deletedAt" IS NULL`,
+        {
+          type: sequelize.QueryTypes.SELECT,
+          model: models.Location,
+          mapToModel: true,
+          replacements: { collectiveIds },
+        },
+      );
```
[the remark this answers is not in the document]
gustavlrsn: Indeed! 🥴

gustavlrsn:
@Betree thanks for your thorough review! I've addressed your comments. Also added an index on `CollectiveId` for Locations when deletedAt is `null`.

Betree (commented):
My only blocking feedback is about organizations' address visibility. Feel free to merge once it's addressed.

test/server/graphql/v2/query/AccountQuery.test.js
```diff
@@ -121,30 +121,30 @@ describe('server/graphql/v2/query/AccountQuery', () => {
...
-      expect(resultUnauthenticated.data.account.location.address).to.eq('PRIVATE!');
-      expect(resultRandomUser.data.account.location.address).to.eq('PRIVATE!');
-      expect(resultHostAdmin.data.account.location.address).to.eq('PRIVATE!');
+      expect(resultUnauthenticated.data.account.location?.address).to.eq('PUBLIC!');
+      expect(resultRandomUser.data.account.location?.address).to.eq('PUBLIC!');
+      expect(resultHostAdmin.data.account.location?.address).to.eq('PUBLIC!');
     });
 
-    it('is private for organization accounts', async () => {
+    it('is public for organization accounts', async () => {
```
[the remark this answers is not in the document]
gustavlrsn: Fixed this

gustavlrsn:
> My only blocking feedback is about organizations' address visibility. Feel free to merge once it's addressed.

Fixed this, existing behavior will be consistent. And reverted to using separate resolvers for `Individual`s and `Organization`s.

Betree (approved):
Lgtm!

Betree:
Let's make sure we deploy this one at the beginning of the week. There's a good potential for breaking changes; we want some availability if quick fixes are needed.

PR #8629, opened 2023-03-02
Clean relatedCollectives

znarf:
According to Heroku metrics, the query is somehow one of our slowest and we definitely not need this.

Related with https://github.com/opencollective/opencollective/issues/6036

Screenshot 2023-03-02 at 20 45 08


Betree (approved):
Wow, this wasn't even used in email templates? Nice cleaning

Betree:
Confirm that was removed in https://github.com/opencollective/opencollective-api/pull/4505

PR #8636, opened 2023-03-03
Optimize contributionsAmount queries

znarf:
Fix: https://github.com/opencollective/opencollective/issues/6503


Betree (approved):
Good stuff!

PR #8639, opened 2023-03-03
Make cancel recurring job more resilient

Betree:
Related to https://github.com/opencollective/opencollective/issues/6509


Betree (commented):
To review with "Ignore whitespace" changes, only added a try-catch.

hdiniz (approved):

PR #8640, opened 2023-03-03
Fix Wise payout bug on Collectives that have different currencies than their Hosts

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6506
Resolves https://github.com/opencollective/opencollective/issues/6502

### DIFF
--- a/server/lib/transferwise.ts
+++ b/server/lib/transferwise.ts
@@ -41,6 +41,8 @@ const axios = Axios.create({
   baseURL: config.transferwise.apiUrl,
 });
 
+export const isProduction = config.env === 'production';
+
 type TransferwiseErrorCodes = 'balance.payment-option-unavailable' | string;
 
 const signString = (data: string) => {
@@ -623,8 +625,6 @@ export const fundBatchGroup = async (
       }
     });
 };
-
-export const isProduction = config.env === 'production';
 
 const publicKey = fs.readFileSync(
   path.join(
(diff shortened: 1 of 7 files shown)


server/lib/transferwise.ts
```diff
@@ -381,6 +383,31 @@ export const getTransfer = async (connectedAccount: ConnectedAccount, transferId
...
+  const statuses = ['processing', 'funds_converted', 'outgoing_payment_sent'];
+  for (const status of statuses) {
+    response = await requestDataAndThrowParsedError(
+      axios.get,
+      `/v1/simulation/transfers/${transferId}/${status}`,
+      {
+        requestPath: `/v1/simulation/transfers/:id/${status}`,
+        connectedAccount,
+      },
+      'Development: There was an simulating transfer status for Wise',
```
Betree, line 404:
```suggestion
      'Development: There was an error simulating transfer status for Wise',
```

server/paymentProviders/transferwise/index.ts
```diff
@@ -118,7 +120,7 @@ async function quoteExpense(
       'For multi-currency expenses, the host currency must be the same as the collective currency',
     );
     assert(
-      expense.currency === payoutMethod.unfilteredData.currency,
+      expense.currency === targetCurrency,
```
Betree, line 123: Same effect, but definitely clearer :+1:

Betree (approved):
Looks good to me!

PR #8648, opened 2023-03-06
Add collective loaders

znarf:
Make sure to use collective loaders in many new places.


Betree (approved):
Looking good!

One small difference between `models.Collective.findByPk` vs `req.loaders.Collective.byId.load` is that the model is ok with being called with `null` (it then returns null) whereas the loader throws an error. I didn't spot any cases that could be affected by this in this PR.

znarf:
@Betree I actually encountered the issue and had to do things differently in some cases, `getHostCollective()` and `getParentCollective()`.

PR #8651, opened 2023-03-07
Put Contributors behind GraphQL loader

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6514


server/lib/contributors.ts
```diff
@@ -220,63 +222,12 @@ const filterContributors = (contributors: ContributorsList, filters: Contributor
...
-): Promise<ContributorsList> => {
-  const contributorsCache: ContributorsCacheEntry = await loadContributors(collectiveId);
-  const contributors = contributorsCache.tiers[tierId.toString()] || [];
-  return filterContributors(contributors, filters);
-};
-
-/**
- * Get all the contributors that are not part of any tier.
- */
-export const getContributorsWithoutTier = async (
```
Betree, line 273: This one was probably implemented for the "Custom contributions" card, but we're using the default `contributors` query with a filter on Tier in the frontend. Removing for now.

test/server/graphql/v1/collective.test.js
```diff
@@ -1294,4 +1302,53 @@ describe('server/graphql/v1/collective', () => {
...
+            roles
+          }
+          admins: contributors(roles: ADMIN) {
+            id
+          }
+          tiers {
+            contributors {
+              id
+            }
+          }
```
Betree, line 1322: Here we previously had 4 queries to `getContributorsForCollective`: one for the contributors, one for the admins, and one for each of the two tiers.

znarf:
Does this remove the cache, fast enough without it?

znarf:
> Does this remove the cache

Ok, it does not.

Betree:
> > Does this remove the cache
> 
> Ok, it does not.

Indeed, I haven't measured but I think the cache is still interesting to keep; the contributors query was still pretty heavy last time I checked.

server/models/TransactionSettlement.ts
```diff
@@ -77,6 +77,7 @@ class TransactionSettlement extends Model<
         AND t."deletedAt" IS NULL
         AND ts."deletedAt" IS NULL
         ${settlementStatus ? 'AND ts."status" = :settlementStatus' : ''}
+        ORDER BY "id" ASC
```
Betree, line 80: Not related to this PR, but the test somehow randomly fails because the transactions order was inconsistent

image

Betree:
> Is the loader here to debounce multiple requests?

Yes, only that. At this stage, it doesn't add the ability to fetch contributors for multiple collectives in a single SQL request.

PR #8654, opened 2023-03-07
Ability to set arbitrary custom data on Expenses

Betree:
Part of https://github.com/opencollective/opencollective/issues/6451

I've limited the max custom data size to 10kb for this first release.

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -771,7 +771,7 @@ const validateExpenseCustomData = (value: Record<string, unknown> | null): void 
   // Validate size
   const payloadSize = Buffer.byteLength(JSON.stringify(value), 'utf8');
   if (payloadSize > 10e3) {
-    throw new ValidationFailed(`Expense custom data cannot exceed 10kb. Current size: ${payloadSize / 1000}kb`);
+    throw new ValidationFailed(`Expense custom data cannot exceed 10kB. Current size: ${payloadSize / 1000}kB`);
   }
 };
 
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -12,6 +12,7 @@ import {
   isEqual,
   isNil,
   isNumber,
+  isUndefined,
   keyBy,
   mapValues,
   omit,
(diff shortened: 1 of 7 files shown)


aerugo:
@Betree thanks! So, do understand correctly that this make it possible to call the expense submission flow with a customData parameter? And possible to query expenses with some customData?

Betree:
> @Betree thanks! So, do understand correctly that this make it possible to call the expense submission flow with a customData parameter? And possible to query expenses with some customData?

Correct! We plan to plug this in the frontend via URL parameters, so that you can pass this custom data while reusing the existing form.

We also plan to add this field to the expenses search.

server/graphql/common/expenses.ts
```diff
@@ -747,6 +758,23 @@ async function validateExpensePayout2FALimit(req, host, expense, expensePaidAmou
...
+
+  // Validate type: must be a JSON object
+  if (typeof value !== 'object') {
+    throw new ValidationFailed('Expense custom data must be an object');
+  }
+
+  // Validate size
+  const payloadSize = Buffer.byteLength(JSON.stringify(value), 'utf8');
+  if (payloadSize > 10e3) {
+    throw new ValidationFailed(`Expense custom data cannot exceed 10kb. Current size: ${payloadSize / 1000}kb`);
```
znarf, line 774:
```suggestion
    throw new ValidationFailed(`Expense custom data cannot exceed 10kB. Current size: ${payloadSize / 1000}kB`);
```

A bit pedantic but here we go.

znarf: https://en.wikipedia.org/wiki/Kilobyte

test/server/graphql/v2/mutation/ExpenseMutations.test.js
```diff
@@ -401,6 +404,22 @@ describe('server/graphql/v2/mutation/ExpenseMutations', () => {
...
+      const expenseData = { ...getValidExpenseData(), payee: { legacyId: payee.id }, customData };
+      const collective = await fakeCollective();
+      const result = await graphqlQueryV2(
+        createExpenseMutation,
+        { expense: expenseData, account: { legacyId: collective.id } },
+        user,
+      );
+
+      expect(result.errors).to.exist;
+      expect(result.errors[0].message).to.eq('Expense custom data cannot exceed 10kb. Current size: 10.008kb');
```
znarf, line 421:
```suggestion
      expect(result.errors[0].message).to.eq('Expense custom data cannot exceed 10kB. Current size: 10.008kB');
```

znarf (approved):
Looks good.

gustavlrsn (approved):
Looks great!

PR #8658, opened 2023-03-08
Refactor collective model to typescript

hdiniz:
Related https://github.com/opencollective/opencollective/issues/5893
Might be better to review by following the individual commits on 'Collective.ts`.


server/lib/merge-accounts.ts
```diff
@@ -80,7 +78,7 @@ export const simulateMergeAccounts = async (
 
   const addCountsToSummary = counts => {
     Object.entries(counts).forEach(([key, count]) => {
-      if (count > 0) {
+      if ((count as number) > 0) {
         addLineToSummary(`  - ${key}: ${count}`);
       }
     });
```
Betree, line 84: Wouldn't it be better to type it in the function parameter? `as` kind of bypasses type checks

```es6
const addCountsToSummary = (counts: Record<string, number>) => {
```

Betree: Note that this one doesn't seem to original from this PR, I'm curious why we haven't seen it before

test/server/lib/merge-accounts.test.ts
```diff
@@ -93,7 +93,7 @@ describe('server/lib/merge-accounts', () => {
 
   before(async function () {
     // This test can take some time to setup on local env
-    this.timeout(45000);
+    this.timeout(90000);
```
Betree, line 96: - https://github.com/opencollective/opencollective/issues/6533

test/test-helpers/fake-data.ts
```diff
@@ -358,7 +369,7 @@ export const fakeExpense = async (expenseData: Record<string, unknown> = {}) =>
  * Creates a fake comment. All params are optionals.
  */
 export const fakeComment = async (
-  commentData: Record<string, unknown> = {},
+  commentData: Partial<InferCreationAttributes<Comment>> = {},
```
Betree, line 372: Nice job typing all these with `Partial<InferCreationAttributes<...>>` :+1:

Betree (approved):
Looking great! This is a massive improvement.

server/lib/merge-accounts.ts
hdiniz: Changed the CollectiveFieldsConfig type as merge-accounts, maybe it propagated to here.

PR #8662, opened 2023-03-09
Ability to search for expenses by custom data

Betree:
Require https://github.com/opencollective/opencollective-api/pull/8654
Part of https://github.com/opencollective/opencollective/issues/6451

This will let API consumers search for expenses by `customData`.

Usage:

```gql
# Will return all expenses where customData.my.nested.key = 42
expenses(host: $host, customData: $customData) {
   ...
}
```

```es6
variables = {
  host: ...,
  customData: { my: { nested: { key: 42 } } }
}

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -801,7 +801,7 @@ async function validateExpensePayout2FALimit(req, host, expense, expensePaidAmou
   }
 }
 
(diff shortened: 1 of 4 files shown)


kewitz (approved):
This makes sense, especially considering we won't be able to automatically optimize these queries.
I wonder if there's something more abstract our users need that we're missing on our model.

PR #8668, opened 2023-03-10
Add basic support for yubikey otp 2fa

hdiniz:

Add yubikey otp 2fa provider and device id column to Users table.

This will be used as MVP for internal users only, device ids will be inserted directly at this time.


hdiniz:
> Small thing, we should use kebab case for files. https://github.com/opencollective/opencollective/blob/main/rfcs/004-file-naming-convention.md#2-file-names-should-be-kebab-case
> 
>     * recovery_code.ts -> recovery-code.ts
> 
>     * yubikey_otp.ts -> yubikey-otp.ts

Thanks, done!

migrations/20230309234219-add-yubikeyDeviceId-to-user-table.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.addColumn('Users', 'yubikeyDeviceId', {
+      type: Sequelize.STRING,
+      allowNull: true,
+      defaultValue: null,
+      unique: true,
```
Betree, line 10: I don't think we should make that one unique, it's a legit use case to share the same key for multiple accounts (e.g. I use the same one for my personal and OC Cloudflare account)

```suggestion
```

server/lib/two-factor-authentication/yubikey-otp.ts
```diff
@@ -0,0 +1,41 @@
...
+      throw new ApolloError('Two-factor authentication code is invalid', 'INVALID_2FA_CODE');
+    }
+  },
+};
+
+async function validateYubikeyOTP(otp: string): Promise<boolean> {
+  const response = await fetch(
+    `https://api2.yubico.com/wsapi/2.0/verify?${new URLSearchParams({
+      otp,
+      nonce: crypto.randomBytes(16).toString('base64'),
```
Betree, line 35: What's the purpose of this nonce? Are we suppose to read it back in the response?

Betree (approved):
The code is simple - as described in the kickoff. Good job!

server/lib/two-factor-authentication/yubikey-otp.ts
hdiniz: Its a required field on the API, I am not sure if we need to validate it on the response. From what I could gather its used by the API to differentiate between a replayed OTP (otp only) and a replayed request (same OTP and nonce)

PR #8671, opened 2023-03-13
Add ts interface to models

hdiniz:
Related https://github.com/opencollective/opencollective/issues/5893

Trying to add typings to models without necessarily converting them to classes, as moving to classes requires moving code around which makes it hard to review the diffs.

This change should provide typings for models without touching "running" code, which we can then further refactor into classes.


server/graphql/v2/input/PaymentMethodInput.ts
```diff
@@ -75,7 +75,7 @@ export const getLegacyPaymentMethodFromPaymentMethodInput = async (
   if (!pm) {
     return null;
   } else if (pm.id) {
-    return fetchPaymentMethodWithReference(pm);
+    return fetchPaymentMethodWithReference(pm) as any;
   }
```
Betree, line 79: Maybe we should rather change the return type for this one to something like `Promise<Record<string, unknown> & {service: string; type: string;}>`. Not a fan of returning `any`, but not a blocker if it's too much complexity for now.

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -173,7 +173,7 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
     const paymentMethod = await fetchPaymentMethodWithReference(args.paymentMethod, {
       sequelizeOpts: { attributes: ['id'], include: [{ model: models.Collective }] },
     });
-    if (!req.remoteUser?.isAdminOfCollective(paymentMethod.Collective)) {
+    if (!req.remoteUser?.isAdminOfCollective((paymentMethod as any).Collective)) {
```
Betree, line 176: We should rather have `Collective` as an optional attribute of payment method, like we do in https://github.com/opencollective/opencollective-api/blob/3117bff193fa20c2a4f5476e6a290e8d3cb79e08/server/models/Expense.ts#L79 for example

Betree (commented):
I like how it keeps the diff minimal, it's a good approach to start with interfaces. 

I shared some concerns below about places where we're casting to `any`, I wonder if we could avoid that.

server/graphql/v2/input/PaymentMethodInput.ts
hdiniz: Agreed. I think we should eventually go over the linter warnings for any and remove as much as we can.

Betree (approved):
No blocker for me!

PR #8673, opened 2023-03-13
Increase test timeout

hdiniz:


Betree (approved):
Good temporary hack. Documented https://github.com/opencollective/opencollective/issues/6533 for long-term solution

PR #8685, opened 2023-03-15
Improve various checks

Betree:


znarf (approved):
Good for me. Except the `testw` thing?

PR #8686, opened 2023-03-15
Make some of the database fields more secure with validating length

RichardLitt:
Seems like a good idea.

### DIFF
--- a/server/models/Collective.ts
+++ b/server/models/Collective.ts
@@ -3386,10 +3386,18 @@ Collective.init(
       },
     },
 
-    description: DataTypes.STRING, // max 95 characters
+    description: {
+      type: DataTypes.STRING,
+      validate: {
+        len: [0,95]
+      },
+     },
 
     longDescription: {
       type: DataTypes.TEXT,
+      validate: {
+        len: [0,50000] // just to prevent people from putting a lot of text in there
+      },
       set(longDescription: string) {
         if (longDescription) {
           this.setDataValue('longDescription', sanitizeHTML(longDescription, longDescriptionSanitizerOptions));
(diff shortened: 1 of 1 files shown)


Betree (commented):
Definitely needed, I'll do a bit of data investigation to make sure we don't have existing cases before deploying

Betree:
Thank you @RichardLitt!

server/models/Collective.ts
```diff
@@ -3372,10 +3372,18 @@ Collective.init(
       },
     },
 
-    description: DataTypes.STRING, // max 95 characters
+    description: {
+      type: DataTypes.STRING,
+      validate: {
+        len: [0,95]
```
Betree, line 3378:
```suggestion
        len: [0, 255]
```

server/models/Collective.ts
```diff
@@ -3372,10 +3372,18 @@ Collective.init(
...
+      },
+     },
 
     longDescription: {
       type: DataTypes.TEXT,
+      validate: {
+        len: [0,50000] // just to prevent people from putting a lot of text in there
```
Betree, line 3385:
```suggestion
        len: [0, 30000] // just to prevent people from putting a lot of text in there
```

server/models/Collective.ts
```diff
@@ -3455,9 +3463,19 @@ Collective.init(
       },
     },
 
-    locationName: DataTypes.STRING,
+    locationName: { 
+      type: DataTypes.STRING,
+      validate: {
+        len: [0,5000] // Just to prevent people typing too much in here.
+      },
+    },
```
Betree, line 3471: Rolling back the changes since these fields are being removed in another PR

```suggestion
    locationName: DataTypes.STRING,
```

server/models/Collective.ts
```diff
@@ -3455,9 +3463,19 @@ Collective.init(
...
 
-    address: DataTypes.STRING,
+    address: { 
+      type: DataTypes.STRING,
+      validate: {
+        len: [0,5000] // Just to prevent people typing too much in here.
+      },
+    },
```
Betree, line 3478: Rolling back the changes since these fields are being removed in another PR

```suggestion
    address: DataTypes.STRING,
```

Betree (commented):
Adapted values based on what we have in the DB (after banning spammers)

RichardLitt:
Let me know if you need me to do anything.

Betree:
@RichardLitt It's all good, this one should be ready to merge!

Betree (approved):

Betree:
Thank you @RichardLitt!

PR #8688, opened 2023-03-15
getHostCollective: by default, don't return if not approved

Betree:
Todo before merging: check if we have cases of collectives where only one of `isApproved` or `approvedAt` is set.

### DIFF
--- a/server/models/Collective.ts
+++ b/server/models/Collective.ts
@@ -2845,7 +2845,11 @@ class Collective extends Model<
   };
 
   // get the host of the parent collective if any, or of this collective
-  getHostCollective = function () {
+  getHostCollective = async function ({ returnEvenIfNotApproved = false } = {}) {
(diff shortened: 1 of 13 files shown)


server/models/Collective.ts
```diff
@@ -2845,7 +2845,11 @@ class Collective extends Model<
   };
 
   // get the host of the parent collective if any, or of this collective
-  getHostCollective = function () {
+  getHostCollective = async function ({ returnEvenIfNotApproved = false } = {}) {
+    if (!this.isActive && !returnEvenIfNotApproved) {
```
Betree, line 2849: The initial condition was:

```suggestion
    if (!(this.isActive && this.approvedAt) && !returnEvenIfNotApproved) {
```

but many tests are not setting `approvedAt` properly, and though I started fixing them there were just too many so I went for simplification.

kewitz (approved):

PR #8691, opened 2023-03-15
Add support for captcha in support form

gustavlrsn:
This PR is moving the contact support endpoint from the frontend project to the API, in order to use the captcha verification function, mutualized from `orders.js`.


server/routes.js
```diff
@@ -26,6 +26,7 @@ import graphqlSchemaV2 from './graphql/v2/schema';
 import cache from './lib/cache';
 import errors from './lib/errors';
 import logger from './lib/logger';
+import sendMessageToSupport from './lib/message-support';
```
znarf, line 29: Maybe move this to the controller folder.

gustavlrsn: Yes, nice, it's moved.

znarf:
Looks good! Let's make sure to double check it's working nicely when in production.

znarf (approved):

PR #8695, opened 2023-03-16
Make codecov great again

Betree:
- Tag tests
- Include untested files in stats
- Add info on PRs
- Fix flawky test

This PR might require some iterations, as Codecov is not always playing well with branch-specific configs; it's difficult to fully test.


test/server/graphql/v2/mutation/ConnectedAccountMutations.test.ts
```diff
@@ -16,6 +16,10 @@ describe('server/graphql/v2/mutation/ConnectedAccountMutations', () => {
   });
   beforeEach(utils.resetTestDB);
 
+  afterEach(() => {
+    sandbox.restore();
+  });
+
```
Betree, line 22: Not sure why it started now, but without this one we were failing with:

```
       "before all" hook in "server/paymentProviders/transferwise/index":
     TypeError: Attempted to wrap getProfiles which is already wrapped
```

.github/workflows/ci.yml
```diff
@@ -264,7 +264,7 @@ jobs:
 
       - run: npm run test -- --ignore "test/server/graphql/**"
 
-      - run: npm run test:coverage
+      - run: npm run test:coverage -- -F Unit
```
Betree, line 267: Adding some tags to track our unit/graphql coverages separately

PR #8697, opened 2023-03-16
Group Static checks

kewitz:
Related to https://github.com/opencollective/opencollective/issues/6567.
I don't think we even need to run in parallel since other jobs take much longer to run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ env:
   CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
 jobs:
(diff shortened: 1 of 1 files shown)


Betree (commented):
The good part about running in parallel is to somehow get all static results at once.

I sometimes submit raw PRs to quickly run the CI and I'm glad it gives me Typescript, Lint, Prettier, etc. as the same results so that I can quickly fix everything.

I'm afraid it will create back-and-forth if we don't do this

kewitz:
We'll definitely need to open the test result to see what is going wrong. The upside is to save runners and avoid action queueing.

PR #8698, opened 2023-03-16
Merge accounts: snapshot test data

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6533

## Motivation

Some tests need heavy initialization before running. Merge-accounts is a good example of that: we want to make sure accounts are moved along with all their associations, and we want to make sure they don't touch other entries in the DB, We thus need to insert many rows accross many tables

Inserting these rows one by one takes time. In the abovementioned case, over 30s-40s.

## Proposed solution

This PR implements a `getOrCreateDBSnapshot` helper that, using `pg_dump` & `pg_restore`, creates a backup of the DB after its initializer runs. Next test runs will then re-use this snapshot  rather than re-creating the data manually

Restoring a snapshot takes 5s-10s (depending on the number of migrations to run). To make sure we keep them up to date, I've introduced a `npm run test:update-db-snapshots` command along with warnings when slow snapshots are detected.

## Use cases

- Optimize performance for slow tests initializers
- Share a base context between tests: because snapshots use a unique identifier, multiple tests could use the same data basis for their environment. The result would be something like:

`test/utils/initializers/large-fiscal-host.ts`
```es6
export const name = 'large-fiscal-host';
export const initialize = async () => { ... }
export const getData = async () => { ... }
```

`test/server/test-(1|2).test.ts`
```es6
import * as LargeFiscalHostInitializer from '../utils/initializers/large-fiscal-host';

let host;
before(async () => {
  await getOrCreateDBSnapshot(this, LargeFiscalHostInitializer.name, LargeFiscalHostInitializer.initialize);
  ({ host }) = await LargeFiscalHostInitializer.getData();
})
```

## Todo

<details><summary>Details</summary>
<p>

- [x] Save test data
- [x] Load test data from snapshot
- [x] Run migrations
- [x] Add commands to easily update snapshots
- [x] Suggest or force to update snapshot when needed

</p>
</details>


kewitz:
Ben, I suspect most of that time was due to `resetTestDB` truncating process.
I think there's a simpler approach to this, we could just export the DB schema and update `resetTestDB` to use this snapshot:
- `pg_dump -Z9 -Fc -cs "postgres://postgres@localhost/opencollective_test" > schema.dump`
- `pg_restore --clean --exit-on-error --schema public -d "postgres://postgres@localhost/opencollective_test" schema.dump`

What do you think?

Betree:
> I suspect most of that time was due to resetTestDB truncating process.

It was not, I measured the time from the first fake* to the last one, aka only DB insertions/update.

> I think there's a simpler approach to this, we could just export the DB schema and update `resetTestDB` to use this snapshot:
> - `pg_dump -Z9 -Fc -cs "postgres://postgres@localhost/opencollective_test" > schema.dump`
> - `pg_restore --clean --exit-on-error --schema public -d "postgres://postgres@localhost/opencollective_test" schema.dump`
> 
> What do you think?

That's pretty much what we're doing here, only it's done only in the test that needs here to not interfere with the others.

Betree:
@kewitz That being said I did notice some really slow `resetTestDb` and agree we should investigate that

kewitz (approved):
Looking good. I hate that we need to run migrations on the test DB, should we just rely on updating the snapshot?

Betree:
> Looking good. I hate that we need to run migrations on the test DB, should we just rely on updating the snapshot?

It makes things slower, but at the same time, I don't want to force developers to update the snapshots each time they introduce a new migration. I'm not sure what's the best compromise here.

PR #8699, opened 2023-03-16
Take Into Account Custom Tier when Updating Order

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6568

Relate Front-End PR: https://github.com/opencollective/opencollective-frontend/pull/8734

[update-tier-custom.webm](https://user-images.githubusercontent.com/12435965/225717104-ee7e90fc-ea32-41a0-a475-803bec415128.webm)


Betree (approved):
Good fix!

PR #8700, opened 2023-03-16
Correct Condition to Clear Goal Once it's Already Set

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6569

[clear-goal-set.webm](https://user-images.githubusercontent.com/12435965/225723349-d827267a-bf9c-481f-a707-d8e9287153ef.webm)


Betree (approved):
Works for me

PR #8707, opened 2023-03-20
Add loader for Host -> totalHostedCollectives

Betree:
Identified as a requirement for https://github.com/opencollective/opencollective/issues/5672

This will also improve the performance of the search page, where each search for hosts was triggering 20 SQL requests to retrieve this.

### DIFF
--- a/server/graphql/loaders/index.js
+++ b/server/graphql/loaders/index.js
@@ -127,6 +127,7 @@ export const loaders = req => {
         HostCollectiveId: collectiveIds,
         type: [CollectiveType.COLLECTIVE, CollectiveType.FUND],
         isActive: true,
+        approvedAt: { [Op.not]: null },
       },
     });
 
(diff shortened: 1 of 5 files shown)


server/graphql/loaders/index.js
```diff
@@ -117,6 +118,23 @@ export const loaders = req => {
...
 
+  context.loaders.Collective.hostedCollectivesCount = new DataLoader(async collectiveIds => {
+    const results = await models.Collective.findAll({
+      raw: true,
+      attributes: ['HostCollectiveId', [sequelize.fn('COUNT', sequelize.col('id')), 'count']],
+      group: ['HostCollectiveId'],
+      where: {
+        HostCollectiveId: collectiveIds,
+        type: [CollectiveType.COLLECTIVE, CollectiveType.FUND],
+        isActive: true,
```
znarf, line 129: Should we add `approvedAt` not null ?

PR #8708, opened 2023-03-20
Performance: group 3 queries into one for Host Reports

znarf:


Betree (approved):
Good change. Any rough metric about how much we're saving by doing that?

kewitz (approved):

PR #8710, opened 2023-03-21
performance: async hostMetrics

znarf:


znarf:
I moved to something else and not sure the value of this anymore. Closing.

PR #8712, opened 2023-03-21
Expense: Handle taxes and customData for drafts

Betree:
Fix https://github.com/opencollective/opencollective/issues/6558


kewitz (commented):
Looking good. Could you please update `models/RecurringExpense@createNextExpense` to make sure this is also compatible with recurring expenses?

Betree:
@kewitz done, thanks for the catch!

It would be nice to figure out how to share this logic between regular, draft, and recurring expenses. It's a lot to expect from developers to remember that they need to update all these files.

PR #8713, opened 2023-03-21
performance: getHostedCollectives "raw" optimization

znarf:


server/models/Collective.ts
```diff
@@ -1845,15 +1845,7 @@ class Collective extends Model<
   getHostedCollectives = async function (queryParams = {}) {
     return models.Collective.findAll({
       ...queryParams,
-      where: { isActive: true, HostCollectiveId: this.id },
-      includes: [
```
znarf, line 1849: This was doing nothing, it's `include` not `includes`.

Betree (commented):
Looking good; we should make it a rule to always use `raw` when we're not using the model

PR #8715, opened 2023-03-22
Add script to refund host fees

Betree:
For https://opencollective.freshdesk.com/a/tickets/180091
Tickets TLDR ; Wrongffully added funds with host fees, cannot refund then re-add because expenses have been paid and there's not enough balance to refund (see https://github.com/opencollective/opencollective/issues/6591)

The manual alternative is:
- Delete the HOST_FEE + HOST_FEE_SHARE + HOST_FEE_SHARE_DEBT transactions
- Delete the matching TransactionSellements entry (group = 'eec0c658-53e9-45e1-b8c7-9e23448c4994')
- Create an expense on https://opencollective.com/opencollective/ with https://opencollective.com/giftcollective as the payee, to make sure OC refunds the settled amount ($245.48USD or $396NZD)

It seemed simpler and less error-prone to add a script for that.

### DIFF
--- a/scripts/ledger/refund-host-fee-on-transaction.ts
+++ b/scripts/ledger/refund-host-fee-on-transaction.ts
@@ -1,0 +1,65 @@
+#!/usr/bin/env ./node_modules/.bin/babel-node
+import '../../server/env';
+
+import { Command } from 'commander';
+import { v4 as uuid } from 'uuid';
+
+import { refundHostFee } from '../../server/lib/payments';
+import models from '../../server/models';
+
+const DRY_RUN = process.env.DRY_RUN !== 'false';
+
+if (DRY_RUN) {
+  console.log('DRY RUN: No changes will be made');
+}
+
+const program = new Command()
+  .description('Helper to refund host fees on a given transaction')
+  .arguments('TransactionGroup UserCollectiveSlug')
+  .parse();
+
+const main = async () => {
+  const transactionGroup = program.args[0];
+  const userCollectiveSlug = program.args[1];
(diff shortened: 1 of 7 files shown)


znarf (approved):
Looks good!

scripts/ledger/refund-host-fee-on-transaction.ts
SudharakaP, line 1: @Betree : Looks good. Should we add a entry to activity log as otherwise running this will have no trace who refunded the transaction? 🤔

Betree: In my opinion, it's not really needed while this stays as a script. Especially since it doesn't destroy any data - to rollback simply delete the refund and reset `RefundTransactionid`. But we should if we ever move this to a root action.

Betree:
Script ran, the [contribution](https://opencollective.com/hr-consultancy-project/orders/596555) page acknowledges this successfully (praise the new ledger, we're really seeing its power here):

image

The ledger also looks correct:

```sql
select id, kind, type, description, "RefundTransactionId", "isRefund", amount from "Transactions" where "OrderId" = 596555 ORDER BY id
```

image

And the `TransactionSettlements` entry is there. 

I wasn't 100% sure about the settlement behavior with refunds and did not find any test to clear my doubts, so I updated the existing ones to add a case for refunds.

Betree:
The [settlement](https://opencollective.com/giftcollective/expenses/131755) for GIft Collective just arrived; the refund has been properly acknowledged and is surfaced in the CSV:

image

So we're all done here, merging!

PR #8718, opened 2023-03-22
wip: new strategy for loaders

znarf:
Promising performance improvement here.

Require https://github.com/opencollective/opencollective-frontend/pull/8750

Implement a strategy where loaders are shared to serve the same frontend request.

Can later have other strategies in place to share loader.


znarf:
In the way it is implemented, this was a valuable and safe improvement for speeding up SSR rendering of multi-query pages like the collective page.

A bit difficult to measure and at the price of some added complexity. 

I'm closing for now if there is no shared enthusiasm for the approach.

PR #8722, opened 2023-03-23
performance: update approach at graphql cache

znarf:
- Cache everything that has an account slug.
- Keep a list of all cache keys used. 
- Flush them all when we receive a cache clear for that slug

### DIFF
--- a/server/graphql/cache.js
+++ b/server/graphql/cache.js
@@ -3,23 +3,6 @@ import { difference, isEqual } from 'lodash';
 import { TransactionKind } from '../constants/transaction-kind';
 import { md5 } from '../lib/utils';
 
-export const purgeCacheForCollectiveOperationNames = [
-  'CollectivePage',
-  'ContributePage',
-  'BudgetSection',
-  'BudgetSectionWithHost',
-  'UpdatesSection',
-  'TransactionsSection',
-  'TransactionsPage',
-  'ExpensesPage',
-  'CollectiveBannerIframe',
-  'CollectiveCover',
-  'RecurringContributions',
-  'ContributionsSectionStatic',
-  'Members_Users',
-  'Members_Organizations',
-];
-
 export function checkSupportedVariables(req, variableNames) {
   if (!req.body.variables) {
     return false;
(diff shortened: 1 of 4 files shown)


server/graphql/cache.js
```diff
@@ -1,176 +1,27 @@
...
+  if (req.remoteUser) {
     return;
   }
-  return `${req.body.operationName}_${queryHash}_${req.body.variables.slug}`;
-}
 
-export function getGraphqlCacheKey(req) {
-  if (req.remoteUser) {
+  // If it's not a GraphQL query with such properties, no cache
+  if (!req.body || !req.body.operationName || !req.body.query || !req.body.variables) {
```
Betree, line 9: Can we also exclude mutations? Or maybe they never reach this check?

server/graphql/cache.js
```diff
@@ -1,176 +1,27 @@
...
     return;
   }
-  if (!req.body || !req.body.operationName || !req.body.query) {
+
+  // We only want to cache queries with an Account slug
+  // TODO: check that slug never express something else like a Tier
+  const slug = req.body.variables.slug || req.body.variables.collectiveSlug;
+  if (!slug) {
```
Betree, line 16: Out of safety, we should ignore queries that contain variables that give you special permissions, such as `draftKey` for expenses (though I acknowledge it has no chance to happen with the code as it is today since the `ExpenseQuery` doesn't take a slug as argument).

I don't have any other cases in mind right now, but we can also be defensive and protect against anything that sounds like a private variable (e.g. `token`, `password`, etc). Maybe something to synchronize with https://github.com/opencollective/opencollective-api/pull/8452?

server/lib/cache/index.js
```diff
@@ -141,18 +140,19 @@ export function memoize(func, { key, maxAge = 0, serialize, unserialize }) {
...
-    }),
-  );
+export async function purgeGraphqlCacheForCollective(slug) {
+  return cache.get(`graphqlCacheKeys_${slug}`).then(keys => {
+    if (keys) {
+      for (const key of keys) {
+        cache.delete(key);
+      }
+    }
+  });
```
Betree, line 150: Nice!

Betree (commented):
Nice changes! I like that this is way more generic than the previous approach; it will be much easier to maintain. Combined with the check that checks for the request time before caching, this sets a good auto-caching strategy.

One possible concern is that it becomes quite easy to saturate our cache with unnecessary data; but I guess there are easier ways to DDOS us.

server/graphql/cache.js
znarf: I think we assume that Mutations are authenticated? And remoteUser is set?

But I guess we have some exceptions to this (guest flow, draft expenses), so would be worth checking.

znarf: Added a check.

Betree: > But I guess we have some exceptions to this (guest flow, draft expenses), so would be worth checking.

Also `createCollective`

PR #8725, opened 2023-03-24
Improve resetTestDb performance

Betree:
An experiment to use DB snapshots (`pg_dump`/`pg_restore`) rather than `TRUNCATE` to improve `resetTestDb`'s performance and hopefully solve timeouts like https://github.com/opencollective/opencollective/issues/6484#issuecomment-1482568530.


Betree:
Closing in favor of https://github.com/opencollective/opencollective-api/pull/8735, the performance of this PR is miles away from what is achieved there.

PR #8724, opened 2023-03-24
Iterate on expense permissions

Betree:
Related to https://github.com/opencollective/opencollective/issues/4729
Follow-up on https://github.com/opencollective/opencollective-api/pull/8655, https://github.com/opencollective/opencollective-api/pull/8719

- Restore edit permissions for collective admins on virtual card charges
- Add a `host.settings.allowCollectiveAdminsToEditPrivateExpenseData` flag to restore legacy permissions on payout methods and edit expenses for the hosts that need it.
- Refactors the tests to ensure we cover these new contexts while not introducing regressions.

PR #8719, opened 2023-03-23
Make Sure Collective Admins has no Permission to Edit Expense

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6597

[collective-admins-no-edit.webm](https://user-images.githubusercontent.com/12435965/227072943-8f778348-f87a-4fb5-9477-e268b0992313.webm)

PR #8731, opened 2023-03-27
ExpensePermissions: Rollback changes & improve tests

Betree:
Rollback the changes made for https://github.com/opencollective/opencollective/issues/4729 in the following PRs:
- https://github.com/opencollective/opencollective-api/pull/8655
- https://github.com/opencollective/opencollective-api/pull/8719
- https://github.com/opencollective/opencollective-api/pull/8724

I did not want to rollback the test changes, and they're increasing the coverage. I made an update to make them even clearer, making sure failed assertions displayed both the context and the permission that failed:


SudharakaP:
@Betree : Thanks good call until we can mature a solution. 🙏🏽 🙏🏽

PR #8733, opened 2023-03-27
Optimize Members indexes

Betree:
Part of https://github.com/opencollective/opencollective/issues/3205

While looking at a [slow query](https://open-collective.sentry.io/issues/3969837441), I noticed that:
- They're not indexed by `TierId`; yet this is widely used by `req.loaders.Tier.contributorsStats` on the collective page
- We're currently indexing deleted entries, forcing queries to re-filter after getting the results


kewitz (approved):

Betree:
@kewitz Yes, I updated it to make sure it has the latest migrations included.

PR #8734, opened 2023-03-27
Remove unucessary `resetTestDB` calls

Betree:
WIP


Betree:
Given the improvements pushed in https://github.com/opencollective/opencollective-api/pull/8735, `resetTestDB` is now really cheap, and while we should not abuse it, I'm not willing to invest energy in this PR atm just to save a few ms.

PR #8742, opened 2023-03-28
Handle multiple 2fa method in mutations

hdiniz:

Handles adding and removing a second 2FA method.


kewitz (approved):
Nice work. It looks future-proof and should work with the abstractions we predicted for the future.

PR #8752, opened 2023-03-29
Perf: Expense security resolver

kewitz:
I found some margin for simplifying the getExpenseStats that can be further escalated. To avoid unnecessary complexities, I want to ship this cache-based optimization and see how it behaves.


server/graphql/v2/object/Expense.ts
```diff
@@ -392,7 +393,15 @@ const Expense = new GraphQLObjectType({
...
           if (await ExpenseLib.canSeeExpenseSecurityChecks(req, expense)) {
-            return checkExpense(expense);
+            if (
+              expense.data?.security &&
+              moment(expense.data?.security.evaluatedAt).isAfter(moment().subtract(1, 'day'))
+            ) {
+              return expense.data.security.checks;
+            }
+            const checks = await checkExpense(expense);
+            await expense.update({ data: { ...expense.data, security: { checks, evaluatedAt: moment() } } });
```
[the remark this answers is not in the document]
kewitz: Usually like this:
```json
[
  {
    "level": "MEDIUM",
    "message": "User is not using 2FA",
    "scope": "USER",
    "details": "Expense submitted by User",
  },
  {
    "level": "PASS",
    "message": "User was successfully paid 2 times by this collective and 2 times on the platform",
    "scope": "USER",
    "details": "Expense submitted by User",
  },
  {
    "level": "HIGH",
    "message": "Payout Method has never been paid on the platform",
    "scope": "PAYOUT_METHOD",
    "details": null,
  }
]
```

kewitz:
@znarf @Betree, I agree that this is complicated to invalidate, but I can't entirely agree that are many more improvements to make using loaders.

The `HostDashboardExpenses` loads all the pending expenses in the host, which means there's a reduced probability that these expenses have anything in common. The only symmetry we have here is across the multiple requests in a given period, which is a good argument for caching.

This is the same issue we face when exporting host transactions. Since we sort them throughout time, the entropy of the result defeats any request-isolated optimizations like the DataLoader.

Another solution is decoupling the security check from the main query used to render the page. This way, we can render the expenses right-away and leave the costly calculation of the security check to be executed asynchronously.

Betree (commented):
> I can't entirely agree that are many more improvements to make using loaders.

Admitting the host dashboard loads 10 expenses, we can still condense 10 queries into 1 for the following cases:
- `populateRoles`
- `findRelatedUsersByIp`
- `findRelatedUsersByConnectedAccounts`
- `checkExpenseStats` up to x 3 (that itself calls `getExpensesStats` x 2), so 6 DB calls per expense

Altogether we could save up to 81 DB calls per host dashboard request (10 expenses x 9 DB calls in the first case, just 9 DB calls if we use loaders). That seems worth exploring IMO.

kewitz:
Refactored this focusing on Loaders like suggested by @Betree.
One naive local benchmark against remote prod data suggests an improvement:

Main
```
14:29:36 ::1 HostDashboardExpenses {} 6973ms Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/111.0.0.0 Safari/537.36
```

Refactored
```
14:30:24 ::1 HostDashboardExpenses {} 4076ms Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/111.0.0.0 Safari/537.36
```

migrations/20230331163500-expense-security-check-index.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.addIndex('Expenses', ['PayoutMethodId'], {
```
Betree, line 6: Good addition

server/models/User.ts
```diff
@@ -312,8 +317,12 @@ class User extends Model<InferAttributes<User>, InferCreationAttributes<User>> {
     return count > 0;
   };
 
+  getLastKnownIp = function (): string {
+    return this.data?.lastSignInRequest?.ip || this.data?.creationRequest?.ip;
+  };
```
Betree, line 322: :+1:

server/models/User.ts
```diff
@@ -39,7 +44,7 @@ class User extends Model<InferAttributes<User>, InferCreationAttributes<User>> {
   public declare twoFactorAuthRecoveryCodes: CreationOptional<string[]>;
   public declare CollectiveId: number;
   public declare newsletterOptIn: boolean;
-  public declare data: CreationOptional<Record<string, unknown>>;
+  public declare data: CreationOptional<Record<string, unknown> & UserData>;
```
Betree, line 47: :+1:

server/types/express.d.ts
```diff
@@ -24,7 +25,7 @@ declare global {
       };
       userToken?: UserToken;
       personalToken?: PersonalToken;
-      loaders: Record<string, DataLoader>;
+      loaders: Record<keyof typeof models, DataLoader>;
```
Betree, line 28: This is not necessarily true. `context.loaders.Contributors` for example doesn't relate to a DB model. In the past, FX rate loaders were not either. Loaders debounce expensive async operations and, most of the time, that's indeed meant to be used with DB models. But there are legit cases of using them with 3rd party APIs or other async calls as well.

Ideally, we'll type `loaders.index` and that will solve everything. Until them, I'll suggest we keep the existing

```suggestion
      loaders: Record<string, DataLoader>;
```

test/server/lib/security/expense.test.ts.snap
```diff
@@ -1,6 +1,25 @@
...
+| PAYOUT_METHOD | HIGH   | Payout Method details is being used by other user or collectives |"
+`;
+
+exports[`lib/security/expense checkExpense() and checkExpenses() returns potential threats related to the expense payment (batch) 2`] = `
+"
+| scope | level  | message                                  |
+| ----- | ------ | ---------------------------------------- |
+| USER  | MEDIUM | User is not using 2FA                    |
+| USER  | HIGH   | User has never been paid on the platform |"
+`;
```
Betree, line 20: I've recently discovered [inline snapshots](https://jestjs.io/docs/snapshot-testing#inline-snapshots), they make small snapshots like this way easier to read (no need to open a separate file to get the test result). No need to change it here, it's just a FYI in case you didn't know them already

Betree (approved):
Looking great! Small feedback on the type for `loaders` otherwise this looks production-ready :rocket:

PR #8753, opened 2023-03-30
Make sure `HostCollectiveId` is set for Organizations

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6616

[add-funds.webm](https://user-images.githubusercontent.com/12435965/228704769-84fe0584-5658-41f7-aaec-23bad7ce0bf6.webm)

### DIFF
--- a/migrations/20230329091450-set-hostCollectiveId-in-collectives.js
+++ b/migrations/20230329091450-set-hostCollectiveId-in-collectives.js
@@ -10,6 +10,8 @@ module.exports = {
       WHERE "isHostAccount" = true
       AND "type" = 'ORGANIZATION'
       AND "HostCollectiveId" IS NULL
+      AND "isActive" IS TRUE
+      AND "approvedAt" IS NOT NULL
     `);
   },
 
--- a/migrations/20230329091450-set-hostCollectiveId-in-collectives.js
+++ b/migrations/20230329091450-set-hostCollectiveId-in-collectives.js
@@ -1,0 +1,19 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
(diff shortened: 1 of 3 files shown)


server/graphql/v2/object/AccountPermissions.ts
```diff
@@ -16,7 +16,7 @@ const AccountPermissions = new GraphQLObjectType({
     },
     addFunds: {
       type: new GraphQLNonNull(Permission),
-      description: 'Whether the current user can mark this order as expired',
+      description: 'Whether the current has teh permission to add funds to the account',
```
znarf, line 19:
```suggestion
      description: 'Whether the current has permission to add funds to the account',
```

migrations/20230329091450-set-hostCollectiveId-in-collectives.js
```diff
@@ -0,0 +1,19 @@
...
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+      UPDATE "Collectives"
+      SET
+        "HostCollectiveId" = id
+      WHERE "isHostAccount" = true
+      AND "type" = 'ORGANIZATION'
+      AND "HostCollectiveId" IS NULL
```
znarf, line 12:
```suggestion
      AND "isActive" IS TRUE
      AND "approvedAt" IS NOT NULL
```

znarf (changes requested):
Assumptions needs to be rechecked.

SudharakaP:
> Assumptions needs to be rechecked.

Yes, my mistake. I was under the impression that after making a host one should be able to add funds. But it seems that only after activating the budget the host is able to add funds and receive financial contributions etc.

PR #8758, opened 2023-03-30
Revert "Update Credit Card Upon Detection of Duplicate Credit Card"

Betree:
Reverts opencollective/opencollective-api#8341

Mostly because of the bug introduced by https://github.com/opencollective/opencollective-api/pull/8341#discussion_r1153487704, but while discussing with @znarf we agreed this introduced too much complexity: how will that play with refunds? what are we going to show in the transactions list for the transactions that have deleted payment methods?

The problem this PR was trying to address is very minimal, users can easily resolve it by manually editing their payment methods.


znarf:
Once the revert is shipped, we have at least 100 orders + payment methods that we'll need to correct.

Betree:
E2e error unrelated, merging

PR #8761, opened 2023-03-31
Correct Count When Removing Admins from Child Collectives

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6623

[remove-admin-success.webm](https://user-images.githubusercontent.com/12435965/229070921-0b29ca79-39b6-42bb-ac6b-9e2b1aaf7baf.webm)

### DIFF
--- a/server/graphql/v2/mutation/MemberMutations.js
+++ b/server/graphql/v2/mutation/MemberMutations.js
@@ -1,6 +1,7 @@
 import { GraphQLBoolean, GraphQLNonNull, GraphQLString } from 'graphql';
 import { GraphQLDateTime } from 'graphql-scalars';
 import { pick } from 'lodash';
+import { Op } from 'sequelize';
 
 import ActivityTypes from '../../../constants/activities';
 import POLICIES from '../../../constants/policies';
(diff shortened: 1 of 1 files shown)


server/graphql/v2/mutation/MemberMutations.js
```diff
@@ -238,7 +239,7 @@ const memberMutations = {
         const host = await account.getHostCollective();
         if (host) {
           const adminCount = await models.Member.count({
-            where: { CollectiveId: account.id, role: MemberRoles.ADMIN },
+            where: { CollectiveId: { [Op.or]: [account.id, account.ParentCollectiveId] }, role: MemberRoles.ADMIN },
```
kewitz, line 242: Not sure what happens when `ParentCollectiveId` is null. Maybe we can compact this array just in case?
```suggestion
            where: { CollectiveId: { [Op.or]: compact([account.id, account.ParentCollectiveId]) }, role: MemberRoles.ADMIN },
```

kewitz (approved):

server/graphql/v2/mutation/MemberMutations.js
SudharakaP: Thanks. Added. 👍🏽

PR #8763, opened 2023-03-31
loaders cleanup

znarf:
Follow up from https://github.com/opencollective/opencollective-api/pull/8718

Removed the controversial things from the PR so we can merge this small cleanup.

The goal is to simplify and remove things that are unnecessary or unclear.


Betree (approved):
Looking good!

znarf:
This was a good one but some changes were introduced in the meantime that are making it conflicting and difficult to fix.

Should have merged when this was warm.

PR #8665, opened 2023-03-09
Calculate host fee on pre-tax total 

Betree:
Part of https://github.com/opencollective/opencollective/issues/4492

- [x] Add failing test to confirm the issue
- [x] Fix root cause
- [x] Create a migration to update the `HOST_FEE` transactions

PR #8765, opened 2023-03-31
Add support for taxes on added funds

Betree:
Part of https://github.com/opencollective/opencollective/issues/5927


server/lib/payments.js
```diff
@@ -951,7 +951,7 @@ export const getHostFee = async (order, host = null) => {
 
   const hostFeePercent = await getHostFeePercent(order, host);
 
-  return calcFee(order.totalAmount - platformTip, hostFeePercent);
```
Betree, line 954: Importing this fix from https://github.com/opencollective/opencollective-api/pull/8665

PR #8775, opened 2023-04-04
Model checks

znarf:
We have some consistency issues in the database and this is an attempt at documenting them and their fix.


checks/hosted-collectives.js
```diff
@@ -0,0 +1,51 @@
...
+     AND "isHostAccount" IS FALSE
+     AND "ParentCollectiveId" IS NULL
+     AND "HostCollectiveId" IS NOT NULL
+     AND "type" NOT IN ('ORGANIZATION', 'USER')`,
+    { type: sequelize.QueryTypes.SELECT, raw: true },
+  );
+
+  if (results[0].count > 0) {
+    if (check) {
+      throw new Error('Hosted Collectives without hostFeePercent');
```
Betree, line 25: Is that an issue, given that we default to the host's hostFeePercent?

checks/transactions.js
Betree, line 1: Given what we said about refunds the other day, we may want to check transactions that have a deleted payment method in the future.

Betree (commented):
I was looking for this today in the codebase, it would be great to see it merged. Maybe we can move the code to `scripts/checks` for consistency.

znarf:
@Betree Let's try to pair a bit to complete a first version.

checks/users.js
```diff
@@ -0,0 +1,33 @@
...
+     LEFT JOIN "Collectives" c
+     ON c."id" = u."CollectiveId"
+     WHERE u."deletedAt" IS NULL
+     AND (c."deletedAt" IS NOT NULL OR c."id" IS NULL)`,
+    { type: sequelize.QueryTypes.SELECT, raw: true },
+  );
+
+  if (results[0].count > 0) {
+    if (check) {
+      throw new Error('No Users without a matching Collective');
```
Betree, line 22: One closely related that I wanted to check today is cases of `Users` entries linked to non-`USER` Collectives. We have a few:

```sql
select c.slug, c.type
FROM "Users" u
INNER JOIN "Collectives" c ON u."CollectiveId" = c.id
WHERE c.type != 'USER'
AND c."deletedAt" IS NULL
AND u."deletedAt" IS NULL
```

Betree:
Today I investigated a case where, following manual edits to the PayPal account, OCNZ removed the PayPal webhook that lets us record transactions asynchronously. Thankfully we have a reconciliation script, but this is very problematic and could be easily detected using a similar approach to what's done here.

I would advocate to enlarge the scope of this work to not just "DB checks", but "Sanity checks". For PayPal, I would like to run a job that checks that:
- All hosts connected PayPal credentials are still valid (they should return a 200 for a simple request)
- All hosts with PayPal configured have a valid webhook on their accounts ([validation code already exists](https://github.com/opencollective/opencollective-api/blob/6ec8263b3ccbddec2fc0fb5906e0405efb78159a/server/lib/paypal.ts))

znarf:
> Maybe we can move the code to scripts/checks for consistency.

I will keep it in this new `checks` folder for now as I want to push this as a concept different from scripts.

Betree (approved):
Looking good! The folder pattern is future-proof enough, we'll be able to add PayPal checks in `checks/paypal` and so on.

PR #8779, opened 2023-04-05
Fix total amount reported in taxes-on-host-fee script

Betree:
The issue doesn't appear with my local postgres, but in production returning a `numeric` made sequelize consider the value as a string, which would display `NaN` in summary. Apparently, that's because node-postgres returns a string for the value when its type might be of BIGINT.

> Total amount refunded by script: NZ$NaN

### DIFF
--- a/server/models/Transaction.js
+++ b/server/models/Transaction.js
@@ -864,7 +864,7 @@ Transaction.createHostFeeShareTransactions = async (
   }
 
   // We use the Host Fee amountInHostCurrency/hostCurrency as a basis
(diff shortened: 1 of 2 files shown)


Betree:
Added another metric to clearly surface host fee share

server/models/Transaction.js
```diff
@@ -864,7 +864,7 @@ Transaction.createHostFeeShareTransactions = async (
   }
 
   // We use the Host Fee amountInHostCurrency/hostCurrency as a basis
-  const amount = calcFee(hostFeeTransaction.amountInHostCurrency, hostFeeSharePercent); // TODO without taxes
```
Betree, line 867: This was a TODO I initially added when starting the project, but turned out this was not the best place to fix this

PR #8782, opened 2023-04-05
Make Expense drafts public

Betree:
This will allow everyone to investigate drafts submitted on the platform.

### DIFF
--- a/server/graphql/v2/query/ExpenseQuery.ts
+++ b/server/graphql/v2/query/ExpenseQuery.ts
@@ -1,6 +1,7 @@
 import { GraphQLString } from 'graphql';
 
 import expenseStatus from '../../../constants/expense_status';
+import { allowContextPermission, PERMISSION_TYPE } from '../../common/context-permissions';
 import { ExpenseReferenceInput, fetchExpenseWithReference } from '../input/ExpenseReferenceInput';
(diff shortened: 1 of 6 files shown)


server/graphql/v2/query/ExpenseQuery.ts
```diff
@@ -22,26 +23,22 @@ const ExpenseQuery = {
...
     } else if (args.id) {
-      return req.loaders.Expense.byId.load(args.id);
+      expense = await req.loaders.Expense.byId.load(args.id);
     } else {
       throw new Error('You must either provide an id or an expense');
     }
+
+    if (!expense) {
+      return null;
+    } else if (expense.status === expenseStatus.DRAFT && args.draftKey && expense.data?.draftKey === args.draftKey) {
```
kewitz, line 37: Maybe this logic can be simplified with:
```suggestion
    if (expense?.status === expenseStatus.DRAFT && args.draftKey && expense?.data?.draftKey === args.draftKey) {
```

kewitz (approved):
LGTM

server/graphql/v2/query/ExpenseQuery.ts
Betree: I tend to prefer the first form, it's more verbose but also more explicit IMO

PR #8795, opened 2023-04-10
Add EU to country iso enum

hdiniz:
Adding EU as a valid ISO value, its used by the CountryInput on frontend.


hdiniz:
> It looks strange, EU is not a country. It would be the same as having country=Africa in there.
> 
> What are we trying to solve here?

In frontend, our CountryInput component includes EU. But using this value on this graphql enum breaks the requests. The value is an exception on the [ISO 3166 registry](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Current_codes). Its unusual indeed.

Betree (commented):
My vote would then go to filtering out `EU` from our country picker; we don't want people to submit addresses with "Europe" selected as the country.

PR #8796, opened 2023-04-10
Fix: Wise use consistent source for amountInHostCurrency

kewitz:
This is required because sometimes the transfer has a transfer.sourceValue that does not match the amount debited from the account.

- [ ] Fix impacted transactions in:
    ```
    SELECT e.id, e."CollectiveId", e."createdAt", e."data"#>>'{quote,sourceAmount}' as "quoteAmount", e."data"#>>'{quote,sourceCurrency}' as "sourceCurrency", e."data"#>>'{transfer,sourceValue}' as "transferValue", e."data"#>>'{quote,targetCurrency}' as "targetCurrency", e."data"#>>'{quote,rateType}' as "rateType", t."amountInHostCurrency"
    FROM "Expenses" e
    LEFT JOIN "Transactions" t ON t."ExpenseId" = e."id" AND t."type" = 'DEBIT'
    WHERE
    (e."data"#>>'{transfer,sourceValue}')::NUMERIC < (e."data"#>>'{quote,sourceAmount}')::NUMERIC
    AND ((e."data"#>>'{transfer,sourceValue}')::NUMERIC * -100) != t."amountInHostCurrency"
    ```

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -1858,7 +1858,7 @@ export const getExpenseFees = async (
     collectiveToHostFxRate * (<number>resultFees['paymentProcessorFeeInCollectiveCurrency'] || 0),
   );
   feesInHostCurrency.hostFeeInHostCurrency = Math.round(
-    collectiveToHostFxRate * (<number>fees['hostFeeInCollectiveCurrency'] || 0),
+    collectiveToHostFxRate * (<number>resultFees['hostFeeInCollectiveCurrency'] || 0),
   );
   feesInHostCurrency.platformFeeInHostCurrency = Math.round(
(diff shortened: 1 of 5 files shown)


SudharakaP (approved):
@kewitz : Thanks for doing this as this was impacting the 1k project. Are we handing the already impacted data manually? Just wondering as I didn't see any migration script here.

kewitz:
@SudharakaP I thought I fix affected transactions manually.

SudharakaP:
> @SudharakaP I thought I fix affected transactions manually.

@kewitz : No problem as long as it works, but just note I've done some more yesterday and day before with 1k project so maybe we need to time it if you do manually. So just in case you aren't doing it while the batch is running thus missing some transactions. 😉

PR #8798, opened 2023-04-11
Add two factor method model

hdiniz:

This an attempt to add a model to store user two factor methods.

It deprecates the columns `yubikeyDeviceId` and `twoFactorAuthToken` from the `Users` table.

### UserTwoFactorMethod
| name | type |
| ------- | ------- |
| id       | integer primary key |
| method | string (totp, yubikey_otp, etc) |
| data | jsonb (method dependent) |
| UserId | Users foreignKey |
| createdAt | timestamp | 
| updatedAt | timestamp |
| deletedAt | timestamp |


server/models/UserTwoFactorMethod.ts
```diff
@@ -0,0 +1,101 @@
...
+UserTwoFactorMethod.init(
+  {
+    id: {
+      allowNull: false,
+      autoIncrement: true,
+      primaryKey: true,
+      type: DataTypes.INTEGER,
+    },
+    method: {
+      type: DataTypes.STRING,
```
kewitz, line 67: Add a validator making sure strings are in SupportedTwoFactorMethods.

server/models/UserTwoFactorMethod.ts
```diff
@@ -0,0 +1,101 @@
...
+      allowNull: false,
+    },
+    data: {
+      type: DataTypes.JSONB,
+      allowNull: true,
+    },
```
kewitz, line 73: Can we count on TS to statically validate the data content for us here or should we also validate the structure against what is expected from the method?

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -318,7 +325,13 @@ const accountMutations = {
             throw new ValidationFailed('Invalid 2FA token');
           }
 
-          await user.update({ yubikeyDeviceId: (args.token as string).substring(0, 12) });
+          await UserTwoFactorMethod.create({
+            UserId: user.id,
+            method: TwoFactorMethod.YUBIKEY_OTP,
+            data: {
+              yubikeyDeviceId: (args.token as string).substring(0, 12),
```
kewitz, line 332: Maybe easier to type args?

migrations/20230410210224-add-UserTwoFactorMethod-table.js
```diff
@@ -0,0 +1,66 @@
...
+        type: Sequelize.DATE,
+        defaultValue: Sequelize.fn('now'),
+      },
+      deletedAt: {
+        allowNull: true,
+        type: Sequelize.DATE,
+      },
+    });
+
+    await queryInterface.addIndex('UserTwoFactorMethods', ['UserId', 'method']);
```
kewitz, line 44: Since `userMethods` query by `UserId` only, perhaps we should also index for that:
```suggestion
    await queryInterface.addIndex('UserTwoFactorMethods', ['UserId', 'method']);
     await queryInterface.addIndex('UserTwoFactorMethods', ['UserId']);
```

kewitz: Well, considering that we shouldn't have many methods per user, I think it would even be OK to only index by `UserId` here.

kewitz (commented):
Looks good. I like that the data structure is future-proof.
I would just recommend validating the method at the model level.

server/models/UserTwoFactorMethod.ts
hdiniz: If you are interacting with this model from a typescript file, you will get a static check which is very convenient already. I used a generic type so you get a correct validation for the appropriate two factor method. This is good but bypassed as you know.

I would love to get a runtime schema validation on the data fields in general and I think we can experiment with something here if its simple enough.

server/graphql/v2/mutation/AccountMutations.ts
hdiniz: Will do! I am hoping at some point we can get automatic typing here from the mutation definition.

server/models/UserTwoFactorMethod.ts
hdiniz: @kewitz I added a runtime validator with [zod](https://github.com/colinhacks/zod), it ties up nicely with typescript.

kewitz (approved):

server/lib/security/expense.ts
```diff
@@ -220,7 +220,7 @@ export const checkExpensesBatch = async (
       // Author Security Check: Checks if the author of the expense has 2FA enabled or not.
       addBooleanCheck(
         checks,
-        expense.User.hasTwoFactorAuthentication,
+        await expense.User.hasTwoFactorAuthentication(),
```
Betree, line 223: We're creating an N+1 query issue here; loading 15 expenses will trigger 15 SQL queries. Could we use a loader for this?

server/graphql/v1/CollectiveInterface.js
```diff
@@ -1183,13 +1183,7 @@ const CollectiveFields = () => {
...
         if (req.remoteUser?.isAdmin(collective.id) || req.remoteUser?.isRoot()) {
           if (collective.type === types.USER) {
-            const user = await models.User.findOne({
-              attributes: ['id', 'twoFactorAuthToken'],
-              where: { CollectiveId: collective.id },
-            });
-            if (user.twoFactorAuthToken) {
-              return true;
-            }
+            return await req.remoteUser.hasTwoFactorAuthentication();
```
SudharakaP, line 1186: @hdiniz : The removal of this part introduces a bug. Now note that the 2FA flag for the remote user is returned always. But we need to make sure that the the flag returns whether the 2FA is enabled for the given collective. 

This is important in root actions for example, where with this change it seems that now, even if you disable 2FA for a collective; it still showing as the collective has 2FA enabled (screenshot below; I've disabled for this collective) as the flag is actually coming from my account (where 2FA is enabled). 

image

PR #8802, opened 2023-04-11
Prototype: Wise user-independent tokens

kewitz:
This is a prototype to test if a user-independent token for Wise is viable and if it would solve the problems faced by OC-EU.

This refactor should:
- Affect only collectives with the `settings.transferwise.isolatedUsers` flag
  - Connect/Disconnect only the ConnectedAccount created by the requested user.
  - Pay Wise batchGroups using the dedicated user token.

### DIFF
--- a/server/graphql/v1/mutations/connectedAccounts.js
+++ b/server/graphql/v1/mutations/connectedAccounts.js
@@ -19,6 +19,12 @@ export async function editConnectedAccount(req, connectedAccountData) {
     throw new Unauthorized('Connected account not found');
   } else if (!req.remoteUser.isAdmin(connectedAccount.CollectiveId)) {
     throw new Unauthorized("You don't have permission to edit this connected account");
+  } else if (
+    connectedAccount.service === 'transferwise' &&
+    connectedAccount.collective.settings?.transferwise?.isolateUsers &&
+    req.remoteUser.id !== connectedAccount.CreatedByUserId
+  ) {
+    throw new Unauthorized("You don't have permission to edit this connected account");
   }
 
   await twoFactorAuthLib.enforceForAccount(req, connectedAccount.collective, { onlyAskOnLogin: true });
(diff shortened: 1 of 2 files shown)


server/controllers/connectedAccounts.js
```diff
@@ -116,12 +118,24 @@ export const disconnect = async (req, res) => {
...
+      });
+      if (collective.settings?.transferwise?.isolateUsers) {
+        const account = connectedAccounts.find(ca => ca.CreatedByUserId === remoteUser.id);
+        assert(account, 'No connected account found for this user');
+        await account.destroy();
+      }
+    } else {
+      const account = await models.ConnectedAccount.findOne({
+        where: { service, CollectiveId },
+      });
```
Betree, line 134: That means peeps won't be able to disconnect their accounts if they have the `isolateUsers` flag; shouldn't we do:

```suggestion
    const collective = await models.Collective.findByPk(CollectiveId);
    if (service === 'transferwise' && collective.settings?.transferwise?.isolateUsers) {
      const connectedAccounts = await models.ConnectedAccount.findAll({
        where: { service, CollectiveId },
      });
      const account = connectedAccounts.find(ca => ca.CreatedByUserId === remoteUser.id);
      assert(account, 'No connected account found for this user');
      await account.destroy();
    } else {
      const account = await models.ConnectedAccount.findOne({
        where: { service, CollectiveId },
      });
```

Betree (commented):
It's a nice security mechanism to make sure other admins can't access your connected account (aka. your personal token). I think we should enforce it more globally in `editConnectedAccount` (V1), `deleteConnectedAccount` (V2).
I know it's only a test so it doesn't have to be in this PR, but I'm afraid we'll forget to go back to it if we don't do it now nor document in an issue.

Is there something to be adapted in the frontend, maybe the wording in Admin > Connected Account could document this behavior?

server/controllers/connectedAccounts.js
kewitz: Nice catch, Ben!

PR #8805, opened 2023-04-12
Pass Current Date if Transaction Not Created Yet

SudharakaP:
Related to the ticket; https://opencollective.freshdesk.com/a/tickets/184569

These orders that are mentioned in the ticket is for stripe recurring bank transfers. When the user initiates a bank transfer an order is created in `PROCESSING` status and transaction doesn't exist at this point; hence we return an invalid date (as we calculate it based on `transaction.createdAt`. In this case we need to return the current date, as an acknowledgement that we received the order.

### DIFF
--- a/server/lib/payments.js
+++ b/server/lib/payments.js
@@ -676,7 +676,7 @@ const sendOrderConfirmedEmail = async (order, transaction) => {
     // normal order
     const data = {
       order: order.info,
-      transaction: transaction ? transaction.info : null,
+      transaction: transaction ? transaction.info : { createdAt: new Date() },
       user: user.info,
       collective: collective.info,
       host: host ? host.info : {},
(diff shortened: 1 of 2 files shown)


server/lib/payments.js
```diff
@@ -676,7 +676,7 @@ const sendOrderConfirmedEmail = async (order, transaction) => {
     // normal order
     const data = {
       order: order.info,
-      transaction: transaction ? transaction.info : null,
+      transaction: transaction ? transaction.info : new Date(),
```
hdiniz, line 679: Isn't transaction  in data an object?
```suggestion
      transaction: transaction ? transaction.info : { createdAt: new Date() },
```

hdiniz (approved):

PR #8807, opened 2023-04-13
Migrate Wise affected transactions from #8796

kewitz:


kewitz:
Closing this in favor of another migration that will fix two classes of issues related to expensePayer = PAYEE.

PR #8811, opened 2023-04-14
Order host collection query by rank

hdiniz:
Related https://github.com/opencollective/opencollective/issues/6440

Use host flags `isFirstPartyHost` and `isTrustedHost` to rank host collection results.


server/lib/search.js
```diff
@@ -154,11 +154,24 @@ const getSortSubQuery = (searchTermConditions, orderBy = null) => {
...
     CREATED_AT: `c."createdAt"`,
     HOSTED_COLLECTIVES_COUNT: `
-        SELECT COUNT(1) FROM "Collectives" hosted
-        WHERE hosted."HostCollectiveId" = c.id 
-        AND hosted."deletedAt" IS NULL
-        AND hosted."isActive" = TRUE
-        AND hosted."type" IN ('COLLECTIVE', 'FUND')
+        SELECT
+          ARRAY [
+            -- is host is trusted or first party
```
Betree, line 159:
```suggestion
            -- is host trusted or first party
```

Betree (commented):
It's an elegant solution, but I'm concerned by the fact that `HOSTED_COLLECTIVES_COUNT` won't act as its name suggests anymore. Could we introduce what you implemented under a new sort type, e.g. `HOST_RANK` and keep the existing `HOSTED_COLLECTIVES_COUNT` as a simple count?

Betree (approved):
Perfect!

PR #8813, opened 2023-04-17
Fix Privacy card token update

kewitz:
Resolves https://open-collective.sentry.io/issues/3964501287/?project=5199682

It seems Privacy is returning a different object structure from the one they present on their outdated documentation. I updated the code to try parsing the card token from their actual transaction object and fallback to the previous type.

I also worked a bit on performance so we don't fetch the card twice from the DB.

### DIFF
--- a/cron/hourly/80-reconcile-virtualcard-transactions.js
+++ b/cron/hourly/80-reconcile-virtualcard-transactions.js
@@ -54,7 +54,9 @@ async function reconcileConnectedAccount(connectedAccount) {
           logger.debug(JSON.stringify(transactions, null, 2));
         } else {
           logger.info(`Syncing ${transactions.length} pending transactions...`);
-          await Promise.all(transactions.map(transaction => privacy.processTransaction(transaction)));
+          await Promise.all(
+            transactions.map(transaction => privacy.processTransaction(transaction, undefined, { card })),
+          );
 
           logger.info(`Refreshing card details'...`);
           const [privacyCard] = await privacyLib.listCards(connectedAccount.token, card.id);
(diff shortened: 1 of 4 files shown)


Betree (commented):
I mentioned this issue to @znarf a few weeks ago, but he told me there should not be any more privacy cards. We should look into what these are.

kewitz:
@Betree @znarf This reconciliation job will just fetch the latest 200 transactions of each card and check them against our records. No new transactions have been created in a while but we still have 11 Privacy cards "active" on our DB.
I'm OK with deprecating Privacy if we no longer want to support this, we would just need to deactivate these cards first.

znarf:
The latest transaction was in `2021-12-06` right?

```
SELECT *
FROM "VirtualCards" vc
LEFT JOIN "Expenses" e ON e."VirtualCardId" = vc."id"
LEFT JOIN "Transactions" t ON t."ExpenseId" = e."id"
WHERE vc."provider" = 'PRIVACY'
ORDER BY t."createdAt" DESC
```

I think we can move forward with deprecation and delete any card we still have registered.

kewitz:
@znarf do we care about the existing status of these cards on Privacy? In the sense that these cards could still be used by someone and we wouldn't be able to track them.

znarf:
Most of the non-deleted one are attached to OSC or OCF, 3 are attached to Collectives.

Given that they're not used, we should move forward and suspend them all.

hoxxeinx (approved):
#run world

hoxxeinx (approved):
Whats up?

PR #8822, opened 2023-04-19
GraphQL: prevent importing V1 types from V2

Betree:
:warning: Requires https://github.com/opencollective/opencollective-frontend/pull/8813 to be merged & deployed first
Breaking change announcement: https://opencollective.com/engineering/updates/breaking-change-expense-item-incurredat-input


.eslintrc.json
```diff
@@ -76,6 +76,22 @@
...
+          {
+            "patterns": [
+              {
+                "group": ["**/v1/types"],
+                "message": "GraphQL V1 types should not be used with V2."
+              }
+            ]
+          }
+        ]
+      }
```
SudharakaP, line 94: I was thinking while we are at this, shouldn't we do the other way too? Using V2 types being used with V1?

SudharakaP (approved):

.eslintrc.json
Betree: That wouldn't bother me. V1 is deprecated and we don't expect the same code quality there than in V2.

Andredreyer1: Yes please

PR #8823, opened 2023-04-19
Update Sentry integration

Betree:
Includes some of https://github.com/opencollective/opencollective-api/pull/8452


server/lib/utils.js
```diff
@@ -552,3 +553,22 @@ export const computeDatesAsISOStrings = (startDate, endDate) => {
...
+export const redactSensitiveFields = fastRedact({
+  serialize: false,
+  paths: [
+    'authorization',
+    'Authorization',
+    'AUTHORIZATION',
+    'token',
+    'accessToken',
+    'refreshToken',
+    '["Personal-Token"]',
```
Betree, line 566: I had to come up with this funny syntax because of the dash in the variable name

PR #8825, opened 2023-04-19
Wise: Invalidate cached quote if paymentOption is disabled

kewitz:
Edge-case reported by one of our users. If the host has no balance to cover the transaction, we'll persist a quote that has no valid quote.paymentMethod and throws the invalid balance error persisted there until the quote expires.

### DIFF
--- a/server/paymentProviders/transferwise/index.ts
+++ b/server/paymentProviders/transferwise/index.ts
@@ -96,8 +96,12 @@ async function quoteExpense(
   const isExistingQuoteValid =
     expense.feesPayer !== 'PAYEE' &&
     expense.data?.quote &&
+    // We want a paymentOption to be there
     expense.data.quote['paymentOption'] &&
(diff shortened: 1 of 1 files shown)


server/paymentProviders/transferwise/index.ts
```diff
@@ -96,7 +96,7 @@ async function quoteExpense(
   const isExistingQuoteValid =
     expense.feesPayer !== 'PAYEE' &&
     expense.data?.quote &&
-    expense.data.quote['paymentOption'] &&
+    expense.data.quote['paymentOption']?.disabled === false && // We can not reuse quotes if there's no payment option available
```
znarf, line 99: The code can look incorrect for people not familiar with the data structure. I know it's the same, but the following would look a bit more explicit, maybe with attached comments on each line.

```
// We want a paymentOption to be there
expense.data.quote['paymentOption'] &&
// It needs to be explicitely not disabled
expense.data.quote['paymentOption'].disabled === false
```

PR #8827, opened 2023-04-20
Remove Pia's calendly from emails

Betree:
As discussed in https://opencollective.slack.com/archives/C6JTTA4SK/p1681740377922199


templates/emails/onboarding.day35.active.hbs
Betree, line 1: This email doesn't really seem necessary

znarf (approved):
Pia will be happy!

PR #8828, opened 2023-04-20
Force host currency when approving application

Betree:
The flow with `shouldAutomaticallyApprove` was already covered (see https://github.com/opencollective/opencollective-api/blob/9e4a2597f3662147290caf73b603f5a392b8003b/server/models/Collective.ts#L2100).


znarf (approved):
Great change!

PR #8829, opened 2023-04-20
Add webauthn 2fa

hdiniz:
Add webauth provider using `@simplewebauthn/server`.


server/lib/two-factor-authentication/webauthn.ts
```diff
@@ -0,0 +1,189 @@
...
+    if (!method) {
+      throw new ApolloError('Two-factor authentication code is invalid', 'INVALID_2FA_CODE');
+    }
+
+    await verifyAuthenticationResponse(user, authenticationResponse, req);
+  },
+  async authenticationOptions(user: User, req) {
+    return generateAuthenticationOptions(user, req);
+  },
+};
```
kewitz, line 53: It is better to export named functions due to object reference.

```suggestion
export async function validateToken(user: User, token: Token, req): Promise<void> {
  const authenticationResponse: AuthenticationResponseJSON = JSON.parse(
    Buffer.from(token.code, 'base64').toString('utf-8'),
  );
  
  const method = await UserTwoFactorMethod.findOne<UserTwoFactorMethod<TwoFactorMethod.WEBAUTHN>>({
    where: {
      UserId: user.id,
      method: TwoFactorMethod.WEBAUTHN,
      data: {
        credentialId: authenticationResponse.id,
      },
    },
  });
  
  if (!method) {
    throw new ApolloError('Two-factor authentication code is invalid', 'INVALID_2FA_CODE');
  }
  
  await verifyAuthenticationResponse(user, authenticationResponse, req);
}

export async function authenticationOptions(user: User, req) {
  return generateAuthenticationOptions(user, req);
}
```

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -339,6 +368,34 @@ const accountMutations = {
 
           break;
         }
+        case TwoFactorMethod.WEBAUTHN: {
+          const registrationResult = JSON.parse(Buffer.from(args.token, 'base64').toString('utf8'));
+          const registrationResponse = await webauthn.verifyRegistrationResponse(user, req, registrationResult);
+
+          if (!registrationResponse.verified) {
```
kewitz, line 375: Wouldn't `webauthn.verifyRegistrationResponse` throw before you reach this conditional?

kewitz (commented):
Looks consistent with the existing 2FA lib, would like to have the frontend to test this out and understand how it works.

server/graphql/v2/mutation/AccountMutations.ts
hdiniz: I think so, but not sure. I think I am being defensive here because this might depend on the type of device used: https://github.com/MasterKale/SimpleWebAuthn/blob/7cbc1dc6723294874fe4f576c2d32d07337ec5af/packages/server/src/registration/verifyRegistrationResponse.ts#L208]

We are requesting registration with `attestation = 'direct'`, so it should always be verifiable.

I am thinking about just moving this check to our wrapper, instead of here in the mutation, wdyt?

server/lib/two-factor-authentication/lib.ts
```diff
@@ -184,12 +188,21 @@ async function validateRequest(
       },
     });
 
+    const supportedMethods = await twoFactorMethodsSupportedByUser(remoteUser);
+    const authenticationOptions: Partial<Record<TwoFactorMethod, unknown>> = {};
+    for (const method of supportedMethods) {
+      if (typeof providers[method].authenticationOptions === 'function') {
+        authenticationOptions[method] = await providers[method].authenticationOptions(remoteUser, req);
+      }
+    }
```
kewitz, line 197: Are we supposed to ignore if authenticationOptions is not a function? If we're always expecting a function, it would be simpler to presume this is always a function and be strict about types here.

server/lib/two-factor-authentication/webauthn.ts
```diff
@@ -0,0 +1,282 @@
+import crypto from 'crypto';
+
+import { AsnParser, OctetString } from '@peculiar/asn1-schema';
+import { Certificate } from '@peculiar/asn1-x509';
+import * as simplewebauthn from '@simplewebauthn/server';
+// eslint-disable-next-line import/no-unresolved, node/no-missing-import
```
kewitz, line 6: Do we need this?

server/lib/two-factor-authentication/webauthn.ts
```diff
@@ -0,0 +1,282 @@
...
+  if (!method) {
+    throw new ApolloError('Two-factor authentication code is invalid', 'INVALID_2FA_CODE');
+  }
+
+  await verifyAuthenticationResponse(user, authenticationResponse, req);
+}
+
+export async function authenticationOptions(user: User, req) {
+  return generateAuthenticationOptions(user, req);
+}
```
kewitz, line 59: This method shouldn't exist if it has the same signature as `generateAuthenticationOptions`.

kewitz (commented):
Looking good, this is a big refactor and since this is a security feature, @znarf should also review this.

Added a few suggestions on how to simplify the code but I don't think there's anything critical, good job.

server/controllers/users.js
```diff
@@ -194,16 +204,26 @@ export const exchangeLoginToken = async (req, res, next) => {
 
   const twoFactorAuthenticationEnabled = parseToBoolean(config.twoFactorAuthentication.enabled);
   if (twoFactorAuthenticationEnabled && (await twoFactorAuthLib.userHasTwoFactorAuthEnabled(req.remoteUser))) {
+    const supported2FAMethods = [
+      TwoFactorMethod.RECOVERY_CODE,
+      ...(await twoFactorAuthLib.twoFactorMethodsSupportedByUser(req.remoteUser)),
+    ];
```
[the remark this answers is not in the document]
hdiniz: The reason is that a recovery code option is always available on the API if the user has 2FA enabled, but we dont want to present it on the frontend for every action since using the code resets all methods for the user.
Here its the login route, so we want to present it, I think an alternative would be to always return recovery code as an option and have this decision about presenting it on the frontend based on context.

server/lib/two-factor-authentication/webauthn.ts
hdiniz: Yes, unfortunately these linter rules are not picking up the exports declaration on this dependency's package.json :/ I havent been able to fix this by changing the rules.

server/lib/two-factor-authentication/fido-metadata.ts
```diff
@@ -0,0 +1,99 @@
...
+      },
+    };
+  },
+  {} as Record<string, MetadataEntry>,
+);
+
+export async function downloadFidoMetadata(): Promise<Metadata> {
+  const fidoAlianceMetadataUrl = 'https://mds3.fidoalliance.org';
+
+  const response = await axios.get<string>(fidoAlianceMetadataUrl);
```
znarf, line 52: Could we use node-fetch? while it's not formalised in an RFC, we're trying to not depend on axios for simple things.

server/lib/two-factor-authentication/fido-metadata.ts
```diff
@@ -0,0 +1,99 @@
...
+      ...acc,
+      [current.aaguid]: {
+        ...current,
+      },
+    };
+  },
+  {} as Record<string, MetadataEntry>,
+);
+
+export async function downloadFidoMetadata(): Promise<Metadata> {
```
znarf, line 49: Maybe some comments to explain the purpose of this function and what we're doing here?

server/lib/two-factor-authentication/webauthn.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  await method.update({
+    data: {
+      ...method.data,
+      counter: verificationResponse.authenticationInfo.newCounter,
+    },
+  });
+}
+
+async function getAuthenticatorMetadata(
```
znarf, line 214: This whole function is cryptic, could we have comments explaining what it's doing and implementing?

server/lib/two-factor-authentication/fido-metadata.ts
```diff
@@ -0,0 +1,108 @@
...
+
+          return {
+            ...acc,
+            [current.aaguid]: {
+              ...current,
+            },
+          };
+        },
+        {} as Record<string, MetadataEntry>,
+      );
```
znarf, line 96: Good for me but I was wondering it there was a simpler way to do this. The outcome is really basic and it's many lines of code.

Again, no blocker.

znarf (approved):
Great changes. Looks good to me.

PR #8831, opened 2023-04-20
Modify Webhook Message to Include the Transaction Amount

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6570

image

For a ticket;

image

### DIFF
--- a/server/lib/activities.js
+++ b/server/lib/activities.js
@@ -175,7 +175,11 @@ const doFormatMessage = (activity, format) => {
       return `New ${update} published on ${collective}`;
 
     case activities.COLLECTIVE_MEMBER_CREATED:
-      return `New member ${member} joined ${collective}`;
+      if (amount) {
+        return `${member} just joined ${collective} and contributed with ${formatCurrency(currency, amount)}`;
+      } else {
+        return `New member ${member} joined ${collective}`;
+      }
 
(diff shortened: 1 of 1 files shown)


server/lib/activities.js
```diff
@@ -175,7 +175,14 @@ const doFormatMessage = (activity, format) => {
       return `New ${update} published on ${collective}`;
 
     case activities.COLLECTIVE_MEMBER_CREATED:
-      return `New member ${member} joined ${collective}`;
+      if (amount) {
+        return `${member} just joined ${collective} and contributed  with ${formatCurrency(
```
Betree, line 179:
```suggestion
        return `${member} just joined ${collective} and contributed with ${formatCurrency(
```

server/lib/activities.js
```diff
@@ -175,7 +175,14 @@ const doFormatMessage = (activity, format) => {
...
+          currency,
+          amount,
+        )}! Well done and thank you very much!`;
```
Betree, line 182: :ice_cream: Not sure we need to bring back that part (my personal opinion)

```suggestion
        )}`;
```

SudharakaP: Makes sense. I've felt the same. I've removed that part. 👍🏽

PR #8834, opened 2023-04-20
Add Virtual Card Purchase Webhook Action

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6663

Related Front-End PR: https://github.com/opencollective/opencollective-frontend/pull/8816

### DIFF
--- a/server/lib/activities.js
+++ b/server/lib/activities.js
@@ -184,6 +184,9 @@ const doFormatMessage = (activity, format) => {
         return `New member ${member} joined ${collective}`;
       }
 
+    case activities.VIRTUAL_CARD_PURCHASE:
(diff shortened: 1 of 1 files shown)


SudharakaP:
@kewitz : Is there a way we could test this webhook? I am trying to figure out a way to do a full end-to-end test here. Do we have some test Stripe virtual cards or something we can create? 🤔

kewitz:
Hey, I must have deleted the old notification on this by accident.
You can create a new test in Stripe's test file:
https://github.com/opencollective/opencollective-api/blob/main/test/server/paymentProviders/stripe/virtual-cards.test.js#L51-L87

PR #8701, opened 2023-03-17
Add Support for Root Admins to Submit Tax Forms

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/4805

Related Front-End PR: https://github.com/opencollective/opencollective-frontend/pull/8735

PR #8835, opened 2023-04-21
Ability To Add Tax Form Even if It isn't Requested From the User Yet

SudharakaP:
Related to https://github.com/opencollective/opencollective-api/pull/8701

### DIFF
--- a/server/lib/tax-forms.ts
+++ b/server/lib/tax-forms.ts
@@ -122,10 +122,15 @@ export const setTaxForm = async (account, taxFormLink, year) => {
       year,
       requestStatus: 'RECEIVED',
     });
-    return true;
   } else {
-    throw new Error('No legal document found');
+    await models.LegalDocument.create({
+      requestStatus: 'RECEIVED',
+      documentLink: taxFormLink,
+      year,
+      CollectiveId: account.id,
+    });
   }
+  return true;
 };
 
 export async function sendHelloWorksUsTaxForm(


Betree (commented):
This should ideally be done in a transaction, otherwise there's a risk of having the tax form created two times (concurrency issue).

SudharakaP:
> This should ideally be done in a transaction, otherwise there's a risk of having the tax form created two times (concurrency issue).

@Betree : Thanks. I needed to test this as well property. I've done that and now it's all good. 👍🏽

Betree (approved):
Looks good to me, though there's a small prettier issue.

On a sidenote, this kind of situation is a good case for [`upsert`](https://sequelize.org/api/v6/class/src/model.js~model#static-method-upsert); the code would have been more concise and wouldn't require a transaction.

SudharakaP:
> Looks good to me, though there's a small prettier issue.
> 
> On a sidenote, this kind of situation is a good case for [`upsert`](https://sequelize.org/api/v6/class/src/model.js~model#static-method-upsert); the code would have been more concise and wouldn't require a transaction nor locking the table.

Thanks for the idea. I tried using `upsert` but it seems to need an index on the `LegalDocuments` table which I couldn't figure out for the moment. For now I'll merge this, but if you have some time we can go over this together?

Betree:
No problem, I haven't yet used upsert myself with sequelize so I'm not 100% sure how it works. Even though it's in theory better for that use case, the solution you implemented is perfectly fine, we don't need to iterate.

PR #8843, opened 2023-04-24
Add migration for pending transactions affected by Wise feesPayer bug

kewitz:
This will take care of the pending 14 transactions created in April 13th.


Betree (commented):
Looks like the `tmp_impacted_debits` is not cleared properly, making the CI fails. Otherwise the migration looks good to me, the query for `tmp_impacted_debits` does return 14 entries.

PR #8846, opened 2023-04-25
Fix host-settlement script default PayoutMethod account for bank transfers and others

kewitz:
1. Settle all bank transfers using our USD bank account;
2. Fixes empty array of transactionIds in the settlement expense;
3. Improves debug info in DRY mode

### DIFF
--- a/test/cron/monthly/host-settlement.test.js
+++ b/test/cron/monthly/host-settlement.test.js
@@ -255,10 +255,6 @@ describe('cron/monthly/host-settlement', () => {
     expect(gphHostSettlementExpense).to.have.nested.property('data.isPlatformTipSettlement', true);
   });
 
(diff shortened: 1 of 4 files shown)


test/cron/monthly/host-settlement.test.js
```diff
@@ -255,10 +255,6 @@ describe('cron/monthly/host-settlement', () => {
     expect(gphHostSettlementExpense).to.have.nested.property('data.isPlatformTipSettlement', true);
   });
 
-  it('should use a payout method compatible with the host currency', () => {
-    expect(gphHostSettlementExpense).to.have.property('PayoutMethodId', 2956);
-  });
-
```
Betree, line 261: Rather than removing it, can we update the test to ensure it picks an appropriate payout method?

Betree (approved):
Lgtm!

PR #8851, opened 2023-04-26
Update incomplete Expense side-effects

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6698

### DIFF
--- a/server/graphql/common/comment.ts
+++ b/server/graphql/common/comment.ts
@@ -5,7 +5,7 @@ import { mustBeLoggedInTo } from '../../lib/auth';
 import models from '../../models';
 import Comment from '../../models/Comment';
 import Conversation from '../../models/Conversation';
-import Expense from '../../models/Expense';
+import Expense, { ExpenseStatus } from '../../models/Expense';
 import Update from '../../models/Update';
 import { NotFound, Unauthorized, ValidationFailed } from '../errors';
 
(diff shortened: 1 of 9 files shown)


server/graphql/common/comment.ts
```diff
@@ -144,6 +144,13 @@ async function createComment(commentData, req): Promise<Comment> {
     },
   });
 
+  if (ExpenseId) {
+    const expense = await req.loaders.Expense.byId.load(ExpenseId);
+    if (remoteUser.isAdmin(expense.FromCollectiveId) && expense?.status === ExpenseStatus.INCOMPLETE) {
+      await expense.update({ status: ExpenseStatus.APPROVED });
+    }
+  }
+
```
hdiniz, line 153: In `canMarkAsIncomplete` it seems a host admin can make an expense `INCOMPLETE` from `ERROR` and `PENDING`. In theory possible to have an expense transition like this:

`ERROR` | `PENDING`  -> `INCOMPLETE` by host admin -> `APPROVED` on submitter comment?

Maybe we should restrict the transition in `canMarkAsIncomplete` to `APPROVED` -> `INCOMPLETE` only?

Should we create an activity `COLLECTIVE_EXPENSE_APPROVED` here to mark the transition?

kewitz: Good point, you're totally correct about the PENDING to APPROVED here.
In this context, I think it makes sense to have ERROR -> INCOMPLETE -> APPROVED because will happen during processing.

PR #8855, opened 2023-04-27
Add host collective to virtual card deletion email

SudharakaP:
Something that was requested by some members of the support team due to dealing with virtual card deletions and not knowing which host they belong to. I've added the host to the virtual card deletion email here. 

Related to https://opencollective.freshdesk.com/a/tickets/179940

image

### DIFF
--- a/server/lib/notifications/email.ts
+++ b/server/lib/notifications/email.ts
@@ -578,12 +578,7 @@ export const notifyByEmail = async (activity: Activity) => {
 
     case ActivityTypes.COLLECTIVE_VIRTUAL_CARD_SUSPENDED:
     case ActivityTypes.COLLECTIVE_VIRTUAL_CARD_DELETED: {
-      const collective = await models.Collective.findByPk(activity.CollectiveId);
-      const HostCollectiveId = await collective.getHostCollectiveId();
-      if (HostCollectiveId) {
-        const hostCollective = await models.Collective.findByPk(HostCollectiveId);
-        activity.data.hostCollective = hostCollective.info;
-      }
+      activity.data.hostCollective = activity.data.host;
       await notify.collective(activity, {
         collectiveId: activity.data.collective.id,
       });
--- a/server/lib/notifications/email.ts
+++ b/server/lib/notifications/email.ts
@@ -577,7 +577,13 @@ export const notifyByEmail = async (activity: Activity) => {
       break;
 
     case ActivityTypes.COLLECTIVE_VIRTUAL_CARD_SUSPENDED:
-    case ActivityTypes.COLLECTIVE_VIRTUAL_CARD_DELETED:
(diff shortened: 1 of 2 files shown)


templates/emails/collective.virtualcard.deleted.hbs
```diff
@@ -14,4 +14,8 @@ Subject: ⚠ {{{collective.name}}}'s {{virtualCard.last4}} card was deleted
 <blockquote style="color: #6a737d;font-size: 16px;text-align: left;padding: 0.5em 0.75em;margin: 0 24px;border-left: 3px solid #e4e4e4;white-space: pre-line;font-style: italic;">{{message}}</blockquote>
 {{/if}}
 
+{{#if hostCollective}}
+  <p>Host Collective: {{> linkCollective collective=hostCollective}}</p>
```
kewitz, line 18: Could we either clear the style property of the other P tag or add the same style here to make it consistent?

server/lib/notifications/email.ts
```diff
@@ -577,15 +577,21 @@ export const notifyByEmail = async (activity: Activity) => {
...
 
     case ActivityTypes.COLLECTIVE_VIRTUAL_CARD_SUSPENDED:
-    case ActivityTypes.COLLECTIVE_VIRTUAL_CARD_DELETED:
+    case ActivityTypes.COLLECTIVE_VIRTUAL_CARD_DELETED: {
+      const collective = await models.Collective.findByPk(activity.CollectiveId);
+      const HostCollectiveId = await collective.getHostCollectiveId();
+      if (HostCollectiveId) {
+        const hostCollective = await models.Collective.findByPk(HostCollectiveId);
+        activity.data.hostCollective = hostCollective.info;
+      }
```
kewitz, line 586: This information already exists in `activity.data.host`.

kewitz (changes requested):
I think it is worth reusing the existing information in activity data.

templates/emails/collective.virtualcard.deleted.hbs
SudharakaP: Done. It looks better now with; 

image

server/lib/notifications/email.ts
```diff
@@ -577,15 +577,16 @@ export const notifyByEmail = async (activity: Activity) => {
       break;
 
     case ActivityTypes.COLLECTIVE_VIRTUAL_CARD_SUSPENDED:
-    case ActivityTypes.COLLECTIVE_VIRTUAL_CARD_DELETED:
+    case ActivityTypes.COLLECTIVE_VIRTUAL_CARD_DELETED: {
+      activity.data.hostCollective = activity.data.host;
```
Betree, line 581: I don't get this part; since host is already set in `activity.data.host`, why alias it and not use this variable directly?

PR #8863, opened 2023-05-01
Ignore unknown payment method in stripe mandate webhook

hdiniz:
Sentry: https://open-collective.sentry.io/issues/4003421234/?referrer=slack


Betree (approved):
Looks good :+1:

PR #8864, opened 2023-05-02
Add agreements API

hdiniz:
Project: https://github.com/opencollective/opencollective/issues/4894
Related: https://github.com/opencollective/opencollective/issues/6702


gustavlrsn (commented):
Looks great!

One thing to consider could be the evolution of the GraphQL schema when we later add the ability to link agreements to expenses, maybe it would then make sense to use the `expense.agreements` field for returning those specifically linked to the expense and use `expense.account.agreements` for returning all agreements on the account for the current host.

migrations/20230427070145-add-Agreements-table.js
```diff
@@ -0,0 +1,73 @@
...
+        defaultValue: Sequelize.fn('now'),
+      },
+      updatedAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+        defaultValue: Sequelize.fn('now'),
+      },
+      deletedAt: {
+        allowNull: true,
+        type: Sequelize.DATE,
```
gustavlrsn, line 61: Maybe we could add `archivedAt` to enable archiving

hdiniz: Since it will be an nullable field, can we add it when we enable archiving?

server/graphql/loaders/expenses.ts
```diff
@@ -144,3 +145,25 @@ export const generateExpensesSecurityCheckLoader = req => {
     },
   );
 };
+
+export const expenseAgreementsLoader = new DataLoader<number, Agreement[]>(async (expenseIds: number[]) => {
```
Betree, line 149: Similarly to https://github.com/opencollective/opencollective-api/pull/8894, we should use a generator function here

migrations/20230427070145-add-Agreements-table.js
gustavlrsn: Sounds good!

server/graphql/v2/mutation/AgreementMutations.ts
```diff
@@ -0,0 +1,149 @@
...
+        type: new GraphQLNonNull(Agreement),
+        description: 'Edit an agreement for the given host account. Scope: "host".',
+        args: {
+          agreement: {
+            type: new GraphQLNonNull(AgreementReferenceInput),
+            description: 'Agreement to update.',
+          },
+          title: {
+            type: GraphQLNonEmptyString,
+            description: 'Host where the agreement will be created.',
```
gustavlrsn, line 90:
```suggestion
            description: 'Updated agreement title',
```

server/graphql/v2/mutation/AgreementMutations.ts
```diff
@@ -0,0 +1,149 @@
...
+            HostCollectiveId: host.id,
+            CollectiveId: account.id,
+            UserId: req.remoteUser.id,
+            UploadedFileId: uploadedFile?.id,
+          });
+
+          return agreement;
+        },
+      },
+      editAgreement: {
```
gustavlrsn, line 80: Should we allow attachment updates here? E.g adding an attachment to the agreement, removing one that is there, or uploading/changing to a new file

server/graphql/v2/object/Agreement.ts
```diff
@@ -0,0 +1,42 @@
...
+    },
+    host: {
+      type: new GraphQLNonNull(Host),
+      async resolve(agreement, _, req) {
+        return agreement.Host || (await req.loaders.Collective.byId.load(agreement.HostCollectiveId));
+      },
+    },
+    expiresAt: {
+      type: GraphQLDateTime,
+    },
```
gustavlrsn, line 40: I guess we need something like
```suggestion
    },
    attachment: FileInfo
```

gustavlrsn (commented):
Looks good! 

I think most important to address would probably be to include the uploaded file in the Agreement object. 

Editing the file when editing the agreement could potentially wait since user could just add a new agreement if the file changes.

Also, should we include the activity log entries when creating/editing an agreement or add it in a follow-up PR? I'm fine either way

hdiniz:
@gustavlrsn Added the FileUpload field to the Agreement and the graphql object tests. I will iterate on the activities and attachment on a follow up PR

gustavlrsn (approved):

PR #8868, opened 2023-05-02
Fix tax amount record for Expenses

Betree:
- [x] Fix root cause
- [x] Fix past entries
- [x] Add tests to see how this impacts the balance
- [x] Look at `createTransactionsForManuallyPaidExpense` as it computes the amounts differently


znarf (approved):
Looks good!

PR #8873, opened 2023-05-03
Locations migration followup

gustavlrsn:
Two follow up fixes from the migration in #8626 (not yet deployed to production).
- Updates the migrated Locations 
  - only set `name` and `address` if structured address (collective.data.address) is missing - to prevent having two different addresses saved in the new Location
  - set `address` as `name` if there is no locationName - enabling a better migration to the autocomplete input where the name will be prefilled in the input
- Makes sure to never save an empty `structured` object - but rather use `null` if it is empty.

### DIFF
--- a/migrations/20230503114054-update-location-migration.js
+++ b/migrations/20230503114054-update-location-migration.js
@@ -1,0 +1,41 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    /**
+     * Updates the migrated Locations to only set the name and address when structured address was missing
+     * and also set address as locationName if missing (enabling slightly better migration to autocomplete input)
+     */
+    await queryInterface.sequelize.query(`
+      UPDATE "Locations" l
+      SET 
+        "name" = CASE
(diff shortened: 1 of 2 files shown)


migrations/20230503114054-update-location-migration.js
```diff
@@ -0,0 +1,41 @@
...
+      UPDATE "Locations" l
+      SET 
+        "name" = CASE
+                  WHEN l."structured" IS NULL THEN COALESCE(c."locationName", c."address")
+                  ELSE NULL
+                END,
+        "address" = CASE
+                      WHEN l."structured" IS NULL THEN c."address"
+                      ELSE NULL
+                    END
```
Betree, line 20: Won't this drop all existing names and addresses for locations that don't have a `structured` set?

gustavlrsn: It should not, it's only when `structured` exists that we set the `name` and `address` to null.

gustavlrsn: It should not, it's only when `structured` exists that we set the `name` and `address` to null.

PR #8879, opened 2023-05-04
GraphQLV1: deprecate some root queries

Betree:


server/graphql/v1/queries.js
```diff
@@ -114,6 +114,7 @@ const queries = {
    */
   TransactionInvoice: {
     type: InvoiceType,
+    deprecationReason: '2023-05-04: Please fetch transactions with GraphQL v2',
```
Betree, line 117: Removing the last usage of that query in https://github.com/opencollective/opencollective-pdf/pull/912

PR #8884, opened 2023-05-05
GraphQLV2: add oppositeTransaction to Transaction

Betree:
To facilitate the [migration](https://github.com/opencollective/opencollective-pdf/pull/912) to GraphQL V2 on the PDF service where we have a query like this that makes sure we always generate the receipt from the credit:

```gql
query TransactionInvoice($transactionId: String!) {
  transaction(id: $transactionId) {
    id
    ... on Credit {
      ...ReceiptTransactionFragment
    }
    ... on Debit {
      oppositeTransaction {
        ...ReceiptTransactionFragment
      }
    }
  }
}
```

This query was previously implemented using `relatedTransactions`, but fetching the receipts fields on all transactions has a cost that we can easily avoid by introducing this new field.

### DIFF
--- a/server/graphql/loaders/transactions.ts
+++ b/server/graphql/loaders/transactions.ts
@@ -4,58 +4,91 @@ import { groupBy } from 'lodash';
 import { TransactionKind } from '../../constants/transaction-kind';
 import models, { Op } from '../../models';
 
-export const generateHostFeeAmountForTransactionLoader = (): DataLoader<number, number[]> =>
-  new DataLoader(async (transactions: (typeof models.Transaction)[]) => {
-    const transactionsWithoutHostFee = transactions.filter(transaction => {
-      // Legacy transactions have their host fee set on `hostFeeInHostCurrency`. No need to fetch for them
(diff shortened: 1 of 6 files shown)


server/graphql/loaders/transactions.ts
Betree, line 1: Diff on the two 1st loaders  should be compared without whitespace. I basically added a `cacheKeyFn` to both of them, which should help better cache the results

znarf (approved):
Looks good.

Do we really need a different loader? We could perhaps reuse the relatedTransactions one and filter the result?

Betree:
> Do we really need a different loader? We could perhaps reuse the relatedTransactions one and filter the result?

My goal was to save a bit of RAM by implementing a dedicated one that only loads what's needed, but the performance diff is probably negligible.

znarf:
I was thinking that reusing could potentially improve performance if the 2 loaders are used at the same time. But maybe not a so realistic scenario.

Betree:
Removed the custom loader in the spirit of simplicity + premature optimization is the root of all evil

PR #8885, opened 2023-05-05
Fix(Locations): Prevent saving an empty location input as a new location

gustavlrsn:
We were not properly checking against saving an empty locationInput `{}` as a new location. This fix makes sure to only save new locations when there is some data on the locationInput. 

Also includes a fix to be able to set Collective.countryISO to null if the locationInput is null or empty.


Betree (approved):
Looks good

PR #8910, opened 2023-05-12
Expense: Add security check for tax form

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6733

image

### DIFF
--- a/server/constants/tax-form.ts
+++ b/server/constants/tax-form.ts
@@ -10,10 +10,10 @@ export const TAX_FORM_IGNORED_EXPENSE_TYPES = [
   EXPENSE_TYPE.SETTLEMENT,
   EXPENSE_TYPE.FUNDING_REQUEST, // This one is a simplification, as foundation already collects the document in their process. See https://github.com/opencollective/opencollective/issues/4766
   EXPENSE_TYPE.GRANT, // This one is a simplification, as foundation already collects the document in their process. See https://github.com/opencollective/opencollective/issues/4766
-];
+] as const;
 export const TAX_FORM_IGNORED_EXPENSE_STATUSES = [
   EXPENSE_STATUS.ERROR,
   EXPENSE_STATUS.REJECTED,
   EXPENSE_STATUS.DRAFT,
   EXPENSE_STATUS.UNVERIFIED,
-];
+] as const;
--- a/server/graphql/loaders/expenses.ts
+++ b/server/graphql/loaders/expenses.ts
@@ -10,7 +10,6 @@ import { Activity } from '../../models/Activity';
 import Expense from '../../models/Expense';
 import { ExpenseAttachedFile } from '../../models/ExpenseAttachedFile';
 import { ExpenseItem } from '../../models/ExpenseItem';
-import { LEGAL_DOCUMENT_TYPE } from '../../models/LegalDocument';
 
 import { populateModelAssociations, sortResultsArray } from './helpers';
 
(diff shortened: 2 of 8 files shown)


server/graphql/loaders/expenses.ts
```diff
@@ -81,23 +81,13 @@ export const attachedFiles = (): DataLoader<number, ExpenseAttachedFile[]> => {
...
-/**
- * Loader for expense's requiredLegalDocuments.
- */
-export const requiredLegalDocuments = (): DataLoader<number, string[]> => {
-  return new DataLoader(async (expenseIds: number[]) => {
-    const expenseIdsPendingTaxForm = await queries.getTaxFormsRequiredForExpenses(expenseIds);
-    return expenseIds.map(id => (expenseIdsPendingTaxForm.has(id) ? [LEGAL_DOCUMENT_TYPE.US_TAX_FORM] : []));
-  });
-};
-
```
Betree, line 100: Both loaders were pretty much doing the same thing, I think it was a leftover from GraphQL v1. I unified them to leverage the full power of dataloaders caching.

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -53,10 +54,16 @@ const updateFilterConditionsForReadyToPay = async (where, include, host): Promis
...
-      const expensesIdsSubjectToTaxForm = expensesSubjectToTaxForm.map(expense => expense.id);
-      expensesIdsPendingTaxForms = await queries.getTaxFormsRequiredForExpenses(expensesIdsSubjectToTaxForm);
+      const expenseIds = expensesSubjectToTaxForm.map(expense => expense.id);
+      const requiredLegalDocs = await loaders.Expense.taxFormRequiredBeforePayment.loadMany(expenseIds);
+      for (let i = 0; i < requiredLegalDocs.length; i++) {
+        if (requiredLegalDocs[i]) {
+          expensesIdsPendingTaxForms.add(expenseIds[i]);
+        }
+      }
+
```
Betree, line 66: Modified this code to use the loader here as well. The main benefit is to preload the tax forms status in the loader, so the subsequent calls (in `Expense.requiredLegalDocuments` + `Expense.securityChecks`) directly get the result from the loader's cache.

Confirmed with a log in the SQL query function to make sure it's only called one time:

image

test/server/graphql/loaders/expense.test.js
```diff
@@ -30,7 +30,7 @@ describe('server/graphql/loaders/expense', () => {
     otherPayoutMethod = await fakePayoutMethod({ type: 'OTHER' });
   });
 
-  describe('userTaxFormRequiredBeforePayment', () => {
```
Betree, line 33: These two tests were pointing to the two loaders that we've merged above. They're testing different scenarios that would be difficult to merge, so I instead named them `userTaxFormRequiredBeforePayment (1)` and `userTaxFormRequiredBeforePayment (2)`. Keeping the same `describe` structure for a cleaner diff.

server/constants/tax-form.ts
```diff
@@ -10,10 +10,10 @@ export const TAX_FORM_IGNORED_EXPENSE_TYPES = [
...
   EXPENSE_TYPE.FUNDING_REQUEST, // This one is a simplification, as foundation already collects the document in their process. See https://github.com/opencollective/opencollective/issues/4766
   EXPENSE_TYPE.GRANT, // This one is a simplification, as foundation already collects the document in their process. See https://github.com/opencollective/opencollective/issues/4766
-];
+] as const;
 export const TAX_FORM_IGNORED_EXPENSE_STATUSES = [
   EXPENSE_STATUS.ERROR,
   EXPENSE_STATUS.REJECTED,
   EXPENSE_STATUS.DRAFT,
   EXPENSE_STATUS.UNVERIFIED,
-];
```
kewitz, line 19: 🤌

server/graphql/loaders/expenses.ts
kewitz: Good call.

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -53,10 +52,16 @@ const updateFilterConditionsForReadyToPay = async (where, include, host): Promis
...
     if (expensesSubjectToTaxForm.length > 0) {
-      const expensesIdsSubjectToTaxForm = expensesSubjectToTaxForm.map(expense => expense.id);
-      expensesIdsPendingTaxForms = await queries.getTaxFormsRequiredForExpenses(expensesIdsSubjectToTaxForm);
+      const expenseIds = expensesSubjectToTaxForm.map(expense => expense.id);
+      const requiredLegalDocs = await loaders.Expense.taxFormRequiredBeforePayment.loadMany(expenseIds);
+      for (let i = 0; i < requiredLegalDocs.length; i++) {
+        if (requiredLegalDocs[i]) {
+          expensesIdsPendingTaxForms.add(expenseIds[i]);
+        }
+      }
```
kewitz, line 63: Suggestion for readability:
```suggestion
      requiredLegalDocs.forEach((required, i) => {
        if (required) {
          expensesIdsPendingTaxForms.add(expenseIds[i]);
        }
      });
```

PR #8911, opened 2023-05-12
Add Comment.type and Private Notes to Expenses

kewitz:
Blocker to https://github.com/opencollective/opencollective/issues/6699

### DIFF
--- a/migrations/20230512130000-add-comment-type.js
+++ b/migrations/20230512130000-add-comment-type.js
@@ -11,6 +11,7 @@ module.exports = {
     await queryInterface.addColumn('Comments', 'type', {
       type: Sequelize.ENUM(...Object.values(CommentType)),
       defaultValue: CommentType.COMMENT,
+      allowNull: false,
     });
     await queryInterface.addColumn('CommentHistories', 'type', {
       type: Sequelize.ENUM(...Object.values(CommentType)),
--- a/migrations/20230512130000-add-comment-type.js
+++ b/migrations/20230512130000-add-comment-type.js
@@ -1,0 +1,25 @@
+'use strict';
+
+const CommentType = {
+  COMMENT: 'COMMENT',
+  PRIVATE_NOTE: 'PRIVATE_NOTE',
+};
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.addColumn('Comments', 'type', {
+      type: Sequelize.ENUM(...Object.values(CommentType)),
+      defaultValue: CommentType.COMMENT,
+    });
+    await queryInterface.addColumn('CommentHistories', 'type', {
+      type: Sequelize.ENUM(...Object.values(CommentType)),
+      defaultValue: CommentType.COMMENT,
+    });
+  },
+
+  async down(queryInterface) {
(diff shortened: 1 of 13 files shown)


migrations/20230512130000-add-comment-type.js
```diff
@@ -0,0 +1,25 @@
...
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.addColumn('Comments', 'type', {
+      type: Sequelize.ENUM(...Object.values(CommentType)),
+      defaultValue: CommentType.COMMENT,
+    });
+    await queryInterface.addColumn('CommentHistories', 'type', {
+      type: Sequelize.ENUM(...Object.values(CommentType)),
+      defaultValue: CommentType.COMMENT,
+    });
```
Betree, line 18: If not meant to be nulled

```suggestion
    await queryInterface.addColumn('Comments', 'type', {
      type: Sequelize.ENUM(...Object.values(CommentType)),
      defaultValue: CommentType.COMMENT,
      allowNull: false
    });
    await queryInterface.addColumn('CommentHistories', 'type', {
      type: Sequelize.ENUM(...Object.values(CommentType)),
      defaultValue: CommentType.COMMENT,
      allowNull: false
    });
```

server/graphql/common/expenses.ts
```diff
@@ -165,7 +165,7 @@ const isCollectiveAdmin = async (req: express.Request, expense: Expense): Promis
   return req.remoteUser.isAdminOfCollective(expense.collective);
 };
 
-const isHostAdmin = async (req: express.Request, expense: Expense): Promise<boolean> => {
+export const isHostAdmin = async (req: express.Request, expense: Expense): Promise<boolean> => {
```
Betree, line 168: If used outside of this file, we should call it `isExpenseHostAdmin`. Otherwise another developer will likely use is as `isHostAdmin(req, collective)` at some point.

server/graphql/v2/object/Comment.js
```diff
@@ -28,6 +29,10 @@ const Comment = new GraphQLObjectType({
         type: Account,
         resolve: collectiveResolver,
       },
+      type: {
+        type: CommentType,
```
Betree, line 33: If not nullable

```suggestion
        type: new GraphQLNonNull(CommentType),
```

Betree (commented):
Nice implementation! Looks very clear.

My only blocking feedback is about the performance regression in `Expense.comments.totalCount`.

server/graphql/loaders/comments.ts
```diff
@@ -6,18 +6,22 @@ import { ReactionEmoji } from '../../constants/reaction-emoji';
...
+        where: { [Op.or]: ExpenseIdAndType },
         group: ['ExpenseId'],
-      }).then(results =>
-        sortResults(ExpenseIds, results, 'ExpenseId', { count: 0 }).map(result => <number>result.count),
-      ),
-    ),
+      });
+      return ExpenseIdAndType.map(({ ExpenseId }) => counters.find(c => c.ExpenseId === ExpenseId)?.count || 0);
+    }),
+
```
Betree, line 24: If not passing primitives (numbers, strings) that `dataloader` can natively cache on, you need to provide a `cacheKeyFn`. Otherwise it won't be able to group or cache the data properly.

PR #8912, opened 2023-05-12
feat: mechanism to log timeout queries

znarf:


server/lib/hyperwatch.js
```diff
@@ -97,6 +103,16 @@ const load = async app => {
       } catch (err) {
         reject(err);
       }
+    };
+
+    // 30s timeout
+    const finishTimeout = setTimeout(() => {
+      finish();
+    }, 30 * 1000);
```
Betree, line 111: In case the timeout gets changed during the request

```es6
}, req.socket?.server?.timeout || 30 * 1000);
```

PR #8917, opened 2023-05-15
Fine-tune Sentry for slow requests

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6749

### DIFF
--- a/server/lib/sentry.ts
+++ b/server/lib/sentry.ts
@@ -30,6 +30,7 @@ const getIntegrations = (expressApp = null): Integration[] => {
 };
 
 const TRACES_SAMPLE_RATE = parseFloat(config.sentry.tracesSampleRate) || 0;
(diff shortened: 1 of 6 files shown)


server/lib/sentry.ts
```diff
@@ -29,40 +29,51 @@ const getIntegrations = (expressApp = null): Integration[] => {
...
     },
     beforeSendTransaction(event) {
-      try {
-        const reqBody = JSON.parse(event.request.data);
-        event.request.data = JSON.stringify(utils.redactSensitiveFields(reqBody));
-      } catch (e) {
-        // request data is not a json
-      }
-
+      redactSensitiveDataFromRequest(event.request);
```
Betree, line 60: Small refactor to use the same helper for both issues & performance

znarf (approved):
Great!

PR #8922, opened 2023-05-16
Add route to fetch memory dump

Betree:
Not to be merged now as we already identified a possible source for the leak, but could be useful later


znarf:
Seems this way of doing it is outdated? https://dev.to/bengl/node-js-heap-dumps-in-2021-5akm

Betree:
> Seems this way of doing it is outdated? https://dev.to/bengl/node-js-heap-dumps-in-2021-5akm

Nice, using v8 seems much faster. Will update the PR if we end up needing doing that.

PR #8927, opened 2023-05-16
fix: extract Bearer token without throwing exception

znarf:
in the oAuth protocol, authorization header can be used with the`Basic` scheme, ignore this one


znarf:
E2E not behaving, it should be fine, merging.

PR #8935, opened 2023-05-17
Clean environment variables

znarf:
Make sure to compute non-sensitive defaults so we can remove them from production/staging custom configuration.

### DIFF
--- a/config/production.json
+++ b/config/production.json
@@ -58,27 +58,12 @@
   "thegivingblock": {
     "apiUrl": "https://public-api.tgbwidget.com/v1"
   },
-  "aws": {
-    "s3": {
-      "bucket": "opencollective-production"
-    }
-  },
   "graphql": {
     "cache": {
       "enabled": true
     }
   },
-  "dbEncryption": {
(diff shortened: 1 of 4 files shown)


config/production.json
```diff
@@ -54,5 +57,28 @@
...
+      "bucket": "opencollective-production"
+    }
+  },
+  "graphql": {
+    "cache": {
+      "enabled": true
+    }
+  },
+  "dbEncryption": {
+    "cipher": "AES"
```
Betree, line 72: Maybe this one should stay private as it's security-related

config/production.json
```diff
@@ -54,5 +57,28 @@
...
+  },
+  "captcha": {
+    "enabled": true
+  },
+  "limits": {
+    "ordersPerHour": {
+      "perAccount": 100,
+      "perAccountForCollective": 10,
+      "perEmailForCollective": 5
```
Betree, line 81: Same here; keeping that private would make sense to me. Unless we override in prod.

Betree (approved):

PR #8939, opened 2023-05-19
Optimize Update notifications for large audiences

Betree:
To address a memory exhaustion issue when OSC tries to publish updates.

### DIFF
--- a/server/lib/notifications/email.ts
+++ b/server/lib/notifications/email.ts
@@ -1,7 +1,7 @@
-import Promise from 'bluebird';
 import config from 'config';
 import debugLib from 'debug';
 import { cloneDeep, compact, get } from 'lodash';
+import PQueue from 'p-queue';
 
 import { roles } from '../../constants';
 import ActivityTypes, { TransactionalActivities } from '../../constants/activities';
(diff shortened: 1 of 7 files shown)


server/lib/notifications/email.ts
```diff
@@ -43,44 +45,65 @@ export const notify = {
     options?: NotifySubscribersOptions & {
       user?: User;
       userId?: number;
+      /** If true, will not check if the user is unsubscribed. Only use this if you've already checked it before. */
+      skipUnsubscribedCheck?: boolean;
```
Betree, line 49: We were double-checking for unsubscription here, since already done in `notify.users`

server/lib/notifications/email.ts
```diff
@@ -89,21 +112,24 @@ export const notify = {
...
+        const [userId, user] = isUserId ? [userOrUserId, null] : [userOrUserId.id, userOrUserId];
+        if (
+          !unsubscribed.some(unsubscribedUser => unsubscribedUser.id === userId) && // Unsubscribed
+          !options?.exclude?.includes(userId) // Explicitly excluded
+        ) {
+          queue.add(() => notify.user(activity, { ...options, skipUnsubscribedCheck: true, userId, user }));
+        }
+      }
+
+      await queue.onIdle();
```
Betree, line 132: Two optimizations here:
- We accept a User ID instead of a full User. Since it's anyway reloaded later in the process, there's no benefit in providing a full model at this point.
- User a PQueue so that we only send 50 notifications in parallel. Should prevent memory exhaustion when loading 6000+ users and their collectives at once.

PR #8947, opened 2023-05-22
Remove bluebird

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6764


server/lib/cache/redis.js
```diff
@@ -1,22 +1,20 @@
-import Promise from 'bluebird';
 import debug from 'debug';
 import redis from 'redis';
 
-const asyncRedis = Promise.promisifyAll(redis);
```
znarf, line 5: We're still using Redis 3, which doesn't seem to be async by default:
https://github.com/redis/node-redis/blob/master/docs/v3-to-v4.md

kewitz:
Rebased on top of https://github.com/opencollective/opencollective-api/pull/8965

PR #8951, opened 2023-05-23
Draft: Create Collective Order Stats materialized view

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6737


cron/hourly/50-refresh-materialized-views.js
Betree, line 1: Good call to consolidate in a single file

migrations/20230523170000-create-collective-orders-stats.js
```diff
@@ -0,0 +1,64 @@
...
+          "successRate",
+          AVG("successRate") OVER w AS "accSuccessRate",
+          SUM("numberOfOrders") OVER w AS "accNumberOfOrders"
+        FROM ma
+        WINDOW w AS (PARTITION BY "CollectiveId" ORDER BY "month" DESC)
+      )
+      SELECT
+        "CollectiveId",
+        SUM("accErrorRate") FILTER (WHERE "months" = 1) AS "errorRate1M",
+        SUM("accErrorRate") FILTER (WHERE "months" = 3) AS "errorRate3M",
```
Betree, line 43: Shouldn't that be `<=`? Otherwise, we're going to sum only the data from 3 months ago instead of `SUM(month,month2,month3)`. Same below, I'm surprised by these `=`, but maybe I'm missing something.

```suggestion
        SUM("accErrorRate") FILTER (WHERE "months" <= 3) AS "errorRate3M",
```

kewitz: The accumulated sum is being done in the rolling window above. Here, there will be only a single value per month per collective and I'm using SUM just as an accumulator for grouping.

Betree: So `month = 3` contains the aggregated info for the last 3 months, not just the info for `currentMonth - 3`?

kewitz: Exactly.

kewitz: It is aggregating with:
```sql
SELECT
  "CollectiveId",
  "month",
  RANK() OVER w AS "months",
  "errorRate",
  AVG("errorRate") OVER w AS "accErrorRate",
  "successRate",
  AVG("successRate") OVER w AS "accSuccessRate",
  SUM("numberOfOrders") OVER w AS "accNumberOfOrders"
FROM ma
WINDOW w AS (PARTITION BY "CollectiveId" ORDER BY "month" DESC)
```

This means month 3 has the average of the last 3 months.

PR #8955, opened 2023-05-24
Add accountant to thankyou notification

znarf:
was already done for recurring contributions


Betree (approved):
Good catch, I thought this was already covered!

PR #8958, opened 2023-05-24
README Update

inkade:
Edited the Foreword section to serve more as a introductory section, better suited to newcomers.


znarf:
Thank you for your contribution.

I went through the content you submitted and really appreciate the time and effort you put into it. However, after careful consideration, I've decided not to merge it.

PR #8969, opened 2023-05-26
Fix opentelemetry packages and add configuration

znarf:
Seems there was some problems with the latest updates.

Add configuration, disable by default.

Re-initialize package-lock.json from scratch.


kewitz:
Rebased.

znarf:
Merged main again.

hdiniz (approved):

PR #8976, opened 2023-05-29
Truncate AccountStats.expenseTags results and combine restant results

kewitz:


server/graphql/v2/object/AccountStats.js
```diff
@@ -358,15 +358,34 @@ export const AccountStats = new GraphQLObjectType({
         type: new GraphQLList(AmountStats),
         description: 'Returns expense tags for collective sorted by popularity',
         args: {
-          limit: { type: new GraphQLNonNull(GraphQLInt), defaultValue: 30 },
+          limit: { type: new GraphQLNonNull(GraphQLInt), defaultValue: 100 },
+          truncate: { type: GraphQLInt, defaultValue: 7 },
```
Betree, line 362: Looks like the code is not ready to handle `NULL`, so the arg should be non-null. Also, add a comment to explain what it is?

Betree (commented):
The PR doesn't explain what we're trying to solve

PR #8978, opened 2023-05-30
Update chai to 4.3.7

Betree:
See https://www.chaijs.com/releases/

Symbols are now supported by deep eq, we don't need the adapter anymore.


znarf:
You should update package.json also.

PR #8980, opened 2023-05-30
Cache npm in setup-node action

Betree:
Small simplification. See https://github.com/actions/setup-node#caching-global-packages-data

image


znarf:
Looks like a nice one. Thank you1

PR #8990, opened 2023-05-31
Update redis event handling

znarf:


znarf:
Went through this with Leo live, merging.

PR #9022, opened 2023-06-05
Straighten permissions for `Individual.hasTwoFactorAuth`

Betree:


server/graphql/v2/object/Individual.js
```diff
@@ -102,7 +102,9 @@ export const GraphQLIndividual = new GraphQLObjectType({
         type: GraphQLBoolean,
         async resolve(collective, args, req) {
           const user = await req.loaders.User.byCollectiveId.load(collective.id);
-          return twoFactorAuthLib.userHasTwoFactorAuthEnabled(user);
+          if (req.remoteUser.id === user.id) {
```
znarf, line 105: There is no guarantee that `req.remoteUser` is set right?

znarf, line 105:
```suggestion
          if (req.remoteUser?.id === user.id) {
```

znarf (changes requested):
See my feedback, unless I'm mistaken.

server/graphql/v2/object/Individual.js
Betree: All the queries where we fetch this info are logged in, but you're right - there's no guarantee of it. I copied the snippet from V1's `hasTwoFactorAuth`, I'll fix it there and in `hasPassword` that uses the same logic too.

znarf (approved):

PR #9024, opened 2023-06-05
Type cache lib and remove memcached support

kewitz:


znarf (approved):
Looks good to me!

znarf:
Actually, can you remove the memcache dependency from package.json also?

PR #9025, opened 2023-06-05
Limit SecurityChecks to trusted hosts

Betree:


migrations/20230605122518-dev-mark-hosts-as-trusted.js
Betree, line 1: Had to do this because one of our E2E test for security checks is based on `brusselstogetherasbl`; added `opensource` as well as it seems like a nice default for dev/test envs.

kewitz: Agreed. But also, feel free to update the tests if you think that's simpler.

kewitz (approved):

migrations/20230605122518-dev-mark-hosts-as-trusted.js
Betree: Actually, that's the simplest approach I could think about since there doesn't seem to be any trusted hosts in the default test DB.

PR #9026, opened 2023-06-05
Agreements: allow attachment URL as input

gustavlrsn:
Allowing attachment url's as input to `addAgreement` to be able to use existing attached files form components in the frontend (slightly modified to work for both the expense form and agreement form).


gustavlrsn:
Closing this since we're fixing the frontend support for GraphQL uploads

PR #9027, opened 2023-06-05
Deprecate frontend-based expense verification

kewitz:


Betree (approved):
The code looks good; maybe add a PR description or issue to explain why we're changing this?

PR #9038, opened 2023-06-07
Add last approver to expense

hdiniz:
Related: https://github.com/opencollective/opencollective/issues/6675

Adds the last approver of the expense to the Expense graphql object. Used to display approver on the expense list


Betree (commented):
I'm not very aligned with the idea of showing only one approver; at some point, we know that we'll enable multiple approvers (https://github.com/opencollective/opencollective/issues/3081). For host admins, knowing that an expense got an approval from multiple admins is a precious information.

I would like to suggest showing something like this:
image

Then, instead of populating a new column, we could just rely on the `Activities` table.

kewitz:
I think @betree has a point and we can use this refactor to lay the foundation for multiple approvals. I don't think we should rely on activities due to the size of the table and the time it could take for us to resolve this field. I think denormalizing this into a column in the Expense is a good idea, just make it an array and isolate the logic of managing this array in a method.

Betree:
> I don't think we should rely on activities due to the size of the table and the time it could take for us to resolve this field.

I don't agree with the performance argument. Sure, the `Activities` table is huge, but we have an index on `ExpenseId` that makes it extremely cheap to query this information:

```sql
EXPLAIN ANALYZE
select * from "Activities" where "ExpenseId" = 143189 AND type = 'collective.expense.approved'
```

```
Index Scan using activities__expense_id on ""Activities""  (cost=0.09..11.61 rows=1 width=1445) (actual time=0.012..0.013 rows=1 loops=1)"
  Index Cond: (""ExpenseId"" = 143189)"
  Filter: ((type)::text = 'collective.expense.approved'::text)
  Rows Removed by Filter: 1
Planning Time: 0.074 ms
Execution Time: 0.023 ms

```

:arrow_up: From prod data. Complexity = 10, Execution time = 0.02ms. Performance is not an issue. 

> I think denormalizing this into a column in the Expense is a good idea, just make it an array and isolate the logic of managing this array in a method.

As far as I know, foreign key arrays are not well supported in Postgres (e.g. there's no `onDelete: cascade`, it's not a real relationship).

kewitz:
I agree with @Betree now. We just need a custom method on the expense to get the latest approvals more recent than the last state change.

hdiniz:
I agree with the desired outcome of multiple approvers, but I think it should be attempted as part of this PR. 

Many approvers is more complex when considering unapprovals, and using the activities for this makes it a bit harder.

- Alice approved once.
- Alice for some reason unapproved. (repeat the two steps N times)
- Bob then approved.
- Carlos (fiscal host) unapproved
- Expense was updated.
- Daniel approved.
- Expense is paid.

Alice is definitely not approving, but to know it we need to reduce N activities.
Is Bob still approving? He never explicitly unapproved, was he unapproved by Carlos? Or was his approval stale by the expense details changing?
Daniel was the last approver.

This also needs to happen on a loader, as this information is intended for the expense list.

ATM I propose we keep a  `lastApprovedBy` field that takes the last approver (Daniel) from the expense activity log, it can be deprecated when many approvers is implemented.

Betree:
> Alice is definitely not approving, but to know it we need to reduce N activities.
> Is Bob still approving? He never explicitly unapproved, was he unapproved by Carlos? Or was his approval stale by the expense details changing?
> Daniel was the last approver.

We can start with the assumption that any unapproval invalidates all previous approvals; it sounds reasonable and simple enough to me. Said with code:

```es6
const blockingTypes = ['UNAPPROVED', 'REJECTED']
const typesToCheck = [...blockingTypes, 'APPROVED']
const activities = await findAll({ where: { type: typesToCheck }, order: ['createdAt', 'DESC'] })
const lastApprovedActivities = takeWhile(activities, a => a.type === 'APPROVED');
return lastApprovedActivities.map(a => a.user);
```

This definition could evolve in the future depending on what we decide with https://github.com/opencollective/opencollective/issues/3081, as we'll have to answer this question to decide when an expense needs to move to the "Approved" list.

It's not a blocker for me to go with `lastApprovedBy`, but I would find it a bit sad considering that we know the problem is going to hit us in the near future, and on top of having to resolve it we will have to deprecate and migrate an API field.

hdiniz:
@Betree I updates the field to `approvedBy` and added the filter by activities as you described. Is the `UNAPPROVED` activity sufficient to reset approvers?

server/graphql/v2/object/Expense.ts
```diff
@@ -188,6 +188,25 @@ const GraphQLExpense = new GraphQLObjectType<ExpenseModel, express.Request>({
...
+        type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(GraphQLAccount))),
+        description: 'The accounts who approved this expense',
+        async resolve(expense, _, req) {
+          const activities: Activity[] = await req.loaders.Expense.activities.load(expense.id);
+          const approvalActivitiesSinceLastUnapprovedState = takeRightWhile(
+            activities,
+            a => a.type !== ActivityTypes.COLLECTIVE_EXPENSE_UNAPPROVED,
+          ).filter(a => a.type === ActivityTypes.COLLECTIVE_EXPENSE_APPROVED);
+
+          const approvingUserIds = uniq(approvalActivitiesSinceLastUnapprovedState.map(a => a.UserId));
```
Betree, line 201: Pretty cool that, out of the box, this will return users in the order of approval

kewitz (commented):
This looks really good.
Not relevant to your project, but it is important to make sure we keep `expense.status` as the single source of truth for any business logic, even when we implement multiple approvals.

Betree (approved):
Lgtm!

PR #9040, opened 2023-06-08
Replace axios with node-fetch for webhooks delivery

znarf:
it's a bit difficult to juggle between different http libraries, could be a good thing to consolidate on node-fetch.


Betree (commented):
Things to be careful about:
- Axios has a concept of timeouts. We should make sure we keep the behavior consistent.
- Axios seems to have built-in CSRF protection

znarf:
As axios default timeout is 0 (none), there is no need to support timeouts here for feature parity. Also don't see the need for csrf in that context.

I think this PR is good for review.

Betree:
Note that we'd need to rebase to migrate https://github.com/opencollective/opencollective-api/pull/9102 also

PR #9058, opened 2023-06-12
Fix processPendingOrder validations

Betree:


server/models/Order.ts
```diff
@@ -319,9 +319,9 @@ const Order: ModelStatic<OrderModelInterface> & OrderModelStaticInterface = sequ
       },
     },
 
-    validate: {
-      validateTotalAmount() {
-        if ((this.taxAmount || 0) + (this.platformTipAmount || 0) > this.totalAmount) {
+    hooks: {
+      beforeSave: order => {
+        if ((order.taxAmount || 0) + (order.platformTipAmount || 0) > order.totalAmount) {
```
Betree, line 324: `validate` on model is not working properly, being sometimes called with outdated values

PR #9060, opened 2023-06-12
Workspace: User timeline

kewitz:
Related to https://github.com/opencollective/opencollective/issues/6807


server/graphql/v2/query/index.js
```diff
@@ -32,6 +33,7 @@ const query = {
   account: AccountQuery,
   accounts: AccountsCollectionQuery,
   activities: ActivitiesCollectionQuery,
+  activityTimeline: ActivityTimelineQuery,
```
znarf, line 36: What is the difference between `activities` and `activityTimeline` ? Could we reuse `activities`?

znarf: Or have it served has a resolver under `account` like `account.activities`?

kewitz: I thought it didn't make much sense to have this as a property of the `Account` since this should take into consideration the context of the user resolving it when deciding what should be returned.

The existing `ActivitiesCollectionQuery` feels a bit raw for what we want in the Timeline. It works great for listing the activities that are directly related to a collective (activity log) but the timeline will be a bit more complex.

`activityTimeline` is a sort of opinionated resolver for activities that will return activities from different entities related to the account. Since we will be deciding what to show specific accounts based on different conditions, and this can potentially grow in complexity, I would like to isolate this logic to the backend and keep the arguments to a minimum. For this reason, I added a new resolver dedicated to the timeline.

The name can be a bit confusing as it suggests it is just another activity query, should we rename?

server/graphql/v2/query/ActivityTimelineQuery.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+    slug: {
+      type: GraphQLString,
+      description: `The slug identifying the account (ie: babel for https://opencollective.com/babel)`,
+    },
+    limit: CollectionArgs.limit,
+    offset: CollectionArgs.offset,
+  },
+  resolve: async (_, args, req) => {
+    assert(args.id || args.slug, 'Please provide an id or a slug');
```
znarf, line 47: Is this an established practice? How this will be catch and sent to the client?

server/graphql/v2/query/ActivityTimelineQuery.ts
```diff
@@ -0,0 +1,70 @@
...
+  description: "Activity timeline for the account's workspace",
+  args: {
+    id: {
+      type: GraphQLString,
+      description: `The public id identifying the account (ie: dgm9bnk8-0437xqry-ejpvzeol-jdayw5re)`,
+    },
+    slug: {
+      type: GraphQLString,
+      description: `The slug identifying the account (ie: babel for https://opencollective.com/babel)`,
+    },
```
znarf, line 42: We could use `account: GraphQLAccountReferenceInput` instead. That way you have access to `fetchAccountWithReference` to easily get the account.

server/graphql/v2/query/collection/ActivitiesCollectionQuery.ts
```diff
@@ -52,6 +70,11 @@ const ActivitiesCollectionArgs = {
     defaultValue: null,
     description: 'Only return activities that are of this class/type',
   },
+  timeline: {
+    type: new GraphQLNonNull(GraphQLBoolean),
+    defaultValue: false,
+    description: 'If true, return the timeline of activities for this account',
+  },
```
gustavlrsn, line 77: Nice, seems like a flexible way to test the timeline query generator.

gustavlrsn (approved):

PR #9067, opened 2023-06-13
Allow "new" payout method to be assigned to an edited expense

znarf:
such as ACCOUNT_BALANCE for example in the issue we're trying to fix


Betree (approved):
Makes sense!

kewitz (approved):

PR #9079, opened 2023-06-16
Change Logic of Obtaining Recurring Orders Based on Frequency

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/6826

~~Related Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/8992~~

image

### DIFF
--- a/server/graphql/schemaV2.graphql
+++ b/server/graphql/schemaV2.graphql
@@ -944,6 +944,7 @@ enum ContributionFrequency {
   MONTHLY
   YEARLY
   ONETIME
+  RECURRING
 }
 
 """
--- a/server/graphql/schemaV2.graphql
+++ b/server/graphql/schemaV2.graphql
@@ -944,7 +944,6 @@ enum ContributionFrequency {
   MONTHLY
   YEARLY
   ONETIME
-  RECURRING
 }
 
 """
--- a/server/graphql/v2/enum/ContributionFrequency.ts
+++ b/server/graphql/v2/enum/ContributionFrequency.ts
@@ -8,7 +8,6 @@ export const GraphQLContributionFrequency = new GraphQLEnumType({
     MONTHLY: {},
     YEARLY: {},
     ONETIME: {},
-    RECURRING: {},
   },
 });
 
--- a/server/graphql/v2/enum/ContributionFrequency.ts
+++ b/server/graphql/v2/enum/ContributionFrequency.ts
@@ -8,6 +8,7 @@ export const GraphQLContributionFrequency = new GraphQLEnumType({
     MONTHLY: {},
     YEARLY: {},
     ONETIME: {},
+    RECURRING: {},
   },
 });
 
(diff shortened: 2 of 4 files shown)


server/graphql/v2/enum/ContributionFrequency.ts
```diff
@@ -8,6 +8,7 @@ export const GraphQLContributionFrequency = new GraphQLEnumType({
     MONTHLY: {},
     YEARLY: {},
     ONETIME: {},
+    RECURRING: {},
```
hdiniz, line 11: This enum is used in other queries and mutation inputs. For this I think we should filter with another value in the OrderCollectionQuery.

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -226,11 +226,13 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
...
-      include.push({ model: models.Subscription, required: true, where: { interval: 'month' } });
+      where['interval'] = 'month';
     } else if (args.frequency === 'YEARLY') {
-      include.push({ model: models.Subscription, required: true, where: { interval: 'year' } });
+      where['interval'] = 'year';
+    } else if (args.frequency === 'RECURRING') {
+      where['interval'] = { [Op.in]: ['year', 'month'] };
     }
   } else if (args.onlySubscriptions) {
     include.push({ model: models.Subscription, required: true });
```
hdiniz, line 238: Maybe an alternative would be to make `onlySubscriptions` also return processing subscriptions. 

```suggestion
    include.push({ model: models.Subscription, required: false });
    where[Op.and].push({
       [Op.or]: [
         { ['$subscription.id$']: { [Op.ne]: null }],
         { "interval": { [Op.in]: ['year', 'month'] }, "status": "PROCESSING" }
       ]
    })
```

server/graphql/v2/enum/ContributionFrequency.ts
SudharakaP: Makes sense. I've removed this and used the alternative of making `onlySubscriptions` also return processing subscriptions. Thanks. 👍🏽

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
SudharakaP: Done. 👍🏽

server/graphql/schemaV2.graphql
```diff
@@ -944,6 +944,7 @@ enum ContributionFrequency {
   MONTHLY
   YEARLY
   ONETIME
+  RECURRING
```
hdiniz, line 947: I think this value was left over, right?

hdiniz (approved):

server/graphql/schemaV2.graphql
SudharakaP: Ops, that seems to be the case. I've corrected it. Will merge ones test pass. Thanks for the reviews. 👍🏽

PR #9089, opened 2023-06-16
Iterate on Agreements

Betree:
- Added support for notes
- Add the ability to edit uploaded file
- Add some validations
- Misc code & perf improvements

### DIFF
--- a/server/graphql/v2/mutation/AgreementMutations.ts
+++ b/server/graphql/v2/mutation/AgreementMutations.ts
@@ -1,8 +1,9 @@
 import express from 'express';
-import { GraphQLNonNull } from 'graphql';
+import { GraphQLNonNull, GraphQLString } from 'graphql';
 import { GraphQLDateTime, GraphQLNonEmptyString } from 'graphql-scalars';
 import GraphQLUpload from 'graphql-upload/GraphQLUpload.js';
 import type { FileUpload } from 'graphql-upload/Upload.js';
+import { pick } from 'lodash';
 
 import ActivityTypes from '../../../constants/activities';
 import twoFactorAuthLib from '../../../lib/two-factor-authentication';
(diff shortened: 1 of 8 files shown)


server/graphql/v2/mutation/AgreementMutations.ts
```diff
@@ -47,16 +52,19 @@ export default {
...
 
+      const account = await fetchAccountWithReference(args.account, {
+        throwIfMissing: true,
+      });
+
       if (!req.remoteUser.isAdminOfCollective(host)) {
         throw new Unauthorized('Only host admins can create agreements');
+      } else if (account.HostCollectiveId !== host.id) {
+        // We're not checking `isActive` as it should be possible to create agreements for accounts not approved yet
+        throw new Unauthorized(`Account ${account.name} is not currently hosted by ${host.name}`);
```
Betree, line 63: @hdiniz Related to our conversation to only let users add agreements on accounts they're currently hosting.

server/graphql/v2/mutation/AgreementMutations.ts
```diff
@@ -119,14 +136,16 @@ export default {
...
 
-      const toUpdate: Parameters<AgreementModel['update']>[0] = {};
-
-      if (args.title) {
-        toUpdate.title = args.title;
-      }
-
-      if (args.expiresAt) {
-        toUpdate.expiresAt = args.expiresAt;
+      const toUpdate: Parameters<AgreementModel['update']>[0] = pick(args, ['notes', 'title', 'expiresAt']);
```
Betree, line 139: Simplified a bit the approach by using `pick`. It provides a cleaner behavior out-of-the box:

```es6
// This will prevent setting `expiresAt` to null; should be `if (args.expiresAt !== undefined)`
if (args.expiresAt) {
  toUpdate.expiresAt = args.expiresAt;
}

// Anyway, it's simpler to just use pick as it won't pick the value if it's `undefined`
const toUpdate = pick(values, ['expiresAt']);
```

PR #9098, opened 2023-06-19
enhancement: review sign in activity and lastLoginAt

znarf:
Came to my attention that we were not always updating `lastLoginAt`, and opposite were creating activities when we should not.


server/controllers/users.js
```diff
@@ -243,7 +249,13 @@ export const refreshToken = async (req, res, next) => {
     return next(new BadRequest(errorMessage));
   }
 
-  const token = await req.remoteUser.generateSessionToken({ sessionId: req.jwtPayload?.sessionId });
+  // Context: this is token generation when extending a session
```
Betree, line 252: Great that you added all these comments, they bring some welcome clarity

Betree (approved):

PR #9099, opened 2023-06-19
Add stripe issuing balance to host object

hdiniz:
Related https://github.com/opencollective/opencollective/issues/6833

Add field with stripe issuing balance


server/graphql/v2/object/Host.js
```diff
@@ -313,6 +315,38 @@ export const GraphQLHost = new GraphQLObjectType({
           }
         },
       },
+      stripeIssuingBalance: {
```
znarf, line 318: I'm wondering if we should introduce a dedicated object to retrieve various Stripe related stuff.

```
host {
  id
  stripe {
   issuingBalance {
     amount
     currency
  }
}
```

kewitz: I agree with this and this is consistent with the `Account.transferwise` property that resolves an incapsulated Transferwise object.
Although I understand there's also a `transferwiseBalance` resolver in host, this should ultimately be moved to the Transferwise object.

kewitz (approved):

PR #9102, opened 2023-06-19
Remove node-slack package

hdiniz:
HTTP post from `node-slack` was using the `request` package without setting content-type:
image

Using axios to make same post request
Slack/Discord webhook for a contribution:
image
image

### DIFF
--- a/server/lib/slack.js
+++ b/server/lib/slack.js
@@ -2,12 +2,13 @@
  * Slack message sending logic
  */
 
+import axios from 'axios';
 import config from 'config';
-import Slack from 'node-slack';
 
 import activitiesLib from '../lib/activities';
 
 import logger from './logger';
+import { reportErrorToSentry } from './sentry';
 
 export const OPEN_COLLECTIVE_SLACK_CHANNEL = {
   ABUSE: 'abuse',
@@ -42,7 +43,7 @@ export default {
   /*
    * Posts a message to a slack webhook
    */
-  postMessage(msg, webhookUrl, options) {
+  async postMessage(msg, webhookUrl, options) {
     if (!options) {
       options = {};
     }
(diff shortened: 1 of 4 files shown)


server/lib/slack.js
```diff
@@ -74,13 +74,15 @@ export default {
...
 
-      return new Slack(targetUrl, {}).send(slackOptions, err => {
-        if (err) {
+      return axios
+        .post(targetUrl, slackOptions)
+        .then(() => {
+          return resolve();
+        })
+        .catch(err => {
           logger.warn(`SlackLib.postMessage failed for ${targetUrl}:`, err);
```
Betree, line 83: Can be done in a different PR, but it would be great to do a `reportErrorToSentry` here to catch similar errors in the future:

```suggestion
          logger.warn(`SlackLib.postMessage failed for ${targetUrl}:`, err);
          reportErrorToSentry(err, { extra: { targetUrl, slackOptions } });
```

test/server/lib/slack.test.js
Betree, line 1: Why remove the tests entirely? It seems interesting to make sure the payloads stay the same; it could actually help to detect the kind of issue we just faced. We just have to replace the slack-lib stub with an axios stub.

Betree (approved):
Good job, I think we can resolve https://github.com/opencollective/opencollective/issues/5752 with this one - the real goal was to remove `node-slack` which is not maintained anymore.

Betree:
We should also test it with Slack, just to make sure there's no regression there. There's a webhook configured in prod (abuse channel) if you want to use that.

znarf:
Related: https://github.com/opencollective/opencollective-api/pull/9040

test/server/lib/slack.test.js
hdiniz: Readded it!

hdiniz:
@Betree have another pass please, I just modified the test with axios and moved the lib function to user async

znarf (approved):
Looks good!

PR #9107, opened 2023-06-20
Hugi/make custom data public

aerugo:
This is untested, just making a PR to point out a proposed fix.


Betree (changes requested):
I'm not comfortable with that; the pattern we've used here of naming the field "data" and mimicking order's API implies that fields are private, at least that's what we do in other parts of the API, and I'd prefer to keep it consistent. The field would have had a different name (i.e. `publicInfo`) if it was meant to be public.

Plus, most of the use cases we've defined at the time (like being able to specify an internal payment ID) justify that this field should be private.

PR #9120, opened 2023-06-23
Agreement: expose createdAt

Betree:
Required by latest design

### DIFF
--- a/server/graphql/v2/object/Agreement.ts
+++ b/server/graphql/v2/object/Agreement.ts
@@ -24,7 +24,7 @@ export const GraphQLAgreement = new GraphQLObjectType<AgreementModel, express.Re
       description: 'The time of creation of this agreement',
     },
     createdBy: {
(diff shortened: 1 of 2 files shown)


server/graphql/v2/object/Agreement.ts
```diff
@@ -19,8 +19,12 @@ export const GraphQLAgreement = new GraphQLObjectType<AgreementModel, express.Re
       type: GraphQLString,
       description: 'Additional notes about the agreement for the host admins',
     },
+    createdAt: {
+      type: new GraphQLNonNull(GraphQLDateTime),
+      description: 'The time of creation of this agreement',
+    },
     createdBy: {
-      type: new GraphQLNonNull(GraphQLAccount),
+      type: GraphQLAccount,
```
Betree, line 27: @hdiniz Made this nullable since the resolver can eventually return `null`

PR #9132, opened 2023-06-28
Add filter by virtual card name or last4 digits

hdiniz:
Related https://github.com/opencollective/opencollective/issues/4312


znarf (approved):
Looks good

PR #9141, opened 2023-06-29
Fixes & enhancements for accountants

Betree:


server/graphql/v2/object/Host.js
```diff
@@ -294,7 +294,7 @@ export const GraphQLHost = new GraphQLObjectType({
         description: 'Stripe connected account',
         async resolve(host, _, req) {
           if (!req.remoteUser?.isAdmin(host.id)) {
-            throw new Unauthorized('You need to be logged in as an admin to see the stripe connected account');
+            return null;
```
Betree, line 297: @hdiniz accountants don't have access to this, but we don't want the dashboard to crash when they try to visit the expenses list (which they have access to). The field is nullable, so it looks fine to just return null.

hdiniz (approved):

PR #9147, opened 2023-06-30
Add stripe username to connected account graphql schema

hdiniz:
Used to generate links to correct stripe accounts in frontend.

Related: https://github.com/opencollective/opencollective-frontend/pull/9037

### DIFF
--- a/server/graphql/v2/object/StripeConnectedAccount.ts
+++ b/server/graphql/v2/object/StripeConnectedAccount.ts
@@ -1,4 +1,4 @@
-import { GraphQLObjectType } from 'graphql';
+import { GraphQLObjectType, GraphQLString } from 'graphql';
 
 import { reportErrorToSentry } from '../../../lib/sentry';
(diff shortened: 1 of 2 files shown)


server/graphql/v2/object/StripeConnectedAccount.ts
```diff
@@ -10,6 +10,9 @@ export const GraphQLStripeConnectedAccount = new GraphQLObjectType({
   name: 'StripeConnectedAccount',
   description: 'Stripe connected account properties',
   fields: () => ({
+    username: {
+      type: GraphQLString,
+    },
```
znarf, line 15: What kind of permission model do we have here? Are we relying on `GraphQLStripeConnectedAccount` never being returned if you don't have the permission?

hdiniz: Yes, this object is returned by the Host resolver and the permission check is being done there.

znarf (approved):

PR #9156, opened 2023-07-05
Add virtual card query

hdiniz:
Adds a graphql virtual card query and correct host field object type


server/graphql/v2/query/VirtualCardQuery.ts
```diff
@@ -0,0 +1,48 @@
...
+      defaultValue: true,
+    },
+  },
+  async resolve(_: void, args, req: express.Request): Promise<VirtualCard | null> {
+    if (!checkScope(req, 'virtualCards')) {
+      return null;
+    }
+
+    const virtualCard = await VirtualCard.findByPk(args.virtualCard.id, {
+      include: ['collective', 'host', 'user'],
```
kewitz, line 29:
```suggestion
      include: ['collective', 'host'],
```

kewitz (approved):

PR #9163, opened 2023-07-06
Various fixes for setting tags on contributions

Betree:
Part of https://github.com/opencollective/opencollective/issues/6856

### DIFF
--- a/server/graphql/v2/mutation/TagMutations.ts
+++ b/server/graphql/v2/mutation/TagMutations.ts
@@ -4,11 +4,12 @@ import { GraphQLList, GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'g
 import models from '../../../models';
 import { canEditExpenseTags } from '../../common/expenses';
 import { checkRemoteUserCanUseExpenses, checkRemoteUserCanUseOrders } from '../../common/scope-check';
-import { NotFound, Unauthorized } from '../../errors';
+import { Unauthorized } from '../../errors';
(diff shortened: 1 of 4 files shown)


server/graphql/v2/object/OrderPermissions.ts
```diff
@@ -30,6 +30,15 @@ export const canSeeOrderPrivateActivities = async (req: express.Request, order):
   return isHostAdmin(req, order);
 };
 
+export const canSetOrderTags = async (req: express.Request, order): Promise<boolean> => {
```
kewitz, line 33: For consistency, I would suggest renaming this to `canEditOrderTags`.

server/graphql/v2/mutation/TagMutations.ts
```diff
@@ -49,25 +50,25 @@ const tagMutations = {
...
         const expense = await fetchExpenseWithReference(args.expense, { throwIfMissing: true });
-        await canEditExpenseTags(req, expense, { throw: true });
+        if (!(await canEditExpenseTags(req, expense))) {
+          throw new Unauthorized('You do not have the permissions to set tags on this expense');
+        }
 
         await expense.update({ tags: args.tags });
 
-        return expense;
+        return { expense };
```
kewitz, line 70: This is crazy, how was this working before?

server/graphql/v2/object/OrderPermissions.ts
Betree: The mutation itself is named `setTags`, `canSetTags` seems consistent on this regard

PR #9166, opened 2023-07-07
Create new Account.feed resolver

kewitz:
Addresses performance issues in Workspace timeline.

### DIFF
--- a/server/graphql/v2/enum/ActivityType.ts
+++ b/server/graphql/v2/enum/ActivityType.ts
@@ -21,6 +21,11 @@ export const GraphQLActivityType = new GraphQLEnumType({
   values: Activities,
 });
 
+export const GraphQLActivityClassType = new GraphQLEnumType({
+  name: 'ActivityClassType',
+  values: Classes,
+});
+
 export const GraphQLActivityAndClassesType = new GraphQLEnumType({
   name: 'ActivityAndClassesType',
   values: { ...Activities, ...Classes },
--- a/server/graphql/v2/interface/Account.ts
+++ b/server/graphql/v2/interface/Account.ts
@@ -34,7 +34,7 @@ import { GraphQLActivityClassType } from '../enum/ActivityType';
 import { GraphQLExpenseType } from '../enum/ExpenseType';
 import { GraphQLPaymentMethodService } from '../enum/PaymentMethodService';
 import { GraphQLPaymentMethodType } from '../enum/PaymentMethodType';
-import { idDecode, idEncode, IDENTIFIER_TYPES } from '../identifiers';
+import { idEncode } from '../identifiers';
 import { fetchAccountWithReference, GraphQLAccountReferenceInput } from '../input/AccountReferenceInput';
 import { GraphQLChronologicalOrderInput } from '../input/ChronologicalOrderInput';
 import { GraphQLOrderByInput, ORDER_BY_PSEUDO_FIELDS } from '../input/OrderByInput';
@@ -681,9 +681,9 @@ const accountFieldsDefinition = () => ({
     type: new GraphQLList(GraphQLActivity),
     describe: 'Get the activity feed for this account',
     args: {
-      untilId: {
-        type: GraphQLString,
-        description: 'Only returns activities before ID',
+      dateTo: {
+        type: GraphQLDateTime,
+        description: 'Only returns activities before this date',
       },
       limit: {
         type: GraphQLInt,
@@ -705,8 +705,12 @@ const accountFieldsDefinition = () => ({
         return [];
       }
 
-      const untilId = args.untilId ? idDecode(args.untilId, IDENTIFIER_TYPES.ACTIVITY) : undefined;
-      const feed = await getCollectiveFeed({ collective, untilId, limit: args.limit, classes: args.classes });
+      const feed = await getCollectiveFeed({
+        collective,
+        dateTo: args.dateTo,
+        limit: args.limit,
+        classes: args.classes,
+      });
       if (feed === null) {
         throw new ContentNotReady();
       } else {
--- a/server/graphql/v2/interface/Account.ts
+++ b/server/graphql/v2/interface/Account.ts
@@ -6,6 +6,7 @@ import { Order } from 'sequelize';
 import { types as CollectiveTypes } from '../../../constants/collectives';
 import FEATURE from '../../../constants/feature';
 import { buildSearchConditions } from '../../../lib/search';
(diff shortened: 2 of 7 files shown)


server/graphql/v2/interface/Account.ts
```diff
@@ -674,6 +677,43 @@ const accountFieldsDefinition = () => ({
...
     description: 'Logged-in user permissions on an account',
   },
+  feed: {
+    type: new GraphQLList(GraphQLActivity),
+    describe: 'Get the activity feed for this account',
+    args: {
+      untilId: {
+        type: GraphQLString,
+        description: 'Only returns activities before ID',
+      },
```
znarf, line 687: I prefer timestamp based interfaces.

And generally we're trying to not expose internal autoincrement ids.

server/graphql/v2/enum/ActivityType.ts
```diff
@@ -21,6 +21,11 @@ export const GraphQLActivityType = new GraphQLEnumType({
   values: Activities,
 });
 
+export const GraphQLActivityClassType = new GraphQLEnumType({
+  name: 'ActivityClassType',
+  values: Classes,
```
znarf, line 26: Now that I see this, not a fan of `Classes` as a noon, would it still be time to consider something else?

server/lib/timeline.ts
```diff
@@ -0,0 +1,200 @@
...
+    generateFeed(collective);
+    return null;
+  }
+
+  const wantedTypes = flatten(classes.map(c => ActivitiesPerClass[c]));
+  const ids = [];
+  let offset = 0;
+
+  // This is a thunk responsible for paginating until we have enough results
+  // Notice that we fetch twice the limit to account for activities that might be filtered out
```
znarf, line 162: Multiplying the limit by 2 doesn't looks like a solid approach.

server/lib/timeline.ts
```diff
@@ -0,0 +1,200 @@
...
+      where,
+      order,
+      limit,
+    });
+  }
+
+  // Check if timeline cache exists
+  const cacheKey = `timeline-${collective.slug}`;
+  const cacheExists = await redis.exists(cacheKey);
+  if (!cacheExists) {
```
znarf, line 151: I'm wondering if we still want to call this a `cache`.

server/lib/timeline.ts
```diff
@@ -0,0 +1,200 @@
...
+        })
+    : // When fetching the latest activities, we can paginate in reverse by rank
+      () => redis.zRange(cacheKey, offset, offset + limit * 2, { REV: true });
+
+  let cached = await fetchMore();
+
+  // If we're not paginating, first regenerate cache
+  if (!untilId) {
+    const latestActivity = JSON.parse(cached[0]);
+    await generateFeed(collective, latestActivity.id);
```
znarf, line 179: It's not really "generation", maybe use a different function for that.

server/graphql/v2/interface/Account.ts
kewitz: I'll refactor this, I agree using timestamp is better.

server/lib/timeline.ts
kewitz: If we were to consider the data structure used it would be a sorted set but it is still cached data. I think `cache` conveys better what is going on here.

server/graphql/v2/enum/ActivityType.ts
kewitz: It is, but I think we're past the point where we could do this in this PR. It will touch all the notifications settings stuff, frontend, and backend.

server/lib/timeline.ts
kewitz: I'll remove this, redis should be fast enough.

server/lib/timeline.ts
```diff
@@ -0,0 +1,203 @@
...
+const FEED_LIMIT = 1000;
+
+const createOrUpdateFeed = async (collective: Collective, since?: number) => {
+  const redis = await createRedisClient();
+  const cacheKey = `timeline-${collective.slug}`;
+
+  const where = await makeTimelineQuery(collective);
+  if (since) {
+    where['createdAt'] = { [Op.gt]: since };
+  }
```
Betree, line 105: Quick comparison of `createdAt` VS `id`:

- createdAt

```
"Seq Scan on ""Activities""  (cost=0.00..1355506.28 rows=651 width=1445) (actual time=6060.196..75744.113 rows=9043 loops=1)"
"  Filter: (""createdAt"" >= '2023-07-10 00:00:00+00'::timestamp with time zone)"
  Rows Removed by Filter: 6504857
Planning Time: 1.634 ms
JIT:
  Functions: 2
  Options: Inlining true, Optimization true, Expressions true, Deforming true
  Timing: Generation 0.439 ms, Inlining 95.361 ms, Optimization 34.757 ms, Emission 14.756 ms, Total 145.314 ms
Execution Time: 75773.890 ms
```

- id

```
QUERY PLAN
Result  (cost=1.63..1.63 rows=1 width=4) (actual time=0.013..0.013 rows=1 loops=1)
  InitPlan 1 (returns $0)
    ->  Limit  (cost=0.09..1.63 rows=1 width=4) (actual time=0.010..0.010 rows=1 loops=1)
"          ->  Index Only Scan Backward using ""Activities_pkey"" on ""Activities""  (cost=0.09..37.14 rows=24 width=4) (actual time=0.009..0.009 rows=1 loops=1)"
                Index Cond: ((id IS NOT NULL) AND (id > 6521000))
                Heap Fetches: 1
Planning Time: 0.212 ms
Execution Time: 0.040 ms
```

server/lib/timeline.ts
```diff
@@ -0,0 +1,222 @@
...
+      break;
+    } else {
+      offset += limit;
+      cached = await fetchMore();
+    }
+  }
+
+  // Return the actual activities from the database
+  const activities = await models.Activity.findAll({ where: { id: ids }, order });
+  stop();
```
Betree, line 220: I would argue for renaming, otherwise you have to go up the code to understand what this function is

```suggestion
  stopWatch();
```

Betree (approved):
Looks good to me!

In parallel to this effort, it could be interesting to try and revamp our indexes to make sure they index on type:
- `Index<CollectiveId` => `Index<CollectiveId, type>`
- `Index<FromCollectiveId` => `Index<FromCollectiveId, type>`
- `Index<HostCollectiveId` => `Index<HostCollectiveId, type>`
- `Index<UserId` => `Index<UserId, type>`

PR #9173, opened 2023-07-10
Make sure we clean draft data when editing expense

kewitz:
Fixes https://github.com/opencollective/opencollective/issues/6493


Betree (approved):
Looks good :+1:

PR #9182, opened 2023-07-12
Timeline Feedbacks #1

kewitz:
- [x] Add missing lock to timeline generator
- [x] Fix the wrong stopwatch label
- [x] Debug log stopwatch benchmarks
- [ ] Improve initial query performance
  - Will require another pass after this merge


Betree:
@kewitz There was a prettier issue on this one

PR #9204, opened 2023-07-20
Remove `createOrder` GraphQL V1 mutation

Betree:
Part of https://github.com/opencollective/opencollective/issues/6590

This PR removes the ability to create new pledges by dropping the `createOrder` V1 mutation altogether and converting all tests that were still using it to V2 when appropriate.


test/server/graphql/v1/createOrder.test.js
Betree, line 1: I created tests in `test/server/graphql/v2/mutation/OrderMutations.test.ts` to cover the few cases that existed here and were not covered in the new tests.

test/server/graphql/v1/createOrder.backyourstack.test.js
Betree, line 1: See https://github.com/opencollective/opencollective/issues/7113

server/lib/handlebars.js
```diff
@@ -108,7 +108,11 @@ handlebars.registerHelper('moment', (value, props) => {
 });
 
 handlebars.registerHelper('moment-timezone', value => {
-  return moment().tz(value).format('Z');
+  if (!value) {
+    return '';
+  } else {
+    return moment().tz(value).format('Z');
+  }
```
Betree, line 115: An error popped up during a test here, where we had an event with `timeZone=null`. I've added the timezone to the event, but it seems helpful to not crash in this case anyway.

znarf (approved):
Looks all good!

PR #9215, opened 2023-07-24
Only fiscal-host created pending orders should be editable

kewitz:
Fiscal Host admins shouldn't use this flow to temper with user-created manual contributions.

### DIFF
--- a/server/graphql/v2/object/OrderPermissions.ts
+++ b/server/graphql/v2/object/OrderPermissions.ts
@@ -23,7 +23,7 @@ export const canMarkAsExpired = async (req: express.Request, order): Promise<boo
 };
 
 export const canEdit = async (req: express.Request, order): Promise<boolean> => {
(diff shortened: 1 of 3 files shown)


server/graphql/v2/object/OrderPermissions.ts
```diff
@@ -23,7 +23,7 @@ export const canMarkAsExpired = async (req: express.Request, order): Promise<boo
 };
 
 export const canEdit = async (req: express.Request, order): Promise<boolean> => {
-  return order.status === ORDER_STATUS.PENDING && isHostAdmin(req, order);
+  return order.status === ORDER_STATUS.PENDING && isHostAdmin(req, order) && order.data?.isPendingContribution;
```
Betree, line 26: `isHostAdmin` returns a promise, we must await if not returning it

```suggestion
  return order.status === ORDER_STATUS.PENDING && await isHostAdmin(req, order) && order.data?.isPendingContribution;
```

Betree (changes requested):

PR #9229, opened 2023-07-31
Base `uploadFile` mutation with mock OCR parsing

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6900
Resolve https://github.com/opencollective/opencollective/issues/6902

### DIFF
--- a/server/graphql/v2/mutation/UploadedFileMutations.ts
+++ b/server/graphql/v2/mutation/UploadedFileMutations.ts
@@ -1,0 +1,90 @@
+import { GraphQLBoolean, GraphQLNonNull, GraphQLObjectType } from 'graphql';
+import GraphQLUpload from 'graphql-upload/GraphQLUpload.js';
+import { FileUpload } from 'graphql-upload/Upload';
+
+import { FileKind } from '../../../constants/file-kind';
+import { getExpenseOCRParser } from '../../../lib/ocr';
+import models, { UploadedFile } from '../../../models';
+import { checkRemoteUserCanUseExpenses } from '../../common/scope-check';
+import { GraphQLUploadedFileKind } from '../enum/UploadedFileKind';
+import { GraphQLFileInfo } from '../interface/FileInfo';
+import { GraphQLParseUploadedFileResult, ParseUploadedFileResult } from '../object/ParseUploadedFileResult';
+
+const GraphQLUploadFileResult = new GraphQLObjectType({
+  name: 'UploadFileResult',
+  fields: () => ({
+    file: {
+      type: new GraphQLNonNull(GraphQLFileInfo),
+    },
+    parsingResult: {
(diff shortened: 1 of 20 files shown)


server/graphql/v2/mutation/UploadedFileMutations.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      // Since we're only supporting for expense at the moment, we check the expenses scope
+      checkRemoteUserCanUseExpenses(req);
+
+      // Upload file to S3
+      const uploadedFile = await models.UploadedFile.uploadGraphQl(await args.file, args.kind, req.remoteUser);
+
+      // Parse document
+      let parsingResult: ParseUploadedFileResult | undefined;
+      if (args.parseDocument) {
```
kewitz, line 63: Considering this is paid by request, we should also limit this feature to selected collectives.

test/server/graphql/v2/mutation/UploadedFileMutations.test.ts
```diff
@@ -0,0 +1,148 @@
...
+            success: false,
+            message: 'OCR parsing is not available',
+            confidence: null,
+            expense: null,
+          });
+        });
+      });
+
+      // TODO Mock Klippa service and make sure we call its endpoints here
+      // describe('using Klippa', () => {
```
kewitz, line 142: When you implement this, there's a helper to write integration tests using nock `useIntegrationTestRecorder`.

server/graphql/v2/object/ParseUploadedFileResult.ts
```diff
@@ -0,0 +1,49 @@
+import { GraphQLBoolean, GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
+import { GraphQLDate } from 'graphql-scalars';
+
+import { GraphQLStrictPercentage } from '../scalar/StrictPercentage';
+
+import { GraphQLAmount } from './Amount';
+
+const GraphQLExpenseParsedFileInfo = new GraphQLObjectType({
+  name: 'ExpenseParsedFileInfo',
+  fields: () => ({
```
kewitz, line 10: Can we use typescript to enforce that these fields are compatible with the Expense object?

kewitz (approved):

server/graphql/v2/mutation/UploadedFileMutations.ts
Betree: This will be done as part of https://github.com/opencollective/opencollective/issues/6904, but we don't need to rate limit the mock API.

test/server/graphql/v2/mutation/UploadedFileMutations.test.ts
Betree: Interesting!

PR #9248, opened 2023-08-10
Twitter: Use OAuth 2 & V2 endpoints

Betree:
Require https://github.com/opencollective/opencollective-frontend/pull/9263 (for the new OAuth flow)
Part of https://github.com/opencollective/opencollective/issues/6870


server/lib/express.js
```diff
@@ -77,9 +90,8 @@ export default async function (app) {
   // Setup session (required by passport)
 
   const redisClient = await createRedisClient();
-  if (redisClient) {
-    const store = new RedisStore({ client: redisClient });
-
+  if (redisClient || process.env.OC_ENV === 'development') {
+    const store = !redisClient ? undefined : new RedisStore({ client: redisClient });
```
Betree, line 94: Had to change this because `@superfaceai/passport-twitter-oauth2` relies on session

server/types/express.d.ts
```diff
@@ -1,13 +1,13 @@
...
 
 import PersonalToken from '../models/PersonalToken';
-import User from '../models/User';
+import UserModel from '../models/User';
 import UserToken from '../models/UserToken';
 
 declare global {
   namespace Express {
     interface Request {
-      remoteUser?: User | null;
```
Betree, line 10: Had to change this because we now have `@types/passport` in our dependency tree and it messes up with the global `Express.Request.User` namespace: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/40a9d006062d9dfe259599d73b8f5a091de487a6/types/passport/index.d.ts#L20.

hdiniz (approved):

PR #9251, opened 2023-08-16
Use Transacton ID instead of Transaction Object

SudharakaP:
We need to use the transaction `id` here as otherwise the error appears as; 

```
Charge disputed for transaction #[object SequelizeInstance:Transaction]
```


Betree (approved):
Good catch :+1:

PR #9252, opened 2023-08-17
Block manual payments from creating recurring contributions

kewitz:
Related to https://github.com/opencollective/opencollective/issues/6941


Betree (approved):
This fix looks perfect, however I think that we should resolve https://github.com/opencollective/opencollective/issues/6942 before merging it; otherwise some legit contributions are going to be blocked.

kewitz:
@Betree, thanks for flagging that up. I'll that as well.

PR #9253, opened 2023-08-17
Add rate limiting for Klippa

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6904

### DIFF
--- a/migrations/20230818104410-add-indexes-uploaded-files.js
+++ b/migrations/20230818104410-add-indexes-uploaded-files.js
@@ -1,0 +1,12 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  up: async queryInterface => {
+    await queryInterface.addIndex('UploadedFiles', ['CreatedByUserId'], { concurrently: true });
+  },
+
+  down: async queryInterface => {
+    await queryInterface.removeIndex('UploadedFiles', ['CreatedByUserId']);
+  },
+};
--- a/server/lib/rate-limit.ts
+++ b/server/lib/rate-limit.ts
@@ -33,7 +33,7 @@ export default class RateLimit {
   }
 
   /** Register `nbCalls` in the cache. Returns false if limit has been reached. */
-  public async registerCall(): Promise<boolean> {
+  public async registerCall(nbCalls = 1): Promise<boolean> {
     if (
       this.ignoreTests &&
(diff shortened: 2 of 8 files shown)


test/server/graphql/v2/mutation/UploadedFileMutations.test.ts
```diff
@@ -237,6 +244,87 @@ describe('server/graphql/v2/mutation/UploadedFileMutations', () => {
...
+          const hourlyLimit = config.limits.klippa.perUser.hour;
+
+          // Mock the clock to make sure we don't trigger the hourly limit
+          clock = sinon.useFakeTimers(moment().subtract(20, 'hours').toDate());
+          const validRequests = [];
+          for (let i = 0; i < dailyLimit; i++) {
+            const result = await graphqlQueryV2(uploadFileMutation, args, user);
+            validRequests.push(result);
+            if (validRequests.length % hourlyLimit === 0) {
+              clock.tick(60 * 60 * 1000); // Tick one hour to bypass the DB hourly limit
```
kewitz, line 299: God-damn time-travelers!

migrations/20230818104410-add-indexes-uploaded-files.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  up: async queryInterface => {
+    await queryInterface.addIndex('UploadedFiles', ['CreatedByUserId'], { concurrently: true });
```
kewitz, line 6: 👍🏼

server/lib/rate-limit.ts
```diff
@@ -44,7 +44,7 @@ export default class RateLimit {
     if (count >= this.limit) {
       return false;
     } else {
-      cache.set(this.cacheKey, count + 1, this.expiryTimeInSeconds);
+      cache.set(this.cacheKey, count + nbCalls, this.expiryTimeInSeconds);
```
kewitz, line 47: There's an intrinsic error in this logic:
Since we're always resetting the TTL to the original `expiryTimeInSeconds`, a one-hour rate-limiter actually includes all the requests that were made within one hour of each other, which can include calls much older than one hour ago.

kewitz (approved):

server/lib/rate-limit.ts
kewitz: (Not a blocker, I don't think we need to fix this in this PR)

PR #9260, opened 2023-08-21
Cache & re-use Klippa results

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6601
Part of https://github.com/opencollective/opencollective/issues/6865

I had to update S3 to the lastest SDK since one of the features I needed to use (`ChecksumAlgorithm`) was not working otherwise. I took the opportunity to run the `trash` scripts; they're working as expected.


Betree:
> Although I wasn't able to test it locally, the code is looking good. Kudos for updating the S3 lib.

Thanks for the prompt review!

PR #9268, opened 2023-08-21
Search expense by merchant ids

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/6864

### DIFF
--- a/migrations/20230821182200-index-expense-merchant-id-fields.js
+++ b/migrations/20230821182200-index-expense-merchant-id-fields.js
@@ -1,0 +1,29 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    // Stripe issued Virtual Card transactions
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "expenses_data_transactionId"
+      ON "Expenses"
+      USING BTREE (("data"#>>'{transactionId}') ASC)
+      WHERE "data"#>>'{transactionId}' IS NOT NULL
+      AND "deletedAt" IS NULL
+    `);
+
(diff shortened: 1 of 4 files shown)


migrations/20230821182200-index-expense-merchant-id-fields.js
```diff
@@ -0,0 +1,29 @@
...
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "expenses_data_transactionId"
+      ON "Expenses"
+      USING BTREE (("data"#>>'{transactionId}') ASC)
+      WHERE "data"#>>'{transactionId}' IS NOT NULL
+      AND "deletedAt" IS NULL
+    `);
+
+    // PayPal transactions
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "expenses_data_transaction_id"
```
Betree, line 17: These index names are not super obvious, one could think they're just a duplication mistake. Suggested:

```suggestion
      CREATE INDEX CONCURRENTLY IF NOT EXISTS "expenses_data_stripe_virtual_card_transaction_id"
      ON "Expenses"
      USING BTREE (("data"#>>'{transactionId}') ASC)
      WHERE "data"#>>'{transactionId}' IS NOT NULL
      AND "deletedAt" IS NULL
    `);

    // PayPal transactions
    await queryInterface.sequelize.query(`
      CREATE INDEX CONCURRENTLY IF NOT EXISTS "expenses_data_paypal_transaction_id"
```

Betree (approved):
Looking good!

migrations/20230821182200-index-expense-merchant-id-fields.js
kewitz: This was my first implementation, but then I decided to follow the traditional syntax.

PR #9276, opened 2023-08-23
Add unique index to CollectiveOrderStats materialized view

kewitz:
Should be enough to fix the refresh script error:
image


Betree (approved):
:+1:

PR #9277, opened 2023-08-23
Release OCR for internal hosts

Betree:


server/lib/ocr/index.ts
```diff
@@ -52,7 +53,10 @@ export const runOCRForExpenseFile = async (
 };
 
 export const userCanUseOCR = (user: User | undefined | null): boolean => {
-  return Boolean(config.env.OC_ENV !== 'production' || user?.isRoot());
+  return (
+    config.env.OC_ENV !== 'production' ||
```
kewitz, line 57: We want to enable this in production, right?
```suggestion
```

kewitz (changes requested):

server/lib/ocr/index.ts
Betree: The logic is:
- Either you're not in production =>  we enable klippa directly, no permission check required
- Or we check permissions

PR #9290, opened 2023-08-28
Filter null UserId from Expense.approvedBy

hdiniz:
When an expense is re-approved by the platform after the submitter comments on a incomplete expense, the approval activity contains a null UserId.

Reported on slack: https://opencollective.slack.com/archives/C0HSLRNVC/p1692970040254069


hdiniz:
fix by https://github.com/opencollective/opencollective-api/pull/9291

PR #9302, opened 2023-08-30
Add mutation to follow collectives

hdiniz:
Related https://github.com/opencollective/opencollective/issues/6866

Add mutations to enable adding FOLLOWER members to collectives


server/graphql/loaders/collective.ts
```diff
@@ -99,7 +100,7 @@ export default {
           group: ['MemberCollectiveId'],
           raw: true,
           mapToModel: false,
-          where: { MemberCollectiveId: otherAccountsCollectiveIds },
+          where: { MemberCollectiveId: otherAccountsCollectiveIds, role: { [Op.ne]: MemberRoles.FOLLOWER } },
```
Betree, line 103: :+1:

server/graphql/v2/mutation/MemberMutations.ts
```diff
@@ -31,7 +32,88 @@ const isLastAdmin = async (account, memberAccount) => {
...
+      type: new GraphQLNonNull(GraphQLIndividual),
+    },
+    member: {
+      type: GraphQLMember,
+    },
+  },
+});
+
 const memberMutations = {
+  follow: {
```
Betree, line 60: I'd recommend calling these mutations `followAccount`/`unfollowAccount` for clarity; we already have a `followConversation` mutation and it's possible that we'll add more like that in the future (e.g. `followExpense`, `followThread`, etc).

server/graphql/v2/mutation/MemberMutations.ts
```diff
@@ -31,7 +32,88 @@ const isLastAdmin = async (account, memberAccount) => {
...
+          MemberCollectiveId: req.remoteUser.CollectiveId,
+          CollectiveId: account.id,
+          role: MemberRoles.FOLLOWER,
+        },
+        defaults: {
+          CreatedByUserId: req.remoteUser.id,
+        },
+      });
+
+      return { member, individual: req.remoteUser.getCollective() };
```
Betree, line 85: :ice_cream: Detail/no requested change: I believe `collective` is always preloaded for remote user

```suggestion
      return { member, individual: req.remoteUser.collective };
```

server/graphql/v2/mutation/MemberMutations.ts
```diff
@@ -31,7 +32,88 @@ const isLastAdmin = async (account, memberAccount) => {
...
+      const [member] = await Member.findOrCreate({
+        where: {
+          MemberCollectiveId: req.remoteUser.CollectiveId,
+          CollectiveId: account.id,
+          role: MemberRoles.FOLLOWER,
+        },
+        defaults: {
+          CreatedByUserId: req.remoteUser.id,
+        },
+      });
```
Betree, line 83: We should either:
- Rate limit the endpoint
- Or make sure that we're re-using any previously deleted membership, settings its `deletedAt` to null

In the current state, someone spamming the _Follow/Unfollow_ button would create many entries in the DB.

server/graphql/v2/mutation/MemberMutations.ts
```diff
@@ -260,11 +342,13 @@ const memberMutations = {
       if (args.isInvitation) {
         members = await models.MemberInvitation.destroy({
           where: { MemberCollectiveId: memberAccount.id, CollectiveId: account.id, role: args.role },
+          // @ts-expect-error returning does not exist on destroy type, but works since this is really a bulk update to set deletedAt
```
Betree, line 345: Interesting, I should probably have used the `.update` method when I added this

Betree (approved):
As always, your code looks super good @hdiniz :100: I've added a few notes below on further improvements

PR #9324, opened 2023-09-07
Fix(HostApplications): Only return applications where the associated account has correct HostCollectiveId

gustavlrsn:
Fix to only return applications where the associated collective also has correct HostCollectiveId. 

Also adds a check to see if approvedAt on the Collective is null when filtering on "PENDING", found two instances where there was a "PENDING" application to OSC but the Collective was already approved.

### DIFF
--- a/checks/model/hosted-collectives.js
+++ b/checks/model/hosted-collectives.js
@@ -42,8 +42,46 @@ async function checkHostFeePercent({ fix = false } = {}) {
   }
 }
 
+async function checkPendingHostApplications({ fix = false } = {}) {
+  const message = 'Host Applications with status PENDING but Collective is approved to Host';
+
+  const results = await sequelize.query(
+    `SELECT COUNT(*) AS count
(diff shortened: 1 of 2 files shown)


znarf:
That's some interesting cases to add to the Daily Checks in the database.

https://github.com/opencollective/opencollective-api/blob/main/checks/model/hosted-collectives.js

gustavlrsn:
> That's some interesting cases to add to the Daily Checks in the database.
> 
> https://github.com/opencollective/opencollective-api/blob/main/checks/model/hosted-collectives.js

@znarf added a check and fix for Host Applications that have status PENDING but the Collective is approved

znarf:
> Perhaps we should also soft-delete host applications where the Collective's HostApplicationId does not match what is in the host application? This assuming that conceptually you can only apply to one host at the time and that an application will immediately change your HostCollectiveId.

Make sense to me

znarf (approved):

PR #9357, opened 2023-09-26
Set Order's `updatedAt` when cancelling non-transferable

Betree:


znarf (approved):

BenJam:
thanks for this, the edge case caught me out :D

PR #9364, opened 2023-09-28
Change Tier Currency to Host Currency on Collective Approval

SudharakaP:
Resolve https://github.com/opencollective/opencollective/issues/7015

[tier-change.webm](https://github.com/opencollective/opencollective-api/assets/12435965/96645303-2435-4dfa-b6ba-330357716892)

### DIFF
--- a/server/graphql/v2/mutation/HostApplicationMutations.ts
+++ b/server/graphql/v2/mutation/HostApplicationMutations.ts
@@ -311,6 +311,18 @@ const approveApplication = async (host, collective, req) => {
       { where: { CollectiveId: collective.id }, validate: false, transaction },
     );
 
+    // Convert all active tiers of child collectives to host currency
+    const childCollectives = await models.Collective.findAll({
+      where: { ParentCollectiveId: collective.id },
+      transaction,
+    });
+    childCollectives.map(childCollective => {
+      models.Tier.update(
+        { currency: host.currency },
+        { where: { CollectiveId: childCollective.id }, validate: false, transaction },
+      );
+    });
+
     // Approve the collective
     await collective.update(newAccountData, { transaction });
   });
--- a/server/graphql/v2/mutation/HostApplicationMutations.ts
+++ b/server/graphql/v2/mutation/HostApplicationMutations.ts
@@ -305,6 +305,12 @@ const approveApplication = async (host, collective, req) => {
       transaction,
     });
 
+    // Convert all active tiers to host currency
+    const tiers = await models.Tier.findAll({ where: { CollectiveId: collective.id, deletedAt: null }, transaction });
+    tiers.map(tier => {
+      tier.update({ currency: host.currency }, { transaction });
+    });
+
     // Approve the collective
     await collective.update(newAccountData, { transaction });
   });
--- a/server/graphql/v2/mutation/HostApplicationMutations.ts
+++ b/server/graphql/v2/mutation/HostApplicationMutations.ts
@@ -306,10 +306,10 @@ const approveApplication = async (host, collective, req) => {
     });
 
     // Convert all active tiers to host currency
-    const tiers = await models.Tier.findAll({ where: { CollectiveId: collective.id, deletedAt: null }, transaction });
-    tiers.map(tier => {
-      tier.update({ currency: host.currency }, { transaction });
-    });
+    await models.Tier.update(
+      { currency: host.currency },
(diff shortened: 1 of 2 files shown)


server/graphql/v2/mutation/HostApplicationMutations.ts
```diff
@@ -305,6 +305,12 @@ const approveApplication = async (host, collective, req) => {
       transaction,
     });
 
+    // Convert all active tiers to host currency
+    const tiers = await models.Tier.findAll({ where: { CollectiveId: collective.id, deletedAt: null }, transaction });
+    tiers.map(tier => {
+      tier.update({ currency: host.currency }, { transaction });
+    });
+
```
Betree, line 313: A simpler/more performant alternative to that:

```suggestion
    await models.Tier.update({ currency: host.currency }, { where: { CollectiveId: collective.id }, transaction });

```

Betree (commented):
It looks like we're already supposed to do that in https://github.com/opencollective/opencollective-api/blob/583b63ab30999bbcbca3c49afb2623c744cb761c/server/models/Collective.ts#L2224. `addHost` seems like the right place to do it, as it will do the change no matter which path we use. We should try to understand why it hasn't worked with the case you're referring to.

server/graphql/v2/mutation/HostApplicationMutations.ts
SudharakaP: Thanks. Done.

SudharakaP:
> It looks like we're already supposed to do that in
> 
> https://github.com/opencollective/opencollective-api/blob/583b63ab30999bbcbca3c49afb2623c744cb761c/server/models/Collective.ts#L2224
> . `addHost` seems like the right place to do it, as it will do the change no matter which path we use. We should try to understand why it hasn't worked with the case you're referring to.

@Betree : This `addHost` is only called when the collective applies for being hosted. This is why in the screen-cast, after applying to be hosted, I have to go again to Tiers (which is now in the correct currency) and 

> It looks like we're already supposed to do that in
> 
> https://github.com/opencollective/opencollective-api/blob/583b63ab30999bbcbca3c49afb2623c744cb761c/server/models/Collective.ts#L2224
> . `addHost` seems like the right place to do it, as it will do the change no matter which path we use. We should try to understand why it hasn't worked with the case you're referring to.

The reason is that the function `addHost` is called through `changeHost`. And `changeHost` is only called when we apply for a host; such as in, 

https://github.com/opencollective/opencollective-api/blob/d8e64744ad0f281bb9377f1b4b594074d9fd34ee/server/graphql/v2/mutation/HostApplicationMutations.ts#L145

or when removing the host, 

https://github.com/opencollective/opencollective-api/blob/d8e64744ad0f281bb9377f1b4b594074d9fd34ee/server/graphql/v2/mutation/HostApplicationMutations.ts#L253

or rejecting the host,

https://github.com/opencollective/opencollective-api/blob/d8e64744ad0f281bb9377f1b4b594074d9fd34ee/server/graphql/v2/mutation/HostApplicationMutations.ts#L353

But it's not called when approving the host;

https://github.com/opencollective/opencollective-api/blob/d8e64744ad0f281bb9377f1b4b594074d9fd34ee/server/graphql/v2/mutation/HostApplicationMutations.ts#L274

Betree (commented):
That makes sense, I'm good with this PR then (apart from the small feedback I've shared above).

I've documented https://github.com/opencollective/opencollective/issues/7019 for us to rethink our approach to this.

server/graphql/v2/mutation/HostApplicationMutations.ts
```diff
@@ -305,6 +305,12 @@ const approveApplication = async (host, collective, req) => {
       transaction,
     });
 
+    // Convert all active tiers to host currency
+    await models.Tier.update(
+      { currency: host.currency },
+      { where: { CollectiveId: collective.id }, validate: false, transaction },
```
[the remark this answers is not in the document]
SudharakaP: This won't work as we don't have `ParentCollectiveId` in `Tiers`, so I've changed the code to first get all the child collectives and then update all the tiers relevant to each one of them.

Betree (approved):
I pushed an improvement to:
- Select only the `id` from children
- Update all tiers in a single SQL query rather that `nbChildren` queries
- Not update tiers that already have host's currency
- Add tests

PR #9365, opened 2023-09-29
Fix broken link in Host Application email

gustavlrsn:
Broken link reported on Slack.

Depends on https://github.com/opencollective/opencollective-frontend/pull/9402

Updates the link in `collective.apply.for.host` email template.

### DIFF
--- a/templates/emails/collective.apply.for.host.hbs
+++ b/templates/emails/collective.apply.for.host.hbs
@@ -98,7 +98,7 @@ Dear admin of {{host.name}},
 <center>
   <a
     class="btn blue"
-    href="{{config.host.website}}/{{host.slug}}/admin/host-applications?accountSlug={{collective.slug}}"
+    href="{{config.host.website}}/{{host.slug}}/admin/host-applications?application={{collective.id}}"
   >
     <div style="font-size: 15px;">View application</div>
   </a>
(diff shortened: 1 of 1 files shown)


templates/emails/collective.apply.for.host.hbs
```diff
@@ -98,7 +98,7 @@ Dear admin of {{host.name}},
 <center>
   <a
     class="btn blue"
-    href="{{config.host.website}}/{{host.slug}}/admin/pending-applications#application-{{collective.id}}"
+    href="{{config.host.website}}/{{host.slug}}/admin/host-applications?accountSlug={{collective.slug}}"
```
Betree, line 101: Not a fan of using the `slug` as collective admins can change it and the link would become invalid while the collective still exists and still has an application pending.

Betree:
Would be great if we can have the frontend backward-compatible, to make sure links sent in previous emails still work

templates/emails/collective.apply.for.host.hbs
znarf: Ok, let's send the id then.

gustavlrsn: Good point, fixed this to use the id instead

gustavlrsn:
> Would be great if we can have the frontend backward-compatible, to make sure links sent in previous emails still work

@Betree thanks for the review! Fixed the comments and updated the frontend PR to be fully backwards compatible.

Betree (approved):

PR #9368, opened 2023-10-03
In app survey response mutation

gustavlrsn:
GraphQL mutation to send in-app survey responses to Coda.


Betree (commented):
Why a REST endpoint rather than a GraphQL mutation?

server/controllers/survey.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+  if (!responseId) {
+    return next(new errors.BadRequest('Missing survey response id'));
+  }
+  const CODA_TOKEN = process.env.CODA_IN_APP_SURVEY_TOKEN;
+
+  if (!CODA_TOKEN) {
+    return next(new errors.ServerError('Missing CODA_IN_APP_SURVEY_TOKEN'));
+  }
+  const CODA_DOC_ID = 'nHLKv7oLV0';
```
Betree, line 29: Some rate limits could prevent abuse here. Something as simple as:

```suggestion
  const rateLimit = new RateLimit(`survey-response-${remoteUser.id}`, 60, ONE_HOUR_IN_SECONDS);
  if (!(await rateLimit.registerCall())) {
    throw new RateLimitExceeded();
  }

  const CODA_DOC_ID = 'nHLKv7oLV0';
```

gustavlrsn: Good call, added a rate limit

Betree (approved):
I would have included it in the public API: if we build a mobile app or any other service tomorrow, having a mutation like that could be helpful. Beyond that, I think GraphQL should be the default for all our API <> Frontend communications unless there's a very good reason not to do it.

This is not a blocker on my end, but it would be great to clarify with the team how we want to approach this in the future.

gustavlrsn:
@Betree Good points, keeping everything on the same strategy certainly has its benefits, I'll move it over to GraphQL

Betree (approved):

PR #9371, opened 2023-10-04
Add support root to edit the order (used for root actions)

SudharakaP:
Related to https://github.com/opencollective/opencollective/issues/6641

Front-end PR: https://github.com/opencollective/opencollective-frontend/pull/9418

### DIFF
--- a/server/graphql/v2/mutation/PaymentMethodMutations.js
+++ b/server/graphql/v2/mutation/PaymentMethodMutations.js
@@ -58,7 +58,7 @@ const addCreditCard = {
     checkRemoteUserCanUseOrders(req);
 
     const collective = await fetchAccountWithReference(args.account, { throwIfMissing: true });
-    if (!req.remoteUser?.isAdminOfCollective(collective)) {
+    if (!req.remoteUser?.isAdminOfCollective(collective) && !req.remoteUser?.isRoot()) {
       throw new Forbidden(`Must be an admin of ${collective.name}`);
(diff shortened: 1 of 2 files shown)


znarf (approved):

server/graphql/v2/mutation/PaymentMethodMutations.js
```diff
@@ -58,7 +58,7 @@ const addCreditCard = {
     checkRemoteUserCanUseOrders(req);
 
     const collective = await fetchAccountWithReference(args.account, { throwIfMissing: true });
-    if (!req.remoteUser?.isAdminOfCollective(collective)) {
+    if (!req.remoteUser?.isAdminOfCollective(collective) && !req.remoteUser?.isRoot()) {
```
Betree, line 61: Yes for having support staff update the amount/interval for a contribution or canceling it. 

But why add that to the `addCreditCard` mutation? We definitely don't want support staff to handle credit card numbers and enter them in place of users. Unless I'm missing something?

Betree (changes requested):
Blocking feedback on `addCreditCard` mutation

server/graphql/v2/mutation/PaymentMethodMutations.js
SudharakaP: The reason I added it is once in a while (rarely) I thought that support will need to also change the payment method; like if the user has several credit cards etc. Anyways, I don't think we do this often and can't remember a time this was requested recently so I've removed it.

Betree (approved):
:+1:

PR #9376, opened 2023-10-06
Add more DB checks

znarf:
- No Collectives type=USER without a matching User (no auto fix)
- No non-deleted Members without a matching non-deleted Collective
- No ACCOUNTANT OR ADMIN member with a type different than USER (no auto fix)


checks/model/members.js
```diff
@@ -0,0 +1,69 @@
+import '../../server/env';
+
+import logger from '../../server/lib/logger';
+import { sequelize } from '../../server/models';
+
+async function checkDeletedMembers({ fix = false } = {}) {
+  const message = 'No non-deleted Members without a matching non-deleted Collective (no auto fix)';
```
Betree, line 7: Looks like there is an autofix

```suggestion
  const message = 'No non-deleted Members without a matching non-deleted Collective)';
```

Betree (approved):

PR #9379, opened 2023-10-09
Check if account is frozen when adding funds

kewitz:
Fixes https://github.com/opencollective/opencollective/issues/7037


server/graphql/v2/mutation/AddFundsMutations.ts
```diff
@@ -84,6 +86,10 @@ export const addFundsMutation = {
       );
     }
 
+    if (get(account, `data.features.${FEATURE.ALL}`) === false) {
```
znarf, line 89: Maybe we should add an helper method `account.isFrozen()` for that to make the code clearer.

znarf (approved):

PR #9390, opened 2023-10-11
Base `AccountingCategories` implementation

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7050


server/lib/data.ts
```diff
@@ -45,6 +51,53 @@ export function diffDBEntries<T extends Model>(
...
+/**
+ * Diff two lists of DB objects and returns which ones where created, removed and updated.
+ * Useful for places where we update an attribute by providing a list.
+ *
+ * @param {Array} oldEntries: Existing entries
+ * @param {Array} newEntries: New entries to update with
+ * @param {Array} diffedFields: The fields used to compare objects for ``
+ * @returns [newEntries, removedEntries, updatedEntries]
+ */
+export function richDiffDBEntries<BaseModel extends Model, CreateOrUpdateAttributes = Record<string, unknown>>(
```
Betree, line 69: This function is a copy of `diffDBEntries` with the following adjustments:
- It returns an object with named properties, which is more evident than a tuple
- Inputs can be better typed than `Record` thanks to  the `CreateOrUpdateAttributes` template param
- Update values are passed alongside the original model, which allows for better recording before/after values (i.e., in activities) and can simplify the model update in some cases

Converting existing code would be easy, but I preferred to keep it for a separate PR.

test/test-helpers/fake-data.ts
```diff
@@ -147,6 +166,22 @@ export const fakeHost = async (hostData: Parameters<typeof fakeCollective>[0] =
   });
 };
 
+/** Create a fake host */
+export const fakeActiveHost = async (hostData: Parameters<typeof fakeCollective>[0] = {}) => {
```
Betree, line 170: I wanted to make this the default behavior for `fakeHost` since most hosts are configured as active nowadays, but it breaks a few tests so I've introduced it as a separate helper.

server/models/AccountingCategory.ts
```diff
@@ -0,0 +1,134 @@
...
+    data: AccountingCategoryEditActivityData,
+  ): Promise<void> {
+    await models.Activity.create({
+      type: ActivityTypes.ACCOUNTING_CATEGORIES_EDITED,
+      UserId: user.id,
+      CollectiveId: collective.id,
+      HostCollectiveId: collective.HostCollectiveId,
+      data,
+    });
+  }
```
kewitz, line 57: Not sure we need this shim, especially if we're only calling it from the mutation.

server/models/Collective.ts
```diff
@@ -2383,6 +2385,18 @@ class Collective extends Model<
       await Promise.all(children.map(e => e.changeHost(null, remoteUser, { isChildren: true })));
     }
 
+    // Unlink accounting categories from unpaid expenses, as these categories are host-specific
+    await models.Expense.update(
+      { AccountingCategoryId: null },
+      {
+        where: {
+          CollectiveId: [this.id, ...children.map(c => c.id)],
+          status: { [Op.not]: 'PAID' },
```
kewitz, line 2394: I'm not sure if we have another validation to prevent a collective change while an Expense is still processing, but theoretically:
```suggestion
          status: { [Op.notIn]: ['PAID', 'PROCESSING'] },
```

server/models/RecurringExpense.ts
```diff
@@ -92,6 +92,7 @@ export class RecurringExpense extends Model<RecurringExpenseAttributes, Recurrin
       CollectiveId: this.CollectiveId,
       lastEditedById: expense.UserId,
       incurredAt,
+      AccountingCategoryId: expense.AccountingCategoryId,
```
kewitz, line 95: 💯

server/models/index.ts
```diff
@@ -93,6 +95,11 @@ const models = {
 /**
  * Relationships
  */
+
+// AccountingCategory
+models.AccountingCategory.belongsTo(models.Collective, { foreignKey: 'CollectiveId', as: 'collective' });
+models.AccountingCategory.hasMany(models.Expense, { foreignKey: 'AccountingCategoryId', as: 'expenses' });
```
kewitz, line 101: In order to have `Expenses.getAccountingCategory()` method defined, we need to add
```ts
models.Expense.belongsTo(models.AccountingCategory, { foreignKey: 'AccountingCategoryId', as: 'accountingCategory' });
```

server/models/Collective.ts
Betree: My thinking for limiting this to `PAID` was:
- If an expense is still processing, we should probably not let the unhost happen
- If it's really processing, then it will be paid by the current host, thus it makes sense to keep the accounting category

server/models/index.ts
Betree: I'm not using that anymore, will remove it from the model

server/models/AccountingCategory.ts
Betree: This was initially done to have a common helper between create/edit/remove mutations, but it's indeed not helpful anymore.

Betree: Actually, since we don't yet have good typing for activitie's data, this helper helps to plug typescript on it. I'll therefore keep it for now

PR #9391, opened 2023-10-11
Add Vendors mutations

kewitz:
Solves https://github.com/opencollective/opencollective/issues/7047


server/graphql/v2/input/VendorInput.ts
```diff
@@ -0,0 +1,55 @@
...
+    notes: { type: GraphQLString },
+  }),
+});
+
+export const GraphQLVendorCreateInput = new GraphQLInputObjectType({
+  name: 'VendorCreateInput',
+  fields: () => ({
+    name: { type: new GraphQLNonNull(GraphQLString) },
+    legalName: { type: GraphQLString },
+    tags: { type: new GraphQLList(GraphQLString) },
```
Betree, line 37: I like to use `GraphQLNonEmptyString` for these cases 

```suggestion
    name: { type: new GraphQLNonNull(GraphQLNonEmptyString) },
    legalName: { type: GraphQLString },
    tags: { type: new GraphQLList(GraphQLNonEmptyString) },
```

server/graphql/v2/mutation/VendorMutations.ts
```diff
@@ -0,0 +1,145 @@
...
+        ParentCollectiveId: host.id,
+        ...pick(args.vendor, ['name', 'legalName', 'tags']),
+        data: {
+          vendorInfo: pick(vendorInfo, VENDOR_INFO_FIELDS),
+        },
+      };
+
+      if (vendorInfo.taxType) {
+        assert(vendorInfo.taxId, new BadRequest('taxId is required when taxType is provided'));
+        vendorData.data[vendorInfo.taxType] = vendorInfo.taxId;
```
Betree, line 57: If we aim to re-use existing fields, that would be:

```suggestion
        vendorData.data[vendorInfo.taxType] = { number: vendorInfo.taxId };
```

server/graphql/v2/mutation/VendorMutations.ts
Betree, line 1: We should make sure to check the scope (`enforceScope` or `checkScope`) for all the mutations in this file

server/graphql/v2/mutation/VendorMutations.ts
```diff
@@ -0,0 +1,145 @@
...
+      },
+    },
+    resolve: async (_, args, req) => {
+      if (!req.remoteUser) {
+        throw new Unauthorized('You need to be authenticated to perform this action. Please login and try again.');
+      }
+
+      const vendor = await fetchAccountWithReference(args.vendor, { loaders: req.loaders, throwIfMissing: true });
+      const host = await req.loaders.Collective.byId.load(vendor.ParentCollectiveId);
+      assert(host, new NotFound('Vendor host not found'));
```
Betree, line 85: We should check the account type

```suggestion
      assert(host, new NotFound('Vendor host not found'));
      assert(vendor.type === 'VENDOR', new ValidationFailed('Account is not a vendor'));
```

server/graphql/v2/mutation/VendorMutations.ts
```diff
@@ -0,0 +1,145 @@
...
+        ...pick(args.vendor, ['name', 'legalName', 'tags']),
+        data: {
+          ...vendor.data,
+          vendorInfo: { ...vendor.data?.vendorInfo, ...pick(vendorInfo, VENDOR_INFO_FIELDS) },
+        },
+      };
+
+      if (vendorInfo.taxType) {
+        assert(vendorInfo.taxId, new BadRequest('taxId is required when taxType is provided'));
+        vendorData.data[vendorInfo.taxType] = vendorInfo.taxId;
```
Betree, line 102:
```suggestion
        vendorData.data[vendorInfo.taxType] = { number: vendorInfo.taxId };
```

Betree (commented):
A few checks to add but the core looks perfect :ok_hand:

server/graphql/v2/mutation/VendorMutations.ts
kewitz: @Betree should we presume `type: 'OWN'` as well?

Betree: We can, yes

PR #9407, opened 2023-10-19
Script to check & fix duplicate members

Betree:
Related issues:
- https://github.com/opencollective/opencollective/issues/7067
- https://github.com/opencollective/opencollective/issues/4648
- https://github.com/opencollective/opencollective/issues/4648

### DIFF
--- a/server/models/MigrationLog.ts
+++ b/server/models/MigrationLog.ts
@@ -11,6 +11,7 @@ export enum MigrationLogType {
   BAN_ACCOUNTS = 'BAN_ACCOUNTS',
   MOVE_ORDERS = 'MOVE_ORDERS',
   MOVE_EXPENSES = 'MOVE_EXPENSES',
+  MODEL_FIX = 'MODEL_FIX',
 }
 
 export type MigrationLogDataForMergeAccounts = {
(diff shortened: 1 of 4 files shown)


server/models/MigrationLog.ts
```diff
@@ -11,6 +11,7 @@ export enum MigrationLogType {
   BAN_ACCOUNTS = 'BAN_ACCOUNTS',
   MOVE_ORDERS = 'MOVE_ORDERS',
   MOVE_EXPENSES = 'MOVE_EXPENSES',
+  CHECK_SCRIPT = 'CHECK_SCRIPT',
```
znarf, line 14: Maybe not a fan of `CHECK_SCRIPT`.

`CHECK_MODEL`, `MODEL_CHECK`, `MODEL_FIX`, `CHECK_MODEL_FIX`, `MODEL_CHECK_FIX` not sure.

Betree: What about `DATABASE_CHECK` or `DATABASE_FIX`?

znarf: Why not `DATABASE_FIX`, _fix_ is better than _check_.

I prefer `MODEL_FIX` because the checks are in the "checks/model" folder and not "checks/database".

Betree:
I'll be merging and running this one, then we'll investigate new cases as they emerge

PR #9409, opened 2023-10-19
Log which user triggered "activate host"

Betree:


server/models/Collective.ts
```diff
@@ -901,7 +901,7 @@ class Collective extends Model<
 
   // run when attaching a Stripe Account to this user/organization collective
   // this Payment Method will be used for "Add Funds"
-  becomeHost = async function () {
+  becomeHost = async function (remoteUser) {
```
znarf, line 904:
```suggestion
  becomeHost = async function ({remoteUser} = {}) {
```

Not super important but I prefer such syntax to pass was seems to be extra parameter, and not a main argument of `becomeHost`.

znarf (approved):

server/models/Collective.ts
Betree: It depends on whether we consider `remoteUser` as optional or not. In the code below, its presence is required so I wouldn't use the first syntax.

We can decide to have this function callable without `remoteUser`, in which case we should also update the code below to `removeUser?.id`. I personally like that we're enforcing a user here as it will force new developments to include it. We can always update the signature in the future if the use case for non-user-triggered events becomes a thing.

znarf: Good for me

PR #9411, opened 2023-10-20
Add accounting category on expense mutations

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7052

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -9,6 +9,7 @@ import {
   flatten,
   get,
   isBoolean,
+  isEmpty,
   isEqual,
   isNil,
   isNumber,
@@ -52,6 +53,7 @@ import twoFactorAuthLib from '../../lib/two-factor-authentication';
 import { canUseFeature } from '../../lib/user-permissions';
 import { formatCurrency, parseToBoolean } from '../../lib/utils';
 import models, { Collective, sequelize } from '../../models';
+import AccountingCategory from '../../models/AccountingCategory';
 import Expense from '../../models/Expense';
 import { ExpenseAttachedFile } from '../../models/ExpenseAttachedFile';
 import { ExpenseItem } from '../../models/ExpenseItem';
@@ -618,6 +620,28 @@ export const canMarkAsIncomplete: ExpensePermissionEvaluator = async (
 };
 
 /**
+ * Returns true if user is allowed to change the accounting category of the expense
+ */
+export const canEditExpenseAccountingCategory = async (
+  req: express.Request,
+  expense: Expense,
+  options = { throw: false },
+): Promise<boolean> => {
+  return remoteUserMeetsOneCondition(
+    req,
+    expense,
+    [
+      // Host admins and accountants can always change the accounting category.
+      isHostAdmin,
+      isHostAccountant,
+      // Otherwise, we fallback on the default edit permissions (must be a collective/fromCollective admin)
+      canEditExpense,
+    ],
+    options,
+  );
+};
+
+/**
  * Returns true if expense can be marked as unpaid by user
  */
 export const canMarkAsUnpaid: ExpensePermissionEvaluator = async (
@@ -1134,6 +1158,7 @@ type ExpenseData = {
   currency?: string;
   tax?: TaxDefinition[];
   customData: Record<string, unknown>;
+  accountingCategory?: AccountingCategory;
 };
 
 const EXPENSE_EDITABLE_FIELDS = [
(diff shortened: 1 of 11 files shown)


server/graphql/common/expenses.ts
```diff
@@ -1172,6 +1197,17 @@ const checkTaxes = (account, host, expenseType: string, taxes): void => {
   }
 };
 
+/**
+ * Throws if the accounting category is not allowed for this expense/host
+ */
+const checkCanUseAccountingCategory = (accountingCategory: AccountingCategory, host: Collective): void => {
```
kewitz, line 1203: If we have a safe return for an empty `accountCategory` parameter:
```suggestion
const checkCanUseAccountingCategory = (accountingCategory?: AccountingCategory, host: Collective): void => {
```

server/graphql/common/expenses.ts
```diff
@@ -1475,6 +1513,22 @@ export const DRAFT_EXPENSE_FIELDS = [
...
+/**
+ * Returns true if the value has changed.
+ *
+ * /!\ We have no 1-to-1 field mapping between `Expense` and `ExpenseData`, but since we used to only
+ * check `isNil` this won't introduce any regression. Ideally, this helper should do a mapping between the two.
+ */
+const isValueChanging = (expense: Expense, expenseData: ExpenseData, key: string): boolean => {
+  const nullableFields = ['accountingCategory'];
+  const value = expenseData[key];
+  if (!nullableFields.includes(key)) {
```
kewitz, line 1525: 🍨 This conditional would be simpler to read if the primary conditional wasn't a negation:
```suggestion
  if (nullableFields.includes(key)) {
```

server/graphql/common/expenses.ts
```diff
@@ -1502,6 +1556,44 @@ export async function editExpenseDraft(req: express.Request, expenseData: Expens
   });
 }
 
+/**
+ * A simple helper to handle the case when editing only the tags and/or accounting category of an expense.
+ * Permissions for these fields (especially regarding the status of the expense) are more relaxed than for other fields.
+ */
+const editOnlyTagsAndAccountingCategory = async (
```
kewitz, line 1563: Nice! I like this, especially due to the complexity of editing an expense. Perhaps we should have a separate mutation to edit this type of meta-information about expenses and orders.

server/models/Expense.ts
```diff
@@ -260,9 +262,11 @@ class Expense extends Model<InferAttributes<Expense>, InferCreationAttributes<Ex
       CollectiveId: this.CollectiveId,
       FromCollectiveId: this.FromCollectiveId,
       HostCollectiveId: this.HostCollectiveId,
+      AccountingCategoryId: this.AccountingCategoryId,
       currency: this.currency,
       amount: this.amount,
       description: this.description,
+      /** @deprecated - now using `tags` */
```
kewitz, line 269: Nice catch.

server/graphql/common/expenses.ts
```diff
@@ -617,6 +619,28 @@ export const canMarkAsIncomplete: ExpensePermissionEvaluator = async (
...
+): Promise<boolean> => {
+  return remoteUserMeetsOneCondition(
+    req,
+    expense,
+    [
+      // Host admins and accountants can always change the accounting category.
+      isHostAdmin,
+      isHostAccountant,
+      // Otherwise, we fallback on the default edit permissions (must be a collective/fromCollective admin)
+      canEditExpense,
```
kewitz, line 638: This will enable `CollectiveAdmin to edit paid expenses, not sure if we want this.

kewitz (approved):

server/graphql/common/expenses.ts
Betree: True

server/graphql/common/expenses.ts
Betree: If the expense is STATUS=PAID, `canEditExpense` will let collective admins pass only if the expense is a charge - aka when they upload the documentation for the virtual card - which seems like the perfect moment to set the accounting category.

server/graphql/common/expenses.ts
Betree: > error TS1016: A required parameter cannot follow an optional parameter.

Betree: Problem is: we don't want devs to call this function without providing the parameter. So the right call is to type it as `AccountingCategory | undefined | null`. Since we're not enforcing strict null checks for now, I don't think that we should bother with that, but I've done it here since I was on it

PR #9414, opened 2023-10-23
Add GraphQLV2 ID to Webhooks

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6449

~~I went with `publicId` as it sounded friendlier than `idV2`. Other alternatives:~~
- `publicGraphQLId`
- `graphqlV2Id`
- `graphqlId`
- `idGraphQLV2`

### DIFF
--- a/server/lib/webhooks.js
+++ b/server/lib/webhooks.js
@@ -1,6 +1,7 @@
 import { pick } from 'lodash';
 
 import { activities } from '../constants';
+import { idEncode, IDENTIFIER_TYPES } from '../graphql/v2/identifiers';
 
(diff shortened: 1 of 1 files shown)


znarf:
Not sure about `publicId` but after having looked at alternatives, nothing against.

The only -1 is that we used `idV2` to express that in GraphQL V1 transactions.

Betree:
> The only -1 is that we used `idV2` to express that in GraphQL V1 transactions.

That's a good point in favor of `idV2`. I'll go back to it, in the spirit of consistency.

PR #9425, opened 2023-10-27
Feat: Add Vendors Dashboard and first implementation Specs

kewitz:
Related to https://github.com/opencollective/opencollective/issues/7030


test/server/graphql/v2/interface/Account.test.ts
```diff
@@ -0,0 +1,74 @@
...
+        id
+        type
+        name
+        slug
+      }
+    }
+  }
+`;
+
+describe('server/graphql/v2/interface/Account', () => {
```
Betree, line 22: To me, this really belongs to the `Host` object

PR #9426, opened 2023-10-27
Expense tag stats

gustavlrsn:
Project: https://github.com/opencollective/opencollective/issues/7033

### Description
Adding a `expenseTagStats` root query to support an autocompleting fuzzy search filter on expense tags in the expenses dashboard, for both hosts and regular accounts.

It's returning results from a materialized view `ExpenseTagStats`, that has tag counts grouped on either HostCollectiveId and CollectiveId.

While the `account.expensesTags` field could potentially be extended to include a `tagSearchTerm`, it might not be ideal to do a fairly expensive query in an autocompleting search. It also lacks support for the host use case.


gustavlrsn:
* main
  * **PR #9426** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-api/9426" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-api/9426?utm_source=stack-comment).

migrations/20231028141843-expense-tag-stats-materialized-view.js
```diff
@@ -0,0 +1,46 @@
...
+          FROM "Expenses"
+          WHERE "CollectiveId" IS NOT NULL
+          AND "deletedAt" IS NULL
+          AND "status" NOT IN ('SPAM', 'DRAFT', 'UNVERIFIED')
+          GROUP BY "tag", "CollectiveId"
+      ) AS CollectiveTags;`);
+
+    await queryInterface.sequelize.query(
+      `CREATE INDEX IF NOT EXISTS expense_tag_stats__HostCollectiveId_CollectiveId ON "ExpenseTagStats"("HostCollectiveId","CollectiveId");`,
+    );
```
kewitz, line 32: Should we also create one index for the "HostCollectiveId" only query?

cron/daily/92-refresh-expense-tag-stats-materialized-view.ts
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env node
```
kewitz, line 1: I think you should include this in `cron/hourly/50-refresh-materialized-views.js`.

kewitz (approved):
Looking good.

migrations/20231028141843-expense-tag-stats-materialized-view.js
gustavlrsn: The query using the table was always using both "HostCollectiveId" and "CollectiveId" (setting either to null) in the WHERE clause, but I updated the query now to only use either argument since both are not really needed, so I've added a separate index instead of this composite index for both "HostCollectiveId" and "CollectiveId".

Also added a unique composite index on "HostCollectiveId", "CollectiveId", and "tag" to make the view able to refresh CONCURRENTLY.

cron/daily/92-refresh-expense-tag-stats-materialized-view.ts
gustavlrsn: Ah, nice, had not seen that one. Included it there, also moved the daily cron of refreshing the CollectiveTagStats view to this file.

migrations/20231028141843-expense-tag-stats-materialized-view.js
kewitz: Nice work!

gustavlrsn:
### Merge activity

* **Nov 8, 12:45 PM**: @gustavlrsn merged this pull request with [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-api/9426).

PR #9430, opened 2023-10-31
Show Canceled card on card query

hdiniz:


Betree (approved):

znarf:
Why are we listing deleted cards in the first place? Where is this happening?

znarf:
Generally, everything that is soft-deleted should not be returned on the API.

I would rather stop returning them, likely in
https://github.com/opencollective/opencollective-api/blob/main/server/graphql/v2/interface/Account.ts#L516

PR #9434, opened 2023-11-02
Breakdown AccountingCategories values by role in data

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7090

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -49,13 +49,14 @@ import { listPayPalTransactions } from '../../lib/paypal';
 import { getPolicy } from '../../lib/policies';
 import { reportErrorToSentry, reportMessageToSentry } from '../../lib/sentry';
 import { notifyTeamAboutSpamExpense } from '../../lib/spam';
+import { deepJSONBSet } from '../../lib/sql';
 import { createTransactionsForManuallyPaidExpense, createTransactionsFromPaidExpense } from '../../lib/transactions';
 import twoFactorAuthLib from '../../lib/two-factor-authentication';
 import { canUseFeature } from '../../lib/user-permissions';
 import { formatCurrency, parseToBoolean } from '../../lib/utils';
 import models, { Collective, sequelize } from '../../models';
 import AccountingCategory from '../../models/AccountingCategory';
-import Expense from '../../models/Expense';
+import Expense, { ExpenseDataValuesByRole } from '../../models/Expense';
 import { ExpenseAttachedFile } from '../../models/ExpenseAttachedFile';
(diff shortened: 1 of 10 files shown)


server/graphql/common/expenses.ts
```diff
@@ -1599,16 +1616,33 @@ const editOnlyTagsAndAccountingCategory = async (
...
 
-  if (isEmpty(newValues)) {
+  if (isEmpty(updateClauses)) {
     return expense;
   }
 
-  const updatedExpense = await expense.update(newValues);
-  updatedExpense.createActivity(activities.COLLECTIVE_EXPENSE_UPDATED, req.remoteUser);
-  return updatedExpense;
+  // Use a raw query to unlock the ability of using `JSONB_SET`, which will prevent concurrency issues on the `data` field
```
Betree, line 1630: This was important since updating tags and accounting categories can be done regardless of the status, which means it could be edited at the same time it is being processed.

I also tried with a `sequelize.literal`, but `replacements` is not supported for updates, so it was unclear how to pass sanitized values.

kewitz: Cool, I understand why and agree this is a good idea here.

server/graphql/v2/object/ExpenseValuesByRole.ts
```diff
@@ -0,0 +1,46 @@
...
+      },
+    },
+  }),
+});
+
+export const GraphQLExpenseValuesByRole = new GraphQLObjectType({
+  name: 'ExpenseValuesByRole',
+  fields: () => ({
+    id: { type: new GraphQLNonNull(GraphQLNonEmptyString), resolve: getIdEncodeResolver(IDENTIFIER_TYPES.EXPENSE) },
+    submitter: {
```
kewitz, line 30: Ideally, these fields are also defined by constants.

server/lib/sql.ts
```diff
@@ -0,0 +1,37 @@
...
+ *     '{bar}',
+ *     JSONB_SET(
+ *       COALESCE(data->'foo'->'bar', '{}'),
+ *       '{baz}',
+ *       :myValue
+ *     )
+ *  )
+ * )
+ */
+export const deepJSONBSet = (field: string, path: string[], value: string) => {
```
kewitz, line 25: Nice! :100:

kewitz (approved):
Nice work, outside turning contextual roles into constants I have nothing to add.

server/graphql/v2/object/ExpenseValuesByRole.ts
Betree: They're actually not named the same way: it's `collectiveAdmin` in the DB and `accountAdmin` in graphql

PR #9437, opened 2023-11-03
Fix: Enable Expense author to edit invited payee email

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7080

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -1563,12 +1563,32 @@ const isValueChanging = (expense: Expense, expenseData: ExpenseData, key: string
 
 const isDifferentInvitedPayee = (expense: Expense, payee): boolean => {
   const isInvitedPayee = !expense.data?.payee?.id && expense.data.payee.email;
-  console.log(isInvitedPayee, expense.data.payee, payee);
   if (isInvitedPayee) {
     return !matches(expense.data.payee)(payee);
   }
   return false;
 };
+
+export async function sendDraftExpenseInvite(
+  req: express.Request,
+  expense: Expense,
+  collective: Collective,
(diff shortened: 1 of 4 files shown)


server/graphql/common/expenses.ts
```diff
@@ -1559,18 +1571,46 @@ export async function editExpenseDraft(req: express.Request, expenseData: Expens
...
+        inviteUrl,
+      })
+      .catch(e => {
+        logger.error('An error happened when creating the COLLECTIVE_EXPENSE_INVITE_DRAFTED activity', e);
+        reportErrorToSentry(e);
+      });
+    if (config.env === 'development') {
+      logger.info(`Expense Invite Link: ${inviteUrl}`);
+    }
+  }
```
Betree, line 1611: Maybe we could have a common helper with `server/graphql/v2/mutation/ExpenseMutations.ts` to centralize this code

Betree (approved):
Lgtm!

server/graphql/common/expenses.ts
```diff
@@ -1558,7 +1561,36 @@ const isValueChanging = (expense: Expense, expenseData: ExpenseData, key: string
...
+};
+
+export async function sendDraftExpenseInvite(
+  req: express.Request,
+  expense: Expense,
+  collective: Collective,
+  draftKey: string,
+): Promise<void> {
+  const inviteUrl = `${config.host.website}/${collective.slug}/expenses/${expense.id}?key=${draftKey}`;
+  expense
```
Betree, line 1579: If we want the code that uses this function to be able to `await` on it

```suggestion
return expense
```

kewitz:
Heads up @BenJam.

PR #9439, opened 2023-11-06
Make "status" optional in Host Applications query

gustavlrsn:
### Description
Fixes the host.hostApplications resolver to make sure the 'status' argument is actually optional as specified by the GraphQL schema and what is required for the new filters.

### DIFF
--- a/server/graphql/v2/object/Host.js
+++ b/server/graphql/v2/object/Host.js
@@ -367,7 +367,7 @@ export const GraphQLHost = new GraphQLObjectType({
             order: [[args.orderBy.field, args.orderBy.direction]],
             where: {
               HostCollectiveId: host.id,
-              status: args.status,
+              ...(args.status && { status: args.status }),
             },
             limit: args.limit,
             offset: args.offset,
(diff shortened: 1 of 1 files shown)


server/graphql/v2/object/Host.js
```diff
@@ -378,7 +378,6 @@ export const GraphQLHost = new GraphQLObjectType({
                 required: true,
                 where: {
                   HostCollectiveId: host.id,
-                  ...(args.status === 'PENDING' && { approvedAt: null }),
```
gustavlrsn, line 381: This is no longer needed since implementing the model check and fix in https://github.com/opencollective/opencollective-api/blob/main/checks/model/hosted-collectives.js#L45

znarf (approved):
Looks good to me.

Does it need a matching frontend PR? Any problem to expect there?

gustavlrsn:
### Merge activity

* **Nov 7, 8:23 AM**: @gustavlrsn merged this pull request with [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-api/9439).

PR #9456, opened 2023-11-07
Add expense total in expense collection

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7032


Betree (changes requested):
It's a good start but I think we can improve accuracy and performance

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -372,6 +372,18 @@ const ExpensesCollectionQuery = {
...
+          attributes: [[Sequelize.fn('SUM', Sequelize.col('amount')), 'amount']],
+          include,
+          where,
+          raw: true,
+        });
+
+        return {
+          value: totalExpenses.amount,
+        };
+      },
```
Betree, line 386: 1. We need to handle expenses that are not in the currency of the collective
2. We must make sure we actually return a `currency` for the amount object
3. Rather than trigger a second query, could we reuse the rows that have already been fetched

```suggestion
      async totalAmount: () => { 
        const expensesByCurrency = groupBy(result.rows, 'currency');
        const amountByCurrency = mapValues(expensesByCurrency, expenses => sumBy(expenses, 'amount'));
        // TODO: Convert all amounts to collective's currency, ideally in a single async call using req.loaders.CurrencyExchangeRate.convert.loadMany
        return { value, currency: collective.currency };
      }
```

hdiniz: We can't reuse the first query as it does not contain the entire set of expenses, it contains at most a page, and sometimes no records at all when the client sets `limit: 0` to get totals.

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -375,6 +375,32 @@ const ExpensesCollectionQuery = {
...
+        const values = await req.loaders.CurrencyExchangeRate.convert.loadMany(
+          currencies.map(v => ({ amount: v.value, fromCurrency: v.currency, toCurrency: currency })),
+        );
+
+        return {
+          currencies,
+          converted: {
+            value: sum(values),
+            currency,
+          },
```
hdiniz, line 401: @Betree I think the schema is still a bit awkward, if you have suggestions let me know. 
I tried to add flexibility here by aggregating the total by currency and returning each value for the frontend to display. Optionally, a desired currency can be passed as an argument to convert everything.

In some uses of the expense query, the desired currency is not clear:
- If we querying expenses submitted to different collectives/hosts.
- If we have a mix of PAID/UNPAID expenses, converting to a single currency might not make sense with different exchange rates across time.

server/graphql/v2/collection/ExpenseCollection.ts
```diff
@@ -12,5 +14,24 @@ export const GraphQLExpenseCollection = new GraphQLObjectType({
...
+              },
+            },
+            type: GraphQLAmount,
+          },
+          currencyAmounts: {
+            type: new GraphQLList(GraphQLAmount),
+          },
+        },
+      }),
+    },
```
znarf, line 35: I would propose to send back a `totalAmount: GraphQLAmount`

And extend the `GraphQLAmount` type to support a list of amounts it's aggregating

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -375,6 +375,47 @@ const ExpensesCollectionQuery = {
     return {
       nodes: result.rows,
       totalCount: result.count,
+      aggregation: async () => {
+        const getCurrencyTotalAmountsMemo = (() => {
```
Betree, line 379: Is this memoization working? My intuition is that both resolvers will be called at the same time, calling the function twice in parallel - unless we're lucky with the timing.

The proper way to achieve that would normally be to implement a loader.

test/server/graphql/v2/collection/ExpenseCollectionQuery.test.ts
```diff
@@ -63,6 +73,36 @@ const fakeHostWithRequiredLegalDocument = async (hostData = {}) => {
...
+    await fakeExpense({
+      amount: 5000,
+      currency: 'GBP',
+      type: 'RECEIPT',
+      CollectiveId: collective.id,
+      status: 'APPROVED',
+    });
+
+    const result = await graphqlQueryV2(expensesQuery, queryParams);
+    console.log(result.errors);
```
Betree, line 97:
```suggestion
```

test/server/graphql/v2/collection/ExpenseCollectionQuery.test.ts
```diff
@@ -38,6 +38,16 @@ const expensesQuery = gqlV2/* GraphQL */ `
...
+      aggregation {
+        currencyAmounts {
+          valueInCents
+          currency
+        }
+        totalAmount(currency: USD) {
+          valueInCents
+          currency
+        }
+      }
```
znarf, line 50:
```suggestion
      totalAmount {
        amount(currency: USD) {
          valueInCents
          currency
        }
        amountsByCurrency {
          valueInCents
          currency
        }
      }
```

znarf, line 50: Or:

```suggestion
      totalAmount(currency: USD)  {
        valueInCents
        currency
        amountsByCurrency {
          valueInCents
          currency
        }
      }
```

server/graphql/v2/collection/ExpenseCollection.ts
znarf: Or:

```
const GraphQLTotalAmount = new GraphQLObjectType({
        name: 'TotalAmount',
        fields: {
          amount: {
            type: GraphQLAmount,
         },
        amountsByCurrency: {
            type: new GraphQLList(GraphQLAmount),
          },
       }
```

```
totalAmount: new GraphQLTotalAmount()
```

znarf:
I see that the breakdown per currency is actually not used on the frontend side, any plan for this?

test/server/graphql/v2/collection/ExpenseCollectionQuery.test.ts
Betree: I like this API; it's clear and simple.

If we stick with `aggregation`, I prefer `amountsByCurrency` as a name (rather than `currencyAmounts`). "Aggregation" has a benefit: we'll have a place to add more aggregations in the future, things like `expenses.aggregate.averageAmount`, `expenses.aggregate.medianAmount`, `expenses.aggregate.minAmount`, ...

Betree (commented):
There's some unanswered feedback above, i.e. https://github.com/opencollective/opencollective-api/pull/9456#discussion_r1393835567

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
hdiniz: Yes, thanks! There was an issue there. In this last push, I refactored to use a promise. It will only execute once and memoize the result.

I think a loader would be a bit more complex due to passing the collection query filters to it. And since this is not used in a list of expenses lists, maybe not needed.

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -378,6 +378,47 @@ export const ExpensesCollectionQueryResolver = async (
...
+      })();
+
+      return {
+        amountsByCurrency: getCurrencyTotalAmountsMemo,
+        amount: async ({ currency = 'USD' }) => {
+          const currencyAmounts = await getCurrencyTotalAmountsMemo();
+
+          const values = await req.loaders.CurrencyExchangeRate.convert.loadMany(
+            currencyAmounts.map(v => ({ amount: v.value, fromCurrency: v.currency, toCurrency: currency })),
+          );
```
Betree, line 413: This seems like an unnecessary complication. We could have used `memoize` from lodash to simplify it, but since the promise needs to be fetched in all subfields why not simply can the promise directly?

```suggestion
      const query = (await models.Expense.findAll({
        attributes: [
          [Sequelize.col('"Expense"."currency"'), 'expenseCurrency'],
          [Sequelize.fn('SUM', Sequelize.col('amount')), 'amount'],
        ],
        group: 'expenseCurrency',
        include,
        where,
        raw: true,
      })) as unknown as { expenseCurrency: string; amount: number }[];

      const amountsByCurrency = query.map(result => ({ currency: result.expenseCurrency, value: result.amount }));

      return {
        amountsByCurrency,
        amount: async ({ currency = 'USD' }) => {
          const values = await req.loaders.CurrencyExchangeRate.convert.loadMany(
            amountsByCurrency.map(v => ({ amount: v.value, fromCurrency: v.currency, toCurrency: currency })),
          );
```

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
Betree: Followed up in https://github.com/opencollective/opencollective-api/pull/9456/files#r1399291105

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
hdiniz: Yep, you absolutely correct!

Betree (approved):
Works perfectly :ok_hand:

PR #9459, opened 2023-11-07
Await activity creation to avoid deadlocks

gustavlrsn:


gustavlrsn:
* main
  * **PR #9459** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-api/9459" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-api/9459?utm_source=stack-comment).

Betree (commented):
I like the idea of waiting for activities creation; I'm not a big fan of throwing Promises without looking at their result. We have many other places like that which would need to be updated if we want to make it a global change.

My main concern with the way it has been done here is that it actually changes the behavior in case there's a crash in the activity creation: we'll now return an error saying that the collective/event/fund/project creation failed, which is not true.

I'd recommend, inside `generateCollectiveCreatedActivity`, to wrap the activity creation call in a `try ... catch`, reporting the error to Sentry without re-throwing it if there's an issue.

PR #9466, opened 2023-11-08
Dont auto pause card that were resumed

hdiniz:
Related https://github.com/opencollective/opencollective/issues/4312#issuecomment-1791022764


znarf:
Not a fan of relying on the activity table for such things. I'm ok with it still.

Maybe we should have introduced a dedicated timestamp in the VirtualCards table.

znarf (changes requested):
Would just name the column with a shorter `resumedAt`. We don't say `lastUpdatedAt` for example.

znarf (approved):
Looks good!

PR #9473, opened 2023-11-10
don't issue payment processor fee cover to itself

znarf:
Fix https://github.com/opencollective/opencollective/issues/7083

### DIFF
--- a/migrations/20231110114933-fix-payment-processor-fee-cover.js
+++ b/migrations/20231110114933-fix-payment-processor-fee-cover.js
@@ -5,7 +5,7 @@ module.exports = {
   async up(queryInterface) {
     await queryInterface.sequelize.query(`
      UPDATE "Transactions"
-     SET "deletedAt" = NOW()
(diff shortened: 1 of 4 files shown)


Betree (approved):
Good stuff, but we may want to run a migration before we close the issue

migrations/20231110114933-fix-payment-processor-fee-cover.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+     UPDATE "Transactions"
+     SET "deletedAt" = NOW()
```
Betree, line 8: More of a personal preference: I like to mark migrated transactions somehow; I think it helps to investigate later when we stumble on the data and wonder why it's in a certain state.

```suggestion
     SET "deletedAt" = NOW(), "data" = jsonb_set(COALESCE("data", '{}'), '{deletedFromMigration20231110114933}', 'true')
```

Betree (approved):
:+1:

PR #9480, opened 2023-11-13
PersonalToken: enforce 2FA for fetching value

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6630


znarf (approved):
All right!

PR #9485, opened 2023-11-14
Remove plugin-transform-async-to-generator

znarf:
Fix: https://github.com/opencollective/opencollective/issues/1432

Related to: https://github.com/opencollective/opencollective/issues/6764

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -18,7 +18,6 @@
     ],
     "add-module-exports",
     "@babel/plugin-proposal-class-properties",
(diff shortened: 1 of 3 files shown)


Betree (approved):
The only place where we're using an async generator is https://github.com/opencollective/opencollective-api/blob/0f8c6ab644fc4ace423fa63f1887acfa4cc0767c/scripts/paypal/payment-reconciliator.ts#L45. I just tested, and it seems to be working properly without the plugin. All good for me!

PR #9486, opened 2023-11-14
Vendors: List potential vendors and convert Organization to Vendor

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/3160

### DIFF
--- a/server/graphql/v2/interface/Account.ts
+++ b/server/graphql/v2/interface/Account.ts
@@ -17,6 +17,7 @@ import { checkRemoteUserCanUseAccount, checkScope } from '../../common/scope-che
 import { BadRequest, ContentNotReady, Unauthorized } from '../../errors';
 import { GraphQLAccountCollection } from '../collection/AccountCollection';
 import { GraphQLConversationCollection } from '../collection/ConversationCollection';
+import { GraphQLExpenseCollection } from '../collection/ExpenseCollection';
 import { GraphQLMemberCollection, GraphQLMemberOfCollection } from '../collection/MemberCollection';
 import { GraphQLOAuthApplicationCollection } from '../collection/OAuthApplicationCollection';
 import { GraphQLOrderCollection } from '../collection/OrderCollection';
@@ -56,6 +57,10 @@ import { GraphQLPolicies } from '../object/Policies';
 import { GraphQLSocialLink } from '../object/SocialLink';
 import { GraphQLTagStats } from '../object/TagStats';
 import { GraphQLTransferWise } from '../object/TransferWise';
+import {
+  ExpensesCollectionQueryArgs,
+  ExpensesCollectionQueryResolver,
+} from '../query/collection/ExpensesCollectionQuery';
 import { OrdersCollectionArgs, OrdersCollectionResolver } from '../query/collection/OrdersCollectionQuery';
 import {
   TransactionsCollectionArgs,
@@ -287,6 +292,10 @@ const accountFieldsDefinition = () => ({
     args: {
       ...OrdersCollectionArgs,
     },
+  },
+  expenses: {
+    type: new GraphQLNonNull(GraphQLExpenseCollection),
+    args: ExpensesCollectionQueryArgs,
   },
   settings: {
     type: new GraphQLNonNull(GraphQLJSON),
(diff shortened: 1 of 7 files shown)


server/graphql/v2/interface/Account.ts
```diff
@@ -862,6 +871,14 @@ export const AccountFields = {
...
   transactions: accountTransactions,
   orders: accountOrders,
+  expenses: {
+    type: new GraphQLNonNull(GraphQLExpenseCollection),
+    args: ExpensesCollectionQueryArgs,
+    resolve(collective, args, req) {
+      args.fromAccount = { legacyId: collective.id };
+      return ExpensesCollectionQueryResolver(undefined, args, req);
+    },
+  },
```
Betree, line 881: It's unclear to me whether a field called `account.expenses` should return received expenses, submitted expenses, or both. IMO, we should either rename the field to make it explicit or add a param (e.g. `direction`) to clarify.

server/graphql/v2/object/Host.js
```diff
@@ -1076,6 +1076,59 @@ export const GraphQLHost = new GraphQLObjectType({
           return { nodes: vendors, totalCount: count, limit: args.limit, offset: args.offset };
         },
       },
+      potentialVendors: {
+        type: new GraphQLNonNull(GraphQLAccountCollection),
```
Betree, line 1080: Maybe add a small description to document what this field stands for?

server/models/Collective.ts
```diff
@@ -3064,7 +3064,8 @@ class Collective extends Model<
   };
 
   getImageUrl = function (args = {}) {
-    return getCollectiveAvatarUrl(this.slug, this.type, this.image, args);
+    const slug = this.data?.originalOrganizationProps?.slug || this.slug;
```
Betree, line 3067: Not a fan of this:
1. What if the original slug is then used by another profile? It sounds like this will return another profile's picture
2. How do you then change the image or the vendor profile?

Betree (commented):
Early review, I will test that tomorrow!

server/graphql/v2/object/Host.js
```diff
@@ -1076,6 +1076,59 @@ export const GraphQLHost = new GraphQLObjectType({
           return { nodes: vendors, totalCount: count, limit: args.limit, offset: args.offset };
         },
       },
+      potentialVendors: {
```
Betree, line 1079: The way it's implemented, this field will miss vendors created through the "Add funds" flow. When using this flow, we create a profile where the host admin is **not** added as an admin of the profile. Instead, we create a user with the provided info (email, name) and add this person as the sole admin of the profile.

We could solve this by relying on the `CreatedByUserId` column of the profile and checking whether the only admin of the profile has confirmed their email.

server/graphql/v2/mutation/VendorMutations.ts
```diff
@@ -215,6 +215,66 @@ const vendorMutations = {
...
+      const vendorData = {
+        type: CollectiveType.VENDOR,
+        slug: `${host.id}-${organization.slug}-${uuid().substr(0, 8)}`,
+        CreatedByUserId: req.remoteUser.id,
+        isActive: false,
+        ParentCollectiveId: host.id,
+        data: organization.data || {},
+      };
+      vendorData.data['originalOrganizationProps'] = pick(organization.toJSON(), Object.keys(vendorData));
+      return organization.update(vendorData);
```
Betree, line 275: I would also delete all the admins of the organization, something like:

```suggestion
      await organization.update(vendorData);
      await models.Member.destroy({ where: { CollectiveId: organization.id } });
      return organization;
```

server/graphql/v2/object/Host.js
```diff
@@ -1076,6 +1076,59 @@ export const GraphQLHost = new GraphQLObjectType({
...
+                SELECT c.*
+                FROM "orgs"
+                INNER JOIN "Collectives" c ON c.id = "orgs".id
+                WHERE
+                  "admins" <@ ARRAY(SELECT "MemberCollectiveId" FROM hostadmins)
+                  AND "hosts" = ARRAY[:hostid]
+                ORDER BY c."createdAt" DESC
+                LIMIT :limit
+                OFFSET :offset;
+          `;
```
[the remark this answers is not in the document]
kewitz: The problem is that we're looking for oraganizations that *only* have transacted with a specific host, this means we need to list all those transactions in order to filter out the ones that have transacted with other hosts.

PR #9488, opened 2023-11-14
Only find 2FAs not already deleted

SudharakaP:
I believe here we need to make sure we only take into account 2FAs that are not already deleted. Otherwise we will show things that are already deleted as not deleted in root actions for example. 

image

### DIFF
--- a/server/models/UserTwoFactorMethod.ts
+++ b/server/models/UserTwoFactorMethod.ts
@@ -71,6 +71,7 @@ export default class UserTwoFactorMethod<
       group: 'method',
       where: {
         UserId: userId,
+        deletedAt: null,
       },
     });
 


server/models/UserTwoFactorMethod.ts
```diff
@@ -71,6 +71,7 @@ export default class UserTwoFactorMethod<
       group: 'method',
       where: {
         UserId: userId,
+        deletedAt: null,
```
hdiniz, line 74: I assumed that the the `deletedAt: null` would be redundant because the model is set to be paranoid, is that not the case?

https://github.com/opencollective/opencollective-api/blob/543812902d702526abf9d473b2c131f0425dc4af/server/models/UserTwoFactorMethod.ts#L178

Betree: Indeed, the `deletedAt` condition is automatically added by default on paranoid models.

Betree:
@SudharakaP Not clear what we're trying to solve. The fact that it appears in the collective picker may be more of an issue with the picker itself? Or maybe something we allow only in the context of root actions?

SudharakaP:
I think the issue that was seen is fixed by https://github.com/opencollective/opencollective-api/pull/9605. Anyways, I'll keep an eye out and report this again if I see this issue.

PR #9491, opened 2023-11-15
WIP: Separate payment processor fees

znarf:


server/models/Transaction.ts
```diff
@@ -1006,6 +1039,87 @@ Transaction.createHostFeeTransactions = async (transaction, host, data) => {
...
+/**
+ * Returns the Vendor account associated with a payment processor service. This function is memoized as
+ * the processor fee vendor will not change during the lifetime of the server.
+ */
+Transaction.getPaymentProcessorFeeVendor = memoize(
+  async (service: PAYMENT_METHOD_SERVICE | PayoutMethodTypes | 'OTHER'): Promise<Collective> => {
+    const vendorSlugs = {
+      [PAYMENT_METHOD_SERVICE.STRIPE]: 'stripe-payment-processor-vendor',
+      [PAYMENT_METHOD_SERVICE.PAYPAL]: 'paypal-payment-processor-vendor',
+      [PayoutMethodTypes.BANK_ACCOUNT]: 'wise-payment-processor-vendor', // TODO: We should differentiate depending on manual vs. automatic
```
Betree, line 1051: @kewitz Is there a way we could detect whether it's Wise or manual from the transaction `data`, without async code?

znarf: TODO!

Betree:
A good test to add as a followup to https://github.com/opencollective/opencollective-api/pull/9536: trying to create a transaction with `amount=0`, but a payment processor fee set should only create the fee transaction, not the root one.

PR #9495, opened 2023-11-16
Remove thegivingblock/crypto support

znarf:
Fix: https://github.com/opencollective/opencollective/issues/6859


Betree (approved):
Looks good!

PR #9498, opened 2023-11-17
Improve developer experience for checks

Betree:
- Always call `process.exit` to make sure the scripts don't hang at the end
- Parse options (`--fix`) in individual scripts, which makes sure they can be run individually


znarf:
Looks good!

PR #9499, opened 2023-11-17
Update potentialVendors query to include organizations created by admins

kewitz:


server/graphql/v2/object/Host.js
```diff
@@ -1091,23 +1091,36 @@ export const GraphQLHost = new GraphQLObjectType({
 
           const pageQuery = `
                 WITH hostadmins AS (
-                  SELECT "MemberCollectiveId" FROM "Members" WHERE "CollectiveId" = :hostid AND "deletedAt" IS NULL AND role = 'ADMIN'
-                ), orgs AS (
-                  SELECT c.id, c.slug,ARRAY_AGG(DISTINCT m."MemberCollectiveId") as "admins", ARRAY_AGG(DISTINCT t."HostCollectiveId") as hosts
+                  SELECT m."MemberCollectiveId", m."CreatedByUserId" as "UserId"
```
Betree, line 1094: `m."CreatedByUserId"` cannot safely be interpreted as "User ID of the member"; it contains the ID of the user who created/invited the member, not the User ID itself. 

I suppose your intention was to write:

```suggestion
                  SELECT m."MemberCollectiveId", u."id" as "UserId"
```

server/graphql/v2/object/Host.js
```diff
@@ -1091,23 +1091,36 @@ export const GraphQLHost = new GraphQLObjectType({
...
-                  "admins" <@ ARRAY(SELECT "MemberCollectiveId" FROM hostadmins)
-                  AND "hosts" = ARRAY[:hostid]
+                  (
+                    o."admins" <@ ARRAY(SELECT "MemberCollectiveId" FROM hostadmins)
+                      OR (
+                        o."CreatedByUserId" IN (
+                        SELECT "UserId"
+                        FROM hostadmins
+                        )
+                        AND o."admins" = ARRAY[null]::INTEGER[]
```
Betree, line 1120: I'm not convinced by this `ARRAY[null]::INTEGER[]`, organizations created via added funds do have an admin, it's just that it's an invited account (that often doesn't get confirmed)

server/graphql/v2/object/Host.js
kewitz: Thanks! Nice catch, this was here from the first attempt.

server/graphql/v2/object/Host.js
kewitz: Yes. I was working with the idea that I didn't want to block Organizations that were never activated from becoming Vendors. (Assuming the invited admin never had the intention to join).

PR #9515, opened 2023-11-21
Enforce single payout method per Vendor

kewitz:


server/graphql/v2/mutation/VendorMutations.ts
```diff
@@ -160,20 +160,27 @@ const vendorMutations = {
...
+            ...pick(args.vendor.payoutMethod, ['name', 'data', 'type']),
+            CollectiveId: vendor.id,
+            CreatedByUserId: req.remoteUser.id,
+            isSaved: true,
+          });
+        } else {
+          await Promise.all(
+            existingPayoutMethods
+              .filter(pm => pm.id !== args.vendor.payoutMethod.id)
+              .map(pm => pm.update({ isSaved: false })),
```
Betree, line 180: Missing an ID decode

```suggestion
            const payoutMethodId = idDecode(args.vendor.payoutMethod.id, 'payout-method');
            existingPayoutMethods
              .filter(pm => pm.id !== payoutMethodId)
              .map(pm => pm.update({ isSaved: false })),
```

Also, shouldn't we update the other info for payout method here? Something like:

```es6
const payoutMethodToUpdate = existingPayoutMethods.find(pm => pm.id === payoutMethodId);
await payoutMethodToUpdate.update(pick(args.vendor.payoutMethod, ...));
```

kewitz: Not really. We shouldn't update payout methods. This conditional is only making sure there's a single Payout Method per vendor available.

PR #9516, opened 2023-11-22
Fix `Account.expenses` direction parameter

Betree:
Omitting direction or providing `RECEIVED` was returning all expenses, for all accounts.

### DIFF
--- a/server/graphql/v2/interface/Account.ts
+++ b/server/graphql/v2/interface/Account.ts
@@ -897,13 +897,15 @@ export const AccountFields = {
       ...ExpensesCollectionQueryArgs,
     },
     resolve(collective, args, req) {
-      if (args.direction) {
-        const direction =
-          args.direction === 'SUBMITTED'
-            ? { fromAccount: { legacyId: collective.id } }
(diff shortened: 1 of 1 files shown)


server/graphql/v2/interface/Account.ts
```diff
@@ -897,13 +897,15 @@ export const AccountFields = {
       ...ExpensesCollectionQueryArgs,
     },
     resolve(collective, args, req) {
-      if (args.direction) {
-        const direction =
-          args.direction === 'SUBMITTED'
-            ? { fromAccount: { legacyId: collective.id } }
-            : { toAccount: { legacyId: collective.id } };
```
Betree, line 904: Here it was `toAccount` instead of `account`

server/graphql/v2/interface/Account.ts
```diff
@@ -897,13 +897,15 @@ export const AccountFields = {
       ...ExpensesCollectionQueryArgs,
     },
     resolve(collective, args, req) {
-      if (args.direction) {
```
Betree, line 900: We were not handling the case where `direction` is not set

PR #9517, opened 2023-11-22
Update ED and address info for Open Source Collective

BenJam:
Looks like I started doing this back in 2021!


BenJam:
urgh, so far behind, will create anew.

PR #9524, opened 2023-11-23
Update orphan transactions model check

znarf:
Follow up of https://github.com/opencollective/opencollective-api/pull/9472


Betree (approved):
That works. I feel like having a subquery would have been clearer, something like:

```sql
SELECT *
FROM "Transactions" t1
WHERE NOT EXISTS (
  SELECT * FROM "Transactions" t2
  WHERE t1."TransactionGroup" = t2."TransactionGroup" AND ...
)
```

znarf:
I don't plan to merge this or at least not activate it until the issue is cleaned up.

znarf:
Added a date to only get alerts for new occurrences, we should be clean up to 2023-08.

PR #9526, opened 2023-11-23
Improve Express session cookie configuration

Betree:
I'm hoping that this can help with https://github.com/opencollective/opencollective/issues/6870; a hypothesis is that we're randomly dropping sessions from Redis.


Betree:
To do after merging: check if the TTL is properly set. If not iterate to enforce a TTL on redis.

PR #9530, opened 2023-11-27
Expense: do not allow URL for Invoice items

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7131
Require https://github.com/opencollective/opencollective-frontend/pull/9657

### DIFF
--- a/migrations/20231127092332-move-invoice-item-url-to-attached-file.js
+++ b/migrations/20231127092332-move-invoice-item-url-to-attached-file.js
@@ -24,6 +24,7 @@ module.exports = {
       FROM "Expenses" e
       WHERE e.id = ei."ExpenseId"
       AND e.type = 'INVOICE'
+      AND ei.url IS NOT NULL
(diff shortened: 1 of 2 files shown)


migrations/20231127092332-move-invoice-item-url-to-attached-file.js
```diff
@@ -0,0 +1,35 @@
...
+      AND ei.url IS NOT NULL
+    `);
+
+    // Remove invoice item urls
+    await queryInterface.sequelize.query(`
+      UPDATE "ExpenseItems" ei
+      SET url = NULL
+      FROM "Expenses" e
+      WHERE e.id = ei."ExpenseId"
+      AND e.type = 'INVOICE'
```
kewitz, line 26:
```suggestion
      AND e.type = 'INVOICE'
      AND ei.url IS NOT NULL
```

kewitz (approved):

PR #9531, opened 2023-11-27
Remove cloudinary

znarf:
Related with https://github.com/opencollective/opencollective/issues/4568


znarf:
Merging that as a future proof simplification.

We're still using Cloudinary (because it supports PDFs) on the images service until further notice.

PR #9536, opened 2023-11-28
Handle transactions with amount=0

Betree:
- Resolve https://github.com/opencollective/opencollective/issues/7130 (migrating data)
- Fix/test the case where we're creating a transaction with payment processor fee only (amount=0)

### DIFF
--- a/server/models/Transaction.ts
+++ b/server/models/Transaction.ts
@@ -695,7 +695,15 @@ Transaction.exportCSV = (transactions, collectivesById) => {
  *
  */
 Transaction.createDoubleEntry = async (transaction, opts) => {
-  transaction.type = transaction.amount > 0 ? TransactionTypes.CREDIT : TransactionTypes.DEBIT;
+  // Force transaction type based on amount sign
+  if (transaction.amount > 0) {
+    transaction.type = CREDIT;
+  } else if (transaction.amount < 0) {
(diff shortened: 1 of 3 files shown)


server/models/Transaction.ts
```diff
@@ -695,7 +695,15 @@ Transaction.exportCSV = (transactions, collectivesById) => {
...
 Transaction.createDoubleEntry = async (transaction, opts) => {
-  transaction.type = transaction.amount > 0 ? TransactionTypes.CREDIT : TransactionTypes.DEBIT;
+  // Force transaction type based on amount sign
+  if (transaction.amount > 0) {
+    transaction.type = CREDIT;
+  } else if (transaction.amount < 0) {
+    transaction.type = DEBIT;
+  } else if (!transaction.type) {
+    throw new Error('Transaction type must be set when amount is 0');
+  }
```
Betree, line 705: `type` is always set in `createTransactionsFromPaidExpense`. We do have some old transactions with different kinds matching this condition, but they appear to be edge cases from before `2022-01-01`.


```sql
select * from "Transactions" where amount = 0 AND "deletedAt" IS NULL AND "RefundTransactionId" IS NULL ORDER BY id DESC
```

znarf (approved):
Looks good.

PR #9542, opened 2023-11-29
Collective Page: contributors section to ignore settlements and consider different currencies

kewitz:
Before:
image

After:
image

### DIFF
--- a/server/graphql/loaders/contributors.ts
+++ b/server/graphql/loaders/contributors.ts
@@ -2,12 +2,17 @@ import DataLoader from 'dataloader';
 import { uniq, zipObject } from 'lodash';
 
 import { ContributorsCacheEntry, getContributorsForCollective } from '../../lib/contributors';
+import models from '../../models';
 
 export default {
-  forCollectiveId: (): DataLoader<number, ContributorsCacheEntry> =>
+  forCollectiveId: (req): DataLoader<number, ContributorsCacheEntry> =>
     new DataLoader(async collectiveIds => {
       const uniqueIds = uniq(collectiveIds);
-      const allContributors = await Promise.all(uniqueIds.map(id => getContributorsForCollective(id)));
+      const collectives = await (req.loaders?.Collective?.byId?.loadMany(uniqueIds) ||
+        models.Collective.findAll({ where: { id: uniqueIds } }));
+      const allContributors = await Promise.all(
+        collectives.map(collective => getContributorsForCollective(collective)),
+      );
       const contributorsByIds = zipObject(uniqueIds, allContributors);
       return collectiveIds.map(id => contributorsByIds[id]);
     }),
--- a/server/lib/contributors.ts
+++ b/server/lib/contributors.ts
@@ -8,7 +8,7 @@
  */
 
 import MemberRoles from '../constants/roles';
-import { sequelize } from '../models';
+import { Collective, sequelize } from '../models';
 
 import cache from './cache';
 import { filterUntil } from './utils';
@@ -29,6 +29,7 @@ export interface Contributor {
   type: string;
   since: string;
   totalAmountDonated: number;
+  totalAmountDonatedInHostCurrency: number;
   description: string | null;
   collectiveSlug: string | null;
   publicMessage: string | null;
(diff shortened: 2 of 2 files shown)


znarf (approved):
Looks better!

kewitz:
Requires https://github.com/opencollective/opencollective-frontend/pull/9680

server/lib/contributors.ts
```diff
@@ -99,7 +99,7 @@ const contributorsQuery = `
     c."isIncognito" as "isIncognito",
     BOOL_OR(COALESCE((c."data" ->> 'isGuest') :: boolean, FALSE)) AS "isGuest",
     COALESCE(MAX(m.description), MAX(tiers.name)) AS "description",
-    COALESCE(sum(transactions.amount) / count(DISTINCT m.id), 0) AS "totalAmountDonated"
+    COALESCE(sum(transactions."amountInHostCurrency" - transactions."paymentProcessorFeeInHostCurrency" - transactions."hostFeeInHostCurrency") / count(DISTINCT m.id), 0) AS "totalAmountDonated"
```
Betree, line 102: For this metric, I wouldn't use the net amount. From the perspective of a financial contributor, if I contribute $10k to a collective, I expect the "Top contributors" section to me as $10k, not $9984.

```suggestion
    COALESCE(sum(transactions."amountInHostCurrency") / count(DISTINCT m.id), 0) AS "totalAmountDonated"
```

Betree, line 102: If we switch the logic to use the host currency, we should also update:
- The docs in https://github.com/opencollective/opencollective-api/blob/0a210b25d5ad8e9c4a567f9fb72e2793e81533af/server/graphql/v1/types.js#L489 and https://github.com/opencollective/opencollective-api/blob/0a210b25d5ad8e9c4a567f9fb72e2793e81533af/server/graphql/v2/object/Contributor.ts#L45 to say "in host currency".
- The frontend (e.g. [here](https://github.com/opencollective/opencollective-frontend/blob/508c2bd16947eab201c8dc84f850342638daa521/components/collective-page/sections/TopFinancialContributors.js#L28)), as all these amounts are always displayed using `collective.currency`. We're going to display the wrong information if we don't do this (e.g. `MXN 100` when it's `USD 100`).

Related to this last point, I would personally prefer to make sure we keep returning this amount in the collective currency. Since the goal of this metric is to be displayed on the collective page, I think they'll be confused if we show the amounts in the host currency. To achieve that, we can either rely on `netAmountInCollectiveCurrency` or convert all amounts after the GraphQL query. Another way could be to introduce an `amountV2` on contributors, like what we're doing for https://github.com/opencollective/opencollective/issues/5198.

Betree (changes requested):
Nice catch! I've added some blocking feedback below; I think this needs an iteration.

server/lib/contributors.ts
kewitz: You're correct Ben, I applied the logic to calculate the gross amount of the DEBIT transaction, even though we're explicitly selecting the CREDIT. This makes "amountInHostCurrency" the gross amount. Thanks for pointing out.

kewitz: Good point. This weird currency bug only affects the fiscal-host, which is the only entity that will end up listing transactions in different currencies.
I think it is a good idea to keep displaying in the collective currency, but I'm more inclined to calculate the total amount based on the host amount, since this is the money that is actually stored.
I'll try to join the currency FX here and make it work.

Betree (commented):
A few questions below. Good changes.

server/graphql/loaders/contributors.ts
```diff
@@ -2,12 +2,17 @@ import DataLoader from 'dataloader';
...
 import { ContributorsCacheEntry, getContributorsForCollective } from '../../lib/contributors';
+import models from '../../models';
 
 export default {
-  forCollectiveId: (): DataLoader<number, ContributorsCacheEntry> =>
+  forCollectiveId: (req): DataLoader<number, ContributorsCacheEntry> =>
     new DataLoader(async collectiveIds => {
       const uniqueIds = uniq(collectiveIds);
-      const allContributors = await Promise.all(uniqueIds.map(id => getContributorsForCollective(id)));
+      const collectives = await (req.loaders?.Collective?.byId?.loadMany(uniqueIds) ||
```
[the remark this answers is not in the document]
kewitz: This doesn't work very well in our test environment. This needs to be further investigated but I think it has to do with how we generate and inject these loaders into the graphQL request.

server/lib/contributors.ts
znarf: @kewitz @Betree This is the opposite, we were explicitly adding back paymentProcessorFeeInHostCurrency and hostFeeInHostCurrency to make sure that this was the full amount. The previous code looked better.

If there is a doubt about it, can we cover this with a test?

znarf: Disregard my comment. Except the part about test!

PR #9548, opened 2023-11-30
Update typescript to 5.3.2

znarf:
2 breaking errors:

- server/graphql/v2/object/AccountingCategory.ts:35:1 - error TS2303: Circular definition of import alias 'GraphQLAccountingCategory'.

`export default GraphQLAccountingCategory;`

 - server/graphql/v2/object/Expense.ts:534:10 - error TS2303: Circular definition of import alias 'GraphQLExpense'.

`export { GraphQLExpense };`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -170,7 +170,7 @@
         "sinon": "^17.0.1",
         "sinon-chai": "^3.7.0",
         "supertest": "^6.3.3",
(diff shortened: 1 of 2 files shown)


znarf:
Solve in https://github.com/opencollective/opencollective-api/pull/9728

PR #9554, opened 2023-12-02
Migrate codecov to GitHub actions

znarf:
The `codecov` package is deprecated and this is the new recommended way.


Betree (approved):
Seems to work fine!

PR #9557, opened 2023-12-04
Timeline: use dedicated Redis instance if configured

kewitz:


server/lib/redis.ts
```diff
@@ -4,41 +4,63 @@ import { createClient, RedisClientType } from 'redis';
...
+  DEFAULT = 'DEFAULT',
+  TIMELINE = 'TIMELINE',
+}
+
+const RedisInstanceKey = {
+  [RedisInstanceType.DEFAULT]: 'redis.serverUrl',
+  [RedisInstanceType.TIMELINE]: 'redis.serverUrlTimeline',
+};
+
+const redisClients: Record<string, RedisClientType> = {};
```
znarf, line 17:
```suggestion
const redisInstances: Record<string, RedisClientType> = {};
```

Perhaps, for consistency.

PR #9565, opened 2023-12-06
Optimize reset password scope check

znarf:
it's un-necessary to minify GraphQL query if the scope is not reset-password


Betree (commented):
Nice optimization!

PR #9569, opened 2023-12-07
Platform Tips support for OSC

znarf:
Proposed configuration for OSC:

Fow now (inchanged):
- hostFeePercent: 10
- hostFeeSharePercent: 50

Later:
- hostFeePercent: 8 (or any other value)
- addedFundsHostFeePercent: 10
- bankTransfersHostFeePercent: 10
- stripeNotPlatformTipEligibleHostFeePercent: 10
- paypalNotPlatformTipEligibleHostFeePercent: 10
- hostFeeSharePercent: 50

Then no configuration, but ALWAYS hostFeeSharePercent=0 on contributions eligible to platform tips as a general rule.

Collectives with a custom host fee (like 4%), will always have their custom host fee for all transactions unless they're configured differently at the collective level.

**Again, this PR ensure there is never "Host Fee Share" on contributions eligible to platform tips as a general rule.**

### DIFF
--- a/server/graphql/common/orders.ts
+++ b/server/graphql/common/orders.ts
@@ -80,6 +80,9 @@ export async function addFunds(order: AddFundsInput, remoteUser: User) {
     orderData.data.tax = getOrderTaxInfoFromTaxInput(order.tax, fromCollective, collective, host);
   }
 
+  // Added Funds are not eligible to Platform Tips
+  orderData.platformTipEligible = false;
+
   const orderCreated = await models.Order.create(orderData);
 
   const hostPaymentMethod = await host.getOrCreateHostPaymentMethod();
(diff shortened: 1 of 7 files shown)


Betree (approved):
I was curious about whether this could impact other hosts, so I checked with:

```sql
select host.slug, COUNT(o.id), date_trunc('month', o."createdAt")
from "Orders" o
INNER JOIN "Collectives" c ON c.id = o."CollectiveId"
INNER JOIN "Collectives" host ON host.id = c."HostCollectiveId"
INNER JOIN "Transactions" t ON t."OrderId" = o.id
where o."platformTipEligible" IS NOT TRUE
AND t.kind = 'CONTRIBUTION'
AND host.slug != 'opencollective'
AND c."deletedAt" IS NULL
AND o."deletedAt" IS NULL
AND t."deletedAt" IS NULL
AND host."deletedAt" IS NULL
AND o."createdAt" >= '2023-01-01'
AND (o.data ->> 'isPendingContribution')::boolean IS NOT TRUE
GROUP BY host.id, date_trunc('month', o."createdAt")
ORDER BY date_trunc('month', o."createdAt") DESC
```

There may be a case for https://opencollective.com/opendevufcg, but it's small and likely negligible.

znarf:
Make sure that manual bank transfers have platformEligible=true and pending contributions (created by fiscal host)  platformEligible=false.

PR #9590, opened 2023-12-13
WIP: CollectivePickerAsync for GraphQL V2

kewitz:


server/graphql/v2/query/collection/AccountsCollectionQuery.ts
```diff
@@ -30,6 +30,11 @@ export const CommonAccountsCollectionQueryArgs = {
     type: GraphQLBoolean,
     description: 'Included collectives which are archived',
   },
+  skipGuests: {
+    type: GraphQLBoolean,
+    description: 'Included collectives which are archived',
+    defaultValue: true,
+  },
```
Betree, line 37: Don't we need to add these params in `resolve`?

Betree (commented):
Not sure how the changes to exenses relate?

PR #9603, opened 2023-12-15
Set bacs debit mandate when confirming payment intent

hdiniz:
https://opencollective.freshdesk.com/a/tickets/303653
https://opencollective.slack.com/archives/C0429NTTABF/p1702582604444719

> Error processing Stripe bacs_debit: The parameter `mandate` cannot be passed when creating a PaymentIntent unless `confirm` is set to true


SudharakaP (approved):
Haven't tested it but makes sense given that we no longer passing the mandate on the creation of the payment intent. We can try it out for the user in question and see. Ping me after you merge an deploy so I can let them know. 🤔

SudharakaP:
@hdiniz : Are we good to merge and give this a try ?

PR #9605, opened 2023-12-15
Fix isTwoFactorAuthEnabled graphql field

hdiniz:
Related https://github.com/opencollective/opencollective-api/pull/8798/files#r1428104585


SudharakaP (approved):
💯

PR #9608, opened 2023-12-18
GraphqlV2: return hosts as Host and not as Organization

znarf:
- Make Hosts returned as Host instead of Organization

Replace (BREAKING CHANGE):

```graphql
... on Organization {
  host {
    id
    hostFeePercent
  }
}
```

with simpler:

```graphql
... on Host {
  id
  hostFeePercent
}
```

or:

```graphql
... on Host {
  host {
    id
    hostFeePercent
  }
}
```

- Make Hosts be an AccountWithHost to simplify things on the frontend side

Idea is to avoid such weird code:

```graphql
... on AccountWithHost {
  host {
    id
    slug
    paypalClientId
    supportedPaymentMethods
  }
}
... on Organization {
  host {
    id
    slug
    paypalClientId
    supportedPaymentMethods
  }
}
```

And replace with just:

```graphql
... on AccountWithHost {
  host {
    id
    slug
    paypalClientId
    supportedPaymentMethods
  }
}
```

### DIFF
--- a/server/graphql/schemaV2.graphql
+++ b/server/graphql/schemaV2.graphql
@@ -1625,7 +1625,7 @@ type AccountingCategory {
 """
 This represents an Host account
 """
-type Host implements Account & AccountWithContributions {
+type Host implements Account & AccountWithHost & AccountWithContributions {
   id: String!
   legacyId: Int!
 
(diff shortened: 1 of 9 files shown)


znarf:
This is a breaking change and Frontend (at least) needs to be pushed at the same time.
https://github.com/opencollective/opencollective-frontend/pull/9769

PR #9611, opened 2023-12-18
Wise: Custom Transfer Details support

kewitz:


server/graphql/v2/object/Expense.ts
```diff
@@ -493,6 +494,31 @@ const GraphQLExpense = new GraphQLObjectType<ExpenseModel, express.Request>({
           }
         },
       },
+      validateTransferRequirements: {
+        type: GraphQLJSON,
+        args: {
+          details: {
+            type: GraphQLJSON,
```
hdiniz, line 501: Is it possible to add more strict typing to the return and input of this field?

kewitz: I typed the return. I didn't type the argument because it is dynamically generated by whatever it is requested by Wise and it varies depending on the currency.

PR #9619, opened 2023-12-21
prepare host-report for separate taxes and payment processor fees

znarf:


Betree:
@znarf code updated. For OCEU, I get:
> { VAT: { collected: 3803, paid: -53010 } }

PR #9625, opened 2023-12-22
Transactions: Ignore activities for some kinds

Betree:
Related to https://github.com/opencollective/opencollective/issues/7162

### DIFF
--- a/server/models/Transaction.ts
+++ b/server/models/Transaction.ts
@@ -1459,7 +1459,7 @@ Transaction.createFromContributionPayload = async (
 };
 
 Transaction.createActivity = (transaction, options) => {
-  if (transaction.deletedAt) {
+  if (transaction.deletedAt || ['TAX', 'PAYMENT_PROCESSOR_FEE'].includes(transaction.kind)) {
     return Promise.resolve();
   }
   return (


server/models/Transaction.ts
```diff
@@ -1459,7 +1459,7 @@ Transaction.createFromContributionPayload = async (
 };
 
 Transaction.createActivity = (transaction, options) => {
-  if (transaction.deletedAt) {
+  if (transaction.deletedAt || ['TAX', 'PAYMENT_PROCESSOR_FEE'].includes(transaction.kind)) {
```
znarf, line 1462: You could use constants instead of strings here.

znarf (approved):
Looks good.

Should that be extended to other kinds? Could be seen as a breaking change but is this used really?

server/models/Transaction.ts
Betree: Now that the model is typed with Typescript, I don't see using constants as a strong requirement; it makes the code more verbose, and we'll get an error if there's a typo in the string anyway.

Betree:
> Looks good.
> 
> Should that be extended to other kinds? Could be seen as a breaking change but is this used really?

We could, but it's impossible to know whether some webhooks may depend on that or not. I'm playing the safe card for now and will iterate with https://github.com/opencollective/opencollective/issues/7162.

PR #9643, opened 2023-12-29
Emails: remove unused debug preview

Betree:
As far as I'm aware, no one uses this, and it doesn't seem necessary now that we have maildev in place.


znarf (approved):
Good riddance!

PR #9649, opened 2024-01-02
Removing `Missing CODA_IN_APP_SURVEY_TOKEN` error message for development in Toast

samkevin1:
Related: https://github.com/opencollective/opencollective/issues/7048


server/graphql/v2/mutation/SendSurveyResponseMutation.ts
```diff
@@ -53,6 +53,10 @@ export const sendSurveyResponseMutation = {
     };
 
     try {
+      if (config.env === 'development') {
+        return true;
+      }
+
```
Betree, line 59: This is a blocker to me: we need the ability to work on this feature in local dev environments. Shared my concerns in https://github.com/opencollective/opencollective/issues/7048#issuecomment-1876453672.

server/graphql/v2/mutation/SendSurveyResponseMutation.ts
```diff
@@ -30,8 +30,8 @@ export const sendSurveyResponseMutation = {
     }
     const CODA_TOKEN = process.env.CODA_IN_APP_SURVEY_TOKEN;
 
-    if (!CODA_TOKEN) {
-      throw new UnexpectedError('Missing CODA_IN_APP_SURVEY_TOKEN');
+    if (!CODA_TOKEN && config.env !== 'development') {
+      throw new UnexpectedError('Missing CODA_IN_APP_SURVEY_TOKEN environment variable');
```
Betree, line 34: I would argue for the opposite: never show such a technical error message to end-users.

```suggestion
    if (!CODA_TOKEN) {
      throw new UnexpectedError(config.env !== 'development' ? 'Missing CODA_IN_APP_SURVEY_TOKEN environment variable' : 'An unexpected error happened while submitting the survey');
```

Betree (changes requested):
Thanks for the contribution! I have a few concerns with the scope of the root issue that I've shared in https://github.com/opencollective/opencollective/issues/7048#issuecomment-1876453672.

server/graphql/v2/mutation/SendSurveyResponseMutation.ts
samkevin1: Shouldn't it be `(config.env === 'development' ? 'Missing CODA_IN_APP_SURVEY_TOKEN environment variable' : 'An unexpected error happened while submitting the survey')` instead?

Betree: Indeed

Betree (approved):

Betree:
Good for me, I'll let @gustavlrsn give a 2nd review

gustavlrsn (approved):
Looks good to me!

PR #9651, opened 2024-01-03
Tweaks and Fixes from Hosted Collectives dashboard

kewitz:


server/graphql/v2/interface/IsMemberOf.js
```diff
@@ -102,6 +111,20 @@ export const IsMemberOfFields = {
...
+      if (!isNil(args.isFrozen)) {
+        if (args.isFrozen) {
+          set(collectiveConditions, `data.features.${FEATURE.ALL}`, false);
+        } else {
+          set(collectiveConditions, `data.features.${FEATURE.ALL}`, { [Op.is]: null });
+        }
+      }
+      if (args.isDeleted === true) {
+        where.deletedAt = { [Op.not]: null };
+        if (args.role.includes('HOST')) {
```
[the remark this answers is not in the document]
kewitz: Yep... I considered writing a dedicated resolver for hosted collectives, should I do that?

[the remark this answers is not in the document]
gustavlrsn: Nice enhancement 👍

server/graphql/v2/object/Host.js
```diff
@@ -1137,6 +1145,130 @@ export const GraphQLHost = new GraphQLObjectType({
...
+          orderBy: {
+            type: GraphQLOrderByInput,
+            description: 'Order of the results',
+          },
+        },
+        async resolve(host, args, req) {
+          const isAdmin = req.remoteUser?.isAdmin(host.id);
+          if (!isAdmin) {
+            throw new Unauthorized('You need to be logged in as an admin of the host to see its hosted accounts');
+          }
```
gustavlrsn, line 1187: Could this be public? Seems useful to use this query also on the public profile pages of a host to display hosted collectives instead of using `memberOf`

server/graphql/v2/object/Host.js
```diff
@@ -1137,6 +1145,130 @@ export const GraphQLHost = new GraphQLObjectType({
...
+            } else if (args.hostFeesStructure === HOST_FEE_STRUCTURE.MONTHLY_RETAINER) {
+              throw new ValidationFailed('The MONTHLY_RETAINER fees structure is not supported yet');
+            }
+          }
+
+          if (args.isUnhosted) {
+            const unhostedActivities = await models.Activity.findAll({
+              where: { type: 'collective.unhosted', HostCollectiveId: host.id },
+              attributes: ['CollectiveId'],
+            });
```
gustavlrsn, line 1221: Not a blocker, and not sure how true it is, but I'm wondering if using the Members table to find deleted memeberships as in the previous solution might be faster than searching in the larger Activities table

server/graphql/v2/object/Host.js
```diff
@@ -1137,6 +1145,130 @@ export const GraphQLHost = new GraphQLObjectType({
...
+        },
+        async resolve(host, args, req) {
+          const isAdmin = req.remoteUser?.isAdmin(host.id);
+          if (!isAdmin) {
+            throw new Unauthorized('You need to be logged in as an admin of the host to see its hosted accounts');
+          }
+
+          const where = {
+            HostCollectiveId: host.id,
+          };
```
gustavlrsn, line 1191:
```suggestion
          const where = {
            HostCollectiveId: host.id,
            id: { [Op.not]: host.id },
          };
```

server/graphql/v2/object/Host.js
```diff
@@ -1137,6 +1145,130 @@ export const GraphQLHost = new GraphQLObjectType({
...
+            };
+          }
+
+          if (!isNil(args.isFrozen)) {
+            if (args.isFrozen) {
+              set(where, `data.features.${FEATURE.ALL}`, false);
+            } else {
+              set(where, `data.features.${FEATURE.ALL}`, { [Op.is]: null });
+            }
+          }
```
gustavlrsn, line 1205: Would it be sufficient to only add this condition when isFrozen is `true`? Since isFrozen has a default value of false, it will always add this condition, so would be nice to avoid if possible

```suggestion
          if (args.isFrozen) {
            set(where, `data.features.${FEATURE.ALL}`, false);
          }
```

server/graphql/v2/object/Host.js
kewitz: Hey @gustavlrsn, we have two problems with the Members table I'm trying to avoid:
1. A host Member record is created before the host application is approved, which causes this to list collectives that applied and were rejected.
2. You can have multiple instances of the same membership deleted, this caused some collectives to appear multiple times.

server/graphql/v2/object/Host.js
kewitz: Not really, because this would prevent us from removing frozen collectives from the active tab.

PR #9653, opened 2024-01-04
Wise: round total amount to pay

Betree:
Fix https://github.com/opencollective/opencollective/issues/6748


Betree:
cc @kewitz, I assumed we can end up in this situation when summing the quotes with cases like:

```
> 0.2+0.1
0.30000000000000004
```

kewitz (approved):

kewitz:
@Betree sometimes you gotta give up for all the Rust people on the internet hammering down on JS.

Betree:
@kewitz [Even Rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=eacdc395eb310fc6f80caca94526f08c) is not safe from this bug :smiling_face_with_tear:

PR #9667, opened 2024-01-09
Modify root redirect cookie

gustavlrsn:
- Changing the name of the `rootRedirect` cookie to `rootRedirectDashboard` to signify that we're only interested in the presence of the cookie, not it's value.
- Making sure to clear the cookie if user turns off the preview feature

### DIFF
--- a/config/custom-environment-variables.json
+++ b/config/custom-environment-variables.json
@@ -231,5 +231,10 @@
     "url": "STATSD_URL",
     "port": "STATSD_PORT",
     "prefix": "STATSD_PREFIX"
+  },
+  "features": {
+    "dashboard": {
+      "redirect": "FEATURES_DASHBOARD_REDIRECT"
+    }
   }
 }
--- a/server/middleware/authentication.js
+++ b/server/middleware/authentication.js
@@ -11,6 +11,7 @@ import * as connectedAccounts from '../controllers/connectedAccounts';
 import { verifyJwt } from '../lib/auth';
 import errors from '../lib/errors';
 import logger from '../lib/logger';
+import { clearRedirectCookie, setRedirectCookie } from '../lib/redirect-cookie';
 import { reportMessageToSentry } from '../lib/sentry';
 import { TWITTER_SCOPES } from '../lib/twitter';
 import { getBearerTokenFromRequestHeaders, parseToBoolean } from '../lib/utils';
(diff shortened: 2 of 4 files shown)


gustavlrsn:
* `main`
  * **PR #9667** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-api/9667?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-api/9667?utm_source=stack-comment).

znarf:
I'm wondering what will be the strategy when Dashboard will become the default

gustavlrsn:
Pushed a fix to have a config variable able to set the default status of the Dashboard

server/middleware/authentication.js
```diff
@@ -152,13 +153,10 @@ const _authenticateUserByJwt = async (req, res, next) => {
...
 
   const { earlyAccess = {} } = user.collective.settings || {};
-  if (earlyAccess.dashboard) {
-    res.cookie('rootRedirect', 'dashboard', {
-      secure: true,
-      httpOnly: true,
-      sameSite: 'strict',
-      maxAge: 24 * 60 * 60 * 1000 * 365,
-    });
+  if (earlyAccess.dashboard || (parseToBoolean(config.dashboardIsDefault) && earlyAccess.dashboard !== false)) {
```
znarf, line 156: Sorry if this was done this way based on my input but we should avoid putting config at the root level, this looks a bit unorganized. Suggesting something like `config.features.dashboard` or `config.features.dashboard.redirect`.

config/custom-environment-variables.json
```diff
@@ -231,5 +231,6 @@
     "url": "STATSD_URL",
     "port": "STATSD_PORT",
     "prefix": "STATSD_PREFIX"
-  }
+  },
+  "dashboardIsDefault": "DASHBOARD_IS_DEFAULT"
```
znarf, line 235:
```suggestion
  "features": {
    "dashboard": {
      "redirect": "FEATURES_DASHBOARD_REDIRECT"
    }
  }
```

znarf: Something like this, not sure of the syntax.

server/middleware/authentication.js
gustavlrsn: Changed to `config.features.dashboard.redirect` 👍

znarf (approved):
Ok, looks good. As I understand, this works well because in staging and production, we have an advanced proxy system that is properly forwarding the Cookie header. 

At it stands, this would not work in other environments.

gustavlrsn:
@znarf this should be working in other environments than staging and production. I've tested using the staging API with a local frontend, and it seems to work.

The cookie set on the API is passed along the response as long as the SameSite attribute is fulfilled, which is [configured like this](https://github.com/opencollective/opencollective-api/commit/86f3fd6a20808ade4ea300db5ff54814a87410e4#diff-034dd2229228863fb749aaad973334b70af27934fc92251dca67c9f52f0d5290R6-R11):
```
sameSite: config.env === 'production' ? 'lax' : 'none',
```

Setting sameSite to "none" is what enables the cookie to be passed along cross-domains, from on environment other than production it is set to "none", so that the cookie set on https://staging-api.opencollective.com will be passed along to `localhost` for development. 

From what I understand site refers to the last parts of a domain, i.e. "api.opencollective.com" and "www.opencollective.com" is the same site which means that "lax" (that prevents cross-site requests) should hopefully work there.

PR #9674, opened 2024-01-11
Fix all occurrences of UNUSED_VAR_ASSIGN reported by Deepscan

znarf:


server/graphql/v2/object/TierStats.ts
```diff
@@ -19,6 +19,7 @@ export const GraphQLTierStats = new GraphQLObjectType({
         type: new GraphQLNonNull(GraphQLAmount),
         async resolve(tier, args, req) {
           const totalDonated = await req.loaders.Tier.totalDonated.load(tier.id);
+          // Why are we defensive here? there is no Tier without currency in the database
```
znarf, line 22: @Betree should we consider removing that code, why are we fetching currency again?

Betree: I did this because, while we don't have any entry like that today, tier's `currency` is currently defined as a nullable column. I'm ok with removing it, but then we should ideally make the column non-nullable to make things consistent.

znarf: All good

znarf: Replacing my comment with another one.

PR #9675, opened 2024-01-11
Fix remaining issues reported by Deepscan

znarf:


server/lib/ocr/klippa/KlippaOCRService.ts
```diff
@@ -18,8 +18,9 @@ export class KlippaOCRService implements ExpenseOCRService {
     private apiKey: string,
     private user: User,
   ) {
-    this.apiKey = apiKey;
-    this.user = user;
+    // this.apiKey = apiKey;
+    // this.user = user;
+    // See: https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd
```
Betree, line 23: Not a fan of this TS syntax, unless you know about it it's an easy one to miss.

Suggestion for a simpler comment, in case we do want to adopt this.

```suggestion
    // Typescript constructor automatically assign parameters to `this`
```

server/lib/payments.js
```diff
@@ -301,7 +301,7 @@ export async function refundPaymentProcessorFee(
     // TODO: We're now able to support this more easily, we should implement
     const processorFeeInHostCurrency =
       processorFeeTransaction?.amountInHostCurrency || transaction.paymentProcessorFeeInHostCurrency;
-    if (refundedPaymentProcessorFee && refundedPaymentProcessorFee !== processorFeeInHostCurrency) {
```
Betree, line 304: It's not exactly the same thing: If I get refundedPaymentProcessorFee=0 and processorFeeInHostCurrency=10, I don't want to show any warning as it is not a partial refund. The warning should only be there if `refundedPaymentProcessorFee` is truthy (> 0) **and** different than `processorFeeInHostCurrency`

Betree (commented):
A bit concerned about the refund thing below, the rest looks good

server/lib/payments.js
znarf: Good feedback but it's now what you think. The change is done because of the following report:

```Condition 'refundedPaymentProcessorFee' is always true at this point because it is already checked at line 290 .```

server/lib/ocr/klippa/KlippaOCRService.ts
znarf: I understand and was also surprised by this, this is why I decided to keep the current code commented and try to explain what's happening.

znarf: I considered keeping it like it was but "as en engineer" felt it was wrong to set something that was already there.

PR #9679, opened 2024-01-12
Host Settlement: add currency to created items

Betree:
Follow-up on https://github.com/opencollective/opencollective-api/pull/9529


znarf (approved):
Perfect

PR #9694, opened 2024-01-16
Sorted transactions

znarf:
- Share algorithm in SQL functions
- Update GraphQL Transactions resolver to use functions
- Update "TransactionBalances" materialized view to use the new sorting


migrations/20240111120840-sorted-transactions.js
```diff
@@ -0,0 +1,194 @@
...
+          ) OVER (PARTITION BY "CollectiveId", "hostCurrency" ORDER BY "createdAt" ASC, "id" ASC) as "balance"
+          FROM "Transactions", "ActiveCollectives"
+          WHERE "deletedAt" IS NULL
+          AND "CollectiveId" = "ActiveCollectives"."ActiveCollectiveId"
+          ORDER BY "createdAt" ASC
+       )
+    `);
+
+    // Add a unique index on collective ID to the materialized view
+    await queryInterface.sequelize.query(`CREATE UNIQUE INDEX CONCURRENTLY ON "TransactionBalances"(id)`);
```
gustavlrsn, line 128: The index seems to be on the transaction id, not the collective ID as per the comment

gustavlrsn: I see this is copied code from `migrations/20230104225718-transaction-balances-order-by-create-date.js`

Later in that migration we do create an index on the collective ID, so maybe we want that index created here too

znarf: We're carrying this bad comment for ages by copy-pasting it!

znarf: Not clear at that point that an index on CollectiveId is useful so I'm not gonna add one here.

PR #9700, opened 2024-01-18
Update to consistent sorted transactions

znarf:
Editing the migration because it was reverted in production.

Make sure balances are consistent with the default transactions order


gustavlrsn (approved):

Betree (approved):
Looks good

PR #9695, opened 2024-01-16
AccountingCategory: add support for orders

Betree:
API part of https://github.com/opencollective/opencollective/issues/7207
To be merged together with https://github.com/opencollective/opencollective-frontend/pull/9908

PR #9702, opened 2024-01-18
Add created at to accounting category

hdiniz:


server/models/AccountingCategory.ts
```diff
@@ -133,8 +132,11 @@ AccountingCategory.init(
       type: DataTypes.ENUM(...Object.keys(ExpenseTypes)),
       allowNull: true,
       set(values: Array<ExpenseTypes | `${ExpenseTypes}`>): void {
-        // Sequelize doesn't work with empty arrays ("cannot determine type of empty array"). We force `null` if it's empty
-        this.setDataValue('expensesTypes', values?.length ? uniq(values).sort() : null);
+        const value =
+          values && values.length > 0
+            ? sequelize.literal(`ARRAY[${values.map(t => `'${t}'`).join(',')}]::"enum_Expenses_type"[]`)
+            : null;
```
Betree, line 138: `expenseTypes` is passed by the GraphQL API as an enum so this shouldn't be an issue today, but out of safety I think we should protect this against SQL injections. Also, it's still nice to have `uniq` to prevent blotting the field with duplicates:

```suggestion
        let value = null;
        if (values && values.length > 0) {
          const invalidValues = values.filter(v => !ExpenseTypes[v]);
          if (invalidValues.length > 0) {
            throw new Error(`Invalid values for GraphQLExpenseType: ${invalidValues.join(', ')}`);
          }
        
          value = sequelize.literal(`ARRAY[${uniq(values).map(t => `'${t}'`).join(',')}]::"enum_Expenses_type"[]`)
        }

```

server/models/AccountingCategory.ts
```diff
@@ -133,8 +132,11 @@ AccountingCategory.init(
       type: DataTypes.ENUM(...Object.keys(ExpenseTypes)),
```
Betree, line 132: Not sure if that's related to our issue, but I realize this should be:

```suggestion
      type: DataTypes.ARRAY(DataTypes.ENUM(...Object.keys(ExpenseTypes))),
```

Betree: Awesome!

Betree:
I rebased https://github.com/opencollective/opencollective-api/pull/9695 to fix conflicts

PR #9710, opened 2024-01-19
Move constants from snake case to kebab case

znarf:
https://github.com/opencollective/opencollective/blob/main/rfcs/004-file-naming-convention.md


Betree (approved):
Good stuff

PR #9713, opened 2024-01-19
Refactor models/index by alphabetical order

znarf:


Betree (approved):
Good stuff! Can't really check for mistakes by reading the code, but I assume it is only re-ordering.

znarf:
This is manual work prone to errors but I'm confident enough with tests passing.

PR #9714, opened 2024-01-19
Fix Github handles missing a 's'

Betree:
Affects 8000+ entries. Only updates if we have a perfect match between `trim(regexp_replace(h."githubHandle", 's', ''))` and the current handle.


znarf:
Should we also fix SocialLinks ?

znarf (approved):

PR #9721, opened 2024-01-21
refactor: review how models are imported

znarf:
- Only export models as `default`, not as named import
- Within the model folder, only direct imports, not through `models/index.ts`
- Fix typescript issues surfaced by the refactor


server/paymentProviders/transferwise/index.ts
```diff
@@ -297,15 +294,15 @@ async function payExpense(
 
 const getOrCreateActiveBatch = async (
   host: Collective,
-  options?: { connectedAccount?: string; token?: string },
+  options?: { connectedAccount?: ConnectedAccount; token?: string },
```
znarf, line 297: Typescript issue @kewitz. Does that looks good?

PR #9734, opened 2024-01-25
GroupedTruncate=false

znarf:
I started using that to avoid deadlocks, it seemed to improve.


znarf:
Closing this as it seems to take 10 minutes with this approach instead of 4.

PR #9746, opened 2024-01-26
Transaction collection async + cache

znarf:
Add some caching to the transaction collection.

count, kinds and paymentMethodTypes can be super slow on big hosts.

Lessons to apply elsewhere:
- Collection resolver should be fully on demand and async
- Fetch and Count should be always dissociated (sequelize postgres end up with 2 SQL queries anyway)

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -261,5 +261,8 @@
     "dashboard": {
       "redirect": false
     }
+  },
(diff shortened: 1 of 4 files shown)


Betree (commented):
Nice patch!

PR #9747, opened 2024-01-26
Fix Expense.createActivity to fetch latest debit transaction

kewitz:
In the case of refunds, this will cause every repayment activity to contain the first payment transaction information.


znarf:
As discussed on Slack, not clear to me that it's fixing the issue completely.

1) Are payment processor fees really read from the activity? https://opencollective.com/babel/expenses/181149 has in activity but they're not displayed
2) Even if it was read from the Activity, the suggested code in this PR doesn't fetch payment processor fees

kewitz:
@znarf no, this doesn't solve the separate payment processor fee transactions. This just fixes a bug where the wrong transaction is associated in the activity.

PR #9753, opened 2024-01-29
Convert Order model to Typescript 

znarf:
Partially using sequelize-auto but it's not a miracle solution.

### DIFF
--- a/server/lib/handlebars.js
+++ b/server/lib/handlebars.js
@@ -190,7 +190,7 @@ handlebars.registerHelper('formatOrderAmountWithInterval', order => {
   }
 
   const formattedAmount = formatCurrency(order.totalAmount, order.currency);
-  const subscription = order.subscription;
+  const subscription = order.Subscription;
   const interval = subscription?.interval || order.interval;
 
   if (interval !== null) {
--- a/server/lib/handlebars.js
+++ b/server/lib/handlebars.js
@@ -190,7 +190,10 @@ handlebars.registerHelper('formatOrderAmountWithInterval', order => {
   }
 
   const formattedAmount = formatCurrency(order.totalAmount, order.currency);
-  const subscription = order.Subscription;
+  // `order.subscription` or `order.Subscription` ?
+  // Come from:
+  // https://github.com/opencollective/opencollective-api/blob/ef2c3845349886048f181b4622404f965a2cf134/server/models/Collective.ts#L1821-L1823
+  const subscription = order.subscription;
   const interval = subscription?.interval || order.interval;
 
   if (interval !== null) {
--- a/server/models/Order.ts
+++ b/server/models/Order.ts
@@ -1,35 +1,29 @@
 import { TaxType } from '@opencollective/taxes';
 import debugLib from 'debug';
 import { get } from 'lodash';
-import {
-  ForeignKey,
-  HasManyGetAssociationsMixin,
-  HasOneGetAssociationMixin,
-  InferAttributes,
-  InferCreationAttributes,
-  Model,
-  ModelStatic,
-} from 'sequelize';
+import * as Sequelize from 'sequelize';
+import { DataTypes, Model, Op, Optional, QueryTypes } from 'sequelize';
 import Temporal from 'sequelize-temporal';
 
-import { roles } from '../constants';
 import { SupportedCurrency } from '../constants/currencies';
 import OrderStatus from '../constants/order-status';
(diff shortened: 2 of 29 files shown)


test/server/graphql/v2/mutation/CreateCollectiveMutations.test.js
```diff
@@ -52,7 +52,7 @@ const backYourStackCollectiveData = {
 
 describe('server/graphql/v2/mutation/CreateCollectiveMutations', () => {
   beforeEach('reset db', async () => {
-    await utils.resetTestDB();
+    await utils.resetTestDB({ groupedTruncate: false });
```
Betree, line 55: Why these `groupedTruncate` here in tests?

server/models/Order.ts
```diff
@@ -1,312 +1,478 @@
...
+type ActivityId = Activity['id'];
+type CollectiveId = Collective['id'];
+type CommentId = Comment['id'];
+// type PaymentMethodId = PaymentMethodModelInterface['id'];
+type SubscriptionId = SubscriptionInterface['id'];
+type TierId = Tier['id'];
+type TransactionId = TransactionInterface['id'];
+type UserId = User['id'];
+
+class Order extends Model<OrderAttributes, OrderCreationAttributes> implements OrderAttributes {
```
Betree, line 110: Not a blocker, but I consider the `InferAttributes` approach simpler and clearer

server/models/Order.ts
```diff
@@ -1,312 +1,478 @@
...
-interface OrderModelStaticInterface {
-  generateDescription(collective, amount, interval, tier): string;
-  cancelActiveOrdersByCollective(collectiveId: number): Promise<[affectedCount: number]>;
-  cancelActiveOrdersByTierId(tierId: number): Promise<[affectedCount: number]>;
-  cancelNonTransferableActiveOrdersByCollectiveId(collectiveId: number): Promise<[affectedCount: number]>;
+export interface OrderAttributes {
+  id: number;
+  CreatedByUserId?: number;
+  CollectiveId?: number;
+  currency?: string;
```
Betree, line 37: We'll want to update this one for https://github.com/opencollective/opencollective-api/pull/9706

server/models/Order.ts
```diff
@@ -1,312 +1,478 @@
...
-  status: OrderStatus;
-  interval?: string;
-  data:
+type Subscription = SubscriptionInterface;
+
+type AccountingCategoryId = AccountingCategory['id'];
+type ActivityId = Activity['id'];
+type CollectiveId = Collective['id'];
+type CommentId = Comment['id'];
+// type PaymentMethodId = PaymentMethodModelInterface['id'];
```
Betree, line 104: Why comment this one, and `declare setPaymentMethod: Sequelize.BelongsToSetAssociationMixin<PaymentMethodModelInterface, PaymentMethodId>;` below?

server/models/Order.ts
```diff
@@ -1,312 +1,478 @@
...
+      type: DataTypes.DATE,
+      allowNull: true,
+    },
+    privateMessage: {
+      type: DataTypes.TEXT,
+      allowNull: true,
+    },
     TierId: {
       type: DataTypes.INTEGER,
+      allowNull: true,
```
Betree, line 374: Great that we're adding these, it makes things explicit

Betree (commented):
It would be interesting to have @hdiniz's feedback on this one since he originally created the interface for `Orders`.

test/server/graphql/v2/mutation/CreateCollectiveMutations.test.js
znarf: I randomly had a deadlock there. Should remove, unrelated.

server/models/Order.ts
znarf: Having read sequelize documentation on the subject since starting this PR, I agree.

server/models/Order.ts
znarf: todo

server/lib/handlebars.js
```diff
@@ -190,7 +190,7 @@ handlebars.registerHelper('formatOrderAmountWithInterval', order => {
   }
 
   const formattedAmount = formatCurrency(order.totalAmount, order.currency);
-  const subscription = order.subscription;
+  const subscription = order.Subscription;
```
znarf, line 193: To check

hdiniz: Comes from https://github.com/opencollective/opencollective-api/blob/ef2c3845349886048f181b4622404f965a2cf134/server/models/Collective.ts#L1821-L1823

PR #9755, opened 2024-01-29
Add hostOnly flag and instructions to accounting categories

hdiniz:
Project https://github.com/opencollective/opencollective/issues/7207
Part of https://github.com/opencollective/opencollective/issues/7209


server/graphql/common/expenses.ts
```diff
@@ -1208,6 +1210,14 @@ const checkCanUseAccountingCategory = (
...
   } else if (accountingCategory.kind && accountingCategory.kind !== 'EXPENSE') {
     throw new ValidationFailed('This accounting category is not allowed for expenses');
+  } else if (
+    accountingCategory.expensesTypes &&
+    accountingCategory.expensesTypes.length > 0 &&
+    !accountingCategory.expensesTypes.includes(expenseType)
+  ) {
+    throw new ValidationFailed(`This accounting category is not allowed for expenses type: ${expenseType}`);
+  } else if (accountingCategory.hostOnly && !remoteUser?.isAdmin(host.id)) {
+    throw new ValidationFailed('This accounting category can only be used by the host admin');
```
Betree, line 1220: I'm a bit worried about these two checks. If a host admin sets the accounting category and then asks the user to edit any info, the system will throw an error saying they're not allowed to use the accounting category.

Possible solutions:
- Do not add any checks; continue to consider `expensesTypes` and `hostOnly` as "visual"/UI limitations
- Only run these two conditions when changing the expense type (`existingExpense.type !== expenseType`)

server/graphql/v2/input/AccountingCategoryInput.ts
```diff
@@ -42,6 +51,13 @@ export const AccountingCategoryInput = new GraphQLInputObjectType({
       type: GraphQLString,
       description: 'A friendly name for non-accountants (i.e. expense submitters and collective admins)',
     },
+    hostOnly: {
+      type: GraphQLBoolean,
```
Betree, line 55: If there's a default value, we can:

```suggestion
      type: new GraphQLNonNull(GraphQLBoolean),
```

PR #9756, opened 2024-01-29
Expenses Submit on Behalf: Attribute email to existing user

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7095

Considering that the user is:
- Logged out;
- Has access to the email that received the expense draft being submitted;
- Is using/guessing an email that already was used on the platform;
- The submitted draft still needs to be verified through the following user authentication on the platform;

It should be safe to allow an expense draft to be submitted by the logged-out user instead of throwing a "User already exists" error.

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -1728,7 +1728,12 @@ export async function editExpenseDraft(req: express.Request, expenseData: Expens
   };
 
   if (args.expense.payee && isDifferentInvitedPayee(existingExpense, args.expense.payee)) {
-    const payee = args.expense.payee as { email: string; name?: string };
+    let payee = args.expense.payee as { id?: number; email?: string };
+    const existingUser = payee.email && (await models.User.findByEmail(payee.email));
+    const existingUserCollective = existingUser && (await existingUser.getCollective());
+    if (existingUserCollective) {
+      payee = existingUserCollective.minimal;
+    }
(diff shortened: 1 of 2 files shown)


Betree (changes requested):
I see two dangers with this change:
1. I'm just worried about permissions: we must make sure that, by being the UserId (inviter) on an expense for someone else I invited, I don't get any access to their private info (payout method, location, etc). At first glance, it looks ok, but this isn't easy to guarantee.
2. By allowing any users to create draft expenses on any account, we're opening the door to a new attack vector that I'm deeply concerned about:
   1. As a malicious user, create a draft expense with a Babel admin as the payee, putting your bank account info as the payout method.
   2. Reach out to other admins (by impersonating an email/chat account), saying you're struggling to validate your expense, asking them if they can confirm it.
   3. The legit admin goes and validates the expense, reassured by the fact that the **payee is indeed set to someone they know**, another admin who has been paid by the collective in the past.
   4. The expense gets paid to a malicious account.

I think that the proper UX to resolve https://github.com/opencollective/opencollective-api/pull/9756 is not so easy to implement, and should be treated as a real project we brainstorm on rather than a quick fix.

kewitz:
I agree, we need to at least write a proper suite of tests and review the side effects. All things I won't be able to carve extra-time to solve this sprint.
Closing this PR for now.

PR #9776, opened 2024-02-05
Update getTaxFormsRequiredForExpenses to ignore non-tax-liable vendors

kewitz:
Related to https://github.com/opencollective/opencollective/issues/7200


Betree (approved):
Looks good!

Betree (changes requested):
Actually, anything enforced here should also be enforced in `getTaxFormsRequiredForAccounts`

Betree (approved):
:+1:

PR #9789, opened 2024-02-08
Balance time series

gustavlrsn:
For https://github.com/opencollective/opencollective/issues/7265

### Description

Adding the `account.stats.balanceTimeSeries` resolver to provide a way to get the balance as a time series.

This version is not using the TransactionBalances materialized view as described in the issue as a potential way to get performant data, since there are cases of Collectives that are not covered by this view, so we need a general strategy as a fallback either way. Using TransactionBalances could be a potential follow up to improve performance.

### DIFF
--- a/server/graphql/schemaV2.graphql
+++ b/server/graphql/schemaV2.graphql
@@ -1530,6 +1530,7 @@ type Expense {
   The payout method to use for this expense
   """
   payoutMethod: PayoutMethod
+  paymentMethod: PaymentMethod
 
   """
   The virtual card used to pay for this charge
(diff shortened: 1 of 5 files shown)


gustavlrsn:
* `main`
  * **PR #9789** <a href="https://app.graphite.dev/github/pr/opencollective/opencollective-api/9789?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/opencollective/opencollective-api/9789?utm_source=stack-comment).

kewitz (approved):
I like the solution you proposed but it is definitely more complex than a simple SQL query.
I think we should fix the edge cases on TransactionBalances ASAP.

znarf:
> I think we should fix the edge cases on TransactionBalances ASAP.

@kewitz what are those edge cases?

kewitz:
@znarf I'm not familiarized with the TransactionBalances view, but I was referring to:
> since there are cases of Collectives that are not covered by this view

gustavlrsn:
@znarf if I understood you correctly, that would be when a Collective has transactions with more than one host currency.

znarf:
@gustavlrsn Yes. And I think it would be acceptable to not return timeseries for them.

On the other end, if performance is reasonably fast with your approach, I don't have anything against.

gustavlrsn:
@znarf performance seems ok, I'll create a follow up issue for using TransactionBalances when that is available

PR #9791, opened 2024-02-08
Update populate-expense-payment-method to run based on Transaction.createdAt

kewitz:


scripts/payment-methods/populate-expense-payment-method.ts
```diff
@@ -53,9 +53,15 @@ const migrate = async () => {
...
-    include: [PayoutMethod, { association: 'Transactions', where: { type: 'DEBIT', kind: 'EXPENSE' } }],
+    include: [
+      PayoutMethod,
+      {
+        association: 'Transactions',
+        where: { type: 'DEBIT', kind: 'EXPENSE', createdAt: { [Op.gte]: DATE_FROM }, isRefund: false },
+        required: true,
+      },
+    ],
     order: [['updatedAt', 'DESC']],
```
znarf, line 65: Can we order by Transactions date ?

znarf: Just did that

PR #9797, opened 2024-02-09
PayPal: lock order to prevent concurrency issues

Betree:
This PR is meant to prevent concurrency issues when processing PayPal webhook events by implementing a locking mechanism, similar to what we do for Expenses.

Usage:

```typescript
const myOrder = await Order.findByPk(orderId);

// By default, this block will throw if the order is already locked
await myOrder.lock(async () => {
  // Do something with the order
});

// You can wait for the order to get unlocked by adding some parameters:
await myOrder.lock(async () => {
  // Do something with the order
}, { retries: 10, retryDelay: 1000 });
```

### DIFF
--- a/server/models/Order.ts
+++ b/server/models/Order.ts
@@ -21,7 +21,7 @@ import { PLATFORM_TIP_TRANSACTION_PROPERTIES, TransactionTypes } from '../consta
 import * as libPayments from '../lib/payments';
 import sequelize, { DataTypes, Op, QueryTypes } from '../lib/sequelize';
 import { sanitizeTags, validateTags } from '../lib/tags';
-import { capitalize } from '../lib/utils';
+import { capitalize, sleep } from '../lib/utils';
 
 import AccountingCategory from './AccountingCategory';
 import Collective from './Collective';
@@ -110,6 +110,7 @@ export interface OrderModelInterface
         hostFeePercent?: number;
         memo?: string;
         tax?: OrderTax;
+        isLocked?: boolean;
       }
     | any; // TODO: Remove `any` once we have a proper type for this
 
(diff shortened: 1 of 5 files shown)


znarf:
How a locked Order get unlocked in case of a crash?

Betree:
> How a locked Order get unlocked in case of a crash?

The `callback` runs in a `try .. finally`, unlocking the expense regardless of the promise's result. In case of a platform crash, however, it would stay marked as locked forever.

We can resolve that by replacing `isLocked` by a `lockedAt`, and defining that nothing should lock an order for more than x minutes. Any order that gets automatically unlocked like that should be reported to Sentry.

That could be useful for expenses too.

znarf:
Cool, would be great to be crash proof. Even without a pure crash, imagine us restarting the dynos while a webhook is processing.

server/models/Order.ts
```diff
@@ -541,6 +561,75 @@ Order.prototype.getSubscriptionForUser = function (user) {
...
+Order.clearExpiredLocks = function () {
+  return sequelize.query(
+    `
+      UPDATE "Orders"
+      SET data = data
+        - 'lockedAt'
+        || JSONB_BUILD_OBJECT('deadlocks', COALESCE(data -> 'deadlocks', '[]'::JSONB) || TO_JSONB(ARRAY[data ->> 'lockedAt']))
+      WHERE data -> 'lockedAt' IS NOT NULL
+      AND (data ->> 'lockedAt')::TIMESTAMP < NOW() - INTERVAL '30 minutes'
+      AND "deletedAt" IS NULL
```
Betree, line 625: The query takes around 3s to run. 

> QUERY PLAN
> "Seq Scan on ""Orders""  (cost=0.00..95364.90 rows=200166 width=519) (actual time=3443.194..3443.194 rows=0 loops=1)"
> "  Filter: ((""deletedAt"" IS NULL) AND ((data -> 'lockedAt'::text) IS NOT NULL) AND (((data ->> 'lockedAt'::text))::timestamp without time zone < (now() - '00:30:00'::interval)))"
>   Rows Removed by Filter: 735203
> Planning Time: 0.107 ms
> Execution Time: 3443.209 ms
>

Betree:
Made the changes to clear out (and log) deadlocks.

znarf (approved):
Looks good!

PR #9711, opened 2024-01-19
Add Expense.PaymentMethod

znarf:
Implement: https://github.com/opencollective/opencollective/issues/7212

PR #9798, opened 2024-02-09
fix(PayPal/Adaptive): crash in mark as paid

Betree:
Fix a crash reported by Ivan: (see [Sentry](https://open-collective.sentry.io/issues/4960290497/?project=5199682&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=90d&stream_index=3))

Introduced in https://github.com/opencollective/opencollective-api/pull/9711/files#diff-90ee3a074ce4d95ccb38d581b33cb78d00855e488561cc76ea4b0582f31f71b4R2240. This is similar to https://github.com/opencollective/opencollective-api/pull/9711/files#r1465556900, it seems that when we moved the function from `markExpenseAsPaid` to `expense.markAsPaid`, we got rid of the `return` but did not update depending code accordingly.

image

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -2117,7 +2117,14 @@ export async function deleteExpense(req: express.Request, expenseId: number): Pr
   return expense.reload({ paranoid: false });
 }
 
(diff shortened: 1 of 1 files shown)


znarf (approved):
Thank you!

PR #9802, opened 2024-02-09
Fix isUnhosted argument query

kewitz:


znarf (approved):
Classic SQL mistake. Looks good!

kewitz:
@znarf and it is covering an edge case that is not even that critical. This smells like a last-minute update before merging. 😥

PR #9810, opened 2024-02-12
Webhooks: record last success date

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7280

Recording only the date (without time) to prevent flooding the DB.


znarf (approved):
Looks good to me

PR #9824, opened 2024-02-15
Optimize ExpensePipelineOverview query

znarf:
- add an index to fetch all approved collectives on a given host
- add an index to fetch expenses per collective and status
- add 2 indexes to make security checks faster
- make the expenses resolver async (nodes and totalCount fully on demand)

### DIFF
--- a/migrations/20240215073437-expense-pipeline-index.js
+++ b/migrations/20240215073437-expense-pipeline-index.js
@@ -35,7 +35,7 @@ module.exports = {
     `);
 
     await queryInterface.sequelize.query(`
-      DROP INDEX IF EXISTS "connected_accounts__username"
+      DROP INDEX IF EXISTS "connected_accounts__service_username"
     `);
 
     await queryInterface.sequelize.query(`
--- a/migrations/20240215073437-expense-pipeline-index.js
+++ b/migrations/20240215073437-expense-pipeline-index.js
@@ -16,8 +16,8 @@ module.exports = {
 
     // Related to security check and findRelatedUsersByConnectedAccounts
     await queryInterface.sequelize.query(`
-      CREATE INDEX CONCURRENTLY IF NOT EXISTS "connected_accounts__username"
-      ON "ConnectedAccounts"("username")
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "connected_accounts__service_username"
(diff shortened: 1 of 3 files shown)


migrations/20240215073437-expense-pipeline-index.js
```diff
@@ -0,0 +1,49 @@
...
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "collectives__host_collective_id_approved"
+      ON "Collectives"("HostCollectiveId")
+      WHERE "HostCollectiveId" IS NOT NULL AND "deletedAt" IS NULL AND "approvedAt" IS NOT NULL
+    `);
+
+    // Related to security check and findRelatedUsersByConnectedAccounts
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "connected_accounts__username"
+      ON "ConnectedAccounts"("username")
+      WHERE "deletedAt" IS NULL
```
Betree, line 21: Shouldn't we couple username+service? It seems wrong to fetch an account by username without specifying the service.

Betree (commented):
Related: https://github.com/opencollective/opencollective/issues/6107

Good stuff, maybe @kewitz should have a look as well.

migrations/20240215073437-expense-pipeline-index.js
znarf: We can

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -374,13 +375,35 @@ export const ExpensesCollectionQueryResolver = async (
...
+      }
+    }
+    return Expense.count({ include, where });
+  };
+
   return {
-    nodes: result.rows,
-    totalCount: result.count,
+    nodes: fetchNodes,
+    totalCount: fetchTotalCount,
```
Betree, line 404: Can we maybe have a shared helper for that? It sounds like we'd want the same approach in different places, and it would help to contain the complexity of this resolver. Something like:

```ts
const { nodes, totalCount } = await paginateResults(Expense, { include, where, order, offset, limit });
```

znarf: I removed this part.

PR #9825, opened 2024-02-15
Add config to opt-in `collective.transaction.created` activity

Betree:
For https://github.com/opencollective/opencollective/issues/7162

This should also improve test reliability, as asynchronously created activities mess up with `resetTestDB` (deadlocks).

To get the notifications list:

```sql
SELECT array_agg(DISTINCT "CollectiveId") as "CollectiveId"
FROM "Notifications"
WHERE type = 'collective.transaction.created'
  AND active IS TRUE
```


znarf (approved):
Nice!

PR #9829, opened 2024-02-15
Implement Transaction.clearedAt

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7210


znarf:
Some questions:

- Is it defaulting to `null` or to `createdAt`?
- How does a `clearedAt` propagates within the `TransactionGroup` ?

kewitz:
@znarf 
1. Null. Since these dates don't mean the same thing, I rather be clear about not having the information and forcing the user to coalesce than to suggest we know when it was cleared.
2. Same mechanics as `createdAt`. `createHostFeeShareTransactions`, `createTaxTransactions`, `createPaymentProcessorFeeTransactions` will copy the value from the originating transaction.

znarf:
> Null. Since these dates don't mean the same thing, I rather be clear about not having the information and forcing the user to coalesce than to suggest we know when it was cleared.

Not sure this is the expected behavior, can you inquire with product?

migrations/20240212114556-add-transaction-cleared-at.js
```diff
@@ -0,0 +1,27 @@
...
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "transactions__collective_clearedAt"
+      ON "Transactions"("CollectiveId", "clearedAt" DESC)
+      WHERE "deletedAt" IS NULL
+    `);
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "transactions__hostCollective_clearedAt"
+      ON "Transactions"("HostCollectiveId", "clearedAt" DESC)
+      WHERE "deletedAt" IS NULL
+    `);
```
znarf, line 17: Duplicate

server/paymentProviders/paypal/payment.ts
```diff
@@ -78,10 +80,6 @@ const recordTransaction = async (
     },
   };
 
-  if (createdAt) {
-    transactionData['createdAt'] = createdAt;
-  }
-
```
Betree, line 84: Removing this introduced seems to have introduced a breaking change: https://github.com/opencollective/opencollective/issues/7296

PR #9835, opened 2024-02-16
More robust isUnhosted query based on Members and HostApplications joins

kewitz:
This should return both the collectives that were unhosted as well as the ones that unhosted themselves.

### DIFF
--- a/server/graphql/v2/object/Host.js
+++ b/server/graphql/v2/object/Host.js
@@ -1202,6 +1202,7 @@ export const GraphQLHost = new GraphQLObjectType({
           },
         },
         async resolve(host, args) {
+          /** @type {Parameters<typeof models.Collective.findAndCountAll>[0]['where']} */
           const where = {
             HostCollectiveId: host.id,
             id: { [Op.not]: host.id },
@@ -1232,12 +1233,12 @@ export const GraphQLHost = new GraphQLObjectType({
           }
 
           if (args.isUnhosted) {
-            const unhostedActivities = await models.Activity.findAll({
-              where: { type: 'collective.unhosted', HostCollectiveId: host.id },
+            const collectiveIds = await models.HostApplication.findAll({
               attributes: ['CollectiveId'],
+              where: { HostCollectiveId: host.id, status: 'APPROVED' },
             });
             where.HostCollectiveId = { [Op.or]: [{ [Op.ne]: host.id }, { [Op.is]: null }] };
-            where.id = { [Op.in]: unhostedActivities.map(a => a.CollectiveId) };
+            where.id = collectiveIds.map(({ CollectiveId }) => CollectiveId);
           } else {
             where.isActive = true;
             where.approvedAt = args.isApproved ? { [Op.not]: null } : null;


server/graphql/v2/object/Host.js
```diff
@@ -1232,12 +1232,34 @@ export const GraphQLHost = new GraphQLObjectType({
...
+                  include: [
+                    {
+                      model: models.HostApplication,
+                      as: 'hostApplications',
+                      required: true,
+                      attributes: ['id'],
+                      where: {
+                        status: 'APPROVED',
+                        HostCollectiveId: host.id,
+                      },
```
Betree, line 1252: Note that this won't cover everything: when we introduced host applications in https://github.com/opencollective/opencollective-api/pull/5002 (3 years ago), we didn't create host applications for already hosted collectives. I do believe it's a good idea to look at this table, though; it's the right place to trust for this kind of info.

If we want it to be reliable, I would suggest a migration to insert entries for existing collectives. Something like:

```sql
INSERT INTO "HostApplications" ("CollectiveId", "HostCollectiveId", "status", "createdAt")
VALUES (collective.id, collective."HostCollectiveId", 'APPROVED', COALESCE(collective."approvedAt", collective."createdAt"))
FROM "Collectives" collective
WHERE collective."isActive" IS TRUE
AND collective."deletedAt" IS NULL
AND collective."HostCollectiveId" IS NOT NULL
AND collective.id != collective."HostCollectiveId"
```

server/graphql/v2/object/Host.js
```diff
@@ -1232,12 +1232,34 @@ export const GraphQLHost = new GraphQLObjectType({
           }
 
           if (args.isUnhosted) {
-            const unhostedActivities = await models.Activity.findAll({
-              where: { type: 'collective.unhosted', HostCollectiveId: host.id },
+            const formerHostedMembers = await models.Member.findAll({
+              where: { MemberCollectiveId: host.id, role: 'HOST', deletedAt: { [Op.ne]: null } },
```
Betree, line 1236: I wouldn't trust that: a collective could be unhosted, then rejoin the host.

Since the `HostApplication` is required, I don't think we need to look at members at all. And I don't think we need to do it in two queries either. I would suggest replacing this query by:
```ts
where.HostCollectiveId = { [Op.or]: [{ [Op.ne]: host.id }, { [Op.is]: null }] };
include.push([{
  association: 'hostApplications',
  attributes: [],
  required: true,
  where: { HostCollectiveId: host.id, status: 'APPROVED' }
}])
```

Then remove the `where.id`, and add the include on the main query.

server/graphql/v2/object/Host.js
kewitz: Hey @Betree, there's something intrinsically off with this logic since not every collective needs to go through the HostApplications path in order to get hosted, i.e. automated GitHub stars approval, collective created by a fiscal host admin, etc...

Perhaps going through the other way around? Deleted host members that do not have a denied HostAaplication?

Betree: True.

Looking at deleted members that do not have a denied HostAplication could work but make sure to only consider the last one (a collective could be rejected, then accepted).

An alternative would be to look at the `CollectiveHistories`; the query would be super simple and reliable, but incredibly slow. We could add an index to resolve that, but altogether I don't think that would be a good solution.

kewitz: > Looking at deleted members that do not have a denied HostAplication could work but make sure to only consider the last one (a collective could be rejected, then accepted).

We're also making sure HostCollectiveId does not match, which covers this edge case.

server/graphql/v2/object/Host.js
kewitz: LOL, checkout the query this generates:
```sql
SELECT
  "Collective".*
FROM (
  SELECT
    "Collective"."id",
    "Collective"."type",
    "Collective"."slug",
    "Collective"."name",
    "Collective"."legalName",
    "Collective"."company",
    "Collective"."CreatedByUserId",
    "Collective"."LastEditedByUserId",
    "Collective"."ParentCollectiveId",
    "Collective"."HostCollectiveId",
    "Collective"."hostFeePercent",
    "Collective"."platformFeePercent",
    "Collective"."description",
    "Collective"."longDescription",
    "Collective"."expensePolicy",
    "Collective"."contributionPolicy",
    "Collective"."currency",
    "Collective"."image",
    "Collective"."backgroundImage",
    "Collective"."countryISO",
    "Collective"."settings",
    "Collective"."data",
    "Collective"."startsAt",
    "Collective"."endsAt",
    "Collective"."timezone",
    "Collective"."createdAt",
    "Collective"."updatedAt",
    "Collective"."isActive",
    "Collective"."isIncognito",
    "Collective"."approvedAt",
    "Collective"."twitterHandle",
    "Collective"."repositoryUrl",
    "Collective"."website",
    "Collective"."tags",
    "Collective"."deactivatedAt",
    "Collective"."isHostAccount",
    "Collective"."plan",
    "Collective"."deletedAt"
  FROM
    "Collectives" AS "Collective"
  WHERE ("Collective"."deletedAt" IS NULL
    AND (("Collective"."HostCollectiveId" != 11004
        OR "Collective"."HostCollectiveId" IS NULL)
      AND "Collective"."id" != 11004
      AND "Collective"."type" IN ('COLLECTIVE', 'FUND')))
  AND (
    SELECT
      "CollectiveId"
    FROM
      "HostApplications" AS "hostApplications"
    WHERE (("hostApplications"."deletedAt" IS NULL
        AND ("hostApplications"."HostCollectiveId" = 11004
          AND "hostApplications"."status" = 'APPROVED'))
      AND "hostApplications"."CollectiveId" = "Collective"."id")
  LIMIT 1) IS NOT NULL
ORDER BY
  "Collective"."approvedAt" DESC
LIMIT 20 OFFSET 0) AS "Collective"
  INNER JOIN "HostApplications" AS "hostApplications" ON "Collective"."id" = "hostApplications"."CollectiveId"
    AND ("hostApplications"."deletedAt" IS NULL
      AND ("hostApplications"."HostCollectiveId" = 11004
        AND "hostApplications"."status" = 'APPROVED'))
ORDER BY
  "Collective"."approvedAt" DESC;
```

server/graphql/v2/object/Host.js
Betree: > > Looking at deleted members that do not have a denied HostAplication could work but make sure to only consider the last one (a collective could be rejected, then accepted).
> 
> We're also making sure HostCollectiveId does not match, which covers this edge case.

Not exactly. The case I'm referring to is:
1. Collective applies to host
2. Host rejects the application: "please include blablabla..."
3. Collective re-applies with the requested info
4. Application gets approved

There are now two applications in the DB for this collective: a rejected one, and an approved one.

kewitz: That's not a problem if the `Collective.HostCollectiveId` != host id. The purpose of the tab is displaying all collectives that were once hosted, not the ones that were unhosted only after having the latest application approved.

PR #9843, opened 2024-02-19
Warn about suspended asset check failures

kewitz:


server/models/SuspendedAsset.ts
```diff
@@ -33,7 +34,9 @@ class SuspendedAsset extends Model<InferAttributes<SuspendedAsset>, InferCreatio
   }): Promise<void> {
     const asset = await this.findOne({ where: { type, fingerprint } });
     if (asset) {
-      throw new Error(`Asset ${fingerprint} of type ${type} is suspended.`);
+      const error = Error(`Asset ${fingerprint} of type ${type} is suspended.`);
```
znarf, line 37:
```suggestion
      const error = new Error(`Asset ${fingerprint} of type ${type} is suspended.`);
```

kewitz: Thanks!

PR #9855, opened 2024-02-22
Invalidate all tokens when email changes

Betree:
Goes with https://github.com/opencollective/opencollective-frontend/pull/10379

Implemented by introducing a new `confirmEmail` mutation on GraphQL V2 (rather than using the V1 mutation) to prevent breaking changes when deploying. As a result, old frontends will be logged out after successfully changing the email.

### DIFF
--- a/server/graphql/v2/mutation/IndividualMutations.ts
+++ b/server/graphql/v2/mutation/IndividualMutations.ts
@@ -135,6 +135,15 @@ const individualMutations = {
         throw new Unauthorized('OAuth and Personal Tokens are not allowed for this route');
       }
 
+      // Rate limit (by IP, since we support anonymous requests)
+      const rateLimitKey = `individual_confirm_email_ip_${req.ip}`;
+      const rateLimitMax = config.limits.confirmEmailPerIpPerHour;
+      const rateLimit = new RateLimit(rateLimitKey, rateLimitMax, ONE_HOUR_IN_SECONDS);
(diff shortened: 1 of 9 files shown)


server/graphql/v2/mutation/IndividualMutations.ts
```diff
@@ -93,6 +94,49 @@ const individualMutations = {
...
+    },
+    resolve: async (_, { token: confirmEmailToken }, req) => {
+      const user = await confirmUserEmail(confirmEmailToken);
+      const individual = await user.getCollective({ loaders: req.loaders });
+
+      // The sign-in token
+      let token;
+
+      // We don't want OAuth tokens to be exchanged against a session token
+      if (req.userToken?.type !== 'OAUTH') {
```
Betree, line 131:
```suggestion
      if (req.userToken?.type !== 'OAUTH' && req.remoteUser) {
```

server/graphql/v2/mutation/IndividualMutations.ts
```diff
@@ -93,6 +94,49 @@ const individualMutations = {
         });
       }
 
+      return { individual, token };
+    },
+  },
+  confirmEmail: {
+    description: 'Confirm email for Individual. Scope: "account".',
```
znarf, line 101: Scope: "account" would imply it's oAuth but we're not checking scope. And do we really want oAuth tokens here?

znarf (approved):

PR #9863, opened 2024-02-23
Fix: clearedAt regression bugs

kewitz:


server/lib/payments.js
```diff
@@ -497,13 +497,13 @@ export async function createRefundTransaction(
   }
 
   // Refund Payment Processor Fee
-  await refundPaymentProcessorFee(transaction, user, refundedPaymentProcessorFee, transactionGroup, clearedAt);
+  await refundPaymentProcessorFee(transaction, user, refundedPaymentProcessorFee, transactionGroup, {}, clearedAt);
```
Betree, line 500: Typescript would have helped here

Betree (approved):

server/lib/payments.js
kewitz: Yes, @znarf is already refactoring this file and converting it to TS.

PR #9872, opened 2024-02-26
Expense buget stats to count distinct expense ids

kewitz:


znarf:
Nice. Alternative, kind = 'EXPENSE' on the Transactions join ?

kewitz:
@znarf I'm not sure, it depends if we want to display the net amount of the expense or include the payment processor fees. I'm going with the hypothesis that they want to see the total amount.

Also, I'm refactoring this two functions to use Sequelize instead a raw query, the PR is almost done.

znarf:
Sounds good.

kewitz:
Merging this to fix the bug. I will wait for a review on #9873.

PR #9873, opened 2024-02-27
Rewrite getCollectiveExpensesTags queries with sequelize

kewitz:
Refactor getCollectiveExpenseTags functions using Sequelize.
Fixes expense double count on the collective budget section. (Replaces https://github.com/opencollective/opencollective-api/pull/9872)


kewitz:
@znarf not sure if necessarily simpler than using a raw query, but I think it is at least more consistent.

PR #9883, opened 2024-02-29
Transferring orders when changing hosts

Betree:
Left to do:
- [x] Add logic for Stripe elements
- [x] Cancel all contributions (incl Stripe) if `pauseContributions` if false

As a follow-up:
- Daily CRON to send emails to resume contributions
  - We look at PAUSED contributions for Collectives who are financially active (have a host or independent)
  - Based on some parameters (last contact, number of attempts), we send an email prompting contributors to update their payment method or create a new contribution.
  - We do our best to default all relevant parameters (guest flow is ok)
  - We keep track of the original order id to cancel it when new contribution is created (replaceOrderId)

### DIFF
--- a/cron/hourly/70-handle-batch-subscriptions-update.ts
+++ b/cron/hourly/70-handle-batch-subscriptions-update.ts
@@ -2,7 +2,8 @@
 
 import '../../server/env';
 
-import { groupBy, size, uniq } from 'lodash';
+import { groupBy, size, sortBy, uniq } from 'lodash';
+import moment from 'moment';
 
 import { activities } from '../../server/constants';
 import FEATURE from '../../server/constants/feature';
@@ -14,7 +15,7 @@ import models, { Collective, Op } from '../../server/models';
 import { OrderModelInterface } from '../../server/models/Order';
 
 /**
- * Since the collective has been archived, its HostCollectiveId has been set to null.
+ * If the collective has been archived, its HostCollectiveId has been set to null.
  * We need some more logic to make sure we're loading the right host.
  */
 const getHostFromOrder = async order => {
@@ -37,12 +38,12 @@ const getOrderCancelationReason = (
   order: OrderModelInterface,
   orderHost: Collective,
 ): {
-  code: 'TRANSFER' | 'DELETED_TIER' | 'ARCHIVED_ACCOUNT' | 'UNHOSTED_COLLECTIVE' | 'CHANGED_HOST' | 'CANCELLED_ORDER';
+  code: 'PAUSED' | 'DELETED_TIER' | 'ARCHIVED_ACCOUNT' | 'UNHOSTED_COLLECTIVE' | 'CHANGED_HOST' | 'CANCELLED_ORDER';
   message: string;
 } => {
-  if (order.status === 'PAUSED' && order.data?.pausedForTransfer) {
+  if (order.status === 'PAUSED') {
     return {
-      code: 'TRANSFER',
+      code: 'PAUSED',
       message: `Your contribution to the Collective was paused. We'll inform you when it will be ready for re-activation.`,
     };
   } else if (order.TierId && !order.Tier) {
@@ -65,7 +66,13 @@ const getOrderCancelationReason = (
  */
 export async function run() {
   const orphanOrders = await models.Order.findAll<OrderModelInterface>({
-    where: { status: [OrderStatuses.CANCELLED, OrderStatuses.PAUSED] },
+    where: {
+      status: [OrderStatuses.CANCELLED, OrderStatuses.PAUSED],
+      data: { needsAsyncDeactivation: true },
+      updatedAt: {
+        [Op.gt]: moment().subtract(1, 'month').toDate(), // For performance, only look at orders updated recently
+      },
+    },
     include: [
       {
         model: models.Tier,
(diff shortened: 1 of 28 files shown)


server/lib/recurring-contributions.js
```diff
@@ -31,13 +31,22 @@ export const MAX_RETRIES = 6;
 export async function ordersWithPendingCharges({ limit, startDate } = {}) {
   return models.Order.findAndCountAll({
     where: {
+      status: { [Op.not]: status.PAUSED },
```
znarf, line 34: Surprise that we don't have already a condition there. Because we rely on Subscription.isActive? Should we tweak that one too?

Betree: Below, in the `include`, we have another condition that filters out inactive/external subscriptions: https://github.com/opencollective/opencollective-api/blob/fa67633a8a0789e39a418ae2106507ce26465d7d/server/lib/recurring-contributions.js#L56

server/models/Order.ts
```diff
@@ -667,30 +672,81 @@ Order.cancelActiveOrdersByTierId = function (tierId: number) {
...
+          WHEN spm.id IS NOT NULL THEN (
+            -- For gift cards, we need to check the source payment method
+            spm.service != 'stripe'
+            OR spm.type != 'creditcard'
+            OR COALESCE(spm.data#>>'{stripeAccount}', :platformStripeAccountId) != :platformStripeAccountId
+          ) WHEN pm.id IS NOT NULL THEN (
+            -- Only legacy stripe credit card contributions are transferable
+            pm.service != 'stripe'
+            OR pm.type != 'creditcard'
+            OR COALESCE(pm.data#>>'{stripeAccount}', :platformStripeAccountId) != :platformStripeAccountId
```
Betree, line 712: Or internal from collective <> children?

templates/emails/subscription.paused.hbs
```diff
@@ -0,0 +1,24 @@
...
+
+<p>Your recurring contribution to {{collective.name}} for {{currency subscription.amount currency=subscription.currency}}/{{subscription.interval}} has been paused.</p>
+
+{{#if reason}}
+  <p>The reason given by the collective is:</p>
+  <blockquote style="color: #6a737d;font-size: 12px; font-style: italic; text-align: left;padding: 0.5em 0.75em;margin: 1em 0;border-left: 3px solid #e4e4e4;white-space: pre-line;">{{reason}}</blockquote>
+{{/if}}
+
+<p>
+  We'll inform you when your contribution is ready for reactivation. Without any further action from you, your contribution will remained paused.
```
Betree, line 15:
```suggestion
  We'll let you know when your contribution is ready for reactivation. Without any further action from you, your contribution will remain paused.
```

cron/hourly/70-handle-batch-subscriptions-update.ts
```diff
@@ -102,13 +122,19 @@ export async function run() {
...
               collective: order.collective.minimal,
               fromCollective: order.fromCollective.minimal,
-              reasonCode: reasonCode,
-              reason: reason,
+              reasonCode: reason.code,
+              reason: reason.message,
+              messageForContributors: order.data?.messageForContributors,
               order: order.info,
               tier: order.Tier?.info,
+              awaitForDispatch: true, // To make sure we won't kill the process while emails are still being sent
```
Betree, line 130: Added this flag to make sure we await the email dispatch in `Activities.afterCreate`, without it the script could exit before all the emails have been sent.

server/graphql/v2/object/AmountStats.js
```diff
@@ -14,7 +15,7 @@ export const GraphQLAmountStats = new GraphQLObjectType({
       type: new GraphQLNonNull(GraphQLAmount),
       description: 'Total amount for this label',
       resolve(entry) {
-        if (entry.amount) {
+        if (!isNil(entry.amount)) {
```
Betree, line 18: Returning a `0` amount was not supported

server/lib/recurring-contributions.js
```diff
@@ -31,13 +31,22 @@ export const MAX_RETRIES = 6;
...
-      { model: models.Collective, as: 'collective', required: true },
+      {
+        model: models.Collective,
+        as: 'collective',
+        required: true,
+        where: {
+          HostCollectiveId: { [Op.not]: null },
+          isActive: true,
+        },
+      },
```
Betree, line 49: This part is to ignore unhosted collectives completely

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -357,6 +357,8 @@ const orderMutations = {
         throw new Unauthorized("You don't have permission to update this order");
       } else if (!order.Subscription.isActive) {
         throw new Error('Order must be active to be updated');
+      } else if (order.status === OrderStatuses.PAUSED) {
+        throw new Error('Paused orders cannot be updated');
```
znarf, line 361: Maybe we'll need to revisit this once we're ready? Updating a payment method should be ok?

server/graphql/v2/object/AccountStats.js
```diff
@@ -376,17 +380,60 @@ export const GraphQLAccountStats = new GraphQLObjectType({
...
+        type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(GraphQLAmountStats))),
+        args: {
+          frequency: {
+            type: GraphQLContributionFrequency,
+            description: 'Return only the stats for this frequency',
+          },
+          hasPortability: {
+            type: GraphQLBoolean,
+            description: 'Filter contributions on whether they can be ported to another fiscal host directly',
+          },
```
znarf, line 404: Are we gonna use that concept in the end?

znarf (approved):
Looks good to me

server/graphql/v2/mutation/OrderMutations.js
Betree: Absolutely. Here I wanted to prevent people re-activating their orders while the collective is still under OCF, until we build the proper solution for it.

server/graphql/v2/object/AccountStats.js
Betree: Not currently. If we're settling with the fact that recurring contributions can never be transferred directly, we should remove it. I'm adding a reminder to myself to go back and remove this, depending on how things evolved with the current situation.

PR #9902, opened 2024-03-04
Fix: canRefund to check already refunded Orders

kewitz:


znarf (changes requested):
If moving forward with this approach, why not go extra mile and check for a bunch of status (REFUNDED, REJECTED, more?) directly in canRefund?

kewitz:
@znarf that's a good idea. The original intent was to just move frontend checks to the API.

PR #9914, opened 2024-03-07
feat: Duplicate account

Betree:


server/graphql/v2/interface/Account.ts
```diff
@@ -658,6 +659,9 @@ const accountFieldsDefinition = () => ({
       const where = {
         ParentCollectiveId: account.id,
       };
+      if (!isNil(args.isActive)) {
+        where['isActive'] = args.isActive;
+      }
```
Betree, line 664: This is to get the count for **active** events/projects in the frontend

PR #9923, opened 2024-03-11
OrdersCollectionQuery: dynamically fetch nodes/count

Betree:
This should prevent at least 4 expensive SQL queries when opening "Dashboard > Contributions", were we're fetching multiple counts based on the status: https://github.com/opencollective/opencollective-frontend/blob/d5771f3cb5028cfc9b79f1740bcdba0076c32d81/components/dashboard/sections/contributions/Contributions.tsx#L64-L75


znarf (approved):
Looks good!

PR #9925, opened 2024-03-11
Fix: Search query and vendor results

kewitz:


Betree (approved):
As far as I understand it's not really a bug: the results were returned, but not necessarily first. A document with multiple term occurrences would be ranked higher than the exact match.

Example:
- Account 1
   - Name: OpenSource
   - Description: We do stuff.
- Account 2
   - Name: Test
   - Description: We love opensource and do only opensource.

With the current strategy Account 2 would be returned 1st because it has 2 occurrences for opensource.

With the new strategy, we're normalizing by ["dividing the rank by 1 + the logarithm of the document length"](https://www.postgresql.org/docs/current/textsearch-controls.html), thus preferring shorter results. This should be a good improvement for account names and slugs.

PR #9926, opened 2024-03-11
Resume paused contributions

Betree:
- [x] Support canceling paused contributions
- [x] Support updating the payment method of a paused contribution
- [x] Add the message from collective admins on the resume email
- [x] Resume email
- [x] Resume email reminder 1,2,3


znarf (approved):
Looks good.

To be discussed: how much is this an OCF only process VS something that van be used for every collective switching between fiscal hosts.

PR #9931, opened 2024-03-12
Do not cancel orders when adding host

Betree:


server/models/Collective.ts
```diff
@@ -2407,13 +2427,16 @@ class Collective extends Model<
       });
     }
 
-    // Pause or cancel all orders that cannot be transferred
-    const newOrderStatus = options?.pauseContributions ? OrderStatuses.PAUSED : OrderStatuses.CANCELLED;
-    await Order.stopActiveSubscriptions(this.id, newOrderStatus, options);
+    // Deactivate/remote everything related to previous host
+    if (this.HostCollectiveId) {
```
Betree, line 2431: This is the actual fix, the rest is for better typings of the arguments

PR #9988, opened 2024-03-25
Internalize tax forms

Betree:
Part of https://github.com/opencollective/opencollective/issues/7300
Require https://github.com/opencollective/opencollective-pdf/pull/985

### DIFF
--- a/server/graphql/v2/mutation/LegalDocumentsMutations.ts
+++ b/server/graphql/v2/mutation/LegalDocumentsMutations.ts
@@ -11,7 +11,12 @@ import { getUSTaxFormPdf } from '../../../lib/pdf';
 import { reportErrorToSentry } from '../../../lib/sentry';
 import { encryptAndUploadTaxFormToS3 } from '../../../lib/tax-forms';
 import { Activity, LegalDocument } from '../../../models';
-import { LEGAL_DOCUMENT_REQUEST_STATUS, LEGAL_DOCUMENT_SERVICE } from '../../../models/LegalDocument';
+import {
+  LEGAL_DOCUMENT_REQUEST_STATUS,
+  LEGAL_DOCUMENT_SERVICE,
+  LEGAL_DOCUMENT_TYPE,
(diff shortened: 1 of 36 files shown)


server/graphql/v2/mutation/LegalDocumentsMutations.ts
```diff
@@ -0,0 +1,124 @@
...
+        throw new Forbidden('You need to be logged in to submit a legal document');
+      }
+
+      checkRemoteUserCanUseAccount(req);
+      const account = await fetchAccountWithReference(args.account, { throwIfMissing: true });
+      if (!req.remoteUser.isAdminOfCollective(account)) {
+        throw new Forbidden('You do not have permission to submit a legal document for this account');
+      }
+
+      if (args.type === 'US_TAX_FORM') {
```
znarf, line 53: Style: I prefer to use the pattern were we test the opposite and exit early.

```
if (args.type !== 'US_TAX_FORM') {
  throw new ValidationFailed(`Legal document type ${args.type} is not supported`);
}
```

I understand that the code as it is may be more future proof if we were to add more types.

znarf, line 53: Constant perhaps `LEGAL_DOCUMENT_TYPE.US_TAX_FORM`.

But we said this morning that it's not really necessary when using Typescript.

server/graphql/v2/mutation/LegalDocumentsMutations.ts
```diff
@@ -0,0 +1,124 @@
...
+        // Some validation on form data
+        const formType = args.formData.formType;
+        assert(formType, 'Form type is required');
+        assert(['W9', 'W8_BEN', 'W8_BEN_E'].includes(formType), 'Invalid form type');
```
znarf, line 57: Could those allowed values be read from somewhere else? From the types?

znarf (approved):
Looks good to me!

Some minor feedback to prove that I looked at it ;-)

PR #9995, opened 2024-03-28
Make v2 search resolver consistent with v1

kewitz:


server/graphql/v2/query/collection/AccountsCollectionQuery.ts
```diff
@@ -30,6 +30,11 @@ export const CommonAccountsCollectionQueryArgs = {
     type: GraphQLBoolean,
     description: 'Included collectives which are archived',
   },
+  skipGuests: {
+    type: GraphQLBoolean,
+    description: 'Included collectives which are archived',
```
Betree, line 35: Wrong description

server/graphql/v2/query/collection/AccountsCollectionQuery.ts
```diff
@@ -54,6 +59,10 @@ const AccountsCollectionQuery = {
       type: new GraphQLList(GraphQLAccountReferenceInput),
       description: 'Host hosting the account',
     },
+    parent: {
+      type: new GraphQLList(GraphQLAccountReferenceInput),
+      description: 'Host hosting the account',
```
Betree, line 64: Wrong description

Betree (commented):
Good changes!

PR #10000, opened 2024-04-03
Improve some URLs validations

Betree:


Betree:
@kewitz I just realized the PR number :smile:

PR #10006, opened 2024-04-04
Do not report webhook 404 errors to Sentry

kewitz:


server/lib/notifications/index.ts
```diff
@@ -93,11 +97,15 @@ const dispatch = async (
...
-          });
+          if (WEBHOOK_IGNORED_RESPONSE_CODES.includes((e as AxiosError).response?.status)) {
+            logger.warn(`Webhook #${notifConfig.id} to ${notifConfig.webhookUrl} returned ${e.response.status}`);
+          } else {
+            const stringifiedError =
+              e instanceof AxiosError ? `${e.response?.status} ${e.response?.statusText} ${e.config?.url}` : e;
+            reportErrorToSentry(e, {
+              extra: { activity, notifConfig, onlyChannels, force, stringifiedError },
+            });
+          }
```
Betree, line 108: I'm not convinced that we need to treat them differently based on the status code. Why would we report a 500 or 410 to Sentry, but not a 403?

Altogether, I'd say:
- Reporting them to Sentry gives us a metric of "how many webhook" requests fail. I personally think that's helpful, but if we want to get rid of it then let's remove it for all cases.
- `logger.warn` are just going to pollute the logs, without adding much value.
- If we want to act on failures, we should start recording them, e.g. in `notification.data`

PR #10008, opened 2024-04-05
Immediately create order members after transactions are created

kewitz:
Relates to https://github.com/opencollective/opencollective/issues/7075

Moving above `await sendThankYouEmail(order, transaction);` should solve the issue and should also explain why it always happens with PayPal and not with Stripe.

Either way, I'm executing the side effects in parallel immediately after transactions are created and making it consistent across PayPal and Stripe.

### DIFF
--- a/server/paymentProviders/stripe/webhook.ts
+++ b/server/paymentProviders/stripe/webhook.ts
@@ -194,27 +194,26 @@ export const paymentIntentSucceeded = async (event: Stripe.Event) => {
   await createOrUpdateOrderStripePaymentMethod(order, stripeAccount, paymentIntent);
 
   const transaction = await createChargeTransactions(charge, { order });
+  const sideEffects: Promise<unknown>[] = [
+    order.update({
+      status: !order.SubscriptionId ? OrderStatuses.PAID : OrderStatuses.ACTIVE,
+      processedAt: new Date(),
+      data: {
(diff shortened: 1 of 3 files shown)


server/paymentProviders/stripe/webhook.ts
```diff
@@ -194,26 +194,24 @@ export const paymentIntentSucceeded = async (event: Stripe.Event) => {
...
+  await Promise.all([
+    order.update({
+      status: !order.SubscriptionId ? OrderStatuses.PAID : OrderStatuses.ACTIVE,
+      processedAt: new Date(),
+      data: {
+        ...omit(order.data, 'paymentIntent'),
+        previousPaymentIntents: [...(order.data.previousPaymentIntents ?? []), paymentIntent],
+      },
+    }),
+    order.getOrCreateMembers(),
```
Betree, line 206: One thing I don't like about this new approach is that any failure of `getOrCreateMembers` will prevent the creation of the subscription. Some error handling on this one could be nice.

Betree (commented):
This can help with https://github.com/opencollective/opencollective/issues/7075, but we can't guarantee that it will 100% solve it.

Betree:
> and should also explain why it always happens with PayPal and not with Stripe.

This was explained in https://github.com/opencollective/opencollective/issues/7075#issuecomment-1798006750

PR #10012, opened 2024-04-08
Mark paused subscriptions as cancelled when creating a new one

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7341

### DIFF
--- a/server/models/Order.ts
+++ b/server/models/Order.ts
@@ -4,6 +4,7 @@ import { get } from 'lodash';
 import {
   ForeignKey,
   HasManyGetAssociationsMixin,
+  HasOneCreateAssociationMixin,
   HasOneGetAssociationMixin,
   InferAttributes,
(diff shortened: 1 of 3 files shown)


server/models/Order.ts
```diff
@@ -647,6 +650,23 @@ Order.prototype.unPause = async function (user: User, { UserTokenId = undefined
...
+    { status: OrderStatus.CANCELLED },
+    {
+      where: {
+        id: { [Op.not]: this.id },
+        FromCollectiveId: this.FromCollectiveId,
+        CollectiveId: this.CollectiveId,
+        status: OrderStatus.PAUSED,
+      },
+    },
+  );
```
znarf, line 667: Looks good. So, no need to do anything else than updating the model?

Betree: Nope, the subscription itself has already been deactivated.

PR #10031, opened 2024-04-15
Add check to have ensure isActive=true on hosts

znarf:
Budget should normally be activated for all new hosts since October 2020.

https://github.com/opencollective/opencollective-api/pull/4752

### DIFF
--- a/checks/model/hosts.js
+++ b/checks/model/hosts.js
@@ -109,12 +109,13 @@ async function checkHostActive({ fix = false } = {}) {
   const message = 'Host with isActive=false';
 
   const results = await sequelize.query(
-    `SELECT *
+    `SELECT COUNT(*) as count
      FROM "Collectives"
      WHERE "isHostAccount" IS TRUE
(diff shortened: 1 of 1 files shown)


checks/model/hosts.js
```diff
@@ -104,10 +104,45 @@ async function checkHostMemberEntry({ fix = false } = {}) {
...
+  const message = 'Host with isActive=false';
+
+  const results = await sequelize.query(
+    `SELECT *
+     FROM "Collectives"
+     WHERE "isHostAccount" IS TRUE
+     AND "deletedAt" IS NULL AND "deactivatedAt" IS NULL
+     AND "approvedAt" IS NOT NULL AND "HostCollectiveId" = "id"
+     AND "isActive" IS FALSE
+     AND COALESCE(("settings"->'forceInactive')::boolean), false) IS FALSE`,
```
Betree, line 118: To make it a bit more explicit

```suggestion
     AND COALESCE(("settings"->'forceInactiveHost')::boolean), false) IS FALSE`,
```

Betree (approved):

PR #10037, opened 2024-04-16
Add missing Transaction merchantId indexes

kewitz:


znarf:
Maybe it's time to create a dedicated column for that.

gustavlrsn (approved):

kewitz:
Tracking that as tech-debt in https://github.com/opencollective/opencollective/issues/7370

PR #10041, opened 2024-04-16
Add EFFECTIVE_DATE as an orderBy field

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7355


gustavlrsn (commented):
The custom ordering function should be is activated in prod (`config.ledger.orderedTransactions`), which will bypass using the OrderBy.field:
https://github.com/opencollective/opencollective-api/blob/main/server/graphql/v2/query/collection/TransactionsCollectionQuery.ts#L476-L486

I guess we want to keep the custom order but base it on the supplied `orderBy.field`

gustavlrsn (approved):

PR #10048, opened 2024-04-17
Add link to payment processor dashboard

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7345

Provides a link to the payment processor dashboard (Stripe) if available.


server/paymentProviders/stripe/webhook.ts
```diff
@@ -330,6 +352,29 @@ export const chargeDisputeCreated = async (event: Stripe.Event) => {
     return;
   }
 
+  console.log(chargeTransaction.OrderId, chargeTransaction?.Order?.id);
```
kewitz, line 355:
```suggestion
```

server/paymentProviders/stripe/webhook.ts
```diff
@@ -383,6 +437,28 @@ export const chargeDisputeClosed = async (event: Stripe.Event) => {
     return;
   }
 
+  if (chargeTransaction.Order) {
```
kewitz, line 440: Side effects should come after the actions, there are important checks and early-returns coming after this conditional.

server/paymentProviders/stripe/webhook.ts
```diff
@@ -591,6 +712,31 @@ export const reviewClosed = async (event: Stripe.Event) => {
...
+        toCollective: order.collective.info,
+        tierName: order.Tier?.name,
+        reason: review.closed_reason,
+        paymentProcessorUrl: getDashboardObjectIdURL(
+          typeof review.payment_intent === 'string' ? review.payment_intent : review.payment_intent.id,
+          event.account,
+        ),
+      },
+    });
+  }
```
kewitz, line 738: Should come last, after transactions are updated.

migrations/20240505060955-add-index-for-order-pending-or-manual-filters.js
```diff
@@ -0,0 +1,24 @@
...
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "orders_isPendingContribution" ON "Orders" (((data #>> '{isPendingContribution}')::text)) where ((data #>> '{isPendingContribution}'::text[]) = 'true'::text);
+    `);
+
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "orders_isManualContribution" ON "Orders" (((data #>> '{isManualContribution}')::text)) where ((data #>> '{isManualContribution}'::text[]) = 'true'::text);
+    `);
```
kewitz, line 12: Not sure if we need those casts since the double `>>` already casts to text. Also, I think the second cast is wrong, it is not an array of texts.

If you need an example of how to index data fields:
https://github.com/opencollective/opencollective-api/pull/10037/files#diff-3af06b8af92da31930b6a5d8634d7dcca72dfe3d1703564986697d827cc82d78R8-R12

server/paymentProviders/stripe/webhook.ts
```diff
@@ -280,6 +286,8 @@ export const paymentIntentProcessing = async (event: Stripe.Event) => {
       },
       { transaction },
     );
+
+    sendEmailNotifications(order);
```
kewitz, line 290: What is this about?

kewitz (approved):

migrations/20240505060955-add-index-for-order-pending-or-manual-filters.js
hdiniz: I will double check the indexes with the example you provided. The `::text[]` cast refers to the argument of `#>>`, but its probably unnecessary, I just copied the database format after creating the index manually.

server/paymentProviders/stripe/webhook.ts
hdiniz: The processing webhook was not sending a notification or creating an activity for the processing status.

PR #10054, opened 2024-04-19
Add flag to release internal tax forms

Betree:
API part for https://github.com/opencollective/opencollective/issues/7362

This PR adds a `TAX_FORMS_USE_INTERNAL` environment variable (default=true). When enabled, it will:
1. Hook into the `createExpense` and `editExpense` mutations to create the request right away (rather than async)
2. Prevent sending new Dropbox Form requests from  `cron/hourly/40-send-tax-form-requests.js` CRON job
3. Add a reminder email after two days in case the person did not complete the tax form

Notes:
- Dropbox Form requests already sent will stay valid, and we will continue to process them in the webhook (will be removed in https://github.com/opencollective/opencollective/issues/7378).
- This must be released before the corresponding frontend PR.


server/lib/sequelize.js
```diff
@@ -47,10 +47,10 @@ if (config.database.options.logging) {
 }
 
 if (config.database.options.pool) {
-  if (config.database.options.pool.min) {
+  if (config.database.options.pool.min && typeof config.database.options.pool.min === 'string') {
     config.database.options.pool.min = parseInt(config.database.options.pool.min, 10);
   }
-  if (config.database.options.pool.max) {
```
Betree, line 53: I'm not sure why now, but this was making the test watchers fail with:

```
Error: Can not update runtime configuration property: "min". Configuration objects are immutable unless ALLOW_CONFIG_MUTATIONS is set.
```

znarf:
🚀

PR #10058, opened 2024-04-22
Add coverage for tax forms CRON

Betree:


Betree:
Somehow, this approach of directly testing the CRON job caused heavy issues related to the configuration import that were difficult to debug. Ended up moving code in the model in https://github.com/opencollective/opencollective-api/pull/10054.

PR #10062, opened 2024-04-23
TaxForms: update for Vendors

Betree:


server/models/LegalDocument.ts
```diff
@@ -144,6 +145,61 @@ class LegalDocument extends Model<InferAttributes<LegalDocument>, InferCreationA
...
+        type: activities.TAXFORM_RECEIVED,
+        UserId: user.id,
+        CollectiveId: payee.id,
+        FromCollectiveId: user.CollectiveId,
+        UserTokenId: UserTokenId,
+        data: {
+          service: legalDocument.service,
+          document: legalDocument.info,
+          account: payee.info,
+          isManual: true,
```
kewitz, line 197: I think you can skip that commend with a `notify: false` flag.
```suggestion
          isManual: true,
          notify: false,
```

kewitz (approved):

server/models/LegalDocument.ts
Betree: This activity is never sent directly in `server/lib/notifications/email.ts`, only manually. The `isManual` flag is only there for us to easily differentiate when a tax form is submitted manually (for analytics)

PR #10070, opened 2024-04-29
Record host applications in auto-approve flow

znarf:


server/models/HostApplication.ts
```diff
@@ -46,17 +46,18 @@ class HostApplication extends Model<InferAttributes<HostApplication>, InferCreat
     host: Collective,
     collective: Collective,
     user: User,
+    status: HostApplicationStatus,
     data: Record<string, unknown>,
   ): Promise<HostApplication> {
-    const existingApplication = await this.getByStatus(host, collective, HostApplicationStatus.PENDING);
+    const existingApplication = await this.getByStatus(host, collective, status);
```
Betree, line 52: On this line, I think we could try to cover an additional use case. If Babel manually applies to OSC (creates pending application) and then applies through the GitHub flow, we should ideally mark the previous, pending application as "Approved" (rather than re-recreating a new one).

server/graphql/v2/object/HostApplication.ts
```diff
@@ -25,18 +28,51 @@ const GraphQLHostApplication = new GraphQLObjectType({
...
+      async resolve(application, _, req) {
+        // Unfiltered
+        if (
+          req.remoteUser?.isAdmin(application.HostCollectiveId) ||
+          req.remoteUser?.isAdmin(application.CollectiveId)
+        ) {
+          return application.customData;
+        }
+        // Allow-list to support the OSC / GitHub case
+        return pick(application.message, ['repositoryUrl', 'validatedRepositoryInfo']);
```
Betree, line 74: Looks like a typo

```suggestion
        return pick(application.customData, ['repositoryUrl', 'validatedRepositoryInfo']);
```

PR #10072, opened 2024-04-29
enhancement(TaxForms): allow download from expense page

Betree:
For https://github.com/opencollective/opencollective/issues/7392


znarf (approved):
Looks good

PR #10073, opened 2024-04-29
Prevent removing payment method from processing orders and return payment method info even if deleted

hdiniz:
An order paid with a delayed payment method (SEPA, ACH, etc)  can be processing, we prevent the payment method to be deleted in that case.

If a payment method is deleted, its info should not be missing from the order list.

### DIFF
--- a/server/graphql/loaders/index.js
+++ b/server/graphql/loaders/index.js
@@ -710,6 +710,15 @@ export const loaders = req => {
     }).then(results => sortResults(CollectiveIds, results, 'CollectiveId', [])),
   );
 
+  context.loaders.PaymentMethod.byIdIncludeDeleted = new DataLoader(keys => {
+    return models.PaymentMethod.findAll({
+      where: {
+        id: keys,
+      },
+      paranoid: false,
+    }).then(results => sortResultsSimple(keys, results, pm => pm.id, []));
+  });
+
   /** *** Order *****/
   // Order - findByMembership
   context.loaders.Order.findByMembership = new DataLoader(combinedKeys =>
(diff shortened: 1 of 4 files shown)


server/graphql/v2/object/Order.js
```diff
@@ -202,7 +202,7 @@ export const GraphQLOrder = new GraphQLObjectType({
         type: GraphQLPaymentMethod,
         resolve(order, _, req) {
           if (order.PaymentMethodId) {
-            return req.loaders.PaymentMethod.byId.load(order.PaymentMethodId);
+            return req.loaders.PaymentMethod.byIdIncludeDeleted.load(order.PaymentMethodId);
```
kewitz, line 205: Do we really need to expose deleted assets? What is the use case?

server/graphql/loaders/index.js
```diff
@@ -710,6 +710,15 @@ export const loaders = req => {
...
   );
 
+  context.loaders.PaymentMethod.byIdIncludeDeleted = new DataLoader(keys => {
+    return models.PaymentMethod.findAll({
+      where: {
+        id: keys,
+      },
+      paranoid: false,
+    }).then(results => sortResultsSimple(keys, results, pm => pm.id, []));
+  });
```
kewitz, line 720: Quick reminder: this is only safe because we use the primary key ID in this query.
Hypothetically, we should not implement stuff like `byCollectiveIdIncludedDeleted` because our custom indexes exclude deleted records.

server/graphql/v2/object/Order.js
hdiniz: When looking at past contributions / transactions. If the user removes the payment method, we still want to show info about it in the contribution list/drawers.

PR #10066, opened 2024-04-25
Convert Order model to expected typescript

znarf:
Took some stuff from a previous attempt https://github.com/opencollective/opencollective-api/pull/9753

Much more focused.

PR #10081, opened 2024-05-03
Convert Transactions model to typescript class

znarf:
Similar to https://github.com/opencollective/opencollective-api/pull/10066

MemberInvitation.ts, PaymentMethod.ts and Subscription.ts left todo


Betree (commented):
A bit difficult to review the diff but the pattern looks ok to me

PR #10090, opened 2024-05-15
Deprecate account.hasImage resolver

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7407

Requires https://github.com/opencollective/opencollective-images/pull/611


znarf (approved):
To be deployed AFTER frontend.

PR #10108, opened 2024-05-30
Update host-settlement minimum invoice amount pinned to USD

kewitz:
I notice we invoice a great amount of settlements that are either invalid (no one is settling $2,95) or ignored. Considering that invoiced transactions are no longer considered in the next invoice (they're no longer debts) I decided to:
- [x] Updates the host-settlement cron job to enforce a minimum invoiceable amount in USD
- [x] Set minimum invoiceable amount through environment variable


znarf (commented):
Let's keep $10, that value was not an issue?

znarf (approved):
Good for me.

PR #10117, opened 2024-06-04
enhancement(Expense): tolerate small diffs in exchange rate

Betree:
Due to fluctuation, there can be small differences between the rate we get for `latest` and the one stored in the cache. Tolerating a small difference should resolve this.


znarf:
I remember we already had something like that. In a different place ?

Betree:
@znarf Just a few lines below, for user-submitted FX rates: https://github.com/opencollective/opencollective-api/blob/7d22a41c810cdab242234bd7131302f355c452b7/server/graphql/common/expenses.ts#L1402.

The situation is a bit more tricky: in Redis, we have a very fresh FX rate that does not match the latest returned to the user for "today". I'm gonna review the logic more deeply, but this simple fix  should unblock most cases.

PR #10124, opened 2024-06-06
fix(Orders): next charge date when unpausing contributions

Betree:


server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -439,7 +439,7 @@ const orderMutations = {
 
         // Unpause contribution
         if (order.status === OrderStatuses.PAUSED) {
-          await order.unpause(req.remoteUser, { UserTokenId: req.userToken?.id });
+          await order.unPause(req.remoteUser, { UserTokenId: req.userToken?.id });
```
Betree, line 442: There was a typo here causing "SUBSCRIPTION_RESUMED" activities to not be created

server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -437,9 +438,13 @@ const orderMutations = {
           order = await updatePaymentMethodForSubscription(req.remoteUser, order, newPaymentMethod);
         }
 
-        // Unpause contribution
-        if (order.status === OrderStatuses.PAUSED) {
```
Betree, line 441: For Stripe, this code was not reached because at this point the order was marked as active already by https://github.com/opencollective/opencollective-api/blob/cbfc984efc008c2f36eb903b4c2d5dfec236cfd9/server/lib/subscriptions.ts#L85

server/lib/subscriptions.ts
```diff
@@ -21,12 +21,14 @@ const getIsSubscriptionManagedExternally = pm => {
...
+ *
+ * If reactivating the contribution on the same interval (month/year), keep the previous "next charge date". Otherwise,
+ * set the next charge date to the beginning of the next interval.
  */
-const getNextChargeDateForUpdateContribution = (baseNextChargeDate, newInterval) => {
+const getNextChargeDateForUpdateContribution = (baseNextChargeDate: Date, newInterval: Order['interval']) => {
   const previousNextChargeDate = moment(baseNextChargeDate);
-  if (previousNextChargeDate.isBefore(moment())) {
-    // If the contribution was pending, keep it in the past
+  if (previousNextChargeDate.isSame(moment(), newInterval)) {
```
Betree, line 31: The logic change here will impact all types of subscription updates: when updating the payment method of a subscription, we will only charge the "previousNextChargeDate" if it's in the same interval (current month or year).

server/paymentProviders/paypal/subscription.ts
```diff
@@ -280,7 +280,7 @@ export const setupPaypalSubscriptionForOrder = async (
   try {
     await paypalRequest(`billing/subscriptions/${paypalSubscriptionId}/activate`, null, hostCollective, 'POST');
     if (order.PaymentMethodId !== paymentMethod.id) {
-      order = await order.update({ PaymentMethodId: paymentMethod.id });
+      order = await order.update({ PaymentMethodId: paymentMethod.id, status: ORDER_STATUS.PROCESSING }); // Will be marked as active by the webhook
```
Betree, line 283: Doing this will make sure the order is neither in the `ACTIVE` state (which would be false since it's not charged yet) nor in the `PAUSED` state (which is not the case either). Asked if that's safe in https://opencollective.slack.com/archives/C06K5J91675/p1717658997289129.

PR #10125, opened 2024-06-06
Fix Stripe Dispute logic

kewitz:
- [x] Disable refund for orders in dispute;
- [x] Use double transaction with Strip vendor to track DISPUTE_FEES;
- [x] Track DISPUTE_FEES even for `won` conditions;


znarf (approved):
Looks good!

PR #10128, opened 2024-06-07
Use stripe payment method configuration to create payment intent

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7413

Toggles between sets of payment methods defined on Stripe dashboard.


server/graphql/v2/mutation/OrderMutations.js
```diff
@@ -1039,8 +1039,16 @@ const orderMutations = {
       const currency = paymentIntentInput.currency;
 
       try {
+        let paymentMethodConfiguration = config.stripe.oneTimePaymentMethodConfiguration;
+
+        if (paymentIntentInput.frequency && paymentIntentInput.frequency !== 'ONETIME') {
```
znarf, line 1044: No constant for ONETIME ?

znarf (approved):
Looks good. Would be cleaner to have a constant to represent ONETIME.

PR #10130, opened 2024-06-10
fix(Expense): allow seeing payout method with the right draft key

Betree:
For https://github.com/opencollective/opencollective/issues/6379#issuecomment-2155472259


kewitz (approved):

mkurz:
When live? Thanks!

PR #10131, opened 2024-06-10
Switch to ts-unused-exports

znarf:
Similar to https://github.com/opencollective/opencollective-frontend/pull/10400 on the frontend side.


Betree (approved):
Looking good!

PR #10141, opened 2024-06-17
Allow pg_dump customization

Betree:
This aims to provide a workaround for this issue that occurs when a dump from a recent postgres version is meant to be restored on an older Postgres (e.g. on CI):

> pg_restore: error: unsupported version (1.15) in file header.

To enable it, set the `OC_PG_DUMP_CMD`  variable in your `.env`. Some examples:

- Run an older `pg_dump` with Docker:

> OC_PG_DUMP_CMD=docker run --net host postgres:15 pg_dump

- RUn an older `pg_dump` with Docker (sudo protected): 
> OC_PG_DUMP_CMD=pkexec docker run --net host postgres:15 pg_dump


Betree:
Closing as we're removing this optimization in https://github.com/opencollective/opencollective-api/pull/10145

PR #10149, opened 2024-06-19
Add managed/operational funds accounting category flag

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7420


migrations/20240613102553-add-accounting-category-applies-to.js
```diff
@@ -0,0 +1,32 @@
...
+      defaultValue: 'HOSTED_COLLECTIVES',
+    });
+
+    await queryInterface.sequelize.query(`
+      UPDATE "AccountingCategories"
+      SET "appliesTo" = 'HOSTED_COLLECTIVES'
+      WHERE "CollectiveId" IN (
+        SELECT id FROM "Collectives" where type = 'ORGANIZATION'
+      );
+    `);
```
Betree, line 18: This one doesn't seem necessary, given the `defaultValue` set above will already be applied to all rows?

server/graphql/v2/input/AccountingCategoryInput.ts
```diff
@@ -63,6 +65,11 @@ export const AccountingCategoryInput = new GraphQLInputObjectType({
       type: new GraphQLList(GraphQLExpenseType),
       description: 'If meant for expenses, the types of expenses this category applies to',
     },
+    appliesTo: {
+      type: GraphQLAccountingCategoryAppliesTo,
+      defaultValue: AccountingCategoryAppliesTo.HOSTED_COLLECTIVES,
```
Betree, line 70: Setting a default value here will prevent the line at https://github.com/opencollective/opencollective-api/pull/10149/files#diff-e389b9ca7581279f04ac04da2e828a014340eb970747273a49b9bbbf0ec935f8R63-R67 from doing its job (`isNil(input.appliesTo)` will never be true, unless `null` is explicitly passed)

Betree (approved):
Lgtm! Some minor feedback below

server/graphql/v2/input/AccountingCategoryInput.ts
hdiniz: I made the arg nullable to be backward, is that right?

migrations/20240613102553-add-accounting-category-applies-to.js
hdiniz: Yes

server/graphql/v2/input/AccountingCategoryInput.ts
Betree: If there's a default value, it will stay backward compatible when enforcing non-null. 

With a mutation like:

```es6
doStuff = {
  args: { something: type: new GraphQLNonNull(GraphQLString), defaultValue: 'Ok' }
}
``` 

Passing this is ok:

```es6
doStuff({ variables: {}})
```

But not this:

```es6
doStuff({ variables: { something: null }})
```

Betree (approved):
Good stuff!

PR #10196, opened 2024-07-10
enhancement(Expense): improve default logic for host expenses

Betree:
This PR originates from OSC being [confused](https://opencollective.slack.com/archives/GFH4N961L/p1720564063912079) as to why they could not categorize an expense that took place while they were not the collective host yet. 

It changes two things with the default Host expenses list query:
1. **Hide expenses that were paid by another fiscal host**: If a collective gets an expense paid while hosted by Host A, and then later moves to Host B, Host B will not see the expense in its expense list.
2. **Surface PAID expenses for unhosted collectives**: If a collective gets an expense paid while hosted by Host A, and then later moves to another host/independent/or gets archived, the expense will stay visible in the expense list of Host A.


kewitz (approved):
:shipit: This makes sense and it is in sync with the idea expressed in the PR body.

PR #10194, opened 2024-07-09
fix(Budget): ignore expense processor fees for total received

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7465
Resolve https://github.com/opencollective/opencollective/issues/7464

PR #10202, opened 2024-07-12
fix(TransactionsStats): update CollectiveTransactionStats

Betree:
Follow-up on https://github.com/opencollective/opencollective-api/pull/10194

The previous fix works for the JS version. However, fast balances are still affected by the bug because the change was only made on `CurrentCollectiveTransactionStats`, without being propagated to `CollectiveTransactionStats`.

### DIFF
--- a/migrations/20240712081151-current-collective-views-update.js
+++ b/migrations/20240712081151-current-collective-views-update.js
@@ -42,8 +42,8 @@ module.exports = {
             COALESCE(t."taxAmount" * t."hostCurrencyFxRate", 0)
           )
             FILTER (
-              -- Get all credits related to orders
-              WHERE (t.type = 'CREDIT' AND t."OrderId" IS NOT NULL)
+              -- Get all credits except PAYMENT_PROCESSOR_COVER from expenses
+              WHERE (t.type = 'CREDIT' AND NOT (t.kind = 'PAYMENT_PROCESSOR_COVER' AND t."OrderId" IS NULL))
               -- Deduct host fees and payment processor fees that are related to orders
               OR (t.type = 'DEBIT' AND t.kind IN ('HOST_FEE', 'PAYMENT_PROCESSOR_FEE') AND t."OrderId" IS NOT NULL)
             )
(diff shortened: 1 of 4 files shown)


znarf (approved):

migrations/20240712081151-current-collective-views-update.js
```diff
@@ -0,0 +1,278 @@
...
+            COALESCE(t."hostFeeInHostCurrency", 0) +
+            COALESCE(t."paymentProcessorFeeInHostCurrency", 0) +
+            COALESCE(t."taxAmount" * t."hostCurrencyFxRate", 0)
+          )
+            FILTER (
+              -- Get all credits related to orders
+              WHERE (t.type = 'CREDIT' AND t."OrderId" IS NOT NULL)
+              -- Deduct host fees and payment processor fees that are related to orders
+              OR (t.type = 'DEBIT' AND t.kind IN ('HOST_FEE', 'PAYMENT_PROCESSOR_FEE') AND t."OrderId" IS NOT NULL)
+            )
```
Betree, line 49: I changed the definition here after `test/stories/ledger >Level 1: Same currency (USD) > 6. Expense marked as unpaid with Payment Processor fees NOT REFUNDED` surfaced another related issue. Taking this ledger as a reference:
image

The code was returning $1505 for the net total amount received. The reason was that it was summing the contribution ($1500) + the expense `PAYMENT_PROCESSOR_COVER` ($5). I chose to address that by filtering on `OrderId` not null, to stay simple and consistent. If we're worried about potential regressions, another option would be to be more specific:

```sql
WHERE (t.type = 'CREDIT' AND (t.kind !== 'PAYMENT_PROCESSOR_COVER' OR t."OrderId" IS NOT NULL))
```

@znarf Wdyt?

znarf: > I chose to address that by filtering on OrderId not null, to stay simple and consistent.

This will exclude incoming expenses as money raised and I think this is gonna be a problem at some point in the future.

znarf:
Latest version looks good.

PR #10205, opened 2024-07-16
fix(Transactions): handle partial refunds

Betree:
Resolve https://github.com/opencollective/opencollective/issues/6338
Resolve https://github.com/opencollective/opencollective/issues/7479

### DIFF
--- a/server/lib/payments.ts
+++ b/server/lib/payments.ts
@@ -29,7 +29,6 @@ import { RecipientAccount as BankAccountPayoutMethodData } from '../types/transf
 import { notify } from './notifications/email';
 import { getFxRate } from './currency';
 import emailLib from './email';
-import logger from './logger';
 import { getTransactionPdf } from './pdf';
 import { createPrepaidPaymentMethod, isPrepaidBudgetOrder } from './prepaid-budget';
 import { getNextChargeAndPeriodStartDates } from './recurring-contributions';
@@ -181,7 +180,15 @@ export const buildRefundForTransaction = (
   t: Transaction,
   user?: User,
   data?: TransactionData,
-  refundedPaymentProcessorFee?: number,
+  {
+    refundedPaymentProcessorFee = null,
+    refundedAmount = null,
+  }: {
+    /** If the payment processor fee is not recorded as a separate transaction, use this field to refund it as part of the main transaction */
+    refundedPaymentProcessorFee?: number;
+    /** If the refunded amount is different from the original transaction amount */
+    refundedAmount?: number;
+  } = {},
 ): TransactionCreationAttributes => {
   const refund = pick(t, [
     'currency',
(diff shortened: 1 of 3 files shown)


server/lib/payments.ts
```diff
@@ -207,24 +214,27 @@ export const buildRefundForTransaction = (
...
   refund.description = `Refund of "${t.description}"`;
   refund.data = { ...refund.data, ...data };
+  refund.isRefund = true;
+  refund.hostFeeInHostCurrency = 0; // We're handling host fees in separate transactions
 
   /* The refund operation moves back fees to the user's ledger so the
    * fees there should be positive. Since they're usually in negative,
    * we're just setting them to positive by adding a - sign in front
    * of it. */
-  refund.hostFeeInHostCurrency = -refund.hostFeeInHostCurrency;
```
kewitz, line 215: I think this is here to guarantee backward compatibility because `refundHostFee()` only works with newer transactions with split fees. It should be safe to remove this by now, but we need to be conscious that refunding old unified transactions in development environment won't work properly.

server/lib/payments.ts
```diff
@@ -207,24 +214,27 @@ export const buildRefundForTransaction = (
...
-
-  // We're handling host fees in separate transactions
-  refund.hostFeeInHostCurrency = 0;
+  if (isNil(refundedAmount)) {
+    refund.amount = -t.amount;
+    refund.amountInHostCurrency = -t.amountInHostCurrency;
+    refund.netAmountInCollectiveCurrency = -Transaction.calculateNetAmountInCollectiveCurrency(t);
+  } else {
+    refund.amount = -refundedAmount;
+    refund.amountInHostCurrency = -Math.round(refundedAmount * refund.hostCurrencyFxRate);
```
kewitz, line 235: Are we sure `refundedAmount` is always in collective currency?

server/lib/payments.ts
```diff
@@ -181,7 +180,15 @@ export const buildRefundForTransaction = (
...
   data?: TransactionData,
-  refundedPaymentProcessorFee?: number,
+  {
+    refundedPaymentProcessorFee = null,
+    refundedAmount = null,
+  }: {
+    /** If the payment processor fee is not recorded as a separate transaction, use this field to refund it as part of the main transaction */
+    refundedPaymentProcessorFee?: number;
+    /** If the refunded amount is different from the original transaction amount */
+    refundedAmount?: number;
```
kewitz, line 190: Perhaps adding the `InCollectiveCurrency` suffix to these amounts.

server/lib/payments.ts
Betree: @kewitz This line had no impact since we were forcing the `hostFeeInHostCurrency` to `0` at line 227.

Betree:
@kewitz Thanks for the feedback! I've iterated on the function prototype to make the args clearer.

While doing that I realized that I forgot to cover the case for processor fees cover: if a transaction with $3 fees gets refunded with $2 fees, the host should still wire a $1 `PAYMENT_PROCESSOR_COVER` to the collective to cover their loss.
This is now implemented, along with some tests to cover multi-currency.

Betree:
It looks like there's an inconsistency in the code with `createRefundTransaction` being given the `refundedPaymentProcessorFee` in host currency in most cases but in transaction currency with Stripe. Currently analyzing the impact.

kewitz (approved):
Nice, I think this makes sense and the test seems solid.

PR #10207, opened 2024-07-16
Adds runCronJob helper function

kewitz:
This should prevent any potential double execution of our cron jobs.


server/lib/mutex.ts
```diff
@@ -44,3 +44,40 @@ export async function lockUntilResolved<T>(
...
+  try {
+    const response = await until();
+    await redis.del(_key);
+    debug(`Released lock ${_key}`);
+    return response;
+  } catch (e) {
+    await redis.del(_key);
+    debug(`Released lock ${_key}`);
+    throw e;
+  }
```
Betree, line 82: This code flow looks a bit strange, `catch` could be triggered because Redis is unavailable. If the goal is just to clear the lock whatever happens, maybe you intended to do something like this?

```suggestion
  try {
    return await until();
  } finally {
    await redis.del(_key);
    debug(`Released lock ${_key}`);
  }
```

cron/utils.ts
```diff
@@ -12,3 +17,30 @@ export function onlyExecuteInProdOnMondays() {
...
   }
 }
+
+export const runCronJob = async (name: string, run: () => Promise<any>, timeoutMs: number) => {
+  let exitCode = 0;
+  const isNotTest = process.env.NODE_ENV !== 'test' && process.env.NODE_ENV !== 'ci' && config.env !== 'e2e';
+  if (require.main !== module && isNotTest) {
+    console.warn('This script is not meant to be required, please run it directly');
+    return;
+  }
```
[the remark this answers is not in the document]
Betree: Confirmed: It looks like the imported CRON jobs are now started with CI

image

test/cron/utils.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    let secondCompletes = false;
+    const second = runCronJob(
+      'test',
+      async () => {
+        await new Promise(resolve => setTimeout(resolve, 100));
+        secondCompletes = true;
+      },
+      10000,
+    );
```
Betree, line 37: Running this locally gave me an error:
image

Looking at the code, there's not guarantee of execution order here: first and second are started in parallel, the same process tick. Suggestions:

```suggestion
    let firstCompletes = false;
    const first = runCronJob(
      'test',
      async () => {
        await new Promise(resolve => setTimeout(resolve, 100));
        firstCompletes = true;
      },
      10000,
    );

    await sleep(1); // Make sure the second job will start at the next process tick

    let secondCompletes = false;
    const second = runCronJob(
      'test',
      async () => {
        await new Promise(resolve => setTimeout(resolve, 100));
        secondCompletes = true;
      },
      10000,
    );
```

Betree (commented):
Great approach, and the mutex implementation looks solid :corn: :corn: :corn:

Some feedback on the details below.

cron/utils.ts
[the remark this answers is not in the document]
kewitz: Not related to the warning, this is in place so we can unit test this function.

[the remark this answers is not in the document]
kewitz: I'll add a dedicated env variable for this.

server/lib/mutex.ts
kewitz: The issue here is that the finally handler would eat any error thrown by until().

cron/utils.ts
```diff
@@ -8,7 +16,54 @@ import config from 'config';
...
+  name: string,
+  run: () => Promise<any>,
+  timeoutMs: number,
+  errorParameters?: CaptureErrorParams,
+) => {
+  let exitCode = 0;
+  if (require.main !== module && process.env.DISABLE_CRON_MODULE_START_CHECK === undefined) {
+    logger.warn('This script is not meant to be required, please run it directly');
+    return;
+  }
```
Betree, line 34: This env variable hacks around the problem but it will still exist in local dev. To me, the CRON job should never run when the file gets imported since you're not in control of when it will be executed. The right pattern for tests is to import the `main` function from tests and call it directly.

Or am I missing legit cases where we want the CRON job to start as soon as it's imported?

server/lib/mutex.ts
Betree: It won't. If an error is thrown by `until`, `finally` will run then the error will bumble up. Try this:

```ts
const testFinally = async (fn) => {
  try {
    return await fn();
  } finally {
    console.log('finally');
  }
}

const test = async () => {
  console.log('return value with no error:', await testFinally(() => Promise.resolve('hello')));
  
  console.log('Testing error:')
  try {
    await testFinally(() => Promise.reject('error'))
  } catch (e) {
    console.log('received error:', e);
  }
}


test()
```

> $ test()
> finally
> return value with no error: hello
> Testing error:
> finally
> received error: error

Betree: The key here is `return await`, that's how we make sure the error is raised inside of this `try finally`

cron/utils.ts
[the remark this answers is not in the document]
Betree: Not sure if this is the right approach, added a comment here: https://github.com/opencollective/opencollective-api/pull/10207#discussion_r1680487532

cron/utils.ts
kewitz: The problem is the early return of this function. This is not a root call, it is an early return within `runCronJob`. 

How do we test a call to `runCronJob` if we necessarily need to import that as a module in the test file?

kewitz: @Betree I don't think the problem will still exist in local dev since DISABLE_CRON_MODULE_START_CHECK is never set at the environment level. It's just manipulated by the test case itself.

Betree: Ok, I get it now. We would have handled this with a function parameter, but I'm fine with your approach:

```
export const runCronJob = async (
  name: string,
  run: () => Promise<any>,
  timeoutSeconds: number,
  { errorParameters, forceIfNotMain } = {}
) => {
```

kewitz: LOL. That was very obvious.

kewitz: LOL, `require.main !== module` is always true here, because we import this function from a file. 😂

kewitz:
@Betree the `require.main` check was a bust. I refactored to make sure the check was relevant which simplified the `runCronJob` function and tests.

I'll wait for a last OK from you so I can merge and deploy.

cron/utils.ts
```diff
@@ -8,7 +16,49 @@ import config from 'config';
...
+  let exitCode = 0;
+  try {
+    await lockUntilOrThrow(
+      `cron:${name}`,
+      async () => {
+        const start = Date.now();
+        logger.info(`Starting CRON job: ${name}`);
+        await run();
+        const duration = Date.now() - start;
+        logger.info(`CRON job finished in ${Math.round(duration / 1000)}s: ${name} `);
```
Betree, line 39: Great addition :+1:

cron/utils.ts
```diff
@@ -8,7 +16,49 @@ import config from 'config';
...
+  }
+
+  const isNotTest = process.env.NODE_ENV !== 'test' && process.env.NODE_ENV !== 'ci' && config.env !== 'e2e';
+  if (isNotTest) {
+    await closeRedisClient();
+    await sequelize.close();
+    process.exit(exitCode);
+  } else {
+    logger.info(`CRON would exit with code ${exitCode}`);
+  }
```
Betree, line 63: Good idea to wait for Sentry to flush the errors, but this `sleep` is not a safe way to guarantee that. Suggestion:

```suggestion
    logger.error(`Error running ${name} job`, error);
    reportErrorToSentry(
      error,
      deepmerge({ handler: 'CRON', extra: { name, timeout: timeoutSeconds } }, errorParameters || {}),
    );
  }

  const isNotTest = process.env.NODE_ENV !== 'test' && process.env.NODE_ENV !== 'ci' && config.env !== 'e2e';
  if (isNotTest) {
    await closeRedisClient();
    await sequelize.close();
    await Sentry.close(10e3); // Wait for Sentry to flush errors, with a timeout of 10
    process.exit(exitCode);
  } else {
    logger.info(`CRON would exit with code ${exitCode}`);
  }
```

See https://docs.sentry.io/platforms/javascript/guides/node/configuration/draining/

Betree (commented):
Good changes! Just a small note on waiting Sentry, the rest looks ready to ship

PR #10212, opened 2024-07-17
Refactor defaultHostCollective to prevent crash when missing platform

znarf:


Betree (approved):
Looks cleaner :ok_hand:

PR #10214, opened 2024-07-18
Remove early exits to prevent cron-lock hangs

kewitz:


cron/daily/95-checks.js
```diff
@@ -22,10 +22,8 @@ async function run() {
 
     const text = `${failureMessage}\n\n${errors.map(msg => `- ${msg}`).join('\n')}\n\n${fixMessage} ${fixCommand}`;
 
-    return email.sendMessage(recipients, subject, html, { text });
+    return await email.sendMessage(recipients, subject, html, { text });
```
znarf, line 25: You're already awaiting in the `runCronJob`, I don't think there is much to get by awaiting here.

I'm generally skeptic on the `return await` pattern.

PR #10224, opened 2024-07-24
Create Export lib to store some smart-dump functions

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7494


Betree (approved):
Good stuff :+1: I'd go even further and have a dedicated folder for this feature to store the different part that are going to be implemented and shouldn't fit in a single file, something like:
- lib/smart-dump/index.ts
- lib/smart-dump/export.ts
- lib/smart-dump/import.ts
- lib/smart-dump/sanitizer.ts
- ...etc

kewitz:
@Betree OK, I'll give another pass at it and split it even further so we can minimize possible conflicts later on.

PR #10225, opened 2024-07-24
Fix double-refund bug of host fees and platform tips

kewitz:
Related to https://github.com/opencollective/opencollective/issues/7482

I was reviewing the refund flow and noticed that, whenever a fiscal-host admin tries to refund a HOST_FEE transaction directly, we end up refunding it twice. The same is true for platform tips.

### DIFF
--- a/server/lib/payments.ts
+++ b/server/lib/payments.ts
@@ -374,7 +374,7 @@ export async function refundHostFee(
   transactionGroup: string,
   clearedAt?: Date,
 ): Promise<void> {
-  const hostFeeTransaction = await transaction.getHostFeeTransaction();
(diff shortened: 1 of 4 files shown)


Betree (approved):
Nice catch! Luckily, it looks like there's no case yet in our records:

```sql
select "TransactionGroup", kind, COUNT(*)
from "Transactions"
WHERE kind IN ('PLATFORM_TIP', 'HOST_FEE')
AND type = 'DEBIT'
AND "isRefund" IS TRUE
AND "deletedAt" IS NULL
GROUP BY "TransactionGroup", kind
HAVING COUNT(*) > 1
```

PR #10229, opened 2024-07-27
Add threads to host applications

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7480


gustavlrsn (approved):
Nice work, looks great to me!

migrations/20240723144352-add-comments-to-host-applications.js
```diff
@@ -0,0 +1,25 @@
...
+      references: {
+        key: 'id',
+        model: 'HostApplications',
+      },
+    });
+
+    await queryInterface.addColumn('CommentHistories', 'HostApplicationId', {
+      type: Sequelize.INTEGER,
+      allowNull: true,
+    });
```
Betree, line 18: Could be nice to add an index to avoid the increasing complexity as the table grows, similar to what we do for the other comment types:

```
    await queryInterface.addIndex('Comments', ['HostApplicationId'], {
      concurrently: true,
      where: {
        HostApplicationId: { [Op.ne]: null },
      },
    });
```

server/graphql/common/comment.ts
```diff
@@ -188,7 +212,7 @@ async function createComment(commentData, req): Promise<Comment> {
     UserId: comment.CreatedByUserId,
     CollectiveId: comment.CollectiveId,
     FromCollectiveId: comment.FromCollectiveId,
-    HostCollectiveId: commentedEntity.collective.approvedAt ? commentedEntity.collective.HostCollectiveId : null,
+    HostCollectiveId: 'HostCollectiveId' in commentedEntity ? commentedEntity.HostCollectiveId : null,
```
Betree, line 215: I don't agree with that part, the `approvedAt` check was there for a reason: we don't want to record the `HostCollectiveId` for other comment types if the collective hasn't been approved yet; otherwise this could be abused to spam a fiscal host's activity log.

And checking the `HostCollectiveId` is not reliable:
- The field doesn't exist for orders
- The field doesn't exist for updates
- For expenses, it's only set at payment time, so we'll be recording comments posted before payment as host-less while comments after payment will have a host attached.

I understand why we want to treat things differently for host applications, but let's preserve the functionality for other comment types:
```
HostCollectiveId:
  commentedEntity instanceof HostApplication ?
    commentedEntity.HostCollectiveId :
    commentedEntity.collective.approvedAt ?
      commentedEntity.collective.HostCollectiveId
      : null

File: server/graphql/common/host-applications.ts [hunk 3]
```diff
@@ -0,0 +1,26 @@
...
+export async function canMakePrivateNoteOnHostApplication(
+  req: Express.Request,
+  hostApplication: HostApplication,
+): Promise<boolean> {
+  if (req.remoteUser.isAdmin(hostApplication.HostCollectiveId)) {
+    return true;
+  }
+
+  return false;
+}
```
Betree, line 26: No need to make these functions async

server/graphql/v2/input/CommentCreateInput.js
```diff
@@ -20,6 +21,10 @@ export const GraphQLCommentCreateInput = new GraphQLInputObjectType({
       type: GraphQLExpenseReferenceInput,
       description: 'If your comment is linked to an expense, set it here',
     },
+    hostApplication: {
+      type: GraphQLHostApplicationReferenceInput,
+      description: 'If your comment is linked to an host application, set it here',
+    },
```
Betree, line 27: description at line 17 should be updated accordingly

server/graphql/v2/object/HostApplication.ts
```diff
@@ -74,5 +79,50 @@ export const GraphQLHostApplication = new GraphQLObjectType({
...
+          defaultValue: { field: 'createdAt', direction: 'ASC' },
+        },
+      },
+      async resolve(hostApplication, { limit, offset, orderBy }, req) {
+        if (
+          !req.remoteUser?.isAdmin(hostApplication.HostCollectiveId) &&
+          !req.remoteUser?.isAdmin(hostApplication.CollectiveId)
+        ) {
+          return null;
+        }
```
Betree, line 99: :ice_cream: I would reuse `canCommentHostApplication` here for consistency

server/graphql/v2/object/HostApplication.ts
```diff
@@ -74,5 +79,50 @@ export const GraphQLHostApplication = new GraphQLObjectType({
...
+
+        const type = [CommentType.COMMENT];
+        if (req.remoteUser?.isAdmin(hostApplication.HostCollectiveId)) {
```
Betree, line 102: :ice_cream: And `canMakePrivateNoteOnHostApplication` on this one

server/lib/emailTemplates.ts
```diff
@@ -154,6 +157,9 @@ handlebars.registerPartial('eventdata', eventdata);
 handlebars.registerPartial('charge_date_notice', chargeDateNotice);
 handlebars.registerPartial('mr-footer', mthReportFooter);
 handlebars.registerPartial('mr-subscription', mthReportSubscription);
+handlebars.registerHelper('idEncode', (id, type) => {
+  return idEncode(id, type);
+});
```
Betree, line 162: :ok_hand: nice addition

PR #10230, opened 2024-07-27
export and import dump from jsonl files

hdiniz:
Resolves https://github.com/opencollective/opencollective/issues/7491
Resolves https://github.com/opencollective/opencollective/issues/7489


scripts/smart-dump.ts
```diff
@@ -40,31 +42,30 @@ program.command('dump [recipe] [env]').action(async (recipe, env) => {
   const date = new Date().toISOString().substring(0, 10);
   const hash = md5(JSON.stringify({ entries, defaultDependencies, date })).slice(0, 5);
   const filename = `${date}.${hash}`;
-  let docs = [];
+  const seenModelRecords: Set<string> = new Set();
```
kewitz, line 45: 🫰

scripts/smart-dump.ts
```diff
@@ -40,31 +42,30 @@ program.command('dump [recipe] [env]').action(async (recipe, env) => {
...
   for (const entry of entries) {
     logger.info(`>>> Traversing DB for entry ${entries.indexOf(entry) + 1}/${entries.length}...`);
-    const newdocs = await traverse({ ...entry, defaultDependencies, parsed });
-    docs.push(...newdocs);
+    await traverse({ ...entry, defaultDependencies, parsed }, ei => {
+      const modelRecordKey = `${ei.model}.${ei.id}`;
+      if (!seenModelRecords.has(modelRecordKey)) {
+        dumpFile.write(JSON.stringify(ei) + os.EOL);
+        seenModelRecords.add(modelRecordKey);
+      }
```
kewitz, line 57: Awesome.

scripts/smart-dump.ts
```diff
@@ -40,31 +42,30 @@ program.command('dump [recipe] [env]').action(async (recipe, env) => {
...
+  logger.info(`>>> Dumping... to dbdumps/${filename}.jsonl`);
+  const dumpFile = fs.createWriteStream(`dbdumps/${filename}.jsonl`);
   for (const entry of entries) {
     logger.info(`>>> Traversing DB for entry ${entries.indexOf(entry) + 1}/${entries.length}...`);
-    const newdocs = await traverse({ ...entry, defaultDependencies, parsed });
-    docs.push(...newdocs);
+    await traverse({ ...entry, defaultDependencies, parsed }, ei => {
+      const modelRecordKey = `${ei.model}.${ei.id}`;
+      if (!seenModelRecords.has(modelRecordKey)) {
+        dumpFile.write(JSON.stringify(ei) + os.EOL);
```
kewitz, line 55: Perhaps we should await the write to be done here.

server/lib/import-export/export.ts
```diff
@@ -105,7 +98,10 @@ export const traverse = async ({
     } else {
       records.forEach(r => parsed[model].add(r.id));
     }
-    acc.push(...records);
+
+    records.forEach(element => {
+      callback(element);
```
kewitz, line 103: Maybe await for each write to be done?

scripts/smart-dump.ts
hdiniz: The writes are syncronous

scripts/smart-dump.ts
```diff
@@ -34,37 +38,51 @@ program.command('dump [recipe] [env]').action(async (recipe, env) => {
     recipe = './smart-dump/defaultRecipe.js';
   }
 
+  const tempDumpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'oc-export-'));
```
Betree, line 41: Not sure why the dir name ends with `-`, you maybe wanted to add a unique ID there?

scripts/smart-dump.ts
```diff
@@ -34,37 +38,51 @@ program.command('dump [recipe] [env]').action(async (recipe, env) => {
...
   // eslint-disable-next-line @typescript-eslint/no-var-requires
   const { entries, defaultDependencies } = require(recipe);
   const parsed = {};
   const date = new Date().toISOString().substring(0, 10);
   const hash = md5(JSON.stringify({ entries, defaultDependencies, date })).slice(0, 5);
-  const filename = `${date}.${hash}`;
-  let docs = [];
+  const seenModelRecords: Set<string> = new Set();
+
+  const gitRevision = execSync('git describe --always --abbrev=0 --match "NOT A TAG" --dirty="*"').toString().trim();
```
Betree, line 51: That could fail in production, where (I believe) `git` is not available. Not a blocker, but we should ideally implement a fallback.

Betree (approved):
Awesome!

scripts/smart-dump.ts
hdiniz: `mkdtempSync` generates a unique temp directory with a given prefix, so this results in something like `/tmp/os-export-c122Fuytasd`

PR #10234, opened 2024-07-31
Automatically verify expenses submitted with a draft-key

kewitz:
If the invited user is submitting the expense with the correct key and the same email the creator of the draft used to invite the payee, therefore we can consider the access to the email was already verified and the expense can move straight to the PENDING status.

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -1819,13 +1819,13 @@ export async function submitExpenseDraft(
     throw new Unauthorized('Expense can not be edited.');
   }
 
-  const userIsOriginalPayee = originalPayee && req.remoteUser?.isAdminOfCollective(originalPayee);
-  const userIsAuthor = req.remoteUser?.id === existingExpense.UserId;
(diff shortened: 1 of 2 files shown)


server/graphql/common/expenses.ts
```diff
@@ -1819,13 +1819,13 @@ export async function submitExpenseDraft(
     throw new Unauthorized('Expense can not be edited.');
   }
 
-  const userIsOriginalPayee = originalPayee && req.remoteUser?.isAdminOfCollective(originalPayee);
-  const userIsAuthor = req.remoteUser?.id === existingExpense.UserId;
+  const userIsOriginalPayee: boolean = Boolean(originalPayee) && req.remoteUser?.isAdminOfCollective(originalPayee);
+  const userIsAuthor = Boolean(req.remoteUser) && req.remoteUser?.id === existingExpense.UserId;
```
Betree, line 1823: No need to double check

```suggestion
  const userIsAuthor = Boolean(req.remoteUser) && req.remoteUser.id === existingExpense.UserId;
```

Betree (approved):
Lgtm!

PR #10238, opened 2024-08-05
Prevent double refund for platform tip dept as well

kewitz:


Betree (approved):

Betree:
I sof-deleted transactions #7836311 and #7836312 that were part of this bug. Prod data now looks clean:

```sql
select "TransactionGroup", kind, COUNT(*)
from "Transactions"
WHERE kind IN ('PLATFORM_TIP_DEBT')
AND type = 'DEBIT'
AND "isRefund" IS TRUE
AND "deletedAt" IS NULL
GROUP BY "TransactionGroup", kind
HAVING COUNT(*) > 1
```

PR #10241, opened 2024-08-07
Edit Added Funds mutations

kewitz:
- [x] Add missing tests


server/graphql/v2/mutation/AddedFundsMutations.ts
```diff
@@ -0,0 +1,352 @@
+import assert from 'assert';
```
Betree, line 1: The rename makes it difficult to understand what changed in the file, and I'm not 100% convinced that the new name is any clearer. I would keep that for a separate PR.

server/graphql/v2/mutation/AddedFundsMutations.ts
```diff
@@ -0,0 +1,352 @@
...
+
+      // Check amounts
+      const totalAmount = getValueInCentsFromAmountInput(args.amount, { expectedCurrency: account.currency });
+      const paymentProcessorFee = args.paymentProcessorFee
+        ? getValueInCentsFromAmountInput(args.paymentProcessorFee, { expectedCurrency: account.currency })
+        : 0;
+
+      validateAddFundsArgs({ account, fromAccount, args, paymentProcessorFee, totalAmount, host, req });
+
+      await twoFactorAuthLib.enforceForAccount(req, host, { onlyAskOnLogin: true });
```
Betree, line 289: Not especially for this PR, but since added funds let host admins change the balance of accounts, we should be stricter than that on 2FA - for both add funds and edit funds.

server/graphql/v2/mutation/AddedFundsMutations.ts
```diff
@@ -0,0 +1,352 @@
...
+      await order.update({
+        CreatedByUserId: req.remoteUser.id,
+        FromCollectiveId: fromAccount.id,
+        CollectiveId: account.id,
+        totalAmount: totalAmount,
+        currency: account.currency,
+        description: args.description || order.description,
+        status: OrderStatuses.NEW,
+        TierId: tier?.id || order.TierId,
+        AccountingCategoryId: accountingCategory?.id || order.AccountingCategoryId,
```
Betree, line 315: :no_entry_sign: We must make sure that the accounting cateogory belongs to the host first. For add funds, this is done in the `common` lib function, by relying on `checkCanUseAccountingCategoryForOrder`.

server/graphql/v2/mutation/AddedFundsMutations.ts
```diff
@@ -0,0 +1,352 @@
...
+      // Update existing Order
+      await order.update({
+        CreatedByUserId: req.remoteUser.id,
+        FromCollectiveId: fromAccount.id,
+        CollectiveId: account.id,
+        totalAmount: totalAmount,
+        currency: account.currency,
+        description: args.description || order.description,
+        status: OrderStatuses.NEW,
+        TierId: tier?.id || order.TierId,
```
Betree, line 314: :no_entry_sign: We must make sure that the tier belongs to the collective, see https://github.com/opencollective/opencollective-api/blob/f5a08c5e2823c1df94f16ff33c724924e80d8f21/server/graphql/common/orders.ts#L89-L91

server/graphql/v2/mutation/AddedFundsMutations.ts
```diff
@@ -0,0 +1,352 @@
...
+      const transactions = await order.getTransactions({ order: [['id', 'desc']] });
+      assert(transactions.length > 0, 'No ADDED FUNDS transaction found for this order');
+      const editedTransactions = transactions.map(t => t.id);
+      await Promise.all(transactions.map(transaction => transaction.destroy()));
+
+      const previousData = order.toJSON();
+      // Update existing Order
+      await order.update({
+        CreatedByUserId: req.remoteUser.id,
+        FromCollectiveId: fromAccount.id,
```
Betree, line 308: :no_entry_sign: There are some checks to be made on `FromCollectiveId`: https://github.com/opencollective/opencollective-api/blob/f5a08c5e2823c1df94f16ff33c724924e80d8f21/server/graphql/common/orders.ts#L74

server/graphql/v2/mutation/AddedFundsMutations.ts
```diff
@@ -0,0 +1,352 @@
...
+    throw new ValidationFailed('hostFeePercent should be a value between 0 and 100.');
+  } else if (args.tax && (args.tax.rate < 0 || args.tax.rate > 1)) {
+    throw new ValidationFailed('Tax rate must be between 0 and 1');
+  } else if (paymentProcessorFee > totalAmount) {
+    throw new ValidationFailed('Payment processor fee cannot be higher than the total amount');
+  } else if (args.hostFeePercent === 100 && paymentProcessorFee > 0) {
+    throw new ValidationFailed('Payment processor fee cannot be applied when host fee is 100%');
+  } else if (!totalAmount) {
+    throw new ValidationFailed('Amount should be greater than 0');
+  }
```
Betree, line 88: This was covered by the add funds mutation, we need to make sure it's covered in the new one too

```suggestion
  } else if (!totalAmount || totalAmount < 0) {
    throw new ValidationFailed('Amount should be greater than 0');
  }
```

server/graphql/v2/mutation/AddedFundsMutations.ts
```diff
@@ -0,0 +1,352 @@
...
+        description: args.description || order.description,
+        status: OrderStatuses.NEW,
+        TierId: tier?.id || order.TierId,
+        AccountingCategoryId: accountingCategory?.id || order.AccountingCategoryId,
+        data: {
+          ...order.data,
+          hostFeePercent: args.hostFeePercent ?? order.data.hostFeePercent,
+          paymentProcessorFee: paymentProcessorFee ?? order.data.paymentProcessorFee,
+          memo: args.memo ?? order.data.memo,
+        },
```
Betree, line 321: Since we're not looking at taxes, I wonder what would happen if you try to edit added funds that have taxes attached?

Betree (changes requested):
Looks good overall, but we need to make sure all the validations made when adding funds are also there when editing them. Great if we can share the code for validations between the two functions.

server/graphql/v2/mutation/AddedFundsMutations.ts
kewitz: I've separated the rename commit from the edits so you can isolate the diff:
https://github.com/opencollective/opencollective-api/pull/10241/commits/97262fd3c084f8774341996ae96198f73fe48574#diff-8b9a3f71fc3209ae45f2ce922a6f80cda530d1c1580574ce9e1603fc58fdb90

I renamed it because this file was a single mutation with the same file name.

PR #10248, opened 2024-08-13
fix(Expense): update PaymentMethodId

Betree:
Follow-up on https://github.com/opencollective/opencollective/issues/7212

### DIFF
--- a/migrations/20240814102434-add-missing-PaymentMethodId-virtualcard-expenses.js
+++ b/migrations/20240814102434-add-missing-PaymentMethodId-virtualcard-expenses.js
@@ -1,0 +1,33 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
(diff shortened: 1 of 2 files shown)


kewitz (approved):

Betree:
Will add a migration to fix past data based on this query:

```sql
select pm.id, *
from "Expenses" e
INNER JOIN "Collectives" host ON e."HostCollectiveId" = host.id
INNER JOIN "PaymentMethods" pm ON pm."CollectiveId" = host.id AND pm.service = 'stripe' AND pm.type ='virtual_card'
WHERE (e.data ->> 'missingDetails')::boolean IS TRUE
AND e."PaymentMethodId" IS NULL
AND e."deletedAt" IS NULL
```

Additional info:
- Inner join on payments methods because all concerned hosts already have a stripe virtual card PM
- Inner join on collectives missed 24 expense from 2021 that don't have their fiscal host set (see https://github.com/opencollective/opencollective/issues/7515)

PR #10252, opened 2024-08-14
Refact/dump recipes

kewitz:
**Updates:**
- Add new typed recipe to export a small collective (engineering).
- Fixes several bugs with import/export routines
- Updates export tmp folder name to include hash of the export to make debugging easier
- Resets the sequential id counters after importing

**Instructions:**

1. Set up a `.env.prod` to connect to production with a read-only credentials
2. Run smart dump
    ```
    $ npm run script scripts/smart-dump.ts dump ./smart-dump/engineering.ts leokewitz prod
    
    info: Connecting to postgres://...
    info: Fixer API is not configured, lib/currency will always return 1.1
    info: >>> Temp directory: /tmp/oc-export-b5343Rk8k3W
    info: >>> Dumping Schema...
    info: >>> Schema Dumped in a few seconds
    info: >>> Dumping... to /tmp/oc-export-b5343Rk8k3W/data.jsonl
    info: >>> Traversing DB for entry 1/1...
    info: >>> Seen 3426 records...
    info: >>> Seen 3426 records...
    info: >>> Seen 5727 records...
    info: >>> Seen 7713 records...
    info: >>> Dumped! 7742 records in 2 minutes
    info: >>> Ziping export to... dbdumps/2024-08-14.b5343.zip
      adding: data.jsonl (deflated 88%)
      adding: schema.sql (deflated 90%)
      adding: metadata.json (deflated 76%)
    info: >>> Done! See dbdumps/2024-08-14.b5343.zip
    ```
3. Restore locally using postgres root user
    ```
    PG_USERNAME=postgres PG_DATABASE=oc_dump npm run script scripts/smart-dump.ts restore dbdumps/2024-08-14.b5343.zip
    ```


scripts/smart-dump.ts
```diff
@@ -29,7 +30,7 @@ const exec = cmd => {
   }
 };
 
-program.command('dump [recipe] [env] [as_user]').action(async (recipe, env, asUser) => {
+program.command('dump [recipe] [as_user]').action(async (recipe, asUser, env) => {
```
Betree, line 33: Why not include `[env]` in the command description since we're using it?

scripts/smart-dump/defaultRecipe.js
```diff
@@ -2,12 +2,12 @@ const entries = [
   {
```
Betree, line 2: The file name should ideally be kebab case (https://github.com/opencollective/opencollective/blob/main/rfcs/004-file-naming-convention.md#2-file-names-should-be-kebab-case).

It would also be great to have 2 separate recipes:
1. `host-admin-recipe.js`
2. `collective-admin-recipe.js`

Eventually importing one into the other if we want to combine the logic.

Betree (approved):
Lgtm!

scripts/smart-dump.ts
kewitz: Because this is part of the `npm run script` preloader. In reality, we're just getting env vars in this script.

PR #10253, opened 2024-08-15
Check if account has a required tax document

hdiniz:
:thread: https://opencollective.slack.com/archives/GFH4N961L/p1723739386351679


Betree (commented):
@hdiniz the goal of this line was to say: "only allow host admins to download the tax form if it has been requested for this specific year". In other words, if I submitted a tax form for OCF in 2020, they shouldn't automatically have access to my new 2024 tax form if I haven't submitted tax forms since.

I would therefore try to understand why, in the reported case, nothing was returned in the first place.

Betree:
Side-note: The way we check permission in this function is cumbersome, this would be greatly simplified by https://github.com/opencollective/opencollective/issues/7389.

Betree (approved):
Let's go with that as a hot fix, I'll comment on https://github.com/opencollective/opencollective/issues/7389 to make sure we address the year concern as a follow-up.

PR #10249, opened 2024-08-13
feat(PayPal): pause contributions

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7512

This PR introduces:
- Some developer docs on setting up PayPal locally (`docs/paypal.md`).
- The ability to pause & resume orders cleanly, including the deactivation on payment provider (PayPal) when needed
- A script to manually pause a contribution (`scripts/recurring-contributions/pause.ts`)

PR #10256, opened 2024-08-20
typescript: Convert payment utils

Betree:
Extracted from https://github.com/opencollective/opencollective-api/pull/10249 for cleaner diff

Side-note: it would be great if `persistVirtualCardTransaction` would use more explicit names, storing provider IDs in `data.id`, `data.token` is a bit too generic.

### DIFF
--- a/server/paymentProviders/stripe/virtual-cards.ts
+++ b/server/paymentProviders/stripe/virtual-cards.ts
@@ -14,7 +14,7 @@ import { reportMessageToSentry } from '../../lib/sentry';
 import stripe, { convertToStripeAmount, StripeCustomToken } from '../../lib/stripe';
 import models from '../../models';
 import VirtualCard from '../../models/VirtualCard';
-import { getOrCreateVendor, getVirtualCardForTransaction, persistTransaction } from '../utils';
+import { getOrCreateVendor, getVirtualCardForTransaction, persistVirtualCardTransaction } from '../utils';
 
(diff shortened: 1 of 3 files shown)


server/paymentProviders/stripe/virtual-cards.ts
```diff
@@ -285,7 +285,7 @@ export const processTransaction = async (stripeTransaction: Stripe.Issuing.Trans
   const isRefund = stripeTransaction.type === 'refund';
   const clearedAt = moment.unix(stripeTransaction.created).toDate();
 
-  return persistTransaction(virtualCard, {
+  return persistVirtualCardTransaction(virtualCard, {
```
Betree, line 288: Renamed for clarity

server/paymentProviders/utils.ts
```diff
@@ -174,10 +174,12 @@ export const persistTransaction = async (virtualCard, transaction) => {
...
         if (!originalCreditTransaction.RefundTransactionId) {
-          await createRefundTransaction(originalCreditTransaction, 0, {
-            refundTransactionId: transactionId,
-            transaction,
-          });
+          await createRefundTransaction(
+            originalCreditTransaction,
+            0,
+            { refundTransactionId: transactionId, transaction },
+            null,
```
Betree, line 181: `user` param does not have a default so we have to provide it. Since it's unclear who the user should be here, we're explicitly passing null.

PR #10257, opened 2024-08-20
Upgrade postgres to 16.x

znarf:


Betree:
Done incrementally through https://github.com/opencollective/opencollective-api/pull/10528 then https://github.com/opencollective/opencollective-api/pull/10536 then https://github.com/opencollective/opencollective-api/pull/10534

PR #10266, opened 2024-08-23
debt(GraphQLV1/editCollective): remove unused code

Betree:
Part of the cleanup related to https://github.com/opencollective/opencollective-frontend/pull/10054


server/graphql/v1/mutations/collectives.js
```diff
@@ -421,16 +418,6 @@ export function editCollective(_, args, req) {
...
-      .then(() => collective.editTiers(args.collective.tiers))
-      .then(() => {
-        // @deprecated since 2019-10-21: now using dedicated `editCoreContributors` endpoint
-        if (args.collective.members) {
-          return collective.editMembers(args.collective.members, {
-            CreatedByUserId: req.remoteUser.id,
-            remoteUserCollectiveId: req.remoteUser.CollectiveId,
-          });
-        }
-      })
```
Betree, line 433: I couldn't find a single place in the frontend where we're still using `editCollective` to edit members, even the legacy places like the onboarding modal still use the `editCoreContributors` mutation.

PR #10270, opened 2024-08-26
Wise: Quote mismatch fix

kewitz:
Whenever Wise quotes USD to USD it is presumed that the transaction is domestic, and at payment time, if they figure out the payee is not in the US and the transaction incurs a higher fee they'll just charge the account and move on.

The problem here is that the new incurred value is not returned either through their webhook event or through their API transfer information request. They'll update the existing quote and return the transfer information as if it had no fee. (!?)

What was done:
- [ ] Fix past transactions were quoted in the same currency as the Host but had the payee receiving it in another country. 
  - To be addressed in a separate PR with fixes based on each fiscal host.
- [x] Make sure we quote expenses not only using the currencies involved but also the recipient details.
- [x] When receiving a webhook request, make sure we retrieve the quote used from Wise.


scripts/wise.ts
```diff
@@ -45,7 +50,75 @@ program.command('check-expense <expenseId>').action(async expenseId => {
...
+      console.log(`✅ Expense #${expense.id} for ${expense.collective.slug}`);
+    } catch (e) {
+      problems.push([expense.id, expense.collective.slug, expense.createdAt, e.message, e.actual, e.expected]);
+      console.log(`❌ Expense #${expense.id} for ${expense.collective.slug} failed check: ${e.message}`);
+    }
+  }
+
+  if (problems.length > 1) {
+    console.error(`Found ${problems.length - 1} problems, writting it to wise-check.csv`);
+    fs.writeFileSync('wise-check.csv', problems.map(row => row.join(',')).join('\n'));
```
Betree, line 119: I suppose we'll follow up with a script to fix transactions for those past entries, based on the CSV results?

scripts/wise.ts
```diff
@@ -45,7 +50,75 @@ program.command('check-expense <expenseId>').action(async expenseId => {
...
+  if (!connectedAccount) {
+    return printAndExit(`${host.slug} not connected to Wise`, 'error');
+  }
+  const expenses = await models.Expense.findAll({
+    where: {
+      HostCollectiveId: host.id,
+      status: 'PAID',
+      updatedAt: { [Op.gte]: since },
+      data: { transfer: { [Op.ne]: null } },
+      feesPayer: 'COLLECTIVE',
```
Betree, line 89: If I understand correctly:

```suggestion
      feesPayer: 'COLLECTIVE', // We don't care about other feesPayer as payees normally don't have a balance
```

In which case it would be worth double-checking.

server/paymentProviders/transferwise/index.ts
```diff
@@ -133,6 +138,7 @@ async function quoteExpense(
   const targetCurrency = payoutMethod.unfilteredData.currency as string;
   const quoteParams = {
     profileId: connectedAccount.data.id,
+    // Attention: sourceCurrency must always be the host currency, we count with this when persisting the Processor Payment Fee
```
Betree, line 141: :+1:

server/paymentProviders/transferwise/webhook.ts
```diff
@@ -70,12 +75,29 @@ export async function handleTransferStateChange(event: TransferStateChangeEvent)
...
+    if (
+      // Check if existing quote is QuoteV3
+      'price' in paymentOption &&
+      // Check if the priceDecisionReferenceId has changed
+      paymentOption.price.priceDecisionReferenceId !== wisePaymentOption.price?.priceDecisionReferenceId
+    ) {
+      logger.warn(`Wise updated the payment option for expense ${expense.id}, updating existing values...`);
+      const expenseDataQuote = { ...omit(quote, ['paymentOptions']), paymentOption } as ExpenseDataQuoteV3;
+      await expense.update({ data: { ...expense.data, quote: expenseDataQuote, paymentOption } });
+      paymentOption = wisePaymentOption;
```
Betree, line 92: Sounds like we want to reverse these two lines, otherwise the `paymentOption` will not be updated

```suggestion
      paymentOption = wisePaymentOption;
      await expense.update({ data: { ...expense.data, quote: expenseDataQuote, paymentOption } });
```

scripts/wise.ts
kewitz: Yes, this should be addressed in the coming weeks together with affected fiscal hosts.

server/paymentProviders/transferwise/webhook.ts
kewitz: 🙌🏼

PR #10275, opened 2024-08-28
WIP: Add ConnectedAccount sameAs prototype

kewitz:


znarf (approved):
Good to me. Let's maybe ask @Betree to check.

Betree (commented):
An issue URL or PR description to understand what we're trying to solve?

kewitz:
@Betree https://github.com/opencollective/opencollective/issues/7528

Betree (commented):
Looks like a good hack to solve the issue.

I'm not a fan of using `data.sameAs`:
1. Because it's not a real reference (in SQL terms)
2. Because the name is not supper explicit

For (1), I have no blockers to merge this if we plan to revamp it in the near future.
For (2), it's not a blocker either but I can suggest:
- `ProxyConnectedAccountId`
- `DependsOnConnectedAccountId`
- `AliasOfConnectedAccountId`
- `ReferenceConnectedAccountId`
- `MirrorConnectedAccountId`

PR #10281, opened 2024-08-29
Wise: check-fees script

kewitz:


znarf (approved):
Looks good to me.

kewitz:
@znarf I'm adding the fix part in this PR. Should be ready today.

PR #10291, opened 2024-09-03
enhancement(ManualPaymentMethod): allow multi-currencies globally

Betree:
This was [blocking OSC](https://opencollective.slack.com/archives/GFH4N961L/p1724860931257429) and we couldn't find any justification for it. Added a quick test to make sure everything was in place.


znarf (approved):
Looks good!

PR #10297, opened 2024-09-04
Add Expense.reference column and resolvers

kewitz:
Part of https://github.com/opencollective/opencollective/issues/7531

### DIFF
--- a/server/lib/search.ts
+++ b/server/lib/search.ts
@@ -405,7 +405,9 @@ export const parseSearchTerm = (
   } else if (searchTerm.match(/^\d+\.?\d*$/)) {
     return { type: 'number', term: parseFloat(searchTerm), isFloat: searchTerm.includes('.') };
   } else {
-    return { type: 'text', term: searchTerm, words: words(searchTerm).length };
+    // We use a custom patter here because Lodash will split A123 to ['A', '123']
+    const wordsCount = words(searchTerm, /[^, -]+/g).length;
+    return { type: 'text', term: searchTerm, words: wordsCount };
   }
 };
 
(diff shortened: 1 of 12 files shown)


Betree (approved):
LGTM. In GraphQL descriptions, I would explicitly mention that it's a "**user-provided** reference", to make sure people understand the difference with `id`.

MsLady45 (approved):

MsLady45 (approved):

MsLady45 (approved):

server/lib/search.ts
```diff
@@ -405,7 +405,9 @@ export const parseSearchTerm = (
   } else if (searchTerm.match(/^\d+\.?\d*$/)) {
     return { type: 'number', term: parseFloat(searchTerm), isFloat: searchTerm.includes('.') };
   } else {
-    return { type: 'text', term: searchTerm, words: words(searchTerm).length };
+    // We use a custom patter here because Lodash will split A123 to ['A', '123']
```
Betree, line 408:
```suggestion
    // We use a custom pattern here because Lodash will split A123 to ['A', '123']
```

server/models/Expense.ts
```diff
@@ -932,6 +933,13 @@ Expense.init(
       defaultValue: false,
       allowNull: false,
     },
+
+    reference: {
+      type: DataTypes.STRING,
+      set(reference: string | null) {
+        this.setDataValue('reference', isNil(reference) ? null : reference.trim());
```
Betree, line 940: Unless we want to support empty strings, I would consolidate:

```suggestion
        this.setDataValue('reference', reference?.trim() || null);
```

PR #10304, opened 2024-09-05
Wise Check: Assertion to consider transfers that originally had no fees

kewitz:


scripts/wise/check-fees.ts
```diff
@@ -141,10 +141,23 @@ program.command('fix [hosts]').action(async hosts => {
 
     for (const expense of expenses) {
       try {
+        const [originalDebitTransaction, ...otherDebits] = expense.Transactions.filter(
+          t => t.type === 'DEBIT' && t.kind === TransactionKind.EXPENSE,
```
Betree, line 145: That works because we throw if there are multiple debits, but `model: models.Transaction` is not ordered so we have no guarantee that the 1st item in the array is the original debit

Betree (approved):

PR #10308, opened 2024-09-10
enhancement(PayPal): add expense reference to adaptive + payout

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7533


znarf (approved):
Perfect

PR #10311, opened 2024-09-10
Add project search to Fund/Collective fields

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7535

### DIFF
--- a/server/graphql/v2/interface/Account.ts
+++ b/server/graphql/v2/interface/Account.ts
@@ -708,7 +708,7 @@ const accountFieldsDefinition = () => ({
       accountType: {
         type: new GraphQLList(GraphQLAccountType),
       },
-      term: {
+      searchTerm: {
         type: GraphQLString,
       },
     },
@@ -733,9 +733,9 @@ const accountFieldsDefinition = () => ({
         };
       }
 
-      if (args.term) {
+      if (args.searchTerm) {
         where['searchTsVector'] = {
-          [Op.match]: Sequelize.fn('websearch_to_tsquery', args.term),
+          [Op.match]: Sequelize.fn('websearch_to_tsquery', args.searchTerm),
         };
       }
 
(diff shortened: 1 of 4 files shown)


server/graphql/v2/interface/AccountWithProjects.ts
```diff
@@ -0,0 +1,57 @@
...
+        type: GraphQLChronologicalOrderInput,
+        defaultValue: { field: 'createdAt', direction: 'ASC' },
+      },
+    },
+    type: GraphQLProjectCollection,
+    resolve(account, args): CollectionReturnType<Collective> {
+      const where = {
+        type: {
+          [Op.ne]: CollectiveType.VENDOR,
+        },
```
Betree, line 28: If I query `account.projects`, I expect all the items I'll get back to have the type `PROJECT` (I shouldn't get any events)

Betree (commented):
I don't get the need for this `account.projects` field, which serves the same purpose as `account.childrenAccounts` but with an ambiguous name. Can't we simply add support for a search term to the existing `childrenAccounts` field?

```gql
{
  collective(slug: "babel") {
    projects: childrenAccounts(accountType: PROJECT, searchTerm: "...") {
      ...
    }
  }
}
```

server/graphql/v2/interface/Account.ts
```diff
@@ -701,6 +701,9 @@ const accountFieldsDefinition = () => ({
       accountType: {
         type: new GraphQLList(GraphQLAccountType),
       },
+      term: {
```
Betree, line 704: I would enforce `searchTerm` as it's the GraphQL variable name we use for the vendors query, hosted accounts, expenses, and actually most (if not all?) places where we have a similar text search option.

```suggestion
      searchTerm: {
```

Betree (commented):
A small blocker on the variable name, the rest looks good!

Betree (approved):

PR #10313, opened 2024-09-12
enhancement(Collectives): handle images misuse

Betree:
Fix https://github.com/opencollective/opencollective/issues/7545 by ignoring the input when the frontend does something like this:

```ts
// We're trying to update `image`, but `imageUrl` contains the image service proxy URL,
// not the image URL itself.
updateVendorMutation({
  variables: { id: vendor.id, image: vendor.imageUrl },
})
```

### DIFF
--- a/server/lib/collectivelib.ts
+++ b/server/lib/collectivelib.ts
@@ -21,7 +21,7 @@ import { md5 } from './utils';
 
 const { USER } = CollectiveType;
 
-type AvatarUrlOpts = {
+type ImageUrlOpts = {
   height?: boolean;
   format?: string;
 };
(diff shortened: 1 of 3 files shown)


znarf (approved):
I'm all for forgiving user errors in an API.

Why not fix the mistake on the frontend also ? Why is it doing this ? Since when ?

Betree:
From the developer POV, it's an easy mistake to make:
1. Have the same form for create/edit
2. Populate your form with initialValues={account}
3. If the image is not edited, it will default to the current - `account.imageUrl` (aka the image proxy URL)

The right approach would normally be:
- To not pass anything for the image if there's no change
- Or to pass the raw S3 URL, but I'm not even sure we expose it on GraphQLV2 (and maybe that's a good thing?)

Since this can be confusing for someone unfamiliar with the API & images service, tolerating it at the model level sounded like the simplest and most future-proof solution.

PR #10314, opened 2024-09-12
Move from maildev to mailpit

Betree:
Replaces maildev by mailpit. In a nutshell:
- The server is not started automatically anymore, you need to run `npm run mailpit`. That would be a bit dirtier to achieve given that mailpit is not a JS/NPM package, and I figured this was ok given we don't work on emails daily.
- Mailpit has a great UI, really useful tools (like HTML compatibility checks), and a nicer integration with Cypress

Screenshot from 2024-09-12 11-57-21
Screenshot from 2024-09-12 11-57-12
Screenshot from 2024-09-12 11-57-07
Screenshot from 2024-09-12 11-57-03
Screenshot from 2024-09-12 11-57-00
Screenshot from 2024-09-12 11-56-57
Screenshot from 2024-09-12 11-56-51
Screenshot from 2024-09-12 11-56-40


scripts/dev/run-mailpit.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
hdiniz, line 1: Its also available as a docker container, in the interest of security and isolation it seems more suitable than installing a binary that run as the current user. And it works on all platforms.
https://mailpit.axllent.org/docs/install/docker/

PR #10317, opened 2024-09-17
Fix stale NEW stripe payment intent orders

hdiniz:


scripts/stripe/fix-stale-new-paymentIntents.ts
```diff
@@ -0,0 +1,54 @@
...
+import moment from 'moment';
+
+import OrderStatuses from '../../server/constants/order-status';
+import { PAYMENT_METHOD_SERVICE, PAYMENT_METHOD_TYPE } from '../../server/constants/paymentMethods';
+import { syncOrder } from '../../server/lib/stripe/sync-order';
+import { Collective, Op, Order, PaymentMethod } from '../../server/models';
+
+const main = async () => {
+  const results = await Order.findAndCountAll({
+    limit: 20,
```
kewitz, line 12: We should either remove/increase the limit or implement pagination:
```suggestion
```

cron/hourly/20-fix-stripe-stale-NEW-paymentIntents.ts
```diff
@@ -0,0 +1,54 @@
...
+import logger from '../../server/lib/logger';
+import { syncOrder } from '../../server/lib/stripe/sync-order';
+import { Collective, Op, Order, PaymentMethod } from '../../server/models';
+import { runCronJob } from '../utils';
+
+async function run() {
+  logger.info('Starting job to fix stripe stale NEW paymentIntents');
+
+  const results = await Order.findAndCountAll({
+    limit: 20,
```
kewitz, line 16: We should either remove/increase the limit or implement pagination:
```suggestion
```

scripts/stripe/fix-stale-new-paymentIntents.ts
hdiniz: Added pager! Thanks

cron/hourly/20-fix-stripe-stale-NEW-paymentIntents.ts
hdiniz: Added pager, thanks!

cron/hourly/20-fix-stripe-stale-NEW-paymentIntents.ts
```diff
@@ -0,0 +1,85 @@
...
+    for (const order of page) {
+      await syncOrder(order, { IS_DRY: process.env.DRY, logging: logger.info });
+    }
+  }
+
+  logger.info(`Done!`);
+}
+
+if (require.main === module) {
+  runCronJob('fix-stripe-stale-NEW-paymentIntents', run, 60 * 60);
```
kewitz, line 84: 👍🏼

scripts/stripe/fix-stale-new-paymentIntents.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+}
+
+const main = async () => {
+  const pager = staleStripeNewPaymentIntentOrdersPager();
+
+  for await (const page of pager) {
+    console.log(`Processing ${page.length} orders...`);
+    for (const order of page) {
+      await syncOrder(order, { IS_DRY: process.env.DRY });
```
kewitz, line 74: A good practice is having DRY on by default:
```js
const IS_DRY = process.env.DRY !== 'false';
```

scripts/stripe/sync-paymentIntent-orders.ts
```diff
@@ -1,57 +1,18 @@
...
-};
 
 const main = async () => {
   const transactionIds = compact(process.argv.slice(2).map(toNumber));
   for (const id of transactionIds) {
     const order = await models.Order.findByPk(id, {
       include: [{ model: models.Collective, as: 'collective' }],
     });
-    await processOrder(order);
+    await syncOrder(order, { IS_DRY: process.env.DRY, logging: logger.info });
```
kewitz, line 15: A good practice is having DRY on by default:
```js
const IS_DRY = process.env.DRY !== 'false';
```

kewitz (approved):
LGTM.

cron/hourly/20-fix-stripe-stale-NEW-paymentIntents.ts
```diff
@@ -0,0 +1,85 @@
...
+    });
+
+    if (isEmpty(pageResult)) {
+      return;
+    }
+
+    yield pageResult;
+
+    offset += PAGE_SIZE;
+  }
```
Betree, line 67: This is not a safe way to paginate orders as fewer or more results will end up changing the results.

Example - take 5 orders:
- order 1
- order 2
- order 3
- order 4
- order 5

And say the pagination is "3". If nothing changes, you'll get [1,2,3] then [4,5]. So far so good.

Now imagine a new order "0" matches the conditions while you're processing order 1 (e.g. because it's `createdAt` just reached the condition threshold):
- order 0
- order 1
- order 2
- order 3
- order 4
- order 5

While your first page will be [1,2,3], the second query will now return [3,4,5] resulting in a double-processing of `3`.

Similar things can happen if orders get removed from the query, and given that `syncOrder` updates the status it will impact pagination.

PR #10321, opened 2024-09-20
Introduce Elastic Search

Betree:
Currently implemented as root-only to limit any security implications (since this doesn't have tests yet).

Stats with an old prod snapshot (we can expect the current one to be -/+ 20%):
image

Follow-up issues:
- Add tests
- Find a better way to store/represent amount and currency
- Fields weights (e.g. `slug^5`)
- CRON job to sync entries


gustavlrsn (approved):
Works great! Nice work wrapping this up into something useable!

I see another potential follow up; some ability to get individuals, collectives, projects, events etc in separate buckets, either through a filter on the accounts sub resolver or splitting `accounts` into various types of accounts.

Also perhaps something for "Hosted Collectives", it seems we get those through the `accounts` resolver, but a separate one might be useful to get directed to the Hosted Collectives dashboard for that collective (if the general account results link to the user profile). Same thing for "Contributors".

PR #10328, opened 2024-09-25
Add wise/check-fees refund command

kewitz:
Add refund command to create a compensation between Host and Collective to cover for negative differences in balance.

### DIFF
--- a/scripts/wise/check-fees.ts
+++ b/scripts/wise/check-fees.ts
@@ -3,21 +3,24 @@ import '../../server/env';
 import assert from 'assert';
 
 import { Command } from 'commander';
-import { omit, pick } from 'lodash';
+import { groupBy, omit, pick, round } from 'lodash';
 import moment from 'moment';
 
 import { Service } from '../../server/constants/connected-account';
 import { TransactionKind } from '../../server/constants/transaction-kind';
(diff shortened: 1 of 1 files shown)


scripts/wise/check-fees.ts
```diff
@@ -256,6 +259,114 @@ program.command('fix [hosts]').action(async hosts => {
...
+        }, 0);
+        if (totalDiff < 0) {
+          console.log(`\tRefunding ${collective.slug} for a total of ${totalDiff} ${currency}`);
+          const amountInHostCurrency = Math.abs(totalDiff * 100);
+          totalRefunded += amountInHostCurrency;
+          const hostCurrencyFxRate = await getFxRate(collective.currency, host.currency);
+          const amount = Math.round(amountInHostCurrency / hostCurrencyFxRate);
+          const transactionData = {
+            type: TransactionTypes.CREDIT,
+            kind: TransactionKind.PAYMENT_PROCESSOR_COVER,
```
znarf, line 318: Maybe ADDED_FUNDS or CONTRIBUTION would be better.

It's not expected to have  PAYMENT_PROCESSOR_COVER not part of a TransactionGroup and this can create issues and break some consistency checks.

znarf:
Note that Added Funds without a matching order may be breaking some expectations.

PR #10329, opened 2024-09-25
Wise: partial refunds Refactor

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7579


server/paymentProviders/transferwise/webhook.ts
```diff
@@ -117,42 +121,144 @@ export async function handleTransferStateChange(event: TransferStateChangeEvent)
...
     await expense.createActivity(activities.COLLECTIVE_EXPENSE_ERROR, null, { isSystem: true, event });
   }
 }
 
+const handleTransferRefund = async (event: TransferRefundEvent): Promise<void> => {
+  const transferId = event.data.resource.id;
+  const refundWiseEventTimestamp = event.data.occurred_at;
+  const expense = await models.Expense.findOne({
+    where: {
+      status: [expenseStatus.PROCESSING, expenseStatus.PAID],
```
kewitz, line 137: Race condition with state-change event above.

server/paymentProviders/transferwise/webhook.ts
```diff
@@ -117,42 +121,144 @@ export async function handleTransferStateChange(event: TransferStateChangeEvent)
...
+      data: { transfer: { id: transferId } },
```
kewitz, line 138: Make this index unique.

Currently:
```sql
CREATE INDEX expenses__wise_transfer_id
  ON "Expenses" ((data #>> '{transfer,id}'::text[]) DESC)
  WHERE ((data #>> '{transfer,id}'::text[]) IS NOT NULL);
```

server/paymentProviders/transferwise/webhook.ts
```diff
@@ -117,42 +121,144 @@ export async function handleTransferStateChange(event: TransferStateChangeEvent)
...
+      { model: models.User, as: 'User' },
+      { model: models.Transaction },
+    ],
+  });
+
+  if (!expense) {
+    // This is probably some other transfer not executed through our platform.
+    logger.warn('Could not find related Expense, ignoring transferwise event.', event);
+    return;
+  } else if (expense.data.refundEventTimestamp === refundWiseEventTimestamp) {
```
kewitz, line 156:
```suggestion
  } else if (expense.data.refundWiseEventTimestamp === refundWiseEventTimestamp) {
```

server/paymentProviders/transferwise/webhook.ts
```diff
@@ -117,42 +121,144 @@ export async function handleTransferStateChange(event: TransferStateChangeEvent)
...
-    }
+  } else if (expense.status === expenseStatus.PROCESSING && event.data.current_state === 'cancelled') {
+    logger.info(`Wise: Transfer failed, setting status to error.`, event);
     await expense.update({ data: { ...expense.data, transfer } });
     await expense.setError(expense.lastEditedById);
     await expense.createActivity(activities.COLLECTIVE_EXPENSE_ERROR, null, { isSystem: true, event });
   }
 }
 
+const handleTransferRefund = async (event: TransferRefundEvent): Promise<void> => {
```
kewitz, line 132: Lock the expense.

server/paymentProviders/transferwise/webhook.ts
```diff
@@ -117,42 +121,144 @@ export async function handleTransferStateChange(event: TransferStateChangeEvent)
...
+      const difference = sourceAmount - refundedAmount;
+      const paymentProcessorFeeInHostCurrency = difference * 100;
+      const hostCurrencyFxRate = await models.Transaction.getFxRate(collective.currency, hostCurrency);
+      await models.Transaction.createPaymentProcessorFeeTransactions({
+        amount: 0,
+        paymentProcessorFeeInHostCurrency,
+        currency: collective.currency,
+        hostCurrencyFxRate,
+        hostCurrency,
+        CollectiveId: expense.CollectiveId,
```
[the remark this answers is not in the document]
kewitz: Ben says Host.

server/paymentProviders/transferwise/webhook.ts
```diff
@@ -117,42 +121,144 @@ export async function handleTransferStateChange(event: TransferStateChangeEvent)
...
+      await createRefundTransaction(
+        creditTransaction,
+        paymentProcessorFee * 100,
+        pick(creditTransaction.data, ['transfer']),
+        expense.User,
+      );
+    } else if (refundedAmount < sourceAmount) {
+      logger.verbose('Wise: Paid Expense was partially refunded', event);
+      const difference = sourceAmount - refundedAmount;
+      const paymentProcessorFee = expense.data.paymentOption.fee.total;
```
kewitz, line 196: Can we end up with a difference greater than the payment processing fee?
Maybe when the `feesPayer = PAYEE`.

server/paymentProviders/transferwise/webhook.ts
```diff
@@ -117,42 +121,144 @@ export async function handleTransferStateChange(event: TransferStateChangeEvent)
...
+        creditTransaction,
+        Math.round((paymentProcessorFee - difference) * 100),
+        pick(creditTransaction.data, ['transfer']),
+        expense.User,
+      );
+    }
+
+    await expense.setError(expense.lastEditedById);
+    await expense.createActivity(activities.COLLECTIVE_EXPENSE_ERROR, null, { isSystem: true, event });
+    await expense.update({ data: { ...expense.data, refundWiseEventTimestamp } });
```
kewitz, line 207: Persist the refund event as well.

test/server/paymentProviders/transferwise/webhook.test.ts
```diff
@@ -1,197 +1,643 @@
...
+    // it('should mark expense as paid and create transactions if transfer was sent', async () => {
+    //   await api.post('/webhooks/transferwise').send(event).expect(200);
+
+    //   await expense.reload();
+    //   expect(expense).to.have.property('status', status.PAID);
+    //   const [debitTransaction] = await expense.getTransactions({ where: { type: 'DEBIT' } });
+    //   expect(debitTransaction).to.be.have.property('paymentProcessorFeeInHostCurrency', -1000);
+    //   expect(debitTransaction).to.be.have.property('netAmountInCollectiveCurrency', -11000);
+    //   expect(debitTransaction).to.be.have.nested.property('data.transfer.id', 1234);
+    // });
```
kewitz, line 356:
```suggestion
```

test/server/paymentProviders/transferwise/webhook.test.ts
```diff
@@ -1,197 +1,643 @@
...
+
+      await api.post('/webhooks/transferwise').send(event).expect(200);
+      await api.post('/webhooks/transferwise').send(event).expect(200);
+
+      await expense.reload({ include: [{ model: models.Transaction }] });
+      expect(expense).to.have.property('status', status.ERROR);
+      expect(expense).to.have.nested.property('data.transfer.id', event.data.resource.id);
+      expect(expense).to.have.nested.property('data.refundWiseEventTimestamp', event.data.occurred_at);
+      expect(expense.Transactions).to.have.length(2);
     });
```
kewitz, line 641: Add a test case for `feesPayer = PAYEE`.

server/paymentProviders/transferwise/webhook.ts
kewitz, line 207: Check if the expense isn't already set to ERROR.

server/paymentProviders/transferwise/index.ts
```diff
@@ -788,25 +789,41 @@ const oauth = {
   },
 };
 
-async function setUpWebhook(): Promise<Webhook> {
-  const url = `${config.host.api}/webhooks/transferwise`;
+const APPLICATION_TRIGGERS = ['transfers#state-change', 'transfers#refund'] as const;
+
+async function createWebhooksForHost(host = config.host.api): Promise<Webhook> {
```
Betree, line 794: I don't see where we're using this function?

Depending on how we use it, `host` seems like a dangerous parameter. Why not just always use `config.host.api` , eventually mocking the config in tests if it what it's for?

server/paymentProviders/transferwise/index.ts
```diff
@@ -788,25 +789,41 @@ const oauth = {
...
   const existingWebhooks = await transferwise.listApplicationWebhooks();
 
-  if (existingWebhooks?.find(w => w.trigger_on === 'transfers#state-change' && w.delivery.url === url)) {
-    logger.info(`TransferWise App Webhook already exists for ${url}.`);
-    return;
+  for (const trigger_on of APPLICATION_TRIGGERS) {
+    if (existingWebhooks?.find(w => w.trigger_on === trigger_on && w.delivery.url === url)) {
+      logger.info(`TransferWise App Webhook already exists for ${url}.`);
+      return;
+    }
```
Betree, line 802: `continue` rather than `return`, in case one event has a webhook setup but not the other?

```suggestion
    if (existingWebhooks?.find(w => w.trigger_on === trigger_on && w.delivery.url === url)) {
      logger.info(`TransferWise App Webhook already exists for ${url}.`);
      continue;
    }
```

Betree (commented):
Need to look at the tests but looks good so far. Let's sync with OSC when we deploy/enable the new webhooks to try to not do it in the middle of a paying session. It would actually be great to have a mechanism to put Wise payments under "maintenance" and have that reflected in the UI so that we can make breaking changes comfortably.

server/paymentProviders/transferwise/index.ts
kewitz: There's a part of this PR that was stashed and contained a helper tool to mount and dismount webhooks using ngrok.

But you're right, we can lock this to `config.host.api`.

PR #10332, opened 2024-09-26
Unify platform constants

Betree:

PR #10338, opened 2024-09-27
chore: Add script to migrate OCI to OFT

Betree:


server/constants/platform.ts
```diff
@@ -10,13 +10,20 @@ const MIGRATION_DATE = moment('2024-10-01T00:00:00Z');
  * Returns the platform constants based on a function to check if the platform has been migrated.
  */
 const getPlatformConstants = (checkIfMigrated: () => boolean) => ({
+  get OCICollectiveId() {
+    return 8686;
+  },
+
+  get OFTCollectiveId() {
+    return 835523;
```
kewitz, line 18: Actually, this is the OFiCo id.
OFiTech is 845576.

kewitz (commented):
Do we have any plans for the host-settlement cron job?

server/constants/platform.ts
Betree: Yep, I corrected that in https://github.com/opencollective/opencollective-api/pull/10332/files#diff-74edb6d2812ca785f8d5de71cf8106d23b666692c4821b40535f2c089a7477f8 and didn't bother to do it here, but it won't hurt

Betree:
```suggestion
    return 845576;
```

Betree:
> Do we have any plans for the host-settlement cron job?

The host settlement job is being deactivated for November+ in https://github.com/opencollective/opencollective-api/pull/10332. We'll make a follow-up PR to make it work for Ofitech.

PR #10349, opened 2024-10-02
Add (experimental) transaction groups to API

gustavlrsn:
For https://github.com/opencollective/opencollective/issues/7534

### DIFF
--- a/server/graphql/v2/collection/TransactionGroupCollection.ts
+++ b/server/graphql/v2/collection/TransactionGroupCollection.ts
@@ -10,7 +10,7 @@ export const GraphQLTransactionGroupCollection = new GraphQLObjectType({
   fields: () => ({
     ...CollectionFields,
     nodes: {
-      type: new GraphQLList(GraphQLTransactionGroup),
+      type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(GraphQLTransactionGroup))),
     },
   }),
 });
--- a/server/graphql/v2/collection/TransactionGroupCollection.ts
+++ b/server/graphql/v2/collection/TransactionGroupCollection.ts
@@ -1,0 +1,16 @@
+import { GraphQLList, GraphQLObjectType } from 'graphql';
+
+import { CollectionFields, GraphQLCollection } from '../interface/Collection';
+import { GraphQLTransactionGroup } from '../object/TransactionGroup';
+
+export const GraphQLTransactionGroupCollection = new GraphQLObjectType({
+  name: 'TransactionGroupCollection',
+  interfaces: [GraphQLCollection],
+  description: 'A collection of Transactions groups',
+  fields: () => ({
+    ...CollectionFields,
+    nodes: {
+      type: new GraphQLList(GraphQLTransactionGroup),
+    },
+  }),
+});
--- a/server/graphql/v2/collection/TransactionGroupCollection.ts
+++ b/server/graphql/v2/collection/TransactionGroupCollection.ts
@@ -1,4 +1,4 @@
-import { GraphQLList, GraphQLObjectType } from 'graphql';
+import { GraphQLList, GraphQLNonNull, GraphQLObjectType } from 'graphql';
 
 import { CollectionFields, GraphQLCollection } from '../interface/Collection';
 import { GraphQLTransactionGroup } from '../object/TransactionGroup';
--- a/server/graphql/v2/object/TransactionGroup.ts
+++ b/server/graphql/v2/object/TransactionGroup.ts
@@ -15,7 +15,7 @@ export const GraphQLTransactionGroup = new GraphQLObjectType({
       id: {
         type: new GraphQLNonNull(GraphQLString),
       },
-      amount: {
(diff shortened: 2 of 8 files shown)


server/graphql/v2/collection/TransactionGroupCollection.ts
```diff
@@ -0,0 +1,16 @@
...
+import { GraphQLTransactionGroup } from '../object/TransactionGroup';
+
+export const GraphQLTransactionGroupCollection = new GraphQLObjectType({
+  name: 'TransactionGroupCollection',
+  interfaces: [GraphQLCollection],
+  description: 'A collection of Transactions groups',
+  fields: () => ({
+    ...CollectionFields,
+    nodes: {
+      type: new GraphQLList(GraphQLTransactionGroup),
```
Betree, line 13:
```suggestion
      type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(GraphQLTransactionGroup))),
```

server/graphql/v2/object/TransactionGroup.ts
```diff
@@ -0,0 +1,54 @@
...
+  name: 'TransactionGroup',
+  description: 'Transaction group',
+  fields: () => {
+    return {
+      id: {
+        type: new GraphQLNonNull(GraphQLString),
+      },
+      amount: {
+        type: new GraphQLNonNull(GraphQLAmount),
+      },
```
Betree, line 20: It's unclear what "amount" would be for a transaction group. Since it contains DEBITs and CREDITs, the sum of all transactions in a group would normally be 0. If we intend to filter in the frontend, I'd suggest renaming this to `totalAmount` to make it clearer that it's a sum.

Since this is a beta query, I'm only thinking out loud in these comments - this is not a requested change.

server/graphql/v2/query/TransactionGroupQuery.ts
```diff
@@ -0,0 +1,65 @@
...
+        TransactionGroup: args.groupId,
+      },
+      order: [[sequelize.literal(getTransactionKindPriorityCase('Transaction')), 'ASC']],
+    });
+
+    const primaryTransaction = transactions[0]; // First transaction based on the ordering using getTransactionKindPriorityCase
+    const convertedAmounts = await Promise.all(
+      transactions.map(async t => {
+        const fxRate = await getFxRate(t.currency, account.currency, t.createdAt);
+        return Math.round(t.netAmountInCollectiveCurrency * fxRate);
```
Betree, line 45: Loaders will make this faster

```suggestion
        const fxRate = await getFxRate(t.currency, account.currency, t.createdAt);
        return Math.round(t.netAmountInCollectiveCurrency * fxRate);
```

```suggestion
        return req.loaders.CurrencyExchangeRate.convert.load({
          amount: t.netAmountInCollectiveCurrency,
          fromCurrency: t.currency,
          toCurrency: account.currency
          date: t.createdAt
        });
```

server/graphql/v2/query/collection/TransactionGroupCollectionQuery.ts
```diff
@@ -0,0 +1,182 @@
...
+    where.push({
+      [Op.and]: sequelize.literal(
+        `EXISTS (
+            SELECT 1 FROM "Transactions" t4
+            WHERE t4."id" = (${primaryTransactionIdSubquery})
+              AND t4."kind" = '${args.kind}'
+          )`,
+      ),
+    });
+  }
```
Betree, line 92: We need to check `"deletedAt" = NULL` in all those subqueries

Betree (commented):
Good stuff, some suggestions below but no blocking feedback. Except maybe the "deletedAt` thing.

server/graphql/v2/object/TransactionGroup.ts
gustavlrsn: Since we're only fetching the transactions in a group on a specific account (and not all the transactions in the group) the amount will not be 0. This is at least the current implementation, with the required `account` argument in the queries.

`totalAmount` makes sense, or possibly `netAmount`?

server/graphql/v2/query/TransactionGroupQuery.ts
gustavlrsn: The loader does not have the date option

server/graphql/v2/object/TransactionGroup.ts
Betree: `netAmount` could be confusing as we use this term in many contexts to signify "amount without fees / taxes".

server/graphql/v2/query/TransactionGroupQuery.ts
Betree: You're right, the `convert` loaders doesn't support date. The `CurrencyExchangeRate.fxRate` one does though.
Not a change we need to do now, but surfacing that as a way to optimize the query later if needed.

gustavlrsn: Ah, nice, will include it!

PR #10352, opened 2024-10-03
chore(Transactions): tweak platform tips description

Betree:


znarf (approved):
That's a good first step! Maybe we can then propose to the team what they think about having something like "Tip to the Open Collective Platform"

PR #10372, opened 2024-10-14
fix(HostReport): add deprecation message

Betree:
Part of https://github.com/opencollective/opencollective/issues/7622

**New default:**
image

**In case we need to manually generate a legacy report:**
image


znarf:
Let's make sure that those reports are generally released and not behind a feature flag before activating this warning.

PR #10388, opened 2024-10-18
feat(TransactionsImport): add parameters for pipeline

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7588
Resolve https://github.com/opencollective/opencollective/issues/7587

### DIFF
--- a/server/graphql/v2/mutation/TransactionImportsMutations.ts
+++ b/server/graphql/v2/mutation/TransactionImportsMutations.ts
@@ -3,7 +3,7 @@ import type { Request } from 'express';
 import { GraphQLBoolean, GraphQLList, GraphQLNonNull } from 'graphql';
 import { GraphQLJSONObject, GraphQLNonEmptyString } from 'graphql-scalars';
 import GraphQLUpload from 'graphql-upload/GraphQLUpload.js';
-import { isUndefined, omit, omitBy, pick } from 'lodash';
+import { omit, pick } from 'lodash';
 
 import { disconnectPlaidAccount } from '../../../lib/plaid/connect';
 import RateLimit from '../../../lib/rate-limit';
@@ -17,7 +17,10 @@ import { fetchAccountWithReference, GraphQLAccountReferenceInput } from '../inpu
 import { getValueInCentsFromAmountInput } from '../input/AmountInput';
 import { getDatabaseIdFromOrderReference } from '../input/OrderReferenceInput';
 import { GraphQLTransactionsImportRowCreateInput } from '../input/TransactionsImportRowCreateInput';
-import { GraphQLTransactionsImportRowUpdateInput } from '../input/TransactionsImportRowUpdateInput';
+import {
+  GraphQLTransactionsImportRowUpdateInput,
+  TransactionImportRowGraphQLType,
+} from '../input/TransactionsImportRowUpdateInput';
 import { GraphQLTransactionsImport } from '../object/TransactionsImport';
 
 const transactionImportsMutations = {
@@ -206,7 +209,16 @@ const transactionImportsMutations = {
         description: 'Whether to restore all dismissed rows',
       },
     },
-    resolve: async (_: void, args, req: Request) => {
+    resolve: async (
+      _: void,
+      args: {
+        id: string;
+        rows?: TransactionImportRowGraphQLType[];
+        dismissAll?: boolean;
+        restoreAll?: boolean;
+      },
+      req: Request,
+    ) => {
       checkRemoteUserCanUseTransactions(req);
       const importId = idDecode(args.id, 'transactions-import');
       const transactionsImport = await TransactionsImport.findByPk(importId, {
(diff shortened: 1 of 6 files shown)


server/graphql/v2/object/TransactionsImport.ts
```diff
@@ -86,15 +89,54 @@ export const GraphQLTransactionsImport = new GraphQLObjectType({
...
+        importInstance,
+        args: { limit: number; offset: number; status: TransactionsImportRowStatus; searchTerm: string },
+      ) => {
+        const where: Parameters<typeof TransactionsImportRow.findAll>[0]['where'] = {
+          [Op.and]: [{ TransactionsImportId: importInstance.id }],
+        };
+
+        // Filter by status
+        if (args.status) {
+          if (args.status === 'IGNORED') {
```
kewitz, line 113: 🍨 Use the enum.

server/graphql/v2/object/TransactionsImport.ts
```diff
@@ -86,15 +89,54 @@ export const GraphQLTransactionsImport = new GraphQLObjectType({
...
+        searchTerm: {
+          type: GraphQLString,
+          description: 'Search by text',
+        },
+      },
+      resolve: async (
+        importInstance,
+        args: { limit: number; offset: number; status: TransactionsImportRowStatus; searchTerm: string },
+      ) => {
+        const where: Parameters<typeof TransactionsImportRow.findAll>[0]['where'] = {
```
kewitz, line 107: Something to work on some other time: make a generic to extract the "where" from any model.

server/graphql/v2/mutation/TransactionImportsMutations.ts
```diff
@@ -213,35 +221,52 @@ const transactionImportsMutations = {
...
-              const orderId = getDatabaseIdFromOrderReference(row.order);
-              const order = await req.loaders.Order.byId.load(orderId);
-              const collective = order && (await req.loaders.Collective.byId.load(order.CollectiveId));
-              if (!order || !collective || collective.HostCollectiveId !== transactionsImport.CollectiveId) {
-                throw new Unauthorized(`Order not found or not associated with the import: ${orderId}`);
+        if (args.rows?.length) {
+          await Promise.all(
+            args.rows.map(async row => {
+              const rowId = idDecode(row.id, 'transactions-import-row');
+              const values = omitBy(omit(row, ['id', 'order']), isUndefined);
```
kewitz, line 228: Ideally we create an array of editable fields and use `pick` here.

server/graphql/v2/mutation/TransactionImportsMutations.ts
```diff
@@ -213,35 +221,52 @@ const transactionImportsMutations = {
...
+              transaction,
+            },
+          );
+        } else if (args.restoreAll) {
+          await TransactionsImportRow.update(
+            { isDismissed: false },
+            { where: { TransactionsImportId: importId, isDismissed: true }, transaction },
+          );
+        } else {
+          throw new ValidationFailed('You must provide at least one row to update or dismiss all rows');
```
kewitz, line 268: Not true, right? You can dismiss/restore all without passing rows.

server/graphql/v2/object/TransactionsImport.ts
```diff
@@ -86,15 +89,54 @@ export const GraphQLTransactionsImport = new GraphQLObjectType({
...
+        if (args.searchTerm) {
+          where[Op.and].push({
+            [Op.or]: buildSearchConditions(args.searchTerm, {
+              textFields: ['description', 'sourceId'],
+            }),
+          });
+        }
+
         return {
           offset: 0,
```
kewitz, line 132:
```suggestion
          offset: args.offset,
```

kewitz (approved):
Looking good, got a few points you may want to check but feel free to ship how you see fit.

server/graphql/v2/object/TransactionsImport.ts
Betree: I don't have a strong opinion on that one. `args.status` is typed, so if you try to do something like `args.status === 'xxx'` typescript will error, because it knows the possible values for it. Knowing this, what's the added value of using the enum?

server/graphql/v2/object/TransactionsImport.ts
Betree: Yep, I agree this is unnecessarily verbose, there may be a better way to do it

server/graphql/v2/mutation/TransactionImportsMutations.ts
Betree: It's missing "restore" indeed

server/graphql/v2/object/TransactionsImport.ts
Betree: Good catch!

server/graphql/v2/mutation/TransactionImportsMutations.ts
Betree: It's not new, but I agree `pick` is safer than `omit`. Will fix that in a separate PR.

PR #10422, opened 2024-10-25
deps: Upgrade to ESLint v9

Betree:


Betree:
Migration reference: https://github.com/opencollective/opencollective-taxes/pull/350

PR #10432, opened 2024-10-29
Review total raised and spent stats

znarf:


migrations/20241029071652-collective-transaction-stats-fixes.js
```diff
@@ -0,0 +1,338 @@
...
+      CREATE MATERIALIZED VIEW "CollectiveTransactionStats" AS
+
+        WITH "ActiveCollectives" AS (
+          SELECT c."id" as "CollectiveId"
+          FROM "Transactions" t
+          LEFT JOIN "Collectives" c ON c."id" = t."CollectiveId" AND c."deletedAt" IS NULL
+          WHERE t."deletedAt" IS NULL
+            AND t."hostCurrency" IS NOT NULL
+            AND c."deactivatedAt" IS NULL
+            -- TODO: not sure why we have those conditions omn guest/incognito, too many accounts?
```
Betree, line 20:
```suggestion
            -- TODO: not sure why we have those conditions on guest/incognito, too many accounts?
```

Betree: I think it's to avoid computing metrics for them, since we know we'll never display them? Not sure if the optimization is worth it though.

server/lib/budget.js
```diff
@@ -634,8 +636,11 @@ export async function sumCollectivesTransactions(
       if (['netAmountInCollectiveCurrency', 'netAmountInHostCurrency'].includes(column)) {
         where[Op.and].push({
           [Op.or]: [
-            { type: DEBIT, [Op.not]: { kind: ['HOST_FEE', 'PAYMENT_PROCESSOR_FEE'], OrderId: { [Op.not]: null } } },
-            { type: CREDIT, kind: 'PAYMENT_PROCESSOR_COVER' },
+            {
+              type: DEBIT,
+              [Op.not]: { [Op.and]: { kind: ['HOST_FEE', 'PAYMENT_PROCESSOR_FEE'], OrderId: { [Op.not]: null } } },
```
Betree, line 641: `[Op.and]` is already the default, right? 

```suggestion
              [Op.not]: { kind: ['HOST_FEE', 'PAYMENT_PROCESSOR_FEE'], OrderId: { [Op.not]: null } },
```

No requested change, just want to make sure we're on the same page for this line.

server/lib/budget.js
```diff
@@ -646,8 +651,19 @@ export async function sumCollectivesTransactions(
...
-          { type: CREDIT, [Op.not]: { kind: 'PAYMENT_PROCESSOR_COVER', OrderId: null } },
+          // Get all credits except PAYMENT_PROCESSOR_COVER from expenses or before separate fees
+          {
+            type: CREDIT,
+            [Op.not]: [
+              {
+                kind: 'PAYMENT_PROCESSOR_COVER',
+                [Op.or]: {
+                  OrderId: null,
+                  ...(separatePaymentProcessorFees && { createdAt: { [Op.lt]: '2024-01-01' } }),
```
Betree, line 662: :+1:

Betree (approved):

server/lib/budget.js
znarf: It was necessary to get the parenthesis right in the condition.

PR #10447, opened 2024-11-04
Add new expense policy object and allow comment when creating expense

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7606


gustavlrsn (approved):

migrations/20241103183508-migrate-expense-policies-to-policies.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+      UPDATE "Collectives" SET
+      data = jsonb_set(
+        data,
+        '{policies,EXPENSE_POLICIES}',
```
Betree, line 10: We need a `COALESCE` here, as `JSONB_SET(null, ...)` gives `NULL`. We'd be dropping existing policies.

```sql
SELECT COUNT(*)
FROM "Collectives"
WHERE TRIM("expensePolicy") <> '' AND "expensePolicy" IS NOT NULL
AND data IS NULL
```

=> 126 entries

migrations/20241103183508-migrate-expense-policies-to-policies.js
```diff
@@ -0,0 +1,34 @@
...
+        jsonb_build_object(
+          'invoicePolicy', "expensePolicy",
+          'receiptPolicy', "expensePolicy",
+          'titlePolicy', ''
+        )
+      )
+      WHERE TRIM("expensePolicy") <> '' AND "expensePolicy" IS NOT NULL;
```
Betree, line 17: In production, running the equivalent SELECT gives me an error:

```sql
SELECT  jsonb_set(
        data,
        '{policies,EXPENSE_POLICIES}',
        jsonb_build_object(
          'invoicePolicy', "expensePolicy",
          'receiptPolicy', "expensePolicy",
          'titlePolicy', ''
        )
      )
FROM "Collectives"
WHERE TRIM("expensePolicy") <> '' AND "expensePolicy" IS NOT NULL;
```

> [22P02] ERROR: path element at position 2 is not an integer: "EXPENSE_POLICIES"

PR #10452, opened 2024-11-06
Add lock on order when processing payment intent succeeded

hdiniz:


server/paymentProviders/stripe/webhook.ts
```diff
@@ -167,56 +168,98 @@ export const paymentIntentSucceeded = async (event: Stripe.Event) => {
...
+
+    if (order) {
+      await order.update({
+        data: previousOrderData,
+      });
+    }
+
+    if (sequelizeTransaction) {
+      sequelizeTransaction.rollback();
+    }
```
Betree, line 259: Two things here:
1. Should we rather rollback before we update?
2. It looks like we may end up calling `rollback` on an already `commit`ed transaction (e.g. if `createOrUpdateOrderStripePaymentMethod` fails). What would be the effect of that?

hdiniz: 1. You're right, its best to put that rollback first to avoid deadlocking.
2. `sequelizeTransaction` is set to null right after its committed so this would be skipped.

Betree (approved):

Betree:
@copilot+gpt-5.6-luna:high There's no context here, can you summarize what it's trying to solve and whether it's still necessary with the latest version of the code we have in `main`?

PR #10468, opened 2024-11-15
Check balance before issuing a refund

kewitz:


server/graphql/common/transactions.ts
```diff
@@ -233,6 +233,14 @@ export async function refundTransaction(
     }
   }
 
+  // Check if the hosted collective has enough funds to refund the transaction
+  if (!args?.ignoreBalanceCheck && collective && collective.HostCollectiveId && transaction.type === 'CREDIT') {
```
Betree, line 237: Maybe only allow this option if `req.remoteUser` is a host admin?

Betree (approved):

PR #10471, opened 2024-11-18
Add env variable to disable timeline resolver

kewitz:


server/lib/timeline.ts
```diff
@@ -242,11 +243,17 @@ export const getCollectiveFeed = async ({
   limit: number;
   classes: ActivityClasses[];
 }) => {
+  const isDisabled = parseToBoolean(config.timeline.disabled);
+  if (isDisabled) {
+    return [];
```
Betree, line 248: Maybe return `null`, so that the frontend gets a "ContentNotReady"?

PR #10474, opened 2024-11-19
Introduce daily materialized views

znarf:
These materialized views are a bit data intensive and it's likely not necessary to update them so frequently as they're ultimately used by dynamic views.

### DIFF
--- a/cron/daily/92-refresh-daily-materialized-views.js
+++ b/cron/daily/92-refresh-daily-materialized-views.js
@@ -4,9 +4,7 @@ import logger from '../../server/lib/logger';
 import { sequelize } from '../../server/models';
 import { runCronJob } from '../utils';
 
-const VIEWS = ['CollectiveOrderStats', 'ExpenseTagStats'];
-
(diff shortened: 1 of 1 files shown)


cron/daily/92-refresh-daily-materialized-views.js
```diff
@@ -0,0 +1,35 @@
...
+    try {
+      await sequelize.query(`REFRESH MATERIALIZED VIEW CONCURRENTLY "${view}"`);
+      const [runSeconds, runMilliSeconds] = process.hrtime(startTime);
+      logger.info(`${view} materialized view refreshed in ${runSeconds}.${runMilliSeconds} seconds`);
+    } catch (e) {
+      const [runSeconds, runMilliSeconds] = process.hrtime(startTime);
+      logger.error(
+        `Error while refreshing ${view} materialized view after ${runSeconds}.${runMilliSeconds} seconds: ${e.message}`,
+        e,
+      );
```
Betree, line 28: Not a requested change since it's not new, but it would be nice to report these to Sentry to make sure the error reports trigger a notification & don't get lost. I can iterate on that in a follow-up PR.

Betree (approved):

PR #10477, opened 2024-11-20
Draft Expenses: Optional invite notification and draft key

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7671

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -29,7 +29,7 @@ import {
 import moment from 'moment';
 import { v4 as uuid } from 'uuid';
 
-import { activities, roles } from '../../constants';
+import { activities, expenseStatus, roles } from '../../constants';
 import ActivityTypes from '../../constants/activities';
 import { CollectiveType } from '../../constants/collectives';
 import { Service } from '../../constants/connected-account';
@@ -394,6 +394,14 @@ export const canSeeExpenseSecurityChecks: ExpensePermissionEvaluator = async (re
   }
 
   return remoteUserMeetsOneCondition(req, expense, [isHostAdmin]);
+};
+
+export const canSeeDraftKey: ExpensePermissionEvaluator = async (req, expense) => {
+  if (expense.status !== expenseStatus.DRAFT) {
+    return false;
+  }
+
+  return canSeeExpenseSecurityChecks(req, expense);
 };
 
 export const canSeeExpenseCustomData: ExpensePermissionEvaluator = async (req, expense) => {
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -401,7 +401,17 @@ export const canSeeDraftKey: ExpensePermissionEvaluator = async (req, expense) =
     return false;
   }
 
-  return canSeeExpenseSecurityChecks(req, expense);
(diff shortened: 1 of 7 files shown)


server/graphql/v2/mutation/ExpenseMutations.ts
```diff
@@ -455,6 +455,11 @@ const expenseMutations = {
         type: new GraphQLNonNull(GraphQLAccountReferenceInput),
         description: 'Account where the expense will be created',
       },
+      skipInvite: {
+        type: GraphQLBoolean,
```
Betree, line 459: Given this has a default value and we later check it with `!args.skipInvite`, I'd make it non-nullable otherwise passing `null` would act as `false`

```suggestion
        type: new GraphQLNonNull(GraphQLBoolean),
```

server/graphql/v2/mutation/ExpenseMutations.ts
```diff
@@ -528,7 +533,9 @@ const expenseMutations = {
         status: expenseStatus.DRAFT,
       });
 
-      await sendDraftExpenseInvite(req, expense, collective, draftKey);
+      if (!args.skipInvite) {
+        await sendDraftExpenseInvite(req, expense, collective, draftKey);
+      }
```
Betree, line 538: One issue s that `sendDraftExpenseInvite` also takes care of recording the activity, which is important for the activity log / audit trail. I'd suggest setting `activity.data.notify` to false, see https://github.com/opencollective/opencollective-api/blob/ff571d0b0ae40fca506387b4e507119e40d2ab80/server/models/Activity.ts#L109. Also make sure there's no email reminder CRON?

server/graphql/common/expenses.ts
```diff
@@ -396,6 +396,14 @@ export const canSeeExpenseSecurityChecks: ExpensePermissionEvaluator = async (re
   return remoteUserMeetsOneCondition(req, expense, [isHostAdmin]);
 };
 
+export const canSeeDraftKey: ExpensePermissionEvaluator = async (req, expense) => {
+  if (expense.status !== expenseStatus.DRAFT) {
+    return false;
+  }
+
+  return canSeeExpenseSecurityChecks(req, expense);
```
Betree, line 404: I see a risk in defining such a dependency, as developers iterating on `canSeeExpenseSecurityChecks` may not be aware of it. I think this is a case for "explicit" is better than "re-usable", a copy-paste wouldn't hurt.

Plus, security checks are limited to trusted hosts, I don't see why we would prevent this one for other hosts.

server/graphql/v2/object/Expense.ts
```diff
@@ -634,6 +634,15 @@ export const GraphQLExpense = new GraphQLObjectType<ExpenseModel, express.Reques
...
       },
+      draftKey: {
+        type: GraphQLString,
+        description: '[Host Admin only] Key to access the draft of this expense',
+        async resolve(expense, _, req) {
+          if (await ExpenseLib.canSeeDraftKey(req, expense)) {
+            return expense.data?.draftKey;
+          }
+        },
+      },
```
Betree, line 645: Given how sensitive it is, we should add tests in `test/server/graphql/v2/query/ExpenseQuery.test.js` to guarantee that only host admins can see it

PR #10482, opened 2024-11-22
Rewrite generate jwt debug script

kewitz:
It now can:
- [x] Work with a .env sufix file `npm run script scripts/generate-jwt.ts id 123 prod`
- [x] Generate JWT for a particular email


scripts/generate-jwt.ts
```diff
@@ -0,0 +1,52 @@
...
+
+program.command('email <email> [env]').action(async email => {
+  const user = await models.User.findOne({ where: { email: email } });
+
+  if (!user) {
+    console.error(`User with email ${email} not found`);
+    process.exit(1);
+  }
+
+  const jwt = await user.generateSessionToken({ createActivity: false, updateLastLoginAt: false });
```
Betree, line 44: :ice_cream: to make sure these tokens don't stay valid for too long

```suggestion
  const jwt = await user.generateSessionToken({ createActivity: false, updateLastLoginAt: false, expiration: 60 * 60 * 24 });
```

Betree (approved):

PR #10484, opened 2024-11-22
Member checks fixes

Betree:


checks/model/members.js
```diff
@@ -70,7 +70,7 @@ async function checkDuplicateMembers({ fix = false } = {}) {
     `SELECT ARRAY_AGG(DISTINCT m2.id) AS duplicate_ids
      FROM "Members" m1
      INNER JOIN "Members" m2
-       ON m1.id != m2.id
+       ON m1.id < m2.id
```
Betree, line 73: @znarf  There was indeed a mistake here that made us delete both the original and the duplicate, I've added a migration to restore them all.

znarf:
Sorry, I merged instead of approving

PR #10490, opened 2024-11-25
record orderId in Stripe

xdamman:
For some payments the `orderId` is record (https://github.com/opencollective/opencollective-api/blob/0e84ea5b424c2d14b04b4652a585248fdfc35ada/server/paymentProviders/stripe/bacsdebit.ts#L33) but not for credit card payments (https://github.com/opencollective/opencollective-api/blob/0e84ea5b424c2d14b04b4652a585248fdfc35ada/server/paymentProviders/stripe/creditcard.ts#L57)

This fixes it.

Not sure if this mutation is used anywhere: https://github.com/opencollective/opencollective-api/blob/0e84ea5b424c2d14b04b4652a585248fdfc35ada/server/graphql/v2/mutation/OrderMutations.js#L1058

If it is used, it would be great to also make sure it records the `orderId` in the Stripe Metadata (but this requires more digging that I'll leave to people more familiar with the code)

### DIFF
--- a/server/paymentProviders/stripe/creditcard.ts
+++ b/server/paymentProviders/stripe/creditcard.ts
@@ -57,6 +57,7 @@ const createChargeAndTransactions = async (
       metadata: {
         from: `${config.host.website}/${order.fromCollective.slug}`,
         to: `${config.host.website}/${order.collective.slug}`,
+        orderId: order.id,
       },
     };
     // We don't add a platform fee if the host is the root account


Betree (approved):
Looks like a good change for consistency, thanks!

xdamman:
Is this used somewhere? It looks like it. 

https://github.com/opencollective/opencollective-api/blob/0e84ea5b424c2d14b04b4652a585248fdfc35ada/server/graphql/v2/mutation/OrderMutations.js#L1058-L1061

Can someone dig into this and figure out how to add the `orderId`?

Betree:
@xdamman We can't add it here since payment intents are created before orders. Can you add some details on the problem you're looking to solve, maybe there's another solution?

PR #10492, opened 2024-11-26
fix(Account): disable tips for platform accounts

Betree:


znarf (approved):
Would be great to transition this property to `platformTipAvailable` one day.

Betree:
> Would be great to transition this property to `platformTipAvailable` one day.

Added to https://github.com/opencollective/opencollective/issues/5049

PR #10494, opened 2024-11-26
feat: Upload images when creating accounts

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7647

Also includes a few enhancements regarding the validation of uploaded files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ typings
 
 # test artifacts
 output/
+!test/mocks/files/*.csv
 
 *.swp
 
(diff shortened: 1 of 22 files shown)


server/graphql/v2/mutation/TransactionImportsMutations.ts
```diff
@@ -155,9 +155,7 @@ const transactionImportsMutations = {
       // Handle CSV
       let file;
       if (args.file) {
-        file = await UploadedFile.uploadGraphQl(await args.file, 'TRANSACTIONS_IMPORT', req.remoteUser, {
-          supportedMimeTypes: ['text/csv'],
-        });
```
Betree, line 160: Supported mime types are now directly inferred from `kind`, see `models/UploadedFile`

.gitignore
```diff
@@ -39,6 +39,7 @@ typings
 
 # test artifacts
 output/
+!test/mocks/files/*.csv
```
Betree, line 42: To allow `test/mocks/files/transactions.csv`

kewitz (approved):

PR #10497, opened 2024-11-27
enhancement(CreateProject): add the ability to disable contributions

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7650


kewitz (commented):
I think we should also test that CreateOrder mutation is blocked for the specific project.

PR #10500, opened 2024-11-27
Provide signed url for uploaded expense files

hdiniz:
Related https://github.com/opencollective/opencollective/issues/6351


migrations/20241127165812-add-expense-attachments-url-indexes.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.addIndex('ExpenseAttachedFiles', ['url']);
+    await queryInterface.addIndex('ExpenseItems', ['url']);
```
Betree, line 7: Maybe only on non-deleted entries? https://github.com/opencollective/opencollective/issues/3205

server/lib/awsS3.ts
```diff
@@ -39,6 +40,11 @@ if (config.aws.s3.key) {
   });
 }
 
+export function getSignedGetURL(params: GetObjectCommand['input'], options?: { expiresIn: number }) {
+  const command = new GetObjectCommand(params);
+  return getSignedUrl(s3, command, { expiresIn: options?.expiresIn || 10 * 60 });
```
Betree, line 45: :ice_cream: I'd put the `10 * 30` as a default parameter in the function prototype, to make it obvious for people using it.

server/lib/utils.js
```diff
@@ -513,6 +513,12 @@ export const getBearerTokenFromRequestHeaders = req => {
   }
 };
 
+export const getBearerTokenFromCookie = req => {
+  return req?.cookies?.accessTokenPayload && req?.cookies?.accessTokenSignature
+    ? [req.cookies.accessTokenPayload, req.cookies.accessTokenSignature].join('.')
+    : null;
+};
+
```
Betree, line 521: We already have that: `getTokenFromCookie` in `lib/auth.ts`

server/models/UploadedFile.ts
```diff
@@ -129,6 +133,52 @@ class UploadedFile extends Model<InferAttributes<UploadedFile>, InferCreationAtt
...
+    }
+
+    let parsedURL: URL;
+    try {
+      parsedURL = new URL(url);
+    } catch (e) {
+      return false;
+    }
+
+    return parsedURL.origin === config.host.website && /\/api\/files\/\w+/.test(parsedURL.pathname);
```
Betree, line 148:
```suggestion
    return parsedURL.origin === config.host.website && /^\/api\/files\/\w+/.test(parsedURL.pathname);
```

server/models/UploadedFile.ts
```diff
@@ -129,6 +133,52 @@ class UploadedFile extends Model<InferAttributes<UploadedFile>, InferCreationAtt
...
+    }
+
+    let parsedURL: URL;
+    try {
+      parsedURL = new URL(url);
+    } catch (e) {
+      return null;
+    }
+
+    const match = parsedURL.pathname.match(/\/api\/files\/([^\/]+)\/?/);
```
Betree, line 163:
```suggestion
    const match = parsedURL.pathname.match(/^\/api\/files\/([^\/]+)\/?/);
```

server/models/UploadedFile.ts
```diff
@@ -129,6 +133,52 @@ class UploadedFile extends Model<InferAttributes<UploadedFile>, InferCreationAtt
...
+
+    if (match.length !== 2) {
```
Betree, line 165:
```suggestion
    if (match && match.length !== 2) {
```

server/routes.js
```diff
@@ -401,6 +402,8 @@ export default async app => {
    */
   app.get('/docs/search', gitbook.search);
 
+  app.get('/files/:base64UrlEncodedUrl', filesController.getFile);
+
```
Betree, line 406: Maybe move this under https://github.com/opencollective/opencollective-api/blob/16f30b945fd43d327b5c41e3d9a5d82d4c096812/server/routes.js#L397, in the "File downloads" section

templates/partials/expense-items.hbs
```diff
@@ -7,7 +7,7 @@
           <td style="width: 66px; border-bottom: 1px dotted rgb(196, 199, 204); padding: 12px 0;">
             <div style="border: 1px solid #dcdee0; width: 48px; height: 48px; padding: 4px;">
               <a href="{{this.url}}">
-                <div class="preview" style="width: 100%; height: 100%; z-index:0; background-size: contain; background-repeat: no-repeat; background-position: center; background-image: url('{{resizeImage this.url width=640}}');"></div>
+                <div class="preview" style="width: 100%; height: 100%; z-index:0; background-size: contain; background-repeat: no-repeat; background-position: center; background-image: url('{{this.url}}?thumbnail');"></div>
```
Betree, line 10: This won't work: most email clients proxy the images to prevent tracking. I suggest:
1. Redesigning the email to remove previews
2. OR, when fetching thumbnails from the controller as unauthenticated/unauthorized, return a generic file icon.

server/controllers/files.ts
```diff
@@ -0,0 +1,109 @@
...
+import { Request, Response } from 'express';
+
+import { canSeeExpenseAttachments } from '../graphql/common/expenses';
+import { getSignedGetURL, parseS3Url } from '../lib/awsS3';
+import { Collective, Expense, ExpenseAttachedFile, ExpenseItem, UploadedFile } from '../models';
+import { SUPPORTED_FILE_TYPES_IMAGES } from '../models/UploadedFile';
+
+async function hasUploadedFilePermission(req: Request, uploadedFile: UploadedFile): Promise<boolean> {
+  const actualUrl = uploadedFile.getDataValue('url');
+  if (req.remoteUser.id !== uploadedFile.CreatedByUserId) {
```
Betree, line 11: I don't get this condition; if the remote user is the uploader, always return false?

In case it was reversed, I don't think the opposite should be true either: if I submit an expense as the admin of a collective, then later get removed from the admins, I shouldn't have access to the file anymore. In general, we don't want to use `CreatedByUserId` for permissions checks.

server/controllers/files.ts
```diff
@@ -0,0 +1,109 @@
...
+    switch (uploadedFile.kind) {
+      case 'EXPENSE_ITEM': {
+        const expenseItem = await ExpenseItem.findOne({
+          where: {
+            url: actualUrl,
+          },
+          include: { model: Expense, include: [{ model: Collective, as: 'fromCollective' }] },
```
Betree, line 18: :ice_cream: I tend to find the `association` parameter clearer when it's available

```suggestion
          include: { model: Expense, include: [{ association: 'fromCollective' }] },
```

Betree (commented):
Since it's serving a critical feature with sensitive permissions, we need to add proper tests for the `/files/...` controller.

server/lib/utils.js
hdiniz: Thats on the frontend codebase.

server/controllers/files.ts
hdiniz: I got this check wrong when refactoring the validation under this function, its fixed now.

templates/partials/expense-items.hbs
hdiniz: Replaced with a generic placeholder icon. The controller would return the same icon for every resource (existing or not, regardless of the mime type) to prevent exposing any info to unauthenticated requests.

migrations/20241127165812-add-expense-attachments-url-indexes.js
hdiniz: Only ExpenseItems has a `deletedAt` column. I dont know why, but expense attached files does not have one.

server/lib/utils.js
Betree: My bad, indeed! The file that confused me here (because we have one with a similar name in frontend & API), `server/lib/auth.js` is probably a better fit than  `server/lib/utils.js`.

server/controllers/files.ts
```diff
@@ -0,0 +1,119 @@
...
+
+import { canSeeExpenseAttachments } from '../graphql/common/expenses';
+import { getSignedGetURL, parseS3Url } from '../lib/awsS3';
+import { Collective, Expense, ExpenseAttachedFile, ExpenseItem, UploadedFile } from '../models';
+import { SUPPORTED_FILE_TYPES_IMAGES } from '../models/UploadedFile';
+
+async function hasUploadedFilePermission(req: Request, uploadedFile: UploadedFile): Promise<boolean> {
+  const actualUrl = uploadedFile.getDataValue('url');
+  if (req.remoteUser?.id === uploadedFile.CreatedByUserId) {
+    return true;
```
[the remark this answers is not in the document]
hdiniz: This check is necessary while submitting an expense, the expense does not exist yet to make a proper check here and the submitter should be able to see the file they just uploaded.

Would it be ok to have this as a fallback?

1. Get uploaded file related expense
2. If expense exists, check if remoteUser has access to the uploaded file expense
3. If not, check if remoteUser is uploader.

server/graphql/common/expenses.ts
```diff
@@ -1490,9 +1490,39 @@ const checkCanUseAccountingCategory = (
...
+
+    if (UploadedFile.isOpenCollectiveProtectedS3BucketURL(item.url)) {
+      const uploadedFile = await UploadedFile.getFromProtectedURL(item.url);
+      if (!uploadedFile) {
+        throw new ValidationFailed('Invalid expense item url');
+      }
+      mapItemUrlToUploadedFile[item.url] = uploadedFile.getDataValue('url');
+    } else {
+      mapItemUrlToUploadedFile[item.url] = item.url;
+    }
```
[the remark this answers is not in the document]
hdiniz: You're right, I had in mind that the unique constraint on UploadedFile#url would be enough, but it is not.
I just pushed changes to address this:

1. UploadedFile#url is unique, the url validation was updated to restrict use of query params or path hashes that one could use to get a different, but equivalent url.
2. The expense mutations that receive expense items or expense attached files have been updated to check if the remoteUser has access to that url.

hdiniz:
Thanks for your review on this @Betree, I have made changes according to our discussion so far, but its still not complete.
One missing piece here is a strategy for dealing with uploaded files used in expense invites, I will look into it next.

server/controllers/files.ts
```diff
@@ -0,0 +1,81 @@
...
+    return res.status(400).send({ message: 'Invalid id' });
+  }
+
+  const uploadedFile = await UploadedFile.findByPk(decodedId);
+
+  if (!uploadedFile) {
+    return res.status(403).send({ message: 'Unauthorized' });
+  }
+
+  const actualUrl = uploadedFile.getDataValue('url');
```
Betree, line 48: We don't want people to start generating signed URLs for public files. If a file is public, we should return early:

```suggestion
  const actualUrl = uploadedFile.getDataValue('url');
  if (uploadedFile.isPublicFile()) {
    return actualUrl;
  }
```

In `server/models/UploadedFile.ts`:
```
  public isPublicFile(): boolean {
    return [
      'ACCOUNT_AVATAR',
      'ACCOUNT_BANNER',
      'ACCOUNT_LONG_DESCRIPTION',
      'TIER_LONG_DESCRIPTION',
    ].includes(this.kind);
  }
```

This also made me realize: https://github.com/opencollective/opencollective/issues/7718.

test/test-helpers/fake-data.ts
```diff
@@ -400,13 +400,13 @@ export const fakeUploadedFile = async (fileData: Partial<InferCreationAttributes
   const fileName = `${randStr()}${extension}`;
   const kind = sample(SUPPORTED_FILE_KINDS);
   return models.UploadedFile.create({
-    url: fakeUploadedFileURL(kind, fileName),
+    url: fakeUploadedFileURL(fileData?.kind || kind, fileName),
     kind,
     fileSize: randNumber(100, 100000),
```
Betree, line 405: If we do that, we will end up with different file kinds between the `url` and the `kind` columns. Suggested:

```ts
  const kind = fileData?.kind || sample(SUPPORTED_FILE_KINDS);
  return models.UploadedFile.create({
    url: fakeUploadedFileURL(kind, fileName),
    kind,
```

server/graphql/v2/object/Expense.ts
```diff
@@ -546,14 +546,46 @@ export const GraphQLExpense = new GraphQLObjectType<ExpenseModel, express.Reques
...
               itemsFields = [...itemsFields, ...EXPENSE_DRAFT_ITEMS_PRIVATE_FIELDS];
             }
 
             const draftData = pick(expense.data, draftFields);
-            if (expense.data?.items) {
-              draftData.items = (expense.data.items as any[]).map(item => pick(item, itemsFields));
+            const items = ((expense.data.items as { url?: string }[]) || []).map(item => pick(item, itemsFields));
+            for (const item of items) {
+              if (item.url) {
+                const uploadedFile = await req.loaders.UploadedFile.byUrl.load(item.url);
```
Betree, line 559: :ice_cream: Because we're in a for loop, the loader here will actually be slower than a simple `model.find` (there's no batching possible, yet you get the loaders overload). A more performant way to do that would have been to use a `Promise.all`, which is compatible with loaders.

server/graphql/common/expenses.ts
```diff
@@ -1533,11 +1568,34 @@ export const prepareExpenseItemInputs = async (
     );
   }
 
+  const mapItemUrlToUploadedFile: Record<string, string> = {};
```
Betree, line 1571: :ice_cream: I got confused expecting a `Record<string, UploadedFile>` when reading the code below

```suggestion
  const mapItemUrlToUploadedFileUrl: Record<string, string> = {};
```

server/graphql/common/uploaded-file.ts
```diff
@@ -0,0 +1,110 @@
...
+          url: actualUrl,
+        },
+        include: { model: Expense, include: [{ association: 'fromCollective' }] },
+      });
+
+      const expense = expenseItem?.Expense;
+
+      if (!expense) {
+        return req.remoteUser?.id === uploadedFile.CreatedByUserId;
+      }
```
Betree, line 80: How can an expense item exist without an expense? Did you mean to write:

```suggestion
      const expense = expenseItem?.Expense;

      if (!expenseItem) {
        return req.remoteUser?.id === uploadedFile.CreatedByUserId;
      }
```

server/graphql/common/uploaded-file.ts
```diff
@@ -0,0 +1,110 @@
...
+          url: actualUrl,
+        },
+        include: { model: Expense, include: [{ model: Collective, as: 'fromCollective' }] },
+      });
+
+      const expense = expenseAttachedFile?.Expense;
+
+      if (!expense) {
+        return req.remoteUser?.id === uploadedFile.CreatedByUserId;
+      }
```
Betree, line 99: Same here, I don't get why we'd check that. If the source expense doesn't exist, its related uploaded files & items are supposed to be deleted.

Betree (approved):
Great job with the draft status handling :+1: 

Looking good overall, I expressed some concerns below but they're non-blocking. Let's try to deploy & test this in production before the holidays.

PR #10504, opened 2024-11-28
Contributions: Use last charged date for recurring contributions

kewitz:
Relates to https://github.com/opencollective/opencollective/issues/7694

### DIFF
--- a/server/graphql/v2/query/collection/OrdersCollectionQuery.ts
+++ b/server/graphql/v2/query/collection/OrdersCollectionQuery.ts
@@ -329,10 +329,18 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
   }
 
   if (args.chargedDateFrom) {
-    where[Op.and].push({ ['$Subscription.lastChargedAt$']: { [Op.gte]: args.chargedDateFrom } });
+    where[Op.and].push(
+      sequelize.where(sequelize.literal(`COALESCE("Subscription"."lastChargedAt", "Order"."createdAt")`), {
+        [Op.gte]: args.chargedDateFrom,
+      }),
+    );
   }
   if (args.chargedDateTo) {
-    where[Op.and].push({ ['$Subscription.lastChargedAt$']: { [Op.lte]: args.chargedDateTo } });
+    where[Op.and].push(
+      sequelize.where(sequelize.literal(`COALESCE("Subscription"."lastChargedAt", "Order"."createdAt")`), {
+        [Op.lte]: args.chargedDateTo,
+      }),
+    );
   }
 
   if (args.status && args.status.length > 0) {
@@ -400,7 +408,7 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
   }
   const { offset, limit } = args;
   return {
-    nodes: () => models.Order.findAll({ include, where, order, offset, limit, logging: true }),
+    nodes: () => models.Order.findAll({ include, where, order, offset, limit }),
     totalCount: () => models.Order.count({ include, where }),
     limit: args.limit,
     offset: args.offset,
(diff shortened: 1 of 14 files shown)


server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -374,10 +389,18 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
...
+    order = [
+      [sequelize.literal(`COALESCE("Subscription"."lastChargedAt", "Order"."createdAt")`), args.orderBy.direction],
+    ];
+  } else {
+    order = [[args.orderBy.field, args.orderBy.direction]];
+  }
   const { offset, limit } = args;
   return {
-    nodes: () => models.Order.findAll({ include, where, order, offset, limit }),
+    nodes: () => models.Order.findAll({ include, where, order, offset, limit, logging: true }),
```
gustavlrsn, line 403: Leftover?
```suggestion
    nodes: () => models.Order.findAll({ include, where, order, offset, limit }),
```

gustavlrsn (approved):
Great stuff! Migration, filters and sorting seems to work well 👍

server/paymentProviders/stripe/creditcard.ts
```diff
@@ -117,7 +117,12 @@ const createChargeAndTransactions = async (
   });
 
   const charge = paymentIntent.latest_charge || (paymentIntent as any).charges.data[0];
-  return createChargeTransactions(charge as Stripe.Charge, { order });
+  const transaction = await createChargeTransactions(charge as Stripe.Charge, { order });
+  if (order.SubscriptionId) {
+    const subscription = await models.Subscription.findByPk(order.SubscriptionId);
+    await subscription.update({ lastChargedAt: transaction.clearedAt });
+  }
```
gustavlrsn, line 124: Seems like the subscription is not created yet at this point (I guess the same is true for other payment providers/options)

Maybe what is needed is something like the following in the `executeOrder` function in `server/lib/payments.ts` (line 867):

```typescript
order.interval && (await createSubscription(order, { lastChargedAt: transaction.clearedAt }));
```

gustavlrsn (changes requested):
Found an issue when testing together with the frontend to create a new order, where the `lastChargeDate` is not set on creation (and thus breaking the lastChargeDate filtering), added a comment on where the issue seems to be

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -314,12 +323,18 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
...
     where['data.expectedAt'] = where['data.expectedAt'] || {};
     where['data.expectedAt'][Op.lte] = args.expectedDateTo;
   }
 
+  if (args.chargedDateFrom) {
+    where[Op.and].push({ ['$Subscription.lastChargedAt$']: { [Op.gte]: args.chargedDateFrom } });
+  }
+  if (args.chargedDateTo) {
+    where[Op.and].push({ ['$Subscription.lastChargedAt$']: { [Op.lte]: args.chargedDateTo } });
+  }
```
gustavlrsn, line 336: The filter is broken for one-time contributions, perhaps a similar solution as for the ordering is needed:

```suggestion
  if (args.chargedDateFrom) {
    where[Op.and].push(
      sequelize.where(sequelize.literal(`COALESCE("Subscription"."lastChargedAt", "Order"."createdAt")`), {
        [Op.gte]: args.chargedDateFrom,
      }),
    );
  }
  if (args.chargedDateTo) {
    where[Op.and].push(
      sequelize.where(sequelize.literal(`COALESCE("Subscription"."lastChargedAt", "Order"."createdAt")`), {
        [Op.lte]: args.chargedDateTo,
      }),
    );
  }
```

server/paymentProviders/stripe/creditcard.ts
kewitz: Nice catch. You're correct, for the first charge, we need to set lastChargeAt in execute order.

PR #10506, opened 2024-11-29
Search query: filering options at the resolver level

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7688

Introduces parameters on the `accounts` search:
- type
- isHost
- tags

### How to add new parameters 

1. Add the argument on the fields resolver @ `server/lib/elastic-search/graphql-search.ts`
2. Optionally define a `prepareArguments` function if the field needs a conversion between GraphQL and ES (e.g. loading accounts)
3. Add a case in `getIndexConditions` @ `server/lib/elastic-search/search.ts`

### Test query

```gql
{
  search(searchTerm: "open source") {
    results {
      accounts(isHost: false, type: INDIVIDUAL, tags: ["openmiami"]) {
        highlights
        collection {
          totalCount
          nodes {
            id
            name
            isHost
            tags
            type
          }
        }
      }
    }
  }
}
```


server/graphql/v2/query/SearchQuery.ts
```diff
@@ -53,47 +45,15 @@ const SearchQuery = {
...
-          accounts: getSQLSearchResolver(models.Collective, query),
-          comments: getSQLSearchResolver(models.Comment, query),
-          expenses: getSQLSearchResolver(models.Expense, query),
-          hostApplications: getSQLSearchResolver(models.HostApplication, query),
-          orders: getSQLSearchResolver(models.Order, query),
-          tiers: getSQLSearchResolver(models.Tier, query),
-          transactions: getSQLSearchResolver(models.Transaction, query),
-          updates: getSQLSearchResolver(models.Update, query),
-        },
-      };
```
Betree, line 95: The SQL search was initially implemented for the hackathon and could have been useful in dev, but maintaining feature parity seems too expensive for the narrow usage it will have.

server/lib/elastic-search/search.ts
```diff
@@ -118,58 +150,73 @@ const buildQuery = (
...
       },
-    },
-    /* eslint-enable camelcase */
-  });
+      /* eslint-enable camelcase */
+    });
+  } catch (e) {
+    reportErrorToSentry(e, { user, extra: { requestedIndexes, searchTerm, limit, account, host } });
+    throw new Error('The search query failed, please try again later');
+  }
```
Betree, line 221: Added some error handling to make sure we don't expose ES errors in the frontend

gustavlrsn (approved):
Great, these resolver filters will be very useful! Works well on my end.

PR #10508, opened 2024-11-29
Performance: Contributions Dashboard

kewitz:
Related to https://github.com/opencollective/opencollective/issues/7696

It turns out, we can optimize the Contributions Dashboard query by adding an index for the order clause.
Before index: https://www.pgexplain.dev/plan/3413977b-7d8f-4fcc-bd7d-de64a4dcd25b
After index: https://www.pgexplain.dev/plan/9fb3d821-1a7e-4a20-845b-c180f5bc1ff4


Betree (approved):
Nice one

PR #10517, opened 2024-12-03
Feat: `activeContributors` resolver to get contributors for goals

gustavlrsn:
For https://github.com/opencollective/opencollective/issues/7605

### Description
Adding a resolver to get active contributors (and ability to set a time period you're interested in).


Betree (approved):
Works for me, we can probably make it more performant but it's good enough for a proof of concept.

PR #10519, opened 2024-12-04
fix(Search): add support for restoring missing/undeleted entries

Betree:
Adds the `npm run script scripts/search.ts restore <indexName>` command to look for entries that are missing from ElasticSearch. This could happen for two reasons:
1. Because we previously missed it in synchronization jobs
2. Because `deletedAt` was manually set to `NULL` directly in the DB

Also made some improvements to the options names.


gustavlrsn (approved):
Nice!

PR #10521, opened 2024-12-04
feat(ElasticSearch): Postgres synchronization

Betree:
Resolve https://github.com/opencollective/opencollective/issues/7689


gustavlrsn (approved):

Betree:
@gustavlrsn I added support for a customizable index prefix to share an ElasticSearch instance between multiple environments. The indexes are created as `staging_collectives` instead of `collectives`.

This is also used in tests: test data will be created as `test_{index}` to not tamper the local dev data.

PR #10530, opened 2024-12-06
Add hasDebt parameter on Transactions resolver

znarf:
ability to answer the question "I want to have all platform tips that were directly collected"


Betree (approved):
Works for me, not sure about the performance cost?

znarf:
@Betree It only cost if you're using it. And it looks acceptable when using it, I haven't measured.

PR #10534, opened 2024-12-09
deps: Upgrade to Postgres 15

Betree:
Key changes:
- Update table permissions since schema default permissions changed in PG 15 (see https://github.com/opencollective/opencollective-api/pull/10534#discussion_r1877553542).
- Update CI jobs to `ubuntu-24.04` (which is [supposed](https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/#ubuntu-latest-upcoming-breaking-changes) to become the default in the next coming days) to install postgres-client-16 and make sure we use the latest tooling for restoring.


scripts/db_restore.sh
```diff
@@ -72,13 +74,50 @@ fi
...
+BEGIN 
+   EXECUTE (
+      SELECT string_agg(
+         format('ALTER TABLE %I.%I ENABLE TRIGGER ALL', schemaname, tablename), '; '
+      ) 
+      FROM pg_tables 
+      WHERE schemaname = 'public'
+   ); 
+END \$\$;
+"
```
Betree, line 120: Two changes here:
1. Rather than bruteforcing the same pg_restore command 2 times, we're now restoring only the schema first (`-s`) and then the data (`-a`).
2. Disabling triggers before restoring the data, to prevent any missing foreign key issues from popping in for interdependent rows.

scripts/db_restore.sh
```diff
@@ -72,13 +74,50 @@ fi
...
-# The first time we run it, we will trigger FK constraints errors
-set +e
-pg_restore -U postgres -h localhost --no-acl --no-owner --role=${LOCALDBUSER} -n public -O -c -d "${LOCALDBNAME}" "${DBDUMP_FILE}" 2>/dev/null
-set -e
+PG_RESTORE_OPTIONS=(-U postgres -h localhost --no-acl --no-owner --role="${LOCALDBUSER}" -n public -O -d "${LOCALDBNAME}")
+PG_RESTORE_COMMAND=pg_restore
+if [ -n "$USE_DOCKER" ]; then
+  PG_RESTORE_COMMAND="sudo docker run --rm --network host -v $(dirname "$DBDUMP_FILE"):/backup postgres:14 pg_restore -h localhost -p 5432 -U opencollective"
+  DBDUMP_FILE="/backup/$(basename "$DBDUMP_FILE")"
+fi
```
Betree, line 82: Added this small thing to easily change `pg_dump`/`pg_restore` versions using docker, but my tests so far surface that their versions are retro compatible, so using the local one (in my case, v16) should give the best results

.github/workflows/e2e.yml
```diff
@@ -62,14 +62,15 @@ jobs:
...
-        run: sudo apt-get install -y graphicsmagick
+      - name: Install dependencies
+        run: |
+          sudo apt-get install \
+          `# Cypress dependencies - see https://docs.cypress.io/app/get-started/install-cypress#UbuntuDebian` \
+          libgtk2.0-0t64 libgtk-3-0t64 libgbm-dev libnotify-dev libnss3 libxss1 libasound2t64 libxtst6 xauth xvfb \
+          `# Postgres client` \
+          postgresql-client-16 \
+          `# GraphicsMagick (not sure if needed)` \
+          graphicsmagick
```
Betree, line 73: Single command for faster install

scripts/db_restore.sh
```diff
@@ -72,12 +74,16 @@ fi
   set -e
 } | tee >/dev/null
 
+# Update table permissions
+echo "Updating table permissions"
+psql -U postgres -h localhost $LOCALDBNAME -c "GRANT ALL ON SCHEMA public TO ${LOCALDBUSER};"
+
```
Betree, line 80: See https://medium.com/@dmitry.romanoff/postgresql-version-15-error-permission-denied-for-schema-public-dc8285c41e00

kewitz:
Should we go to 16.4 to match production?

Betree:
> Should we go to 16.4 to match production?

It's the next step, but I want to make it incremental to easily rollback if it's not working for some in the team.

PR #10536, opened 2024-12-10
deps: Migrate to postgres 16

Betree:
Mostly updated the scripts to make the dev transition easier


docs/postgres.md
```diff
@@ -1,8 +1,6 @@
 # PostgreSQL Database
 
-You need to have PostgreSQL > 15.x.
-
-In production, we're currently running 16.4.
```
Betree, line 5: No need to disclose the version in production now that it matches

PR #10552, opened 2024-12-19
fix(SanitizeHTML): arrow dir attribute for RTL languages

Betree:
Fix https://github.com/opencollective/opencollective/issues/7711


epageperron:
Thank you ! :-)

PR #10557, opened 2024-12-19
Stop using temporary quotes to check for valid paymentOption

kewitz:
For some reason, Wise is not returning paymentOptions with payIn = 'BALANCE' when using the temporary quote endpoint.
I'm updating our code to actually quote the expense if there's no valid expense available.


test/server/controllers/transferwise.test.ts
```diff
@@ -79,33 +79,32 @@ describe('server/controllers/transferwise', () => {
...
+        payIn: 'BALANCE',
+        sourceAmount: 101.14,
+        targetAmount: 90.44,
+        sourceCurrency: 'USD',
+        targetCurrency: 'EUR',
+        payOut: 'BANK_TRANSFER',
+        disabled: false,
+      },
+    };
+    sandbox.stub(transferwise, 'quoteExpense').resolves(quote);
```
Betree, line 107: Are we sure that wise will never return `paymentOptions` (array) in any of the requests? If not, should we keep supporting for both, e.g. with a monkey patch like:

```ts
if (!quote.paymentOption && quote.paymentOptions?.length) {
  quote.paymentOption = quote.paymentOptions[0];
}
```

Also, did they announce the breaking change somewhere?

Betree (approved):
Lgtm

test/server/controllers/transferwise.test.ts
kewitz: @Betree, Wise will return an array from the API but the `quoteExpense` is an internal function in the payment providers' libs that will return a quote with the selected BALANCE option already selected and inserted in `quote.paymentOption`.

In retrospect, I don't think it was such a great idea to mutate the original quote to persist this information, but it is how it currently works.

kewitz: (Everything is properly typed tho, so it shouldn't be insecure)

PR #10590, opened 2024-12-30
perf(Postgres): Remove unused indexes

Betree:
Considering that indexes add overhead to SQL write queries and can make the query planner slower, this migration drops indexes that are seemingly no longer used in the codebase. They were identified by running the following
query in the database:

```
SELECT
    relname AS table_name,
    indexrelname AS index_name,
    idx_scan AS index_scans,
    pg_size_pretty(pg_relation_size(indexrelid)) AS index_size,
    indexdef
FROM
    pg_stat_user_indexes
JOIN
    pg_indexes ON pg_indexes.indexname = pg_stat_user_indexes.indexrelname
WHERE
    -- Unused indexes
    idx_scan = 0
    -- That are not unique (we want to keep the constraints)
   AND indexdef NOT ILIKE 'CREATE UNIQUE INDEX%'
ORDER BY
    pg_relation_size(indexrelid) DESC;
```

Methodology:
1. On a Postgres server that has been running for some time (Heroku doesn't let us check the `SELECT stats_reset FROM pg_stat_database WHERE datname = current_database()`)
2. Have a tour in the interface to trigger many different queries
3. Have a tour in metabase dashboard to trigger many different queries
4. Run the above query to check unused indexes


migrations/20241230093753-drop-unused-indexes.js
```diff
@@ -0,0 +1,38 @@
...
+      // 'UploadedFiles_s3_hash', // Table: UploadedFiles. Size: 16 MB. This one is unused because we've disabled Klippa, but we want to keep it in case we re-enable the feature.
+      'transaction_wise_transfer_id', // Table: Transactions. Size: 8144 kB
+      'expense_tag_stats_tag', // View: ExpenseTagStats. Size: 6864 kB. We're seemingly never fetching BY tag.
+      // 'uploaded_files__created_by_user_id', // Table: UploadedFiles. Size: 5792 kB. For Klippa rate-limiting, keeping in case we re-enable an AI feature.
+      'expenses_data_stripe_virtual_card_transaction_id', // Table: Expenses. Size: 1104 kB
+      'expenses_data_paypal_transaction_id', // Table: Expenses. Size: 1048 kB
+      'wise_batchgroup_status', // Table: Expenses. Size: 648 kB
+      'privacy_transfer_id', // Table: Transactions. Size: 16 kB
+      'Subscriptions_lastChargedAt', // Table: Subscriptions. Size: 16 kB
+      'virtual_card_requests__host_collective_id__collective_id', // Table: VirtualCardRequests. Size: 16 kB
```
Betree, line 25: cc @znarf for the following indexes:
- `transactions__collective_id_createdAt` (https://github.com/opencollective/opencollective-api/pull/9743)
- `transactions__host_collective_id_createdAt` (https://github.com/opencollective/opencollective-api/pull/9743)
- `transactions__hostCollective_clearedAt` (https://github.com/opencollective/opencollective-api/pull/9743)

Betree, line 25: cc @kewitz for the following indexes:
- `orders__req_ip` (https://github.com/opencollective/opencollective-api/pull/8151)
- `transaction_wise_transfer_id`
- `wise_batchgroup_status`
- `Subscriptions_lastChargedAt`
- `payment_methods__fingerprint`

Betree, line 25: cc @gustavlrsn for the following indexes:
- `expense_tag_stats_tag`

kewitz: Everything seems safe to delete, but I'm surprised payment_methods__fingerprint is not being used. I'll investigate this further.

kewitz: Everything is safe to delete.

Betree: We'll keep the fingerprint for now, see https://oficonsortium.slack.com/archives/C06K5J91675/p1735902404908739

znarf: That looks pretty dangerous to me, I think this has the potential to break production. `transactions__collective_id_createdAt` and `transactions__host_collective_id_createdAt` were carefully created for the generic GraphQL transactions resolver.

znarf: The transferwise id was introduced for webhooks, I doubt we moved away from that. https://github.com/opencollective/opencollective-api/pull/6002 https://github.com/opencollective/opencollective-api/pull/7379

PR #10600, opened 2024-12-31
feat(cron): implement data retention policy

Betree:
Resolve https://github.com/opencollective/opencollective/issues/5388

### DIFF
--- a/cron/daily/96-apply-data-retention-policy.ts
+++ b/cron/daily/96-apply-data-retention-policy.ts
@@ -1,0 +1,93 @@
+import '../../server/env';
+
+import { ModelStatic } from 'sequelize';
+
+import logger from '../../server/lib/logger';
+import { parseToBoolean } from '../../server/lib/utils';
+import models, { Op, sequelize } from '../../server/models';
+import { runCronJob } from '../utils';
+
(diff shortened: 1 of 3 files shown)


Betree:
> Well done. I would argue that Location should also be considered FINANCIAL since we don't store anything in Orders about the location.

That's a good point, updating that now.

cron/daily/96-apply-data-retention-policy.ts
```diff
@@ -0,0 +1,93 @@
...
+export const runDataRetentionPolicyJob = async (isDryRun = false) => {
+  const transaction = await sequelize.transaction();
+  for (const [model, retentionPeriod] of MODEL_RETENTION_PERIODS) {
+    const result = await model.destroy({
+      transaction,
+      force: true,
+      where: {
+        deletedAt: {
+          [Op.not]: null,
+          [Op.lte]: sequelize.literal(`NOW() - INTERVAL '${retentionPeriod}'`),
```
kewitz, line 66: 🍨 Looking back now, I think we should have used an explicit "AND" array here. I just had a brief "WTF" moment reading this code and thinking this was an "OR".

PR #10607, opened 2025-01-03
ExpensesCollectionQuery: Add NON_HOST_ADMIN lastCommentBy filter option

kewitz:


Betree (approved):
Great, filtering by non-host admins looks like the most future-proof solution for me.

Looks like smart dump stuff spilled over?

kewitz:
@Betree I did a small refactor so I could safely copy the edge cases locally and investigate this issue. I left it here because I didn't thought it deserved a separated PR...

PR #10615, opened 2025-01-07
Set signedurl content disposition if downloading from html anchor tag

hdiniz:


Betree (approved):
Lgtm, let's try to deploy it as soon as possible

PR #10624, opened 2025-01-10
Add Contributor(s) Query

kewitz:


server/graphql/v2/query/ContributorQuery.ts
```diff
@@ -0,0 +1,68 @@
...
+      description: 'Context host to fetch the contributor from',
+    },
+  },
+  async resolve(_: void, args, req: Express.Request) {
+    assert(args.account && args.host, 'You must provide either an account or a host to fetch the contributors');
+    const account = await fetchAccountWithReference(args.account, { throwIfMissing: true });
+    const host = await fetchAccountWithReference(args.host, { throwIfMissing: true });
+    if (!req.remoteUser?.isAdminOfCollective(host)) {
+      throw new BadRequest('Only admins can lookup for members using the "host" argument');
+    }
```
Betree, line 29: There's a lack of consistency in what should be required:
- Parameters are optional 
- `assert` enforces both arguments are set (but the message says "Or")
- Direct fetch with arg check also assumes both should be

server/graphql/v2/query/ContributorQuery.ts
```diff
@@ -0,0 +1,68 @@
...
+      {
+        replacements: {
+          accountid: account.id,
+          hostid: host.id,
+        },
+        type: sequelize.QueryTypes.SELECT,
+        raw: true,
+        plain: true,
+      },
+    );
```
Betree, line 62: I would suggest implementing this query in `server/lib/contributors.ts`:
1. To take advantage of the cache when fetching by collective
2. To re-use the query when possible (the one here seems to be missing a few fields such as `since`)
3. For consistency

server/graphql/v2/query/collection/ContributorsCollectionQuery.ts
```diff
@@ -0,0 +1,153 @@
...
+    },
+    orderBy: {
+      type: new GraphQLNonNull(GraphQLChronologicalOrderInput),
+      defaultValue: { field: 'createdAt', direction: 'ASC' },
+      description: 'Order of the results',
+    },
+    includeInherited: {
+      type: GraphQLBoolean,
+      defaultValue: true,
+    },
```
Betree, line 44: Maybe add a description for this one, it's not obvious. And apparently not used?

server/graphql/v2/query/collection/ContributorsCollectionQuery.ts
```diff
@@ -0,0 +1,153 @@
...
+      args.limit = 100;
+    }
+    if (isNil(args.offset) || args.offset < 0) {
+      args.offset = 0;
+    }
+    if (args.limit > 100 && !req.remoteUser?.isRoot()) {
+      throw new Error('Cannot fetch more than 1,000 members at the same time, please adjust the limit');
+    }
+
+    assert(args.account || args.host, 'You must provide either an account or a host to fetch the contributors');
```
Betree, line 58: Can't use both at the same time (xor)

```suggestion
    assert(Boolean(args.account) ^ Boolean(args.host), 'You must provide either an account or a host to fetch the contributors');
```

server/graphql/v2/query/collection/ContributorsCollectionQuery.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    const nodes = () =>
+      sequelize.query(
+        `
+          SELECT mc.*,
+            ARRAY_AGG(DISTINCT (m.role)) as roles,
+            ARRAY_AGG(DISTINCT (c.id)) as "ContributedCollectiveIds",
+            'ADMIN' = ANY (ARRAY_AGG(m.role)) as "isAdmin",
+            'ADMIN' = ANY (ARRAY_AGG(m.role))  OR 'MEMBER' = ANY (ARRAY_AGG(m.role)) as "isCore",
+            'BACKER' = ANY (ARRAY_AGG(m.role)) as "isBacker"
```
Betree, line 127: Same here, would be great to rely on `server/lib/contributors.ts` when we can.

server/graphql/v2/query/collection/ContributorsCollectionQuery.ts
```diff
@@ -0,0 +1,153 @@
...
+    type: { type: new GraphQLList(GraphQLAccountType) },
+    email: {
+      type: EmailAddress,
+      description: 'Admin only. To filter on the email address of a member, useful to check if a member exists.',
+    },
+    orderBy: {
+      type: new GraphQLNonNull(GraphQLChronologicalOrderInput),
+      defaultValue: { field: 'createdAt', direction: 'ASC' },
+      description: 'Order of the results',
+    },
```
Betree, line 40: Not used?

Betree (commented):
Some feedback below, it's looking good overall.

server/graphql/v2/query/ContributorQuery.ts
kewitz: We have an issue here. This host contributor we're returning here is a subset of the existing "Contributor" object.
I'll move these functions and DRY them, but they won't be compatible between existing library queries and this new query.

server/graphql/v2/query/ContributorQuery.ts
kewitz: We have an issue here. This host contributor we're returning here is a subset of the existing "Contributor" object.
I'll move these functions and DRY them, but they won't be compatible between existing library queries and this new query.

kewitz: Because the point of this endpoint is to expose contributors to the context of the host and not the collective.

kewitz: I mean, we use the same type and object, we just do not populate all fields. I think we'll eventually reach that point but I'm not sure if we need to query all those fields now or if that is viable performance-wise.

PR #10626, opened 2025-01-13
Provide thumbnail for protected file uploads

hdiniz:


Betree (commented):
I'm really worried about the load we're going to put on the API with that one, we normally do this kind of processing in https://github.com/opencollective/opencollective-images. Can we move that logic there?

PR #10630, opened 2025-01-14
More details in reference when paying out with WISE

xdamman:
Fixes https://github.com/opencollective/opencollective/issues/7765

I'd recommend fetching the requirements first and applying the provided regex to verify the validity of the reference.
See https://docs.wise.com/api-docs/api-reference/transfer#transfer-requirements

In case of validation error, it's ok to fall back to just the `expense.id`.

### DIFF
--- a/server/paymentProviders/transferwise/index.ts
+++ b/server/paymentProviders/transferwise/index.ts
@@ -242,13 +242,30 @@ async function createTransfer(
     throw new TransferwiseError(message, null, { quote });
   }
 
+  const country = recipient.details.address?.country;
+  let reference = `${expense.collective.slug.toUpperCase().substring(0, 3)}${expense.id}`; // MAX 10 chars for the US
+  switch (country) {
+    case 'GBP': // max 18 chars
+      reference = `${expense.collective.slug.toUpperCase().substring(0, 10)}${expense.id}`;
+      break;
+    case 'EUR': // max 35 chars
+      reference = `${expense.collective.slug.substring(0, 25)}${expense.id}`;
+      break;
+    case 'USD': // max 10 chars
+      reference = `${expense.collective.slug.toUpperCase().substring(0, 3)}${expense.id}`;
+      break;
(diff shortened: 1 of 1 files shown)


server/paymentProviders/transferwise/index.ts
```diff
@@ -242,13 +242,30 @@ async function createTransfer(
...
+    case 'EUR': // max 35 chars
+      reference = `${expense.collective.slug.substring(0, 25)}${expense.id}`;
+      break;
+    case 'USD': // max 10 chars
+      reference = `${expense.collective.slug.toUpperCase().substring(0, 3)}${expense.id}`;
+      break;
+    case 'MXN': // max 100 chars
+      reference = `${expense.collective.slug.substring(0, 100)}${expense.id}`;
+      break;
+  }
```
Betree, line 260: The substring should always be `substring(0, maxLength - expense.id.toString().length)`. Looks like this was considered for some (GBP, EUR, USD) but not MXN

Betree (commented):
I recommend separating the two information (e.g. with a `-` character), otherwise it could mess up with collectives that have numbers in their names (e.g. "collective75" + Expense 42563 = "collective7542563").

For the rest, I'll let @kewitz give a proper review.

kewitz:
Xavier, I suggest addressing this in the frontend (PayExpenseModal), where we implemented the Wise reference validation and allow fiscal hosts to customize reference value.

I would happily implement this. Could you give a bit more information on the core problem this addresses? Or why isn't the expense id enough?

kewitz:
Closed in favor of https://github.com/opencollective/opencollective-frontend/pull/10926

server/paymentProviders/transferwise/index.ts
xdamman: Yes, it's also wrong to use the currency code instead of country code. Please consider my PRs more as a way to kickstart the conversation. Feel free to rewrite (lmk if you prefer that I just write a github issue).

PR #10634, opened 2025-01-20
DX: Remove annoying logs

kewitz:
- [x] The lack of a mailing service shouldn't break API
- [x] If you want to check Sentry logs use `SENTRY_LOG_TO_CONSOLE`

### DIFF
--- a/server/lib/sentry/index.ts
+++ b/server/lib/sentry/index.ts
@@ -180,7 +180,7 @@ export const reportErrorToSentry = (err: Error, params: CaptureErrorParams = {})
 
     if (checkIfSentryConfigured()) {
       Sentry.captureException(err);
-    } else {
+    } else if (config.sentry?.logToConsole) {
       logger.error(
         `[Sentry disabled] The following error would be reported: ${err.message} (${safeJsonStringify(
           {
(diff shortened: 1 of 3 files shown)


server/lib/sentry/index.ts
```diff
@@ -180,7 +180,7 @@ export const reportErrorToSentry = (err: Error, params: CaptureErrorParams = {})
 
     if (checkIfSentryConfigured()) {
       Sentry.captureException(err);
-    } else {
+    } else if (config.sentry?.logToConsole) {
```
Betree, line 183: Are thrown exceptions (e.g. in graphql resolvers) still being logged in console? It's important to have them

kewitz: They should, this is the function that `reportErrorToSentry` not that logs it to the console...

Betree: @kewitz Some places in the code are not logging to the console, they're just throwing the error at Sentry. And `logger` doesn't always show the stack trace. That's why this helper was introduced. I'm a bit wary that this change could invisibilize some errors in dev, but no blocker at that point since most of them have a `logger.error` next to them.

Ideally, I would love to see an API like:

```ts
reportErrorToSentry = (error, { log = true } = {}) => {}
```

Then instead of:

```ts
catch (e) {
  logger.error(`Something happened`, e);
  reportErrorToSentry(e);
}
```

We would simply write:

```ts
catch (e) {
  reportErrorToSentry(e);
}
```

Or, if we don't want to log anything:

```ts
catch (e) {
  reportErrorToSentry(e, { log: false });
}
```

kewitz: https://github.com/opencollective/opencollective-api/pull/10638

PR #10635, opened 2025-01-20
Only check if accounting category can be used if it is included in the expense input

hdiniz:
Fixes https://github.com/opencollective/opencollective/issues/7753

Only check if accounting category can be used if it is included in the expense input. Also impacts existing expense flow.


Betree (approved):
:+1:

PR #10638, opened 2025-01-21
DX: reportError()

kewitz:
Here's my suggestion:
- Fix function name
- Remove unnecessary messaging
- Fallback to console logging if Sentry is not available

Preview:
Screenshot 2025-01-21 at 21 02 05

### DIFF
--- a/server/lib/sentry/index.ts
+++ b/server/lib/sentry/index.ts
@@ -172,28 +172,24 @@ const simplifyReq = req =>
  * Helper to capture an error on Sentry
  */
 export const reportErrorToSentry = (err: Error, params: CaptureErrorParams = {}): void => {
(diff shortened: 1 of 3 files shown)


server/lib/sentry/index.ts
```diff
@@ -171,29 +171,22 @@ const simplifyReq = req =>
...
-        )})`,
-      );
-    }
-  });
+    });
+  } else {
+    logger.error(err.stack ? err.stack : err.message, {
+      ...params,
+      req: redactSensitiveDataFromRequest(simplifyReq(params.req)),
+    });
```
Betree, line 188: I'd recommend keeping the `safeJsonStringify` here, you never know what could be passed in `data`.

Betree (commented):
Works for me

PR #10661, opened 2025-01-23
ContributionFlow: Information requirement resolvers

kewitz:
Adds CONTRIBUTOR_INFO_THRESHOLDS policy and ContributorProfile.totalContributedToHost resolver.

### DIFF
--- a/server/graphql/v2/object/ContributorProfile.ts
+++ b/server/graphql/v2/object/ContributorProfile.ts
@@ -2,7 +2,6 @@ import { GraphQLBoolean, GraphQLObjectType } from 'graphql';
 import { GraphQLDateTime } from 'graphql-scalars';
 import moment from 'moment';
 
-import { getFxRate } from '../../../lib/currency';
 import type { ContributorsLoaders } from '../../loaders/contributors';
 import { GraphQLAccount } from '../interface/Account';
 
@@ -31,12 +30,12 @@ export const GraphQLContributorProfile = new GraphQLObjectType({
         },
         since: {
           type: GraphQLDateTime,
-          description: 'The date since when to calculate the total',
+          description: 'The date since when to calculate the total. Defaults to the start of the current year.',
         },
       },
       resolve: async ({ account, forAccount }, args, req) => {
         const host = await req.loaders.Collective.byId.load(forAccount.HostCollectiveId);
-        const since = moment(args.since).toISOString() || moment.utc().startOf('year').toISOString();
+        const since = args.since ? moment(args.since).toISOString() : moment.utc().startOf('year').toISOString();
         const stats = await (req.loaders.Contributors as ContributorsLoaders).totalContributedToHost
           .buildLoader({ hostId: host.id, since })
           .load(account.id);
@@ -47,8 +46,11 @@ export const GraphQLContributorProfile = new GraphQLObjectType({
         }
 
         if (args.inCollectiveCurrency && stats.currency !== currency) {
-          const fxRate = await getFxRate(stats.currency, currency);
-          return { value: Math.round(stats.amount * fxRate), currency };
+          const convertParams = { amount: stats.amount, fromCurrency: stats.currency, toCurrency: currency };
+          return {
+            value: await req.loaders.CurrencyExchangeRate.convert.load(convertParams),
+            currency,
+          };
         }
         return { value: stats.amount, currency };
       },


server/graphql/v2/object/ContributorProfile.ts
```diff
@@ -10,5 +16,42 @@ export const GraphQLContributorProfile = new GraphQLObjectType({
...
+      args: {
+        inCollectiveCurrency: {
+          type: GraphQLBoolean,
+          defaultValue: false,
+          description: 'When true, the amount is converted to the currency of the collective',
+        },
+        since: {
+          type: GraphQLDateTime,
+          description: 'The date since when to calculate the total',
+        },
```
Betree, line 35:
```suggestion
        since: {
          type: GraphQLDateTime,
          description: 'The date since when to calculate the total. Defaults to the start of the current year.',
        },
```

server/graphql/v2/object/ContributorProfile.ts
```diff
@@ -10,5 +16,42 @@ export const GraphQLContributorProfile = new GraphQLObjectType({
...
+      },
+      resolve: async ({ account, forAccount }, args, req) => {
+        const host = await req.loaders.Collective.byId.load(forAccount.HostCollectiveId);
+        const since = moment(args.since).toISOString() || moment.utc().startOf('year').toISOString();
```
Betree, line 39: `moment(undefined)` === `moment()`. You probably meant:

```suggestion
        const since = args.since ? moment(args.since).toISOString() : moment.utc().startOf('year').toISOString();
```

server/graphql/v2/object/ContributorProfile.ts
```diff
@@ -10,5 +16,42 @@ export const GraphQLContributorProfile = new GraphQLObjectType({
...
+          .load(account.id);
+
+        const currency = args.inCollectiveCurrency ? forAccount.currency : host.currency;
+        if (!stats) {
+          return { value: 0, currency };
+        }
+
+        if (args.inCollectiveCurrency && stats.currency !== currency) {
+          const fxRate = await getFxRate(stats.currency, currency);
+          return { value: Math.round(stats.amount * fxRate), currency };
```
Betree, line 51: :ice_cream: 

```suggestion
          const convertParams = { amount: stats.amount, fromCurrency: stats.currency, toCurrency: currency };
          return { 
            value: await req.loaders.CurrencyExchangeRate.convert.load(convertParams),
            currency
          };
```

server/graphql/v2/object/Policies.ts
```diff
@@ -133,5 +139,31 @@ export const GraphQLPolicies = new GraphQLObjectType({
...
+          account.HostCollectiveId && account.HostCollectiveId !== account.id
+            ? await req.loaders.Collective.byId.load(account.HostCollectiveId)
+            : account;
+        let thresholds = await getPolicy(host, POLICIES.CONTRIBUTOR_INFO_THRESHOLDS, req);
+        let fxRate = 1;
+        if (!thresholds || isEmpty(thresholds)) {
+          fxRate = await getFxRate('USD', account.currency);
+          thresholds = DEFAULT_USD_THRESHOLDS;
+        } else if (host.currency !== account.currency) {
+          fxRate = await getFxRate(host.currency, account.currency);
```
Betree, line 163: :ice_cream: Here, I'd recommend using `req.loaders.CurrencyExchangeRate.fxRate`, as it will save some network requests if we ever start fetching this in a list.

server/graphql/loaders/contributors.ts
```diff
@@ -17,4 +20,53 @@ export default {
...
+      since,
+      hostId,
+    }: {
+      hostId: number;
+      since: Date | string;
+    }): DataLoader<
+      number,
+      { CollectiveId: number; amount: number; currency: SupportedCurrency; HostCollectiveId: number }
+    > => {
+      const key = compact([hostId, since]).join('-');
```
Betree, line 35: I don't understand the benefit of using `buildLoader` rather than just a regular loader here?

Betree (approved):

server/graphql/loaders/contributors.ts
kewitz: Not sure if I'm tripping, but isn't this required for sharing the `since` and `hostId` parameters across multiple DataLoader calls.

Betree: That's not the way `DataLoader` is meant to be used: DataLoaders automatically cache input, so that if you do:

```ts
loader.load(1);
loader.load(1);
```

You'll only receive a single call, with the argument `[1]`. 

Now, you probably got confused by the fact that you're passing an object here, and dataloaders will indeed fail to properly cache them. That's where the [`cacheKeyFn`](https://github.com/graphql/dataloader?tab=readme-ov-file#new-dataloaderbatchloadfn--options) function enters the game; see the `generateFxRateLoader` for example:

```ts
export const generateFxRateLoader = (): DataLoader<CurrencyFxRateRequest, number> => {
  return new DataLoader(
    async (requestedCurrencies: CurrencyFxRateRequest[]) => {
      const fxRates = await loadFxRatesMap(requestedCurrencies);
      return requestedCurrencies.map(
        request => fxRates[request.date || 'latest'][request.fromCurrency][request.toCurrency],
      );
    },
    {
      // Since the argument is an object, we need a custom serializer to be able to use it in the cache
      cacheKeyFn: arg => `${arg.fromCurrency}-${arg.toCurrency}`,
    },
  );
};
```

kewitz: These are not equivalent @Betree. Here I'm instantiating a single loader for each Host/Since because the query is written in that context. There's no way to simply use cacheKeyFn without either: querying the DB multiple times or rewriting the query as a chained OR conditionals.

Betree: I understand your point, and I'm not asking for a rewrite here. Your approach makes sense. To clarify a few points:

> There's no way to simply use cacheKeyFn without either: querying the DB multiple times 

This is already the case right now; if you call this loader with different hosts it will trigger multiple SQL requests. The "native" way to achieve the same behavior with raw data loaders, without changing the SQL query, would have been to group by `hostId`/`since` at the start the loader resolver, before triggering the requests. If `cacheKeyFn` is properly set to include all params, it won't re-query the DB if something has already been fetched.

Both achieve the same result, which is why I'm not arguing for a rewrite. I just wanted to express that this `buildLoader` adds a new level of complexity to solve a problem (caching) that `DataLoader` already solves internally given the right parameters.

kewitz: I did rewrite it:
https://github.com/opencollective/opencollective-api/pull/10703/commits/e53e5853dfe6519f83a748ca1323418f0930c6ce

kewitz: Well, I just noticed this disables a nice index scan optimization in the DB... I'll give another pass on this after lunch.

PR #10663, opened 2025-01-27
chore: Cleanup emails & use OSC onboarding templates

Betree:
As discussed in https://oficonsortium.slack.com/archives/GFH4N961L/p1736775468293559 & according to https://docs.google.com/document/d/1pht3scHj8Ly1CgmXSX3WbTlGdv9I1amNB1XsLX2RNDA/edit?tab=t.0.


gracepowers:
Thank you, Benjamin! One requested update: could you please remove the second bullet asking people to take hte onboarding survey? Screenshot to the bullet below: Screenshot 2025-01-27 at 9 07 31 AM

We were receiving spam responses from collectives that we do not host, so we’re no longer using that link. Instead, we’re sharing the survey manually with new collectives.

Betree:
Done in https://github.com/opencollective/opencollective-api/pull/10663/commits/5b2d6b8c7e2ef96e53826da3e6dcb32b6f03ff96

hdiniz (approved):

PR #10694, opened 2025-02-12
Dont call fixer api on e2e test

hdiniz:
Small optimizations to e2e runtime. 

- Fixer API is being queried with an invalid api key on e2e:

  > error: [Sentry fallback] FX Rate query issue: You have not supplied a valid API Access Key. [Technical Support: support@apilayer.com] (access_key=UNDEFINED_FIXER_ACCESS_KEY&base=BRL&symbols=USD) (undefined)

  Disabled by checking e2e environment flag, as e2e is currently executing using OC_ENV=ci and the unit/graphql tests do use the fixer client.

- man-db apt trigger is taking about a minute or so and its unnecessary

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -16,6 +16,7 @@ env:
   E2E_TEST: 1
   PGHOST: localhost
   PGUSER: opencollective
+  CYPRESS_COVERAGE: false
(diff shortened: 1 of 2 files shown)


Betree (approved):
Great change!

PR #10702, opened 2025-02-13
Make invoice file separate field

hdiniz:
Related: https://github.com/opencollective/opencollective/issues/7726

Add invoice expense attachments as its own field on the expense object, differentiating from other attachment types.


server/graphql/common/expenses.ts
```diff
@@ -2367,7 +2397,10 @@ export async function editExpense(
   // before the `canEditExpense` permissions check, because `editOnlyTagsAndAccountingCategory` has its
   // own permissions checks that are more permissive (e.g. tags can be edited even if the expense is paid)
   const modifiedFields = omitBy(expenseData, (_, key) => key === 'id' || !isValueChanging(expense, expenseData, key));
-  if (Object.keys(modifiedFields).every(field => ['tags', 'accountingCategory'].includes(field))) {
+  if (
+    isUndefined(expenseData.invoiceFile) &&
```
gustavlrsn, line 2401: Is there a reason this is not covered by the check on modifiedFields below?

gustavlrsn (approved):
LGTM!

hdiniz:
> Getting this error when used with the mock uploads locally:
> Screenshot 2025-02-24 at 12 38 29 Screenshot 2025-02-24 at 12 38 19

This only works without mocked uploads, we should be deprecating that asap. I added a few changes here to use minio on e2e tests for that reason. I will try to remove it entirely.

.github/workflows/e2e.yml
```diff
@@ -58,6 +66,14 @@ jobs:
           - 5432:5432
         # needed because the postgres container does not provide a healthcheck
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+      minio:
+        image: minio/minio:edge-cicd
```
Betree, line 70: I initially wanted to use this image as well in https://github.com/opencollective/opencollective-api/pull/10688, but it [hasn't been updated in 3 years](https://hub.docker.com/layers/minio/minio/edge-cicd/images/sha256-29e8e51691d11e779468f275002779b221fd3902518d103e35c8a8bb2ef0f3ea) and that was causing some issues in my tests (I don't remember where), which is why I went with the `npm run minio &` approach.

No blocker for me if the image is working here, we can always update later if needed.

server/graphql/common/expenses.ts
```diff
@@ -2237,7 +2261,10 @@ const isValueChanging = (expense: Expense, expenseData: Partial<ExpenseData>, ke
   const value = expenseData[key];
   if (key === 'accountingCategory') {
     return !isUndefined(value) && (value?.id ?? null) !== expense.AccountingCategoryId;
-  } else {
+  } else if (key === 'invoiceFile') {
+    return !isUndefined(value) && !isEqual(value, expense[key]);
+  }
+  {
```
Betree, line 2267: Missing `else`?

```suggestion
  } else if (key === 'invoiceFile') {
    return !isUndefined(value) && !isEqual(value, expense[key]);
  } else {
```

Betree (commented):
The context of this change is not clear, can we link a GitHub issue or add a description?

PR #10703, opened 2025-02-13
Contribution Thresholds: API Fixes

kewitz:


server/graphql/v2/object/Policies.ts
```diff
@@ -12,7 +12,7 @@ import { getIdEncodeResolver, IDENTIFIER_TYPES } from '../identifiers';
 import { GraphQLAmount } from './Amount';
 
 const DEFAULT_USD_THRESHOLDS: Policies[POLICIES.CONTRIBUTOR_INFO_THRESHOLDS] = {
-  address: 5000e2,
+  address: 500e2,
```
Betree, line 15: According to https://github.com/opencollective/opencollective-frontend/blame/ccea842ddedc8785c9e4c2cddd2f74bc96331bd4/components/contribution-flow/utils.js#L342, $5000 was the previous default. Why change that to $500 now?

kewitz: Ronen raised this while digging into this issue. The actual value should be $500.

kewitz:
Rebased.

server/graphql/v2/object/Policies.ts
Betree: Where did he gets his number from?

While searching online, I stumbled on [this update](https://www.irs.gov/newsroom/irs-provides-transition-relief-for-third-party-settlement-organizations-form-1099-k-threshold-is-5000-for-calendar-year-2024) from the IRS:
> TPSOs will be required to report transactions when the amount of total payments for those transactions is more than $5,000 in 2024; more than $2,500 in 2025; and more than $600 in calendar year 2026 and after.

PR #10711, opened 2025-02-18
Return exchange rate  date key with same utc date as request

hdiniz:
Fixes an error where the exchange rate is requested in a UTC date and is returned into another UTC date by API.

image

image


Betree (approved):
:+1:

PR #10726, opened 2025-02-25
Payments Methods: Redesign settings dashboard

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7837

### DIFF
--- a/server/models/PayoutMethod.ts
+++ b/server/models/PayoutMethod.ts
@@ -218,7 +218,14 @@ class PayoutMethod extends Model<InferAttributes<PayoutMethod>, InferCreationAtt
     const expenses = await (sequelize.models.Expense as typeof Expense).findOne({
       where: {
         PayoutMethodId: this.id,
-        status: { [Op.notIn]: [ExpenseStatuses.PENDING, ExpenseStatuses.DRAFT] },
+        status: {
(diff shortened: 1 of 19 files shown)


server/models/PayoutMethod.ts
```diff
@@ -213,6 +213,17 @@ class PayoutMethod extends Model<InferAttributes<PayoutMethod>, InferCreationAtt
       return [];
     }
   }
+
+  async canBeEditedOrDeleted(): Promise<boolean> {
+    const expenses = await (sequelize.models.Expense as typeof Expense).findOne({
+      where: {
+        PayoutMethodId: this.id,
+        status: { [Op.notIn]: [ExpenseStatuses.PENDING, ExpenseStatuses.DRAFT] },
```
hdiniz, line 221: Do you think it would make sense to include REJECT or CANCELED as well?

kewitz: I think the question here was related to what status I know for sure we don't care about editing. With the existing expense flow, I wonder if something would end up in canceled before being edited to the correct state. But perhaps if the user submits to the wrong collective and needs to re-submit to another collective and change something in the Payout Method details, that would make sense.

kewitz:
Rebased.

PR #10732, opened 2025-03-03
enhancement(RetentionPolicy): differenciate expense comments

Betree:


cron/daily/96-apply-data-retention-policy.ts
```diff
@@ -18,8 +18,36 @@ enum RetentionPeriod {
...
+      }
+    }
+  }
+
+  return JSON.stringify(result);
+};
+
+const MODEL_RETENTION_PERIODS: ModelRetentionPeriodSettings[] = [
+  [models.Comment, RetentionPeriod.FINANCIAL, { ExpenseId: { [Op.not]: null } }],
+  [models.Comment, RetentionPeriod.SENSITIVE, { ExpenseId: { [Op.is]: null } }],
```
kewitz, line 50: I think we should be more explicit about which ids we're expecting to be set instead of querying for "everything that is not an expense comment".

Betree: I agree, pushing an update in https://github.com/opencollective/opencollective-api/pull/10748

PR #10754, opened 2025-03-21
Add separate bash script for `docker-compse`

SudharakaP:
Related Discussion at, https://github.com/opencollective/opencollective-frontend/pull/11092?notification_referrer_id=NT_kwDOAL3B_bQxNTMxNDUzMDU5OToxMjQzNTk2NQ&notifications_query=reason%3Aparticipating#discussion_r2003246052

Currently `npm run mailpit` fails with;

```
> mailpit
> ./scripts/dev/run-docker.sh compose -f docker-compose.dev.yml --profile mail up

unknown shorthand flag: 'f' in -f

Usage:  docker [OPTIONS] COMMAND [ARG...]

Run 'docker --help' for more information
```


scripts/dev/run-docker-compose.sh
```diff
@@ -0,0 +1,15 @@
...
+  echo "Docker Compose is not installed. See https://docs.docker.com/compose/install/."
+  exit 1
+fi
+
+if docker-compose version &>/dev/null; then
+  docker-compose "$@"
+else
+  echo "Docker Compose requires root privileges. Running with sudo: docker $@"
+  sudo docker-compose "$@"
+fi
```
Betree, line 15: The root issue is that some people choose to install `docker-compose` as a standalone command, while others like me install it as a plugin (`docker compose`) - see https://docs.docker.com/compose/install/. Would be great if the script could support the two:
1. Check if `docker-compose` is installed (`which docker-compose`)
2. If the command exists, run it. Retry with root if necessary.
3. If not existing, check if `docker compose` is supported. If so, run it with `./run-docker.sh`
4. If none of this is supported, show the install error message.

Betree (commented):
I can take care of implementing the `docker compose` support, since it's how its setup on my machine. LMK.

SudharakaP:
> I can take care of implementing the `docker compose` support, since it's how its setup on my machine. LMK.

Yes please. I don't have docker compose as the plugin and don't use it that way so perhaps you can add that part better 😉

Betree (approved):

PR #10768, opened 2025-04-01
debt(Account): remove legacy condition for hosts payout methods

Betree:


kewitz (approved):
The update to the payout methods resolver looks good.
The template update broke some tests, but it looks good. I guess "wants to send funds" sounds generic enough to cover all use cases.

PR #10771, opened 2025-04-02
Performance: Fix timeline performance issues

kewitz:
- [x] Fix issue with locking mechanism that could spawn multiple queries for the same timeline
- [x] Add an event horizon to the timeline, avoiding scanning the whole Activities table in case the user does not have enough activities to reach the limit

### DIFF
--- a/migrations/20250403170000-add-missing-activity-timeline-index.js
+++ b/migrations/20250403170000-add-missing-activity-timeline-index.js
@@ -1,0 +1,18 @@
+'use strict';
+
+module.exports = {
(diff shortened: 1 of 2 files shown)


Betree (approved):
Would be great if we could accept a parameter for the horizon date in the API. That way you would be able preload the page and click "Load more" if you want to dig.
The risk of the current implementation: logging into a recently unused account, it would look like there is no history.

 That said, it is not a priority - I'm fine with this shortcut if it improves performance.

PR #10788, opened 2025-04-09
Type Loaders

kewitz:


Betree (approved):
The `server/graphql/loaders/index.ts` file is difficult to compare, but it looks good overall. Thanks for taking care of that tedious but important task!

PR #10793, opened 2025-04-11
Refact: Get rid of Sequelize.define models

kewitz:


server/models/Member.ts
```diff
@@ -24,55 +25,114 @@ const invalidateContributorsCacheUsingInstance = instance => {
...
+  declare public description: string;
+  declare public publicMessage: string;
+  declare public since: Date;
+  declare public createdAt: Date;
+  declare public updatedAt: Date;
+  declare public deletedAt?: Date;
+
+  declare public tier?: Tier;
+  declare public memberCollective?: Collective;
+  declare public collective?: Collective;
```
Betree, line 44: The way to type these is usually:

```suggestion
  declare public tier?: BelongsToGetAssociationMixin<Tier>;
  declare public memberCollective?: BelongsToGetAssociationMixin<Collective>;
  declare public collective?: BelongsToGetAssociationMixin<Collective>;
```

Is this missing bit the reason why you had to do:
```diff
- member.collective = collective
+ member['collective'] = collective
```

in https://github.com/opencollective/opencollective-api/pull/10793/files#diff-e0233b2efe7b6d82ca4f3d789b6c2cc07e74b88d94a3eec9bdcd51f058dcdfe9?

Betree (approved):
Lgtm, are we making this change for consistency? Typings?

server/models/Member.ts
kewitz: Thanks for the comment. These are actually declared with `NonAttribute<>`, `BelongsToGetAssociationMixin<>` is used for `getTier` and `getCollective` declarations.

Betree: True, my bad!

kewitz:
@Betree for overall consistency. This was bothering me a bit while I was typing loaders using generic types.

PR #10807, opened 2025-04-21
Wise: Multiple User accounts

kewitz:


server/paymentProviders/utils.ts
```diff
@@ -291,3 +292,6 @@
 
   return vendor;
 };
+
+export const hashObject = (obj: object) =>
+  crypto.createHash('sha1').update(JSON.stringify(obj)).digest('hex').slice(0, 7);
```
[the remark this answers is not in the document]
kewitz: This is fine for what we're using, and we cannot simply now.

muhdaizwan (changes requested):
MUHAMMAD AIZWAN BIN ODZIR.

server/paymentProviders/utils.ts
```diff
@@ -291,3 +293,6 @@ export const getOrCreateVendor = async (vendorProviderId, vendorName) => {
...
```
Betree, line 298: @kewitz what prevents us from using a different hashing algorithm?

server/graphql/v2/object/ConnectedAccount.ts
```diff
@@ -31,5 +34,47 @@ export const GraphQLConnectedAccount = new GraphQLObjectType({
     },
     settings: { type: GraphQLJSON },
     service: { type: new GraphQLNonNull(GraphQLConnectedAccountService) },
+    accountsMirrored: {
+      type: new GraphQLList(GraphQLAccount),
```
Betree, line 38: Doesn't matter much for now, but that will become an issue when we'll enable TS strict mode

```suggestion
      type: new GraphQLNonNull(new GraphQLList(GraphQLAccount)),
```

server/paymentProviders/utils.ts
kewitz: Just the existing hashes. We could write a migration to recalculate those hashes.

Betree: Ok! It's fine to do it later on, then.

Betree (approved):

PR #10831, opened 2025-04-29
feat(GraphQL): Refined Expense permissions

gustavlrsn:
For https://github.com/opencollective/opencollective/issues/7953

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -536,6 +536,81 @@ export const canEditExpense: ExpensePermissionEvaluator = async (
   }
 };
 
+const createPermissionEvaluator = (
+  permissionMatrix: Partial<Record<ExpenseStatuses, ExpensePermissionEvaluator[]>>,
+  errorMessage = 'You do not have permission for this action',
+): ExpensePermissionEvaluator => {
+  return async (req: express.Request, expense: Expense, options = { throw: false }) => {
+    if (!validateExpenseScope(req, options)) {
+      if (options?.throw) {
+        throw new Forbidden(
+          'Your current token is missing the necessary scope',
+          EXPENSE_PERMISSION_ERROR_CODES.INVALID_SCOPE,
+        );
+      }
+      return false;
+    }
+
+    // Check if the status exists in our matrix
+    if (!permissionMatrix[expense.status]) {
+      if (options?.throw) {
+        throw new Forbidden(
+          `Cannot perform this action in current status (${expense.status})`,
+          EXPENSE_PERMISSION_ERROR_CODES.UNSUPPORTED_STATUS,
+        );
+      }
+      return false;
+    }
+
+    const permissions = permissionMatrix[expense.status];
+    // Check each role independently
+    const hasPermission = remoteUserMeetsOneCondition(req, expense, permissions, options);
+
+    if (!hasPermission && options?.throw) {
+      throw new Forbidden(errorMessage, EXPENSE_PERMISSION_ERROR_CODES.MINIMAL_CONDITION_NOT_MET);
+    }
+
+    return hasPermission;
+  };
+};
+
+export const canEditTitle = createPermissionEvaluator(
+  {
+    PENDING: [isOwner, isCollectiveAdmin],
+    APPROVED: [isHostAdmin],
+    INCOMPLETE: [isOwner, isCollectiveAdmin, isHostAdmin],
+  },
+  'You do not have permission to edit the title for this expense',
+);
+
+export const canEditType = createPermissionEvaluator({
+  PENDING: [isOwner, isCollectiveAdmin],
+  APPROVED: [isHostAdmin],
+  INCOMPLETE: [isOwner, isCollectiveAdmin, isHostAdmin],
+});
+
+export const canEditPaidBy = createPermissionEvaluator({
+  PENDING: [isOwner, isCollectiveAdmin],
+  APPROVED: [isHostAdmin],
+  INCOMPLETE: [isOwner, isHostAdmin],
+});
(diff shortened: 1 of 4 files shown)


Betree (commented):
The GraphQL part looks perfect.

The declarative approach is shorter in length, and I recognize that not having to re-check the scope in each helper + the automatic handling of `options` are major improvements that we should try to keep regardless of how we refactor this.

As I've shared previously, I'm however not a fan of the underlying library refactor as it makes the condition less explicit (you may have to dig into the helper functions to confirm their behaviors). The fact that tests were implemented declaratively while the logic itself stayed imperative was also a good mechanism to make sure we end up with the same results even when changing the coding paradigm.

This file is well tested, so this is not a blocking feedback from my end, just a call to make sure we consider all aspects of the change. If we do go with a declarative approach, I will advocate making the DSL as explicit as possible. Taking this snippet as an example:

image

~~Using an attribute name would make it clearer to me that the first arg is supposed to match the expense attributes. And that way, we can also make it optional. Something like:~~

```es6
export const canEditExpenseTags = evaluateFirstMatch([
  {
    match: { status: ExpenseStatus.PAID },
    condition: (req, expense, options) => remoteUserMeetsOneCondition(req, expense, [isHostAdmin, isCollectiveAdmin], options),
  },
  {
    condition: (req, expense, options) =>
      remoteUserMeetsOneCondition(req, expense, [isOwner, isOwnerAccountant, isHostAdmin, isCollectiveAdmin], options),
  },
]);
```

And I've just realized after writing this that the second condition would not execute if the first one (`status: PAID`) matches, which is not what I understood when first reading the above helpers. I was under the assumption that the helper would stop at the first condition that returns true, and continue otherwise. My fault was to avoid reading the code for `evaluateFirstMatch` first, but this is exactly my concern: you now need to learn the behavior of this helper to grasp the code, when the sole understanding of Javascript was previously sufficient.

Practically, you currently need to learn that the first argument (fields matcher) will void the other conditions while the matching function is simply ignored if it returns false. There is nothing in the current API telling me that. I need to guess by looking at multiple evaluators - and eventually dig into the helper.

I advise reviewing the API to make sure someone that don't have access to the code of `evaluateFirstMatch` can still use it with 100% confidence, without any ambiguity, solely by reading the condition. I wouldn't be that picky with other parts of the code, but expense permissions evaluators are too sensitive (in terms of security) to allow for ambiguity. 

**Suggestions**

- Keep things imperative, but put everything that is common in a helper _(this one is actually more concise than the proposed declarative approach)_

```ts
const createExpensePermissionsEvaluator = (evaluator) => (req, expense, options = { throw: false }) => {
  if (!validateExpenseScope(req, options)) {
    if (options?.throw) {
      throw new Forbidden(
        'Your current token is missing the necessary scope',
        EXPENSE_PERMISSION_ERROR_CODES.INVALID_SCOPE,
      );
    }
    return false;
  } else if (!canUseFeature(req.remoteUser, FEATURE.USE_EXPENSES)) {
    if (options?.throw) {
      throw new Forbidden('User cannot edit expense tags', EXPENSE_PERMISSION_ERROR_CODES.UNSUPPORTED_USER_FEATURE);
    }
    return false;
  } else {
    return evaluator(req, expense, options);
  }
}

const canEditExpenseTags = createExpensePermissionsEvaluator((req, expense, options) => {
  if (expense.status === ExpenseStatus.PAID) {
    return remoteUserMeetsOneCondition(req, expense, [isHostAdmin, isCollectiveAdmin], options);
  } else {
    return remoteUserMeetsOneCondition(req, expense, [isOwner, isOwnerAccountant, isHostAdmin, isCollectiveAdmin], options);
  }
});
```

- Declarative, but more explicit about the execution flow

```ts
const canEditExpenseTags = evaluateUntilFirstMatch([
  {
    match: { status: ExpenseStatus.PAID },
    evaluator: (req, expense) => remoteUserMeetsOneCondition(req, expense, [isHostAdmin, isCollectiveAdmin]),
  },
  {
    evaluator: (req, expense) => remoteUserMeetsOneCondition(req, expense, [isOwner, isOwnerAccountant, isHostAdmin, isCollectiveAdmin]),
  },
]);
```

gustavlrsn:
@Betree thanks for the review! I reverted back to a strictly imperative approach, even without a helper to check scope and canUseFeature, since for some permissions like `canEditExpense` we seem to not care about `canUseFeature` for all statuses of the expense, so did not want to introduce any new behavior.

I also realized there is three types of expense features, and the helper only checked the first of these (which felt a bit narrow for a general `createExpensePermissionEvaluator` function)

```
  /** Whether this profile can submit expenses */
  USE_EXPENSES = 'USE_EXPENSES',
  /** Whether this profile can receive expenses */
  RECEIVE_EXPENSES = 'RECEIVE_EXPENSES',
  /** Whether this account can use multi-currency expenses */
  MULTI_CURRENCY_EXPENSES = 'MULTI_CURRENCY_EXPENSES',
  ```
  
 So  I'm keeping all existing permission checkers as is, and have just added the new ones in a similar fashion.

Betree (approved):
Looking good, I suggest adding tests in `test/server/graphql/common/expenses.test.js` to make sure the new checks are covered.

I'll follow up on your comments about the feature check. There were also good things in your proposal that simplified the implementation which I would love to bring back at some point.

gustavlrsn:
> Looking good, I suggest adding tests in `test/server/graphql/common/expenses.test.js` to make sure the new checks are covered.
> 
> I'll follow up on your comments about the feature check. There were also good things in your proposal that simplified the implementation which I would love to bring back at some point.

I've added tests, and made a final check against the permissions matrix in Coda. Thank you!

PR #10855, opened 2025-05-15
feat(Move expense): Add mutation to move expenses within the same Collective

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/7980 https://github.com/opencollective/opencollective/issues/7955

Note: This PR currently targets the extended expense edit permissions branch since those are used in the resolver

### Description
Adds a `moveExpense` mutation to move an expense within a Collective.

The new mutation resolver does the permissions checks, and uses the general `moveExpenses` function to perform the operation. 

However, this means we are removing the user root check from that function, which is already handled by the `moveExpenses` root mutation resolver, but has been there as straps and braces to prevent accidental unauthorized usage of the function by developers. I'm adding a DANGER comment to note that this functions needs pre-authorization/permissions checks.


gustavlrsn:
> Looking good!

Thanks for the review, fixed your comments and also added some tests for this mutation

PR #10857, opened 2025-05-15
feat(email): configuration for amazon SES

Betree:


hdiniz (approved):

Betree:
To be rebased and tested with the official SES integration after the NodeMalier v7 update, which may resolve the issue I faced when trying it locally: https://github.com/opencollective/opencollective-api/pull/10872

PR #10859, opened 2025-05-16
enhancement(sendEmail): better error messages

Betree:
See https://github.com/opencollective/opencollective/pull/7983


server/controllers/services/email.js
```diff
@@ -92,7 +93,25 @@ export const messageSupport = async (req, res) => {
...
+        ip,
+      },
+    });
+
+    res.status(500).send({
+      message: 'Error while opening the support ticket, please reach out directly to support@opencollective.com',
+    });
+  }
 
   res.status(200).send({ sent: true });
```
kewitz, line 116: We need to move this into the try scope because we can't `res.status.send` twice.

Betree: Good catch, I forgot the `return`

PR #10861, opened 2025-05-16
Fix: Create missing HostApplications for unhosted collectives

kewitz:
Part of https://github.com/opencollective/opencollective/issues/7923

### DIFF
--- a/server/graphql/v2/object/Host.ts
+++ b/server/graphql/v2/object/Host.ts
@@ -1675,7 +1675,9 @@ export const GraphQLHost = new GraphQLObjectType({
               where: { HostCollectiveId: host.id, status: 'APPROVED' },
             });
             where.HostCollectiveId = { [Op.or]: [{ [Op.ne]: host.id }, { [Op.is]: null }] };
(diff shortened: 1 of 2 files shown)


server/graphql/v2/object/Host.ts
```diff
@@ -1675,7 +1675,9 @@ export const GraphQLHost = new GraphQLObjectType({
               where: { HostCollectiveId: host.id, status: 'APPROVED' },
             });
             where.HostCollectiveId = { [Op.or]: [{ [Op.ne]: host.id }, { [Op.is]: null }] };
-            where.id = collectiveIds.map(({ CollectiveId }) => CollectiveId);
+            const id = collectiveIds.map(({ CollectiveId }) => CollectiveId);
+            // @ts-expect-error Type 'unique symbol' cannot be used as an index type. Not sure why TS is not happy here.
+            where[Op.and].push({ [Op.or]: [{ id: id }, { ParentCollectiveId: id }] });
```
Betree, line 1680: I've also seen this one before, and haven't found a better way to address it yet.

Betree (approved):
Looking great :+1:

PR #10874, opened 2025-05-19
Make sure there is a matching transaction in missingMembers fix

znarf:


Betree (approved):
Surprising that we have PAID/ACTIVE orders without transactions, adding that to my list of things to investigate.

znarf:
In my first round of investigation, I found:
- some cases where the transactions were deleted but not the orders
- some cases where transactions were existing but with non matching CollectiveId / FromCollectiveId

Nothing recent came to my attention.

PR #10882, opened 2025-05-21
debt: Remove legacy PDF service config

Betree:
Part of https://github.com/opencollective/opencollective/issues/7987


ladygotit60-commits:
https://github.com/opencollective/opencollective/issues/7987

PR #10883, opened 2025-05-21
fix(Update): skip unsubscribed users

Betree:
For https://opencollective.freshdesk.com/a/tickets/381332


server/lib/notifications/email.ts
```diff
@@ -108,6 +108,7 @@ export const notify = {
       type: activity.type,
       CollectiveId: options?.collective?.id || activity.CollectiveId,
       channel: Channels.EMAIL,
+      UserId: users.map(u => (typeof u === 'number' ? u : u.id)),
```
Betree, line 111: Optimization to prevent fetching all users when sending only to a few

kewitz (approved):

PR #10890, opened 2025-05-23
Consider currency when filtering Expenses by amount

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7979

- [ ] Extend the same logic to Transactions

### DIFF
--- a/server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
+++ b/server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
@@ -29,6 +29,8 @@ import {
   fetchAccountWithReference,
   GraphQLAccountReferenceInput,
 } from '../../input/AccountReferenceInput';
+import { getValueInCentsFromAmountInput } from '../../input/AmountInput';
+import { AmountRangeInputType, GraphQLAmountRangeInput } from '../../input/AmountRangeInput';
 import {
   CHRONOLOGICAL_ORDER_INPUT_DEFAULT_VALUE,
   GraphQLChronologicalOrderInput,
@@ -169,13 +171,19 @@ export const ExpensesCollectionQueryArgs = {
     description: 'The order of results',
     defaultValue: CHRONOLOGICAL_ORDER_INPUT_DEFAULT_VALUE,
   },
+  amount: {
+    type: GraphQLAmountRangeInput,
+    description: 'Only return expenses that match this amount range',
+  },
   minAmount: {
     type: GraphQLInt,
     description: 'Only return expenses where the amount is greater than or equal to this value (in cents)',
+    deprecate: '2025-05-26: Please use amount instead',
   },
   maxAmount: {
     type: GraphQLInt,
     description: 'Only return expenses where the amount is lower than or equal to this value (in cents)',
+    deprecate: '2025-05-26: Please use amount instead',
   },
   payoutMethodType: {
     type: GraphQLPayoutMethodType,
@@ -274,7 +282,7 @@ const loadAllAccountsFromArgs = async (
 
 export const ExpensesCollectionQueryResolver = async (
   _: void,
-  args,
+  args: Record<string, any> & { amount?: AmountRangeInputType },
   req: express.Request,
 ): Promise<CollectionReturnType & { totalAmount?: any }> => {
   const where = { [Op.and]: [] };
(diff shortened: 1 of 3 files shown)


server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -396,12 +401,43 @@ export const ExpensesCollectionQueryResolver = async (
   } else if (args.tag === null || args.tags === null) {
     where['tags'] = { [Op.is]: null };
   }
-  if (args.minAmount) {
-    where['amount'] = { [Op.gte]: args.minAmount };
-  }
-  if (args.maxAmount) {
-    where['amount'] = { ...where['amount'], [Op.lte]: args.maxAmount };
+  if (args.minAmount || args.maxAmount) {
+    console.log(args);
```
Betree, line 405:
```suggestion
```

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -543,7 +579,7 @@ export const ExpensesCollectionQueryResolver = async (
   const { offset, limit } = args;
 
   const fetchNodes = () => {
-    return Expense.findAll({ include, where, order, offset, limit });
+    return Expense.findAll({ include, where, order, offset, limit, logging: true });
```
Betree, line 582:
```suggestion
    return Expense.findAll({ include, where, order, offset, limit });
```

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -396,12 +401,43 @@ export const ExpensesCollectionQueryResolver = async (
...
+      args.minAmount && args.maxAmount
+        ? { [Op.between]: [args.minAmount, args.maxAmount] }
+        : args.minAmount
+          ? { [Op.gte]: args.minAmount }
+          : { [Op.lte]: args.maxAmount };
+
+    if (args.amountCurrency) {
+      console.log(
+        `Filtering expenses by amount in ${args.amountCurrency} is deprecated. Please use the currency field instead.`,
+      );
```
Betree, line 416: Leftover? I don't really get this one

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -396,12 +401,43 @@ export const ExpensesCollectionQueryResolver = async (
...
+  if (args.minAmount || args.maxAmount) {
+    console.log(args);
+    const operator =
+      args.minAmount && args.maxAmount
+        ? { [Op.between]: [args.minAmount, args.maxAmount] }
+        : args.minAmount
+          ? { [Op.gte]: args.minAmount }
+          : { [Op.lte]: args.maxAmount };
+
+    if (args.amountCurrency) {
```
Betree, line 413: If `amountCurrency` is not provided but we're still filtering on `minAmount`/`maxAmount`, I would still default to a given currency (the host's?), to make sure we resolve the issue consistently, even for people not using this new flag.

Betree (commented):
If we meant to deprecate `minAmount` and `maxAmount`, then I would rather introduce `min` and `max` (or better names) as `GraphQLAmountInput`. 

But this approach has the benefit to be simple, so I'm ok with it too.

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -396,12 +404,51 @@ export const ExpensesCollectionQueryResolver = async (
...
   }
-  if (args.minAmount) {
-    where['amount'] = { [Op.gte]: args.minAmount };
-  }
-  if (args.maxAmount) {
-    where['amount'] = { ...where['amount'], [Op.lte]: args.maxAmount };
+
+  if (args.amount?.gte || args.amount?.lte) {
+    if (args.amount.gte && args.amount.lte) {
+      assert(args.amount.gte.currency === args.amount.lte.currency, 'Amount range must have the same currency');
```
Betree, line 410: That's pure evil
> { gte: { value: 10, currency: 'USD' }, lte: { value: 100: currency: 'INR' } }

giphy

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -396,12 +404,51 @@ export const ExpensesCollectionQueryResolver = async (
...
+
+    where[Op.and].push(
+      sequelize.where(
+        sequelize.literal(`
+            CASE
+              WHEN "Expense"."data" #>> '{quote,sourceCurrency}' = '${currency}'
+                AND "Expense"."data" #>> '{quote,targetCurrency}' = "Expense"."currency"
+                THEN 1.0 / ("Expense"."data" #> '{quote,rate}')::NUMERIC
+              WHEN "Expense"."data" #>> '{quote,sourceCurrency}' = "Expense"."currency"
+                AND "Expense"."data" #>> '{quote,targetCurrency}' = '${currency}'
```
Betree, line 432: Not a blocker since `currency` comes from an enum, but it's better to use query variables to prevent any risks of SQL injection.

Betree (approved):
Consistent API, backward-compatible. Looks great :ok_hand: 

I wonder about the performance impact, have you checked with prod data yet?

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
kewitz: If someone forges a query like that, I'll consider it a DoS attempt and ban them from the platform.

PR #10899, opened 2025-05-27
Only consider net amount to collective when checking for balance

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/7969


Betree (approved):
Looks perfect!

PR #10904, opened 2025-05-27
Add field to query for expense payee history

hdiniz:
Related https://github.com/opencollective/opencollective/issues/7935


kewitz (changes requested):
This smells a bit. We should be going through `Expense.fromAccount.expenses` since the expense history is a property of the account and not of the expense object.

PR #10912, opened 2025-05-30
Implement PKCE correctly and align token responses with RFC6749

ThisIsMissEm:
This pull request adds the necessary properties to `OAuthAuthorizationCodes` to support PKCE, per the last paragraph in https://node-oauthoauth2-server.readthedocs.io/en/master/misc/pkce.html (essentially two nullable string columns: `codeChallenge` and `codeChallengeMethod`)

This doesn't "add" PKCE, since it already existed, this just makes it work as expected.

It also fixes the `/oauth/token` response to return a compliant Bearer token response, per [RFC6749](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1). The returned `accessToken` is still a JWT, it just now also returns the `token_type` property and the `scope` property based on the UserToken's actual scope value. It does not return a refresh token (it's unclear to me if these are actually generated).

Tests should all pass, I think, but I couldn't run them locally.

### DIFF
--- a/migrations/20250530202548-add_pkce_to_oauth_authorization_code.js
+++ b/migrations/20250530202548-add_pkce_to_oauth_authorization_code.js
@@ -1,0 +1,21 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.addColumn('OAuthAuthorizationCodes', 'codeChallenge', {
+      type: Sequelize.STRING,
+      allowNull: true,
+    });
+
+    await queryInterface.addColumn('OAuthAuthorizationCodes', 'codeChallengeMethod', {
+      type: Sequelize.STRING,
+      allowNull: true,
+    });
+  },
+
+  async down(queryInterface, Sequelize) {
+    await queryInterface.removeColumn('OAuthAuthorizationCodes', 'codeChallenge');
+    await queryInterface.removeColumn('OAuthAuthorizationCodes', 'codeChallengeMethod');
+  },
+};
--- a/migrations/20250530202548-add_pkce_to_oauth_authorization_code.js
+++ b/migrations/20250530202548-add_pkce_to_oauth_authorization_code.js
@@ -14,7 +14,7 @@ module.exports = {
     });
   },
 
-  async down(queryInterface, Sequelize) {
+  async down(queryInterface) {
     await queryInterface.removeColumn('OAuthAuthorizationCodes', 'codeChallenge');
     await queryInterface.removeColumn('OAuthAuthorizationCodes', 'codeChallengeMethod');
   },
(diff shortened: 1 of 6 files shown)


server/lib/oauth/index.ts
```diff
@@ -19,20 +20,16 @@ class CustomTokenHandler extends TokenHandler {
...
+        scope: 'oauth',
         // eslint-disable-next-line camelcase
-        return { access_token: accessToken };
+        access_token: model.accessToken,
       },
-    };
+      auth.TOKEN_EXPIRATION_SESSION_OAUTH, // 90 days,
+    );
+
+    return BearerTokenType(accessToken, auth.TOKEN_EXPIRATION_SESSION_OAUTH, null, model.scope);
```
ThisIsMissEm, line 32: If you wanted to (and you actually support refresh tokens) then you could do the same JWT thing as above pretty much, except with `model.refreshToken` and the `refreshTokenLifetime` value which is just [hardcoded](https://github.com/opencollective/opencollective-api/blob/57a84179df3f59f078715a58ccc315171527a18b/server/lib/oauth/index.ts#L80) to `60 * 60 * 24 * 365`, and then replace the `null` with that refresh token JWT.

ThisIsMissEm:
Also, it'd be fantastic to implement [RFC8414](https://datatracker.ietf.org/doc/html/rfc8414), which would allow discovery of all the OAuth features you support, though it looks like your underlying OAuth library doesn't provide this, so it'd be a custom endpoint that would need to be added and aligned with whatever your OAuth library is doing.

ThisIsMissEm:
Here's the matching frontend change: https://github.com/opencollective/opencollective-frontend/pull/11263

Betree (commented):
Hi @ThisIsMissEm :wave: 

Thanks so much for this fantastic contribution! Your explanation is crystal clear, and the code is well-written. The test coverage looks comprehensive too.

> Tests should all pass, I think, but I couldn't run them locally.

Did you run into any specific issues with the setup? Missing docs or dependencies? Since we don't often go through the initial setup process ourselves, any feedback on what would have made it smoother would be super helpful for improving the developer experience.

FYI, the way I would test this would be:
> npm run test test/server/routes/oauth.test.ts test/server/lib/oauth/model.test.ts

(or `npm run test:watch` while developing).

Tests seem to be failing at the moment. Just let me know if you don't have more bandwidth to look at them now, I'll add that to my plate.

ThisIsMissEm:
> Did you run into any specific issues with the setup? 

Missing database? I couldn't get stuff setup according to docs, seemed to rely on your staging infra and I don't have access to that properly, so I figured I'd leave it to CI/CD and y'all to run the tests.

migrations/20250530202548-add_pkce_to_oauth_authorization_code.js
```diff
@@ -0,0 +1,21 @@
...
+      allowNull: true,
+    });
+
+    await queryInterface.addColumn('OAuthAuthorizationCodes', 'codeChallengeMethod', {
+      type: Sequelize.STRING,
+      allowNull: true,
+    });
+  },
+
+  async down(queryInterface, Sequelize) {
```
ThisIsMissEm:
```suggestion
  async down(queryInterface) {
```

ThisIsMissEm:
I'm going to close this PR and reopen via collaborate push to this repository instead of from mine.

PR #10917, opened 2025-06-04
Fix implementation of OAuth PKCE and token responses

ThisIsMissEm:
This fixes https://github.com/opencollective/opencollective/issues/7996
Supersedes #10912 to ensure all CI Actions run correctly.

## Token Responses

Changes the token response to return a correct `BearerTokenType` response, this automatically includes the `token_type` and access token's scopes (even though the JWT just uses the `oauth` scope)

The response changes from:
```json
{
  "access_token": "<JWT>"
}
```
To
```json
{
  "access_token": "<JWT>",
  "token_type": "Bearer",
  "expires_in": 7776000,
  "scope": ["email", "profile"]
}
```

The value `7776000` is 90 days, which matches the expiry of the JWT.

## PKCE

- Added the `codeChallenge` and `codeChallengeMethod` fields to the database model
- Correctly propagated those values through the OAuth "model"

## Scope Parsing

It turns out that the OAuth2Server library doesn't return a parsed `scope` value, and the value that gets passed into the OAuth "model"'s `saveAuthorizationCode` and `saveToken` methods can pretty much be anything, I would have expected the library to be parsing the scope parameter and passing it through correctly, but apparently that's a bug in that library, so I've added a workaround.

Future versions may fix this bug, but the [changelog](https://github.com/node-oauth/node-oauth2-server/blob/master/CHANGELOG.md) is missing a tonne of information on the change history, i.e., goes from 4.2.0 to 5.0.0, even though we're on 4.3.3

The workaround code handles the comma-separate scopes originally accepted, but also handles the spec-compliant `scope` parameter which is a space separated string.

>  The value of the scope parameter is expressed as a list of space-
>   delimited, case-sensitive strings.  The strings are defined by the
>   authorization server.  If the value contains multiple space-delimited
>   strings, their order does not matter, and each string adds an
>   additional access range to the requested scope.
>
>     scope       = scope-token *( SP scope-token )
>     scope-token = 1*( %x21 / %x23-5B / %x5D-7E )

## Frontend Changes

To make this fully work, some frontend changes are required, you can find these in:
https://github.com/opencollective/opencollective-frontend/pull/11263

I also have a fix that correctly redirects back to the application upon cancelling the OAuth request, which allows OAuth clients to correctly handle that scenario:
https://github.com/opencollective/opencollective-frontend/pull/11264

---

You could consider this a community contribution, but it did take time away from me working on open-source software (which is why people [donate to me](https://support.thisismissem.social))

### DIFF
--- a/server/lib/oauth/index.ts
+++ b/server/lib/oauth/index.ts
@@ -8,6 +8,7 @@ import BearerTokenType from '@node-oauth/oauth2-server/lib/token-types/bearer-to
 import { assign } from 'lodash';
 
 import * as auth from '../../lib/auth';
+import logger from '../logger';
 
 import model from './model';
 
@@ -218,6 +219,7 @@ const handleResponse = function (req, res, response) {
  */
 
 const handleError = function (e, req, res, response, next) {
+  logger.error(e);
   if (this.useErrorHandler === true) {
     next(e);
   } else {
--- a/server/lib/oauth/index.ts
+++ b/server/lib/oauth/index.ts
@@ -4,6 +4,7 @@ import OAuth2Server, { UnauthorizedRequestError } from '@node-oauth/oauth2-serve
 import InvalidArgumentError from '@node-oauth/oauth2-server/lib/errors/invalid-argument-error';
 import AuthorizeHandler from '@node-oauth/oauth2-server/lib/handlers/authorize-handler';
 import TokenHandler from '@node-oauth/oauth2-server/lib/handlers/token-handler';
+import BearerTokenType from '@node-oauth/oauth2-server/lib/token-types/bearer-token-type';
 import { assign } from 'lodash';
 
 import * as auth from '../../lib/auth';
@@ -19,20 +20,16 @@ class CustomTokenHandler extends TokenHandler {
   }
 
   getTokenType = function (model) {
-    return {
-      valueOf: () => {
-        const accessToken = model.user.jwt(
-          {
-            scope: 'oauth',
-            // eslint-disable-next-line camelcase
-            access_token: model.accessToken,
-          },
-          auth.TOKEN_EXPIRATION_SESSION_OAUTH, // 90 days,
-        );
+    const accessToken = model.user.jwt(
+      {
+        scope: 'oauth',
         // eslint-disable-next-line camelcase
-        return { access_token: accessToken };
+        access_token: model.accessToken,
       },
-    };
+      auth.TOKEN_EXPIRATION_SESSION_OAUTH, // 90 days,
+    );
+
+    return BearerTokenType(accessToken, auth.TOKEN_EXPIRATION_SESSION_OAUTH, null, model.scope);
   };
 }
 
--- a/server/lib/oauth/index.ts
+++ b/server/lib/oauth/index.ts
@@ -29,7 +29,7 @@ class CustomTokenHandler extends TokenHandler {
       auth.TOKEN_EXPIRATION_SESSION_OAUTH, // 90 days,
     );
 
-    return BearerTokenType(accessToken, auth.TOKEN_EXPIRATION_SESSION_OAUTH, null, model.scope);
+    return new BearerTokenType(accessToken, auth.TOKEN_EXPIRATION_SESSION_OAUTH, null, model.scope);
   };
 }
 
--- a/test/server/routes/oauth.test.ts
+++ b/test/server/routes/oauth.test.ts
@@ -10,11 +10,11 @@ import { fakeApplication, fakeOAuthAuthorizationCode, fakeUser } from '../../tes
 import { startTestServer, stopTestServer } from '../../test-helpers/server';
 import { resetTestDB } from '../../utils';
 
-export function generatePKCECodeVerifier() {
+function generatePKCECodeVerifier() {
   return randomBytes(32).toString('base64url');
 }
 
-export async function calculatePKCECodeChallenge(codeVerifier: string): Promise<string> {
+async function calculatePKCECodeChallenge(codeVerifier: string): Promise<string> {
   return createHash('sha256').update(codeVerifier).digest().toString('base64url');
 }
 
--- a/test/server/routes/oauth.test.ts
+++ b/test/server/routes/oauth.test.ts
@@ -39,6 +39,7 @@ describe('server/routes/oauth', () => {
       client_id: application.clientId,
       client_secret: application.clientSecret,
       redirect_uri: application.callbackUrl,
+      scope: 'email account',
     });
 
     const authorizeResponse = await request(expressApp)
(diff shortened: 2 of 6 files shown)


ThisIsMissEm:
As an aside, it took several hours to get the tests running and debugged locally — there's definitely room for improving the development setup.

E.g., one issue I encountered was `tsx ./scripts/update-fido-metadata.ts` needing to be run, because apparently the file committed in the repo isn't valid, and that'll crash the tests, but if you run the tests too many times, you'll get rate-limited by the API that provides that data.

The tests also regularly fail often due to port conflicts when running in watch mode. An alternative to using `superagent` would be to use [`light-my-request`](https://npm.im/light-my-request), which does request injection testing: essentially it doesn't start the server listening on a give port, but rather injects requests into the server, this means that if any errors happen, you get a complete stacktrace + the ability to step through code). This is the testing strategy that Fastify uses, and it's far more robust than actually starting up a server and hoping you can listen on a port.

I did also need a dotenv file to configure the tests for my local environment, which was a bit awkward. The `errorhandler` middleware also wasn't logging errors — I'd probably suggest switching to winston-express, even though that package is looking for maintainers, if you're wanting to stick with winston (but using pino would probably be a best direction)

test/server/routes/oauth.test.ts
```diff
@@ -67,20 +77,25 @@ describe('server/routes/oauth', () => {
...
     const iat = fakeNow.getTime() / 1000;
     expect(decodedToken.iat).to.eq(iat); // 1640995200
     expect(decodedToken.exp).to.eq(iat + 7776000); // 90 days
+    expect(decodedToken.scope).to.eq('oauth');
 
     // Test OAuth token with a real query
     const gqlRequestResult = await request(expressApp)
       .post('/graphql/v2')
-      .set('Authorization', `Bearer ${oauthToken}`)
+      .set('Authorization', `Bearer ${oauthToken.access_token}`)
```
ThisIsMissEm, line 98: This change here was partly the reason for the 5xx error before — I'd forgotten that it'd set the `oauthToken` to be the complete json body, not just the token, so when we had an incorrect Bearer token as `Bearer [object Object]` instead of the actual token.

test/server/routes/oauth.test.ts
```diff
@@ -39,6 +48,7 @@ describe('server/routes/oauth', () => {
       client_id: application.clientId,
       client_secret: application.clientSecret,
       redirect_uri: application.callbackUrl,
+      scope: 'email account',
```
ThisIsMissEm, line 51: Adding this line caused another 503 error, even though it _should_ accept a `scope` parameter, the issue turned out to be a bug in the Model where the `saveAuthorizationCode` and `saveToken` methods received a `scope` property which was one of:
- string[]
- string
- undefined

So I've added code to correctly parse the scopes now.

test/server/routes/oauth.test.ts
```diff
@@ -342,5 +357,68 @@ describe('server/routes/oauth', () => {
         error_description: 'Invalid request: `redirect_uri` is invalid',
       });
     });
+
+    describe('PKCE', async () => {
+      let authorization, validParams, codeVerifier, codeChallenge;
+
+      beforeEach(async () => {
+        codeVerifier = generatePKCECodeVerifier();
+        codeChallenge = await calculatePKCECodeChallenge(codeVerifier);
```
ThisIsMissEm, line 366: I've switched these from hard-coded values to dynamically generated, since I'd apparently had a mismatch between the values. The code for these is based on that in [`oauth4webapi`](https://npm.im/oauth4webapi). However, due to running in Node.js 20 (which is currently in maintenance as an [LTS release](https://nodejs.org/en/about/previous-releases) — I'd encourage upgrading to Node 22 or 24), I've reworked them to use native node.js `crypto` for `createHash` and `randomBytes` instead of the Web APIs for these.

test/server/routes/oauth.test.ts
```diff
@@ -342,5 +357,68 @@ describe('server/routes/oauth', () => {
...
+        const response = await request(expressApp)
+          .post('/oauth/token')
+          .type(`application/x-www-form-urlencoded`)
+          .send({
+            grant_type: 'authorization_code',
+            code: authorization.code,
+            client_id: authorization.application.clientId,
+            client_secret: authorization.application.clientSecret,
+            redirect_uri: authorization.application.callbackUrl,
+          })
```
ThisIsMissEm, line 401: I had been using `code_verifier: undefined` but it turns out superagent doesn't strip undefined values when sending json, so errors happened here. So instead I'm just taking the `validParams` from above and inlining them without the `code_verifier` parameter.

This parameter is only required because the Authorization Code was created with PKCE.

server/lib/oauth/index.ts
```diff
@@ -19,20 +21,16 @@ class CustomTokenHandler extends TokenHandler {
...
+        scope: 'oauth',
         // eslint-disable-next-line camelcase
-        return { access_token: accessToken };
+        access_token: model.accessToken,
       },
-    };
+      auth.TOKEN_EXPIRATION_SESSION_OAUTH, // 90 days,
+    );
+
+    return new BearerTokenType(accessToken, auth.TOKEN_EXPIRATION_SESSION_OAUTH, null, model.scope);
```
ThisIsMissEm, line 33: As originally noted, if you did want to support refresh token grant flow, then you could return a refresh token here by replacing `null` with the refresh token `model.refreshToken` (one is currently generated always apparently, but it's just never used)

server/lib/oauth/index.ts
```diff
@@ -221,6 +219,7 @@ const handleResponse = function (req, res, response) {
  */
 
 const handleError = function (e, req, res, response, next) {
+  logger.error(e);
```
ThisIsMissEm, line 222: This makes debugging OAuth errors much easier, since the `errorhandler` middleware doesn't actually log the errors in `test`.

Betree:
> As an aside, it took several hours to get the tests running and debugged locally — there's definitely room for improving the development setup.

Sorry for the poor experience on that side. I've started a [PR](https://github.com/opencollective/opencollective-frontend/pull/11279) to document/fix the various frictions one can encounter during the API setup.

> E.g., one issue I encountered was `tsx ./scripts/update-fido-metadata.ts` needing to be run, because apparently the file committed in the repo isn't valid, and that'll crash the tests, but if you run the tests too many times, you'll get rate-limited by the API that provides that data.

I didn't encounter this one on a fresh install (Fedora 42, using the package's default node/postgres versions). We'll update the default cached version.

> The tests also regularly fail often due to port conflicts when running in watch mode. An alternative to using `superagent` would be to use [`light-my-request`](https://npm.im/light-my-request), which does request injection testing: essentially it doesn't start the server listening on a give port, but rather injects requests into the server, this means that if any errors happen, you get a complete stacktrace + the ability to step through code). This is the testing strategy that Fastify uses, and it's far more robust than actually starting up a server and hoping you can listen on a port.

I wasn't aware of `light-my-request`, thanks for sharing it. It definitely sounds like a superior approach for testing requests. Documented in https://github.com/opencollective/opencollective/issues/8011.

> I did also need a dotenv file to configure the tests for my local environment, which was a bit awkward.

The default setup is intended to work without any further configuration, was it because you used special DB credentials/URL? In any case, we'll add some instructions about that in `README`.

> The `errorhandler` middleware also wasn't logging errors

The historical strategy is to not log errors by default in tests to avoid polluting the tests logs when writing error cases, but then do something like:

```ts
it('should work', async () => {
  const response = await graphqlRequest(...);
  response.errors && console.error(response.errors);
  expect(response.errors).to.not.exists;
})
```

Betree (approved):
Looking great!

Sorry for the late review. We'll try to ship this one before EOW.

PR #10932, opened 2025-06-19
Propagate refund ignoreBalanceCheck param to payment providers

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/8034


Betree (approved):
Looking good! Nice that you moved some additional payment providers to TS.

PR #10964, opened 2025-07-03
Fix: OAuth errors

ThisIsMissEm:
Following the merge of #10917, upon testing I discovered that:

1. The "cancel" button click did not actually work as expected, as the API didn't correctly return the redirect URI like I thought it would (we likely need integration test coverage on the frontend too, but for now I've explicitly added tests for this code path)
2. The `scope` property of the token response was an array of strings, when it should've been a string of space separated scope values. (this caused oauth4webapi to crash)

Screenshot 2025-07-03 at 20 15 00

### Cancel button handling:

I'd misread this error handling code in `@node-oauth/oauth-server`: https://github.com/node-oauth/node-oauth2-server/blob/master/lib/handlers/authorize-handler.js#L126-L129

I expected it to gracefully handle the AccessDeniedError and return it to the frontend, but it actually bubbles up to the `handleError` method in `server/lib/oauth/index.ts` instead as the error is rethrow, so in there I'm explicitly checking for `AccessDeniedError`. We could arguably do a try/catch on line 70 instead, and handle the error there.

### Incorrect scope property error

Screenshot 2025-07-03 at 21 01 48

The above shows the issue with the `scope` property on bearer token response, where [OAuth4WebAPI](https://github.com/panva/oauth4webapi) throws an error because it expects the `scope` value to be a string (the specification does imply this)

### DIFF
--- a/server/models/Application.ts
+++ b/server/models/Application.ts
@@ -1,7 +1,15 @@
 import { randomBytes } from 'crypto';
 
 import { isNil, merge } from 'lodash';
-import type { CreationOptional, ForeignKey, InferAttributes, InferCreationAttributes, NonAttribute } from 'sequelize';
+import type {
+  Association,
+  BelongsToGetAssociationMixin,
+  CreationOptional,
+  ForeignKey,
+  InferAttributes,
+  InferCreationAttributes,
+  NonAttribute,
+} from 'sequelize';
 
 import { crypto } from '../lib/encryption';
 import sequelize, { DataTypes, Model } from '../lib/sequelize';
@@ -27,6 +35,13 @@ class Application extends Model<InferAttributes<Application>, InferCreationAttri
   declare public updatedAt: CreationOptional<Date>;
   declare public deletedAt: CreationOptional<Date>;
   declare public data: JSON;
+
+  declare public createdByUser: NonAttribute<User>;
+  declare getCreatedByUser: BelongsToGetAssociationMixin<User>;
+
+  declare static associations: {
+    createdByUser: Association<Application, User>;
+  };
 
   get info(): NonAttribute<Partial<Application>> {
     return {
--- a/test/server/routes/oauth.test.ts
+++ b/test/server/routes/oauth.test.ts
@@ -82,7 +82,8 @@ describe('server/routes/oauth', () => {
     expect(oauthToken.access_token).to.be.a('string');
     expect(oauthToken.token_type).to.eq('Bearer');
     expect(oauthToken.expires_in).to.eq(7776000);
-    expect(oauthToken.scope).to.have.members(['email', 'account']);
+    // scope should be a string of scopes:
+    expect(oauthToken.scope).to.eq('email account');
 
     const decodedToken = jwt.verify(oauthToken.access_token, config.keys.opencollective.jwtSecret) as jwt.JwtPayload;
     expect(decodedToken.sub).to.eq(application.CreatedByUserId.toString());
--- a/test/server/routes/oauth.test.ts
+++ b/test/server/routes/oauth.test.ts
@@ -46,7 +46,6 @@ describe('server/routes/oauth', () => {
     const authorizeParams = new URLSearchParams({
       response_type: 'code',
       client_id: application.clientId,
-      client_secret: application.clientSecret,
       redirect_uri: application.callbackUrl,
       scope: 'email account',
     });
@@ -111,6 +110,7 @@ describe('server/routes/oauth', () => {
     it('must provide a client_id', async () => {
       const response = await request(expressApp).post('/oauth/authorize?response_type=code').expect(400);
       const body = response.body;
+      expect(body).to.be.an('object');
       expect(body.error).to.eq('invalid_request');
       expect(body.error_description).to.eq('Missing parameter: `client_id`');
     });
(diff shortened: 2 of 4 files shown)


test/test-helpers/fake-data.ts
```diff
@@ -1081,7 +1081,7 @@ export const fakePaypalPlan = async (data: Record<string, unknown> = {}) => {
   });
 };
 
-export const fakeApplication = async (data: Record<string, unknown> = {}): Promise<any> => {
+export const fakeApplication = async (data: Record<string, unknown> = {}) => {
```
ThisIsMissEm, line 1084: This was making the types within the tests not actually correct.

test/server/routes/oauth.test.ts
```diff
@@ -46,7 +46,6 @@ describe('server/routes/oauth', () => {
     const authorizeParams = new URLSearchParams({
       response_type: 'code',
       client_id: application.clientId,
-      client_secret: application.clientSecret,
```
ThisIsMissEm, line 49: `client_secret` should never be part of the `authorize` URL params, since that would leak the client secret to intermediaries and the end-user.

test/server/routes/oauth.test.ts
```diff
@@ -156,6 +158,55 @@ describe('server/routes/oauth', () => {
       expect(response.body).to.be.empty;
       expect(response.get('www-authenticate')).to.eq('Bearer realm="service"');
     });
+
+    it('correctly handles invalid response type', async () => {
```
ThisIsMissEm, line 162: This test case is just to assert that a different OAuthError subclass doesn't trigger the redirect URI behaviour

test/server/routes/oauth.test.ts
```diff
@@ -156,6 +158,55 @@ describe('server/routes/oauth', () => {
...
+        allowed: 'false',
+      });
+
+      const expectedRedirect = new URL(application.callbackUrl);
+      expectedRedirect.searchParams.set('error', 'access_denied');
+      expectedRedirect.searchParams.set('error_description', 'Access denied: user denied access to application');
+
+      // For some reason the error_description is encoded differently by URL
+      // SearchParams, but the result is effectively the same:
+      const expectedRedirectUrl = expectedRedirect.href.replaceAll('+', '%20');
```
ThisIsMissEm, line 198: The difference is here I'm using the WHATWG URL constructor, and in `@node-oauth/oauth-server` they use the node.js `url` module's `format()` method. But `+` and `%20` here are semantically the same.

server/models/Application.ts
```diff
@@ -28,6 +35,12 @@ class Application extends Model<InferAttributes<Application>, InferCreationAttri
   declare public deletedAt: CreationOptional<Date>;
   declare public data: JSON;
 
+  declare public createdByUser: NonAttribute<User>;
+
+  declare static associations: {
+    createdByUser: Association<Application, User>;
+  };
```
ThisIsMissEm: These changes allow us to do `application.createdByUser.jwt()` in the tests correctly, previously this would've given a typescript error.

ThisIsMissEm: Based on the documentation at: https://sequelize.org/docs/v6/other-topics/typescript/

ThisIsMissEm:
@Betree for your attention (likely when you're back from holidays around July 4th)

server/models/Application.ts
```diff
@@ -28,6 +36,13 @@ class Application extends Model<InferAttributes<Application>, InferCreationAttri
   declare public deletedAt: CreationOptional<Date>;
   declare public data: JSON;
 
+  declare public createdByUser: NonAttribute<User>;
+  declare getCreatedByUser: BelongsToGetAssociationMixin<User>;
+
+  declare static associations: {
+    createdByUser: Association<Application, User>;
+  };
+
```
Betree, line 45: Curious as to why that one was actually needed?

Betree (approved):

server/models/Application.ts
ThisIsMissEm: So if I removed the `Promise<any>` on creating an application in fake-data, then the `createdByUser` property wasn't available as far as typescript was concerned. Looking at their docs, they seem to need you to manually declare everything for typescript. I don't know much about sequelize, but this is what I figure out.

Though, now that I try removing that, it works too with just the lines:
```ts
  declare public createdByUser: NonAttribute<User>;
  declare getCreatedByUser: BelongsToGetAssociationMixin<User>;
```

So I'm not sure. I don't know what this does internally in sequelize, and the documentation doesn't really explain it, it just gives a bunch of code samples: https://sequelize.org/docs/v6/other-topics/typescript/

I think it may be used for doing:
```ts
  const ourUser = await User.findByPk(1, {
    include: [User.associations.projects],
    rejectOnEmpty: true // Specifying true here removes `null` from the return type!
  });
```

So like, in the opencollective code, it might be something like:
```ts
    const application = await fetchApplicationWithReference(args.application, {
      include: [Application.associations.collective],
    });
```
Assuming we have the following on `Application`:
```
  declare static associations: {
    collective: Association<Application, Collective>;
    createdByUser: Association<Application, User>;
  };
```

(this was taken from: https://github.com/opencollective/opencollective-api/blob/f2f044c26c1b4eb0b50ab6de860dd7d10fd3c771/server/graphql/v2/mutation/ApplicationMutations.js#L62 )

ThisIsMissEm:
@Betree let me know when it's live and I can retest.

Betree:
@ThisIsMissEm It's live!

PR #10970, opened 2025-07-04
Fix(Activites): Save UserId for COLLECTIVE_CORE_MEMBER_ADDED activity

gustavlrsn:
Project: https://github.com/opencollective/opencollective/issues/8019

# Description
- Save UserId for COLLECTIVE_CORE_MEMBER_ADDED activity, so that we can show who added a member in the timeline

### DIFF
--- a/server/models/MemberInvitation.ts
+++ b/server/models/MemberInvitation.ts
@@ -270,6 +270,7 @@ MemberInvitation.prototype.accept = async function () {
     const memberCollective = await Collective.findByPk(this.MemberCollectiveId);
     await Activity.create({
       type: ActivityTypes.COLLECTIVE_CORE_MEMBER_ADDED,
+      UserId: this.CreatedByUserId,
       CollectiveId: this.CollectiveId,
       FromCollectiveId: this.MemberCollectiveId,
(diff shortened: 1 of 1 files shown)


kewitz (approved):

server/graphql/v2/object/Activity.ts
```diff
@@ -82,7 +83,12 @@ export const GraphQLActivity = new GraphQLObjectType({
           isIncognito = Boolean(fromCollective?.isIncognito);
         }
 
-        if (isIncognito && !req.remoteUser?.isRoot() && !req.remoteUser?.isAdminOfCollective(userCollective)) {
+        if (
+          isIncognito &&
+          !req.remoteUser?.isRoot() &&
+          !req.remoteUser?.isAdminOfCollective(userCollective) &&
+          !req.remoteUser?.isAdminOfCollectiveOrHost(activityCollective)
```
kewitz: Wait, I don't think this makes sense. We're leaking incognito profiles to collective and host admins...

gustavlrsn: True, I've removed this commit to only keep the part about saving UserId for the COLLECTIVE_CORE_MEMBERT_ADDED activity.

PR #10986, opened 2025-07-18
Add canHostAccounts Collective setting

kewitz:
This can be used by Organizations to disable collective hosting oriented features while keeping access to fund management features.

Relates to https://github.com/opencollective/opencollective/issues/8067

### DIFF
--- a/server/graphql/v2/mutation/AccountMutations.ts
+++ b/server/graphql/v2/mutation/AccountMutations.ts
@@ -136,7 +136,7 @@ const accountMutations = {
           throwIfMissing: true,
         });
 
-        const isKeyEditableByHostAdmins = ['expenseTypes'].includes(args.key);
+        const isKeyEditableByHostAdmins = ['expenseTypes', 'canHostAccounts'].includes(args.key);
         const permissionMethod = isKeyEditableByHostAdmins ? 'isAdminOfCollectiveOrHost' : 'isAdminOfCollective';
         if (!req.remoteUser[permissionMethod](account)) {
           throw new Forbidden();
--- a/server/graphql/v2/input/OrganizationCreateInput.ts
+++ b/server/graphql/v2/input/OrganizationCreateInput.ts
@@ -1,6 +1,8 @@
 import { GraphQLInputObjectType, GraphQLNonNull, GraphQLString } from 'graphql';
 import { GraphQLJSON } from 'graphql-scalars';
 
+import { GraphQLCurrency } from '../enum';
+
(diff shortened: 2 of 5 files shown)


server/graphql/v2/input/FeatureToggleInput.ts
```diff
@@ -0,0 +1,19 @@
...
+export const GraphQLFeatureToggleInput = new GraphQLInputObjectType({
+  name: 'FeatureToggleInput',
+  description: 'Input type for toggling features on or off for an acccount',
+  fields: () => ({
+    key: {
+      type: new GraphQLNonNull(GraphQLFeature),
+      description: 'Feature to toggle.',
+    },
+    status: {
+      type: new GraphQLNonNull(GraphQLCollectiveFeatureStatus),
```
Betree: Not the right type here: a user setting the feature to ACTIVE makes no sense (active depends on whether existing data is found) nor does UNSUPPORTED, which depends on your account type - which you have no control over. A simple "enabled" true/false should do the job just fine.

server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -847,6 +849,39 @@ const accountMutations = {
...
+      const account = await fetchAccountWithReference(args.account, { throwIfMissing: true });
+      if (!req.remoteUser.isAdminOfCollective(account) && !req.remoteUser.isRoot()) {
+        throw new Forbidden();
+      }
+      await TwoFactorAuthLib.enforceForAccount(req, account, { onlyAskOnLogin: true });
+
+      if (args.feature.status === FEATURE_STATUS.ACTIVE) {
+        await account.enableFeature(args.feature.key);
+      } else if (args.feature.status === FEATURE_STATUS.DISABLED) {
+        await account.disableFeature(args.feature.key);
```
Betree: 🚫  We don't want people to be able to enable or disable any features on their own accounts. These flags are used in moderation, for example, to prevent spammers from commenting on expenses. Which is why we store them in `data`. 

It sounds to me like what you're trying to introduce is more of a `settings` (which users are expected to edit) than a feature. I would recommend moving it there.

If we have good reasons to keep it in features, then I would recommend this mutation to only allow editing this specific flag. And we'll need to take note that we're losing the ability to actively block hosting accounts for fiscal hosts, requiring us to implement another mechanism if we need that. Altogether, it seems to me that the cost of adding this as a feature largely outweighs the benefits when compared to a simple addition in `settings`  (which we already have mutations for).

PR #11006, opened 2025-08-01
Performance: Avoid complex conditionals when querying READY_TO_PAY expenses

kewitz:


server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
```diff
@@ -511,6 +452,69 @@ export const ExpensesCollectionQueryResolver = async (
     }
   }
 
+  if (!isNil(args.chargeHasReceipts)) {
+    where[Op.and].push(
+      { type: ExpenseType.CHARGE },
+      sequelize.where(
+        sequelize.literal(
+          `NOT EXISTS (SELECT id from "ExpenseItems" ei WHERE ei."ExpenseId" = "Expense".id and ei.url IS NULL AND ei."deletedAt" IS NULL)`,
```
Betree, line 460: I'm surprised that this one is expensive. Is it because of the subquery? Should it rather be a join?

Betree (approved):

server/graphql/v2/query/collection/ExpensesCollectionQuery.ts
kewitz: Not really, I just moved all sub-queries bellow the potential READY_TO_PAY query.

PR #11010, opened 2025-08-04
Platform subscription model

hdiniz:


migrations/20250730104502-add-PlatformSubscriptions-table.js
```diff
@@ -0,0 +1,97 @@
...
+        type: Sequelize.BIGINT,
+        primaryKey: true,
+        autoIncrement: true,
+        unique: true,
+      },
+      archivedAt: {
+        type: Sequelize.DATE,
+        allowNull: false,
+        defaultValue: Sequelize.fn('NOW'),
+      },
```
kewitz, line 79: Where are these coming from?

server/models/PlatformSubscription.ts
```diff
@@ -0,0 +1,346 @@
...
+
+  async calculateUtilization(billingPeriod: BillingPeriod): Promise<PeriodUtilization> {
+    const billingRange = PlatformSubscription.getBillingPeriodRange(billingPeriod);
+    const billingRangeArg = PlatformSubscription.rangeLiteral(billingRange);
+    const rangeLiteralArg = PlatformSubscription.rangeLiteral(this.period);
+
+    const createdAtWhereArgs: WhereAttributeHashValue<Date> = {
+      [Op.and]: [
+        { [Op.contained]: sequelize.literal(rangeLiteralArg) },
+        { [Op.contained]: sequelize.literal(billingRangeArg) },
```
kewitz: Clever, but it is a bit implicit that this operation is enough to calculate utilization in a partial billing period.
Perhaps adding a JSDoc comment to this function explaining what this function can be used for.

server/models/PlatformSubscription.ts
```diff
@@ -0,0 +1,346 @@
...
+        },
+      },
+    );
+
+    const expensesPaid = await Expense.count({
+      where: {
+        HostCollectiveId: this.CollectiveId,
+        createdAt: createdAtWhereArgs,
+      },
+    });
```
Betree: Two things here:
1. We should use the payment date rather than `createdAt`.
2. While the current assumption of `HostCollectiveId` means `PAID`, I would double down and make it explicit with a status check.
 
For these reasons, I'd advocate for instead looking at the `Transactions` table with a COUNT that distinguishes on `ExpenseId`. That way, you ensure that you count only paid expenses, and the date will always be meaningful.

Betree: On top of that: say I pay an expense in January. Then in February, I go back to this expense to mark it as "Unpaid" then re-pay it. (e.g. because I made a mistake in how I recorded the first version. This, I think we'll all agree, needs to count for "1 expense paid" in the quotas. I'd argue it should also count for January, not February.

migrations/20250730104502-add-PlatformSubscriptions-table.js
hdiniz: `hid` and `archivedAt` are used by sequelize-temporal

server/models/PlatformSubscription.ts
```diff
@@ -0,0 +1,346 @@
...
+
+export enum UtilizationType {
+  ACTIVE_COLLECTIVES = 'activeCollectives',
+  EXPENSES_PAID = 'expensesPaid',
+}
+
+type PeriodUtilization = Record<UtilizationType, number>;
+
+type PlatformSubscriptionPlan = {
+  name: string;
```
kewitz:
```suggestion
  name: string;
  pricePerMonth: Number;
  activeCollectivesLimit: Number;
  pricePerExtraActiveCollective: Number;
  paidExpenseLimit: Number;
  pricePerExtraPaidExpense: Number;
  features: String[];
```

server/models/PlatformSubscription.ts
hdiniz: Based on this, I think maybe it's best to look at activities to get it accounted correctly in January in the example given. Another example, it's paid with wise on the last day of January but only marked as paid in February, will be accounted for in February.

PR #11018, opened 2025-08-08
Subscribers Root Dashboard requirements

kewitz:
Add required gQL resolvers and mutations for the new Subscribers dashboard.

### DIFF
--- a/server/graphql/v2/mutation/PlatformSubscriptionsMutations.ts
+++ b/server/graphql/v2/mutation/PlatformSubscriptionsMutations.ts
@@ -1,3 +1,5 @@
+import assert from 'assert';
+
 import { GraphQLNonNull } from 'graphql';
 
 import { PlatformSubscriptionPlan } from '../../../constants/plans';
@@ -36,6 +38,18 @@ const platformSubscriptionMutations = {
             includedExpensesPerMonth: args.subscription.plan.pricing.includedExpensesPerMonth,
           },
         };
+        assert(
+          args.subscription.plan.pricing.pricePerMonth.currency === 'USD',
+          'Only USD is supported for platform subscription pricing',
+        );
+        assert(
+          args.subscription.plan.pricing.pricePerAdditionalCollective.currency === 'USD',
+          'Only USD is supported for platform subscription pricing',
+        );
+        assert(
+          args.subscription.plan.pricing.pricePerAdditionalExpense.currency === 'USD',
+          'Only USD is supported for platform subscription pricing',
+        );
         await PlatformSubscription.replaceCurrentSubscription(account.id, new Date(), plan);
         await account.update({ plan: null });
         return account;


server/graphql/v2/interface/AccountWithPlatformSubscription.ts
```diff
@@ -44,6 +49,22 @@ export const AccountWithPlatformSubscriptionFields = {
...
+    type: GraphQLAmount,
+    description: 'The total amount managed by the account, including all its children accounts (events and projects)',
+    resolve: async (account: Collective, _args, req: Express.Request) => {
+      assert(req.remoteUser.isRoot());
+
+      const result = await req.loaders.Collective.moneyManaged.load(account.id);
+      return pick(result, ['value', 'currency']);
+    },
+  },
+  plan: {
```
hdiniz: maybe we should call this a legacy plan to differentiate?

server/graphql/v2/interface/AccountWithPlatformSubscription.ts
```diff
@@ -44,6 +49,22 @@ export const AccountWithPlatformSubscriptionFields = {
       };
     },
   },
+  managedAmount: {
```
hdiniz: This feels a bit out of place in this interface, can this be put in the account object?

server/graphql/v2/mutation/PlatformSubscriptionsMutations.ts
```diff
@@ -0,0 +1,49 @@
...
+    async resolve(_, args, req: Express.Request): Promise<Collective> {
+      if (!req.remoteUser) {
+        throw new Error('You need to be logged in to update a platform subscription');
+      }
+      const account = await fetchAccountWithReference(args.account, { throwIfMissing: true, paranoid: false });
+      if (req.remoteUser.isRoot()) {
+        const plan: Partial<PlatformSubscriptionPlan> = {
+          title: args.subscription.plan.title,
+          type: args.subscription.plan.type,
+          pricing: {
```
hdiniz, line 33: I think we should validate that the currency input is USD because we are only storing the amount

server/graphql/v2/mutation/PlatformSubscriptionsMutations.ts
```diff
@@ -0,0 +1,49 @@
...
+        description: 'The new platform subscription tier to apply to the account',
+      },
+    },
+    async resolve(_, args, req: Express.Request): Promise<Collective> {
+      if (!req.remoteUser) {
+        throw new Error('You need to be logged in to update a platform subscription');
+      }
+      const account = await fetchAccountWithReference(args.account, { throwIfMissing: true, paranoid: false });
+      if (req.remoteUser.isRoot()) {
+        const plan: Partial<PlatformSubscriptionPlan> = {
```
hdiniz, line 30: Is it possible for root to specify a preexisting plan?

kewitz: I'm enforcing all parameters, should we keep track of the original plan in this case?

hdiniz: If we are going to reuse this mutation for the customer self-service upgrades, it would be best that the parameters are optional or that we can pass a reference to a plan the customer can upgrade/downgrade to.

kewitz: Agreed, I think we can build on top of this. We'll also need to update the PlatformSubscription type to include a base ID.

PR #11019, opened 2025-08-08
Calculate period utilization

hdiniz:


server/models/PlatformSubscription.ts
```diff
@@ -94,6 +119,30 @@ class PlatformSubscription extends Model<
     return now.isSameOrBefore(this.endDate) && now.isSameOrAfter(this.startDate);
   }
 
+  prorateBasePrice(billingPeriod: BillingPeriod): number {
+    const billingStart = PlatformSubscription.periodStartDate(
+      PlatformSubscription.getBillingPeriodRange(billingPeriod),
+    );
+    const billingEnd = PlatformSubscription.periodEndDate(PlatformSubscription.getBillingPeriodRange(billingPeriod));
+
+    const billingTime = moment.utc(billingEnd).diff(billingStart, 'seconds');
```
kewitz, line 128: I think we should calculate base in a difference in days, but this is just my opinion, and we should check with product.

kewitz (approved):
Looks good. We should make sure to add extensive testing to this file to make sure everything works as expected.

server/models/PlatformSubscription.ts
hdiniz: This is only used to get the ratio, I avoided days here as its lower resolution and causes rounding issues

PR #11029, opened 2025-08-14
fix(EditCollective): prevent editing currency

Betree:


Betree:
> Should we set "setCurrency" as a private method?

I initially tried that way, but the method is used in a few other places where we do want to skip these checks (e.g., `createOrganization`), I didn't want to fall into a rabbit hole.

PR #11030, opened 2025-08-14
Update subscription mutations

hdiniz:
Related https://github.com/opencollective/opencollective/issues/8108


migrations/20250813202307-add-expense-platform-billing-type.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`ALTER TYPE "enum_Expenses_type" ADD VALUE 'PLATFORM_BILLING';`);
```
kewitz, line 6: I think we should name this as a noun, to match "receipt" and "invoice". Perhaps "PLATFORM_BILL"?

hdiniz: I have no preference, but [billing is also a noun](https://en.wiktionary.org/wiki/billing)

kewitz (approved):

PR #11053, opened 2025-08-22
Charge stripe platform bill

hdiniz:
Related https://github.com/opencollective/opencollective/issues/8108


server/lib/platform-subscriptions/stripe-payment.ts
```diff
@@ -0,0 +1,78 @@
+import config from 'config';
+
+import { PAYMENT_METHOD_SERVICE, PAYMENT_METHOD_TYPE } from '../../constants/paymentMethods';
+import { Expense, PaymentMethod } from '../../models';
+import { PayoutMethodTypes } from '../../models/PayoutMethod';
+import stripe, { convertToStripeAmount } from '../stripe';
+
+export async function chargePlatformBillingExpenseWithStripe(expense: Expense) {
```
kewitz: If the long-term vision is to unify Orders and Expenses, I suggest moving this pay expense function to `paymentProviders/stripe`.

If you follow this suggestion, consider also renaming `server/lib/platform-subscriptions/payment-options.ts -> server/lib/platform-subscriptions.ts`.

server/models/PlatformSubscription.ts
```diff
@@ -450,6 +454,21 @@ class PlatformSubscription extends Model<
...
+    switch (payoutMethod.type) {
+      case PayoutMethodTypes.STRIPE: {
+        return chargePlatformBillingExpenseWithStripe(expense);
+      }
+      default: {
+        return;
+      }
+    }
+  }
+
```
kewitz: For consistency, I think this function should live in the platform subscription lib.

kewitz (approved):
Looks good, I think we can move functions around to be a bit more consistent with the existing payment structure.

PR #11071, opened 2025-09-03
Do not debit contributions from balance when using Stripe or PayPal

kewitz:
Resolves https://github.com/opencollective/opencollective/issues/8175


znarf (approved):
Looks good!

PR #11079, opened 2025-09-22
deps: Remove deprecated error handler

Betree:


Betree:
Done in https://github.com/opencollective/opencollective-api/pull/11745

PR #11080, opened 2025-09-23
Name middlewares and type base server files

kewitz:


server/index.ts
```diff
@@ -18,13 +20,13 @@ import { updateCachedFidoMetadata } from './lib/two-factor-authentication/fido-m
 import { parseToBoolean } from './lib/utils';
 import routes from './routes';
 
-const workers = process.env.WEB_CONCURRENCY || 1;
+const workers = isUndefined(process.env.WEB_CONCURRENCY) ? toInteger(process.env.WEB_CONCURRENCY) : 1;
```
Betree, line 23: I think you meant this 🙂 

```suggestion
const workers = isUndefined(process.env.WEB_CONCURRENCY) ? 1 : toInteger(process.env.WEB_CONCURRENCY);
```

PR #11099, opened 2025-10-02
enhancement(PlatformSubscriptions): remove required legal docs when downgrading

Betree:
Resolve https://github.com/opencollective/opencollective/issues/8140


server/models/PlatformSubscription.ts
gustavlrsn, line 479: Maybe we should remove `when` as an argument as `handleSubscriptionUpdateFeatureChanges` will not consider this and update things immediately, and we currently only call the function with `when` set to `Date.now()`

gustavlrsn (approved):
Looks good!

server/models/PlatformSubscription.ts
Betree: Good catch. It's used in tests, though, and it's arguably important to keep the ability to delay a subscription change... so ideally, this should:
- Run instantly when `when` is today
- Run asynchronously (CRON job) when `when` is somewhere in the future

I'll iterate on this.

Betree: @gustavlrsn I added a CRON job mechanism to handle this. Also added a `featureProvisioningStatus` column to the `PlatformSubscription` model to track whether the plan has been provisioned with updated features or not.

The expectation is now:
- When updating (or cancelling) the plan right away, the feature provisioning is done directly/synchronously.
- When creating or cancelling a plan with a future date, the CRON job will take care of the provisioning date.

gustavlrsn: Nice, great to have a mechanism to deprovision ended plans as well, looks good!

gustavlrsn (approved):

Betree:
@sentry review

PR #11115, opened 2025-10-07
Internally managed subscriptions support off cycle charge date

hdiniz:
[Slack](https://oficonsortium.slack.com/archives/C06K5J91675/p1759763583850399)

### DIFF
--- a/server/lib/subscriptions.ts
+++ b/server/lib/subscriptions.ts
@@ -117,7 +117,8 @@ export const updatePaymentMethodForSubscription = async (
     // Update the next charge dates
     const previousNextChargeDate = order.Subscription.nextChargeDate;
     const interval = order.Subscription.interval;
-    const nextChargeDate = getNextChargeDateForUpdateContribution(previousNextChargeDate, interval, interval, false);
+    const nextChargeDate = getNextChargeDateForUpdateContribution(previousNextChargeDate, interval, interval, true);
+
     newSubscriptionData['nextChargeDate'] = nextChargeDate.toDate();
     newSubscriptionData['nextPeriodStart'] = nextChargeDate.toDate();
   }


server/lib/subscriptions.ts
```diff
@@ -117,7 +117,8 @@ export const updatePaymentMethodForSubscription = async (
     // Update the next charge dates
     const previousNextChargeDate = order.Subscription.nextChargeDate;
     const interval = order.Subscription.interval;
-    const nextChargeDate = getNextChargeDateForUpdateContribution(previousNextChargeDate, interval, interval, false);
+    const nextChargeDate = getNextChargeDateForUpdateContribution(previousNextChargeDate, interval, interval, true);
```
Betree:
```suggestion
    const nextChargeDate = getNextChargeDateForUpdateContribution(previousNextChargeDate, interval, interval, order.status === 'ERROR');
```

hdiniz: I think we should keep this change as is, after sampling some cases, the issue with the 23 subscriptions you listed seem to be related with the fact that they originally started as paypal subscription and later where updated to stripe credit card. 

`nextChargeDate` is not relevant for adjusting the subscription charge date after a successful charge, `nextPeriodStart` is.

The issue with the 23 orders seems to be from the fact that Paypal provider sets the `nextPeriodStart` to the date a subscription is created, could be any date of the month.

https://github.com/opencollective/opencollective-api/blob/08717905c0d639f8abb9ef9c9f34cdc4a225bd2f/server/paymentProviders/paypal/subscription.ts#L326

https://github.com/opencollective/opencollective-api/blob/08717905c0d639f8abb9ef9c9f34cdc4a225bd2f/server/paymentProviders/paypal/webhook.ts#L137-L141

On the other hand stripe contributions have the `nextPeriodStart` always set to the first date of the month, by calling `getNextChargeAndPeriodStartDates` with `status = 'new'`

When a contribution fails, we adjust the `nextChargeDate` for retrying, which for our managed subscriptions is fine, I think? Or in this concrete case, we just keep it "off cycle" to retry after the PM update.

When the subscription is successfully charged (after error or not), the `nextPeriodStart` date will be used by `getNextChargeAndPeriodStartDates` to move the `nextChargeDate` and `nextPeriodStart` to the first date of the next cycle. So `getNextChargeAndPeriodStartDates` seems to be alright here.

There is another case to look at, update subscription details already works like this:

https://github.com/opencollective/opencollective-api/blob/08717905c0d639f8abb9ef9c9f34cdc4a225bd2f/server/lib/subscriptions.ts#L238-L243

Since these 23 contributions started with paypal, charged "off cycle", I suppose its arguably OK to keep them charging "off cycle" after moving to another payment method. For the end-user they are just continuing to be  charged on the same day as always.

Betree: Then maybe add a comment there to explain why it's ok to pass `true` for `supportsOffCycle`? For anyone reading the code, this would look like a mistake/inconsistency:
- In `updateSubscriptionDetails`, we say that it's true if the subscription is managed externally
- With your change, `updatePaymentMethodForSubscription` will say that it's true if the subscription is **not** managed externally.

hdiniz: @Betree I made updates to handle the `nextPeriodStart` and added a few more tests

PR #11116, opened 2025-10-07
enhancement(Expense): protect invoice reference ID

Betree:


server/graphql/common/expenses.ts
```diff
@@ -437,6 +437,8 @@ export const canSeeExpenseInvoiceInfo: ExpensePermissionEvaluator = async (
 ) => {
   if (!validateExpenseScope(req)) {
     return false;
+  } else if (getContextPermission(req, PERMISSION_TYPE.SEE_EXPENSE_DRAFT_PRIVATE_DETAILS, expense.id)) {
+    return true;
```
kewitz, line 441: This update should be enough to cover for the submit on behalf, we have an E2E test covering this flow.

kewitz (approved):
Contextually, this update should be OK. We just need to update the unit test there.

PR #11121, opened 2025-10-08
Allow editing recurring draft fields

hdiniz:
Resolves https://github.com/opencollective/opencollective/issues/8064

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -18,7 +18,6 @@ import {
   isUndefined,
   keyBy,
   mapValues,
-  matches,
   min,
   omit,
   omitBy,
@@ -666,6 +665,15 @@ export const canEditPayee: ExpensePermissionEvaluator = async (req, expense, opt
     }
     return false;
   } else if (expense.status === ExpenseStatus.DRAFT) {
+    if (expense.RecurringExpenseId) {
+      if (options?.throw) {
+        throw new Forbidden(
+          'Can not edit payee for a recurring draft',
+          EXPENSE_PERMISSION_ERROR_CODES.UNSUPPORTED_STATUS,
+        );
+      }
+      return false;
+    }
     return remoteUserMeetsOneCondition(req, expense, [isOwner], options);
   } else if (expense.status === ExpenseStatus.PENDING) {
     return remoteUserMeetsOneCondition(req, expense, [isOwner], options);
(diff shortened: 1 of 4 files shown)


server/graphql/v2/mutation/ExpenseMutations.ts
```diff
@@ -193,6 +193,11 @@ const expenseMutations = {
         type: new GraphQLNonNull(GraphQLExpenseUpdateInput),
         description: 'Expense data',
       },
+      isDraftEdit: {
+        type: GraphQLBoolean,
+        default: false,
```
Betree: I believe the right key is `defaultValue`? Also, we can make it non-null since there's a default

```suggestion
        type: new GraphQLNonNull(GraphQLBoolean),
        defaultValue: false,
```

server/graphql/common/expenses.ts
```diff
@@ -666,6 +666,15 @@ export const canEditPayee: ExpensePermissionEvaluator = async (req, expense, opt
...
   } else if (expense.status === ExpenseStatus.DRAFT) {
+    if (expense.RecurringExpenseId) {
+      if (options?.throw) {
+        throw new Forbidden(
+          'Can not edit payee for a recurring draft',
+          EXPENSE_PERMISSION_ERROR_CODES.UNSUPPORTED_STATUS,
+        );
+      }
+      return false;
+    }
```
Betree, line 677: 👍

server/graphql/common/expenses.ts
```diff
@@ -2738,14 +2747,32 @@ export async function editExpenseDraft(
...
-  const items =
-    (await prepareExpenseItemInputs(req, currency, expenseData.items, { isEditing: true })) || existingExpense.items;
+  const items = await prepareExpenseItemInputs(
+    req,
+    currency,
+    expenseData.items || (existingExpense.data.items as any),
+    {
+      isEditing: true,
+    },
+  );
```
Betree: Removing the default `existingExpense.items` seems dangerous as it could result in various side-effects, including zeroing the expense amount, if you call `editExpense` while omitting to pass `items`.

```suggestion
  const items = (await prepareExpenseItemInputs(
    req,
    currency,
    expenseData.items || (existingExpense.data.items as any),
    {
      isEditing: true,
    },
  )) || existingExpense.items;
```

Betree (approved):

PR #11124, opened 2025-10-10
Prototype: People Dashboard API

kewitz:


migrations/20250930155000-add-community-activity-summary-view.ts
```diff
@@ -0,0 +1,89 @@
...
+              NULL AS "FromCollectiveId",
+              "CollectiveId",
+              "type",
+              "createdAt",
+              (CASE
+                WHEN "type" = 'collective.expense.approved' THEN 'EXPENSE_APPROVER'
+                END) AS "relations"
+            FROM "Activities"
+            WHERE "type" IN
+                  ('collective.expense.approved')
```
Betree: "rejected" seems to fall into the same category?

server/graphql/loaders/collective.ts
```diff
@@ -178,4 +178,49 @@ export default {
...
+          const keys = HostCollectivePairs.map(pair => `${pair.HostCollectiveId}-${pair.FromCollectiveId}`);
+          HostCollectivePairs.forEach(({ FromCollectiveId, HostCollectiveId }) => {
+            conditionals[HostCollectiveId] = conditionals[HostCollectiveId] || [];
+            conditionals[HostCollectiveId].push(FromCollectiveId);
+          });
+
+          const conditionalQuery = Object.entries(conditionals)
+            .map(([HostCollectiveId, FromCollectiveIds]) => {
+              return `(cas."HostCollectiveId" = ${HostCollectiveId} AND cas."FromCollectiveId" IN (${FromCollectiveIds.join(
+                ',',
```
Betree: As seen with recent cases, Postgres is not good a de-duplicating IDs, so great if the loader can take care of that:

```suggestion
          const conditionals: Record<number, Set<number>> = {};
          const keys = HostCollectivePairs.map(pair => `${pair.HostCollectiveId}-${pair.FromCollectiveId}`);
          HostCollectivePairs.forEach(({ FromCollectiveId, HostCollectiveId }) => {
            conditionals[HostCollectiveId] = conditionals[HostCollectiveId] || new Set<number>();
            conditionals[HostCollectiveId].add(FromCollectiveId);
          });

          const conditionalQuery = Object.entries(conditionals)
            .map(([HostCollectiveId, FromCollectiveIds]) => {
              return `(cas."HostCollectiveId" = ${HostCollectiveId} AND cas."FromCollectiveId" IN (${Array.from(FromCollectiveIds).join(
                ',',
```

server/graphql/v2/interface/Account.ts
```diff
@@ -1093,6 +1095,28 @@ const accountFieldsDefinition = () => ({
...
+      if (args.host) {
+        const host = await fetchAccountWithReference(args.host, { throwIfMissing: true });
+        if (!req.remoteUser?.isAdminOfCollective(host)) {
+          throw new BadRequest('Only host admins can lookup for members using the "host" argument');
+        }
+
+        return req.loaders.Collective.communityStats.onHostContext.load({
+          HostCollectiveId: host.id,
+          FromCollectiveId: account.id,
+        });
```
Betree, line 1118: Could this somehow be abused by malicious people? By passing:
- `host`: a host they administer
- `account`: an account they want to get more info about (not hosted by `host`)

server/graphql/v2/object/CommunityStats.ts
```diff
@@ -0,0 +1,187 @@
...
+      relations: {
+        type: new GraphQLList(GraphQLCommunityRelationType),
+        async resolve(account) {
+          if (account.dataValues.associatedCollectives) {
+            return uniq(flatten(Object.values(account.dataValues.associatedCollectives)));
+          }
+        },
+      },
+      transactionSummary: {
+        type: new GraphQLList(GraphQLCommunityTransactionSummary),
```
Betree:
```suggestion
        type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(GraphQLCommunityTransactionSummary)),
```

server/graphql/v2/object/CommunityStats.ts
```diff
@@ -0,0 +1,187 @@
...
+    },
+    expenseCount: {
+      type: GraphQLInt,
+    },
+    contributionTotal: {
+      type: GraphQLAmount,
+    },
+    contributionCount: {
+      type: GraphQLInt,
+    },
```
Betree, line 33: Looks like all these fields should be non-nullable?

server/graphql/v2/object/CommunityStats.ts
```diff
@@ -0,0 +1,187 @@
...
+
+const GraphQLCommunityAssociatedCollective = new GraphQLObjectType({
+  name: 'CommunityAssociatedCollective',
+  fields: () => ({
+    account: {
+      type: GraphQLAccount,
+    },
+    relations: {
+      type: new GraphQLList(GraphQLCommunityRelationType),
+    },
```
Betree, line 45: Non-nullable

server/graphql/v2/object/CommunityStats.ts
```diff
@@ -0,0 +1,187 @@
...
+        args: { ...CollectionArgs },
+        async resolve(account, args = {}) {
+          const { limit, offset } = args;
+          const HostCollectiveId = account.dataValues.contextualHostCollectiveId;
+          const UserId = account.data?.UserId;
+          if (!UserId) {
+            return { nodes: [], limit, offset, totalCount: 0 };
+          }
+
+          const { count: totalCount, rows: nodes } = await Activity.findAndCountAll({
```
Betree: We tend to use more the distinct resolvers pattern nowadays, as it allows for more flexible performance results:

```ts
return {
  nodes: () => Activity.findAll({ where, ... }),
  totalCount: () => Activity.count({ where, ... }),
  limit, offset,
}
```

server/graphql/v2/query/collection/CommunityQuery.ts
```diff
@@ -0,0 +1,164 @@
...
+  },
+  async resolve(_: void, args, req: Express.Request) {
+    if (isNil(args.limit) || args.limit < 0) {
+      args.limit = DEFAULT_LIMIT;
+    }
+    if (isNil(args.offset) || args.offset < 0) {
+      args.offset = 0;
+    }
+    if (args.limit > DEFAULT_LIMIT && !req.remoteUser?.isRoot()) {
+      throw new Error(`Cannot fetch more than ${DEFAULT_LIMIT},members at the same time, please adjust the limit`);
```
Betree:
```suggestion
      throw new Error(`Cannot fetch more than ${DEFAULT_LIMIT} members at the same time, please adjust the limit`);
```

server/graphql/v2/query/collection/CommunityQuery.ts
```diff
@@ -0,0 +1,164 @@
...
+  args: {
+    account: {
+      type: GraphQLAccountReferenceInput,
+      description: 'Account filter',
+    },
+    host: {
+      type: GraphQLAccountReferenceInput,
+      description: 'Host context filter',
+    },
+    type: { type: new GraphQLList(GraphQLAccountType) },
```
Betree, line 80:
```suggestion
    type: { type: new GraphQLList(new GraphQLNonNull(GraphQLAccountType)) },
```

server/graphql/v2/query/collection/CommunityQuery.ts
```diff
@@ -0,0 +1,164 @@
...
+    email: {
+      type: EmailAddress,
+      description: 'Admin only. To filter on the email address of a member, useful to check if a member exists.',
+    },
+    searchTerm: {
+      type: GraphQLString,
+    },
+    relation: {
+      type: new GraphQLList(GraphQLCommunityRelationType),
```
Betree:
```suggestion
      type: new GraphQLList(New GraphQLNonNull(GraphQLCommunityRelationType)),
```

server/graphql/v2/query/collection/CommunityQuery.ts
```diff
@@ -0,0 +1,164 @@
...
+    },
+    limit: { type: new GraphQLNonNull(GraphQLInt), defaultValue: DEFAULT_LIMIT },
+    offset: { type: new GraphQLNonNull(GraphQLInt), defaultValue: 0 },
+  },
+  async resolve(_: void, args, req: Express.Request) {
+    if (isNil(args.limit) || args.limit < 0) {
```
Betree, line 96:
```suggestion
    enforceScope(req, 'host');
    if (isNil(args.limit) || args.limit < 0) {
```

server/graphql/v2/query/collection/CommunityQuery.ts
```diff
@@ -0,0 +1,164 @@
...
+      offset: args.offset,
+    };
+
+    const account = args.account && (await fetchAccountWithReference(args.account, { throwIfMissing: false }));
+    const host = args.host && (await fetchAccountWithReference(args.host, { throwIfMissing: false }));
+    if (host && account) {
+      assert(
+        host.id === account.HostCollectiveId,
+        new BadRequest('The account provided is not hosted by the host provided'),
+      );
```
Betree, line 123: We'll want to update that in the future, as we know that hosts will want to get information about previously hosted collectives.

server/graphql/v2/query/collection/CommunityQuery.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    const data = await getHostCommunity(replacements);
+    const ids: number[] = data.nodes.map(c => c.id);
+    const permissions = await req.loaders.Collective.canSeePrivateLocation.loadMany(ids);
+    ids.forEach((id, i) => {
+      if (permissions[i]) {
+        allowContextPermission(req, PERMISSION_TYPE.SEE_ACCOUNT_PRIVATE_LOCATION, id);
+      }
+    });
```
Betree, line 162: This seems safe

migrations/20250930155000-add-community-activity-summary-view.ts
kewitz: That's a good point. It definitely does.

server/graphql/v2/interface/Account.ts
kewitz: Not really because `onHostContext` loader will not return anything if the account never interacted with your host. The assumption here is that the activities we use in this loader are equivalent to the relationship we expect to concede access to this type of information.

Also, most of the information down this tree are public.

PR #11148, opened 2025-10-16
Use correct urls to expense item notifications

hdiniz:
Before
image

After
image


Betree (approved):
I was initially concerned about the fact that we're going to override the "root" information in the activity's data, but I'm ok to try this since:
- All places I can think of that use this field will need the authenticated URL
- The change is very simple
- It is reversible if needed

hdiniz:
@Betree I just noticed the issue while testing another change

PR #11157, opened 2025-10-22
fix(HostApplication): Disable Github auto-approval

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/8270


Betree (approved):
What's the impact on the flow? They click "Verify with Github", authenticate their repo, fill the form, then [this condition](https://github.com/opencollective/opencollective-frontend/blob/d615c870bc8388e375e20ba9fe59a2036ef3b363/components/osc-host-application/ApplicationForm.js#L192) will take care of bringing them "YourInitiativeIsNearlyThere" screen where they should get:
> We will review it and let you know by mail if we need you for anything during the process

If that's the case, I'm all good with the change!

PR #11165, opened 2025-10-23
Fix cleared at use in contribution time series

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/8266

This change updates contribution time series to use the new `clearedAt` field in the transaction. This matches with the `effectiveDate` field that is now default on the transaction CSV export.

Additionally, it fixes the record of the clearedAt date for stripe contributions, using the balanceTransaction available_on date to correctly represent when funds are available. The difference between charge create date and balance available date is significant on delayed payments such as SEPA and ACH where funds can take sometimes weeks to clear.

Such as this example:
image

### DIFF
--- a/server/paymentProviders/stripe/common.ts
+++ b/server/paymentProviders/stripe/common.ts
@@ -174,7 +174,8 @@ export const createChargeTransactions = async (
     platformFeeInHostCurrency = fees.applicationFee;
   }
 
-  const clearedAt = balanceTransaction.available_on ? moment.unix(balanceTransaction.available_on).toDate() : null;
+  const clearedAtField = balanceTransaction.available_on || charge.created || null;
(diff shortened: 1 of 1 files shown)


server/paymentProviders/stripe/common.ts
```diff
@@ -174,7 +174,7 @@ export const createChargeTransactions = async (
     platformFeeInHostCurrency = fees.applicationFee;
   }
 
-  const clearedAt = charge.created ? moment.unix(charge.created).toDate() : null;
+  const clearedAt = balanceTransaction.available_on ? moment.unix(balanceTransaction.available_on).toDate() : null;
```
kewitz: Are we confident `balanceTransaction.available_on` is always available? Considering the type version we use locally and the version of the API we're on, I suggest keep the `charge.created` as fallback before getting to null.

PR #11171, opened 2025-10-27
Handle archivable payout method

hdiniz:
Resolve https://oficonsortium.slack.com/archives/C06K5J91675/p1761208238299349

Adds canBeArchived field to payout method schema, handle stripe payout method

### DIFF
--- a/server/graphql/v2/object/PayoutMethod.ts
+++ b/server/graphql/v2/object/PayoutMethod.ts
@@ -90,7 +90,7 @@ const GraphQLPayoutMethod = new GraphQLObjectType({
         }
         const collective = await req.loaders.Collective.byId.load(payoutMethod.CollectiveId);
         if (req.remoteUser?.isAdminOfCollective(collective)) {
-          return (await payoutMethod.canBeDeleted()) || (await payoutMethod.canBeArchived());
+          return payoutMethod.canBeDeleted();
         } else {
           return false;
         }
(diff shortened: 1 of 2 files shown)


server/graphql/v2/mutation/PayoutMethodMutations.ts
```diff
@@ -98,8 +98,10 @@ const payoutMethodMutations = {
       if (await payoutMethod.canBeDeleted()) {
         await payoutMethod.destroy();
         return payoutMethod;
-      } else {
+      } else if (await payoutMethod.canBeArchived()) {
```
kewitz, line 101: Same logic should be applied in the `editPayoutMethod` mutation.

server/graphql/v2/object/PayoutMethod.ts
```diff
@@ -90,7 +90,22 @@ const GraphQLPayoutMethod = new GraphQLObjectType({
         }
         const collective = await req.loaders.Collective.byId.load(payoutMethod.CollectiveId);
         if (req.remoteUser?.isAdminOfCollective(collective)) {
-          return payoutMethod.canBeDeleted();
+          return (await payoutMethod.canBeDeleted()) || (await payoutMethod.canBeArchived());
```
kewitz: We'll need to keep the `canBeDeleted` logic as it is if we want to fix the current frontend bug. Which is acceptable since we have `canBeArchived` as well.
```suggestion
          return payoutMethod.canBeDeleted();
```

PR #11176, opened 2025-10-31
Wise: Loose expense payment constraints

kewitz:
We're reducing a bit the constraints of expense quote and payment through Wise.

### DIFF
--- a/server/paymentProviders/transferwise/index.ts
+++ b/server/paymentProviders/transferwise/index.ts
@@ -161,7 +161,7 @@ async function quoteExpense(
   let expenseToPayoutMethodExchangeRate: number | null = null;
   if (expense.host.data?.useLegacyWiseQuoting === false) {
     if (expense.feesPayer === 'PAYEE') {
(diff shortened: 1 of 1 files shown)


server/paymentProviders/transferwise/index.ts
```diff
@@ -159,39 +158,68 @@ async function quoteExpense(
...
-    // Using "else if" because customizing the fee payer is not allowed for multi-currency expenses. See `getCanCustomizeFeesPayer`.
-    assert(
-      expense.host.currency === expense.currency,
-      'For expenses where fees are covered by the payee, the host currency must be the same as the expense currency',
-    );
-    quoteParams['sourceAmount'] = expense.amount / 100;
+  let expenseToPayoutMethodExchangeRate: number | null = null;
+  if (expense.host.data?.useLegacyWiseQuoting === false) {
+    if (expense.feesPayer === 'PAYEE') {
+      // Using "else if" because customizing the fee payer is not allowed for multi-currency expenses. See `getCanCustomizeFeesPayer`.
```
Betree: We're not using a else if anymore

```suggestion
      // Customizing the fee payer is not allowed for multi-currency expenses. See `getCanCustomizeFeesPayer`.
```

PR #11177, opened 2025-10-31
Check HostedCollectivesCount when disabling canHostAccounts flag

kewitz:


server/graphql/v2/mutation/AccountMutations.ts
```diff
@@ -168,6 +168,15 @@ const accountMutations = {
           }
         }
 
+        if (args.key === 'canHostAccounts' && args.value === false) {
+          const hostedCollectives = await account.getHostedCollectivesCount();
+          if (hostedCollectives >= 1) {
+            throw new Error(
+              `You can't deactivate hosting while still hosting ${hostedCollectives} other collectives. Please contact support: support@opencollective.com.`,
```
znarf:
```suggestion
              `You can't deactivate hosting while still hosting ${hostedCollectives} collectives. Please contact support: support@opencollective.com.`,
```

znarf: other is useless

PR #11179, opened 2025-11-04
Skip email notification when fixing stale stripe orders

hdiniz:


server/lib/stripe/sync-order.ts
```diff
@@ -12,11 +12,34 @@ export const syncOrder = async (order, { IS_DRY, logging }: { IS_DRY?; logging?
...
+  try {
+    paymentIntent = await stripe.paymentIntents.retrieve(order.data.paymentIntent.id, {
+      stripeAccount,
+    });
+  } catch (err) {
+    logging?.(`Order ${order.id} payment intent not found on stripe`, err);
+    if (!IS_DRY) {
+      await order.update({
+        status: OrderStatuses.CANCELLED,
+      });
```
Betree: We should look at the error type before cancelling the order. The request could fail for numerous other reasons: rate-limiting, network error, Stripe API being down...etc

server/lib/stripe/sync-order.ts
```diff
@@ -12,11 +12,34 @@ export const syncOrder = async (order, { IS_DRY, logging }: { IS_DRY?; logging?
...
+    }
+    return;
+  }
+
   logging?.(`Order ${order.id} paymentIntent status: ${paymentIntent.status}`);
 
   const charge = (paymentIntent as any).charges?.data?.[0] as Stripe.Charge;
```
Betree, line 53: Why only look at the first charge? Could it be that a payment intent fails first, then succeeds after?

hdiniz: This array contains the charges ordered by most recent first. Its a deprecated field, replaced by `latest_charge` in the latest API, but I did not want to modify this at this time. We are due a task to try and update our stripe API usage, this field is not missing even from the docs at stripe.

server/lib/stripe/sync-order.ts
hdiniz: Fixed in this last push to specifically handle the case where the payment intent is expired on stripe. Happens when the payment intent is created, but no action is taken on it (i.e. user went into the last step of the contribution flow, but abandoned the contribution).

PR #11181, opened 2025-11-04
enhancement(Collective): do not set plan when becoming host if new pricing

Betree:


Betree:
Reverted the changes on `server/lib/allowed-features.ts` in favor of https://github.com/opencollective/opencollective-api/pull/11182.

PR #11184, opened 2025-11-05
feat(SignIn): Email when using a password on new device

Betree:

PR #11185, opened 2025-11-05
Feat: Add new OTP-based email validation endpoints

kewitz:


server/models/User.ts
```diff
@@ -470,6 +470,20 @@ class User extends Model<InferAttributes<User>, InferCreationAttributes<User>> {
...
   };
 
+  /**
+   *  Safely destroy a user account to avoid email conflicts in the future
+   *  Updates the user email before soft-deleting the account
+   */
+  safeDestroy = async function () {
+    // Update user email in order to free up for future reuse
+    // Split the email, username from host domain
+    const splitedEmail = this.email.split('@');
```
Betree, line 480: No need to do anything about it in this case (especially since this comes from older code), but just for sharing the info: the assumption that emails can only contain one `@` and that it separates host from domain is wrong: https://www.netmeister.org/blog/email.html#source-routing

server/models/User.ts
```diff
@@ -470,6 +470,20 @@ class User extends Model<InferAttributes<User>, InferCreationAttributes<User>> {
...
+    // Add the current timestamp to email username
+    const newEmail = `${splitedEmail[0]}-${Date.now()}@${splitedEmail[1]}`;
+    await this.update({ email: newEmail });
+    return this.destroy();
```
Betree: Not from this PR as well, but this should be done in a transaction; the result of changing the email without destroying the entry would be pretty bad security-wise.

```suggestion
    return sequelize.transaction(async (transaction) => {
      await this.update({ email: newEmail }, { transaction });
      return this.destroy({ transaction });
    });
```

server/lib/auth.ts
```diff
@@ -65,10 +66,14 @@ export function mustHaveRole(remoteUser, roles, CollectiveId, action = 'perform
...
+  const decodedToken = verifyJwt(token) as jwt.JwtPayload;
 
   const maxAge = decodedToken.exp * 1000 - new Date().getTime();
   const [header, payload, signature] = token.split('.');
   res.cookie('accessTokenPayload', [header, payload].join('.'), { maxAge, httpOnly: false, secure: true });
   res.cookie('accessTokenSignature', signature, { maxAge, httpOnly: true, secure: true });
 }
+
+export function generateOTPCode() {
+  return randomInt(100000, 999999).toString();
```
Betree: I understand why we'd want to avoid 000000 (which could look like a bug / missing security). But 097512 would be a valid code, and removing 100000 possibilities is reducing the attack complexity by a strong 10%. So I would advise at least doing:
  
```suggestion
  return randomInt(1, 999999).toString();
```

6-digit seems ok for confirming the email address on a new account.
However, going further, we're likely going to reuse some of this logic in the "Sign in" email, to allow people to copy a code rather than clicking the magic link. In this context, the 6-digit approach will not be sufficient. 3 attempts with 1,000,000 possibilities (9 digits) - 1 chance over 333,333 to get the right code - seems risky, considering how sensitive our data is.

To set a strong pattern from the start that we'll be able to re-use (both API and UI components), I'm going to advocate for going with an alphanumeric code (e.g., 6JNJLG), like Slack or Steam. With a 6-character alphanumeric code, the complexity suddenly bumps from 1,000,000 possibilities to 2,176,782,336 (considering we only do uppercase). We lose a bit of usability in the process, but I think the x2177 safety gain is worth it!

test/server/controllers/users.test.ts
Betree, line 1: I got confused by this as well in https://github.com/opencollective/opencollective-api/pull/11184 recently, but we already have tests for this feature in `test/server/routes/users.test.js`, I'd suggest adding yours to this file instead.

test/test-helpers/fake-data.ts
```diff
@@ -106,6 +106,10 @@ export const randArray = (generateFunc, min = 1, max = 1) => {
   return [...Array(arrayLength)].map((_, idx) => generateFunc(idx, arrayLength));
 };
 
+export const randIp = () => {
```
Betree: 🍨 

```suggestion
export const randIpV4 = () => {
```

server/controllers/users.ts
```diff
@@ -179,6 +183,167 @@ export const signin = async (req, res, next) => {
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    next,
+  ) {
+    const ipRateLimit = new RateLimit(`signup_ip_${req.ip}`, 3, 15 * 60);
+    if (!(await ipRateLimit.registerCall())) {
+      res.status(403).send({
+        error: { message: 'Rate limit exceeded' },
+      });
+      return;
+    }
```
Betree: Good that you implemented rate limiting, but the recent attack/outage we suffered on emails shows we cannot afford not to have CAPTCHA on this endpoint. And that's without speaking about the SPAM accounts.

server/controllers/users.ts
```diff
@@ -179,6 +183,167 @@ export const signin = async (req, res, next) => {
...
+            userAgent: req.header('user-agent'),
+          },
+        },
+      });
+      if (password) {
+        await user.setPassword(password);
+      }
+    }
+
+    const otp = randomInt(100000, 999999).toString();
```
Betree:
```suggestion
    const otp = generateOTPCode();
```

server/controllers/users.ts
```diff
@@ -179,6 +183,167 @@ export const signin = async (req, res, next) => {
...
+    const secret = await bcrypt.hash(otp, 10);
+    const session: SignupResquestSession = {
+      secret,
+      tries: 0,
+      userId: user.id,
+    };
+    await sessionCache.set(otpSessionKey, session, 5 * 60); // 5 minutes
```
Betree: - 5 minutes seems very short to confirm an email, can we extend to 30min?
- I'd suggest putting this duration in `server/lib/auth.ts`, along token expirations durations, and reusing that in the email template to make sure we use the same value everywhere.

server/routes.ts
```diff
@@ -64,6 +64,8 @@ export default async (app: express.Application) => {
    * Sign In related features
    */
   app.post('/users/signin', required('user'), users.signin);
+  app.post('/users/signup', users.signup);
+  app.post('/users/verify-email', users.verifyEmail);
```
Betree, line 69: Why implement it as legacy routes rather than GraphQL requests?

server/controllers/users.ts
```diff
@@ -179,6 +183,167 @@ export const signin = async (req, res, next) => {
...
+
+    const otpSessionKey = `otp_signup_${sanitizedEmail}`;
+    const otpSession: SignupResquestSession = await sessionCache.get(otpSessionKey);
+    if (otpSession) {
+      const user = await models.User.findByPk(otpSession.userId);
+      if (user) {
+        const validOtp = await bcrypt.compare(otp, otpSession.secret);
+        if (validOtp) {
+          await sessionCache.delete(otpSessionKey);
+          // Verify user and create public profile
```
Betree: "create public profile"?

server/controllers/users.ts
```diff
@@ -179,6 +183,167 @@ export const signin = async (req, res, next) => {
...
+          await user.update({ confirmedAt: new Date() });
+          res.send({ success: true });
+          return;
+        } else {
+          const tries = otpSession.tries + 1;
```
Betree: This is not safe against parallelized attacks; there are 2 async checks between retrieving the session and updating it here. If you send thousands of requests at once, chances are many of them will pass with this check being ineffective. We need some kind of lock.

Betree (changes requested):
As it stands, I see no change on `signin`. My assumption is that it's possible to:
- Sign up with this new endpoint, setting a password
- A user entry is created
- Sign in without confirming the email, using the password, since sign-in doesn't seem to block if `confirmedAt` is null

We should make sure this is not possible (with a unit test).

server/controllers/users.ts
kewitz: This is supposed to be short for security reasons. I'll add a "resend code" mechanism to allow the user to generate a new OTP after a few minutes.

server/graphql/v2/object/Individual.ts
```diff
@@ -55,6 +55,15 @@ export const GraphQLIndividual = new GraphQLObjectType({
...
         },
       },
+      requiresProfileCompletion: {
+        type: new GraphQLNonNull(GraphQLBoolean),
+        async resolve(account: Collective, _, req) {
+          const user = await req.loaders.User.byCollectiveId.load(account.id);
+          if (req.remoteUser?.id === user.id) {
+            return Boolean(account.data?.requiresProfileCompletion);
+          }
+        },
```
Betree: Type is non-null, but the field seems like it can return an undefined value?

server/graphql/v2/input/AccountUpdateInput.js
```diff
@@ -24,6 +24,7 @@ export const GraphQLAccountUpdateInput = new GraphQLInputObjectType({
     legalName: { type: GraphQLString },
     slug: { type: GraphQLString },
     description: { type: GraphQLString },
+    image: { type: GraphQLString },
```
Betree:
```suggestion
    image: { type: GraphQLURL },
```

server/lib/auth.ts
```diff
@@ -65,10 +66,20 @@ export function mustHaveRole(remoteUser, roles, CollectiveId, action = 'perform
...
   res.cookie('accessTokenPayload', [header, payload].join('.'), { maxAge, httpOnly: false, secure: true });
   res.cookie('accessTokenSignature', signature, { maxAge, httpOnly: true, secure: true });
 }
+
+export const OTP_RATE_LIMIT_WINDOW = minutesToSeconds(15);
+export const OTP_RATE_LIMIT_MAX_ATTEMPTS = 5;
+export const OTP_TOKEN_EXPIRATION = minutesToSeconds(5);
+
+export function generateOTPCode(length = 6): string {
+  return randomInt(0, 10 ** length - 1)
```
Betree: Not a required change for confirming the email, but I just want to reiterate that using a safer code will be required for signing in; I think we should ideally consolidate that from the start.

server/controllers/users.ts
```diff
@@ -179,6 +190,283 @@ export const signin = async (req, res, next) => {
...
+    res.status(500).send({
+      error: { message: 'Error sending OTP email', type: 'EMAIL_SEND_ERROR' },
+    });
+    return;
+  }
+
+  res.send({ success: true });
+}
+
+export async function resendOTP(req: express.Request, res: express.Response) {
```
Betree:
```suggestion
export async function resendEmailVerificationOTP(req: express.Request, res: express.Response) {
```

server/controllers/users.ts
```diff
@@ -179,6 +190,283 @@ export const signin = async (req, res, next) => {
...
+    res.status(403).send({
+      error: { message: 'Rate limit exceeded', type: 'RATE_LIMIT_EXCEEDED' },
+    });
+    return;
+  }
+
+  try {
+    await lockUntilOrThrow(
+      `otp:verifyEmail:${sanitizedEmail}`,
+      async () => {
```
Betree, line 436: 👍

server/controllers/users.ts
```diff
@@ -179,6 +191,301 @@ export const signin = async (req, res, next) => {
...
+    res.status(403).send({
+      error: { message: 'Rate limit exceeded', type: 'RATE_LIMIT_EXCEEDED' },
+    });
+    return;
+  }
+
+  // Check if OTP request already exists
+  const otpSession: SignupRequestSession = await sessionCache.get(otpSessionKey);
+  if (otpSession) {
+    // Should we just return SUCCESS true here to take the user to the verify OTP step?
```
gustavlrsn: I guess returning SUCCESS true here and taking the user to the verify OTP step will not work now with the added OTP session id, but I'm wondering if we can/should allow to create a new OTP session and not return anything here?

server/controllers/users.ts
```diff
@@ -179,6 +191,301 @@ export const signin = async (req, res, next) => {
...
+    res.status(403).send({
+      error: { message: 'Rate limit exceeded', type: 'RATE_LIMIT_EXCEEDED' },
+    });
+    return;
+  } else if (!(await ipRateLimit.registerCall())) {
+    res.status(403).send({
+      error: { message: 'Rate limit exceeded', type: 'RATE_LIMIT_EXCEEDED' },
+    });
+    return;
+  }
```
gustavlrsn: Small thing, but maybe we should check the ip rate limit first and not put the second check in an `else if`, to harmonize with `signup` and `resendEmailVerification`, and not allow bypassing the ip rate limit by using different emails

server/controllers/users.ts
```diff
@@ -179,6 +191,301 @@ export const signin = async (req, res, next) => {
...
+      data: {
+        creationRequest: {
+          ip: req.ip,
+          userAgent: req.header('user-agent'),
+        },
+      },
+    });
+    if (password) {
+      await user.setPassword(password);
+    }
```
gustavlrsn: I guess we can remove this now, having moved the password input to the "Complete your profile step"

templates/emails/user.otp.requested.hbs
```diff
@@ -0,0 +1,23 @@
+Subject: Open Collective: Email Confirmation
+
+{{> header}}
+
+<center>
+  <h3>Your 6-digit code is: <span>{{otp}}</span></h3>
```
gustavlrsn: As we changed from only using digits, maybe we can simply say:
```suggestion
  <h3>Your verification code is: <span>{{otp}}</span></h3>
```

PR #11187, opened 2025-11-05
Handle async stripe subscription retry

### LINKED ISSUE #8251: Stripe ACH recurring contributions issues
According to https://docs.stripe.com/payments/ach-direct-debit#transaction-failures:

> ACH Direct Debit transactions can fail any time after the payment is initiated through payment confirmation. These failures can occur for a number of reasons, such as:
> 
> -  Insufficient funds
> - An invalid account number
> - A customer disabling debits from their bank account
> 
> **Stripe charges a failure fee in this situation.**

We got feedback from OSC that they had [contributions like #654926](https://opencollective.com/webpack/contributions/654926) being charged 28 times, triggering a total of $112 fees.

We have a mechanism that cancels contributions after a few failures, but it is only applied to credit cards.

**Solutions**
- [x] Immediate workaround: we need to cancel these contributions manually, with our regular helpers.
- [ ] Quick fix: cancel the recurring contributions after 2 failures.
- [ ] Long-term solution: pause the contribution, notify the payee for them to update the payment method or signify that they added funds on their account. Auto-cancel after some time.

cc @hdiniz

### LINKED ISSUE #8856: Stripe SEPA recurring contributions send duplicate order.processed notifications
## Summary

Monthly Stripe SEPA debit (`sepa_debit`) recurring contributions can generate **two** `order.processed` activities (and thank-you emails) per billing cycle, while the ledger correctly records only **one** transaction group.

Reported via support ticket: https://opencollective.freshdesk.com/a/tickets/412062

## Impact

- Contributors receive duplicate thank-you emails each month.
- Activity feed shows two `order.processed` entries per charge.
- Ledger / accounting is **not** double-counted (only one `TransactionGroup` is created when the payment actually settles).

## Example (Order `#832222`)

Payment method: `stripe` / `sepa_debit`

Observed pattern every month since at least 2025-06:

| Month | Activities | Transaction groups |
|-------|------------|--------------------|
| 2026-06 | 2 (`#11992531`, `#12041024`) | 1 (`e66c09c8-f156-4e0c-9965-09283695fc17`) |
| 2026-05 | 2 (`#11815870`, `#11867186`) | 1 (`c2486bda-74b0-4ed6-ab7f-419954e3b63a`) |
| 2026-04 | 2 (`#11678455`, `#11736116`) | 1 (`b8aab027-1854-4084-8a71-9a9105b86b22`) |

Typical timing:
- **First activity** (~1st of month): cron initiates the SEPA debit.
- **Second activity** (~8-12 days later): payment settles and the Stripe webhook records the transaction.

Activity data distinguishes the two sources:
- First activity: `firstPayment: false`, no real transaction UUID in activity data (synthetic `createdAt` fallback).
- Second activity: `firstPayment: true` (incorrect for a renewal), real transaction UUID matching the ledger entry.

June 2026 example:
- Activity `#11992531` (2026-06-01): no transaction UUID.
- Activity `#12041024` (2026-06-10): transaction UUID `98440b71-a98c-4d10-a403-1dc8a5e68b8a`, aligns with transaction `#12573295` and PaymentIntent `pi_3TddyTHMlsn9E37I0PoICYMB`.

## Root cause

Two independent code paths both create `order.processed` for the same recurring charge:

### 1. Recurring contributions cron (`charge-recurring-contributions`)

When `processOrder()` returns without a transaction (PaymentIntent still `processing`), the cron still treats the charge as **success** and calls `recordOrderProcessed()`:

- `opencollective-api/server/lib/recurring-contributions.js` — `processOrderWithSubscription()` sets `orderProcessedStatus = 'success'` even when `processOrder()` returns `undefined`.
- `handleRetryStatus()` calls `recordOrderProcessed(order, transaction)` with `transaction = undefined`.

For SEPA, `opencollective-api/server/paymentProviders/stripe/payment-intent.ts` explicitly returns early when status is `processing` (no ledger transaction created yet).

### 2. Stripe webhook (`payment_intent.succeeded`)

When the SEPA debit eventually settles, the webhook creates the ledger transaction and sends a second thank-you via `sendEmailNotifications()` → `recordOrderConfirmation()` → `createProcessedActivity()`:

- `opencollective-api/server/paymentProviders/stripe/webhook.ts`
- `opencollective-api/server/lib/payments.ts`

The webhook correctly deduplicates **transactions** (checks for existing charge), but does **not** deduplicate **activities**.

### Why credit card is not affected

Synchronous `creditcard` charges create the transaction inline in `processOrder()` and move the PaymentIntent to `previousPaymentIntents`, so the webhook usually cannot find the order by `data.paymentIntent.id` and does not send a second notification.

## Proposed fix

1. In `processOrderWithSubscription` / `handleRetryStatus`: do **not** call `recordOrderProcessed()` when `processOrder()` returns no transaction for async Stripe payment methods (`sepa_debit`, `us_bank_account`, `bacs_debit`, etc.).
2. Let the `payment_intent.succeeded` webhook be the single source of truth for thank-you notifications on settlement (similar to PayPal's webhook flow).
3. Secondary: `recordOrderConfirmation()` should not set `firstPayment: true` for subscription renewals (`order.SubscriptionId` present).

## Repository

Fix belongs in `opencollective-api`.

## Diagnostic SQL

```sql
-- Payment method type
SELECT o.id, pm.service, pm.type
FROM "Orders" o
JOIN "PaymentMethods" pm ON pm.id = o."PaymentMethodId"
WHERE o.id = 832222;

-- Activities vs transactions per month
SELECT date_trunc('month', a."createdAt") AS month,
       COUNT(*) AS activity_count,
       (SELECT COUNT(DISTINCT t."TransactionGroup")
        FROM "Transactions" t
        WHERE t."OrderId" = 832222
          AND t.type = 'CREDIT'
          AND date_trunc('month', t."createdAt") = date_trunc('month', a."createdAt")) AS txn_groups
FROM "Activities" a
WHERE a."OrderId" = 832222 AND a.type = 'order.processed'
GROUP BY 1
ORDER BY 1 DESC;
```

hdiniz:
Resolve https://github.com/opencollective/opencollective/issues/8251
Resolve https://github.com/opencollective/opencollective/issues/8856


server/lib/payments.ts
```diff
@@ -1025,7 +1033,7 @@ const recordOrderConfirmation = async (order: Order, transaction: Transaction):
       fromCollective: fromCollective.minimal,
       interval,
       monthlyInterval: interval === 'month',
-      firstPayment: true,
+      firstPayment,
```
Betree, line 1044: This default value has been removed and passed as an optional argument, yet I don't see all the references to `sendEmailNotifications` being updated, only one of them. Aren't we going to miss it with other payment methods, when called from `executeOrder`?

hdiniz: Thanks, I kept the existing default now.

Betree (commented):
I didn't have the time to go into a deep review, but the core looks good to me.

PR #11208, opened 2025-11-20
enhancement(GraphQL): Add `hostContext` filter to `OrdersCollectionQuery`

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/8360

# Description

This adds the `HostContext` type as an argument to `OrdersCollectionQuery` and deprecating `includeHostedAccounts` as this is now handled by the `HostContext`.

### DIFF
--- a/server/graphql/v2/query/collection/OrdersCollectionQuery.ts
+++ b/server/graphql/v2/query/collection/OrdersCollectionQuery.ts
@@ -269,22 +269,14 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
 
   const fetchAccountParams = { loaders: req.loaders, throwIfMissing: true };
   const host = args.host && (await fetchAccountWithReference(args.host, fetchAccountParams));
-  let account, oppositeAccount, hostedAccounts, includeHostedAccounts, hostContext;
+  let account, oppositeAccount, hostedAccounts, hostContext;
 
-  // // Override with deprecated includeHostedAccounts argument
+  // Use deprecated includeHostedAccounts argument
   if (args.includeHostedAccounts === true && isNil(args.hostContext)) {
     hostContext = 'ALL';
   } else {
     hostContext = args.hostContext;
   }
-
-  // let includeChildrenAccounts = args.includeChildrenAccounts;
-
-  // Skip including children if `hostContext` is set without `hostedAccounts`
-  // It will not change the result (as all children accounts are also hosted) and only increase query complexity
-  // if (!isNil(args.hostContext) && isNil(args.hostedAccounts)) {
-  //   includeChildrenAccounts = true;
-  // }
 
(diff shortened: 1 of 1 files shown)


server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -235,7 +249,27 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
 
   const fetchAccountParams = { loaders: req.loaders, throwIfMissing: true };
   const host = args.host && (await fetchAccountWithReference(args.host, fetchAccountParams));
-  let account, oppositeAccount, hostedAccounts;
+  let account, oppositeAccount, hostedAccounts, includeHostedAccounts, excludeHostAccount;
```
Betree: 🍨 if we want to be accurante, there can be multiple host accounts excluded by this (host account + host's projects/events).

```suggestion
  let account, oppositeAccount, hostedAccounts, includeHostedAccounts, excludeHostAccounts;
```

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -235,7 +249,27 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
...
+      break;
+    case 'INTERNAL':
+    default:
+      includeHostedAccounts = false;
+      excludeHostAccount = false;
+  }
+  // Override with deprecated includeHostedAccounts argument
+  if (args.includeHostedAccounts) {
+    includeHostedAccounts = args.includeHostedAccounts;
+  }
```
Betree: To handle the case where we'd be passing `false`

```suggestion
  if (!isNil(args.includeHostedAccounts)) {
    includeHostedAccounts = args.includeHostedAccounts;
  }
```

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -66,6 +68,11 @@ const getCollectivesJoinCondition = (
         ...(limitToHostedAccountsIds.length ? { [`$${association}.id$`]: { [Op.in]: limitToHostedAccountsIds } } : {}),
       },
     ];
+
+    if (excludeHostAccount) {
+      conditions.push({ [`$${association}.id$`]: { [Op.not]: account.id } });
+      conditions.push({ [`$${association}.ParentCollectiveId$`]: { [Op.not]: account.id } });
```
Betree: All these conditions end up in an OR, so it looks like we're saying:
- Return all accounts
- _If one of the previous conditions match_
- **Or** if it's not the parent account
- **Or** if it's not a child of the parent account

As a result, it looks like we're going to return all accounts on the platform?

PR #11212, opened 2025-11-24
Users signup to record Collective

kewitz:
- [x] Users will no longer be created without collective while waiting for verification.
- [x] Adding new flag to prevent OTP waiting users to sign in with same email.
- [x] Creates suspended user collectives to prevent Collective page from being rendered.

### DIFF
--- a/server/controllers/users.ts
+++ b/server/controllers/users.ts
@@ -2,6 +2,7 @@
 import bcrypt from 'bcrypt';
 import config from 'config';
 import type express from 'express';
+import { omit } from 'lodash';
 
 import { activities } from '../constants';
 import { ENGINEERING_DOMAINS } from '../constants/engineering-domains';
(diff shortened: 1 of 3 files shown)


server/controllers/users.ts
```diff
@@ -266,15 +265,27 @@ export async function signup(req: express.Request, res: express.Response) {
...
+              userAgent: req.header('user-agent'),
+            },
+            requiresVerification: true,
+          },
         },
-      },
+        { transaction },
+      );
+      const collective = await user.createCollective(
+        { name: sanitizedEmail.split('@')[0], data: { requiresProfileCompletion: true, isSuspended: true } },
```
Betree, line 284: To iterate on in a separate PR: email should not be used to construct name

`myemail+opencollective@gmail.com`
=> name = "myemail+opencollective`

mysecretemail@gmail.com
=> mysecretemail

Attacker can guess real email: mysecretemail@gmail.com, mysecretemail@outlook.com

PR #11221, opened 2025-11-26
Base KYC impl

hdiniz:
Related https://github.com/opencollective/opencollective/issues/8341


server/constants/plans.ts
```diff
@@ -274,6 +274,7 @@ const proFeatures: CommercialFeaturesType[] = [
   PlatformFeature.TAX_FORMS,
   PlatformFeature.OFF_PLATFORM_TRANSACTIONS,
   PlatformFeature.FUNDS_GRANTS_MANAGEMENT,
+  PlatformFeature.KYC,
```
kewitz: Just make sure to update the frontend with this, otherwise it can break the platform subscription mutation.

server/graphql/v2/input/RequestKYCVerificationInput.ts
```diff
@@ -0,0 +1,28 @@
...
+  fields: () => ({
+    legalName: {
+      type: new GraphQLNonNull(GraphQLString),
+    },
+    legalAddress: {
+      type: new GraphQLNonNull(GraphQLString),
+    },
+    notes: {
+      type: GraphQLString,
+    },
```
kewitz, line 14: Not sure exactly what kind of data is expected here, are we storing the customer legal name and address or is this some sort of reason/description the fiscal-host is passing on why they need these fields?

Maybe add description.

server/graphql/v2/object/KYCStatus.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { KYCProviderName } from '../../../models/KYCVerification';
+
+import { GraphQLKYCVerification } from './KYCVerification';
+
+export const GraphQLKYCStatus = new GraphQLObjectType({
+  name: 'KYCStatus',
+  description: 'A individual KYC verified status',
+  fields: () =>
+    Object.fromEntries(Object.values(KYCProviderName).map(provider => [provider, { type: GraphQLKYCVerification }])),
```
kewitz, line 11: Couldn't a fiscal-host can hold multiple KYC status, for the same provider and account?

server/graphql/v2/query/collection/KYCVerificationCollection.ts
```diff
@@ -0,0 +1,17 @@
...
+  interfaces: [GraphQLCollection],
+  fields: () => {
+    return {
+      ...CollectionFields,
+      nodes: {
+        type: new GraphQLList(new GraphQLNonNull(GraphQLKYCVerification)),
+      },
+    };
+  },
+});
```
kewitz, line 17: This should actually be `server/graphql/v2/collection/KYCVerificationCollection.ts`, the query folder typically only contains the query resolver.

server/lib/kyc/providers/base.ts
```diff
@@ -0,0 +1,55 @@
...
+> {
+  providerName: KYCProviderName;
+
+  constructor(providerName: KYCProviderName) {
+    this.providerName = providerName;
+  }
+
+  abstract request(req: KYCRequest, providerRequest: ProviderKYCRequest): Promise<ProviderKYCVerification>;
+  abstract getVerifiedName(kycVerification: ProviderKYCVerification): string;
+  abstract getVerifiedAddress(kycVerification: ProviderKYCVerification): string;
```
kewitz: Perhaps having a unified `getVerifiedData` with typed returned data would be more future-proof. That way we wouldn't need to implement a getter for each new property we add, we could just update the type.

server/lib/security/expense.ts
```diff
@@ -502,6 +504,28 @@ export const checkExpensesBatch = async (
...
+      );
+      for (const kycVerification of kycVerifications) {
+        if (!kycVerification) {
+          continue;
+        }
+        // TODO match kyc fields with expense details
+        const provider = getKYCProvider(kycVerification.provider);
+        addBooleanCheck(checks, true, {
+          scope: Scope.USER,
+          level: Level.MEDIUM,
```
kewitz, line 521: If the user is verified, I think this could be a green checkmark: `Level.PASS`.

server/graphql/v2/input/RequestKYCVerificationInput.ts
hdiniz: Will add field descriptions, for manual KYC the host will insert a legalName and legalAddress for the individual and the notes allow them to indicate where or how the individual was verified.

server/graphql/v2/object/KYCStatus.ts
hdiniz: The intended purpose of this field was to return the current **verified** status of the user for each provider, they could have a verification history (past verified/revoked/failed) which is returned by another resolver. And they could have multiple verifications with different providere (e.g. Persona, paypal, etc).
I was not sure exactly how to name this field to reflect this, but this is the intention

server/lib/security/expense.ts
hdiniz: Agreed, I was hesitant to do that here at this point since I am only checking that they are verified, but not checking against the expense details. I would like to iterate on this so that we can cross reference with the payout method data

server/lib/collectivelib.ts
```diff
@@ -173,6 +173,7 @@ export const COLLECTIVE_SETTINGS_KEYS_LIST = [
   'showSetupGuide',
   'showInitialOverviewSubscriptionCard',
   'canHostAccounts', // Whether this account can host other accounts
+  'kyc',
```
kewitz, line 176: Is this being used?

server/lib/allowed-features.ts
```diff
@@ -186,6 +186,10 @@ const FeaturesAccess: Partial<
       CollectiveType.VENDOR,
     ],
   },
+  [FEATURE.KYC]: {
+    accountTypes: [CollectiveType.ORGANIZATION],
+    onlyAllowedFor: [FEATURE_ACCESS_PARTY.FIRST_PARTY_HOSTS, FEATURE_ACCESS_PARTY.PLATFORM_ACCOUNTS],
```
kewitz, line 191: This can contradict the advertised plan features.

test/server/lib/kyc/base.test.ts
```diff
@@ -0,0 +1,41 @@
...
+      const org = await fakeOrganization();
+      const user = await fakeUser();
+      const kycVerification = await fakeKYCVerification({
+        provider: KYCProviderName.MANUAL,
+        CollectiveId: user.collective.id,
+        RequestedByCollectiveId: org.id,
+        status: KYCVerificationStatus.VERIFIED,
+      });
+
+      await testProvider.revoke(kycVerification);
```
kewitz, line 36: Maybe a method in the model for revoking could become interesting once we have more providers.

kewitz (approved):

test/server/lib/kyc/base.test.ts
hdiniz: There was an issue with circular dependency here that was preventing that. I fixed it in this latest push.

server/lib/allowed-features.ts
hdiniz: True, I pushed a change to remove it from the commercial features for now while we dont release it to everyone.

server/lib/collectivelib.ts
hdiniz: Used in the frontend to set a flag to skip KYC introduction in the request KYC flow

PR #11233, opened 2025-12-03
feat: Script to convert independent collectives to orgs

Betree:
For https://github.com/opencollective/opencollective/issues/8354

### DIFF
--- a/test/scripts/accounts/convert-independent-to-organizations.test.ts
+++ b/test/scripts/accounts/convert-independent-to-organizations.test.ts
@@ -33,7 +33,6 @@ describe('scripts/accounts/convert-independent-to-organizations', () => {
     expect(independentCollective.isHostAccount).to.be.true;
     expect(independentCollective.hasMoneyManagement()).to.be.true;
     expect(independentCollective.hasHosting()).to.be.false; // Hosting should be disabled
-    expect(independentCollective.data?.canHostAccount).to.be.false;
     expect(independentCollective.settings?.canHostAccounts).to.be.false;
 
     // Verify activity was created


scripts/accounts/convert-independent-to-organizations.ts
```diff
@@ -0,0 +1,97 @@
...
+
+const main = async (options: { isDryRun: boolean; limit?: number; offset?: number; slug?: string }) => {
+  const { count, rows: collectives } = await Collective.findAndCountAll({
+    order: [['id', 'ASC']],
+    limit: options.limit,
+    offset: options.offset,
+    where: {
+      type: CollectiveType.COLLECTIVE,
+      HostCollectiveId: sequelize.literal('"HostCollectiveId" = "id"'),
+      deactivatedAt: null,
```
Betree: This will target almost 6400 accounts.

scripts/accounts/convert-independent-to-organizations.ts
```diff
@@ -0,0 +1,97 @@
...
+  for (const collective of collectives) {
+    try {
+      logger.info(`Converting ${collective.slug} (id: ${collective.id}) from ${collective.type} to ORGANIZATION...`);
+
+      if (!options.isDryRun) {
+        // Update the type to ORGANIZATION
+        await collective.update({
+          type: CollectiveType.ORGANIZATION,
+          data: {
+            ...collective.data,
```
znarf: `canHostAccounts` should be set to false in data.

scripts/accounts/convert-independent-to-organizations.ts
znarf: deactivatedAt can still be reactivated? So should be converted?

Betree: There are very few, likely an edge case anyway given we're now setting `HostCollectiveId` to null when archiving. Updating still.

scripts/accounts/convert-independent-to-organizations.ts
Betree: I copied what was done in `convertAccountToOrganization`; I suppose it should be updated as well.

scripts/accounts/convert-independent-to-organizations.ts
znarf: Maybe we can write some checks or update the DB to make sure things are consistent.

scripts/accounts/convert-independent-to-organizations.ts
znarf: It's being taken care of in `activateMoneyManagement` actually.

znarf: But it seems that `activateMoneyManagement` will be partly skipped because it's technically already activated for independent Collectives.

znarf: The behavior is generally different than `convertAccountToOrganization` because in the mutation, the Collective is supposed to be "unhosted" while here they are "Independent" and already active.

scripts/accounts/convert-independent-to-organizations.ts
```diff
@@ -0,0 +1,102 @@
...
+
+      if (!options.isDryRun) {
+        await collective.update({
+          type: CollectiveType.ORGANIZATION,
+          data: {
+            ...collective.data,
+            canHostAccount: false,
+          },
+        });
+        await collective.activateMoneyManagement(null, { silent: true });
```
znarf:
```suggestion
        await collective.activateMoneyManagement(null, { force: true, silent: true });
```

znarf: Added support for `force` parameter in the base PR.

scripts/accounts/convert-independent-to-organizations.ts
```diff
@@ -0,0 +1,102 @@
...
+          type: CollectiveType.ORGANIZATION,
+          data: {
+            ...collective.data,
+            canHostAccount: false,
+          },
+        });
+        await collective.activateMoneyManagement(null, { force: true, silent: true });
+        await models.Activity.create({
+          type: activities.COLLECTIVE_CONVERTED_TO_ORGANIZATION,
+          UserId: null,
```
znarf, line 54: This activity is displayed in the timeline / activity log, is it looking ok without userId?

scripts/accounts/convert-independent-to-organizations.ts
```diff
@@ -0,0 +1,102 @@
...
+  for (const collective of collectives) {
+    try {
+      logger.info(`Converting ${collective.slug} (id: ${collective.id}) from ${collective.type} to ORGANIZATION...`);
+
+      if (!options.isDryRun) {
+        await collective.update({
+          type: CollectiveType.ORGANIZATION,
+          data: {
+            ...collective.data,
+            canHostAccount: false,
```
znarf: We can skip this now that it is addressed in activateMoneyManagement

scripts/accounts/convert-independent-to-organizations.ts
Betree: Omitting UserId is a standard practice for system activities. Should be fine given that the string in the frontend is not using "User".

PR #11238, opened 2025-12-04
Organization follow up

znarf:
- convertOrganizationToCollective


server/graphql/v2/mutation/OrganizationMutations.ts
```diff
@@ -60,15 +64,13 @@ export default {
...
-      financiallyActive: {
+      hasMoneyManagement: {
         type: GraphQLBoolean,
         description:
           'If true, the organization will be created with Money Management activated, allowing the organization to receive contributions and pay for expenses. Defaults to false.',
         defaultValue: false,
       },
-      // TODO(hasHosting): rename this flag with appropriate future proof name
-      fiscalHostCapable: {
+      hasHosting: {
```
Betree, line 73: This will be a breaking change when deploying the frontend; we must make sure to sync both deployments.

PR #11242, opened 2025-12-09
enhancement: Misc improvements around authorization

Betree:


server/middleware/authentication.js
```diff
@@ -382,17 +396,14 @@ export async function checkPersonalToken(req, res, next) {
           await personalToken.update({ lastUsedAt: new Date() });
         }
       }
+
       req.personalToken = personalToken;
-      const collectiveId = personalToken.CollectiveId;
-      if (collectiveId) {
-        req.loggedInAccount = await models.Collective.findByPk(collectiveId);
```
Betree, line 388: @znarf I could not find a single place where this `loggedInAccount` was being used, I assume it was a refactor leftover?

server/middleware/authentication.js
Betree, line 1: Goals with these changes:
- Single SQL query to load all (vs 3 before)
- Do not assume that `personalToken.collective` = `user.collective`, make sure the user has access to the account it refers to.
- Remove unused `req.loggedInAccount`

server/graphql/v2/mutation/MemberMutations.ts
```diff
@@ -371,16 +371,38 @@ const memberMutations = {
...
+              attributes: ['id'],
+              transaction,
+            });
+
+            if (users.length > 0) {
+              await models.PersonalToken.destroy({
+                where: { UserId: users.map(user => user.id), CollectiveId: account.id },
+                transaction,
+              });
+            }
```
Betree, line 402: Making sure that we delete any personal tokens that members may have setup with this account if they get removed as members.

server/middleware/authentication.js
znarf: Yes, I agree, can be removed. Maybe it was an intent to move out of `remoteUser` and get direct access to the Collective profile. In the GraphQL API, we try to avoid the User model. But as `remoteUser` looks established, no reason to keep it at this point.

PR #11245, opened 2025-12-10
enhancement(Expense): tweak permissions

Betree:
Resolve https://github.com/opencollective/opencollective/issues/8393

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -285,13 +285,13 @@ const isAdminOfCollectiveWithPermissivePayoutMethodPermissions = async (
   return Boolean(loosePermissionsPolicy);
 };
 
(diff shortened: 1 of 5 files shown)


server/graphql/common/expenses.ts
```diff
@@ -546,12 +554,6 @@ export const canEditExpense: ExpensePermissionEvaluator = async (
   options = { throw: false },
 ) => {
   if (!validateExpenseScope(req, options)) {
-    if (options?.throw) {
-      throw new Forbidden(
-        'Your current token is missing the necessary scope',
-        EXPENSE_PERMISSION_ERROR_CODES.INVALID_SCOPE,
-      );
-    }
```
Betree, line 554: Removed all these to simplify, validateExpenseScope is already throwing

PR #11249, opened 2025-12-11
Allow memberOf to filter by isFrozen flag

kewitz:


server/graphql/v2/interface/IsMemberOf.js
```diff
@@ -193,6 +209,7 @@ export const IsMemberOfFields = {
         limit: args.limit,
         offset: args.offset,
         order,
+        logging: true,
```
Betree, line 212: Leftover

PR #11252, opened 2025-12-12
debt(Collective): move to hasHosting flag

Betree:
Part of https://github.com/opencollective/opencollective/issues/8404

Follow-up PR to remove `canHostAccounts` flag to be deployed separately (prevent breaking changes).

### DIFF
--- a/server/models/Collective.ts
+++ b/server/models/Collective.ts
@@ -992,7 +992,6 @@ class Collective extends Model<
     if (!this.hasMoneyManagement() || force) {
       const updatedValues = {
         isHostAccount: true,
-        hasHosting: false,
         plan: parseToBoolean(config.features?.newPricing) ? undefined : 'start-plan-2021',
         hostFeePercent: undefined,
         platformFeePercent: undefined,
@@ -1102,7 +1101,7 @@ class Collective extends Model<
     const settings = this.settings ? cloneDeep(this.settings) : {};
     unset(settings, 'paymentMethods.manual');
 
-    await this.update({ isHostAccount: false, hasHosting: false, plan: null, settings });
+    await this.update({ isHostAccount: false, plan: null, settings });
 
     await PayoutMethod.destroy({
       where: {


migrations/20251212124128-add-hasHosting-to-collectives.ts
```diff
@@ -0,0 +1,25 @@
...
+    await queryInterface.sequelize.query(`
+      UPDATE "Collectives"
+      SET "hasHosting" = TRUE
+      WHERE "isHostAccount" = TRUE
+        AND (
+          "settings"->>'canHostAccounts' IS NULL
+          OR ("settings"->>'canHostAccounts')::boolean != FALSE
+        )
+    `);
+  },
```
znarf, line 37: Contrary to current behavior I would be tempted to turn it off for anyone not using the feature today. Instead of assuming it is on for everyone.

Betree: Yep, that was listed in the original issue goals, I had it in mind but I'm building this PR in two steps:
1. Preserving the existing behavior with the new flag
2. Changing behavior

=> Simpler to update tests and detect regressions that way.

I will let you know once it's ready for review.

server/models/Collective.ts
```diff
@@ -995,10 +992,10 @@ class Collective extends Model<
     if (!this.hasMoneyManagement() || force) {
       const updatedValues = {
         isHostAccount: true,
+        hasHosting: false,
```
znarf: Is that still necessary if this is the default?

server/models/Collective.ts
```diff
@@ -1105,7 +1102,7 @@ class Collective extends Model<
     const settings = this.settings ? cloneDeep(this.settings) : {};
     unset(settings, 'paymentMethods.manual');
 
-    await this.update({ isHostAccount: false, plan: null, settings });
+    await this.update({ isHostAccount: false, hasHosting: false, plan: null, settings });
```
znarf: Does not seem necessary as we have a rejection of this case at line 1094.

server/models/Collective.ts
Betree: Not necessary, I did it to be consistent with what we did before. Can remove.

migrations/20251212124128-add-hasHosting-to-collectives.ts
```diff
@@ -0,0 +1,42 @@
...
+        )
+        AND (
+          -- Hosts recently created normally went through a flow where enabling hosting was a conscious decision
+          "createdAt" >= '2025-12-01'
+          -- We only keep hosting for hosts that have hosted at least one collective in the past (intentionally not looking at deletedAt)
+          OR EXISTS (
+            SELECT 1 FROM "Members"
+            WHERE "MemberCollectiveId" = "Collectives"."id"
+              AND "CollectiveId" != "Collectives"."id"
+              AND "role" = 'HOST'
```
znarf, line 33: Members.deletedAt IS NULL

Betree: This is on purpose, see the comment above:
> -- We only keep hosting for hosts that have hosted at least one collective in the past (intentionally not looking at deletedAt)

Betree: The philosophy is: if you had hosting activities in the past, let's not remove hosting for you - even if you're not using these capacities at the moment.

znarf: Sorry, missed the comment.

PR #11253, opened 2025-12-15
Make address option for kyc verification, add check to payee

hdiniz:
Related https://github.com/opencollective/opencollective/issues/8341

Resolves https://github.com/opencollective/opencollective/issues/8369

### DIFF
--- a/server/lib/security/expense.ts
+++ b/server/lib/security/expense.ts
@@ -503,15 +503,15 @@ export const checkExpensesBatch = async (
         },
       );
 
+      const hostCollectiveId = expense.HostCollectiveId || expense.collective?.HostCollectiveId;
+
       // KYC Verification Check: only for individual payees
       const payeeCollective =
         expense.fromCollective || (await req.loaders.Collective.byId.load(expense.FromCollectiveId));
       if (payeeCollective?.type === CollectiveType.USER) {
         const kycVerifications = await Promise.all(
           Object.values(KYCProviderName).map(provider =>
-            req.loaders.KYCVerification.verifiedStatusByProvider(expense.HostCollectiveId, provider).load(
-              payeeCollective.id,
-            ),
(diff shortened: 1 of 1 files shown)


server/lib/kyc/providers/manual.ts
```diff
@@ -37,8 +37,8 @@ class ManualKYCProvider extends KYCProvider<ManualKYCRequest, ManualKYCVerificat
         notes: providerRequest.notes ?? '',
       },
       data: {
-        legalAddress: providerRequest.legalAddress,
         legalName: providerRequest.legalName,
+        ...(providerRequest.legalAddress ? { legalAddress: providerRequest.legalAddress } : {}),
```
kewitz, line 42: As a suggestion, `Object.assign` accepts false arguments:
```ts
data: Object.assign({ legalName: providerRequest.legalName }, providerRequest.legalAddress && { legalAddress: providerRequest.legalAddress })
```

In this particular situation `pick(providerRequest, ['legalName', 'legalAddress'])` would also have the same effect.

server/lib/security/expense.ts
```diff
@@ -503,25 +503,28 @@ export const checkExpensesBatch = async (
...
-      for (const kycVerification of kycVerifications) {
-        if (!kycVerification) {
-          continue;
+      // KYC Verification Check: only for individual payees
+      const payeeCollective =
+        expense.fromCollective || (await req.loaders.Collective.byId.load(expense.FromCollectiveId));
+      if (payeeCollective?.type === CollectiveType.USER) {
+        const kycVerifications = await Promise.all(
+          Object.values(KYCProviderName).map(provider =>
+            req.loaders.KYCVerification.verifiedStatusByProvider(expense.HostCollectiveId, provider).load(
```
kewitz: I'm not sure how the code stands now, but please review exactly in which conditions `expense.HostCollectiveId` is populated. It used to be that it was populated only after the expense was paid, to cover for cases where the collective would change hosts with existing pending expenses.

PR #11254, opened 2025-12-16
Remove mentions of independent collective

znarf:
Blocked by https://github.com/opencollective/opencollective-api/pull/11252

### DIFF
--- a/server/models/Collective.ts
+++ b/server/models/Collective.ts
@@ -135,6 +135,8 @@ import VirtualCard from './VirtualCard';
 
 const debug = debugLib('models:Collective');
 
+const { ORGANIZATION } = CollectiveType;
+
 type Goal = {
   type: string;
   amount: number;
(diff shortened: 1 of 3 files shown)


server/graphql/v2/mutation/AccountingCategoriesMutations.ts
```diff
@@ -69,11 +67,11 @@ export default {
       });
 
       if (
-        isIndependentCollective &&
+        !account.hasHosting &&
```
Betree: ⚠️ We'll need to merge this PR after https://github.com/opencollective/opencollective-api/pull/11252 for this to work properly

server/graphql/v2/mutation/AccountingCategoriesMutations.ts
```diff
@@ -68,12 +65,13 @@ export default {
         };
       });
 
+      const newCategories = normalizedInputs.filter(c => !c.id);
       if (
-        isIndependentCollective &&
-        normalizedInputs.some(c => c.appliesTo === AccountingCategoryAppliesTo.HOSTED_COLLECTIVES)
+        !account.hasHosting &&
+        newCategories.some(c => c.appliesTo === AccountingCategoryAppliesTo.HOSTED_COLLECTIVES)
```
Betree: Changed this one to only enforce it for new categories. If you've used hosting in the past, we don't want to force you to remove the previous accounting categories.

Betree (approved):

PR #11257, opened 2025-12-16
enhancement(ESlint): prevent logging: true in sequelize

Betree:


kewitz (approved):
Awesome! It works perfectly! We can add an exception to the rule if we ever come to a valid "logging" property elsewhere.

PR #11271, opened 2025-12-19
Refactor: isHostAccount -> hasMoneyManagement

znarf:


server/graphql/v2/interface/IsMemberOf.js
```diff
@@ -117,7 +117,7 @@ export const IsMemberOfFields = {
         collectiveConditions.isIncognito = false;
       }
       if (!isNil(args.isHostAccount)) {
-        collectiveConditions.isHostAccount = args.isHostAccount;
+        collectiveConditions.hasMoneyManagement = args.isHostAccount;
```
Betree, line 121: Looks like this field is not used in the frontend anymore, maybe deprecate it to prevent any confusion?

Betree (approved):

server/graphql/v2/interface/IsMemberOf.js
znarf: Added deprecation

PR #11287, opened 2025-12-30
feat(Orders): Add `createdBy` argument and `createdByUsers` resolver to OrdersCollectionQuery

gustavlrsn:
For https://github.com/opencollective/opencollective/issues/8449

# Description 
- Adds a `createdBy` argument to the OrdersCollectionQuery, allowing filtering on who created the order
- Adds a `createdByUsers` resolver to the OrdersCollectionQuery, to return a list of who created the orders for the current filters

### DIFF
--- a/test/server/graphql/v2/collection/OrdersCollectionQuery.test.ts
+++ b/test/server/graphql/v2/collection/OrdersCollectionQuery.test.ts
@@ -21,6 +21,7 @@ const ordersQuery = gql`
     $hostedAccounts: [AccountReferenceInput]
     $includeHostedAccounts: Boolean
     $accountingCategory: [String]
+    $createdBy: [AccountReferenceInput]
   ) {
     orders(
       account: $account
@@ -30,6 +31,7 @@ const ordersQuery = gql`
       hostedAccounts: $hostedAccounts
       includeHostedAccounts: $includeHostedAccounts
       accountingCategory: $accountingCategory
+      createdBy: $createdBy
     ) {
       totalCount
       nodes {
@@ -54,6 +56,51 @@ const ordersQuery = gql`
           id
           legacyId
           slug
+        }
+        createdByAccount {
+          id
+          legacyId
+          slug
+        }
+      }
+    }
+  }
+`;
+
+const ordersWithCreatedByUsersQuery = gql`
+  query OrdersWithCreatedByUsers(
+    $account: AccountReferenceInput
+    $hostContext: HostContext
+    $filter: AccountOrdersFilter
+    $includeChildrenAccounts: Boolean
+    $expectedFundsFilter: ExpectedFundsFilter
+    $status: [OrderStatus]
+    $createdByUsersLimit: Int
+    $createdByUsersOffset: Int
+    $createdByUsersSearchTerm: String
+  ) {
+    orders(
+      account: $account
+      hostContext: $hostContext
+      filter: $filter
+      includeChildrenAccounts: $includeChildrenAccounts
+      expectedFundsFilter: $expectedFundsFilter
+      status: $status
+    ) {
+      totalCount
+      createdByUsers(
+        limit: $createdByUsersLimit
+        offset: $createdByUsersOffset
+        searchTerm: $createdByUsersSearchTerm
+      ) {
+        totalCount
+        limit
+        offset
+        nodes {
+          id
+          legacyId
+          slug
+          name
         }
       }
     }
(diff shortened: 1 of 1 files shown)


server/graphql/v2/collection/OrderCollection.js
```diff
@@ -13,6 +14,10 @@ export const GraphQLOrderCollection = new GraphQLObjectType({
       nodes: {
         type: new GraphQLList(GraphQLOrder),
       },
+      createdByUsers: {
```
gustavlrsn, line 17: Not sure about the naming, alternative suggestions:
- `creators`?
- `createdBy`?

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -598,6 +619,33 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
...
+
+      const collectiveIds = users.map(u => u.CollectiveId).filter(id => id !== null);
+      if (collectiveIds.length === 0) {
+        return [];
+      }
+
+      return models.Collective.findAll({
+        where: { id: { [Op.in]: collectiveIds } },
+      });
+    },
```
Betree: This solution will not work from a performance perspective.

I you take a host like OSC, the number of individual users for contributions is around ~150.000:

```sql
select COUNT(DISTINCT o."CreatedByUserId")
from "Orders" o
INNER JOIN "Collectives" c ON o."CollectiveId" = c.id
WHERE c."HostCollectiveId" = 11004
AND o."deletedAt" IS NULL
```

Loading 150.000 profiles means taking hundreds of MB in the server RAM, and that's without speaking about data transfer or SQL cost.

The proper solution must be:
1. Paginated: only return the proper subset
2. Searchable: API returns entries based on the typed search term. I doubt that filtering a list of 150.000 entries with frontend JS would provide a responsive UI.

server/graphql/v2/collection/OrderCollection.js
```diff
@@ -13,6 +14,10 @@ export const GraphQLOrderCollection = new GraphQLObjectType({
...
+        type: new GraphQLList(GraphQLAccount),
+        description: 'The accounts that created the orders in this collection, regardless of pagination',
+      },
```
Betree, line 28:
```suggestion
      createdBy: {
        type: new GraphQLNonNull(GraphQLIndividualCollection),
        description: 'The accounts that created the orders in this collection, regardless of pagination',
        args: {
          ...CollectionArgs,
          searchTerm: { type: GraphQLString },
        }
      },
```

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
gustavlrsn: Makes sense, pushed an update to paginate and make it searchable.

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -251,6 +251,10 @@ export const OrdersCollectionArgs = {
     type: GraphQLAccountReferenceInput,
     description: 'Return orders only for this host',
   },
+  createdBy: {
+    type: new GraphQLList(GraphQLAccountReferenceInput),
+    description: 'Return only orders created by these users',
```
Betree:
```suggestion
    description: 'Return only orders created by these users. Limited to 1000 users at a time.',
```

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -584,6 +588,22 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
     where['status'] = { ...where['status'], [Op.ne]: OrderStatuses.PENDING };
   }
 
+  // Store the current where as it will be used to fetch createdByUsers (before applying createdBy filter)
+  const baseWhere = cloneDeep(where);
+
+  if (!isEmpty(args.createdBy)) {
+    const createdByAccounts = await fetchAccountsWithReferences(args.createdBy, fetchAccountParams);
```
Betree, line 638:
```suggestion
    assert(args.createdBy.length < 1000, '"Created by" is limited to 1000 users');
    const createdByAccounts = await fetchAccountsWithReferences(args.createdBy, fetchAccountParams);
```

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -584,6 +588,22 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
...
+    const createdByUsers = await models.User.findAll({
+      attributes: ['id'],
+      where: { CollectiveId: { [Op.in]: createdByAccounts.map(a => a.id) } },
```
Betree: postgres does not dedup

```suggestion
      where: { CollectiveId: { [Op.in]: uniq(createdByAccounts.map(a => a.id)) } },
```

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -598,6 +618,54 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
...
     limit: args.limit,
     offset: args.offset,
+    createdByUsers: async (args: { limit?: number; offset?: number; searchTerm?: string } = {}) => {
+      const { limit = 10, offset = 0, searchTerm } = args;
+
+      const searchConditions = getSearchTermSQLConditions(searchTerm, 'c');
+
+      const orderWhereClause = sequelize.queryInterface.queryGenerator
+        .getWhereConditions(baseWhere, 'o', models.Order)
+        .replace(/"Order"\./g, 'o.');
```
Betree: I can't find this `getWhereConditions` in the docs nor in [current sequelize's source code](https://github.com/search?q=repo%3Asequelize%2Fsequelize+getWhereConditions&type=code). Any link to check what it does and how it's implemented?

gustavlrsn: It's a private function in Sequelize v6 (the version we're using): https://github.com/sequelize/sequelize/blob/v6/src/dialects/abstract/query-generator.js#L2846

But you're right, can't find it in v7. Will refactor this.

gustavlrsn: @Betree I've pushed a refactor to use the Sequelize query builder instead of raw sql here, required some tweaks in how to filter is applied compared to the main collection query. The filters are now manually added to `createdByUsers` resolver, and it is only target those that are relevant to produce the different dashboard tools for now.

server/graphql/v2/query/collection/OrdersCollectionQuery.ts
```diff
@@ -598,6 +661,125 @@ export const OrdersCollectionResolver = async (args, req: express.Request) => {
...
+          args.hostedAccounts,
+        );
+
+        if (!args.filter || args.filter === 'OUTGOING') {
+          fromCollectiveConditions.push(accountConditions);
+        }
+
+        if (!args.filter || args.filter === 'INCOMING') {
+          collectiveConditions.push(accountConditions);
+        }
```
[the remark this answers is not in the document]
gustavlrsn: Made `filter` required to fetch `createdByUsers`

Betree (commented):
The code looks good to me. Can we add some tests to `test/server/graphql/v2/collection/OrdersCollectionQuery.test.ts` to cover this new behavior?

gustavlrsn:
@Betree added some tests for the `createdBy` argument and the `createdByUsers` resolver

PR #11356, opened 2026-01-26
fix: allow collective admins to see their accessible vendors

Betree:


server/graphql/v2/object/Host.ts
```diff
@@ -1482,15 +1482,64 @@ export const GraphQLHost = new GraphQLObjectType({
...
+              `),
+              ];
+            }
+          }
+
           const publicVendorPolicy = await getPolicy(account, POLICIES.EXPENSE_PUBLIC_VENDORS);
-          const isAdmin = req.remoteUser?.isAdminOfCollective(account);
-          if (!publicVendorPolicy && !isAdmin) {
+
+          if (!publicVendorPolicy && !isAdminOfHost && !visibleToAccountIds?.length) {
```
hdiniz, line 1542: lgtm, but we seem to be changing a policy behavior here, is this alright?

If `EXPENSE_PUBLIC_VENDORS = false` and we pass  `visibleToAccounts = [1234]`, this query will return vendors explicitly visible to 1234 but also vendors without visibilityLimitations (`OR data#>'{visibleToAccountIds}' IS NULL` )

Maybe if `EXPENSE_PUBLIC_VENDORS = false` we could limit the return to vendors with explicitly visibility setting

Betree:
Better addressed in https://github.com/opencollective/opencollective/issues/8752

PR #11389, opened 2026-02-05
Clear accounting category when moving to different host

znarf:
When expenses are moved to an account under a different host, the AccountingCategoryId is now cleared since accounting categories belong to specific hosts and would be invalid after the move.

The migration log now also captures the previous AccountingCategoryId for audit purposes.


Betree (approved):

Betree (approved):
This is a good change, but it doesn't cover the case we identified with manyverse/java-jda/material-table-core which all share the same story: they moved from a different host (or self hosted) to OSC. OSC then added the accounting category.

`moveExpense` was not the source of the problem in this case.

PR #11456, opened 2026-02-24
Contribution categorization

hdiniz:


migrations/20260204130046-add-contribution-accounting-category-rule.ts
```diff
@@ -0,0 +1,70 @@
+'use strict';
+
+import type { QueryInterface } from 'sequelize';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface: QueryInterface, Sequelize) {
+    await queryInterface.createTable('ContributionAccountingCategoryRules', {
```
Betree: Do we need to be that specific in the model? It sounds like we'll likely want to extend it to other things (expenses) in the future. Why not go with something like `AccountingCategoryRules`?

migrations/20260204130046-add-contribution-accounting-category-rule.ts
```diff
@@ -0,0 +1,70 @@
...
+        type: Sequelize.DATE,
+        allowNull: false,
+        defaultValue: Sequelize.fn('NOW'),
+      },
+      deletedAt: {
+        type: Sequelize.DATE,
+      },
+    });
+
+    await queryInterface.addIndex('ContributionAccountingCategoryRules', ['CollectiveId', 'order'], {
```
Betree: Out of curiosity: I get why we want to index `CollectiveId`, but what's the benefit of indexing on the `order`?

server/constants/contribution-roles.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Defines the possible roles for a user on an contribution.
+ */
+
+export enum ContributionRoles {
+  hostAdmin = 'hostAdmin',
+  accountingRules = 'accountingRules',
+}
```
Betree, line 8: At first glance, when you stumble on this file, it's hard to comprehend what `ContributionRoles.accountingRules` stands for. "accountingRules" does not sounds like a role. Can we find a better name?

server/graphql/v2/input/ContributionAccountingCategoryRuleInput.ts
```diff
@@ -0,0 +1,33 @@
...
+    id: {
+      type: GraphQLString,
+      description: 'The ID of the contribution accounting category rule to edit',
+    },
+    accountingCategory: {
+      type: new GraphQLNonNull(GraphQLAccountingCategoryReferenceInput),
+      description: 'The accounting category to apply the rule to',
+    },
+    name: {
+      type: new GraphQLNonNull(GraphQLString),
```
Betree: 🍨 

```suggestion
      type: new GraphQLNonNull(GraphQLNonEmptyString),
```

server/graphql/v2/mutation/AccountingCategoriesMutations.ts
```diff
@@ -135,6 +139,79 @@ export default {
...
+
+        await ContributionAccountingCategoryRule.destroy({
+          where: { id: toDelete.map(r => r.id) },
+          transaction,
+        });
+
+        await ContributionAccountingCategoryRule.bulkCreate(rules, {
+          transaction,
+          updateOnDuplicate: ['order', 'enabled', 'name', 'predicates', 'AccountingCategoryId', 'updatedAt'],
+        });
```
Betree: 🚫 This code would let users edit rules they are not allowed to delete. When building the `rules` list above, we simply `idDecode` the ID, without loading the entry to make sure it belongs to the account.

We use `richDiffDBEntries` in other places to prevent this issue.

server/graphql/v2/object/AccountingCategory.ts
```diff
@@ -61,3 +61,33 @@ export const GraphQLAccountingCategory = new GraphQLObjectType({
...
+  name: 'ContributionAccountingCategoryRule',
+  description: 'A rule for categorizing contributions',
+  fields: {
+    id: {
+      type: new GraphQLNonNull(GraphQLString),
+      resolve: getIdEncodeResolver(IDENTIFIER_TYPES.CONTRIBUTION_ACCOUNTING_CATEGORY_RULE),
+    },
+    name: {
+      type: new GraphQLNonNull(GraphQLString),
+    },
```
Betree:
```suggestion
      type: new GraphQLNonNull(GraphQLNonEmptyString),
      resolve: getIdEncodeResolver(IDENTIFIER_TYPES.CONTRIBUTION_ACCOUNTING_CATEGORY_RULE),
    },
    name: {
      type: new GraphQLNonNull(GraphQLNonEmptyString),
    },
```

server/lib/accounting/categorization/contribution-rules.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    const collective = order.collective || (await order.getCollective());
+    const host = collective.host || (await collective.getHostCollective());
+
+    if (!host) {
+      return;
+    }
+
+    if (!(await hasFeature(host, FEATURE.CONTRIBUTION_CATEGORIZATION_RULES))) {
```
Betree: Same for the feature flag, I get that we want to brand this tool around contributions for the first release, but I see no reason to not keep the backend agnostic.

```suggestion
    if (!(await hasFeature(host, FEATURE.ACCOUNTING_CATEGORIZATION_RULES))) {
```

server/lib/accounting/categorization/types.ts
Betree, line 1: Sounds like `zod` could have helped to simplify the validations here.

server/lib/payments.ts
```diff
@@ -932,6 +933,8 @@ export const executeOrder = async (
       data: omit(order.data, ['paymentIntent']),
     });
 
+    await applyContributionAccountingCategoryRules(order);
+
```
Betree, line 937: Not a requested change, but out of curiosity - have you considered adding that as a `afterCreate` hook on `Orders`, to have it in a single place and make sure we don't forget any path?

server/models/AccountingCategoryRule.ts
```diff
@@ -0,0 +1,102 @@
...
+    order: {
+      type: DataTypes.INTEGER,
+      allowNull: false,
+      defaultValue: 0,
+    },
+    predicates: {
+      type: DataTypes.JSONB,
+      allowNull: false,
+      defaultValue: [],
+    },
```
Betree, line 104: Not a requested change: I was expecting a model validation on `predicates` here, to make sure it's always validated regardless of how we create the predicate.

server/constants/contribution-roles.ts
Betree: After reviewing the rest of the code, I would suggest:


```suggestion
/**
 * Defines the possible roles that can interact with a contribution.
 */
export enum ContributionRoles {
  hostAdmin = 'hostAdmin',
  accountingRulesEngine = 'accountingRulesEngine',
}
```

server/paymentProviders/stripe/webhook.ts
Betree: Given that there is an `order.update` above, another `order.update` in `applyContributionAccountingCategoryRules`, both without transaction and without reloading the order, it sounds like there is a risk of concurrency issue.

Betree (changes requested):

migrations/20260204130046-add-contribution-accounting-category-rule.ts
hdiniz: I've update the table and model name and added a type enum (currently limited to 'CONTRIBUTION')

migrations/20260204130046-add-contribution-accounting-category-rule.ts
hdiniz: I've remove the order from this index, it was left over when I tried a different approach on the query for a accounts rules

server/constants/contribution-roles.ts
hdiniz: Renamed

server/graphql/v2/mutation/AccountingCategoriesMutations.ts
hdiniz: Good catch, I've added a check and accompanying test case to validate this

server/lib/accounting/categorization/contribution-rules.ts
hdiniz: I've renamed the feature as well as the table and model, while keeping the contribution specific stuff on its own for now. There might be overlap between predicates and subjects when we eventually tackle expenses, but I think its currently contained enough in the that we can reword that when it comes to it.

server/lib/accounting/categorization/types.ts
hdiniz: I'v refactored the validation and normalization to use zod

server/models/AccountingCategoryRule.ts
hdiniz: I've added the same validations here

server/paymentProviders/stripe/webhook.ts
hdiniz: I see what you mean, this is possible, I guess the particularities of the updates that happen mitigate this as there are other awaits inside the side effects. 

I went ahead and refactored this section to execute the effects sequentially, I believe there should not be a need to reload the order explicitly as sequelize will update the internal order representation after the update.

server/lib/payments.ts
hdiniz: There are only a couple of cases where we want to run this at the moment, I had a preference for doing it explicitly in such places to also take into consideration the roles and the semantics of the mutations where the host can perform remove/ignore a category on updating an order

Betree (approved):
All blockers resolved on my end!

PR #11629, opened 2026-04-03
Refactors: Base work for AccountDetails' Financial Controls tab

kewitz:
Related to https://github.com/opencollective/opencollective/issues/8634

### DIFF
--- a/server/graphql/v2/object/CommunityStats.ts
+++ b/server/graphql/v2/object/CommunityStats.ts
@@ -1,4 +1,3 @@
-import type express from 'express';
 import { GraphQLEnumType, GraphQLInt, GraphQLList, GraphQLNonNull, GraphQLObjectType, GraphQLString } from 'graphql';
 import { GraphQLDateTime } from 'graphql-scalars';
 import { compact, flatten, uniq, values } from 'lodash';
@@ -7,14 +6,13 @@ import type { Sequelize } from 'sequelize';
 import { QueryTypes } from 'sequelize';
 
 import ActivityTypes from '../../../constants/activities';
-import { Activity, Collective, Op, sequelize } from '../../../models';
+import { Activity, Op, sequelize } from '../../../models';
 import {
   CommunityHostTransactionSummaryRow,
   CommunityHostYearlyTransactionSummaryRow,
 } from '../../../types/kysely-views';
 import { GraphQLActivityCollection } from '../collection/ActivityCollection';
 import { GraphQLCommunityRelationType } from '../enum/CommunityRelationType';
-import { GraphQLAccount } from '../interface/Account';
 import { CollectionArgs } from '../interface/Collection';
 
 import { GraphQLAmount } from './Amount';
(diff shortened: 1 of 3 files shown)


server/models/Expense.ts
```diff
@@ -106,6 +106,8 @@ class Expense extends ModelWithPublicId<
   declare public RecurringExpenseId: ForeignKey<RecurringExpense['id']>;
   declare public AccountingCategoryId: ForeignKey<AccountingCategory['id']>;
   declare public InvoiceFileId: UploadedFile['id'];
+  declare public approvedByCollectiveId: ForeignKey<Collective['id']>;
```
Betree: An expense can (and should, in many cases) have multiple approvers. I don't think this is the right architecture to represent the current situation. I would rather see a separate table like `ExpenseDecisions`, `ExpenseReviews`.

server/models/Expense.ts
```diff
@@ -106,6 +106,8 @@ class Expense extends ModelWithPublicId<
...
+  declare public paidByCollectiveId: ForeignKey<Collective['id']>;
```
Betree: Only one single payer, so this one is good 👍

server/models/Expense.ts
kewitz: My first implementation was based on an array of IDs, but then, I notice that the multi-user approval is not something really supported:
https://github.com/opencollective/opencollective-api/blob/main/server/graphql/common/expenses.ts#L1446-L1447

Betree: It's not enforced strictly, but it still exists and that's were we want to go to. Better to not introduce some debt that we'll have to cleanup down the road.

Array of IDs are not great, because they're not treated as foreign key references by postgres, so poor indexing and no `onDelete`/`onUpdate`. Join tables are the way to go.

| UserId | FromCollectiveId | ExpenseId | review |
|--------|--------|--------|--- |
| 1 | 1 | 42 | APPROVE |

Betree: (with a unique index on FromCollectiveId/ExpenseId + temporal)

Betree: And to be checked: maybe the same can be achieved by looking at Activities. Could be fast if we have the right indexes.

server/graphql/v2/object/Expense.ts
```diff
@@ -263,6 +265,11 @@ export const GraphQLExpense = new GraphQLObjectType<ExpenseModel, Express.Reques
         type: new GraphQLNonNull(new GraphQLList(GraphQLAccount)),
         description: 'The accounts who approved this expense',
         async resolve(expense, _, req) {
+          const approvedActions = await ExpenseAction.getAll(expense.id, ExpenseActionType.APPROVED);
```
hdiniz: Do we need loaders for this (approvedBy) and the paidBy fields if they are used in expense lists?

migrations/20260403143831-create-expense-actions-table.ts
```diff
@@ -0,0 +1,52 @@
...
+        onDelete: 'CASCADE',
+        onUpdate: 'CASCADE',
+        allowNull: false,
+      },
+      UserId: {
+        type: Sequelize.INTEGER,
+        references: { model: 'Users', key: 'id' },
+        onDelete: 'SET NULL',
+        onUpdate: 'CASCADE',
+        allowNull: true,
```
Betree: I'm surprised that this is nullable, an action without user doesn't seem to make sense?

migrations/20260403143831-create-expense-actions-table.ts
```diff
@@ -0,0 +1,52 @@
...
+        type: Sequelize.DATE,
+        defaultValue: Sequelize.NOW,
+        allowNull: false,
+      },
+      deletedAt: {
+        type: Sequelize.DATE,
+        allowNull: true,
+      },
+    });
+
```
Betree: For security, I would add a unique index:
```
on: 'UserId,ExpenseId',
where: { deletedAt: { [Op.is]: null }, action: ['APPROVED', 'REJECTED' },
```

Each user can only approve or reject a given expense one time. It cannot be both approved and rejected at the same time, and it cannot be approved two times by the same user.

migrations/20260403143831-create-expense-actions-table.ts
```diff
@@ -0,0 +1,52 @@
...
+    await queryInterface.addIndex('ExpenseActions', ['ExpenseId']);
+    await queryInterface.addIndex('ExpenseActions', ['UserId']);
+    await queryInterface.addIndex('ExpenseActions', ['action']);
```
Betree: Not sure why we'd need that one? Are we ever going to query expense actions without filtering by Expense or User?

```suggestion
```

migrations/20260409142055-create-community-admin-controls-summary-materialized-view.ts
```diff
@@ -0,0 +1,49 @@
...
+    await queryInterface.sequelize.query(`DROP MATERIALIZED VIEW IF EXISTS "CommunityAdminControlsSummary";`);
+    await queryInterface.sequelize.query(`
+      CREATE MATERIALIZED VIEW "CommunityAdminControlsSummary" AS
+        SELECT
+          u."CollectiveId" AS "AdminCollectiveId",
+          t."HostCollectiveId",
+          ea.action,
+          h.currency AS "hostCurrency",
+          COALESCE(SUM(ABS(t."amountInHostCurrency")) FILTER (WHERE t.type::text = 'DEBIT'::text), 0::bigint) AS "debitTotal",
+          COALESCE(COUNT(t.id) FILTER (WHERE t.type::text = 'DEBIT'::text), 0::bigint) AS "debitCount",
```
Betree: Since we are not filtering on the kind, that means this debit count will include both the expense and the payment processor fees for any expense paid, resulting in a count of "2" for each expense. Does this match what we are doing in the front end? Same questions for creditCount/refundDebitCount.

scripts/fixes/create-expense-actions.ts
Betree, line 1: Should this be a migration, so that staging/dev envs also benefit from it?

Betree, line 1: I'll get a closer look at this tomorrow

server/graphql/common/expenses.ts
```diff
@@ -1448,8 +1450,11 @@ export const approveExpense = async (req: express.Request, expense: Expense): Pr
   } else if (!(await canApprove(req, expense, { throw: true }))) {
     throw new Forbidden();
   }
-
-  const updatedExpense = await expense.update({ status: 'APPROVED', lastEditedById: req.remoteUser.id });
+  const updatedExpense = await expense.update({
+    status: 'APPROVED',
+    lastEditedById: req.remoteUser.id,
+  });
+  await ExpenseAction.record(expense.id, req.remoteUser.id, ExpenseActionType.APPROVED);
```
Betree: 🍨 Good candidate for a SQL transaction (no requested change).

migrations/20260403143831-create-expense-actions-table.ts
```diff
@@ -0,0 +1,52 @@
...
+        onUpdate: 'CASCADE',
+        allowNull: false,
+      },
+      UserId: {
+        type: Sequelize.INTEGER,
+        references: { model: 'Users', key: 'id' },
+        onDelete: 'SET NULL',
+        onUpdate: 'CASCADE',
+        allowNull: true,
+      },
```
Betree: Optionally, since each of these actions are meant to be linked to a specific activity, we could add an `ActivityId` column. This would allow getting more details about the approval from an activity (or the reverse).

migrations/20260409142055-create-community-admin-controls-summary-materialized-view.ts
kewitz: Sorry, this is a bit of copy-pasta from CommunityTransactionSummary.
On the existing CommunityTransactionSummary this is not relevant because we're filtering by "FromCollectiveId" and these fees end up being attributed to Stripe, Wise, etc.

Here this is probably true, I guess the local data I used to test this didn't had the split fees... I'll need to double-check this.

scripts/fixes/create-expense-actions.ts
kewitz: I'm a bit worried about the time this will take to run in production.
Perhaps I can make a non-production migration that run this script?

Betree: Could be nice, not a big deal if it adds too much complexity

server/graphql/v2/object/Expense.ts
Betree: This feedback is critical, we're fetching this field in multiple lists

migrations/20260416092307-backfill-expense-actions.ts
```diff
@@ -0,0 +1,23 @@
...
+  async up() {
+    if (!['development', 'staging'].includes(process.env.OC_ENV)) {
+      return;
+    }
+
+    await runAll({ sinceId: 0, dry: false });
+  },
+
+  async down() {
+    if (!['development', 'staging'].includes(process.env.OC_ENV)) {
```
Betree: Good to run in other envs too, like e2e. We can simplify:

```suggestion
    if (process.env.OC_ENV !== 'production') {
      // We'll run the migration manually in production
      return;
    }

    await runAll({ sinceId: 0, dry: false });
  },

  async down() {
    if (process.env.OC_ENV !== 'production') {
```

scripts/fixes/create-expense-actions.ts
Betree, line 1: 🍨  Script is in `script/fixes`, but it's not really a fix right? Maybe there's a better place for this?

Betree, line 1: Maybe have a test generated to confirm this script does what it intends to do?

server/graphql/common/expenses.ts
```diff
@@ -1454,10 +1456,20 @@ export const approveExpense = async (req: express.Request, expense: Expense): Pr
...
+        status: 'APPROVED',
+        lastEditedById: req.remoteUser.id,
+      },
+      { transaction },
+    );
 
-  const updatedExpense = await expense.update({ status: 'APPROVED', lastEditedById: req.remoteUser.id });
-  await expense.createActivity(activities.COLLECTIVE_EXPENSE_APPROVED, req.remoteUser);
-  return updatedExpense;
+    await ExpenseAction.clearActions(expense.id, ExpenseActionType.REJECTED, { transaction });
```
Betree: This means that someone approving an expense will clear out the "REJECTED" statuses of all other admins. No blocker at this stage, but it will conflict with the intent to enable multiple approvers.

cron/hourly/50-refresh-materialized-views.js
```diff
@@ -8,6 +8,7 @@ import { runCronJob } from '../utils';
 const VIEWS = [
   'CollectiveOrderStats',
   'ExpenseTagStats',
+  'CommunityAdminControlsSummary',
```
kewitz:
```suggestion
```

PR #11642, opened 2026-04-10
Fix bug creating Locations with structured data and empty stringfied address

kewitz:


gustavlrsn (approved):
Looking good. Nice addition of the model check 👍

PR #11643, opened 2026-04-10
Signal KYC status for organization admins

### LINKED ISSUE #8661: Signal when expense is submitted by an org and one of its hosts needs KYC

hdiniz:
Related https://github.com/opencollective/opencollective/issues/8609
Resolve https://github.com/opencollective/opencollective/issues/8661


server/graphql/v2/object/ExpenseKYCStatus.ts
```diff
@@ -7,9 +9,23 @@ export const GraphQLExpenseKYCStatus = new GraphQLObjectType({
   }),
 });
 
+const GraphQLExpensePayeeKYCType = new GraphQLEnumType({
+  name: 'ExpensePayeeKYCType',
+  description: 'Whether the payee is an individual (its own KYC) or a multi-admin account (rollup over admin KYC)',
+  values: {
+    INDIVIDUAL: { value: 'INDIVIDUAL' },
+    ACCOUNT: { value: 'ACCOUNT' },
```
kewitz: Not a fan of these values because our existing abstractions already point INDIVIDUAL as a type of ACCOUNT.

This also does not match the description of the field resolver.

server/graphql/v2/object/ExpenseKYCStatus.ts
```diff
@@ -20,5 +36,8 @@ const GraphQLExpensePayeeKYC = new GraphQLObjectType({
         },
       }),
     },
+    adminVerifications: {
+      type: new GraphQLList(new GraphQLNonNull(GraphQLKYCVerification)),
+    },
```
kewitz: Perhaps it would be easier to implement this at the GraphQLExpenseKYCStatus level as a payeeAdmins property. This way we wouldn't need to have a GraphQLExpensePayeeKYCType.

server/lib/kyc/expenses/kyc-expenses-check.ts
```diff
@@ -53,34 +51,67 @@ export async function expenseKycStatus(
...
+  const isIndividual = payee.type === CollectiveType.USER;
+  let subjectCollectiveIds: number[];
+  if (isIndividual) {
+    subjectCollectiveIds = [payee.id];
+  } else if (loaders) {
+    subjectCollectiveIds = await loaders.Member.adminMemberCollectiveIdsOfCollective.load(payee.id);
+  } else {
+    const admins = await payee.getAdmins();
+    subjectCollectiveIds = admins.map(a => a.id);
+  }
```
kewitz: Not a big fan of mixing the business-logic `isIndividual` conditional with the existence of loaders in the context or not.
```suggestion
  } else {
    subjectCollectiveIds = loaders ?
      await loaders.Member.adminMemberCollectiveIdsOfCollective.load(payee.id)
      await payee.getAdmins().then(admins => admins.map(a => a.id);
  }
```

server/graphql/v2/object/ExpenseKYCStatus.ts
kewitz: Actually, if the payee (or, the admins of a organization payee) already have a `kycStatus` field, why do we have another one in the expense object?

hdiniz: The kyc status is relative to the host that the individual is interacting with, for that reason we can use the expense to resolve the correct context for the kyc status

kewitz (approved):

PR #11656, opened 2026-04-15
feat(PayPal): remove connected account when removing payout method

### LINKED ISSUE #5957: [Project] Expense to require verified PayPal accounts
## User story 
_As a host admin, I want to enforce that the PayPal user account is verified, so I'm more confident we're not being frauded._

## Best solution for this problem
On the user side, implement an OAuth capability, so the user can connect their PayPal account to Open Collective.
On the host side, create a new Policy to enforce that expenses using PayPal requires a verified account.
When submitting an expense to hosts, we check with PayPal if the user is verified or not. (https://developer.paypal.com/docs/api/identity/v1/#userinfo_get)

## MVP
Implement PayPal OAuth for users and display any visual cue to show whether the account is verified or not. Host admins manually do the check and request user to verify their accounts.

## Metrics 
The number of expenses using PayPal as a gateway to cash out fraudulent money.
The number of fraudulent expenses.

### LINKED ISSUE #8357: Verify PayPal account: disconnect when archiving payout method
Keep the verification status, email, and payee ID stored.

Betree:
Resolve https://github.com/opencollective/opencollective/issues/8357
Resolve https://github.com/opencollective/opencollective/issues/5957 (last issue remaining)

### DIFF
--- a/server/graphql/v2/mutation/PayoutMethodMutations.ts
+++ b/server/graphql/v2/mutation/PayoutMethodMutations.ts
@@ -98,6 +98,7 @@ const payoutMethodMutations = {
           if (connectedAccount) {
             // As of 2026-04-15, PayPal does not offer an API endpoint to revoke user OAuth tokens
             // (https://developer.paypal.com/docs/api/identity/v1/), so we only destroy the record locally.
+            await connectedAccount.update({ token: null, refreshToken: null }, { transaction });
             await connectedAccount.destroy({ transaction });
           }
         }
--- a/server/graphql/v2/mutation/PayoutMethodMutations.ts
+++ b/server/graphql/v2/mutation/PayoutMethodMutations.ts
@@ -8,6 +8,7 @@ import {
   handleKycPayoutMethodReplaced,
 } from '../../../lib/kyc/expenses/kyc-expenses-check';
 import { reportErrorToSentry } from '../../../lib/sentry';
+import sequelize from '../../../lib/sequelize';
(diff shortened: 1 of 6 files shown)


server/graphql/v2/mutation/PayoutMethodMutations.ts
```diff
@@ -78,16 +79,31 @@ const payoutMethodMutations = {
...
+        const linkedConnectedAccountId =
+          payoutMethod.type === PayoutMethodTypes.PAYPAL && paypalData?.isPayPalOAuth
+            ? paypalData?.connectedAccountId
+            : null;
+        if (linkedConnectedAccountId) {
+          const connectedAccount = await models.ConnectedAccount.findByPk(linkedConnectedAccountId, { transaction });
+          if (connectedAccount) {
+            // As of 2026-04-15, PayPal does not offer an API endpoint to revoke user OAuth tokens
+            // (https://developer.paypal.com/docs/api/identity/v1/), so we only destroy the record locally.
+            await connectedAccount.destroy({ transaction });
```
kewitz: I would be OK with hard-deleting this considering we can't revoke tokens.

kewitz (approved):

server/graphql/v2/mutation/PayoutMethodMutations.ts
Betree: I'll go with your alternative proposal of nulling the token column

PR #11659, opened 2026-04-16
enhancement(OrdersCollectionQuery): Add opposite account scope

gustavlrsn:
Project: https://github.com/opencollective/opencollective/issues/8607

Add `oppositeAccountScope` to allow filtering on INTERNAL vs EXTERNAL orders

### DIFF
--- a/server/graphql/schemaV2.graphql
+++ b/server/graphql/schemaV2.graphql
@@ -728,6 +728,11 @@ interface Account {
     host: AccountReferenceInput
 
     """
+    Filter orders by whether the opposite account is INTERNAL or EXTERNAL. For fiscal hosts, internal means within the same host. For regular accounts, internal means within the account and its children. When combined with `filter` (INCOMING/OUTGOING), the opposite account is the other side of the order.
+    """
+    oppositeAccountScope: OppositeAccountScope
+
+    """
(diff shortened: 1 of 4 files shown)


kewitz (commented):
The code looks clean and tests seem to cover all the cases this was design to do.

Nonetheless (and sorry for annoying you with this again), it is my personal opinion that the concept of scope is a bit too abstract for a generic API. Since this is a layer responsible to provide access to data, it should speak the data language and not whatever the interface wants to do.

This is, given that orders are placed _FromAccount_ to _Account_, my opinion is that the API should be able to:
- Fetch Orders from (or not from) collectives of a particular host: 'fromHost' and 'notFromHost'
- Fetch Orders to (or not to) a particular host: 'toHost' and 'notToHost'

In summary, I believe that most of the logic we have here processing `oppositeAccountScope` argument should live in the application and the API should stick to exposing our data in a dumb and strictly relational way.

gustavlrsn:
Agree on the API design points, synced with Leo here on the changes and concluded we'll keep this implementation as is, considering that OrdersCollectionQuery is already implementing a lot of similar semantics as this PR, with `filter` (INCOMING/OUTGOING), `account` and `oppositeAccount`, and that we should aim for a simpler design for payment intents.

PR #11660, opened 2026-04-16
fix(Stripe): fixes on Stripe amounts + 0-decimals rounding

Betree:


server/lib/currency.ts
```diff
@@ -279,7 +283,11 @@ export function convertToCurrency(
   }
 
   return getFxRate(fromCurrency, toCurrency, date).then(fxrate => {
-    return fxrate * amount;
+    const isNegative = amount < 0;
+    const rounded = roundCentsAmount(fxrate * amount, toCurrency);
+    const minAmount = isZeroDecimalCurrency(toCurrency) ? 100 : 1; // Converting an amount can never result in 0
+    const result = Math.max(minAmount, Math.abs(rounded));
+    return isNegative ? -result : result;
```
Betree, line 290: This change is needed because when converting to a zero decimal currency, we need to make sure that the last two digits are always zero, since cents are not supported.

I've also added some logic to make sure that any currency conversion can never result in "0". Even if the fx rate is `0.001`, converting $1 cannot result in something getting free.

The rest is just logic to preserve the currency sign, as this helper supports being called with negative numbers.

server/lib/queries.js
```diff
@@ -26,16 +26,10 @@ const generateFXConversionSQL = async aggregate => {
...
-      const amount = await convertToCurrency(1, 'USD', currency);
-      return `WHEN ${currencyColumn} = '${currency}' THEN ${amountColumn} / ${amount}`;
-    }),
-  );
-
-  return `CASE ${result.join('\n')}ELSE 0 END`;
+  const fxRates = await getFxRates('USD', ['AUD', 'CAD', 'EUR', 'GBP', 'INR', 'MXN', 'SEK', 'USD', 'UYU']);
+  return `CASE ${Object.entries(fxRates)
+    .map(([currency, fxRate]) => `WHEN ${currencyColumn} = '${currency}' THEN ${amountColumn} / ${fxRate}`)
+    .join('\n')}ELSE 0 END`;
```
Betree, line 32: Changed this because `convertToCurrency` now returns a proper amount rounded in cents, but here it was used to get the FX rate really. You can see that we always called it with an amount of `1` and expected the return to be a floating number.

server/lib/stripe.ts
```diff
@@ -76,9 +76,15 @@ export const convertToStripeAmount = (currency, amount) => {
...
+/**
+ * Stripe uses cents (like us), but handle zero decimal currencies differently.
+ * For Stripe, ¥1 = 1, while we keep considering ¥1 as 100, just ensuring accross the codebase that the two last digits are always 00.
+ *
+ * This helper handles this complexity and returns a number that is safe to use in our codebase.
+ */
 export const convertFromStripeAmount = (currency, amount) => {
   if (isZeroDecimalCurrency(currency?.toUpperCase())) {
-    return Math.round(amount * 100);
+    return round(amount * 100, -2); // This ensures that the two last digits are always 00.
```
Betree, line 87: Just to note here: this is not supposed to happen. Stripe normally never returns floating amounts. It's mostly there to make the result type explicit.

test/server/lib/currency.test.js
```diff
@@ -122,18 +122,16 @@ describe('server/lib/currency', () => {
...
-      CurrencyLib.convertToCurrency(1, 'EUR', 'USD', new Date(endDate)).then(amount =>
-        expect(amount).to.equal(1.0533),
-      ));
+      CurrencyLib.convertToCurrency(100, 'EUR', 'USD', new Date(endDate)).then(amount => expect(amount).to.equal(105)));
 
     it('converts INR to USD', () =>
-      CurrencyLib.convertToCurrency(1, 'INR', 'USD', new Date(endDate)).then(amount =>
-        expect(amount).to.equal(0.014962),
+      CurrencyLib.convertToCurrency(1, 'INR', 'USD', new Date(endDate)).then(
+        amount => expect(amount).to.equal(1), // Would be 0.014962, but currency should never be 0
```
Betree, line 134: These tests were wrong, we're never supposed to work with floating amounts 🙈

server/paymentProviders/stripe/common.ts
```diff
@@ -145,28 +145,39 @@ export const createChargeTransactions = async (
...
-  const amountInHostCurrency = convertFromStripeAmount(balanceTransaction.currency, balanceTransaction.amount);
-  const hostCurrencyFxRate = amountInHostCurrency / order.totalAmount;
+  const amountInOrderCurrency =
+    order.currency === hostCurrency
+      ? amountInHostCurrency
+      : await convertToCurrency(amountInHostCurrency, hostCurrency, order.currency);
+
+  const hostCurrencyFxRate = amountInHostCurrency / amountInOrderCurrency;
 
   const hostFee = await getHostFee(order);
```
Betree, line 158: This is the root fix for the stripe async payment amount issue - stop relying on `order.totalAmount` (which may be out of sync with the actual charge), and just rely on the Stripe payload

server/lib/currency.ts
```diff
@@ -279,7 +283,11 @@ export function convertToCurrency(
...
   });
 }
```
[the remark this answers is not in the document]
Betree: Nope, this is what we want. $1 should never equals 0€, even if the FX rate is `0.0000000000001`. Min amount is `0.01€`.

PR #11663, opened 2026-04-20
fix(Expense): improve payout method permissions

Betree:


server/graphql/common/expenses.ts
```diff
@@ -391,11 +391,15 @@ export const canSeeExpensePayoutMethodPrivateDetails: ExpensePermissionEvaluator
...
-  // Submitter can see own information until the expense is paid
-  if (expense.status === expenseStatus.PAID && expense.PayoutMethodId) {
+  if (expense.PayoutMethodId) {
     const payoutMethod = await req.loaders.PayoutMethod.byId.load(expense.PayoutMethodId);
-    if (payoutMethod && !payoutMethod.isSaved) {
-      allowedRoles = allowedRoles.filter(role => role !== isOwner && role !== isOwnerAccountant);
+    if (payoutMethod) {
+      if (
+        (expense.status === expenseStatus.PAID && !payoutMethod.isSaved) || // Submitter can see own information until the expense is paid
+        payoutMethod.CollectiveId !== expense.FromCollectiveId // Remote user is somehow not an admin of the payee
```
kewitz: This should always evaluate to false. Add it as a pre-check at the beginning of this function.

Betree: what about cross-host expenses?

PR #11668, opened 2026-04-20
editPendingOrder to validate fromAccount origin and restrictions

kewitz:


server/graphql/v2/mutation/OrderMutations.ts
```diff
@@ -1344,11 +1344,25 @@ const orderMutations = {
...
       const tier = args.order.tier
         ? await fetchTierWithReference(args.order.tier, { throwIfMissing: true })
         : order.tier;
 
+      if (
+        fromAccount &&
+        fromAccount.HostCollectiveId !== host.id &&
+        !req.remoteUser.isRoot() &&
+        !host.data?.allowAddFundsFromAllAccounts &&
+        !host.data?.isTrustedHost
```
Betree, line 1361: Will this be ok for vendors? I assume they have `HostCollectiveId` set?

Betree (approved):

server/graphql/v2/mutation/OrderMutations.ts
kewitz: No, only the ParentCollectiveId is set. I'll fix this right away.

PR #11689, opened 2026-04-24
Combined Incognito Community Summaries and Contributions base

kewitz:
Related to https://github.com/opencollective/opencollective/issues/8762

Mostly resolves the issue requirements, we're still missing some frontend updates that should be addressed in a separated PR.

Here we're:
- [x] Base refactors to denormalize Incognito references
- [x] Combining Incognito and main Account summaries
- [x] Add createIncognitoProfile V2 mutation
- [x] Add test suite to CommunityQuery
- [x] Rewrite CommunityQuery using Kysely

### DIFF
--- a/migrations/20260424123737-backfill-incognito-user-collective-id.ts
+++ b/migrations/20260424123737-backfill-incognito-user-collective-id.ts
@@ -1,0 +1,42 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+      UPDATE "Collectives" c
+      SET data = COALESCE(c.data, '{}'::jsonb)
+             || jsonb_build_object('UserCollectiveId', m."MemberCollectiveId")
+             || jsonb_build_object('UserId', u.id)
+      FROM "Members" m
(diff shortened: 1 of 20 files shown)


test/server/graphql/v1/mutation.test.js
```diff
@@ -111,6 +111,88 @@ describe('server/graphql/v1/mutation', () => {
...
+        const members = await models.Member.findAll({
+          where: { CollectiveId: created.id },
+        });
+
+        expect(members).to.have.length(1);
+        expect(members[0].role).to.equal(roles.ADMIN);
+        expect(members[0].MemberCollectiveId).to.equal(user1.CollectiveId);
+      });
+
+      it('creates distinct incognito profiles on repeated calls (idempotency via getOrCreateIncognitoProfile is on the model, not mutation)', async () => {
```
hdiniz: I though we could only create a single incognito account per individual, is that not the case? This is only documenting existing behavior, right?

hdiniz (approved):

test/server/graphql/v1/mutation.test.js
kewitz: We do, but apparently just due to a frontend implementation and not due to the old V1 mutation.
Either way, I'll remove this tests due to the dedicated V2 implemented in this PR.

PR #11713, opened 2026-04-30
Disconnect user PayPal and Wise ConnectedAccounts when removing member

kewitz:


server/graphql/v2/mutation/MemberMutations.ts
```diff
@@ -401,6 +402,15 @@ const memberMutations = {
...
               });
+              // Delete connected accounts for the user on this collective (ex: Wise accounts)
+              await models.ConnectedAccount.destroy({
+                where: {
+                  CreatedByUserId: users.map(user => user.id),
+                  CollectiveId: account.id,
+                  service: [Service.TRANSFERWISE, Service.PAYPAL],
+                },
+                transaction,
+              });
```
Betree, line 413: I don't really get this: if Lauren connects OSC to PayPal, then later gets removed by an admin (e.g. because she switched her job with someone else), why would the PayPal account be disconnected on OSC?

PR #11727, opened 2026-05-07
feat: Adapt features for private organizations

### LINKED ISSUE #8739: Private organizations: Hide/disable unsupported features
Some features are not relevant for private accounts. They'll need to be gated at the API level (feature status: UNSUPPORTED), making sure that the frontend is properly checking the feature status where relevant:
- [ ] updates
- [ ] contribution flow
- [ ] tiers
- [ ] goal
- [ ] profile page personalization
- [ ] contribution policies
- [ ] conversations
- [ ] ability to switch host
- [ ] cross-host expenses
- [ ] cross-host added funds
- [ ] adding funds from public profiles

Similarly, some features need to be adapted:
- [ ] Edit team: only ADMIN or ACCOUNTANT roles should be supported.

Betree:
Resolve https://github.com/opencollective/opencollective/issues/8739


server/models/Collective.ts
```diff
@@ -2514,7 +2516,12 @@ class Collective extends ModelWithPublicId<
     }
 
     // If we can't automatically approve the collective and it is not open to new applications, reject it
-    if (!shouldAutomaticallyApprove && !hostCollective.canApply()) {
+    if (
+      !shouldAutomaticallyApprove &&
+      !hostCollective.canApply() &&
+      // Exception to the canApply rule: hosting is enabled, and remote user is an admin of the host
+      !(hostCollective.hasHosting && creatorUser.isAdmin(hostCollective.id))
+    ) {
```
kewitz, line 2524: I'm OK with blocking the creating of a HostApplication when `host.isPrivate` is true, but I think this method here should only make sure that collective inherits `isPrivate` from the host.

I'm OK with shipping this is if you're planning on change that in a future PR.

kewitz (approved):

server/lib/allowed-features.ts
```diff
@@ -154,28 +176,33 @@ const FeaturesAccess: Partial<
     ],
   },
   [FEATURE.RECEIVE_HOST_APPLICATIONS]: {
+    onlyPublicAccounts: true,
     onlyAllowedFor: FEATURE_ACCESS_PARTY.HOSTS,
     optIn: true,
     flagOverride: 'settings.apply',
   },
```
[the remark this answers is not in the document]
Betree: Can't change `isPrivate` status.

server/models/Collective.ts
Betree: This is already enforced in multiple places: the `Collectives` models's `beforeCreate` + `createCollective` mutation. I'll add it here as a double -safety.

PR #11729, opened 2026-05-08
feat: Cancel & Refund contributions

### LINKED ISSUE #8748: Cancel & Refund Contributions: API mutations for host admin cancel, refund, and contributor block
## Context

Part of [#8747](https://github.com/opencollective/opencollective/issues/8747).

Host admins need backend support to cancel recurring contributions, refund one or more charges (from one-time or recurring contributions), and block a contributor from making future contributions to any collective under the host.

## Implementation

- Add (or extend) a GraphQL mutation allowing a host admin to **cancel a recurring contribution** (order/subscription) on behalf of the contributor.
- Add (or extend) a GraphQL mutation allowing a host admin to **refund one or more transactions** tied to a contribution.
- Add a mechanism for a host admin to **block a contributor**: prevent the contributor from making new contributions to collectives under that host. Define the block model (e.g. a flag on the member/order record or a dedicated block table) and enforce it at contribution creation time.
- All three actions must be **restricted to host admins** of the relevant host.
- Add appropriate **activities/audit log entries** for each action.

## Acceptance criteria

- A host admin can cancel a recurring contribution via the API; the subscription is stopped and the contributor is notified.
- A host admin can refund one or multiple past charges (individual transactions) for a given contribution.
- A host admin can block a contributor; subsequent contribution attempts from that contributor to the host's collectives are rejected.
- Non-host-admins receive a permission error when attempting these mutations.
- Each action is recorded in the activity log.

gustavlrsn:
Resolve https://github.com/opencollective/opencollective/issues/8748 https://github.com/opencollective/opencollective/issues/6877


server/constants/activities.ts
huangyoufeng76-debug, line 204: 0900282271

huangyoufeng76-debug: > 0900282271

server/graphql/v2/mutation/OrderMutations.ts
```diff
@@ -305,44 +314,125 @@ const orderMutations = {
...
+        : undefined;
+
       await models.Activity.create({
         type: activities.SUBSCRIPTION_CANCELED,
         CollectiveId: order.CollectiveId,
         FromCollectiveId: order.FromCollectiveId,
         HostCollectiveId: order.collective.HostCollectiveId,
         OrderId: order.id,
-        UserId: order.CreatedByUserId,
+        UserId: req.remoteUser.id,
```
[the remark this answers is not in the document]
gustavlrsn: Cancellation of pending one-time contributions is done through `processPendingOrder`

server/graphql/common/orders.ts
```diff
@@ -4,9 +4,12 @@ import { InferCreationAttributes } from 'sequelize';
 
 import { CollectiveType } from '../../constants/collectives';
 import status from '../../constants/order-status';
+import { TransactionKind } from '../../constants/transaction-kind';
+import { TransactionTypes } from '../../constants/transactions';
 import { purgeCacheForCollective } from '../../lib/cache';
 import { roundCentsAmount } from '../../lib/currency';
 import { executeOrder } from '../../lib/payments';
```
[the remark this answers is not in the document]
gustavlrsn: This is existing behavior

server/graphql/schemaV2.graphql
```diff
@@ -25097,6 +25116,16 @@ type Mutation {
     Category for cancelling subscription
     """
     reasonCode: String
+
+    """
+    Remove the contributor from the collective public profile. Host admins only.
+    """
+    removeAsContributor: Boolean = false
```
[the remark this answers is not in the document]
gustavlrsn: This is currently not about blocking

server/constants/activities.ts
```diff
@@ -99,6 +99,8 @@ enum ActivityTypes {
   HOST_APPLICATION_COMMENT_CREATED = 'host.application.comment.created',
   // Contributions
   CONTRIBUTION_REJECTED = 'contribution.rejected',
+  CONTRIBUTION_REFUNDED = 'contribution.refunded',
```
Betree, line 102: We should mark https://github.com/opencollective/opencollective/issues/6877 as resolved too 👍

server/graphql/common/orders.ts
```diff
@@ -276,6 +283,69 @@ export const canSetOrderTags = async (req: express.Request, order: Order): Promi
...
 
+const HOST_CANCELLABLE_ORDER_STATUSES = [status.ACTIVE, status.ERROR, status.PAUSED, status.PROCESSING];
+
+/**
+ * Whether the current user can cancel this order as a host admin from the host dashboard.
+ *
+ * Requires the order to be recurring (has a Subscription) and in a status where
+ * cancellation still makes sense.
+ */
+export const canHostCancelOrder = async (req: express.Request, order: Order): Promise<boolean> => {
```
Betree: Why embed the "Host" in these function names, rather than simply say `canCancelOrder`/`canRefundOrder`/`canRemoveContributorFromOrder`? As I see it, the role (host admin) is part of the permission checks, and it might evolve in the future (e.g. allow collective admins to reject)

server/graphql/common/orders.ts
```diff
@@ -276,6 +283,69 @@ export const canSetOrderTags = async (req: express.Request, order: Order): Promi
...
+ * from the host dashboard.
+ */
+export const canHostRefundOrder = async (req: express.Request, order: Order): Promise<boolean> => {
+  if (!(await isOrderHostAdmin(req, order))) {
+    return false;
+  }
+  // If the order was fully refunded there's nothing left to refund
+  if (order.status === status.REFUNDED) {
+    return false;
+  }
```
Betree: I don't think we can rely on this status, see https://github.com/opencollective/opencollective/issues/8233. I suggest removing the check.
```suggestion
```

server/graphql/common/orders.ts
```diff
@@ -276,6 +283,69 @@ export const canSetOrderTags = async (req: express.Request, order: Order): Promi
...
+  const refundableCount = await models.Transaction.count({
+    where: {
+      OrderId: order.id,
+      type: TransactionTypes.CREDIT,
+      kind: [TransactionKind.ADDED_FUNDS, TransactionKind.BALANCE_TRANSFER, TransactionKind.CONTRIBUTION],
+      isRefund: false,
+      isDisputed: false,
+      RefundTransactionId: null,
+    },
+  });
```
Betree: ⚠️ This will be fetched in lists, so it needs to be backed with a GraphQL loader

server/graphql/common/orders.ts
```diff
@@ -276,6 +283,69 @@ export const canSetOrderTags = async (req: express.Request, order: Order): Promi
...
+  if (!(await isOrderHostAdmin(req, order))) {
+    return false;
+  }
+  const backerCount = await models.Member.count({
+    where: {
+      CollectiveId: order.CollectiveId,
+      MemberCollectiveId: order.FromCollectiveId,
+      role: 'BACKER',
+    },
+  });
```
Betree: Same here, needs a loader

server/graphql/v2/mutation/OrderMutations.ts
```diff
@@ -295,8 +304,13 @@ const orderMutations = {
...
           { model: models.Subscription },
-          { model: models.Collective, as: 'collective' },
-          { model: models.Collective, as: 'fromCollective' },
+          // `paranoid: false` so we still load soft-deleted collectives. The
+          // safeguards in `deleteCollective` normally prevent orphan orders,
+          // but root scripts / manual ops can leave recurring orders pointing
+          // at a soft-deleted collective; without this we crash later when
+          // reading `order.collective.HostCollectiveId`.
+          { model: models.Collective, as: 'collective', paranoid: false },
+          { model: models.Collective, as: 'fromCollective', paranoid: false },
```
Betree: I don't get in which case this could happen? Do you have an example?

server/graphql/v2/object/OrderPermissions.ts
```diff
@@ -113,6 +113,30 @@ const GraphQLOrderPermissions = new GraphQLObjectType({
         return req.remoteUser.isAdmin(order.FromCollectiveId);
       },
     },
+    canHostCancel: {
```
Betree: Same on the GraphQL object, I don't see why we would mention the host here.

```suggestion
    canCancel: {
```

server/lib/notifications/email.ts
```diff
@@ -259,7 +259,26 @@ export const notifyByEmail = async (activity: Activity) => {
...
+      // this one to avoid sending the contributor multiple emails for what they
+      // perceive as a single action.
+      if (activity.data?.hostAction?.refund) {
+        break;
+      }
+      if (activity.data?.messageSource === 'HOST' && activity.data?.fromCollective?.id) {
+        await notify.collective(activity, { collectiveId: activity.data.fromCollective.id });
+      } else {
+        await notify.user(activity);
+      }
```
Betree: Can we add a comment for this? I don't really understand why we're adding this parameter in the first `if`

server/graphql/v2/mutation/OrderMutations.ts
gustavlrsn: That'd be manual SQL operations, but reverting this as we are already preventing orphan orders through the root mutations. AI review suggestion that seemed reasonable at first glance..

server/graphql/common/orders.ts
gustavlrsn: Good catch, I refactored them to be general permissions, and moved the existing validation logic (that checks if you are a root admin or the contributor) into the same thing. 

Used the pattern from the expense permissions with `options = { throw: false }` to make it throw specific error messages when used in the mutation.

server/graphql/common/orders.ts
gustavlrsn: Removed this permission altogether, canHostRefund was a leftover from when you could refund multiple transactions on an Order

server/graphql/common/orders.ts
gustavlrsn: Removed the counting of backer memberships altogether and just checking if you're a host admin, seems unnecessary as it's only an additional option on the general `cancelOrder` mutation.

server/lib/notifications/email.ts
gustavlrsn: Removed the `if/else` here to just send this:

```js
await notify.collective(activity, { collectiveId: activity.FromCollectiveId });
```

Noting that, prior to this PR, `notify.user` worked even when a root admin did the cancellation because the activity was created with `UserId: order.fromCollective.id`, but fixed that to be properly assigned to the remote user.

PR #11741, opened 2026-05-13
Honor per-collective platform tip override

znarf:
- Apply `Collective.data.platformTips` override at payment time (was only used by the UI resolver)
- Extract shared resolution into `Collective.hasPlatformTips`
- Events and projects inherit the `data.platformTips` override from their parent collective
- Exclude current platform collectives from accepting tips (checked on self and on host)
- New `OrderContextInput.platformTipOffered` field: when explicitly `false`, persist the order as not eligible (used by the OSC platform tip A/B)

### DIFF
--- a/server/models/Collective.ts
+++ b/server/models/Collective.ts
@@ -3653,6 +3653,45 @@ class Collective extends ModelWithPublicId<
     return plan;
   };
 
+  hasPlatformTips = async function ({ loaders = null } = {}): Promise<boolean> {
+    if (!isNil(this.data?.platformTips)) {
+      return Boolean(this.data.platformTips);
+    }
+    if (this.ParentCollectiveId) {
+      const parent = loaders
+        ? await loaders.Collective.byId.load(this.ParentCollectiveId)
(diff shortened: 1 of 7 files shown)


server/models/Collective.ts
```diff
@@ -3653,6 +3653,34 @@ class Collective extends ModelWithPublicId<
     return plan;
   };
 
+  hasPlatformTips = async function ({ loaders = null } = {}): Promise<boolean> {
+    if (!isNil(this.data?.platformTips)) {
+      return Boolean(this.data.platformTips);
+    }
+    if (PlatformConstants.AllPlatformCollectiveIds.includes(this.id)) {
```
Betree: Not new, but good opportunity to fix: `AllPlatformCollectiveIds` includes the legacy `OCI` profile, a better candidate would be:

```suggestion
    if (PlatformConstants.CurrentPlatformCollectiveIds.includes(this.id)) {
```

Betree (approved):

Betree:
Great simplification!

PR #11742, opened 2026-05-15
New platform tips ledger behind feature flag

znarf:
Per-host opt-in flag (`settings.newPlatformTipsLedger`) that restructures how platform tips are recorded, so hosts see the tips they're holding on their own books between collection and settlement.

## New ledger flow (recording)

- `PLATFORM_TIP` credits route to a per-host-scoped `oc-platform` Vendor (a single global vendor account, scoped to the host via `HostCollectiveId` on the transactions) so hosts see held tips on their own books.
- Stripe application-fee tips emit a paired `APPLICATION_FEE` transfer to OFi Technology, making the previously implicit platform transfer explicit.
- The `OWED → INVOICED → SETTLED` lifecycle lives on the `PLATFORM_TIP` credit via `TransactionSettlement` (kind = `PLATFORM_TIP`) — there is no per-tip `PLATFORM_TIP_DEBT`.
- `accountsCollection.includePlatformVendors` surfaces the global `oc-platform` vendor alongside host-filtered accounts in the dashboard accounts tool.

## Refunds

- **Application-fee tips:** refunding a tip reverses its `APPLICATION_FEE` pair, so the vendor slice can't be left permanently negative.
- **Post-invoice refunds:** a tip refunded after it was `INVOICED`/`SETTLED` records an `OWED` settlement on its refund pair; the resulting negative `PLATFORM_TIP` `DEBIT` on the vendor nets against the next settlement instead of leaving the prior settlement stuck `OWED`. If refund deductions exceed the tips held in a period, the rows roll forward (a negative release transfer is not supported) and net against a later run.

## Settlement

- **Driven by the vendor ledger, not the flag:** the monthly settlement cron decides what to settle from the `oc-platform` vendor's `OWED` `PLATFORM_TIP` rows, *not* from the host's `newPlatformTipsLedger` flag. The flag only routes tips at collection time and can be toggled mid-month, so tips collected while it was on still settle correctly after a host opts out.
- **Release transfer:** a single `PLATFORM_TIP_TRANSFER` pair releases the held funds from the vendor onto the host's books so the host can pay the settlement expense; the `SETTLEMENT` expense is then created and the source rows flip `OWED → INVOICED`.
- **Host-currency sizing & rounding:** `PLATFORM_TIP` rows are summed in the host's currency, converting any row whose `hostCurrency` differs (e.g. legacy USD rows re-pointed by the conversion script), with per-currency rounding of converted amounts so non-USD (EUR/GBP) hosts are billed the correct cent amount.
- **Atomic, deferred release:** the transfer is deferred until the run passes the `MIN_AMOUNT_USD` guard and is written in the same DB transaction as the expense creation and `markTransactionsAsInvoiced`. It commits only when the source rows flip `OWED → INVOICED`, so a below-threshold/skipped run (or partial failure) can't release funds while leaving rows `OWED` and double-release next run.
- **Host-scoped CSV report:** new-ledger `PLATFORM_TIP` rows live on the `oc-platform` vendor, so the settlement expense's attached CSV uses the host-scoped `hostTransactions` report (the account-scoped report can't see them). Legacy `*_DEBT` rows carry `HostCollectiveId = host` too, so nothing is lost.
- Hardened cron guards (e.g. early return on empty transaction lists to avoid no-op `WHERE 0 = 1` updates; skip with a loud warning when `HOST_FEE_SHARE_DEBT` overlaps the new platform-subscription billing).

## Vendor guards

- The global `oc-platform` vendor (and other platform-owned vendors with no parent host) can no longer be edited or deleted through the host vendor mutations — host admins get a clean `Unauthorized` instead of mutating a platform-managed account.

## Migrations

- `create-oc-platform-vendor`: idempotent vendor insert (`ON CONFLICT DO NOTHING`) so a rerun after the enum add is retry-safe.
- `add-platform-tip-transfer-kind`: adds the `PLATFORM_TIP_TRANSFER` transaction kind.
- `exclude-platform-tip-transfer-from-contributors-index`: keeps transfers out of the contributors index.

## Tooling

- `scripts/fixes/convert-platform-tips-to-new-ledger.ts`: converts a host's legacy-format platform tips to the new ledger for a given period. Re-points the `PLATFORM_TIP` pair onto the `oc-platform` vendor, drops the per-tip debt (re-keying its settlement to `PLATFORM_TIP`), and adds the `APPLICATION_FEE` pair for application-fee tips. Every touched row is stamped with `data.migration` + a `platformTipsLedgerConversion` block (pre-conversion values) for auditability/rollback. Balance-preserving and idempotent. Guards: rejects `FROM` before `2024-10-01` (earlier tips used the previous platform account), validates legacy group cardinality before mutating, only converts `OWED` (or application-fee) tips, and skips refunded tips.

## Tests

New/updated coverage in `host-settlement` (non-USD currency conversion, per-currency rounding, below-threshold skip with no transfer, post-invoice refund deduction on converted hosts, single-settlement-expense assertions), the `Transaction` model (PayPal + Stripe application-fee paths and refund reversal under the flag), `payments`, and the vendor edit/delete guards.


migrations/20260513120000-create-oc-platform-vendor.js
Betree, line 1: Haven't tested, but we'll need to make sure that this vendor is exposed properly in the frontend; and in particular that host admins cannot edit/delete it.

migrations/20260513120000-create-oc-platform-vendor.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.query(`
+      ALTER TYPE "enum_Transactions_kind"
+      ADD VALUE IF NOT EXISTS 'APPLICATION_FEE' AFTER 'ADDED_FUNDS'
```
Betree: 🍨  Maybe we should add them to `getTransactionKindPriorityCase` too?

cron/monthly/host-settlement.ts
```diff
@@ -25,6 +27,35 @@ import { runCronJob } from '../utils';
...
+      continue;
+    }
+    const fromCurrency: SupportedCurrency = t.hostCurrency || targetCurrency;
+    if (fromCurrency === targetCurrency) {
+      total += value;
+    } else {
+      if (fxRates[fromCurrency] === undefined) {
+        fxRates[fromCurrency] = await getFxRate(fromCurrency, targetCurrency);
+      }
+      total += Math.round(value * fxRates[fromCurrency]);
```
Betree: I'm trying to generalize this helper in the API as it solves some issues around 0-decimal currencies

```suggestion
      total += roundCentsAmount(value * fxRates[fromCurrency], targetCurrency);
```

cron/monthly/host-settlement.ts
```diff
@@ -195,6 +226,73 @@ export async function run(baseDate: Date | moment.Moment = defaultDate): Promise
...
+    // markTransactionsAsInvoiced flow flip the PLATFORM_TIP TS rows OWED -> INVOICED.
+    // The transfer itself is a pure balance-mover with no TS row of its own.
+    //
+    // We only stage the transfer here; it is written further down, alongside the settlement
+    // expense, so it stays in lockstep with markTransactionsAsInvoiced. Writing it here would
+    // leak a transfer on any later `continue` (below-threshold or the HOST_FEE_SHARE anomaly skip):
+    // the source TransactionSettlement rows would stay OWED, and the next run would transfer the
+    // same tips again, inflating the host and OC Platform vendor balances.
+    let platformTipsTransfer: { ocPlatformVendor: Collective; amountInHostCurrency: number } | null = null;
+    if (host.hasNewPlatformTipsLedger?.() && (!KIND || KIND === PLATFORM_TIP_DEBT)) {
```
Betree: The script will maybe not behave gracefully when we switch that flag on or off in the middle of the month. Maybe we should directly look at the vendor transactions and decide based on that.

Betree (commented):
No blocker for me so far.

I'm not 100% sure about the `PLATFORM_TIP_TRANSFER` behavior, but I have no proposal at this state.

I'll have a second look next week.

migrations/20260513120000-create-oc-platform-vendor.js
znarf: I'm delaying that for now. But that's a follow up I have in mind.

PR #11766, opened 2026-05-22
chore(Subscriptions): misc enhancements and checks for subscriptions

Betree:
Related to the support case at https://oficonsortium.slack.com/archives/C06K5J91675/p1778963592478779

### DIFF
--- a/server/graphql/v2/mutation/OrderMutations.ts
+++ b/server/graphql/v2/mutation/OrderMutations.ts
@@ -324,6 +324,8 @@ const orderMutations = {
 
       if (args.removeAsContributor && !(await OrdersLib.canRemoveContributorFromOrder(req, order))) {
         throw new Forbidden("You don't have permission to remove contributor from the collective");
+      } else if (!order.Subscription) {
+        throw new Error('This contribution is not a recurring contribution');
       }
 
       const host = order.collective.HostCollectiveId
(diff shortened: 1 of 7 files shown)


server/graphql/v2/mutation/OrderMutations.ts
```diff
@@ -316,12 +318,18 @@ const orderMutations = {
...
+
+      // Cancel the subscription (calls remove for externally managed like PayPal)
+      if (order.Subscription.isActive) {
         await order.Subscription.deactivate();
       }
 
+      // Mark the order as cancelled
+      await order.update({ status: OrderStatuses.CANCELLED, data: { ...order.data, previousStatus } });
+
+      // Record the activity
```
Betree: Updating the order after deactivating the subscription to make sure that we don't mark the order as cancelled if for any reason the subscription deactivation fails. For example, if PayPal request returns an error.

server/paymentProviders/paypal/webhook.ts
```diff
@@ -531,12 +533,7 @@ const getEventHandler = (eventType: string | WatchedPaypalWebhookEvent): ((req:
  */
 async function webhook(req: Request): Promise<void> {
   debug('new event', req.body);
-
-  if (!req.params?.hostId) {
-    // Received on legacy webhook
-    // 2026-04-29: This has normally been the default for a white, adding a global log to get a final confirmation.
-    logger.warn('Please update PayPal webhooks to latest version using scripts/paypal/webhooks.ts');
-  }
```
Betree, line 539: Confirmed in production we have no more cases for this endpoint being used without a host ID.

PR #11771, opened 2026-05-26
Abstract isIncognito permission checks

kewitz:


server/graphql/v2/object/Order.ts
```diff
@@ -450,23 +451,20 @@ export const GraphQLOrder = new GraphQLObjectType({
...
+            const hostCollectiveId = toCollective instanceof Error ? null : toCollective?.HostCollectiveId;
             if (
-              collective &&
+              userCollective &&
+              !(userCollective instanceof Error) &&
               fromCollective &&
-              'isIncognito' in fromCollective &&
-              (!fromCollective.isIncognito || isHostAdminOrAccountant || userIsAdminOfFromAccount)
+              !(fromCollective instanceof Error) &&
+              canSeeIncognitoProfile(req, fromCollective, hostCollectiveId || null)
```
Betree: `canSeeIncognitoProfile(..., host)` could let you think that remote user is truely checked by the function against the host, ie to make sure that there's a contribution there, while it's the parent responsibility to do so. That's a dangerous pattern IMO. 

I would rather

```suggestion
              canSeeIncognitoProfile(req, fromCollective) || req.remoteUser.hasRole([roles.ACCOUNTANT, roles.ADMIN], hostCollectiveId)
```

PR #11785, opened 2026-05-27
Support platformTipAmount in updateOrder

znarf:
Adds a `platformTipAmount` argument to the `updateOrder` mutation so contributors can adjust the platform tip on a recurring contribution order. Two-state semantics: undefined and null keeps the current tip, a value - including zero - sets it.

Tax is correctly recomputed when only the tip changes (previously gated solely on totalAmount changing). The `amount` argument keeps its existing meaning (post-tax total excluding the tip).

### DIFF
--- a/server/graphql/schemaV2.graphql
+++ b/server/graphql/schemaV2.graphql
@@ -25701,9 +25701,14 @@ type Mutation {
     tier: TierReferenceInput
 
     """
-    An Amount to update the order to
+    New amount to charge the contributor, excluding the platform tip. Includes taxes if any.
     """
     amount: AmountInput
+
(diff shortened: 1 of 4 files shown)


Betree (commented):
`null` and `undefined` are often mixed in GraphQL (since the GraphQL spec does not have an official "undefined" value). Which means that if your mutation has this parameter but you forgot to pass a value, it will clear the tip. I'd rather have people explicitly pass `platformTipAmount=0` to clear out the tip, makes it explicit.

znarf:
@Betree agree with you, I iterated right when reading again the PR description.

PR #11791, opened 2026-05-28
feat(PrivateOrganizations): adapt search

### LINKED ISSUE #8734: Private organizations: Search, indexing, and discovery
Remove private accounts from public SQL search, OpenSearch, public pickers, public link generation, and any discovery-oriented UI. Admins of the private organization should still be able to find and access it through authenticated dashboard and API flows.

Proper tests need to be implemented.

Betree:
Resolve https://github.com/opencollective/opencollective/issues/8734

1. Adapt SQL search
2. Exclude in OpenSearch (see https://github.com/opencollective/opencollective/issues/8816)

### DIFF
--- a/server/lib/sql-search.ts
+++ b/server/lib/sql-search.ts
@@ -6,6 +6,7 @@ import assert from 'assert';
 
 import config from 'config';
 import type { Request } from 'express';
+import express from 'express';
 import { SelectQueryBuilder } from 'kysely';
 import slugify from 'limax';
 import { get, isEmpty, isNil, isUndefined, toString, words } from 'lodash';
@@ -35,7 +36,7 @@ import RateLimit, { ONE_HOUR_IN_SECONDS } from './rate-limit';
 import { removeDiacritics } from './string-utils';
 
 // Returned when there's no result for a search
-const EMPTY_SEARCH_RESULT = [[], 0];
+const EMPTY_SEARCH_RESULT = [[], 0] as const;
 
 const CONSOLIDATED_BALANCE_SUBQUERY = makeConsolidatedBalanceSubquery('c');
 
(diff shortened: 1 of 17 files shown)


server/lib/sql-search.ts
```diff
@@ -35,6 +39,52 @@ const EMPTY_SEARCH_RESULT = [[], 0];
...
+        c."type" = '${CollectiveType.ORGANIZATION}'
+        AND c."hasHosting" IS TRUE
+        AND EXISTS (
+          SELECT 1 FROM "Collectives" hosted
+          WHERE hosted."HostCollectiveId" = c.id
+          AND hosted."deletedAt" IS NULL
+          AND hosted."approvedAt" IS NOT NULL
+          AND hosted.id IN (:privilegedCollectiveIds)
+        )
+      )
```
kewitz, line 87:
```suggestion
      c."isPrivate" IS FALSE
      -- User is admin of private collective
      OR c.id IN (:privilegedCollectiveIds)
      -- User is fiscal-host admin of private collective
      OR c."HostCollectiveId" IN (:privilegedCollectiveIds)
      -- User is admin of private collective who owns this event/project
      OR c."ParentCollectiveId" IN (:privilegedCollectiveIds)
      -- User is admin of private collective hosted by this organization
      OR (
        c."type" = '${CollectiveType.ORGANIZATION}'
        AND c."hasHosting" IS TRUE
        AND EXISTS (
          SELECT 1 FROM "Collectives" hosted
          WHERE hosted."HostCollectiveId" = c.id
          AND hosted."deletedAt" IS NULL
          AND hosted."approvedAt" IS NOT NULL
          AND hosted.id IN (:privilegedCollectiveIds)
        )
      )
```

server/lib/sql-search.ts
```diff
@@ -309,14 +360,14 @@ export const searchCollectivesInDB = async (
     const collective = await models.Collective.findOne({
       where: { publicId: term },
     });
-    if (collective) {
+    if (collective && (!collective.isPrivate || (args.req && (await canSeePrivateAccount(args.req, collective))))) {
```
kewitz: Perhaps we should make `req` required to avoid an unsafe call that can eventually expose a private account by their publicId.

Betree: Good call

server/lib/sql-search.ts
```diff
@@ -499,6 +554,7 @@ export const searchCollectivesInDB = async (
         lastTransactionFrom: args.lastTransactionFrom,
         lastTransactionTo: args.lastTransactionTo,
         vendorVisibleToAccountIds,
+        privilegedCollectiveIds: privateAccountVisibility.privilegedCollectiveIds,
       },
     },
   );
```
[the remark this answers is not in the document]
Betree: We have this pattern for other args (vendorVisibleToAccountIds), it is not crashing

PR #11812, opened 2026-06-03
debt(GraphQLV1): remove deprecated fields, adapt for new features

Betree:


migrations/20260603120000-drop-orders-private-message.ts
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+import type { QueryInterface } from 'sequelize';
+import { DataTypes } from 'sequelize';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface: QueryInterface) {
+    await queryInterface.removeColumn('Orders', 'privateMessage');
```
Betree, line 9: Unused since 2017, only a few entries in DB with unimportant content. We're keeping the Histories table column just in case.

PR #11813, opened 2026-06-03
Fix ConnectedAccount resolvers access

kewitz:
Deprecating and protecting settings.

### DIFF
--- a/server/graphql/v2/object/ConnectedAccount.ts
+++ b/server/graphql/v2/object/ConnectedAccount.ts
@@ -46,16 +46,9 @@ export const GraphQLConnectedAccount = new GraphQLObjectType<ConnectedAccount, E
     settings: {
       type: GraphQLJSON,
       async resolve(connectedAccount, _, req) {
-        if (connectedAccount.service === 'persona') {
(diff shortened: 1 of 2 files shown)


Betree (commented):
Are we sure this is not going to break anything?

kewitz:
@Betree only Mirror Accounts is being accessed through the API V2 in the ConnectedAccountsTable component.

server/graphql/v2/object/ConnectedAccount.ts
```diff
@@ -46,16 +46,9 @@ export const GraphQLConnectedAccount = new GraphQLObjectType<ConnectedAccount, E
...
-          }
-          return {
-            apiKeyId: connectedAccount.clientId,
-            inquiryTemplateId: connectedAccount.settings.inquiryTemplateId,
-          };
+        if (!req.remoteUser?.isAdmin(connectedAccount.CollectiveId)) {
+          throw new Unauthorized('You need to be logged in as an admin of the account');
         }
-
         return connectedAccount.settings;
```
Betree: Maybe `pick` the settings we're ok to expose?

PR #11814, opened 2026-06-03
debt(GraphQLV1): remove more types and fields

Betree:


kewitz:
The `githubContributors` resolver is still used by contributors-svg worker.

PR #11821, opened 2026-06-05
fix(ExpenseMutations): sanitize items description for drafts

Betree:


kewitz (approved):
Looks good! Should cover all draft cases.

PR #11924, opened 2026-06-25
Platform tips ledger v2: per-host platform accounts

znarf:
Follow-up to #11742 (which introduced the platform tips ledger behind a feature flag). This PR reworks that design around **per-host platform accounts** and ships the settlement flow end to end.

All behaviour is gated behind the `NEW_PLATFORM_TIPS_LEDGER` feature flag (host setting `newPlatformTipsLedger`); hosts that haven't opted in keep the legacy behaviour untouched.

### Per-host platform-tips account
Instead of a single global, host-less `platform-tips` account, **each fiscal host gets its own platform-tips account**:
- A hosted child of the host: `type = PLATFORM`, `ParentCollectiveId = HostCollectiveId = host`, `currency = host.currency`.
- Identified explicitly by `data.isPlatformTipsAccount = true` (not by slug), with a collision-proof slug (`platform-tips` + random suffix, like Events/Users).
- Auto-created on demand via `getOrCreateHostPlatformTipsAccount(host)`.

This makes balances and permissions straightforward: the tips a host collects live on a real, host-scoped account in the host's own currency, and the host's admins administer it (parent-admin = child-admin) without any special-casing.

### Ledger
- Platform tips (`PLATFORM_TIP`) and the new `APPLICATION_FEE` kind are recorded onto the per-host account in the host's currency (no USD round-trip).
- `TransactionSettlement` now carries settlement state on the `PLATFORM_TIP` credit itself, so settlement status survives the per-host model.
- Refunds reverse the application-fee/tip pair correctly.

### Settlement
- `cron/monthly/host-settlement.ts` bills each opted-in host a `SETTLEMENT` expense (payable to OFiTech) for its pending tips, per host, in the host's currency, with the usual threshold/roll-forward behaviour.
- The payee cannot cover the fees on settlement/platform-billing expenses (enforced backend-side in `checkHasBalanceToPayExpense`, and in the frontend companion).

### GraphQL
- New `Platform` object type in v2 (implements `Account`, `AccountWithHost`, `AccountWithParent`) and a matching `Platform` type in v1.
- Platform accounts are sorted last in `childrenAccounts` (they're internal billing accounts).
- `merchantId` resolves correctly for tips/fees depending on which ledger row is read (host-scoped vs platform-scoped), for finance reconciliation against Stripe/PayPal.

### Migration & backfill
- Migration adds the `APPLICATION_FEE` transaction kind.
- `scripts/fixes/convert-platform-tips-to-new-ledger.ts` backfills a host's historical legacy tips onto the new per-host ledger (re-points the credit, re-denominates into host currency, moves the OWED settlement onto the tip credit, drops the old debt row). Idempotent and reversible (stores a rollback marker).

## Notes
- The companion frontend changes (per-host account in the dashboard, restricted actions on the platform account, admin resolution, `Platform` in `possibleTypes`) live in the opencollective-frontend repo.
- Test suites updated for the per-host model; `tsc` clean; core suites (Transaction, payments, host-settlement, convert, Transactions/ExpensesCollectionQuery) green.


cron/monthly/host-settlement.ts
```diff
@@ -132,13 +289,10 @@ export async function run(baseDate: Date | moment.Moment = defaultDate): Promise
...
-          [
-            PlatformConstants.PlatformCollectiveId,
-            PlatformConstants.OfitechCollectiveId,
-            PlatformConstants.OficoCollectiveId,
-          ].filter(Boolean),
-        ),
+        // All platform accounts, including the legacy OC Inc one (8686): the deleted
+        // `getPendingPlatformTips` used to return 0 for platform accounts, guarding OC Inc from
+        // being invoiced for its legacy OWED tips. That guard now lives here.
+        ignoreSettlementForIds: PlatformConstants.AllPlatformCollectiveIds,
```
Betree: "lives here" => where? Isn't this change re-introducing the issue? I don't get why we need this.

server/graphql/v2/interface/Account.ts
```diff
@@ -905,6 +905,10 @@ const accountFieldsDefinition = () => ({
         }
       }
 
+      // Keep platform-owned accounts (e.g. the per-host platform-tips account) at the bottom of the
+      // list, regardless of the requested sort — they are internal billing accounts, not regular children.
+      order = [[Sequelize.literal(`CASE WHEN "Collective"."type" = 'PLATFORM' THEN 1 ELSE 0 END`), 'ASC'], ...order];
+
```
Betree: This seems a bit unexpected. If they're not expected in most UIs, then we should hide them by default (e.g. with a `ignorePlatformAccounts` param set to true by default).

But this just silently breaks the API `order` promise.

Betree (approved):

server/graphql/v2/interface/Account.ts
znarf: I'm not hiding them for now. 

I changed the approach and switched to a default ordering that is based on account type. API order is now respected if it's passed.

cron/monthly/host-settlement.ts
znarf: This is replacing the production-only return 0 for platform accounts in `getPendingPlatformTips`, which the cron no longer calls.

Reviewing this, I made sure to clean all debt related to OC Inc. Its OWED rows were legacy debts, so I marked them all SETTLED in production. With that done, the special case isn't needed at all anymore for this account.

I updated the comment and switched to `CurrentPlatformCollectiveIds`, so OC Inc is invoiceable like any other host.

PR #11938, opened 2026-06-29
Hosted collectives metrics on host collective overview

hdiniz:
Related https://github.com/opencollective/opencollective/issues/8824


migrations/20260618130000-add-hosted-collectives-daily-transaction-sizes-view.ts
```diff
@@ -0,0 +1,118 @@
...
+            WHEN abs(t."amountInHostCurrency") <= 25000   THEN 'GT_200_LTE_250'
+            WHEN abs(t."amountInHostCurrency") <= 50000   THEN 'GT_250_LTE_500'
+            WHEN abs(t."amountInHostCurrency") <= 100000  THEN 'GT_500_LTE_1000'
+            WHEN abs(t."amountInHostCurrency") <= 200000  THEN 'GT_1000_LTE_2000'
+            WHEN abs(t."amountInHostCurrency") <= 500000  THEN 'GT_2000_LTE_5000'
+            WHEN abs(t."amountInHostCurrency") <= 1000000 THEN 'GT_5000_LTE_10000'
+            WHEN abs(t."amountInHostCurrency") <= 2500000 THEN 'GT_10000_LTE_25000'
+            WHEN abs(t."amountInHostCurrency") <= 5000000 THEN 'GT_25000_LTE_50000'
+            ELSE 'GT_50000'
+          END AS "amountBand",
```
kewitz, line 54: Not a big fan of this, it is a bit USD-centric and makes no sense to several other currencies. Perhaps these bins should be defined as a host setting.

migrations/20260618120000-add-counts-and-frequency-to-hosted-collectives-financial-activity.ts
```diff
@@ -0,0 +1,176 @@
...
+          ), 0) AS "amountSpentNet",
+          COUNT(*) AS "transactionCount",
+          COALESCE(COUNT(*) FILTER (
+            WHERE t.type = 'CREDIT'
+              AND NOT t."isRefund" AND t."RefundTransactionId" IS NULL AND NOT t."isInternal"
+          ), 0) AS "contributionsCount",
+          COALESCE(COUNT(*) FILTER (
+            WHERE t.type = 'DEBIT' AND t.kind <> ALL('{HOST_FEE,PAYMENT_PROCESSOR_FEE}'::"enum_Transactions_kind"[])
+              AND NOT t."isRefund" AND t."RefundTransactionId" IS NULL AND NOT t."isInternal"
+          ), 0) AS "payoutsCount"
```
kewitz, line 66: Not a big fan of all this complexity here. Ideally we would only filter by kind/type and find the information we want. We opted to only sum/count debits and credits grouped by kind in the Community* materialized views to avoid leaking this complexity into the view.

I understand that transactions are much simpler to read from the contributor/vendor side, but still, I would rather consider that all transactions are split; filter by kind and; ignore collectives that have complex balance versions.

kewitz (commented):
Good work.

There's a surge in balance in the APEX staging data that I'm not sure where it is comming from:
Image

server/lib/metrics/graphql/host-metrics.ts
```diff
@@ -34,6 +35,12 @@ const HostMetricsNamespaceType = new GraphQLObjectType({
       description: 'Distinct collectives hosted by this host in a period.',
       bindFromParent: { host: ({ host }) => host.id },
     }),
+    hostedCollectivesTransactionSizes: buildSourceField<HostMetricsParent>({
+      source: HostedCollectivesTransactionSizes,
+      schemaPrefix: 'HostedCollectivesTransactionSizes',
+      description: 'Contribution/payout transactions bucketed by amount band, for size-distribution histograms.',
+      bindFromParent: { host: ({ host }) => host.id },
```
kewitz, line 42: I've been seeing this warning:

> To address this problem (which is not a bug in Apollo Client), either ensure all objects of type HostMetricsNamespace have an ID or a custom merge function, or define a custom merge function for the Host.metrics field, so InMemoryCache can safely merge these objects:

migrations/20260618130000-add-hosted-collectives-daily-transaction-sizes-view.ts
hdiniz: Yeah, I totally agree, I though this could be a compromise to land first before finding a better currency agnostic solution. The issue really is how to pre compute this, if its alright I create an immediate follow up to broaden this to support more logical values for other currencies

migrations/20260618120000-add-counts-and-frequency-to-hosted-collectives-financial-activity.ts
hdiniz: I see your point, I think the tradeoff is that for this project we did intent to restrict or be more opinionated about the data that we are exposing in transactions, in the sense that this is our current definition of `payoutsCounts` and callers dont have to consider whats under the hood. It returns a number that just makes sense according to our current definition. For example, if we change the backing table to be the payment intents, we can still continue to expose the `payoutsCounts` number, payment intent will have different enum and way to process that number but callers keep isolated from the implementation detail.

PR #11958, opened 2026-07-03
debt: Compress some collective.info activity data

Betree:
Related to https://github.com/opencollective/opencollective/issues/8459 / https://github.com/opencollective/opencollective/issues/6910

### DIFF
--- a/server/lib/notifications/email.ts
+++ b/server/lib/notifications/email.ts
@@ -207,11 +207,13 @@ export const notifyByEmail = async (activity: Activity) => {
       await notify.collective(activity);
       break;
 
-    case ActivityTypes.COLLECTIVE_UNHOSTED:
+    case ActivityTypes.COLLECTIVE_UNHOSTED: {
+      const host = activity.data.host.id && (await models.Collective.findByPk(activity.data.host.id));
       await notify.collective(activity, {
-        replyTo: activity.data.host.data?.replyToEmail || 'support@opencollective.com',
+        replyTo: host?.data?.replyToEmail || 'support@opencollective.com',
       });
       break;
+    }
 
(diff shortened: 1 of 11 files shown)


server/models/Collective.ts
```diff
@@ -2630,7 +2629,6 @@ class Collective extends ModelWithPublicId<
             'githubHandle',
             'website',
             'tags',
-            'data',
```
Betree, line 2633: This is the core of the fix: stop storing `data` in info payload. Should improve the situation for many event types.

server/lib/spam.ts
```diff
@@ -942,7 +949,7 @@ export const collectiveSpamCheck = async (collective: Collective, context: strin
     bayes: bayesCheck,
     keywords: Array.from(result.keywords),
     domains: Array.from(result.domains),
-    data: collective.info || (collective as unknown as Record<string, unknown>),
+    data: getSpamReportCollectiveSnapshot(collective),
```
Betree, line 952: A nice to have, to only log a small subset of the data - mostly the parts analyzed by the script.

server/lib/notifications/email.ts
```diff
@@ -207,11 +207,13 @@ export const notifyByEmail = async (activity: Activity) => {
       await notify.collective(activity);
       break;
 
-    case ActivityTypes.COLLECTIVE_UNHOSTED:
+    case ActivityTypes.COLLECTIVE_UNHOSTED: {
+      const host = activity.data.host.id && (await models.Collective.findByPk(activity.data.host.id));
```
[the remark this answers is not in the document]
Betree: We were doing `activity.data.host.data` before - I don't think this will break anything.

PR #11963, opened 2026-07-06
Restrict functionality via root admin actions

hdiniz:
Related https://github.com/opencollective/opencollective/issues/8874


server/graphql/v2/interface/Account.ts
```diff
@@ -312,6 +312,10 @@ const accountFieldsDefinition = () => ({
     type: new GraphQLNonNull(GraphQLBoolean),
     description: 'Whether this account is suspended',
   },
+  isBlockedForUnpaidPlatformBilling: {
```
znarf: We are on the top Account resolver, and it looks super specific. Can we expose that elsewhere?

Can we add that to the `GraphQLPlatformSubscription` maybe?

Or to the Organization with a more neutral name: `billingStatus: ON_HOLD`?

server/models/Collective.ts
```diff
@@ -1487,7 +1487,9 @@ class Collective extends ModelWithPublicId<
    * Returns true if Collective is a host account open to applications.
    */
   canApply = function () {
-    return Boolean(this.hasHosting && !this.isPrivate && this.settings?.apply);
+    return Boolean(
+      this.hasHosting && !this.isPrivate && this.settings?.apply && !this.data?.isBlockedForUnpaidPlatformBilling,
```
kewitz, line 1492: Update `data` schema in this file.

server/graphql/v2/object/PlatformSubscription.ts
```diff
@@ -38,6 +39,14 @@ export const GraphQLPlatformSubscription = new GraphQLObjectType({
     isCurrent: {
       type: new GraphQLNonNull(GraphQLBoolean),
     },
+    isAccountOnHold: {
```
kewitz, line 42: The naming should be consistent with the model layer. Be consistent if we're calling it "on hold" or "blocked".

server/graphql/common/expenses.ts
```diff
@@ -111,22 +112,25 @@ import { checkScope } from './scope-check';
 import { hasProtectedUrlPermission } from './uploaded-file';
 
 const loadHostForExpense = async (req: express.Request, expense: Expense): Promise<Collective | null> => {
+  const loadCollectiveById = (id: number): Promise<Collective | null> =>
```
kewitz, line 115: If this is here solely for test compatibility, remove it and use the test/utils `makeRequest` helper function in tests.

PR #11995, opened 2026-07-20
fix(graphql-v1): enforce private account visibility on allMembers query

kewitz:
### Problem

The GraphQL v1 `allMembers` resolver (`server/graphql/v1/queries.js`) unconditionally filtered out any private collective by hardcoding `isPrivate: false` on its joins. This meant that even users who are legitimately allowed to see a private account (its admins, accountants, or the fiscal host's team) could not load its members or memberships. The resolver was recently annotated as *"Does not work with Private Organizations"* for this reason.

### Solution

Instead of dropping the privacy filter entirely, `allMembers` now **preloads the private collectives the requesting user is allowed to see** and applies the same visibility rules already used by the GraphQL v2 `HasMembers` interface:

- **Root users** see everything.
- **Users with an `ADMIN`/`ACCOUNTANT` role** (`MemberRolesForPrivateAccounts`) can see the private collectives they administer, plus their events/projects (`ParentCollectiveId`) and hosted collectives (`HostCollectiveId`).
- **Everyone else** only sees public collectives.

The resolver has two execution paths, and both were updated:

1. **Sequelize path** – the hardcoded `{ isPrivate: false }` conditions on the `collective` / `memberCollective` includes (and the `memberCond` used for sorting) are replaced by a freshly-built condition object mirroring `HasMembers`.
2. **Raw-query path** (`orderBy: totalDonations` / `balance`) – `getMembersWithTotalDonations` and `getMembersWithBalance` in `server/lib/queries.js` now build their privacy `WHERE` clause dynamically via a shared `getPrivateCollectivesSQLCondition` helper, driven by the preloaded access set passed through `options`.

### Cleanup

While in the area, removed dead code that is no longer used anywhere:

- `Collective.getTopBackers` and `Collective.getTiersWithUsers` (`server/models/Collective.ts`)
- The now-unused `getTopBackers` raw query (`server/lib/queries.js`)
- Their associated tests (`test/server/models/Collective.test.js`)

### DIFF
--- a/server/graphql/v1/queries.js
+++ b/server/graphql/v1/queries.js
@@ -5,6 +5,7 @@ import { isEmail } from 'validator';
 import { roles } from '../../constants';
 import { CollectiveType } from '../../constants/collectives';
 import { PAYMENT_METHOD_SERVICE, PAYMENT_METHOD_TYPE } from '../../constants/paymentMethods';
+import { MemberRolesForPrivateAccounts } from '../../constants/roles';
 import { fetchCollectiveId } from '../../lib/cache';
 import logger from '../../lib/logger';
 import { getConsolidatedInvoicesData } from '../../lib/pdf';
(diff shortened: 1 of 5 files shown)


kewitz:
Related to issue https://github.com/opencollective/opencollective/issues/8732

PR #11996, opened 2026-07-20
fix(graphql-v1): enforce private account visibility on Tier query

kewitz:
## Problem
The v1 `Tier` query (`server/graphql/v1/queries.js`) resolved tiers with a bare `findByPk` and no authorization check. Since `id` is a small sequential integer, anyone calling GraphQL v1 (including unauthenticated callers) could enumerate tier IDs and retrieve tier names, amounts, intervals, descriptions, and the parent collective linkage for tiers belonging to **private organizations**.

The v2 equivalent (`TierQuery`) already enforces this via `assertCanSeeAccount` on the tier's parent collective.

## Fix
The v1 `Tier` resolver now loads the tier's parent collective and calls `assertCanSeeAccount` before returning it, mirroring v2's behavior and the existing v1 `Collective` query pattern in the same file.

## Tests
Added a new describe block in `test/server/graphql/v1/tiers.test.js` covering:
- A random user cannot fetch a tier belonging to a private organization via `Tier(id)`
- Admins of the private collective / private host can still fetch it

All existing tests continue to pass.

### DIFF
--- a/server/graphql/v1/queries.js
+++ b/server/graphql/v1/queries.js
@@ -58,8 +58,13 @@ const queries = {
     args: {
       id: { type: new GraphQLNonNull(GraphQLInt) },
     },
-    resolve(_, args) {
-      return models.Tier.findByPk(args.id);
+    async resolve(_, args, req) {
+      const tier = await models.Tier.findByPk(args.id);
+      if (tier) {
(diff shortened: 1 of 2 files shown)


kewitz:
Related to issue https://github.com/opencollective/opencollective/issues/8732

PR #12012, opened 2026-08-10
Restore githubContributors resolver

kewitz:
Our contributors-svg has been failing to render the contributors banner image due to this missing resolver. 
Removed in https://github.com/opencollective/opencollective-api/pull/11814

### DIFF
--- a/server/graphql/schemaV1.graphql
+++ b/server/graphql/schemaV1.graphql
@@ -39,6 +39,7 @@ interface CollectiveInterface {
   Private instructions related to an event
   """
   privateInstructions: String
+  githubContributors: JSON!
   slug: String
   path: String
   isHost: Boolean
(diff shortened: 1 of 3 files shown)


znarf (changes requested):
Do we really need to restore this on the API side? Looking at contributors-svg, the only reason it broke is that `githubContributors` is still in the query, so the whole query fails GraphQL validation. The service actually fetches contributors directly from GitHub and only uses this field as a fallback when the GitHub fetch takes more than ~10s. And since the resolver just returns `collective.data.githubContributors`, which is stale legacy data from the old GitHub-connect flow, that fallback mostly serves outdated lists anyway.

I'd rather remove the field from the query in contributors-svg (and drop the fallback) than re-add a dead field here. That's a one-line fix on the consumer side and keeps the cleanup from #11814 intact. The only thing contributors-svg genuinely needs from the API is `githubHandle` and `settings` to know which repo to fetch.

kewitz:
@znarf, I'm fine with updating `contributors-svg`.

PR #12014, opened 2026-08-10
Introduce balance accounting category type

hdiniz:
Related https://github.com/opencollective/opencollective/issues/8884


server/graphql/v2/mutation/ConnectedAccountMutations.ts
```diff
@@ -146,6 +151,54 @@ const connectedAccountMutations = {
...
+      if (args.accountingCategory) {
+        const accountingCategory = await fetchAccountingCategoryWithReference(args.accountingCategory, {
+          throwIfMissing: true,
+          loaders: req.loaders,
+        });
+        checkIsValidBalanceAccountingCategory(accountingCategory, collective);
+        accountingCategoryId = accountingCategory.id;
+      }
+
+      return connectedAccount.update({
```
kewitz, line 209: Considering you cannot the same fiscal-host to a particular Wise account, we should add an exception and update every active Wise account.

server/graphql/v2/mutation/AccountingCategoriesMutations.ts
```diff
@@ -82,6 +83,12 @@ export default {
             id = idDecode(input.id, 'accounting-category');
           }
 
+          // Expense types are meaningless for balance-sheet categories
+          const isBalanceSheetKind = ['BALANCE_ACCOUNT', 'CLEARING_ACCOUNT'].includes(input.kind);
+          if (isBalanceSheetKind) {
+            input.expensesTypes = null;
+          }
+
```
kewitz, line 91: We should prevent accounting categories from change from _P&L_ into/from _BALANCE/CLEARING_.

PR #12043, opened 2026-08-31
fix(Expenses): let payee host admins submit cross-host draft expenses

znarf:
Since #10768, cross-host draft expense invites whose payee is a hosted collective are emailed to the payee's **host** admins (and collective admins can no longer read their host's payout methods, a carve-out from #6986/#6991). #11795 and opencollective-frontend#12180 then let those host admins edit the draft and pick a payout method from the recipient host.

But the final submit still required being an admin of the payee itself, so the intended completer was rejected at the last step:

> User needs to be the admin of the payee to submit an expense on their behalf

## Changes

- `submitExpenseDraft`: relax the payee gate to `isAdminOfCollectiveOrHost`. It runs after the existing draft-key/author/payee check, so the submitter has already proven they hold the invite key (or authored the draft).
- Tests: a payee-host admin with the draft key can submit (expense moves to PENDING, payout method created on the host); a random user with the key is still rejected.

Real-world case: OCE Foundation cross-host grants (USD fund to a collective hosted on the EUR host), stuck three times since December.

### DIFF
--- a/server/graphql/common/expenses.ts
+++ b/server/graphql/common/expenses.ts
@@ -2731,8 +2731,12 @@ export async function submitExpenseDraft(
     isNewExpenseFlow: isNewExpenseFlow === true ? true : undefined,
   };
   if (requestedPayee) {
-    if (!req.remoteUser?.isAdminOfCollective(requestedPayee)) {
-      throw new Forbidden('User needs to be the admin of the payee to submit an expense on their behalf');
+    // Also allow admins of the payee's fiscal host (e.g. a host admin completing a cross-host
+    // draft invite on behalf of one of their hosted collectives)
+    if (!req.remoteUser?.isAdminOfCollectiveOrHost(requestedPayee)) {
+      throw new Forbidden(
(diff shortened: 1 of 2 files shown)


server/graphql/common/expenses.ts
```diff
@@ -2731,8 +2731,12 @@ export async function submitExpenseDraft(
...
   };
   if (requestedPayee) {
-    if (!req.remoteUser?.isAdminOfCollective(requestedPayee)) {
-      throw new Forbidden('User needs to be the admin of the payee to submit an expense on their behalf');
+    // Also allow admins of the payee's fiscal host (e.g. a host admin completing a cross-host
+    // draft invite on behalf of one of their hosted collectives)
+    if (!req.remoteUser?.isAdminOfCollectiveOrHost(requestedPayee)) {
+      throw new Forbidden(
+        'User needs to be an admin of the payee (or its fiscal host) to submit an expense on their behalf',
+      );
```
kewitz, line 2739: This makes sense. This feature was simply not implemented with host-to-host scenario in mind.

kewitz (approved):

PR #12063, opened 2026-09-04
Wait for background activity dispatches before resetting the test DB

znarf:
## Problem

Most red `test` / `test-graphql` jobs on `main` are not real failures: 27 of the 31 failed test jobs on `main` since June 6 log a Postgres deadlock (`40P01`) between the `TRUNCATE ... RESTART IDENTITY CASCADE` in `resetTestDB` and an `INSERT INTO "Activities"` (or the notification dispatch reading `Users` / `Notifications`) left running by the previous test. Depending on which side Postgres kills, the symptom is a random test failing in the next suite (`merge-accounts`, `routes/images`, `paypal/api`, `CommunityQuery`, ...) or the job hanging until the 30 minute timeout.

The background work comes from two places:
- the `Activity` `afterCreate` hook, which dispatches notifications without being awaited (deferred to `afterCommit` since #11864);
- `sendEmailNotifications` / `sendOrderFailedEmail` in the order flow, which are called without `await` and create activities themselves.

#11905 added a tracker with `Activity.waitAllDispatch()`, but it is opt-in and only two test files enable it.

## Changes

- `test/setup.js`: enable activity dispatch tracking for the whole mocha run.
- `test/utils.ts`: `resetTestDB` re-enables tracking (some tests disable it on purpose) and awaits pending dispatches before truncating, with a 10s safety timeout that logs a warning instead of hanging.
- `activity-dispatch-tracker.ts`: new `trackBackgroundWork(promise)` that registers a fire-and-forget promise. It attaches no handler to the caller's promise, so a failure still reaches the process-level `unhandledRejection` handler (which reports to Sentry) exactly as before; two tests assert this with tracking disabled and enabled.
- `payments.ts` / `stripe/webhook.ts`: route the four unawaited email helpers in `sendEmailNotifications` and `sendOrderFailedEmail` through `trackBackgroundWork`. Production behavior is unchanged: the calls are still not awaited, they are only observable by the tracker (which is a no-op unless enabled).
- The two tracker tests that assumed tracking is off by default now disable it explicitly.

## Verification

Local runs against Postgres 16:
- Targeted suites (`models/Activity`, cron `71-send-resume-subscription-emails`, `lib/payments`, `stripe/webhook`, `activity-dispatch-tracker`): 162 passing, 0 deadlocks. Before the `trackBackgroundWork` part, the same run still produced one deadlock after the Stripe webhook suite.
- Full unit suite excluding `test/server/graphql/**`: 2536 passing, 0 deadlocks, 0 wait timeouts. The 11 failures are local-environment ones (S3 / MinIO, network, search extensions, Redis mutex) that also fail on `main` locally, plus `merge-accounts` "Merges a user profile", which passes in isolation on this branch and is an order-dependent flake also seen on `main` CI this week.


Betree:
Check https://github.com/opencollective/opencollective-api/pull/11905, the `waitAllActivityDispatches` may be handy here

PR #12081, opened 2026-09-08
Enforce scope in TransactionGroupQuery

kewitz:
Enforce required scopes when querying TransactionGroup using user or personal tokens.


Betree (approved):
Cool. Note that this query is meant to be removed ultimately, Payment Intents provide a better interface for it.

PR #12084, opened 2026-09-09
fix(auth): address authentication issues

Betree:


server/graphql/v2/mutation/GuestMutations.ts
```diff
@@ -62,7 +62,11 @@ const guestMutations = {
         <string>args.emailConfirmationToken,
       );
 
-      const accessToken = user.jwt({}, TOKEN_EXPIRATION_SESSION);
+      const accessToken = await user.generateSessionToken({
+        createActivity: true,
```
Betree:
```suggestion
```

server/graphql/v2/mutation/GuestMutations.ts
```diff
@@ -62,7 +62,11 @@ const guestMutations = {
...
+        updateLastLoginAt: true,
```
Betree:
```suggestion
```

server/graphql/v2/mutation/GuestMutations.ts
```diff
@@ -62,7 +62,11 @@ const guestMutations = {
         <string>args.emailConfirmationToken,
       );
 
-      const accessToken = user.jwt({}, TOKEN_EXPIRATION_SESSION);
+      const accessToken = await user.generateSessionToken({
```
Betree, line 65: We need to check if using a `generateSessionToken` token could have other implications.

server/lib/encryption.ts
```diff
@@ -56,6 +57,20 @@ export const secretbox = {
...
+ * Compare secrets in constant time. Length differences are not leaked via
+ * `crypto.timingSafeEqual` because both sides are hashed first.
+ */
+export function timingSafeEqualString(a: string | null | undefined, b: string | null | undefined): boolean {
+  if (typeof a !== 'string' || typeof b !== 'string') {
+    return false;
+  }
+
+  const left = createHash('sha256').update(a).digest();
+  const right = createHash('sha256').update(b).digest();
```
Betree: Let's use a stronger hashing algorithm here.

server/routes.ts
```diff
@@ -122,7 +123,7 @@ export default async (app: express.Application) => {
       whitelist: function (req: express.Request) {
         const apiKey = req.query.api_key || req.body?.api_key;
         // No limit with internal API Key
-        return apiKey === config.keys.opencollective.apiKey;
+        return typeof apiKey === 'string' && timingSafeEqualString(apiKey, config.keys.opencollective.apiKey);
```
Betree: Let's remove all fixes around using `timingSafeEqualString` to check the API key, it's not a strongly protected secret.

PR #12085, opened 2026-09-09
fix(security): ignore isNewUser on established member invitees

Betree:
## Summary

Security audit of the members invitation process found a **high-severity** issue:

`inviteMember` accepted a client-controlled `isNewUser` flag and wrote `requiresProfileCompletion` onto the **invitee's** account. Any collective admin could invite an established user with `isNewUser: true`. On that user's next `editAccount` call, slug regeneration treated a missing `slug` input as a placeholder and called `generateSlug()`, which sees the current slug as taken and assigns a random suffix — rewriting their public URL.

The bulk `inviteMembers` path was never exposed: `processInviteMembersInput` derives `isNewUser` server-side, setting it only when it just created the user. Only the single `inviteMember` mutation passed the raw GraphQL argument through.

### Reproduction

Against the code on `main`, a user who is admin only of a collective they created themselves can rewrite an unrelated user's public URL:

```
[repro] victim requiresProfileCompletion after attacker invite = true
[repro] victim slug before = ada-lovelace | after = ada-lovelace-4b59857b
```

The victim never requested a slug change; they only edited their profile description. With this branch applied, the same script leaves the flag unset and the slug intact.

### Fixes

1. **`MemberInvitation.invite`**: only set `requiresProfileCompletion` (and the new-user email CTA) when the invitee is an existing `User` row that has never signed in, or that already requires profile completion. A client-supplied `isNewUser` cannot flag an account that has `lastLoginAt`.
2. **`editAccount`**: only auto-rewrite `guest-*` / `user-*` placeholder slugs during profile completion. Established public slugs are left unchanged.
3. Deprecated the untrusted GraphQL `isNewUser` argument.

The legitimate invite-form flow (create a never-signed-in user, then invite with `isNewUser: true`) is unchanged.

## Two test-harness bugs surfaced along the way

Neither affects production; both are included because they made CI red.

**1. 2FA sessions bled between test files.** `resetTestDB()` truncates with `RESTART IDENTITY`, so user ids restart at 1 in every test file, while `sessionCache` lives for the whole mocha run. 2FA sessions are keyed `2fa:${userId}:${sessionId}` with `sessionId` undefined under test, so an entry written by one file satisfied the 2FA check for whichever unrelated user reused that id later. Simply adding users to `AccountMutations.test.ts` shifted the ids and broke `ApplicationMutations`' 2FA tests.

`resetTestDB()` now deletes the `2fa:*` keys. It deliberately does **not** clear the cache: `config/ci.json` sets `redis.serverUrl` but not `serverUrlSession`, so `createRedisClient(SESSION)` falls back to the default instance and in CI both caches are the same Redis database. A flush also dropped the cached FX rates, which moved the converted `PAYMENT_PROCESSOR_COVER` amount in the `createRefundTransaction` multi-currency snapshot from 64 to 68.

**2. `Order` 'retries the lock if it fails' was racy.** It launched both `lock()` calls without awaiting and assumed the `retries: 0` call would always win. When the `retries: 1000` call commits first, the `retries: 0` call rejects with `This order is already been processed, please try again later` at `await firstCall`. `Order.prototype.lock` is pure Postgres and never touches the cache, so this is pre-existing; the test now waits for the first call to hold the lock before launching the retrying one.

## Test plan

- [x] `inviteMember` with `isNewUser: true` on a user who has `lastLoginAt` does **not** set `requiresProfileCompletion` and does not send the new-user email CTA
- [x] `inviteMember` with `isNewUser: true` on a never-signed-in user still flags them (invite-form flow)
- [x] `editAccount` does not rewrite an established slug when `requiresProfileCompletion` is set
- [x] `editAccount` still rewrites `guest-*` placeholder slugs on profile completion
- [x] End-to-end attacker script reproduces the takeover on `main` and is blocked on this branch
- [x] CI's exact `test` command run locally against a Redis matching `config/ci.json`: 2576 passing, multi-currency snapshot green, only minio/network failures left
- [x] `Order.test.js` green over 5 consecutive runs; the losing interleaving reproduced separately to confirm the old race

<div><a href="https://cursor.com/agents/bc-a94b3978-4ccf-4edd-9506-2eb87bce8265?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a94b3978-4ccf-4edd-9506-2eb87bce8265&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/server/graphql/schemaV2.graphql
+++ b/server/graphql/schemaV2.graphql
@@ -27172,9 +27172,12 @@ type Mutation {
     privateNote: String
 
     """
-    When true, the invited user account was just created from the invite form. The invitee will be required to complete their profile before accepting the invitation.
+    When true, the invited user account was just created from the invite form. Ignored if the invitee has already signed in. The invitee will be required to complete their profile before accepting the invitation.
     """
     isNewUser: Boolean = false
+      @deprecated(
+        reason: "2026-09-09: Untrusted client input. The server only flags accounts that have never signed in (or already require profile completion)."
+      )
   ): MemberInvitation!
 
   """
--- a/server/graphql/v2/mutation/AccountMutations.ts
+++ b/server/graphql/v2/mutation/AccountMutations.ts
@@ -742,10 +742,14 @@ const accountMutations = {
 
       if (account.data?.requiresProfileCompletion === true) {
         updateParams.data = Object.assign({}, account.data, updateParams.data, { requiresProfileCompletion: false });
-        // Update slug if either the name is updated or if this is a previous guest account that already had a name
+        // Only auto-rewrite placeholder slugs (guest-* / user-*). `!updateParams.slug` used to
+        // match every established profile, and generateSlug() then treated the current slug as
+        // taken and assigned a random suffix — a public URL takeover if an invitation had
+        // incorrectly set requiresProfileCompletion on an existing user.
         if (
           (updateParams.name || ![DEFAULT_GUEST_NAME, 'Incognito'].includes(account.name)) &&
-          (!updateParams.slug || account.slug.startsWith('guest-') || account.slug.startsWith('user-'))
+          !updateParams.slug &&
+          (account.slug.startsWith('guest-') || account.slug.startsWith('user-'))
         ) {
           updateParams.slug = await Collective.generateSlug((updateParams.name as string) || account.name);
           previousData.slug = account.slug;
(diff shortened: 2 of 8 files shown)


kewitz:
Fixed in https://github.com/opencollective/opencollective-api/pull/12087

PR #12091, opened 2026-09-14
Pull the MinIO service image from quay.io

znarf:
Every `test`, `test-graphql` and e2e job on every branch has failed since 2026-09-12 with:

```
Error response from daemon: pull access denied for minio/minio, repository does not exist or may require 'docker login'
```

Two MinIO distribution channels went away at once:

1. **Docker Hub image.** `minio/minio` is gone (`https://hub.docker.com/v2/repositories/minio/minio/` returns 404). MinIO's own registry has the same tag with the identical manifest digest, so the three service definitions (`ci.yml` x2, `e2e.yml`) now use `quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z@sha256:14cea493…`. Tag and digest unchanged.
2. **`mc` client binary.** `https://dl.min.io/client/mc/release/linux-amd64/mc` now returns HTTP 410 Gone, which failed the e2e "Setup minio bucket" step once the image pulled. The step now uses the AWS CLI preinstalled on the runner against MinIO's S3 API (`head-bucket || create-bucket`, idempotent like `mc mb --ignore-existing`).

Every open Renovate PR needs a rebase or rerun after this lands.


Betree (approved):
Works for me but we should move to maintained minio alternatives

PR #12096, opened 2026-09-14
fix(orders): release the orders limit for Payment Intent orders, rate limit manual contributions

znarf:
Guests of a busy collective were getting "Your card was declined." on every payment method (manual included) while logged-in contributors were fine. No Stripe declines involved: it's the generic error of `checkOrdersLimit`, and the per-collective guest limit was stuck because Payment Intent orders never release it.

- release the orders limit for Payment Intent orders
- rate limit manual contributions per user and per IP (`MANUAL_ORDERS_LIMIT_USER` / `MANUAL_ORDERS_LIMIT_IP`, 10/h each), since successful orders reset the existing limits and manual orders always succeed

`checkOrdersLimit` itself is unchanged.


server/lib/security/limit.ts
```diff
@@ -103,13 +104,13 @@ export const checkOrdersLimit = async (
   // const errorMessage = 'Error while processing your request, please try again or contact support@opencollective.com.';
   const errorMessage = 'Your card was declined.';
 
+  // Rejected attempts are not counted (nor refresh the TTL), otherwise a burst of blocked requests locks the limit forever
   const limits = getOrdersLimit(order, reqIp, reqMask);
-  for (const limit of limits) {
-    const count = (await cache.get(limit.key)) || 0;
+  const counts = await Promise.all(limits.map(async limit => (await cache.get(limit.key)) || 0));
```
Betree: 🍨 This would be a good case for an `MGET` (get multiple values at once)

server/graphql/v1/mutations/orders.js
```diff
@@ -264,6 +269,9 @@ export async function createOrder(order, req) {
   }
 
   await checkOrdersLimit(order, reqIp, reqMask);
+  // Payment Intent orders are confirmed asynchronously: the Stripe webhook releases these once the payment succeeds
+  const isPaymentIntentOrder = order.paymentMethod?.type === PAYMENT_METHOD_TYPE.PAYMENT_INTENT;
+  const ordersLimitKeys = isPaymentIntentOrder ? getOrdersLimitKeys(order, reqIp, reqMask) : undefined;
```
Betree: One good thing about the previous version is that it was also protecting us from a DDos on manually created orders. I'm also afraid we would loose PayPal protections.

Unless I'm missing something, I would rather pass `isPaymentIntentOrder` to `getOrdersLimitKeys` as an extra `needsIncreasedProtection`, and adjust we're keys we're returning based on that.

PR #12100, opened 2026-09-15
feat(orders): rate limit manual contributions per user and per IP

znarf:
Manual contributions (bank transfer and other manual payment providers) don't go through a payment processor, so they can be created in bulk for free: `checkOrdersLimit` releases its counters on every successful order, and manual orders always succeed, so nothing throttles a burst of manual orders on a collective today. Guests don't get a captcha on them either (only on credit card orders).

Adds a dedicated limit for manual contributions, `limits.manualOrdersPerHour` (`MANUAL_ORDERS_LIMIT_USER`, `MANUAL_ORDERS_LIMIT_IP`), 10 per user and 10 per IP per hour by default, checked in `createOrder` right after the existing limits. Unlike `checkOrdersLimit` it is never released on success: it throttles the volume itself. There is deliberately no per-collective key, since a shared counter is what an attacker would use to lock legitimate contributors out of a collective.

The existing `checkOrdersLimit` is untouched. Disabled in test/e2e like the existing limiter, with unit tests stubbing the env.

Follow-up to #12096.


znarf:
Moved into #12096 as a follow-up commit.

PR #12124, opened 2026-09-17
chore(ci): speed up CI

Betree:
Port frontend CI optimizations from https://github.com/opencollective/opencollective-frontend/pull/12402: shared setup action, PR concurrency, tool caches, coverage on main only, E2E stack build once with artifacts, and fix schema drift detection in check-schemas.sh.


Betree:
@cursor rebase this on latest `main`.

PR #12157, opened 2026-09-21
Drop tweetnacl-util in favor of native Buffer/crypto

znarf:
## Summary

Removes the `tweetnacl-util` dependency. It was only used for base64/UTF-8 conversions that `Buffer` provides natively, and `tweetnacl`'s `randomBytes` is a thin wrapper over `crypto.randomBytes`. Both are replaced with the native equivalents.

`tweetnacl` itself stays: it implements XSalsa20-Poly1305 (`secretbox`), which Node's `crypto` module does not provide (Node 24 exposes `aes-*-gcm` and `chacha20-poly1305` only). All tax form PDFs in S3 and `LegalDocuments.data.encryptedFormData` are persisted in that format, so replacing the cipher would require a versioned format plus a re-encryption of existing data.

## Changes

- `server/lib/encryption.ts`: `Nonce` and `generateKey` use `crypto.randomBytes`; base64 key decoding factored into a `decodeKey` helper; `encodeUTF8` replaced by `Buffer.toString('utf8')`. Wire format (`nonce || box`) and base64 key format are unchanged.
- `server/graphql/v2/mutation/LegalDocumentsMutations.ts`: `encodeBase64(new Uint8Array(...))` replaced by `.toString('base64')`.
- `test/server/graphql/v2/mutation/LegalDocumentsMutations.test.ts`: `decodeBase64` replaced by `Buffer.from(s, 'base64')`.
- `package.json` / `package-lock.json`: `tweetnacl-util` removed (no other dependents in the lockfile).

## Verification

- A payload encrypted by the previous code decrypts with the new code via both `secretbox.decrypt` and `secretbox.decryptRaw`.
- `test/server/lib/encryption.test.js`, `LegalDocumentsMutations.test.ts`, `legal-documents.test.ts` and `export-1099.test.ts` pass (45 tests).
- Type check, ESLint and Prettier pass on the changed files.


znarf:
@codex review

